PR #75, opened 2019-10-29
Add a channel-based interface for sink to allow batch processing

suzaku:

With `Emit` we can only process `txn.Txn`s one by one.

1. Add channel based methods into the interface
1. Remove unused methods

The original `Emit` method is still kept so that I don't have to rewrite large part of the old code.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


suzaku:
Replaced by #182

PR #159, opened 2019-12-05
mark processor down for some time not updating resolve ts

july2993:
review after https://github.com/pingcap/ticdc/pull/154

mark processor down for some time not updating resolve ts

mark processor down for some time not updating resolve ts.
and clear up data dispatch tables to other captures laster.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code


cdc/owner.go
```diff
@@ -384,7 +430,26 @@ func (o *ownerImpl) loadChangeFeedInfos(ctx context.Context) error {
...
 
 		if cfInfo, exist = o.changeFeedInfos[changeFeedID]; exist {
+			for cid, pinfo := range etcdChangeFeedInfo {
+				if _, ok := cfInfo.processorLastUpdateTime[cid]; !ok {
+					cfInfo.processorLastUpdateTime[cid] = time.Now()
+					continue
+				}
+
+				oldPinfo, ok := cfInfo.ProcessorInfos[cid]
+				if !ok || oldPinfo.ResolvedTs != pinfo.ResolvedTs {
```
amyangfei: both `ResolvedTs` and `CheckpointTs` change can trigger the update of `last update time`?

july2993: only check ResolvedTS here do you means check it both?

cdc/owner.go
```diff
@@ -34,6 +35,8 @@ import (
 	"go.uber.org/zap"
 )
 
+var markProcessorDownTime = time.Minute
```
amyangfei: It seems the interval that processor updates `resolvedTs` or `checkpointTs` is not always less than 1 minute. Besides shall we allow several times update failure of processor and do we need to increase this value?

cdc/owner.go
amyangfei: `if !ok || oldPinfo.ResolvedTs != pinfo.ResolvedTs || oldPinfo.CheckpointTs != pinfo.CheckpointTs`
like this?

PR #163, opened 2019-12-09
owner: fix checkpoint ts calculate in owner

amyangfei:

When the `StartTs` of a changefeed is smaller than any none-system table DDL, we should have a right way to calculate `global resolved ts` and `global checkpoint ts`

- The `global resolved ts` can use the `ddl resolved ts` value
- Set the `global checkpoint ts` to the local checkpoint cache stored in `ChangeFeedInfo`

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -30,7 +30,6 @@ import (
 	"github.com/pingcap/ticdc/cdc/roles"
 	"github.com/pingcap/ticdc/cdc/roles/storage"
 	"github.com/pingcap/ticdc/cdc/schema"
-	"github.com/pingcap/tidb/store/tikv/oracle"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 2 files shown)


cdc/owner.go
```diff
@@ -494,18 +493,7 @@ func (o *ownerImpl) calcResolvedTs() error {
 		minCheckpointTs := cfInfo.TargetTs
 
 		if len(cfInfo.tables) == 0 {
```
july2993: after this change when there no tables, the resolve ts can be forward and the checkpoint ts can not be forward?

amyangfei: yes, in fact if there is no tables, the `checkpoint ts` is meaningless until cdc meets an add table DDL.

PR #164, opened 2019-12-09
update readme

amyangfei:

Fix `data replication`, add simple deployment procedure

same as above

### Check List 

Tests 

 - No code

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://internal.pingcap.net/idc-jenkins/job/build_cdc_master/badge/icon)](https://internal.pingcap.net/idc-jenkins/job/build_cdc_master/)
 
-**ticdc** is a change data capture for TiDB, it supports to synchronize change data to various downstreams, including MySQL protocol compatible database, message queue via open CDC protocol and other systems such as local file storage.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -25,9 +25,13 @@ Notice ticdc supports building with Go version `Go >= 1.13`
...
 
-## Installing and Deployment
+## Deployment
 
-TODO
+You can setup a CDC cluster for replication test manually as following:
+
+1. setup a TiDB cluster.
+2. start a CDC cluster, which contains one or more CDC servers. The command to start on CDC server is `cdc server --pd-endpoints http://10.0.10.25:2379`, where `http://10.0.10.25:2379` is the client-url of pd-server.
+3. start a replication changefeed by `cdc cli --pd-addr http://10.0.10.25:2379 --start-ts 413105904441098240 --sink-uri root@tcp(127.0.0.1:3306)/test`. The tso is TiDB `timestamp oracle`, if it is not provided, zero will be used. Currently we support MySQL protocol compatible database as downstream sink only, we will add more sink type in the future.
```
july2993: we should use a latest timestamp instead of zero

amyangfei: Does the latest timestamp mean the `timestamp` of now?

july2993: yes,  or fetch a timestamp from pd

amyangfei: fixed, use the current tso

PR #167, opened 2019-12-10
mark processor down for some time not updating resolve ts

july2993:
close the old pr https://github.com/pingcap/ticdc/pull/159 since I can't update it after make repo public.

mark processor down for some time not updating resolve ts

mark processor down for some time not updating resolve ts.
and clear up data dispatch tables to other captures laster.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code


cdc/owner.go
```diff
@@ -384,7 +431,26 @@ func (o *ownerImpl) loadChangeFeedInfos(ctx context.Context) error {
...
 
 		if cfInfo, exist = o.changeFeedInfos[changeFeedID]; exist {
+			for cid, pinfo := range etcdChangeFeedInfo {
+				if _, ok := cfInfo.processorLastUpdateTime[cid]; !ok {
+					cfInfo.processorLastUpdateTime[cid] = time.Now()
+					continue
+				}
+
+				oldPinfo, ok := cfInfo.ProcessorInfos[cid]
+				if !ok || oldPinfo.ResolvedTs != pinfo.ResolvedTs || oldPinfo.CheckPointTs != pinfo.CheckPointTs {
```
july2993: update to check both @amyangfei

cdc/owner.go
```diff
@@ -34,6 +35,8 @@ import (
 	"go.uber.org/zap"
 )
 
+var markProcessorDownTime = 5 * time.Minute
```
july2993: @amyangfei just increase to be 5 minute now.
I think the better implementation should check the aliveness of capture, like maintain a session in etcd.

amyangfei: Oh I mistake it at first, it is one minute, not one second. Five minutes is too long for CDC owner to detect processor down 🙁️
maybe we can recover it to 1 minute for the test, and have a better detect way in other PR.

july2993: reverted

PR #173, opened 2019-12-12
CLI: add a cli tool can delete all keys created by CDC in PD

leoppro:
add a cli can delete all keys created by CDC in PD

execute `cdc clear` will delete  all keys created by CDC in PD

### Check List 

Tests 

 - Unit test


leoppro:
> LGTM
> but will be more easy to use by using subcommands instead of the cmd params?

yes, but I should follow the original design.

PR #177, opened 2019-12-16
fix endless retry in singleEventFeed

july2993:

fix endless retry in singleEventFeed
just remove the for loop in `singleEventFeed` and return the error if create stream fails.

handle the error outside, retry for Unavailable

also, increase max retry time, know the previous time may too quick to fail, setting 10 will fail after near 50~60 second with the Backoff strategy in the `retry` pkg wrap. 

### Check List 

Tests 

restart tikv, may see the relate error and recreate singleEventFeed


amyangfei:
The failed integration test doesn't matter, we can merge this pr anyway

PR #178, opened 2019-12-16
support integration test in CI

amyangfei:

Add integration test support in CI

- run integration test in CI
- collect coverage combining integration test and unit test and upload to coveralls.io

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 ### Makefile for ticdc
-.PHONY: build test check clean fmt cdc coverage
+.PHONY: build test check clean fmt cdc coverage \
+	integration_test_build integration_test
 
 PROJECT=ticdc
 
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -63,6 +64,11 @@ check_third_party_binary:
 	@which bin/sync_diff_inspector
 	@which bin/go-ycsb
 
+integration_test_build:
+	$(GOTEST) -c -cover -covemode=atomic \
```
suzaku: Why is `$(GOTEST)` called integration test here?

amyangfei: Here we use `go test` tool to compile a `cdc binary`, which compiles the package in `main_test.go` and links to the main test program in `main_test.go`. In this way we can run a cdc daemon in the `go test` way.

suzaku: Maybe add an example of how to use it in the comment?

amyangfei: already have a reference link in `main_test.go`

PR #180, opened 2019-12-16
test: switch on binlog test, part 1

leoppro:
switch on binlog test, part 1

### DIFF
--- a/cdc/entry/mounter.go
+++ b/cdc/entry/mounter.go
@@ -61,7 +61,7 @@ func (m *Mounter) Mount(rawTxn model.RawTxn) (model.Txn, error) {
 			}
 			return t, nil
 		case *unknownKVEntry:
-			log.Warn("Found unknown kv entry", zap.Reflect("unknownKVEntry", e))
+			log.Debug("Found unknown kv entry", zap.Reflect("unknownKey", e.Key))
 		}
 	}
 	t.DMLs = append(deleteDMLs, replaceDMLs...)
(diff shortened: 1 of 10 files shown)


cdc/entry/mounter.go
```diff
@@ -61,7 +61,7 @@ func (m *Mounter) Mount(rawTxn model.RawTxn) (model.Txn, error) {
 			}
 			return t, nil
 		case *unknownKVEntry:
-			log.Warn("Found unknown kv entry", zap.Reflect("unknownKVEntry", e))
+			log.Debug("Found unknown kv entry", zap.Reflect("unknownKey", e.Key))
```
suzaku: Use specific type field for `e.Key`?

tests/util/db.go
```diff
@@ -118,6 +118,23 @@ func CheckSyncState(sourceDB, targetDB *sql.DB, schema string) bool {
...
+	sourceTableMap := make(map[string]struct{}, len(tables))
+	for _, table := range tables {
+		sourceTableMap[table] = struct{}{}
+	}
+	for _, table := range targetTables {
+		if _, exist := sourceTableMap[table]; !exist {
+			log.Print("The table in target db is not exist in source db:", table)
+			return false
+		}
+	}
```
suzaku: Doesn't the diff check already check this for us?

leoppro: when the table in target database is more than source database, the diff check will not return an error. This pr will fix this.

cdc/entry/mounter.go
leoppro: fixed

tests/util/db.go
suzaku: Add a comment to explain this?

leoppro: added

PR #181, opened 2019-12-17
get tso from pd, remove useless command

amyangfei:

When we create a new changefeed with current TSO, we compose a TSO from the local timestamp, which is not completely safe. As there exists
1. Time shift among physical machines
2. pd TSO pre-allocation

Fetch a tso from pd when we create a new changefeed with `current` TSO

### Check List 

Tests 

 - Unit test
 - Integration test


tests/_cli_tools/get_tso.go
```diff
@@ -0,0 +1,40 @@
+// Copyright 2019 PingCAP, Inc.
```
july2993: we can make `cdc` support this command directly? like `cdc get_tso`  or `cdc ctrl get_tso`

PR #182, opened 2019-12-17
Use a new async interface for Sink

suzaku:

The original Sink implementation accepts and saves transactions one by one.
If we want to utilize bulk writing or introduce some concurrency to improve performance, we need an async interface.

Change the Sink interface to an async one.

### Check List 

Tests 

 - Unit test


suzaku:
Do not merge for now, I have come up with another solution that may be simpler.

PR #187, opened 2019-12-20
support resign owner, add is-owner field in capture list result

amyangfei:

Add owner resign support and add a way for users to query which capture is the owner.

- add `ResignOwner` function in the `Manager` interface
- add a `is-owner` field in the result of `./cdc ctrl --cmd=query-capture-list`, just for test at present
- add an HTTP api to resign owner from a capture
- unit test entrance is removed by accident, put it back

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cmd/ctrl.go
+++ b/cmd/ctrl.go
@@ -20,8 +20,10 @@ import (
 	"time"
 
 	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/clientv3/concurrency"
 	pd "github.com/pingcap/pd/client"
 	"github.com/pingcap/ticdc/cdc/kv"
+	"github.com/pingcap/ticdc/cdc/roles"
 	"github.com/pingcap/tidb/store/tikv/oracle"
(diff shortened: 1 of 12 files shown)


cmd/ctrl.go
```diff
@@ -114,14 +122,24 @@ var ctrlCmd = &cobra.Command{
 			}
 			return jsonPrint(cfs)
 		case CtrlQueryCaptures:
-			_, captures, err := kv.GetCaptures(context.Background(), cli)
+			_, raw, err := kv.GetCaptures(context.Background(), cli)
 			if err != nil {
 				return err
 			}
+			mockOwnerMgr := roles.NewOwnerManager(cli, "mock-id", kv.CaptureOwnerKey)
+			ownerID, err := mockOwnerMgr.GetOwnerID(context.Background())
```
suzaku: Maybe just extract a standalone function to get owner ID? It seems the only inputs we need are just `key` and `ctx`.

PR #188, opened 2019-12-23
owner: fix startTs in processor table info dispatched from owner

amyangfei:

A processor starts a table replication from `StartTs`, which is first set by the owner.
The `StartTs` should consider both the `StartTs` of the changefeed and the `CheckpointTs` of the changefeed.

~Use max(`StartTs`, `CheckpointTs`) as `StartTs` in the `ProcessTableInfo`~
Use `detail.GetCheckpointTs()` as `StartTs` in the `ProcessTableInfo`

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -454,8 +454,9 @@ func (o *ownerImpl) loadChangeFeedInfos(ctx context.Context) error {
 		}
 
 		detail := changefeeds[changeFeedID]
+		checkpointTs := detail.GetCheckpointTs()
 		log.Info("find new changefeed", zap.Reflect("detail", detail),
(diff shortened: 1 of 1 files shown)


cdc/owner.go
```diff
@@ -489,10 +490,14 @@ func (o *ownerImpl) loadChangeFeedInfos(ctx context.Context) error {
 				continue
 			}
 
+			startTs := changefeed.StartTs
+			if checkpointTs > startTs {
```
july2993: seems `detail.GetCheckpointTs() >= detail.StartTs` should always be true or it's unreasonble

amyangfei: yep, change to use `detail.GetCheckpointTs()` directly

PR #189, opened 2019-12-23
Emit in bulk in syncResolved

suzaku:

We didn't utilize the bulk interface of Sink.

Emit in bulk instead of saving txns one by one.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/cdc/processor.go
+++ b/cdc/processor.go
@@ -523,6 +523,19 @@ func (p *processor) pullDDLJob(ctx context.Context) error {
 
 // syncResolved handle `p.ddlJobsCh` and `p.resolvedTxns`
 func (p *processor) syncResolved(ctx context.Context) error {
+	const bulkLimit = 128
+	pendingTxns := make([]model.Txn, 0, bulkLimit)
+	flush := func(ctx2 context.Context) error {
+		if len(pendingTxns) == 0 {
+			return nil
+		}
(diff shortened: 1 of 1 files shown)


amyangfei:
The coverage stage failure is caused by incompleted coverage information, it seems we haven't implemented the signal handling, I will file another PR to solve this problem. Just re-run CI and have a try, or ignore the error in this PR.

cdc/processor.go
```diff
@@ -539,35 +552,53 @@ func (p *processor) syncResolved(ctx context.Context) error {
...
+			}
+
+			if err := p.schemaStorage.HandlePreviousDDLJobIfNeed(rawTxn.Ts); err != nil {
+				return errors.Trace(err)
+			}
+			txn, err := p.mounter.Mount(rawTxn)
+			if err != nil {
+				return errors.Trace(err)
+			}
+			pendingTxns = append(pendingTxns, txn)
```
leoppro: I think we should filter the fake txn here.

leoppro:
rest LGTM

cdc/processor.go
suzaku: Isn't fake txns already filtered in the `ddlJobsCh` case?

leoppro:
LGTM

PR #190, opened 2019-12-23
Add more metrics

suzaku:

Need to add more metrics to make cdc more obersevable.

Add some new metrics:

1. The time taken to update sub change feed info.
1. The number of events received by puller.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/cdc/puller/metrics.go
+++ b/cdc/puller/metrics.go
@@ -11,21 +11,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package cdc
+package puller
 
-import (
-	"github.com/pingcap/ticdc/cdc/kv"
-	"github.com/pingcap/ticdc/cdc/puller"
-	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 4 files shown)


cdc/metrics.go
```diff
@@ -20,6 +20,17 @@ import (
 
 var registry = prometheus.NewRegistry()
 
+var (
+	updateInfoDuration = prometheus.NewHistogramVec(
```
amyangfei: This metric is not registered in `registry`

cdc/puller/metrics.go
```diff
@@ -0,0 +1,31 @@
...
+		prometheus.CounterOpts{
+			Namespace: "ticdc",
+			Subsystem: "puller",
+			Name:      "event_count",
+			Help:      "The number of events received.",
+		}, []string{"captureID", "type"})
+)
+
+// InitMetrics registers all metrics in this file
+func InitMetrics(registry *prometheus.Registry) {
```
amyangfei: This function should be called in the entrance  function `init()` in `cdc/metrics.go`?

PR #191, opened 2019-12-23
Refactor, changeFeedInfo -> changeFeed

suzaku:

We have already used the name `ChangeFeedInfo` for the parts of infomation about a change feed that's going to be saved in `etcd`.
It's confusing if we have both `ChangeFeedInfo` and `changeFeedInfo`.

Simply use `changeFeed`.

Some other minor changes: fix a typo in `flushChangeFeedInfos` and preallocate the map.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/cdc/owner_test.go
+++ b/cdc/owner_test.go
@@ -130,7 +130,7 @@ func (s *ownerSuite) TestPureDML(c *check.C) {
 
 	tables := map[uint64]schema.TableName{1: {Schema: "any"}}
 
-	changeFeedInfos := map[model.ChangeFeedID]*changeFeed{
+	changeFeeds := map[model.ChangeFeedID]*changeFeed{
 		"test_change_feed": {
 			tables:                  tables,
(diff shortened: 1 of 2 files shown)


cdc/owner_test.go
```diff
@@ -130,7 +130,7 @@ func (s *ownerSuite) TestPureDML(c *check.C) {
 
 	tables := map[uint64]schema.TableName{1: {Schema: "any"}}
 
-	changeFeedInfos := map[model.ChangeFeedID]*changeFeedInfo{
+	changeFeedInfos := map[model.ChangeFeedID]*changeFeed{
```
amyangfei:
```suggestion
	changeFeeds := map[model.ChangeFeedID]*changeFeed{
```
This can be shorter too

cdc/owner_test.go
```diff
@@ -309,7 +309,7 @@ func (s *ownerSuite) TestDDL(c *check.C) {
...
```
amyangfei: ditto

PR #192, opened 2019-12-24
Merge cdc/txn into cdc/puller to reduce the number of packages

suzaku:

`cdc/txn` is a very small package that's only used in `cdc/puller`, and it's not really `txn` specific.

Merge `cdc/txn` into `cdc/puller`.

### Check List 

Tests 

 - Unit test


leoppro:
LGTM

PR #193, opened 2019-12-24
Move txn filtering out of mysqlSink

suzaku:

Ignored or not, a "success" message was logged for each DDL passed to `sink.Emit`.
We need to separate the ignored and those that were actually executed.

Move txn filtering out of `mysqlSink`, the callers should filter the txns and decide what to do if the whole txn is ignored.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/cdc/filter.go
+++ b/cdc/filter.go
@@ -1,0 +1,37 @@
+// Copyright 2019 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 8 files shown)


cdc/filter.go
```diff
@@ -0,0 +1,37 @@
...
+
+package cdc
+
+import (
+	"regexp"
+
+	"github.com/pingcap/ticdc/cdc/model"
+)
+
+func filterBySchemaAndTable(t *model.Txn) {
```
leoppro: how about return a bool value to mark whether the table should bu filter

suzaku: This is a function with side effect, it would remove mutations that belong to a database or a table that's in the black list.

PR #194, opened 2019-12-24
Flush pending txns when the buffers are empty

suzaku:

When there's no more incoming txns and the number of pending txns doesn't reach the limit, the txns are not flushed.

Flush pending txns when the buffers are empty

### Check List 

Tests 

 - Unit test


suzaku:
Without this PR, integration tests would fail for all later PRs, so I am merging it now with only LGT1.

PR #195, opened 2019-12-24
wait all processors exit in capture

amyangfei:

This PR makes capture acts like a resource manager by the following two changes:

- capture doesn't exit when meeting a processor error
- capture waits for all processors exit during system shutdown

more resources management will be added later, such as the changefeed pause/resume

Still use the callback way for the interactions between capture and processors, we may refactor it later.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/capture.go
+++ b/cdc/capture.go
@@ -16,6 +16,8 @@ package cdc
 import (
 	"context"
 	"fmt"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/coreos/etcd/clientv3"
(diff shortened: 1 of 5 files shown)


cdc/capture.go
```diff
@@ -42,6 +44,10 @@ type Capture struct {
 	ownerWorker  *ownerImpl
 
 	processors map[string]*processor
+	procState  struct {
+		sync.Mutex
+		closed int32
```
july2993: what 's the point to use `atomic**` fro closed in the flowing codes, since it's protected by Mutex

amyangfei: It is used to guarantee operations in `OnStopProcessor` not executed if the `Capture` is closed, but since these operations are reentrant, we can remove this

amyangfei:
PTAL @july2993

PR #197, opened 2019-12-24
tests: wait cdc server exits in integration test

amyangfei:

In each integration test case, we should wait for the cdc gotest server daemon exits (which will ensure all coverage data is flushed into the disk),  before we stash the coverage file.

Add a process aliveness checker in the integration test script.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/tests/_utils/cleanup_process
+++ b/tests/_utils/cleanup_process
@@ -2,7 +2,7 @@
 # parameter 1: process name
 
 process=$1
-retry_count=40
+retry_count=20
 
 killall $process || true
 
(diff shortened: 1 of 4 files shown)


tests/_utils/cleanup_process
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# parameter 1: process name
+
+process=$1
+retry_count=40
```
suzaku: Is 20 seconds(`40 * 0.5`) too long?

PR #198, opened 2019-12-24
Fix deadlock while successive drop tables

july2993:
1, Adjust and add some metrics.

2, Fix goroutine leak by not close sorter

3, Fix diffProcessTableInfos assume tables is sorted by id.

4, Fix deadlock while successive drop tables.

drop two table at upstream: drop table t1; drop table t2;
ts1 = the timestamp of "drop table t1"
ts2 = the timestamp of "drop table t2"
the flowing step may happen:

a. gloableResolveTS = t1, checkpointTS of processor = ts (after execute
	drop table t1)
b. processor update resolveTS = t2;
c. owner update gloableResolveTS = t2;
d. processor update checkpointTS = t2(in memory).
e. owner remove table t1 in etcd.
f. processor learn t1 removed and update the in memory checkpointTS as
the one in etch which is t1.
g. processor will never forward checkpointTS and owner wait checkpointTS
of processor to be t2 for execute DDL `drop table t2`

2, close the sorter
3, sort it first
4, don't let ts fallback when some tables is removed

### Check List 

Tests 

 - Unit test
 - Integration test
will add in https://github.com/pingcap/ticdc/pull/180

Code changes

Side effects

Related changes

### DIFF
--- a/cdc/kv/metrics.go
+++ b/cdc/kv/metrics.go
@@ -16,12 +16,19 @@ package kv
 import "github.com/prometheus/client_golang/prometheus"
 
 var (
-	regionSplitCounter = prometheus.NewCounter(
+	eventFeedErrorCounter = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Namespace: "ticdc",
 			Subsystem: "kvclient",
-			Name:      "region_split_count",
-			Help:      "The number of region split events since start.",
+			Name:      "event_feed_error_count",
+			Help:      "The number of error return by tikv",
+		}, []string{"type"})
+	eventFeedGauge = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: "ticdc",
+			Subsystem: "kvclient",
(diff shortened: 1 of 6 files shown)


cdc/kv/metrics.go
```diff
@@ -16,12 +16,19 @@ package kv
...
+	eventFeedErrorCounter = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Namespace: "ticdc",
 			Subsystem: "kvclient",
-			Name:      "region_split_count",
-			Help:      "The number of region split events since start.",
+			Name:      "event_feed_error_count",
+			Help:      "The number of error return by tikv",
+		}, []string{"type"})
+	eventFeedGauge = prometheus.NewGauge(
```
suzaku: Do you need `captureID` as a label?

july2993: just display by instance, captureID is not readable and will not change, so i think no need to add by every metrics

PR #204, opened 2019-12-26
Add support for specifying database and table filter rules

suzaku:

Add support for specifying database and table filter rules.

1. Add filter related fields in `ChangeFeedDetail`
2. Add two methods for `ChangeFeedDetail` to check if a table should be ignored and filter a txn.

### Check List 

Tests 

 - Unit test


cdc/model/changefeed.go
```diff
@@ -31,6 +33,50 @@ type ChangeFeedDetail struct {
 	// The ChangeFeed will exits until sync to timestamp TargetTs
 	TargetTs uint64          `json:"target-ts"`
 	Info     *ChangeFeedInfo `json:"-"`
+
+	filter              *filter.Filter
+	FilterCaseSensitive bool          `json:"filter-case-sensitive"`
+	FilterRules         *filter.Rules `json:"filter-rules"`
```
amyangfei: Is `black-white-list` better name here, which has been used in DM and lightning.

cdc/model/changefeed.go
```diff
@@ -31,6 +33,50 @@ type ChangeFeedDetail struct {
...
+}
+
+func (detail *ChangeFeedDetail) getFilter() *filter.Filter {
+	if detail.filter == nil {
```
amyangfei: I recommend not to combine system level filter with user-defined `black-white-list`.
In current design, If users want to ignore only one non-system level schema such as `not_replica_db`, they have to write `ignore-dbs: ["INFORMATION_SCHEMA", "PERFORMANCE_SCHEMA", "mysql", "not_replica_db"]` ?

cdc/model/changefeed.go
amyangfei: Besides,  how about adding a user-friendly way to add `black-white-list` config when they create a new changefeed. (Maybe add it in `cdc cli`, just for QA test temporarily)

cdc/model/changefeed_test.go
```diff
@@ -0,0 +1,62 @@
+// You may obtain a copy of the License at
```
amyangfei: copyright is not complete

cdc/model/changefeed.go
suzaku: 1. The usage of `bwlist` seems confusing in DM, sometimes it references the filter itself and sometimes the rules, so I chose to reuse the name exposed by the filter package.
1. I can add that in a separate PR.

cdc/model/changefeed.go
suzaku: Fixed.

cdc/model/changefeed_test.go
suzaku: Fixed.

PR #206, opened 2019-12-26
refine subchangfeedinfo clean strategy

amyangfei:

Currently we don't have `SubChangefeedInfo` clean strategy.

- ~When a capture closes normally, clean all `SubChangefeedInfo` belongs to it. We can't do this, we must at least record the replication checkpoint information, just reuse the subchangefeedinfo~
- When the owner marks a processor down, clean both `CaptureInfo` and all `SubChangefeedInfo` belongs to the capture. Table rebalance strategy will be added in another PR

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/capture.go
```diff
@@ -97,7 +98,7 @@ func (c *Capture) OnRunProcessor(p *processor) {
 // OnStopProcessor implements processorCallback.
 func (c *Capture) OnStopProcessor(p *processor, err error) {
 	// TODO: handle processor error
-	log.Info("stop to run processor", zap.String("changefeed id", p.changefeedID), zap.Error(err))
+	log.Info("stop to run processor", zap.String("changefeed id", p.changefeedID), util.ZapErrorFilter(err, context.Canceled))
```
suzaku: How do we decide when to use `util.ZapErrorFilter(err, context.Canceled)` or not?

amyangfei: In most cases we just ignore the `context.Canceled` error

amyangfei:
have a better solution in #214

PR #210, opened 2019-12-27
Delete from map directly

suzaku:

All capture IDs are collected into a slice before they are deleted from the map.

In Go, it's OK to delete from a map while iterating over it. ([ref](https://github.com/golang/go/issues/9926#issuecomment-74988212))

Delete while iterating.

### Check List 

Tests 

 - Unit test


leoppro:
LGTM

PR #212, opened 2019-12-27
bug fix: fix case UpdateWhileDroppingCol

leoppro:
the test case help us find two bugs:
1. get column info and index info mistakenly
2. the `Refresh` of `cachedInspector` will never be called in processor

1. add `GetColumnInfo` and `GetIndicesInfo` functions for `TableInfo`
1. using schema storage in memory to get table info, deprecate the `cachedInspector`
1. reopen test UpdateWhileDroppingCol

### Check List 

Tests 

 - Unit test
 - Integration test


tests/dailytest/case.go
```diff
@@ -388,31 +387,28 @@ CREATE TABLE many_cols (
...
-	wg.Add(1)
+	closeCh := make(chan struct{})
 	go func() {
-		defer wg.Done()
-
 		// Keep updating to generate DMLs while the other goroutine's dropping columns
 		updateSQL := `UPDATE many_cols SET val = ? WHERE id = ?;`
-		for i := 0; i < 100; i++ {
+		i := 0
+		for {
```
suzaku: `for i := 1; ; i++`?

cdc/schema/storage.go
```diff
@@ -61,6 +61,48 @@ func (t TableName) String() string {
...
 
+// TableInfo provides meta data describing a DB table.
+type TableInfo struct {
+	*model.TableInfo
+	ColumnsOffset map[int64]int
+	IndicesOffset map[int64]int
+}
+
+// FromTableInfo creates a TableInfo from a model.TableInfo
+func FromTableInfo(info *model.TableInfo) *TableInfo {
```
suzaku: `schema.FromTableInfo` sounds like creating a `schema` from TableInfo.

Maybe `schema.WrapTableInfo`?

cdc/schema/storage.go
```diff
@@ -61,6 +61,48 @@ func (t TableName) String() string {
...
+	}
+	return &TableInfo{
+		TableInfo:     info,
+		ColumnsOffset: columnsOffset,
+		IndicesOffset: indicesOffset,
+	}
+}
+
+// GetColumnInfo returns the column info by ID
+func (ti *TableInfo) GetColumnInfo(colID int64) (*model.ColumnInfo, error) {
```
suzaku: Maybe return `(*model.ColumnInfo, exist bool)` instead?

cdc/sink/mysql.go
```diff
@@ -88,17 +76,23 @@ func configureSinkURI(sinkURI string) (string, error) {
...
 
-// NewMySQLSinkUsingSchema creates a new MySQL sink
-func NewMySQLSinkUsingSchema(db *sql.DB, schemaStorage *schema.Storage) Sink {
-	inspector := &cachedInspector{
-		db:    db,
-		cache: make(map[string]*tableInfo),
-		tableGetter: func(_ *sql.DB, schemaName string, tableName string) (*tableInfo, error) {
-			info, err := getTableInfoFromSchemaStorage(schemaStorage, schemaName, tableName)
+// NewMySQLSinkUsingSchema creates a new MySQL sink using schema storage
+func NewMySQLSinkUsingSchema(sinkURI string, infoGetter TableInfoGetter, _ map[string]string) (Sink, error) {
```
suzaku: I think we should keep the name `NewMySQLSink` instead.

The name for the unused parameter should be kept, because otherwise we might forget what it's supposed to be used for (or we might just get rid of it).

cdc/sink/mysql.go
```diff
@@ -88,17 +76,23 @@ func configureSinkURI(sinkURI string) (string, error) {
...
+	sinkURI, err := configureSinkURI(sinkURI)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	db, err := sql.Open("mysql", sinkURI)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	inspector := &tableInspectorImpl{
```
suzaku: We can add methods like `GetUniqueKeys` to `schema.TableInfo` and get rid of the `tableInspector` interface completely.

cdc/schema/storage.go
leoppro: nice name

cdc/sink/mysql.go
leoppro: good idea, I will implement this in next PR

suzaku: OK

cdc/entry/kv_entry.go
```diff
@@ -63,11 +64,14 @@ type unknownKVEntry struct {
...
 
-func (idx *indexKVEntry) unflatten(tableInfo *timodel.TableInfo) error {
+func (idx *indexKVEntry) unflatten(tableInfo *schema.TableInfo) error {
 	if tableInfo.ID != idx.TableID {
 		return errors.New("wrong table info in unflatten")
 	}
-	index := tableInfo.Indices[idx.IndexID-1]
+	index, exist := tableInfo.GetIndexInfo(idx.IndexID)
+	if !exist {
+		return errors.NotFoundf("index info, indexID: %d", index)
```
amyangfei:
```suggestion
		return errors.NotFoundf("index info, indexID: %d", idx.IndexID)
```
index is `nil` here

PR #214, opened 2019-12-30
owner rebalance tables when marking processor down

amyangfei:

Add clean strategy for `SubChangefeedInfo` when a processor is down and table rebalance at the same time.

When the owner marks a processor down, clean both `CaptureInfo` and all `SubChangeFeedInfo` belongs to the capture, and rebalance tables when marking processor down.
One exception, if the owner cleans the processor `SubChangeFeedInfo` and it crashes before table rebalance, some tables replication information is missing. In this scenario, the global `CheckpointTs` will not be forwarded until another owner is elected. In order to keep state management simple, we tend to add some recover mechanism in the owner's logic for this scenario, which will be solved in another PR.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/kv/etcd.go
+++ b/cdc/kv/etcd.go
@@ -252,24 +252,3 @@ func DeleteSubChangeFeedInfo(
 	_, err := cli.Delete(ctx, key)
 	return errors.Trace(err)
 }
-
-// DeleteCaptureFeeds removes all subchangefeedinfo belongs to a capture in etcd
-func DeleteCaptureFeeds(
-	ctx context.Context,
-	cli *clientv3.Client,
-	captureID string,
-	opts ...clientv3.OpOption,
-) error {
-	_, changefeeds, err := GetChangeFeeds(ctx, cli, opts...)
-	if err != nil {
-		return errors.Trace(err)
-	}
-	for cfID := range changefeeds {
-		key := GetEtcdKeySubChangeFeed(cfID, captureID)
-		_, err = cli.Delete(ctx, key)
-		if err != nil {
-			return errors.Trace(err)
-		}
-	}
-	return nil
-}
--- a/cdc/kv/etcd.go
+++ b/cdc/kv/etcd.go
@@ -239,3 +239,24 @@ func PutChangeFeedStatus(
 	_, err = client.Put(ctx, key, value, opts...)
 	return errors.Trace(err)
 }
+
+// DeleteCaptureFeeds removes all subchangefeedinfo belongs to a capture in etcd
+func DeleteCaptureFeeds(
+	ctx context.Context,
+	cli *clientv3.Client,
+	captureID string,
+	opts ...clientv3.OpOption,
+) error {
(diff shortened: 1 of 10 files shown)


cdc/owner.go
```diff
@@ -360,15 +373,36 @@ func (o *ownerImpl) addCapture(info *model.CaptureInfo) {
...
 		if err != nil {
-			log.Warn("failed to delete key", zap.Error(err))
+			log.Warn("failed to delete subchangefeed info",
+				zap.String("changefeedID", snap.cfID),
+				zap.String("captureID", snap.captureID),
+				zap.Error(err),
+			)
 			continue
 		}
+		o.markDownProcessor = append(o.markDownProcessor[:i], o.markDownProcessor[i+1:]...)
```
suzaku: It seems like most of the time all `markDownProcessor` should be removed, so it might be simpler to keep track of what items remain after all items are checked and then assign it to `o.markDownProcessor`.

cdc/owner.go
```diff
@@ -299,9 +306,16 @@ func (c *changeFeed) applyJob(job *pmodel.Job) error {
...
+type procInfoSnap struct {
+	cfID      string
+	captureID string
+	tables    []*model.ProcessTableInfo
+}
+
 type ownerImpl struct {
 	changeFeeds       map[model.ChangeFeedID]*changeFeed
-	markDownProcessor map[string]struct{}
+	markDownProcessor []*procInfoSnap
```
suzaku: Why change it to be a `slice`? What's the implication here?

cdc/owner.go
```diff
@@ -482,7 +516,22 @@ func (o *ownerImpl) loadChangeFeeds(ctx context.Context) error {
...
+					for _, tbl := range info.TableInfos {
+						ts := info.CheckPointTs
+						if ts < tbl.StartTs {
+							ts = tbl.StartTs
+						}
+						snap.tables = append(snap.tables, &model.ProcessTableInfo{
+							ID:      tbl.ID,
+							StartTs: ts,
+						})
+					}
```
suzaku: Maybe create a method on `info` to `takeSnapshot`?

cdc/kv/etcd.go
```diff
@@ -239,3 +239,37 @@ func PutChangeFeedStatus(
...
+	captureID string,
+	opts ...clientv3.OpOption,
+) error {
+	_, changefeeds, err := GetChangeFeeds(ctx, cli, opts...)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	for cfID := range changefeeds {
+		key := GetEtcdKeySubChangeFeed(cfID, captureID)
+		_, err = cli.Delete(ctx, key)
```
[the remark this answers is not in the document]
amyangfei: not used anymore, just remove this function

cdc/owner.go
amyangfei: The origin map is not correct since there may exist multiple `changefeeds`

PR #215, opened 2019-12-30
Replace tableInspector with infoGetter

suzaku:

The information we need from tableInspector can be queried from infoGetter, so we can just replace it with infoGetter.

Replace tableInspector with infoGetter, remove related implementation.

### Check List 

Tests 

 - Unit test


cdc/schema/storage.go
```diff
@@ -103,6 +103,41 @@ func (ti *TableInfo) GetIndexInfo(indexID int64) (info *model.IndexInfo, exist b
...
+	}
+	return cols
+}
+
+// GetUniqueKeys returns all unique keys of the table as a slice of column names
+func (ti *TableInfo) GetUniqueKeys() [][]string {
+	var uniqueKeys [][]string
+	for _, idx := range ti.Indices {
+		if idx.Primary || idx.Unique {
+			colNames := make([]string, 0, len(idx.Columns))
```
leoppro: we should filter nullable column here

cdc/schema/storage.go
```diff
@@ -103,6 +103,41 @@ func (ti *TableInfo) GetIndexInfo(indexID int64) (info *model.IndexInfo, exist b
...
+	}
+	if ti.PKIsHandle {
+		for _, col := range ti.Columns {
+			if mysql.HasPriKeyFlag(col.Flag) {
+				// Prepend to make sure the primary key ends up at the front
+				uniqueKeys = append([][]string{{col.Name.O}}, uniqueKeys...)
+				break
+			}
+		}
+	}
```
leoppro: should we make sure that the `pk` is the first in unique keys?

suzaku: I'm not sure, just to make it compatible with the original `inspector` implementation.

leoppro:
LGTM

PR #216, opened 2019-12-30
fix table management when owner starts up

amyangfei:

Currently when a capture becomes owner, it always scans all changefeeds, and re-assign each table to be replicated.

Filter tables already in replication state in all processors, don't re-assign them.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -418,7 +418,7 @@ func (o *ownerImpl) handleWatchCapture() error {
 func (o *ownerImpl) newChangeFeed(id model.ChangeFeedID, processorsInfos model.ProcessorsInfos, detail *model.ChangeFeedDetail) (*changeFeed, error) {
 	checkpointTs := detail.GetCheckpointTs()
 	log.Info("Find new changefeed", zap.Reflect("detail", detail),
-		zap.Uint64("checkpoint ts", checkpointTs))
+		zap.String("id", id), zap.Uint64("checkpoint ts", checkpointTs))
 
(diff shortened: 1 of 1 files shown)


cdc/owner.go
```diff
@@ -432,16 +432,29 @@ func (o *ownerImpl) newChangeFeed(id model.ChangeFeedID, processorsInfos model.P
...
 		if detail.ShouldIgnoreTable(table.Schema, table.Table) {
 			continue
 		}
 
-		tables[id] = table
-		orphanTables[id] = model.ProcessTableInfo{
-			ID:      id,
+		tables[tid] = table
+		if _, ok := existingTables[tid]; ok {
+			log.Info("ignore existing replication table",
```
suzaku: No need to log this? Or maybe use the `DEBUG` level.

amyangfei: Ignore a table may be a sensitive operation, if we skip a table by mistake, data loss happens. Still keep the `INFO` level, but add a `checkpointTs` field in log

suzaku: In the case this PR aimed to fix, it seems this would output one line for each table involved. If we are to keep it, we should strive to make it as simple as possible.

For example:

1. Use shorter message like "Ignore known table"
1. Remove fields that don't change during the iteration, here we can log `changefeedID` and `checkpoint` outside the loop
1. Maybe shorter field names like "tid" and "tbl"

amyangfei: change to `DEBUG` level. `checkpoint` is not always the same. `changefeedID` well is logged outside already.

PR #217, opened 2019-12-30
close channel p.executedTxns where we use it as sender

amyangfei:

Fix a potential panic of `send on closed channel`, which is caused by the branch of a select that is chosen could be non-deterministic (except for the `default` branch).

Close the channel p.executedTxns where we use it as sender

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/processor.go
+++ b/cdc/processor.go
@@ -477,7 +477,6 @@ func (p *processor) globalResolvedWorker(ctx context.Context) error {
 
 	defer func() {
 		close(p.resolvedTxns)
-		close(p.executedTxns)
 	}()
 
 	retryCfg := backoff.WithMaxRetries(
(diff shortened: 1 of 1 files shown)


cdc/processor.go
```diff
@@ -580,6 +579,11 @@ func (p *processor) syncResolved(ctx context.Context) error {
 		pendingTxns = pendingTxns[:0]
 		return nil
 	}
+
+	defer func() {
+		close(p.executedTxns)
+	}()
```
suzaku: The two `defer` can be simplified to one-liners now?

amyangfei: The style is not that important

amyangfei: Yet open coded defer is a little faster

PR #222, opened 2019-12-31
mounter: don't build delete dml when set index key

leoppro:
don't build delete dml when set index key

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/entry/mounter.go
+++ b/cdc/entry/mounter.go
@@ -119,6 +119,10 @@ func (m *Mounter) mountRowKVEntry(row *rowKVEntry) (*model.DML, error) {
 }
 
 func (m *Mounter) mountIndexKVEntry(idx *indexKVEntry) (*model.DML, error) {
+	// skip set index KV
+	if !idx.Delete {
+		return nil, nil
+	}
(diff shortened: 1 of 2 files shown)


cdc/entry/mounter.go
```diff
@@ -119,6 +119,10 @@ func (m *Mounter) mountRowKVEntry(row *rowKVEntry) (*model.DML, error) {
 }
 
 func (m *Mounter) mountIndexKVEntry(idx *indexKVEntry) (*model.DML, error) {
+	// skip set index KV
+	if !idx.Delete {
+		return nil, nil
```
suzaku: Please add a test case to help reviewers understand what this change tried to fix.

leoppro: this change fixes nothing. We translate setting row key to `replace` query, so we needn't build delete dml for setting index key.

leoppro: see:https://docs.google.com/document/d/1D1IqIs_q0y0YSnu6hJgKZO4lkCkDqOexDxzMdK-7NRU/edit#

PR #224, opened 2020-01-02
Add support for `cdc cli` to read filter configs from file

suzaku:

A command line interface is needed to test filter rules.

1. Add a `config` option for `cdc cli` to read configurations like `filter-rules` from a file
1. Fix a bug

### Check List 

Tests 

 - Unit test
 - Manual test
   Change `run.sh` to specify a config file and observe that tables are filtered according to the configuration.

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -3,7 +3,13 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
+
+	"github.com/pingcap/tidb-tools/pkg/filter"
+
+	"github.com/BurntSushi/toml"
+	"github.com/pingcap/errors"
 
 	"github.com/coreos/etcd/clientv3"
 	_ "github.com/go-sql-driver/mysql" // mysql driver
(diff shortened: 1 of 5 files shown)


leoppro:
how about add an example config file?

leoppro:
LGTM

cmd/client.go
```diff
@@ -59,14 +72,51 @@ var cliCmd = &cobra.Command{
...
+			return err
+		}
+		fmt.Printf("create changefeed ID: %s detail %s\n", id, d)
 		return kv.SaveChangeFeedDetail(context.Background(), cli, detail, id)
 	},
 }
+
+// strictDecodeFile decodes the toml file strictly. If any item in confFile file is not mapped
+// into the Config struct, issue an error and stop the server from starting.
+func strictDecodeFile(path, component string, cfg interface{}) error {
```
amyangfei: Is it better to put this function in a general lib and set it public.
Also recommend adding a unit test and a config sample file

suzaku: Unit tests and example file `cmd/cdc.toml` added.

Since no other places use this function now, I think it's better if we keep it close to where it's used. We can easily move it to a general package later if other modules need it.

PR #225, opened 2020-01-03
support changefeed stop/pause/remove operation

amyangfei:

We have no management to a running changefeed, such as pause replication, resume replication or remove this changefeed from a CDC cluster.

- Treat changefeed as resource both in owner and processor, so if a DDL executed to downstream with failure, we don't exit the whole owner, but only pause the changefeed replication.
- Introduce the `AdminJob` concept, which can be processed by the owner only. We can use this feature by HTTP API now.
- In both `Stop` and `Remove` job, we release all the TiKV gRPC connections and stop the replication. The difference between these two jobs is the `Stop` job will preserve the `Changefeed` information, so we can resume the replication of this changefeed

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/http_handler.go
+++ b/cdc/http_handler.go
@@ -15,9 +15,16 @@ package cdc
 
 import (
 	"net/http"
+	"strconv"
 
 	"github.com/coreos/etcd/clientv3/concurrency"
 	"github.com/pingcap/errors"
+	"github.com/pingcap/ticdc/cdc/model"
+)
+
+const (
+	opVarAdminJob     = "job"
+	opVarChangefeedID = "cf-id"
 )
 
 type commonResp struct {
@@ -25,12 +32,7 @@ type commonResp struct {
 	Message string `json:"message"`
 }
 
-func (s *Server) handleResignOwner(w http.ResponseWriter, req *http.Request) {
-	if req.Method != http.MethodPost {
-		writeError(w, http.StatusBadRequest, errors.New("this api only supports POST method"))
-		return
-	}
-	err := s.capture.ownerManager.ResignOwner(req.Context())
+func handleOwnerResp(w http.ResponseWriter, err error) {
 	if err != nil {
 		if errors.Cause(err) == concurrency.ErrElectionNotLeader {
 			writeError(w, http.StatusBadRequest, err)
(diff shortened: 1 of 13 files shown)


cdc/owner.go
```diff
@@ -322,6 +323,9 @@ type ownerImpl struct {
 	captureWatchC      <-chan *CaptureInfoWatchResp
 	cancelWatchCapture func()
 	captures           map[model.CaptureID]*model.CaptureInfo
+
+	jobs     []*model.AdminJob
```
leoppro: call it `adminJobs` to avoid confusion with `ddlJob`?

cdc/http_handler.go
```diff
@@ -15,29 +15,64 @@ package cdc
...
 	"net/http"
+	"strconv"
 
 	"github.com/coreos/etcd/clientv3/concurrency"
 	"github.com/pingcap/errors"
+	"github.com/pingcap/ticdc/cdc/model"
+)
+
+const (
+	opVarAdminJob     = "job"
```
leoppro:
```suggestion
	opVarAdminJob     = "admin-job"
```

cdc/http_handler.go
```diff
@@ -15,29 +15,64 @@ package cdc
...
 	if req.Method != http.MethodPost {
 		writeError(w, http.StatusBadRequest, errors.New("this api only supports POST method"))
 		return
 	}
 	err := s.capture.ownerManager.ResignOwner(req.Context())
+	handleOwnerResp(w, err)
+}
+
+func (s *Server) handleChangefeedAdmin(w http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodPost {
```
suzaku: We can restrict HTTP methods by adding a call to `Methods("POST")` after `HandleFunc`.

cdc/owner.go
```diff
@@ -322,6 +323,9 @@ type ownerImpl struct {
 	captureWatchC      <-chan *CaptureInfoWatchResp
 	cancelWatchCapture func()
 	captures           map[model.CaptureID]*model.CaptureInfo
+
+	adminJobs     []*model.AdminJob
```
suzaku: No need to use pointer for small structs.

cdc/http_handler.go
amyangfei: default http `serverMux` does not provide `methodMatcher`

suzaku: We can use `gorilla/mux` later.

cdc/model/changefeed.go
```diff
@@ -33,8 +32,12 @@ type ChangeFeedDetail struct {
 	// Start sync at this commit ts if `StartTs` is specify or using the CreateTime of changefeed.
 	StartTs uint64 `json:"start-ts"`
 	// The ChangeFeed will exits until sync to timestamp TargetTs
-	TargetTs uint64          `json:"target-ts"`
-	Info     *ChangeFeedInfo `json:"-"`
+	TargetTs uint64 `json:"target-ts"`
+	// ChangeFeed can be resumed from stop state, this field is used to trigger a watch event in each capture
+	ResumeTs uint64 `json:"resume-ts"`
```
suzaku: Is it possible to use `checkpointTs` to avoid introducing a similar concept?

amyangfei: removed, PTAL

PR #231, opened 2020-01-08
Add ignore-txn-commit-ts to skip some txn

amyangfei:

Support ignore some txn by `CommitTs`

Add an `ignore-txn-commit-ts` parameter.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/model/changefeed.go
+++ b/cdc/model/changefeed.go
@@ -41,16 +41,23 @@ type ChangeFeedDetail struct {
 	Config *ReplicaConfig `json:"config"`
 }
 
+func (detail *ChangeFeedDetail) getConfig() *ReplicaConfig {
+	if detail.Config == nil {
+		detail.Config = &ReplicaConfig{}
+	}
+	return detail.Config
+}
+
 func (detail *ChangeFeedDetail) getFilter() *filter.Filter {
 	if detail.filter == nil {
-		rules := detail.Config.FilterRules
-		detail.filter = filter.New(detail.Config.FilterCaseSensitive, rules)
+		rules := detail.getConfig().FilterRules
+		detail.filter = filter.New(detail.getConfig().FilterCaseSensitive, rules)
 	}
 	return detail.filter
 }
 
 func (detail *ChangeFeedDetail) isIgnoreTxnCommitTs(t *Txn) bool {
-	for _, ignoreTs := range detail.Config.IgnoreTxnCommitTs {
+	for _, ignoreTs := range detail.getConfig().IgnoreTxnCommitTs {
 		if ignoreTs == t.Ts {
 			return true
 		}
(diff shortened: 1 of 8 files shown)


cdc/model/changefeed.go
```diff
@@ -37,19 +37,34 @@ type ChangeFeedDetail struct {
...
 	if detail.filter == nil {
-		rules := detail.FilterRules
-		detail.filter = filter.New(detail.FilterCaseSensitive, rules)
+		rules := detail.getConfig().FilterRules
+		detail.filter = filter.New(detail.getConfig().FilterCaseSensitive, rules)
 	}
 	return detail.filter
 }
 
+func (detail *ChangeFeedDetail) isIgnoreTxnCommitTs(t *Txn) bool {
```
suzaku: Maybe name it `ShouldIgnoreTxn`?

cdc/model/changefeed.go
```diff
@@ -66,13 +81,13 @@ func (detail *ChangeFeedDetail) ShouldIgnoreTable(db, tbl string) bool {
 // CDC only supports filtering by database/table now.
 func (detail *ChangeFeedDetail) FilterTxn(t *Txn) {
 	if t.IsDDL() {
-		if detail.ShouldIgnoreTable(t.DDL.Database, t.DDL.Table) {
+		if detail.isIgnoreTxnCommitTs(t) || detail.ShouldIgnoreTable(t.DDL.Database, t.DDL.Table) {
```
suzaku: I think it would be simpler if we require the caller to call `ShouldIgnoreTxn` separately before filtering.
The semantic is not exactly the same:

1. Ignore the whole Txn because it's specified with commit ts
1. OK, the Txn should not be ignored, let's filter something some ignorable mutations

cdc/model/changefeed.go
```diff
@@ -37,19 +37,34 @@ type ChangeFeedDetail struct {
 	AdminJobType AdminJobType    `json:"admin-job-type"`
 	Info         *ChangeFeedInfo `json:"-"`
 
-	filter              *filter.Filter
-	FilterCaseSensitive bool          `json:"filter-case-sensitive"`
-	FilterRules         *filter.Rules `json:"filter-rules"`
+	filter *filter.Filter
+	Config *ReplicaConfig `json:"config"`
```
suzaku: It seems like `ChangeFeedDetail` is the config itself. For example, it includes things like `SinkURI` and `TargetTs`. 

Maybe we should think about how we are going to configure change feeds later.

amyangfei: More like an L2 config of a L1 config. We can refine it later.

cdc/model/changefeed.go
amyangfei: refined, PTAL

PR #232, opened 2020-01-08
update kvproto dep version

amyangfei:

kvproto has merged the cdc protocol and deletes the old `cdc` branch. We should update kvproto version in go.mod (to fix failure in make check: https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/cdc_ghpr_check/detail/cdc_ghpr_check/662/pipeline)

~Also update tidb related dependencies to the latest. unfortunately found encode/decode compatible  issue, will upgrade other dependencies later. ref: https://github.com/pingcap/tidb/blob/master/docs/design/2018-07-19-row-format.md~

### Check List 

Tests 

 - Unit test
 - Integration test


leoppro:
LGTM

PR #233, opened 2020-01-09
kvclient: fix data race about checkpoint ts

leoppro:
fix data race about checkpoint ts

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -371,7 +371,7 @@ func (c *CDCClient) singleEventFeed(
 
 	conn, err := c.getConnByMeta(ctx, meta)
 	if err != nil {
-		return uint64(req.CheckpointTs), err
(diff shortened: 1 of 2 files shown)


cdc/kv/client.go
```diff
@@ -543,6 +537,15 @@ func (c *CDCClient) singleEventFeed(
 	}
 }
 
+func updateCheckpointTS(checkpointTs *uint64, newValue uint64) {
+	for {
```
amyangfei: no need to use the for loop?

leoppro: yes, we need

we need retry if `CompareAndSwap` is failed.

amyangfei: right, found the `checkpointTs` can be updated by both background sorter and when receiving resolvedTs

PR #234, opened 2020-01-09
capture: add the hostname to capture info

leoppro:
It's difficult to determine which server the capture is running on 

add the hostname to capture info

### Check List 

Tests 

 - Unit test
 - Integration test
run ticdc cluster and run query  capture list:
```
./cdc ctrl --cmd=query-capture-list
[{"id":"21a54198-1644-4c8b-9c59-58ac3af547dc","hostname":"834031ee1ee9","is-owner":true}]
```

### DIFF
--- a/cdc/model/capture.go
+++ b/cdc/model/capture.go
@@ -21,7 +21,8 @@ import (
 
 // CaptureInfo store in etcd.
 type CaptureInfo struct {
-	ID string `json:"id"`
+	ID       string `json:"id"`
+	HostName string `json:"hostname"`
 }
 
 // Marshal using json.Marshal.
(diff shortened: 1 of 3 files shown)


cdc/model/capture.go
```diff
@@ -21,7 +21,8 @@ import (
 
 // CaptureInfo store in etcd.
 type CaptureInfo struct {
-	ID string `json:"id"`
+	ID       string `json:"id"`
+	HostName string `json:"hostname"`
```
suzaku: I think we can learn from Binlog and use `statusHost` and `statusPort` for this purpose.

PR #238, opened 2020-01-13
bug fix: kvclient may be output a row which value is null

leoppro:
kvclient may be output a row which value is null

When two changes of the same key get received, the previous one may be overridden after the second prewrite.

Use "startTs + key" instead of "key" alone as the key of the state map.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/kv/matcher.go
+++ b/cdc/kv/matcher.go
@@ -29,17 +29,10 @@ func (m *matcher) putPrewriteRow(row *cdcpb.Event_Row) {
 	m.unmatchedValue[newMatchKey(row)] = row.GetValue()
 }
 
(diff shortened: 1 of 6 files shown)


cdc/kv/matcher.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+func (m *matcher) putPrewriteRow(row *cdcpb.Event_Row) {
+	m.unmatchedValue[newMatchKey(row)] = row.GetValue()
+}
+
+func (m *matcher) matchRow(row *cdcpb.Event_Row) error {
+	if value, exist := m.unmatchedValue[newMatchKey(row)]; exist {
+		row.Value = value
```
suzaku: It's unnatural to modify the input in place in a `match` methdo.

suzaku:
rest LGTM

PR #241, opened 2020-01-16
decode: support new format of row data

leoppro:
Ref: https://github.com/pingcap/tidb/pull/12634
        https://github.com/pingcap/tidb/blob/master/docs/design/2018-07-19-row-format.md

support new format of row data

1. update tidb version
2. check the version of  row data format 
3. decode row data in new format
4. make the behaviour of these two functions consistent:
`decodeRowV1`, `decodeRowV2`

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/capture.go
```diff
@@ -55,7 +56,15 @@ func NewCapture(pdEndpoints []string) (c *Capture, err error) {
 		Endpoints:   pdEndpoints,
 		DialTimeout: 5 * time.Second,
 		DialOptions: []grpc.DialOption{
-			grpc.WithBackoffMaxDelay(time.Second * 3),
+			grpc.WithConnectParams(grpc.ConnectParams{
+				Backoff: backoff.Config{
+					BaseDelay:  time.Second,
+					Multiplier: 0.1,
```
suzaku: This should be greater than 1, otherwise the delay would become smaller and smaller and never reach the `MaxDelay`?

suzaku: This should be greater than 1, otherwise the delay would become smaller and smaller and never reach the `MaxDelay`?

cdc/kv/client.go
```diff
@@ -117,7 +118,15 @@ func (c *CDCClient) getConn(
 		ctx,
 		addr,
 		grpc.WithInsecure(),
-		grpc.WithBackoffMaxDelay(time.Second),
+		grpc.WithConnectParams(grpc.ConnectParams{
+			Backoff: gbackoff.Config{
+				BaseDelay:  time.Second,
+				Multiplier: 0.1,
```
suzaku: Should be greater than 1.

cdc/capture.go
leoppro: yes, you are right

leoppro:
@suzaku  PTAL again

go.mod
```diff
@@ -7,26 +7,40 @@ require (
...
-	github.com/pingcap/pd v1.1.0-beta.0.20190923032047-5c648dc365e0
-	github.com/pingcap/tidb v1.1.0-beta.0.20191017064824-e1ba309148ad
-	github.com/pingcap/tidb-tools v2.1.3-0.20190321065848-1e8b48f5c168+incompatible
-	github.com/prometheus/client_golang v0.9.0
-	github.com/remyoudompheng/bigfft v0.0.0-20190728182440-6a916e37a237 // indirect
-	github.com/sirupsen/logrus v1.4.2 // indirect
-	github.com/spf13/cobra v0.0.3
+	github.com/pingcap/parser v0.0.0-20200109073933-a9496438d77d
+	github.com/pingcap/pd v1.1.0-beta.0.20191219054547-4d65bbefbc6d
+	github.com/pingcap/tidb v1.1.0-beta.0.20200116051450-6e648401e44d
```
lysu: hi, we have found a new bug in tidb, maybe need upgrade tidb hash after https://github.com/pingcap/tidb/pull/14631 merged thx

leoppro: ok thanks

cdc/puller/mock_puller_test.go
```diff
@@ -19,7 +19,7 @@ type mockPullerSuite struct{}
 var _ = check.Suite(&mockPullerSuite{})
 
 func (s *mockPullerSuite) TestTxnSort(c *check.C) {
-	pm := NewMockPullerManager(c)
+	pm := NewMockPullerManager(c, false)
```
amyangfei: What about setting `newRowFormat` to true, ditto for the following cases.

cdc/schema/storage.go
```diff
@@ -98,6 +101,32 @@ func (ti *TableInfo) GetIndexInfo(indexID int64) (info *model.IndexInfo, exist b
...
+		return ti.handleColID, ti.rowColInfos
+	}
+	handleColID := int64(-1)
+	reqCols := make([]rowcodec.ColInfo, len(ti.Columns))
+	for i, col := range ti.Columns {
+		isPK := (ti.PKIsHandle && mysql.HasPriKeyFlag(col.Flag)) || col.ID == model.ExtraHandleID
+		if isPK {
+			handleColID = col.ID
+		}
+		reqCols[i] = rowcodec.ColInfo{
```
amyangfei: seems the `IsPKHandle` field is not used currently, but also fill in it?

cdc/entry/codec.go
```diff
@@ -180,14 +182,29 @@ func decodeMetaKey(ek []byte) (meta, error) {
...
 // decodeRow decodes a byte slice into datums with a existing row map.
+func decodeRow(b []byte, recordID int64, tableInfo *schema.TableInfo) (map[int64]types.Datum, error) {
+	if len(b) == 0 {
+		if tableInfo.PKIsHandle {
+			id, pkValue, err := fetchHandleValue(tableInfo, recordID)
+			if err != nil {
+				return nil, errors.Trace(err)
+			}
+			return map[int64]types.Datum{id: *pkValue}, nil
+		}
```
amyangfei: what about extract a function, which has the similar logic with L247-L253

amyangfei:
btw, shall we upgrade the TiDB binary in jenkins integration test to the latest?

leoppro:
> btw, shall we upgrade the TiDB binary in jenkins integration test to the latest?

I will do that and add an integration test for row format compatibility

PR #242, opened 2020-01-16
retry when etcd watch meets mvcc.ErrCompacted

amyangfei:

etcd has an auto compaction mechanism, we have some etcd key watch with revision. If error happens and retries from a compacted key revision, the error "mvcc: required revision has been compacted" will be returned from etcd API.

- add necessary retry on the `mvcc.ErrCompacted`
- remove revision if we can watch without rev
- add failpoint in unit  test

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/capture.go
+++ b/cdc/capture.go
@@ -130,6 +130,7 @@ func (c *Capture) Start(ctx context.Context) (err error) {
 			err := watcher.Watch(cctx, c)
 			if errors.Cause(err) == mvcc.ErrCompacted {
 				log.Warn("changefeed watcher watch retryable error", zap.Error(err))
+				time.Sleep(time.Millisecond * 500)
 				continue
 			}
 			return errors.Trace(err)
--- a/cdc/capture.go
+++ b/cdc/capture.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/mvcc"
 	"github.com/google/uuid"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
@@ -125,7 +126,14 @@ func (c *Capture) Start(ctx context.Context) (err error) {
 
 	watcher := NewChangeFeedWatcher(c.info.ID, c.pdEndpoints, c.etcdClient)
 	errg.Go(func() error {
-		return watcher.Watch(cctx, c)
+		for {
+			err := watcher.Watch(cctx, c)
+			if errors.Cause(err) == mvcc.ErrCompacted {
+				log.Warn("changefeed watcher watch retryable error", zap.Error(err))
+				continue
+			}
+			return errors.Trace(err)
+		}
 	})
 
 	return errg.Wait()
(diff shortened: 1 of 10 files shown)


cdc/owner.go
```diff
@@ -862,7 +876,18 @@ func (o *ownerImpl) Run(ctx context.Context, tickTime time.Duration) error {
...
+		for {
+			err = o.handleWatchCapture()
+			if errors.Cause(err) != mvcc.ErrCompacted {
+				break
+			}
+			log.Warn("capture info watcher retryable error", zap.Error(err))
+			err = o.resetCaptureInfoWatcher(ctx)
+			if err != nil {
+				break
+			}
```
leoppro: sleep 0.5 second?

leoppro:
LGTM

cdc/capture_info.go
```diff
@@ -99,6 +101,10 @@ func newCaptureInfoWatch(
 		etcdWatchC := cli.Watch(ctx, captureEinfoKeyPrefix, clientv3.WithPrefix(), clientv3.WithRev(revision+1), clientv3.WithPrevKV())
 
 		for resp := range etcdWatchC {
+			failpoint.Inject("WatchCaptureInfoCompactionErr", func() {
+				watchResp <- &CaptureInfoWatchResp{Err: errors.Trace(mvcc.ErrCompacted)}
```
suzaku: Do we need to use `Trace` here?

suzaku: Do we need to use `Trace` here?

cdc/capture_info_test.go
```diff
@@ -93,46 +103,64 @@ func (ci *captureInfoSuite) TestWatch(c *check.C) {
...
 			c.Fatal("timeout to get resp from watchC")
-
 		}
 	}
 
-	// put info2 and info3
+	checkCaptureLen := func(expected int) {
+		owner.l.RLock()
+		defer owner.l.RUnlock()
+		c.Assert(len(owner.captures), check.Equals, expected)
```
suzaku: Can use `HasLen` for better error message?

cdc/capture_info.go
amyangfei: This simulates the common error handling code, which adds trace stack to the `resp.Err()`

cdc/capture.go
```diff
@@ -125,7 +126,15 @@ func (c *Capture) Start(ctx context.Context) (err error) {
 
 	watcher := NewChangeFeedWatcher(c.info.ID, c.pdEndpoints, c.etcdClient)
 	errg.Go(func() error {
-		return watcher.Watch(cctx, c)
+		for {
+			err := watcher.Watch(cctx, c)
+			if errors.Cause(err) == mvcc.ErrCompacted {
```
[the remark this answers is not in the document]
amyangfei: Seems we need a sliding window mechanism, such as retry at most M times during N continuous minutes.
We also need the similar mechanism in the kv client. What about adding a pkg in another pr?

cdc/capture_info_test.go
amyangfei: fixed

cdc/capture.go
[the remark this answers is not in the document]
suzaku: ok

PR #243, opened 2020-01-18
owner: record schemas info and remove tables when drop database

amyangfei:

When `DROP DATABASE` is executed in upstream, all tables belong to this schema should be removed from CDC.

- maintain a map mapping from schema ID to table ID set
- remove all tables belong to the schema which is dropped.

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/owner.go
```diff
@@ -290,21 +303,34 @@ func (c *changeFeed) applyJob(job *pmodel.Job) error {
 
 	// case table id set may change
 	switch job.Type {
+	case pmodel.ActionCreateSchema:
+		c.schemas[uint64(job.SchemaID)] = make(map[uint64]struct{})
+	case pmodel.ActionDropSchema:
+		schemaID := uint64(job.SchemaID)
```
leoppro: how about move this 7 lines into a new function named`dropSchema`?

cdc/owner.go
```diff
@@ -290,21 +303,34 @@ func (c *changeFeed) applyJob(job *pmodel.Job) error {
 
 	// case table id set may change
 	switch job.Type {
+	case pmodel.ActionCreateSchema:
+		c.schemas[uint64(job.SchemaID)] = make(map[uint64]struct{})
```
leoppro: ow about move this line into a new function named `addSchema`?

cdc/owner.go
```diff
@@ -120,25 +123,35 @@ func (c *changeFeed) reAddTable(id, startTs uint64) {
...
 	}
 }
 
-func (c *changeFeed) removeTable(id uint64) {
-	delete(c.tables, id)
+func (c *changeFeed) removeTable(sid, tid uint64) {
+	if _, ok := c.schemas[sid]; ok {
+		delete(c.schemas[sid], tid)
+		if len(c.schemas[sid]) == 0 {
+			delete(c.schemas, sid)
```
suzaku: It's OK to have an empty database with no table. It might be simpler if we only do `c.schemas` cleanup when a database is dropped.

cdc/owner.go
```diff
@@ -114,32 +117,52 @@ func (c *changeFeed) updateProcessorInfos(processInfos model.ProcessorsInfos) {
 	c.processorInfos = processInfos
 }
 
+func (c *changeFeed) addSchema(schemaID uint64) {
+	c.schemas[schemaID] = make(map[uint64]struct{})
```
suzaku: Add a check here so that when `addSchema` is called in the wrong place it doesn't silently empty the existing tables?

leoppro:
LGTM

cdc/owner.go
amyangfei: added, PTAL

PR #244, opened 2020-01-19
Refactor: add a wrap of etcd client for CDC

leoppro:
add a wrap of etcd client

tidy the code about etcd
### Check List 

Tests 

 - Unit test
 - Integration test


suzaku:
@leoppro Please resolve conflicts.

suzaku:
Can you explain why you consider this change necessary? Telling from the `diff`, it seems this change doesn't simplify the code but need to touch a lot of files.

leoppro:
> Can you explain why you consider this change necessary? Telling from the `diff`, it seems this change doesn't simplify the code but need to touch a lot of files.

1. the `CDCEtcdClient` can implement some interface:
See: https://github.com/pingcap/ticdc/blob/master/cdc/roles/storage/etcd.go#L48
there are some interfaces for etcd writer and reader, like `ChangeFeedRWriter`, 
and those interfaces are growing complicated, for example : 
There is a function named `Read` In `ChangeFeedRWriter` interface, returns A,B,C structs.
I want to split the function to many functions, like: readA, readB, readC.
the `CDCEtcdClient` can implement those function and we can avoid using a new struct.

2. some function about etcd writer and reader will be port to `ticdc-lib` repo
Suppose that a etcd client struct in `ticdc-lib` is named `CDCEtcdClient2`
We can composite `CDCEtcdClient` and `CDCEtcdClient2` and no other changes needed

leoppro:
@amyangfei  PTAL

PR #245, opened 2020-02-01
refine scan region coverage check

amyangfei:

#146 introduces a cover check for region split, but it does not consider the scenario when newly generated regions count is larger than `ScanRegions` limit.

check left most cover of splitted regions only

### Check List 

Tests 

 - Unit test
 - Integration test


leoppro:
LGTM

PR #251, opened 2020-02-07
etcd: split TaskStatus into two structs

leoppro:
reduce the probability of write conflicts about `TaskStatus`

split `TaskStatus` into `TaskStatus` and `TaskPosition`
The processor write ResolvedTS and CheckpointTs into `TaskPosition`, and rarely write `TaskStatus`,
so we can reduce the probability of write conflicts about `TaskStatus`
### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/kv/etcd.go
+++ b/cdc/kv/etcd.go
@@ -46,14 +46,24 @@ func GetEtcdKeyChangeFeedStatus(changefeedID string) string {
 	return fmt.Sprintf("%s/changefeed/status/%s", EtcdKeyBase, changefeedID)
 }
 
-// GetEtcdKeyTaskList returns the key of a task status without captureID part
-func GetEtcdKeyTaskList(changefeedID string) string {
-	return fmt.Sprintf("%s/changefeed/task/%s", EtcdKeyBase, changefeedID)
+// GetEtcdKeyTaskStatusList returns the key of a task status without captureID part
+func GetEtcdKeyTaskStatusList(changefeedID string) string {
+	return fmt.Sprintf("%s/changefeed/task/status/%s", EtcdKeyBase, changefeedID)
 }
 
-// GetEtcdKeyTask returns the key of a task status
-func GetEtcdKeyTask(changefeedID, captureID string) string {
-	return fmt.Sprintf("%s/%s", GetEtcdKeyTaskList(changefeedID), captureID)
+// GetEtcdKeyTaskPositionList returns the key of a task position without captureID part
+func GetEtcdKeyTaskPositionList(changefeedID string) string {
+	return fmt.Sprintf("%s/changefeed/task/position/%s", EtcdKeyBase, changefeedID)
+}
+
+// GetEtcdKeyTaskStatus returns the key of a task status
+func GetEtcdKeyTaskStatus(changefeedID, captureID string) string {
+	return fmt.Sprintf("%s/%s", GetEtcdKeyTaskStatusList(changefeedID), captureID)
+}
+
+// GetEtcdKeyTaskPosition returns the key of a task position
+func GetEtcdKeyTaskPosition(changefeedID, captureID string) string {
+	return fmt.Sprintf("%s/%s", GetEtcdKeyTaskPositionList(changefeedID), captureID)
 }
 
 // GetEtcdKeyCaptureList returns the prefix key of all capture info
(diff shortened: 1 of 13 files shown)


cdc/processor.go
```diff
@@ -388,42 +390,43 @@ func (p *processor) localResolvedWorker(ctx context.Context) error {
...
-		p.handleTables(ctx, oldInfo, p.status, oldInfo.CheckPointTs)
+		p.handleTables(ctx, oldStatus, p.status, p.position.CheckPointTs)
 		syncTableNumGauge.WithLabelValues(p.changefeedID, p.captureID).Set(float64(len(p.status.TableInfos)))
 
-		if len(oldInfo.TableInfos) > len(p.status.TableInfos) {
-			// some table is removed, we will not both remove and add table in one operation.
-			// keep CheckpointTs and ResolvedTs as the old in memory cache one.
-			p.status.CheckPointTs = oldInfo.CheckPointTs
-			p.status.ResolvedTs = oldInfo.ResolvedTs
+		err = p.tsRWriter.WriteInfoIntoStorage(ctx)
```
suzaku: Maybe we should move all the code that returns `Permanent` error outside the retried closure.

cdc/owner.go
```diff
@@ -105,20 +106,21 @@ func (c *changeFeed) String() string {
 	return s
 }
 
-func (c *changeFeed) updateProcessorInfos(processInfos model.ProcessorsInfos) {
-	for cid, pinfo := range processInfos {
+func (c *changeFeed) updateProcessorInfos(processInfos model.ProcessorsInfos, position map[string]*model.TaskPosition) {
+	for cid, pinfo := range position {
```
amyangfei: recommend to refine variable names in this function
s/position/positions
s/pinfo/position
s/oldPinfo/oldPosition

cdc/owner.go
```diff
@@ -438,6 +440,16 @@ func (o *ownerImpl) handleMarkdownProcessor(ctx context.Context) {
 			remainProcs = append(remainProcs, snap)
 			continue
 		}
+		err = kv.DeleteTaskPosition(ctx, o.etcdClient, snap.CfID, snap.CaptureID)
+		if err != nil {
+			log.Warn("failed to delete processor info",
```
amyangfei:
```suggestion
			log.Warn("failed to delete task position",
```
similar suggestion in L435, s/processor info/task status

cdc/owner.go
```diff
@@ -61,7 +61,7 @@ type OwnerDDLHandler interface {
 // ChangeFeedRWriter defines the Reader and Writer for changeFeed
 type ChangeFeedRWriter interface {
 	// Read the changefeed info from storage such as etcd.
-	Read(ctx context.Context) (map[model.ChangeFeedID]*model.ChangeFeedInfo, map[model.ChangeFeedID]*model.ChangeFeedStatus, map[model.ChangeFeedID]model.ProcessorsInfos, error)
+	Read(ctx context.Context) (map[model.ChangeFeedID]*model.ChangeFeedInfo, map[model.ChangeFeedID]*model.ChangeFeedStatus, map[model.ChangeFeedID]model.ProcessorsInfos, map[model.ChangeFeedID]map[model.CaptureID]*model.TaskPosition, error)
```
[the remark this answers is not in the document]
leoppro: yes, I solved this problem in #255

cdc/owner.go
```diff
@@ -523,9 +543,10 @@ func (o *ownerImpl) newChangeFeed(id model.ChangeFeedID, processorsInfos model.P
 	ddlHandler := newDDLHandler(o.pdClient, checkpointTs)
 
 	existingTables := make(map[uint64]uint64)
-	for _, taskStatus := range processorsInfos {
+	for captureID, taskStatus := range processorsInfos {
+		pos := taskPositions[captureID]
```
amyangfei: should we check whether `pos` is nil

cdc/processor.go
leoppro: why? I think we should retry when the write operation is conflict

leoppro: I see

cdc/kv/etcd.go
```diff
@@ -169,10 +179,34 @@ func (c CDCEtcdClient) SaveChangeFeedInfo(ctx context.Context, info *model.Chang
...
 
+// GetAllTaskPositions queries all task positions of a changefeed, and returns a map
+// mapping from captureID to TaskPositions
+func (c CDCEtcdClient) GetAllTaskPositions(ctx context.Context, changefeedID string, opts ...clientv3.OpOption) (map[string]*model.TaskPosition, error) {
+	key := GetEtcdKeyTaskPositionList(changefeedID)
+	resp, err := c.Client.Get(ctx, key, append([]clientv3.OpOption{clientv3.WithPrefix()}, opts...)...)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	positions := make(map[string]*model.TaskPosition, resp.Count)
```
suzaku: It seems `model.TaskPosition` is a tiny struct, we can use value instead of pointer for better performance?

leoppro: no, I think we should use pointer, If we try to update `ResolvedTS` or `CheckpointTS`, we should make sure that the new value will be set to `positionsMap`

PR #253, opened 2020-02-10
*: random cleanups

suzaku:

Some random cleanups done while read related code.

1. Change some variable names to improve readability
1. Add a missing error check
1. Reduce nested level of some code

### Check List 

Tests 

 - Unit test

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -360,17 +360,17 @@ func (c *CDCClient) divideAndSendEventFeedToRegions(
 func (c *CDCClient) singleEventFeed(
 	ctx context.Context,
 	span util.Span,
-	checkpoint uint64,
+	checkpointTs uint64,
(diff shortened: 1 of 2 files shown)


cdc/kv/client.go
```diff
@@ -361,24 +360,24 @@ func (c *CDCClient) divideAndSendEventFeedToRegions(
 func (c *CDCClient) singleEventFeed(
 	ctx context.Context,
 	span util.Span,
-	ts uint64,
-	meta *metapb.Region,
+	checkpoint uint64,
```
leoppro: hoe about `checkPointTs`?

suzaku: Yes. Note that checkpoint is a valid word, not two words, so we should use `checkpoint` instead of `checkPoint`.

PR #254, opened 2020-02-10
*: Merge similar structs

suzaku:

1. `RawKVEntry` is an alias of `RegionFeedValue`
1. `ResolvedSpan` contains exactly the same information as `RegionFeedCheckpoint`
1. `KvOrResolved` is just a disguised `RegionFeedEvent`

This PR aims to simplify the code by merging these structs.

Only one of the dupcalited structs is kept for each twin structs.
We no longer need to create a new struct and copy each field from the old one.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/cdc/model/kv.go
+++ b/cdc/model/kv.go
@@ -19,16 +19,16 @@ const (
 // RegionFeedEvent from the kv layer.
 // Only one of the event will be setted.
 type RegionFeedEvent struct {
-	Val        *RawKVEntry
-	Checkpoint *ResolvedSpan
+	Val      *RawKVEntry
+	Resolved *ResolvedSpan
 }
 
 // GetValue returns the underlying value
 func (e *RegionFeedEvent) GetValue() interface{} {
 	if e.Val != nil {
 		return e.Val
-	} else if e.Checkpoint != nil {
-		return e.Checkpoint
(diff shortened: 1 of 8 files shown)


cdc/model/kv.go
```diff
@@ -16,55 +16,40 @@ const (
...
-
-// ResolvedSpan represents that a span is resolved at a certain timestamp
-type ResolvedSpan struct {
-	Span      util.Span
-	Timestamp uint64
+// RegionFeedEvent from the kv layer.
+// Only one of the event will be setted.
+type RegionFeedEvent struct {
+	Val        *RawKVEntry
+	Checkpoint *ResolvedSpan
```
leoppro: why the name of this field is `Checkpoint`?

suzaku: I don't know why the name is chosen in the first place, it seems not a good name here.
But to avoid making the `diff` hard to review, I didn't rename it in this PR, maybe we can think of better names for these 3 structs involved later?

leoppro: it seems like the original name is `Resolved`, how about leave it as `Resolved`?

suzaku: Renamed.

PR #278, opened 2020-02-20
puller: make the Puller support to output sortedrow-based stream

leoppro:
make the Puller support to output sortedrow-based stream, preparing for kafka downstream.

1. rename RawTxn to RawRowGroup. RawRowGroup can include rows in one Txn or rows in incomplete Txn. The `IsCompleteTxn` field is used to mark if the rows in RawRowGroup are from a complete Txn.
2. Optimized the implementation of `entry_group`, support to output sorted row-based RawRowGroup (current implementation is not efficient enough)

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/model/txn.go
+++ b/cdc/model/txn.go
@@ -6,15 +6,17 @@ import (
 	"github.com/pingcap/tidb/types"
 )
 
-// RawTxn represents a complete collection of Entries that belong to the same transaction
-type RawTxn struct {
-	Ts         uint64
-	IsResolved bool
-	Entries    []*RawKVEntry
+// RawRowGroup represents a complete collection of Entries that belong to the same transaction
+type RawRowGroup struct {
+	Ts uint64
+	// IsCompleteTxn represents whether the RawRowGroup is a complete transaction
+	IsCompleteTxn bool
+	IsResolved    bool
+	Entries       []*RawKVEntry
 }
 
-// IsFake returns true if this RawTxn is fake txn.
-func (r RawTxn) IsFake() bool {
+// IsFake returns true if this RawRowGroup is fake txn.
+func (r RawRowGroup) IsFake() bool {
 	return len(r.Entries) == 0
 }
 
--- a/cdc/puller/entry_group.go
+++ b/cdc/puller/entry_group.go
@@ -38,7 +38,7 @@ func (eg *EntryGroup) AddEntry(ts uint64, entry *model.RawKVEntry) {
 	if i >= len(eg.sortedEntries) || eg.sortedEntries[i].Ts != ts {
 		eg.sortedEntries = append(eg.sortedEntries, nil)
 		copy(eg.sortedEntries[i+1:], eg.sortedEntries[i:])
-		eg.sortedEntries[i] = &model.RawRowGroup{Ts: ts, IsCompleteTxn: eg.collectTxn}
+		eg.sortedEntries[i] = &model.RawRowGroup{Ts: ts, IsCompleteTxn: true}
 	}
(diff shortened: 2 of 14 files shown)


cdc/model/txn.go
```diff
@@ -6,15 +6,17 @@ import (
 	"github.com/pingcap/tidb/types"
 )
 
-// RawTxn represents a complete collection of Entries that belong to the same transaction
-type RawTxn struct {
-	Ts         uint64
-	IsResolved bool
-	Entries    []*RawKVEntry
+// RawRowGroup represents a complete collection of Entries that belong to the same transaction
+type RawRowGroup struct {
```
suzaku: Note that the word `raw` in `RawTxn` is used to separate it from `Txn`, since it's no longer called a `txn`, you can remove the `raw` part.

cdc/puller/entry_group.go
```diff
@@ -21,43 +21,47 @@ import (
...
+	if i >= len(eg.sortedEntries) || eg.sortedEntries[i].Ts != ts {
+		eg.sortedEntries = append(eg.sortedEntries, nil)
+		copy(eg.sortedEntries[i+1:], eg.sortedEntries[i:])
+		eg.sortedEntries[i] = &model.RawRowGroup{Ts: ts, IsCompleteTxn: true}
 	}
+	eg.sortedEntries[i].Entries = append(eg.sortedEntries[i].Entries, entry)
 }
 
 // Consume retrieves all RawKVEntry with ts no larger than resolvedTs,
 // returns RawTxn list sorted by ts according these RawKVEntry and
```
suzaku: This comment should be updated.

cdc/puller/entry_group_test.go
```diff
@@ -61,13 +61,65 @@ func (s *mockEntryGroupSuite) TestEntryGroup(c *check.C) {
...
+		//		{IsCompleteTxn: false, Entries: []*model.RawKVEntry{{Ts: 4}, {Ts: 5}}},
+		//	},
+		//	remainEntry: 0,
+		//},
+		//{
+		//	input:       []*model.RawKVEntry{{Ts: 7}},
+		//	resolvedTs:  6,
+		//	expectTxns:  nil,
+		//	remainEntry: 1,
+		//},
```
suzaku: Remove commented code?

cdc/puller/entry_group.go
```diff
@@ -21,43 +21,47 @@ import (
...
+	for ; i < len(eg.sortedEntries); i++ {
+		if eg.sortedEntries[i].Ts > resolvedTs {
 			break
 		}
-		txns = append(txns, model.RawTxn{Ts: ts, Entries: eg.entriesMap[ts]})
-		delete(eg.entriesMap, ts)
+		txns = append(txns, *eg.sortedEntries[i])
 	}
-	eg.sortedTs = eg.sortedTs[i:]
+	if !eg.collectTxn {
```
suzaku: It seems this is never turned off now, maybe we can remove it and add it back with an efficient implementation when we do need it?

cdc/model/txn.go
leoppro: I will rename Txn to RowGroup in the next PR

cdc/puller/entry_group.go
leoppro: fine

cdc/puller/entry_group.go
leoppro: When the downstream is kafka, we will trun `collectTxn` off .
I will use it in the next pr

PR #287, opened 2020-02-25
puller: make puller support to output sorted kv entries

leoppro:
make puller support to output sorted kv entries

add a struct named `entry_sorter` to sort entries.

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/puller/entry_sorter.go
```diff
@@ -0,0 +1,121 @@
...
+		resolvedCh: make(chan uint64, 1024),
+		output:     make(chan *model.RawKVEntry, 128),
+	}
+}
+
+// Run runs the EntrySorter, sort the entries
+func (es *EntrySorter) Run(ctx context.Context) {
+	lessFunc := func(i *model.RawKVEntry, j *model.RawKVEntry) bool {
+		if i.Ts == j.Ts {
+			return i.OpType == model.OpTypeDelete && j.OpType == model.OpTypePut
```
amyangfei:
```suggestion
			return i.OpType == model.OpTypeDelete
```
only  check`entry-i` is `delete` can save one comparison, and meet our requirement: `always put Delete event before Put event`. But not accurate `less than` semantic

PR #290, opened 2020-02-26
sink: refine mysql log, remove redundant SQL count log

amyangfei:

sink prints too much logs in each exec DML loop

- remove log for each time that MySQL sink executes DMLs, instead of a periodically print status

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/sink/mysql.go
+++ b/cdc/sink/mysql.go
@@ -25,10 +25,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"golang.org/x/sync/errgroup"
-
-	"github.com/pingcap/ticdc/pkg/retry"
-
 	dmysql "github.com/go-sql-driver/mysql"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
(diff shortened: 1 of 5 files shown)


cdc/sink/mysql.go
```diff
@@ -345,6 +353,34 @@ func (s *mysqlSink) Close() error {
 	return nil
 }
 
+func (s *mysqlSink) PrintStatus(ctx context.Context) error {
+	changefeedID := util.ChangefeedIDFromCtx(ctx)
```
leoppro: if the caller of sink is owner, could we get changefeedId from context?

how about store a value in context, in order to get the role of caller?

leoppro: by the way, maybe we can storage `changefeedId` in `mysqlSink` struct to avoid query changedfeedId  continually

amyangfei: Add more options in `sink opt`, as `changefeedID` and `captureID` are used quite often. The `role` is not put into sink now, as we only call `PrintStatus` of the sink in processor. We can add the role information if we really need it.

PR #291, opened 2020-02-27
Fix lowercase schema name in ddl mounter

amyangfei:

QA found a bug that CDC couldn't replicate `CREATE TABLE` when the schema name is not all lowercase.
The root cause is TiDB stores the `CREATE TABLE` DDL history with a lowercase schema name.

Find the correct schema name from the `SchemaStorage` when we mount a DDL

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/entry/mounter.go
+++ b/cdc/entry/mounter.go
@@ -207,6 +207,14 @@ func fetchHandleValue(tableInfo *schema.TableInfo, recordID int64) (pkCoID int64
 
 func (m *Mounter) mountDDL(jobEntry *ddlJobKVEntry) (*model.DDL, error) {
 	databaseName := jobEntry.Job.SchemaName
+	// schema name in raw ddl job entry is always lowercase, try to find the
+	// correct value from schema storage
+	if m.schemaStorage != nil {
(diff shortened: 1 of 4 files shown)


cdc/entry/mounter.go
```diff
@@ -207,6 +207,14 @@ func fetchHandleValue(tableInfo *schema.TableInfo, recordID int64) (pkCoID int64
 
 func (m *Mounter) mountDDL(jobEntry *ddlJobKVEntry) (*model.DDL, error) {
 	databaseName := jobEntry.Job.SchemaName
+	// schema name in raw ddl job entry is always lowercase, try to find the
+	// correct value from schema storage
+	if m.schemaStorage != nil {
```
leoppro: we can find origin schema name in `job.BinlogInfo.DBInfo.Name.O`

amyangfei: I'm not sure, but from TiDB DDL history the `job.BinlogInfo.DBInfo` is nil

Screen Shot 2020-02-27 at 19 11 21

leoppro:
LGTM

PR #296, opened 2020-02-28
processor: handle row-based changed event in processor, part three

leoppro:
handle row-based changed event in processor

* processor receives globalResolvedTS and sends to sink, but not to block row changed event
* every `tableInfo` holds a mounter

### DIFF
--- a/cdc/processor.go
+++ b/cdc/processor.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"math"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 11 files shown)


cdc/processor.go
```diff
@@ -556,177 +554,65 @@ func (p *processor) globalResolvedWorker(ctx context.Context) error {
...
-			if len(txn.DMLs) == 0 {
-				continue
-			}
-			pendingTxns = append(pendingTxns, txn)
-			if len(pendingTxns) >= bulkLimit {
-				if err := flush(ctx); err != nil {
-					return errors.Trace(err)
-				}
-			}
+			// should filter in sink
```
amyangfei: Consider the following strategies, which is better:
1. filter events in the processor, don't emit ignored `RowChangedEvent` to sink
2. filter events in the processor, emit `RowChangedEvent` with an ignorable flag to sink
3. do nothing in the processor, filter event in the sink

leoppro: I think `option 3` is more flexibility, what do you think about it

amyangfei: `option 3` seems better
- different sink may have different filter strategy, especially for the DDL compatibility
- option 3 keeps a complete event set

PR #297, opened 2020-02-28
Configurable worker count

### LINKED ISSUE #273: support confige concurrency when replication data to downstream mysql
https://github.com/pingcap/ticdc/blob/d2621b3f0f65f33567fa6bf772b93e8b2aee1128/cdc/sink/mysql.go#L122

consider just use pkg/loader from tidb-binlog

july2993:
fix #273 

- treat uri as truely uri but compatible with the dsn
- support `worker-count` param
- set `SetMaxIdleConns`(default only 2 will cause create new connections every time)  and `SetOpenConns`

### Check List 

Tests 

 - Unit test

Code changes

### DIFF
--- a/cdc/sink/mysql.go
+++ b/cdc/sink/mysql.go
@@ -17,6 +17,8 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"net/url"
+	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


leoppro:
LGTM

Issue #269, opened 2020-02-19
make a better processor aliveness check and ensure RTO < 30s


amyangfei:
## Feature Request

**Is your feature request related to a problem? Please describe:**

We have s simple processor aliveness check, basically check either resolvedTs or checkpointTs is updated in one minute, which doesn't meet the requirement of RTO < 30s

In some tests we found kv client may block or with some other reasons, the resolvedTs and checkpointTs can't be updated in time, which means the **not real-time of replication status doesn't always mean abnormal of a processor**

**Describe the feature you'd like:**

Design a better aliveness check strategy, which satisfies

- can find a processor abnormal (have many scenarios, like network partition, capture node crashes etc.) and rebalance tables belong to it to other processors less than 30s.

amyangfei:
Optimized by https://github.com/pingcap/ticdc/pull/314 and https://github.com/pingcap/ticdc/pull/319

PR #299, opened 2020-02-29
Reduce the RTO by watching the liveness of processors

shafreeck:

The owner periodically checks the task status and marks the processor down if the task is not updated in time.  When a processor is marked down, the task associated with it is rebalanced to other processors.
As issue #269 says, when there is something wrong with the upstream(TiDB/TiKV) or network, the processor receives no more data to update the task status, the processor will be marked as down which is a false positive.

The PR introduces a session for each processor that keeps alive with a lease. Each processor is assigned an ID which constructs a key registered in the Etcd. When a processor stops, the lease expires, and the key will be deleted. The owner keeps watching on the key， and when a key is found deleted, the processor is marked as down.

### Current problems and the future plan
* The session for capture is only used for the election. It is not attached to the CaptureID, so when a capture crashed, it could not be detected by the owner.
* It is not necessary to create a session for each processor, it has the same lifetime of the capture, so it should use the session of the capture.

Create one session for the capture and then it can be shared with the election and processors.

* Create a session when a processor started.
* Assign an ID for each processor
* Mark the processor down when a processor is stopped by watch the deletion of the processor key.
* Add `query-processor-list` for ctl command.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change (new added)

### DIFF
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -403,6 +403,7 @@ func NewOwner(pdEndpoints []string, cli kv.CDCEtcdClient, manager roles.Manager)
 		pdEndpoints:        pdEndpoints,
 		pdClient:           pdClient,
 		changeFeeds:        make(map[model.ChangeFeedID]*changeFeed),
+		activeProcessors:   make(map[string]*model.ProcessorInfo),
 		cfRWriter:          cli,
 		etcdClient:         cli,
 		manager:            manager,
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -991,7 +991,8 @@ func (o *ownerImpl) Run(ctx context.Context, tickTime time.Duration) error {
 		}
 	}()
 
-	var once sync.Once
+	// ownerChanged
+	ownerChanged := true
 	for {
 		select {
 		case <-ctx.Done():
@@ -1000,11 +1001,12 @@ func (o *ownerImpl) Run(ctx context.Context, tickTime time.Duration) error {
 			return errors.Annotate(err, "handleWatchCapture failed")
 		case <-time.After(tickTime):
 			if !o.IsOwner(ctx) {
+				ownerChanged = true
 				continue
 			}
-
-			// Do something initialize when the capture becomes an owner.
-			once.Do(func() {
+			if ownerChanged {
+				// Do something initialize when the capture becomes an owner.
+				ownerChanged = false
 				// Start a routine to keep watching on the liveness of
 				// processors.
 				o.startProcessorInfoWatcher(ctx)
@@ -1015,7 +1017,8 @@ func (o *ownerImpl) Run(ctx context.Context, tickTime time.Duration) error {
 					log.Error("clean up stale tasks failed",
 						zap.Error(err))
 				}
-			})
+			}
+
 			err := o.run(ctx)
 			// owner may be evicted during running, ignore the context canceled error directly
 			if err != nil && errors.Cause(err) != context.Canceled {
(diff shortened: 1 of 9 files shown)


cdc/owner.go
```diff
@@ -1021,6 +1002,20 @@ func (o *ownerImpl) Run(ctx context.Context, tickTime time.Duration) error {
...
+
+			// Do something initialize when the capture becomes an owner.
+			once.Do(func() {
+				// Start a routine to keep watching on the liveness of
+				// processors.
+				o.startProcessorInfoWatcher(ctx)
+
+				// When an owner crashed, its processors crashed too,
+				// clean up the tasks for these processors.
+				if err := o.cleanUpStaleTasks(ctx); err != nil {
```
amyangfei: Should `cleanUpStaleTasks` be called every time when a capture becomes owner?
Consider the following scenario:
1. start two captures
2. resign owner twice
3. kill -9 `the pid of current owner`

Found some tables are not re-balanced

shafreeck: Good catch, I will fix it.

cdc/owner.go
```diff
@@ -1103,3 +1108,194 @@ func (o *ownerImpl) writeDebugInfo(w io.Writer) {
...
+	ctx = clientv3.WithRequireLeader(ctx)
+
+	rev, processors, err := o.etcdClient.GetAllProcessors(ctx)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	// before watching, rebuild events according to
+	// the existed processors. This is necessary because
+	// the etcd events may be compacted.
+	if err := o.rebuildProcessorEvents(ctx, processors); err != nil {
```
amyangfei: Consider the scenario
1. we have two captures
2. the owner capture crashes (kill -9) and the second capture becomes the owner
3. the processor info of the original owner will be deleted because of session expired
4. The second capture takes a longer time to be an owner than the processor info expired, in the current logic, the new owner doesn't aware of the crashed processor when it `rebuildProcessorEvents`.

cdc/owner.go
```diff
@@ -1103,3 +1108,194 @@ func (o *ownerImpl) writeDebugInfo(w io.Writer) {
...
+			// be nil, it may be caused by a temporary error or a context
+			// error(ctx.Err())
+			if ctx.Err() != nil {
+				// The context error indicates the termination of the owner
+				log.Error("watch processor failed", zap.Error(ctx.Err()))
+				return
+			}
+			log.Warn("watch processor returned", zap.Error(err))
+			// Otherwise, a temporary error occured(ErrCompact),
+			// restart the watching routine.
```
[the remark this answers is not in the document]
shafreeck: Sorry for forgetting the for clause.

cdc/owner.go
shafreeck: Yes, the events in that scenario would not be detected, but it is OK here. The case has been covered in the `cleanUpStaleTasks` function. It checks the tasks with the active processors, and if the processor associated with the task is not active, the task will be rebalanced. When an owner starts to be in charge, the processors that have expired would not be loaded as an active processor, the processors that being loaded would not lose the expiration events.

amyangfei: Oh yes you are right. I find the bug does not locate here but in the `CaptureInfoWatcher`.

Still in this scenario, after the second capture becomes owner, it rebalances tables based on the `capture list` in `o.captures`, but the capture info of the crashed capture is not deleted, so some tables are rebalanced to the crashed capture.

The implementation of `CaptureInfoWatcher` doesn't consider the owner role very well, it should also have some similar logic like `ProcessorInfoWatcher`. Besides these two watchers may do some conflict works, should we combine these two watchers or leave them alone to do different jobs, what's your suggestion

shafreeck: There are two problems with the capture watcher, the first one as you said, the capture info is not deleted. The second one is all the roles including the non-owner captures are watching the events, this may handle the capture deletion multiple times.

I am planning to solve these problems by refactoring the capture:
* Use one session for the capture self, its election, and all its processors.
* Remove the capture info watcher(I am not sure is it safe to remove it)

amyangfei: Sounds great! we can refine this in some more PRs.

shafreeck: Yes, agree with you.

shafreeck:
@amyangfei PTAL

PR #308, opened 2020-03-04
client: Implement batch client

### LINKED ISSUE #292: Support batch stream in kv client
## Feature Request

**Is your feature request related to a problem? Please describe:**

The original TiKV EventFeed API will be changed to a duplex stream. ref:
https://docs.google.com/document/d/1SN3ztOXy2QTlCS1Qp9dUWTBfxowx-nIGpkuCw2ccULM/edit

**Describe the feature you'd like:**

The kv client has a clear input and ouput:
- input: a key range
- output: kv change log from all regions located in the given kv range

Things need to be done:

- We need to refine our kv client in CDC to support new API. 
- Have a better error handling in kv client

### LINKED ISSUE #303: EventFeed may miss some regions when region is splitting
The function `partialRegionFeed` (in cdc/kv/client.go) accepts a region info from parameter and may reload region info from region cache before sending request. So it's possible that the region has changed after a split. As a result it will get a smaller region after calling `regionCache.LocateKey` that doesn't cover the range hold by the parameter `regionInfo`, so the remaining part of the range will be missing.

MyonKeminta:

This PR implements batch client to reuse one stream for multiple regions.
Fixes #292 
Also fixes #303

This PR switches kvproto to https://github.com/5kbpers/kvproto/tree/cdc-batch-service . For each invoking to `EventFeed`, at most one stream will be connected to each TiKV instance. 

There will be one sender goroutine that dispatches requests to each stream, one receiver goroutine **for each stream** receives the events from it. However, different from the original design, there are still one goroutine for each region. The receiver goroutines doesn't process the events by themselves, instead, they dispatches the events to the goroutines for each region. It's because to maintain the region's states in a single goroutine is complicated, and to make it work as soon as possible, currently these goroutines are still kept. I'll continue working on further refactoring and optimizing.

In this PR, error handling still needs polishing.

### Check List 

Tests 

 - Integration test

Side effects

 - Increased code complexity


cdc/kv/client.go
```diff
@@ -390,49 +587,180 @@ func (c *CDCClient) divideAndSendEventFeedToRegions(
...
+// info will be sent to `regionCh`.
+func (c *CDCClient) handleError(ctx context.Context, errInfo regionErrorInfo, regionCh chan<- singleRegionInfo) error {
+	err := errInfo.err
+	switch eerr := errors.Cause(err).(type) {
+	case *eventError:
+		innerErr := eerr.err
+		if notLeader := innerErr.GetNotLeader(); notLeader != nil {
+			eventFeedErrorCounter.WithLabelValues("NotLeader").Inc()
+			// TODO: Handle the case that notleader.GetLeader() is nil.
+			c.regionCache.UpdateLeader(errInfo.verID, notLeader.GetLeader().GetStoreId(), errInfo.rpcCtx.PeerIdx)
```
amyangfei: will this panic if `notLeader.GetLeader()` is nil ?

MyonKeminta: I think it's ok, because TiDB didn't check it either, and `GetStoreId`'s implementation checked whether it's nil and returns 0 in that case. When `UpdateLeader` receives 0 as the `StoreId` it will simply try to send the request to the next peer. I think maybe I need to remove the TODO comment.

MyonKeminta: Oh, no. TiDB uses different backoff time according to whether it's nil. We need to process it later, but it won't panic.

Issue #265, opened 2020-02-19
refine replication forward model in CDC


amyangfei:
In CDC we have the following replication model
```
1. kv client recvs data
2. kv client sends data to puller via an event chan
3. puller adds data to a buffer, sorts data and re-constructs transactions
4. puller sends transactions to tableInfo (managed in a processor) via a txn chan
5. procssor pulls all txns from txn chan of each tableInfo (with txn ts no more than CDC GlobalResolvedTs)
```

In the large number of regions test, we found the replication blocked with `buffer in 3`, `chan in 4` full and no data was pulled in step 5. Maybe this is also part reason for slow replication and low throughput. We should have a better data forward model, for the following consideration:

1. table level sort and data flow should not block the kv client pulling messages from TiKV
2. If we use MQ such as Kafka as sink, we even don't need the transaction reconstruction in step 3 (sort in step 3 is enough).

We can separate this refactor into multiple small changes, including:

1. implement a new memory cache mechanism, always receive messages from TiKV and buffer messages in memory if they are not consumed fast enough.
2. limit memory usage in optimization 1
3. we can forward messages to the sink module regardless of the global resolved ts, and only consider the global resolved ts when we really replicate data to the target.

PR #310, opened 2020-03-04
sink: handle row-based changed event in sink, part four

### LINKED ISSUE #272: Sink interface should accept row and resolve event
so can support some sink don't care txn.
the implementation of the sink can reconstruct txn by resolve event internally if need.

leoppro:
fix #272 , #265 

this pr can pass the integration test now

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

 - Need to update the `tidb-cdc/cdc-ansible`
 - Need to be included in the release note

### DIFF
--- a/cdc/entry/schema_builder.go
+++ b/cdc/entry/schema_builder.go
@@ -9,14 +9,12 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff"
-	"github.com/pingcap/ticdc/pkg/retry"
-
-	"go.uber.org/zap"
-
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	timodel "github.com/pingcap/parser/model"
 	"github.com/pingcap/ticdc/cdc/model"
+	"github.com/pingcap/ticdc/pkg/retry"
+	"go.uber.org/zap"
 )
 
 type jobList struct {
@@ -136,14 +134,7 @@ func (b *StorageBuilder) Run(ctx context.Context) error {
 		if rawKV == nil {
 			return errors.Trace(ctx.Err())
 		}
-		job, _ := UnmarshalDDL(rawKV, false)
-		if job != nil {
-			log.Info("storage builder accept", zap.Reflect("job", job))
-		}
 		if rawKV.Ts < b.resolvedTs {
-			if job != nil {
-				log.Info("storage builder drop", zap.Reflect("job", job))
-			}
 			continue
 		}
 
@@ -153,7 +144,7 @@ func (b *StorageBuilder) Run(ctx context.Context) error {
 			continue
 		}
 
-		job, err := UnmarshalDDL(rawKV, true)
+		job, err := UnmarshalDDL(rawKV)
 		if err != nil {
 			return errors.Trace(err)
 		}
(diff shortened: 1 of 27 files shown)


cdc/entry/schema_builder.go
```diff
@@ -6,33 +6,58 @@ import (
...
-		currentJob = l.list.Front()
+
+	if ts < l.gcTs {
+		log.Fatal("cannot fetch the jobs which of finishedTs is less then gcTs, please report a bug", zap.Uint64("gcTs", l.gcTs))
 	}
-	for ; currentJob != nil; currentJob = currentJob.Next() {
+
+	if currentJob.Value != (*timodel.Job)(nil) {
 		job := currentJob.Value.(*timodel.Job)
+		if job.BinlogInfo.FinishedTS <= l.gcTs {
```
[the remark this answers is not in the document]
leoppro: no, `gcTs` could between` currentJob.FinishedTs` and `nextJob.FinishedTs`

when the `currentJob` is the lastest consumed job, this job maybe dropped already, so we need to check whether the `currentJob` is deleted.

cdc/sink/mysql.go
```diff
@@ -36,37 +37,168 @@ import (
...
+	return atomic.LoadUint64(&s.checkpointTs)
+}
+
+func (s *mysqlSink) Run(ctx context.Context) error {
+	execRows := func(rowsMap map[string][]*model.RowChangedEvent) error {
+		groups := make([][]*model.RowChangedEvent, 0, len(rowsMap))
+		for _, rows := range rowsMap {
+			groups = append(groups, rows)
+		}
+		return s.concurrentExec(ctx, groups)
```
amyangfei: we don't need to convert the `map[string][]*model.RowChangedEvent` to `[][]*model.RowChangedEvent`, we can change the parameter of `concurrentExec` and pass the map to it directly

cdc/processor.go
```diff
@@ -248,7 +177,7 @@ func NewProcessor(
 	}
 
 	for _, table := range p.status.TableInfos {
-		p.addTable(ctx, int64(table.ID), table.StartTs)
+		go p.addTable(ctx, int64(table.ID), table.StartTs)
```
amyangfei: why using independent routines here

leoppro: A. we create a  processor follow the steps below:
1. create ddl puller
2. create schema builder
3. call `addTable` function
4. call `processor.Run` function, the `Run` of ddl puller will be called in this step. 

B. in `addTable` function
1. call `scheam builder` to create scheam storage, and `schema builder` will call `HandlePreviousDDLJobIfNeed` of new schema storage.
2. do something ...

In step `B.1`, if schema builder is unresolved,  `HandlePreviousDDLJobIfNeed` will return a `ErrUnresolved` error, and schema builder will retry and wait for it.
but ddl puller is still not running, so schema builder will never be resolved.

amyangfei: Oh I see. 
`addTable` relies on `ddl puller`

amyangfei: The processor startup procedure is not very clear, we can refine this later.

PR #315, opened 2020-03-05
bug fix: panic about schema builder

leoppro:
when we add a table with start-ts, we can't guarantee that `start-ts` is less than `checkpointts`,  `DoGC` function of `schema builder` will panic

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/entry/schema_builder.go
+++ b/cdc/entry/schema_builder.go
@@ -189,8 +189,9 @@ func (b *StorageBuilder) GetResolvedTs() uint64 {
 
 // DoGc removes the jobs which of finishedTs is less then gcTs
 func (b *StorageBuilder) DoGc(ts uint64) error {
-	if ts > atomic.LoadUint64(&b.resolvedTs) {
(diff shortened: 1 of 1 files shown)


cdc/entry/schema_builder.go
```diff
@@ -190,7 +190,8 @@ func (b *StorageBuilder) GetResolvedTs() uint64 {
 // DoGc removes the jobs which of finishedTs is less then gcTs
 func (b *StorageBuilder) DoGc(ts uint64) error {
 	if ts > atomic.LoadUint64(&b.resolvedTs) {
-		log.Fatal("gcTs is greater than resolvedTs in StorageBuilder, please report a bug", zap.Uint64("gcTs", ts))
+		log.Warn("gcTs is greater than resolvedTs in StorageBuilder", zap.Uint64("gcTs", ts))
```
amyangfei: Also log the `reslovedTs`?

leoppro: ok

PR #319, opened 2020-03-06
Refactor/capture watcher

shafreeck:

Fix the problem that the rebalance does not work sometimes.

* Keep the processor status when a processor is stopped because it would be handled by the owner
* Fix a bug in `cleanUpStaleTasks` for using a wrong key
* Keep the processor status when a capture deletion is detected, it should be handled by the processor watcher.

### Check List 

Tests 

 - Unit test
 - Integration test


leoppro:
LGTM

PR #322, opened 2020-03-09
sink: implement kafka sink

leoppro:
implement kafka sink

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/sink/black_hole.go
```diff
@@ -0,0 +1,56 @@
...
+	log.Info("BlockHoleSink: Checkpoint Event", zap.Uint64("checkpoint ts", ts))
+	return nil
+}
+
+func (b *blackHoleSink) EmitRowChangedEvent(ctx context.Context, rows ...*model.RowChangedEvent) error {
+	for _, row := range rows {
+		if row.Resolved {
+			atomic.StoreUint64(&b.checkpointTs, row.Ts)
+		}
+		log.Info("BlockHoleSink: Row Changed Event", zap.Any("row", row))
```
amyangfei:
```suggestion
```
There will be too many logs, suggest to make it more like a blackhole.

cdc/sink/mq.go
```diff
@@ -0,0 +1,167 @@
...
+func (k *mqSink) Run(ctx context.Context) error {
+	for {
+		var sinkCheckpointTs uint64
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case sinkCheckpointTs = <-k.sinkCheckpointTsCh:
+		}
+		globalResolvedTs := atomic.LoadUint64(&k.globalResolvedTs)
+		if globalResolvedTs < sinkCheckpointTs {
```
amyangfei: In which case will `globalResolvedTs` be less than `sinkCheckpointTs`

cdc/sink/mysql.go
```diff
@@ -80,12 +80,40 @@ func (s *mysqlSink) EmitRowChangedEvent(ctx context.Context, rows ...*model.RowC
 		//	continue
 		//}
 		key := util.QuoteSchema(row.Schema, row.Table)
+		checkDecode(row)
```
amyangfei: Is this debug code?

cdc/sink/sink.go
```diff
@@ -15,65 +15,53 @@ package sink
...
+	}
+	switch strings.ToLower(sinkURI.Scheme) {
+	case "blackhole":
+		return newBlackHoleSink(), nil
+	case "mysql", "tidb":
+		return newMySQLSink(sinkURI, nil, opts)
+	case "kafka":
+		return newKafkaSaramaSink(sinkURI)
+	default:
+		return nil, errors.New(fmt.Sprintf("the sink scheme (%s) is not supported", sinkURI.Scheme))
```
amyangfei:
```suggestion
		return nil, errors.Errorf("the sink scheme (%s) is not supported", sinkURI.Scheme)
```

cdc/sink/mysql.go
leoppro: yes, i will remove it

cdc/sink/mq.go
leoppro: https://github.com/pingcap/ticdc/pull/322/files#diff-172dd021d09a9fac1ffc69e5792d4c5bR83

sinkCheckpointTs is localResolvedTs in the processor, when we add a table, old local resolvedTS(min resolvedTs of old tables) may be greater than resolvedTS of the new table. At this moment, local resolvedTS will be fallback.

when local resolvedTS is fallback, we will postpone to pushing global resolvedts:
https://github.com/pingcap/ticdc/blob/master/cdc/processor.go#L526

leoppro:
PTAL again @amyangfei

cdc/sink/mq.go
amyangfei: Oh I see. Also adding some comments here.

leoppro: added

PR #323, opened 2020-03-09
bug fix: panic in storagebuilder sometimes

leoppro:
panic in storagebuilder sometimes

The information exchange between owner and processor is dependent on `TaskPositions` and `TaskStatus`.
`TaskStatus` is updated by the owner mainly, `TaskPositions` is updated by processors.
When the owner adds a table to the processor, the owner will update `TaskStatus`.  When the processor finds the change of `TaskStatus`, the processor will start table puller and update `TaskPositions`.

If the `TaskStatus` is already updated, but `TablePostitions` is still not updated by the processor, the owner may ignore the checkpointTS of the newest table.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -719,6 +719,8 @@ func (c *changeFeed) calcResolvedTs() error {
 
 	if len(c.taskPositions) == 0 {
 		minCheckpointTs = c.status.CheckpointTs
+	} else if len(c.taskPositions) < len(c.taskStatus) {
+		return nil
 	} else {
 		// calc the min of all resolvedTs in captures
 		for _, position := range c.taskPositions {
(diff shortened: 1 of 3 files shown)


cdc/owner.go
```diff
@@ -719,6 +719,8 @@ func (c *changeFeed) calcResolvedTs() error {
 
 	if len(c.taskPositions) == 0 {
 		minCheckpointTs = c.status.CheckpointTs
+	} else if len(c.taskPositions) < len(c.taskStatus) {
```
amyangfei: `len(c.taskPositions) < len(c.taskStatus)` can detect the scenario that a new processor which hasn't written `task position` to etcd.
But what happens when a processor already exists, and it is assigned a new table and the checkpoint ts of the newly added table is not considered

PR #327, opened 2020-03-09
kafka sink: support partition by rowid or unique value

leoppro:
support partition by rowid or unique value

[DMN] we should merge this pr after testing.

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/entry/schema_storage.go
```diff
@@ -69,6 +69,7 @@ type TableInfo struct {
 	IndicesOffset map[int64]int
 	UniqueColumns map[int64]struct{}
 	handleColID   int64
+	IndieMarkCol  string
```
amyangfei: What is `IndieMarkCol` short for, better to add some comment

cdc/entry/schema_storage.go
```diff
@@ -105,6 +110,17 @@ func WrapTableInfo(info *timodel.TableInfo) *TableInfo {
 				ti.UniqueColumns[ti.Columns[col.Offset].ID] = struct{}{}
 			}
 		}
+		if idx.Primary || idx.Unique {
```
amyangfei: Is `primary key` be counted twice, as L104 has increased once.

leoppro: no, if pk is handle, `primary index` will be not included in `ti.Indices`.

PR #334, opened 2020-03-11
cmd: refactor command line interface

amyangfei:

Fix https://github.com/pingcap/ticdc/issues/324

- Combine `cli` and `ctrl` command, and use the sub command style. We can find all new sub commands by `cdc cli -h`, and the most commonly used commd for creating a new changefeed is `cdc cli changefeed create` now.
- Unify old `pd-addr`, `pd-endpoints` to `pd` 
- Remove outdated command `cli pull`

### Check List 

Tests 

 - Unit test
 - Integration test


cmd/client.go
```diff
@@ -25,106 +24,134 @@ import (
...
-		if len(configFile) > 0 {
-			if err := strictDecodeFile(configFile, "cdc", cfg); err != nil {
+func newCreateChangefeedCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "create",
+		Short: "create a new replication task (changefeed)",
+		Long:  ``,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
+			id := uuid.New().String()
```
leoppro: maybe we can receive a change feed id.
if the user doesn't  specify a change feed id, we create a UUID as the default value

leoppro: note that we should check  the uniqueness of changefeedID if we allow the user to specify a changefeedID

cmd/client.go
```diff
@@ -25,106 +24,134 @@ import (
...
+				TargetTs:   targetTs,
+				Config:     cfg,
 			}
 
-			var key string
-			var value string
+			for _, opt := range opts {
+				s := strings.Split(opt, "=")
+				if len(s) <= 0 || len(s) > 2 {
+					fmt.Printf("omit opt: %s", opt)
```
leoppro: should `break` here?

cmd/client.go
shafreeck: Why do you think a customized change feed ID is necessary? If it is necessary, maybe we can let the user give it a name and description?

cmd/client.go
shafreeck: Using `cmd.Printf` is better.

shafreeck: If we want to omit the invalid option, a `continue` should be here.

cmd/client.go
```diff
@@ -25,106 +24,134 @@ import (
...
-			return err
-		}
-		fmt.Printf("create changefeed ID: %s detail %s\n", id, d)
-		return cli.SaveChangeFeedInfo(ctx, detail, id)
-	},
+			d, err := info.Marshal()
+			if err != nil {
+				return err
+			}
+			fmt.Printf("create changefeed ID: %s info %s\n", id, d)
```
shafreeck: Ditto

cmd/ctrl.go
```diff
@@ -133,49 +101,96 @@ var ctrlCmd = &cobra.Command{
...
-	},
+			return nil
+		},
+	}
+	return command
+}
+
+func newTruncateCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "truncate",
```
shafreeck: `truncate` is a little confused, is `destroy` a better word?

cmd/client.go
amyangfei: This PR mainly changes the interaction of the command line, I think we can refine the logic later.

amyangfei:
> Just be curious:
> 
> * Why do we need two commands: `cdc cli` and `cdc ctrl`?
> * Can they be combined into one command?
> 
> And some personal suggestions
> 
> * Should we use a `command <action> <resource>` style to handle the commands, like: `cdc ctrl list changefeed`, `cdc ctrl query processor`
> * Give a resource name that is created, like `cdc cli create changefeed ...`

> Why do we need two commands: `cdc cli` and `cdc ctrl`? Can they be combined into one command?

Yes, in fact this PR does combine these two commands, into a single `cdc cli`

> Should we use a `command <action> <resource>` style to handle the commands, like: `cdc ctrl list changefeed`, `cdc ctrl query processor`

Good idea. I have changed the command style, to make it more resource oriented. While we have different operation for resource, the usage is `cdc cli <resouce> <action>`, such as 

- cdc cli changefeed create
- cdc cli changefeed list
- cdc cli changefeed query

amyangfei:
PTAL @leoppro  @shafreeck

PR #337, opened 2020-03-12
capture: implement the task watcher

shafreeck:

A task watcher waits for new tasks and returns the
task event, then the capture can start a processor
to handle the task.

It is a replacement of the ChangeFeedWatcher and
ProcessorWatcher, the new solution would be more
clear and simpler.

#### Code changes

* Remove the Change Feed Watcher and the Processor Info Watcher
* Add a TaskWatcher watching on /tidb/cdc/task/{captureid}
* Change change feed status key to /tidb/cdc/job/{changefeedid}
* Add integration test scripts

The PR is work in progress
- [x] Add a task watcher to handle new tasks
- [x] Dispatch tasks to a capture
- [x] Fix/Add the unit tests
- [x] Integration test
  * kill servers
  * stop & resume a job
- [x] Add integration test scripts

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects


shafreeck:
@amyangfei PTAL

Notice that the integration test of `availability` failed when to restart the cdc server, it is not a problem of HA. I think it is a bug of cdc because it occurs on the master branch as well. So this bug should not be a barrier to this PR.

cdc/task.go
```diff
@@ -0,0 +1,229 @@
...
+
+	wch := etcd.Watch(ctx, w.cfg.Prefix,
+		clientv3.WithPrefix(),
+		clientv3.WithPrevKV(),
+		clientv3.WithRev(resp.Header.Revision+1))
+	for wresp := range wch {
+		if wresp.Err() != nil {
+			goto restart
+		}
+		for _, ev := range wresp.Events {
```
amyangfei: should `w.events` be updated when receiving a new event from the watch channel?

cdc/task.go
```diff
@@ -0,0 +1,229 @@
...
+			if ev.Type == clientv3.EventTypePut {
+				task, err := w.parseTask(ctx, ev.Kv.Key, ev.Kv.Value)
```
leoppro: can we reuse L145-L166 and L100-L120

cdc/task.go
```diff
@@ -0,0 +1,229 @@
...
+	"errors"
+
+	"github.com/pingcap/log"
+	"github.com/pingcap/ticdc/cdc/model"
+	"go.etcd.io/etcd/clientv3"
+	"go.uber.org/zap"
+)
+
+// TaskEventOp is the operation of a task
+type TaskEventOp string
```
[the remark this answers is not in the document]
shafreeck: I think that string is simple and more readable for users and logs. Performance is not necessary to consider here.

shafreeck:
PTAL @amyangfei @leoppro

cdc/scheduler.go
```diff
@@ -357,23 +58,21 @@ func realRunProcessor(
...
-	}
-
 	processor.Run(ctx, errCh)
 
 	go func() {
 		err := <-errCh
-		if cb != nil {
-			cb.OnStopProcessor(processor, err)
-		}
+		log.Error("error on running processor",
```
amyangfei: If the `errors.Cause(err)` is `context.Cancel`, we should ignore it

shafreeck: Why should be ignored?

amyangfei:
PTAL @leoppro

shafreeck:
ping @leoppro

leoppro:
please merge master branch

PR #338, opened 2020-03-12
sink: fix parse DSN as URI by mistake

leoppro:
fix parse DSN as URI by mistake

We add a scheme name for DSN string, and distinguish between URI and DSN by scheme name.

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/sink/sink.go
```diff
@@ -50,17 +50,27 @@ type Sink interface {
...
+const DSNScheme = "dsn://"
+
 // NewSink creates a new sink with the sink-uri
 func NewSink(sinkURIStr string, filter *util.Filter, opts map[string]string) (Sink, error) {
-	sinkURI, err := url.Parse(sinkURIStr)
-	if err != nil {
-		// try to parse the sinkURI as DSN
-		dsnCfg, err := dmysql.ParseDSN(sinkURIStr)
+	// check if sinkURI is a DSN
+	schemePrefix := sinkURIStr[:len(DSNScheme)]
```
amyangfei: This will panic if `len(sinkURIStr)` is less than len(DSNScheme). We can use `strings.HasPrefix(strings.ToLower(sinkURIStr), DSNScheme)` for simple

PR #347, opened 2020-03-17
support TLS in cdc components

amyangfei:

Support TLS in TiCDC, fulfill https://github.com/pingcap/ticdc/issues/250

- Code changed:

    - upgrade PD dependency to `github.com/pingcap/pd/v4`
    - Add security config to `cdc server` and `cdc cli`
    - gRPC communication with PD/TiKV will be encrypted
    - HTTP status/metric server will be encrypted if security config is enabled
    - Redirect gRPC log to CDC 

- Test changed:
    
    - Add an integration test case with TLS enabled.

TODO:
- Encrypt Sink communication with MySQL/Kafka
- Support common name check

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Support TLS


tests/_utils/run_curl
```diff
@@ -0,0 +1,35 @@
...
+TLS_DIR=$1
+shift 1
+
+if [ -z "${2-}" ]; then
+    POST_ARGS=
+else
+    POST_ARGS="--post-data $2"
+fi
+
+# FIXME: use `wget` instead of `curl` because the latter rejects ECC certs on our CI.
```
shafreeck: Is this FIXME still required?

cdc/kv/client.go
```diff
@@ -945,6 +944,7 @@ func (s *eventFeedSession) receiveFromStream(
 
 	for {
 		cevent, err := stream.Recv()
+		log.Debug("recv events", zap.Reflect("events", cevent))
```
overvenus, line 947:
```suggestion
```

cdc/kv/store_op.go
```diff
@@ -18,6 +18,8 @@ import (
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/ticdc/pkg/flags"
+	"github.com/pingcap/ticdc/pkg/security"
+	"github.com/pingcap/tidb/config"
```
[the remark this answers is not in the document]
leoppro: @overvenus

overvenus:
> how about making the config as a global variable or setting config into context?

Generally speaking, using global variables is not good especially when it relates to security issues. Setting credentials in context is doable but it makes code less explicitly and it is very likely to leak credential to packages we don't want.

PR #358, opened 2020-03-20
Combine the session of captures, manager and processors

shafreeck:
This work is based on #337,  merge it only after #337.

The captures and processors use separate sessions to keep their liveness and the owner uses a session to do election. This PR combines all these three types of sessions to one. When the session expires, the ownership, capture info and processor info are all removed by one time.

* Remove role manager, it is no more useful by design.
* Share one session between captures, processors and the election.
* Support to suicide itself for captures
* Support to recover from a suicide

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

GanttDue: 2020-04-03


amyangfei:
Plesae resolve conflicts @shafreeck

cdc/capture.go
```diff
@@ -78,99 +77,77 @@ func NewCapture(pdEndpoints []string) (c *Capture, err error) {
...
+	return c.election.Campaign(ctx, c.info.ID)
+}
+
+// Resign lets a owner start a new election.
+func (c *Capture) Resign(ctx context.Context) error {
+	return c.election.Resign(ctx)
+}
+
+// Suicide kills the capture itself
+func (c *Capture) Suicide() error {
```
amyangfei: this is not used now?

cdc/capture.go
```diff
@@ -23,34 +23,33 @@ import (
...
 	"google.golang.org/grpc/backoff"
 )
 
 const (
-	ownerRunInterval  = time.Millisecond * 500
 	captureSessionTTL = 3
 )
 
+// ErrSuicide causes a panic
+var ErrSuicide = errors.New("Suicide")
```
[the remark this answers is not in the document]
shafreeck: The ErrSuicide is only used for comparing, the stack does not matter.

cdc/capture.go
shafreeck: fixed

PR #363, opened 2020-03-20
Add a cmd for outputting statistics periodically

5kbpers:
Signed-off-by: 5kbpers <tangminghua@pingcap.com>

We need a way to check the replication speed and other things about performance.
This is a part of https://github.com/pingcap/ticdc/issues/340

Add a cmd for outputting statistics periodically.

### Check List 

Tests 

Code changes

Related changes

 - Need to update the documentation (perhaps)
 - Need to be included in the release note (perhaps)


cmd/ctrl.go
```diff
@@ -126,19 +143,91 @@ func newQueryChangefeedCommand() *cobra.Command {
...
+						return err
+					}
+					var count uint64
+					for _, pinfo := range taskPositions {
+						count += pinfo.Count
+					}
+					now := time.Now()
+					statistics := profileStatus{
+						OPS:   (count - lastCount) / uint64(now.Unix()-lastTime.Unix()),
+						Gap:   oracle.ExtractPhysical(status.ResolvedTs) - oracle.ExtractPhysical(status.CheckpointTs),
```
amyangfei: What about adding unit to `gap` field, like follows
```
{
        "ops": 0,
        "gap": 1005ms,
        "count": 2
}
```

cmd/ctrl.go
```diff
@@ -45,8 +49,21 @@ type capture struct {
...
+}
+
+type captureTaskStatus struct {
+	CaptureID  string            `json:"capture-id"`
+	TaskStatus *model.TaskStatus `json:"status"`
+}
+
+type profileStatus struct {
+	OPS   uint64 `json:"ops"`
+	Gap   int64  `json:"gap"`
```
amyangfei: What about adding another status, which tells user the replication gap between `current tso` and `checkpoint ts`. It may be a more native metric about replication status behind upstream

cmd/ctrl.go
```diff
@@ -126,19 +143,91 @@ func newQueryChangefeedCommand() *cobra.Command {
 		Use:   "query",
 		Short: "Query information and status of a replicaiton task (changefeed)",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			info, err := cdcEtcdCli.GetChangeFeedInfo(context.Background(), *changefeedID)
+			info, err := cdcEtcdCli.GetChangeFeedInfo(context.Background(), changefeedID)
+			if err != nil && errors.Cause(err) != model.ErrChangeFeedNotExists {
```
amyangfei:
```suggestion
			if err != nil {
```
we can just return error if `ChangeFeedStatus` does not exist, as the  `ChangeFeedStatus` is always put in etcd when a changefeed is created.

PR #364, opened 2020-03-20
tests: start tiflash server on CI

overvenus:

Run tests on TiDB cluster with TiFlash server.

Start TiFlash server for every CI cases.

### Check List 

Tests 

 - Integration test

Side effects

 - Increased code complexity


tests/_utils/start_tidb_cluster
```diff
@@ -18,8 +18,9 @@ cd $OUT_DIR && echo "start tidb cluster in $OUT_DIR"
 
 cat - > "$OUT_DIR/pd-config.toml" <<EOF
 [replication]
-# The number of replicas for each region.
-# max-replicas = 1
+# Set it to 1 to make sure we have enough replicas to run placement-rules.
+max-replicas = 1
+enable-placement-rules = true
```
amyangfei: what will happen if `max-replicas` is set to 3 here

scripts/jenkins_ci/integration_test_common.groovy
```diff
@@ -103,6 +106,12 @@ def tests(sink_type, node_label) {
                                 rm -rf cov_dir
                                 mkdir -p cov_dir
                                 ls /tmp/tidb_cdc_test
+                                echo "/tiflash/log/proxy.log"
+                                cat /tmp/tidb_cdc_test/${case_name}/tiflash/log/proxy.log
+                                echo "/tiflash/log/server.log"
+                                cat /tmp/tidb_cdc_test/${case_name}/tiflash/log/server.log
+                                echo "/tiflash/log/error.log"
+                                cat /tmp/tidb_cdc_test/${case_name}/tiflash/log/error.log
```
amyangfei: can we remove these debug logs now?

overvenus: Sure

tests/_utils/start_tidb_cluster
overvenus: Given the short time of running tests, it is possible to finish tests while replicas are still under 3. If so, the placement rule is not functioning properly. It would end with zero replicas on TiFlash nodes.

PR #368, opened 2020-03-23
Support to check whether the table can be duplicated by CDC

leoppro:
Support to check whether the table can be duplicated by CDC

we check table in the following step:
1. creating the changefeed
2. handling DDL jobs in the owner

### Check List 

Tests 

 - Unit test
 - Integration test

Run those sql in upstream db:
```
MySQL [test]> create table t1(a int ,b int);
Query OK, 0 rows affected (0.308 sec)

MySQL [test]> create table t2(a int unique key null ,b int);
Query OK, 0 rows affected (0.241 sec)

MySQL [test]> create table t3(a int unique key not null ,b int);
Query OK, 0 rows affected (0.750 sec)
```

create a changefeed:
```
[my-devel]➜ bin git:(check_table) ✗ ./cdc cli changefeed create --sink-uri="mysql://root@127.0.0.1:3306/"
[WARN] some tables are not eligible to duplicate, []entry.TableName{entry.TableName{Schema:"test", Table:"t1"}, entry.TableName{Schema:"test", Table:"t2"}}
Could you agree to ignore those tables, and continue to duplicate [Y/N]
N
[my-devel]➜ bin git:(check_table) ✗ ./cdc cli changefeed create --sink-uri="mysql://root@127.0.0.1:3306/" --no-confirm
[WARN] some tables are not eligible to duplicate, []entry.TableName{entry.TableName{Schema:"test", Table:"t1"}, entry.TableName{Schema:"test", Table:"t2"}}
create changefeed ID: 9a4603c1-5dda-4279-9001-ae01acd294b0 info {"sink-uri":"mysql://root@127.0.0.1:3306/","opts":{},"create-time":"2020-03-24T11:43:22.130405373+08:00","start-ts":415503850433675265,"target-ts":0,"admin-job-type":0,"config":{"filter-case-sensitive":false,"filter-rules":null,"ignore-txn-commit-ts":null}}
```


cdc/owner.go
```diff
@@ -317,12 +318,34 @@ func (c *changeFeed) banlanceOrphanTables(ctx context.Context, captures map[stri
 	}
 }
 
-func (c *changeFeed) applyJob(job *timodel.Job) error {
+func (c *changeFeed) checkJob(job *timodel.Job) (skip bool) {
+	switch job.Type {
+	case timodel.ActionCreateTable:
+		tableInfo := entry.WrapTableInfo(job.BinlogInfo.TableInfo)
+		log.Warn("this table is not eligible to duplicate", zap.Reflect("job", job))
```
amyangfei: This warning should not be logged when `tableInfo.ExistTableUniqueColumn()` is true

cmd/client.go
```diff
@@ -191,6 +196,25 @@ func newCreateChangefeedCommand() *cobra.Command {
 				Config:     cfg,
 			}
 
+			ineligibleTables, err := verifyTables(ctx, cfg)
+			if err != nil {
+				return err
+			}
+			if len(ineligibleTables) != 0 {
+				cmd.Printf("[WARN] some tables are not eligible to duplicate, %#v\n", ineligibleTables)
```
amyangfei: s/duplicate/replicate
ditto for all `duplicate`

leoppro:
@amyangfei PTAL again

PR #376, opened 2020-03-26
add white list ddl filter

leoppro:
add white list ddl filter, the user can specify the item in the white list by congig

### Check List 

Tests 

 - Unit test
 - Integration test


cmd/changefeed.toml
```diff
@@ -1,6 +1,7 @@
-ignore-txn-commit-ts = []
 
 filter-case-sensitive = false
+ignore-txn-commit-ts = []
+ddl-white-list = [100, 101]
```
amyangfei, line 4: seems `100` and `101` are invalid DDL ActionType, recommend to use valid value. We should also update the document and put a URL here to tell user what value they can use

leoppro: updated, and I think we should hide `ddl-white-list` to user

PR #377, opened 2020-03-26
task: fix to parse task when the status is not created

shafreeck:

When a task is dispatched before a change feed status is created, a newfound task can not be parsed.

When a change feed status is not created, use a default value.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/task.go
+++ b/cdc/task.go
@@ -15,8 +15,8 @@ package cdc
 
 import (
 	"context"
-	"errors"
 
+	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	"github.com/pingcap/ticdc/cdc/model"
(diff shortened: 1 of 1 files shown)


cdc/task.go
```diff
@@ -169,7 +169,11 @@ func (w *TaskWatcher) parseTask(ctx context.Context,
 	}
 	status, err := w.capture.etcdClient.GetChangeFeedStatus(ctx, changeFeedID)
 	if err != nil {
-		return nil, err
+		if errors.Cause(err) == model.ErrTaskStatusNotExists {
```
amyangfei:
```suggestion
		if errors.Cause(err) == model.ErrChangeFeedNotExists {
```

amyangfei:
PTAL @leoppro

PR #378, opened 2020-03-26
util: expose initial interval of retry function

leoppro:
expose initial interval of retry function

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/entry/schema_builder.go
+++ b/cdc/entry/schema_builder.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/cenkalti/backoff"
 	"github.com/pingcap/errors"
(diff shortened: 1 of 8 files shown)


cdc/entry/schema_builder.go
```diff
@@ -169,13 +170,14 @@ func (b *StorageBuilder) Build(ts uint64) (*Storage, error) {
...
 	b.baseStorageMu.Unlock()
 
-	err := retry.Run(func() error {
-		err := c.HandlePreviousDDLJobIfNeed(ts)
-		if errors.Cause(err) != model.ErrUnresolved {
-			return backoff.Permanent(err)
-		}
-		return err
-	}, 5)
+	err := retry.Run(10*time.Millisecond, 5,
```
amyangfei: this retry will be put in `HandlePreviousDDLJobIfNeed` in https://github.com/pingcap/ticdc/pull/374. Just want to know why this retry limit is changed to 10ms

PR #395, opened 2020-03-29
increase chan size, fix grafana template

amyangfei:

As we have an unlimited buffer between puller and TiKV, we could increase channel buffer size as large as possible.

### Check List 

Tests 

 - Unit test
 - Integration test


leoppro:
LGTM

PR #397, opened 2020-03-29
add metrics in mounter and blackhole sink

amyangfei:

add metrics in mounter and blackhole sink

### Check List 

Tests 

 - Unit test
 - Integration test


leoppro:
LGTM

PR #416, opened 2020-03-31
util: Add RegionRangeLock

MyonKeminta:

This PR adds a util named RegionRangeLock, which is a simple range and can also maintains minimal checkpointTs by range and identify stale acquire-lock operation by checking the version of the region's epoch.

This PR is part of https://github.com/pingcap/ticdc/pull/385 , which supports region merge.

Added new util RegionRangeLock.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes

 - No.

### DIFF
--- a/pkg/util/region_range_lock.go
+++ b/pkg/util/region_range_lock.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/google/btree"
 	"github.com/pingcap/log"
-	"github.com/pingcap/tidb/store/tikv/oracle"
 	"go.uber.org/zap"
 )
 
@@ -103,16 +102,6 @@ func (m *RangeTsMap) GetMin(startKey, endKey []byte) uint64 {
 		return true
 	})
 	return ts
-}
-
-func (m *RangeTsMap) String() string {
-	result := ""
-	m.m.Ascend(func(i btree.Item) bool {
-		e := i.(*rangeTsEntry)
-		result += fmt.Sprintf("%v %v %v\n", e.ts, oracle.GetTimeFromTS(e.ts), hex.EncodeToString(e.startKey))
-		return true
-	})
-	return result
 }
 
 type rangeLockEntry struct {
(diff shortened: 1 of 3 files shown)


pkg/util/region_range_lock.go
```diff
@@ -0,0 +1,337 @@
...
+	"sync/atomic"
+
+	"github.com/google/btree"
+	"github.com/pingcap/log"
+	"github.com/pingcap/tidb/store/tikv/oracle"
+	"go.uber.org/zap"
+)
+
+type rangeTsEntry struct {
+	// Only startKey is necessary. End key can be inferred by the next item since the map always keeps a continuous
```
overvenus: What if there is a gap between two adjacent items?

pkg/util/region_range_lock.go
```diff
@@ -0,0 +1,337 @@
...
+		return true
+	})
+	return ts
+}
+
+func (m *RangeTsMap) String() string {
+	result := ""
+	m.m.Ascend(func(i btree.Item) bool {
+		e := i.(*rangeTsEntry)
+		result += fmt.Sprintf("%v %v %v\n", e.ts, oracle.GetTimeFromTS(e.ts), hex.EncodeToString(e.startKey))
```
overvenus: It may print a very large string, is it expected?

pkg/util/region_range_lock.go
```diff
@@ -0,0 +1,337 @@
...
+			entry.startKey, entry.endKey, entry.version, startKey, endKey, version))
+	}
+
+	for _, ch := range entry.waiters {
+		ch <- nil
+	}
+
+	i := l.rangeLock.Delete(entry)
+	if i == nil {
+		panic("impossible")
```
overvenus: Should print some useful information.

pkg/util/region_range_lock.go
MyonKeminta: When updating a range, if there is not another entry with it's key = `endKey`, it will be inserted to keep the value of the original value of the range after the new range. For example, 
* "a" -> 1 (represents ["a", "d") is 1)
* "d" -> 2
if we set ["b", "c") to 10, it will become:
* "a" -> 1 (represents ["a", "b") is 1)
* "b" -> 10 (represents ["b", "c") is 10)
* "c" -> 1 (represents ["c", "d") is 1)
* "d" -> 2 
And if there's no value set in some range, the default value will be considered MaxUint64 (because we need to calculate the minimal value among ranges)

pkg/util/region_range_lock.go
MyonKeminta: Oh, it is very long, and I use this to debug the region merge's logic. Maybe I should remove it?

overvenus: If it is only used for debug, then LGTM.

PR #432, opened 2020-04-02
schema: support multi-version storage for schema storage

leoppro:
support multi-version storage for schema storage
for run test

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/entry/schema_storage.go
```diff
@@ -678,79 +539,118 @@ func (s *Storage) CloneTables() map[uint64]TableName {
...
 	}
-	for k, v := range s.truncateTableID {
-		n.truncateTableID[k] = v
+}
+
+// GetLastSnapshot returns the last snapshot
+func (s *SchemaStorage) GetLastSnapshot() *schemaSnapshot {
+	s.snapsMu.RLock()
+	defer s.snapsMu.RUnlock()
+	return s.snaps[len(s.snaps)-1]
```
amyangfei, line 610: We must guarantee the `s.snaps` is not empty 🤔️

cdc/entry/schema_storage.go
```diff
@@ -678,79 +539,118 @@ func (s *Storage) CloneTables() map[uint64]TableName {
...
+}
+
+// HandleDDLJob creates a new snapshot in storage and handles the ddl job
+func (s *SchemaStorage) HandleDDLJob(job *timodel.Job) error {
+	s.snapsMu.Lock()
+	defer s.snapsMu.Unlock()
+	lastSnap := s.snaps[len(s.snaps)-1]
```
amyangfei, line 617: move this line before L#615 and reuse `GetLastSnapshot`?

cdc/entry/schema_storage.go
```diff
@@ -678,79 +539,118 @@ func (s *Storage) CloneTables() map[uint64]TableName {
...
-		dbInfo, exist := s.SchemaByID(job.SchemaID)
-		if !exist {
-			return errors.NotFoundf("schema %d not found", job.SchemaID)
+// DoGC removes snaps which of ts less than this specified ts
+func (s *SchemaStorage) DoGC(ts uint64) {
+	s.snapsMu.Lock()
+	defer s.snapsMu.Unlock()
+	var startIdx int
+	for i, snap := range s.snaps {
+		if snap.currentTs > ts {
```
amyangfei, line 647: Do we need to keep snapshot with `snap.currentTs = ts`

cdc/entry/schema_storage.go
leoppro: This will cause deadlock

cdc/entry/schema_storage.go
leoppro: yes, when we create a SchemaStorage, we will create a snap into `snaps`.
and the `DoGC` function will guarantee that `len(s.sanps)` at least one more

cdc/entry/schema_storage.go
leoppro: I don't think we need to consider this. One more snapshot does not affect anything, but one less snapshot does not work. So we leave one more

cdc/entry/schema_storage.go
```diff
@@ -678,79 +539,118 @@ func (s *Storage) CloneTables() map[uint64]TableName {
...
+	return nil
 }
 
-// IsIneligibleTableID returns true if the table is ineligible
-func (s *Storage) IsIneligibleTableID(id int64) bool {
-	_, ok := s.ineligibleTableID[id]
-	return ok
+// AdvanceResolvedTs advances the resolved
+func (s *SchemaStorage) AdvanceResolvedTs(ts uint64) {
+	atomic.StoreUint64(&s.resolvedTs, ts)
```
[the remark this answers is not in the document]
leoppro: ok

leoppro:
PTAL @amyangfei

PR #433, opened 2020-04-03
kv-client: support p-c out of order before initialized

amyangfei:

commit log may come before prewrite log, if the incremental scan has not finished.

use a cache to store the unmatched commit log before initialized

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/kv/matcher.go
+++ b/cdc/kv/matcher.go
@@ -1,7 +1,6 @@
 package kv
 
 import (
-	"github.com/pingcap/errors"
 	"github.com/pingcap/kvproto/pkg/cdcpb"
 )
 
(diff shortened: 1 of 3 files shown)


cdc/kv/matcher.go
```diff
@@ -35,7 +36,14 @@ func (m *matcher) matchRow(row *cdcpb.Event_Row) ([]byte, error) {
 		return value, nil
 	}
 	return nil, errors.NotFoundf("prewrite row, startTs:%d", row.GetStartTs())
```
leoppro, line 38: maybe we can return a bool indicates whether the match was successful

amyangfei: addressed

amyangfei:
PTAL @leoppro

PR #434, opened 2020-04-03
unittest: add failpoint for watchers

### LINKED ISSUE #409: Ensure the watchers recover from the ErrCompact error
# Cases

- [x] The task watcher
- [x] The capture watcher

# Expectation

* No events are lost

shafreeck:

Closes #409 

Add failpoint for the task watcher and the capture watcher

Add failpoint to trigger an ErrCompacted error

**Notice that the unit test for the capture watcher is not added, because the whole test file has been commented out.**

### Check List 

Tests 

 - Unit test

### DIFF
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/pingcap/errors"
+	"github.com/pingcap/failpoint"
 	"github.com/pingcap/log"
 	pd "github.com/pingcap/pd/v4/client"
 	"github.com/pingcap/ticdc/cdc/entry"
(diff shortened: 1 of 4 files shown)


cdc/owner.go
```diff
@@ -615,7 +617,11 @@ func (o *Owner) watchCapture(ctx context.Context) error {
 		clientv3.WithPrevKV())
 
 	for resp := range ch {
-		if resp.Err() != nil {
+		err := resp.Err()
+		failpoint.Inject("restart-capture-watch", func() {
```
amyangfei, line 621: seems we have no unit test for owner now

shafreeck: Yes, the whole file is commented out temporarily. A unit test should be added after the file is recovered.

PR #438, opened 2020-04-07
schema: fix `table not found` when `startTS` = 0

leoppro:
when `startts=0`,  `LoadHistoryDDLJobs` function will return a empty jobs array, but we can't receive integrated DDL jobs from DDL Puller

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/processor.go
```diff
@@ -653,6 +653,7 @@ func (p *processor) addTable(ctx context.Context, tableID int64, startTs uint64)
 				if pEvent == nil {
 					continue
 				}
+				log.Info("sorter output", zap.Reflect("pevent", pEvent))
```
amyangfei, line 656: Is this log necessary?

leoppro: it's a debug code, I forgot to remove it.

PR #439, opened 2020-04-07
kv/client: Add backoff on regoin error

MyonKeminta:

This PR reduces the frequency of retrying when region meets error.

Adds a simple sleep before continue handling the region error when the error occurs in `partialRegionFeed`. This significantly reduced NotLeader and RPCCtxUnavailable.
This mechanism used to exist before implementing batch streaming, but was removed when we were hurrying to finish the batch streaming. Add it back now.

By the way, since bugs shows up more frequently without this mechanism, I think it would be a good idea that we add a hidden config for it, so that we can skip the waiting and retry immediately when we are testing correctness of CDC, and in this way it's easier to trigger bugs.

### Check List 

Tests 

image

(Left: before; Right: after).

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -735,6 +735,8 @@ func (s *eventFeedSession) partialRegionFeed(
 
 	// We need to ensure when the error is handled, `isStopped` must be set. So set it before sending the error.
 	state.markStopped()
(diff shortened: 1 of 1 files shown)


MyonKeminta:
Oh, DNM. It seems it waits for too long.

PR #446, opened 2020-04-09
kv client: ignore prewrite event if value is empty

leoppro:
ignore prewrite event if value is empty

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/kv/matcher.go
+++ b/cdc/kv/matcher.go
@@ -26,7 +26,12 @@ func newMatcher() *matcher {
 }
 
 func (m *matcher) putPrewriteRow(row *cdcpb.Event_Row) {
-	m.unmatchedValue[newMatchKey(row)] = row.GetValue()
+	key := newMatchKey(row)
+	value := row.GetValue()
+	if _, exist := m.unmatchedValue[key]; exist && len(value) == 0 {
+		return
+	}
+	m.unmatchedValue[key] = value
 }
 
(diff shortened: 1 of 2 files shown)


cdc/kv/matcher.go
```diff
@@ -26,7 +26,12 @@ func newMatcher() *matcher {
 }
 
 func (m *matcher) putPrewriteRow(row *cdcpb.Event_Row) {
-	m.unmatchedValue[newMatchKey(row)] = row.GetValue()
+	key := newMatchKey(row)
+	value := row.GetValue()
+	if _, exist := m.unmatchedValue[key]; exist && len(value) == 0 {
```
amyangfei, line 31: Please add some comments to explain why we don't overwrite the old value directly.

cdc/kv/matcher.go
```diff
@@ -26,7 +26,14 @@ func newMatcher() *matcher {
 }
 
 func (m *matcher) putPrewriteRow(row *cdcpb.Event_Row) {
-	m.unmatchedValue[newMatchKey(row)] = row.GetValue()
+	key := newMatchKey(row)
+	value := row.GetValue()
+	// tikv may send a prewrite event with empty value
+	// here we need to avoid the invalid prewrite event overwrite the value
+	if _, exist := m.unmatchedValue[key]; exist && len(value) == 0 {
```
overvenus, line 33: PTAL @5kbpers

PR #447, opened 2020-04-09
cmd: readd processor list command

amyangfei:

`cdc cli processor list` is removed in #371 by accident

Add it back

### Check List 

Tests 

 - Unit test
 - Integration test


amyangfei:
PTAL @leoppro

PR #450, opened 2020-04-10
owner: fix a bug about wrong timing sequence of DDL when add a table to processor

leoppro:
fix a bug about wrong timing sequence of DDL when add a table to processor

when the owner add a table to a processor, the owner should ensure the global resolvedts won't exceed the startts of the new table

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/changefeed.go
+++ b/cdc/changefeed.go
@@ -83,9 +83,12 @@ type changeFeed struct {
 	schemas              map[uint64]tableIDMap
 	tables               map[uint64]entry.TableName
 	orphanTables         map[uint64]model.ProcessTableInfo
-	waitingConfirmTables map[uint64]string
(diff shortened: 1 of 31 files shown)


cdc/changefeed.go
```diff
@@ -521,15 +533,16 @@ func (c *changeFeed) calcResolvedTs(ctx context.Context) error {
 		minCheckpointTs = minResolvedTs
 	}
 
+	c.status.ResolvedTs = minResolvedTs
+	c.status.CheckpointTs = minCheckpointTs
```
amyangfei, line 537: these two lines have conflict with L541 and L545. `c.sink.EmitCheckpointEvent` will be never called?

PR #459, opened 2020-04-13
kvclient: support resolve lock for ticdc

leoppro:

support resolve lock for ticdc
### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -14,9 +14,9 @@
 package kv
 
 import (
+	"bytes"
 	"context"
 	"fmt"
-	"github.com/prometheus/client_golang/prometheus"
 	"io"
 	"math/rand"
 	"strconv"
@@ -26,6 +26,7 @@ import (
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/kvproto/pkg/cdcpb"
+	"github.com/pingcap/kvproto/pkg/kvrpcpb"
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/log"
 	pd "github.com/pingcap/pd/v4/client"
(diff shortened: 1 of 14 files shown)


cdc/kv/client.go
```diff
@@ -1013,27 +1022,40 @@ func (s *eventFeedSession) singleEventFeed(
...
+			currentTimeFromPD := oracle.GetTimeFromTS(version.Ver)
+			sinceLastResolvedTs := currentTimeFromPD.Sub(oracle.GetTimeFromTS(lastResolvedTs))
+			if sinceLastResolvedTs > time.Second*20 {
+				log.Warn("region not receiving resolved event from tikv for too long time, try to resolve lock",
+					zap.Uint64("regionID", regionID), zap.Reflect("span", span), zap.Duration("duration", sinceLastResolvedTs), zap.Uint64("lastResolvedTs", lastResolvedTs))
+				maxVersion := oracle.ComposeTS(oracle.GetPhysical(currentTimeFromPD.Add(-10*time.Second)), 0)
+				err = s.resolveLock(ctx, regionID, maxVersion)
+				if err != nil {
+					return atomic.LoadUint64(&checkpointTs), errors.Trace(err)
+				}
```
MyonKeminta, line 1054: I think it's better not to return error here (so do line 1043).  Instead, just continue the loop and retry resolveLock next time. In the current implementation if an error is returned from here, CDC will retry the region but TiKV may not know the region's eventFeed is canceled.

cdc/kv/client.go
```diff
@@ -1163,6 +1187,89 @@ func (s *eventFeedSession) singleEventFeed(
...
+		if err1 != nil {
+			return errors.Trace(err1)
+		}
+		if len(locks) < scanLockLimit {
+			key = loc.EndKey
+		} else {
+			key = locks[len(locks)-1].Key
+		}
+
+		if len(key) == 0 || bytes.Compare(key, loc.EndKey) >= 0 {
```
MyonKeminta, line 1264:
```suggestion
		if len(key) == 0 || (len(loc.EndKey) != 0 && bytes.Compare(key, loc.EndKey) >= 0) {
```

`loc.EndKey` can be "", which means it's the last region. In this case if `key` is not "", it should continue scanning. You should check `loc.EndKey` like how TiDB did:

```go
if len(key) == 0 || (len(endKey) != 0 && bytes.Compare(key, endKey) >= 0) {
	break
}
```

MyonKeminta:
By the way I think it would be good if there can be a test that checks the resolve lock won't kill an ongoing transaction..

PR #465, opened 2020-04-16
sink : add a function that user can switch partition rule in mq sink

leoppro:
a function that user can switch partition rule in mq sink

this PR provided four kinds of partition rules:
`default`, `rowid`, `ts`, `table`

this pr changed the config of changefeed:

```
filter-case-sensitive = false
ignore-txn-commit-ts = []

[filter-rules]
ignore-dbs = ["test", "sys"]

[[filter-rules.do-tables]]
db-name = "sns"
tbl-name = "user"

[[filter-rules.do-tables]]
db-name = "sns"
tbl-name = "following"

[[sink-partition-rules]]
db-name = "sns"
tbl-name = "user"
rule = "ts"

[[sink-partition-rules]]
db-name = "sns"
tbl-name = "following"
rule = "rowid"
```

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/sink/mq.go
```diff
@@ -261,5 +234,125 @@ func newKafkaSaramaSink(ctx context.Context, sinkURI *url.URL, filter *util.Filt
...
+	hash := crc32.NewIEEE()
+	// distribute partition by table
+	_, err := hash.Write([]byte(row.Schema))
+	if err != nil {
+		log.Fatal("calculate hash of message key failed, please report a bug", zap.Error(err))
+	}
+	_, err = hash.Write([]byte(row.Table))
+	if err != nil {
+		log.Fatal("calculate hash of message key failed, please report a bug", zap.Error(err))
+	}
```
amyangfei, line 319: Should the default partition strategy work as follows:
- If `IndieMarkCol` exists, use rowid or unique column value to calculate partition
- Otherwise, use schema.table to calculate partition.

`schema` and `table` are always used to calculate now

leoppro: It's fine

PR #477, opened 2020-04-20
Support file sort in puller

amyangfei:

This is the first part of https://github.com/pingcap/ticdc/issues/466, add a file sort mechanism, a changefeed can use either memory sorter or file sorter

- add two kinds of sort engine, memory and file
- define a new interface `EventSorter` in puller, so we don't need to change original replication model.
- file sorter has an `input` cache channel, it reads from `input` channel and flushes events to file until they are all mounted. 
- implement a simple double cache, one is used to sort with files, the other one is used to cache unsorted data to file.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/model/mounter.go
+++ b/cdc/model/mounter.go
@@ -19,10 +19,10 @@ import (
 
 // PolymorphicEvent describes a event can be in multiple states
 type PolymorphicEvent struct {
-	Ts       uint64           `json:"t"`
-	RawKV    *RawKVEntry      `json:"-"`
-	Row      *RowChangedEvent `json:"r"`
-	Finished chan struct{}    `json:"-"`
+	Ts       uint64
+	RawKV    *RawKVEntry
+	Row      *RowChangedEvent
+	finished chan struct{}
 }
 
 // NewPolymorphicEvent creates a new PolymorphicEvent with a raw KV
@@ -33,7 +33,7 @@ func NewPolymorphicEvent(rawKV *RawKVEntry) *PolymorphicEvent {
 	return &PolymorphicEvent{
 		Ts:       rawKV.Ts,
 		RawKV:    rawKV,
-		Finished: make(chan struct{}),
+		finished: make(chan struct{}),
 	}
 }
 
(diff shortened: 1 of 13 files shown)


cdc/model/mounter.go
```diff
@@ -13,14 +13,16 @@
...
 // PolymorphicEvent describes a event can be in multiple states
 type PolymorphicEvent struct {
-	Ts       uint64
-	RawKV    *RawKVEntry
-	Row      *RowChangedEvent
-	finished chan struct{}
+	Ts       uint64           `json:"t"`
+	RawKV    *RawKVEntry      `json:"-"`
+	Row      *RowChangedEvent `json:"r"`
+	Finished chan struct{}    `json:"-"`
```
leoppro, line 25: why make `Finished` public? and why not store `RowChangedEvent` to the disk directly

leoppro: by the way, i think `json` is not a suited serialization algorithm, it's advantages is readability but we don't need it here

amyangfei: Good catch, `Finished` is set to public by mistake. And use gob encode to store PolymorphicEvent to file

cdc/puller/file_sorter.go
```diff
@@ -0,0 +1,494 @@
...
+	wg.Go(func() error {
+		return fs.gcRemovedFiles(ctx)
+	})
+
+	return wg.Wait()
+}
+
+func (fs *FileSorter) sortAndOutput(ctx context.Context) error {
+	bufferLen := 10
+	buffer := make([]*model.PolymorphicEvent, 0, bufferLen)
```
leoppro, line 431: using bytes.Buffer to avoid frequent memory allocation

cdc/puller/file_sorter.go
```diff
@@ -0,0 +1,494 @@
...
+	buf := new(bytes.Buffer)
+	dataBuf := new(bytes.Buffer)
+	var dataLen [8]byte
+	for _, entry := range entries {
+		err := entry.WaitPrepare(ctx)
+		if err != nil {
+			return 0, errors.Trace(err)
+		}
+		dataBuf.Reset()
+		err = gob.NewEncoder(dataBuf).Encode(entry)
```
leoppro, line 127: how about create a new encoder with file buffer directly?
```
	f, err := os.OpenFile(fullpath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
	if err != nil {
		return 0, errors.Trace(err)
	}
	w := bufio.NewWriter(f)
        encoder:= gob.NewEncoder(w)

	for _, entry := range entries {
		encoder.Encode(entry)
	}
```

cdc/puller/file_sorter.go
```diff
@@ -0,0 +1,494 @@
...
+			if err != nil {
+				return "", errors.Trace(err)
+			}
+		}
+		return newfile, nil
+	}
+
+	// clear and reset unsorted files, set cache sorting flag to prevent repeated sort
+	fs.cache.fileLock.Lock()
+	if atomic.LoadInt32(&fs.cache.sorting) == 1 {
```
leoppro, line 297: it's needless that check the sorting is in process under lock protect
we can use `CompareAndSwap`  to avoid using lock

cdc/puller/file_sorter.go
```diff
@@ -0,0 +1,494 @@
...
+		}
+		evs := make([]*model.PolymorphicEvent, 0)
+		idx := 0
+		for idx < len(data) {
+			dataLen := int(binary.BigEndian.Uint64(data[idx : idx+8]))
+			if idx+8+dataLen > len(data) {
+				return "", errors.New("unsorted file unexpected truncated")
+			}
+			ev := &model.PolymorphicEvent{}
+			err = gob.NewDecoder(bytes.NewReader(data[idx+8 : idx+8+dataLen])).Decode(ev)
```
leoppro, line 261: can we avoid to create decode and reader continually?

PR #481, opened 2020-04-23
cdc: update position and status as soon as possible

### LINKED ISSUE #468: Optimize replication model to decrease replication latency
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently we have several `tick` mechanism in TiCDC replication model, mainly includes:

- owner has a period job, including loads each processor's task status, calculate global resolved ts, process DDL, handling admin job and update changefeed status in a tick, with 500ms
- processor also has several period jobs, including update local resolved ts, update local checkpoint ts and update task status. with 500ms tick

So the replication model of TiCDC is serial, as follows

processor advances local resolved ts -> owner advances global resolved ts -> processor replicate data, and update local checkpoint ts

Currently the replication latency in TiCDC is approximately 2-3s. (under the condition that TiKV push resolved ts every 1s)

**Describe the feature you'd like:**

Maybe we can design a better `watch` mechanism to decrease the waiting time between processor and owner. The ideal upper limit of replication latency should be TiKV resolved ts interval.

**Work estimate**

2 weeks. including test

bobotu:
Resolve #468.

For `processor` do a tiny refactor to wakeup it when the resolved ts or status have changed.

For `owner` it is hard to refactor the main loop, so I just add a side-channel to process an updated position as soon as possible.

We can improve the efficiency of the current event model in some harder ways, such as reduce the etcd queries, do less work in the event loop, redesign the whole interaction model, etc. But these ways will change a lot of code and introduce new bugs. 

For now, it is safer to do some little changes like this PR to gain some performance improvements in some scenarios.

### Check List 
Simple benchmark https://github.com/pingcap/ticdc/pull/481#issuecomment-619488481

Release note

None

Tests 

 - Unit test
 - Integration test


bobotu:
Has been tested using a simple scenario

```sql
CREATE TABLE `heartbeat` (`ts` datetime(6) NOT NULL)
```

Then update ts every second
```sql
update heartbeat set ts = now(6)
```

Watching time difference on downstream MySQL every second
```sql
select TIMESTAMPDIFF(microsecond, ts, now(6)) / 1000 AS diff from heartbeat;
```

In this scenario, the latency can be reduced by about 500-1000ms.

cdc/owner.go
```diff
@@ -529,6 +539,33 @@ loop:
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+			wch := o.etcdClient.Client.Watch(ctx, kv.TaskPositionKeyPrefix, clientv3.WithFilterDelete(), clientv3.WithPrefix())
+
+			for resp := range wch {
+				if resp.Err() != nil {
```
amyangfei, line 554: Do we need to process etcd `ErrCompact` here?

bobotu: Not necessary. Any kind of error will simply recreate the watch channel, and it will terminated when `ctx` is canceled.

amyangfei:
a failed test case: https://github.com/pingcap/ticdc/issues/513

cdc/changefeed.go
```diff
@@ -537,17 +548,19 @@ func (c *changeFeed) calcResolvedTs(ctx context.Context) error {
 
 	if minCheckpointTs > c.status.CheckpointTs {
 		c.status.CheckpointTs = minCheckpointTs
-		err := c.sink.EmitCheckpointTs(ctx, minCheckpointTs)
-		if err != nil {
-			return errors.Trace(err)
+		if c.ddlState != model.ChangeFeedWaitToExecDDL {
```
amyangfei, line 551: why checking `c.ddlState != model.ChangeFeedWaitToExecDDL` here

leoppro: when `c.ddlState == model.ChangeFeedWaitToExecDDL`, the ddl may not executed yet, so we should wait and ensure the ddl is already executed.(here we just skip this checkpointTs)

pkg/notify/notify_test.go
```diff
@@ -12,28 +11,31 @@ type notifySuite struct{}
 var _ = check.Suite(&notifySuite{})
```
amyangfei, line 11: find the test suite not really runs by registering `check.TestingT(t)`

leoppro: fixed

cdc/changefeed.go
amyangfei: Oh, it means in this case `minCheckpointTs` can be updated to `c.ddlJobHistory[0].BinlogInfo.FinishedTS`, but DDL is waiting to be executed, so we can't forward `c.status.CheckpointTs`

leoppro: yes

amyangfei: better to add some comments, or others will get confused

leoppro: done

PR #496, opened 2020-04-26
owner: fix global checkpointts stopping when it 's not exist changefees

leoppro:
If the number of changefeed is zero, checkpoint will not be pushing

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/changefeed.go
```diff
@@ -481,7 +481,7 @@ func (c *changeFeed) calcResolvedTs(ctx context.Context) error {
 	minCheckpointTs := c.targetTs
 
 	if len(c.taskPositions) == 0 {
-		minCheckpointTs = c.status.CheckpointTs
+		minCheckpointTs = c.status.ResolvedTs
 	} else if len(c.taskPositions) < len(c.taskStatus) {
 		return nil
```
amyangfei, line 486: Is this better:
- check if `len(c.taskPositions) < len(c.taskStatus)`, do nothing
- then check if `len(c.taskPositions) == 0`, update `minCheckpointTs` to `c.status.ResolvedTs`
If owner has dispatched a table but processor doesn't start replicating it, we should wait the processor is ready?

leoppro: > If owner has dispatched a table but processor doesn't start replicating it, owner should wait until the processor is ready?

Yes, the owner should, and we using table lock to make the owner wait until the processor is ready now

leoppro:
until the processor is ready

PR #502, opened 2020-04-27
sink/mysql: use pk/uk to refine concurrent execution

crazycs520:

Close https://github.com/pingcap/ticdc/issues/463.

bench test:

sink 1 table, 10000 rows to mysql

```sql
before this PR: 59s. -- 16 workers, but only 1 do jobs. 
```

This PR: 
```shell
8.5s.  -- 16 workers.
10.5s.  -- 8 workers.
15.65s -- 4 workers.
```

Most same with the [DM](https://github.com/pingcap/dm), use [causality](https://github.com/pingcap/dm/blob/master/syncer/causality.go) to do conflict check, change a little.

### Check List 

Tests 

 - Unit test
 - manual test.
   - rand update with conflict rows and sink to MySQL. Running about 1 hour. then compare the data consistency.
   - [update script](https://gist.github.com/crazycs520/fdf75b8ee6dc3c393575c7fc7078fbbe)

```go
        count := 1000
	n1 := rand.Intn(count)
	n2 := rand.Intn(count)
	n3 := rand.Intn(count)
         // column a is primary key, column b is unique key.
	sql := fmt.Sprintf("update t5 set a=%d, b= 'efg-%d', c=now() where a=%d", n1, n2, n3)
```

Code changes

Side effects

 - Increased code complexity

Related changes


leoppro:
LGTM

cdc/sink/mysql.go
```diff
@@ -357,60 +356,121 @@ func newMySQLSink(sinkURI *url.URL, dsn *dmysql.Config, filter *util.Filter, opt
...
-		if err := fn(rows[start:end]); err != nil {
-			return errors.Trace(err)
+	}()
+	var rows []*model.RowChangedEvent
+	for row := range w.rowCh {
+		rows = append(rows, row)
+		rows = w.fetchAllPendingEvent(rows)
+		// TODO: Add retry
+		err := w.execDMLs(ctx, rows)
+		w.trySendErr(err)
```
amyangfei, line 442: should we check err is nil here? or maybe the channel will be filled with a nil error but real error is not sent to

cdc/sink/mysql.go
```diff
@@ -357,60 +356,121 @@ func newMySQLSink(sinkURI *url.URL, dsn *dmysql.Config, filter *util.Filter, opt
...
+				zap.Reflect("r", r), zap.Stack("stack trace"))
 		}
-		if err := fn(rows[start:end]); err != nil {
-			return errors.Trace(err)
+	}()
+	var rows []*model.RowChangedEvent
+	for row := range w.rowCh {
+		rows = append(rows, row)
+		rows = w.fetchAllPendingEvent(rows)
+		// TODO: Add retry
```
amyangfei, line 440: We have retry now, this line can be removed now

cdc/sink/mysql.go
```diff
@@ -357,60 +356,121 @@ func newMySQLSink(sinkURI *url.URL, dsn *dmysql.Config, filter *util.Filter, opt
...
+	rowCh     chan *model.RowChangedEvent
+	errCh     chan error
+	jobWg     *sync.WaitGroup
+	maxTxnRow int
+	execDMLs  func(context.Context, []*model.RowChangedEvent) error
+}
+
+func (w *mysqlSinkWorker) run(ctx context.Context) {
+	defer func() {
+		if r := recover(); r != nil {
```
amyangfei, line 431: If the worker panic in `run` and we recover from it, `jobWg.Wait()` may hang and never return

cdc/sink/causality.go
```diff
@@ -0,0 +1,49 @@
+package sink
```
amyangfei, line 1: better to add copyright

cdc/sink/mysql.go
crazycs520: Great catch.

cdc/sink/causality.go
crazycs520: done.

cdc/sink/mysql.go
crazycs520: Great catch.

PR #503, opened 2020-04-27
owner: batch add table when start a changefeed

leoppro:

If there are many tables in the upstream,  the process of adding tables is very slow.

Add tables in batches

### Check List 

Tests 

 - Unit test
 - Integration test


leoppro:
> Is there any test case covers this scenario
> 
> > both upstream and downstream have N tables, then start CDC to replicate
> 
> If not, do we need to add one?

fine

PR #524, opened 2020-04-30
sink: the implements of the new sink interface

leoppro:
the implements of the new sink interface

### Release note

None

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/sink/sink.go
```diff
@@ -32,30 +32,32 @@ const (
...
-	CheckpointTs() uint64
-	// Count returns the count of sunk events
-	Count() uint64
-	// Run runs the sink
-	Run(ctx context.Context) error
-	// PrintStatus prints necessary status periodically
-	PrintStatus(ctx context.Context) error
+
+	// FlushRowChangedEvents flushes each row which of commitTs greater or equal to `resolvedTs` into downstream.
+	// TiCDC guarantees that all of Event which of commitTs greater or equal to `resolvedTs` are sent to Sink through `EmitRowChangedEvents`
```
amyangfei, line 45: Is it row with commitTs less than or equal to `resolvedTs` ?
and guarantees that all of the events with commits less than or equal to `resolvedTs`

cdc/sink/sink.go
```diff
@@ -32,30 +32,32 @@ const (
...
+	FlushRowChangedEvents(ctx context.Context, resolvedTs uint64) error
+
+	// EmitCheckpointTs sends CheckpointTs to Sink
+	// TiCDC guarantees that all Events **in the cluster** which of commitTs greater or equal to `checkpointTs` are sent to downstream successfully.
```
amyangfei, line 49: less than or euqal to `checkpointTs`?

pkg/util/notify.go
```diff
@@ -0,0 +1,118 @@
+package util
```
overvenus, line 1: Could you put it in a new package?

PR #528, opened 2020-05-01
owner: fix resouce cleanup when an owner exits

amyangfei:

Fix https://github.com/pingcap/ticdc/issues/523

- remove the panic mechanism in capture, just return the `Suicide` error when etcd session is done.
- make `campaignOwnerLoop` as a background routine, in most failure cases (including err compacted, `s.capture.Campaign` failed, owner Run failed), we don't return error directly, just run another campaign loop

### Check List 

Tests 

 - Unit test
 - Integration test

### Release Note

- Fix owner may not run the main work loop because some errors are not handled correctly.


leoppro:
LGTM

PR #531, opened 2020-05-06
client: Fix the issue that TiKV reconnecting may make CDC stuck for a while

MyonKeminta:

This PR tries to fix the issue about reconnecting TiKV may make CDC stuck for a while, and added some debug logs for investigation.

### Check List 

Tests 

 - Unit test
- Integration test

### Release note

- Fix the issue that reconnecting to TiKV may cause CDC stuck for a while

### DIFF
--- a/cdc/kv/client_test.go
+++ b/cdc/kv/client_test.go
@@ -107,35 +107,30 @@ func (s *mockChangeDataService) EventFeed(server cdcpb.ChangeData_EventFeedServe
 	return nil
 }
 
-func newMockService(c *check.C, port int, ch chan *cdcpb.ChangeDataEvent) *grpc.Server {
+func newMockService(c *check.C, port int, ch chan *cdcpb.ChangeDataEvent, wg *sync.WaitGroup) *grpc.Server {
 	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
 	c.Assert(err, check.IsNil)
 	grpcServer := grpc.NewServer()
 	mockService := &mockChangeDataService{c: c, ch: ch}
 	cdcpb.RegisterChangeDataServer(grpcServer, mockService)
+	wg.Add(1)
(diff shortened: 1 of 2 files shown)


cdc/kv/client_test.go
```diff
@@ -61,6 +70,128 @@ func (s *clientSuite) TestUpdateCheckpointTS(c *check.C) {
...
+	defer func() {
+		if ch2 != nil {
+			close(ch2)
+		}
+	}()
+	server2 := newMockService(c, 23376, ch2)
+	defer func() {
+		// if server1 != nil {
+		// 	server1.Stop()
+		// }
```
overvenus, line 135: Could you remove these comments?

cdc/kv/client_test.go
```diff
@@ -61,6 +70,128 @@ func (s *clientSuite) TestUpdateCheckpointTS(c *check.C) {
...
+	cluster.AddStore(2, "localhost:23376")
+	cluster.Bootstrap(3, []uint64{1, 2}, []uint64{4, 5}, 4)
+
+	cdcClient, err := NewCDCClient(pdClient, kvStorage.(tikv.Storage))
+	c.Assert(err, check.IsNil)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	eventCh := make(chan *model.RegionFeedEvent, 10)
+	go func() {
+		err := cdcClient.EventFeed(ctx, util.Span{Start: []byte("a"), End: []byte("b")}, 1, eventCh)
```
amyangfei, line 157: `cdcClient.EventFeed` may not return before the test case tear down, what about adding some wait mechanism

cdc/kv/client_test.go
MyonKeminta: Oh, sorry.

leoppro:
LGTM

PR #546, opened 2020-05-08
cmd: support pause/resume/remove changefeed 

overvenus:

* Support pause/resume/remove changefeed via cdc cli.
* Add status address to capture info.

Cc https://github.com/pingcap/ticdc/issues/542

### Check List 

Tests 

 - Integration test

Code changes

Related changes

### Release note

- Support pause/resume/remove changefeed via cdc cli.


cdc/http_status.go
```diff
@@ -51,13 +52,20 @@ func (s *Server) startStatusHTTP() {
 
 	addr := fmt.Sprintf("%s:%d", s.opts.statusHost, s.opts.statusPort)
 	s.statusServer = &http.Server{Addr: addr, Handler: serverMux}
-	log.Info("status http server is running", zap.String("addr", addr))
+	errCh := make(chan error)
 	go func() {
-		err := s.statusServer.ListenAndServe()
+		ln, err := net.Listen("tcp", addr)
```
leoppro, line 57: move this line out of this goroutine? we can avoid to using error channel

overvenus:
/run-integration-tests

PTAL, thanks

cmd/server.go
```diff
@@ -34,7 +34,7 @@ func init() {
 	rootCmd.AddCommand(serverCmd)
 
 	serverCmd.Flags().StringVar(&serverPdAddr, "pd", "http://127.0.0.1:2379", "PD address, separated by comma")
-	serverCmd.Flags().StringVar(&statusAddr, "status-addr", "0.0.0.0:8300", "Bind address for http status server")
+	serverCmd.Flags().StringVar(&statusAddr, "status-addr", "127.0.0.1:8300", "Bind address for http status server")
```
amyangfei, line 37: DBA suggests setting the default value for `status host` to `0.0.0.0`, the same with TiDB

overvenus: 0.0.0.0 in capture info is meaningless. 127.0.0.1 makes user assign a meaningful address.

amyangfei:
btw, as this PR changes `status-addr` to `addr`, we also need to update parameter in [tipocket](https://github.com/pingcap/tipocket/blob/master/pkg/test-infra/cdc/recommend.go#L95) and [cdc document](https://github.com/pingcap/docs-cn/blob/master/reference/tools/ticdc/deploy.md)

PR #554, opened 2020-05-11
mounter: use a chan list to avoid CPU cost with select lock

amyangfei:
Fix https://github.com/pingcap/ticdc/issues/552

- allocate a chan for each codec worker in mounter 
- dispatch raw changed event randomly to codec worker. (We can do this because we don't care the sequence of kv event in mounter)

After this optimization, the most CPU consumption of codcWorker is `unmarshalAndMountRowChanged`. Besides we can increase `mounter-worker-num` to achieve liner codec performance

Screen Shot 2020-05-11 at 17 41 17

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/entry/mounter.go
+++ b/cdc/entry/mounter.go
@@ -19,6 +19,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


leoppro:
LGTM

PR #572, opened 2020-05-14
*: a new table migration mechanism

leoppro:

- Unify the variable types of tableID
- Deprecated the table lock and using undo list(Operation in TaskStatus) to maintain table migration
- support to add or remove more than one tables in one tick

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/changefeed.go
+++ b/cdc/changefeed.go
@@ -18,8 +18,6 @@ import (
 	"fmt"
 	"math"
 
-	"github.com/pingcap/ticdc/pkg/cyclic"
-
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	timodel "github.com/pingcap/parser/model"
@@ -27,6 +25,7 @@ import (
 	"github.com/pingcap/ticdc/cdc/kv"
 	"github.com/pingcap/ticdc/cdc/model"
 	"github.com/pingcap/ticdc/cdc/sink"
+	"github.com/pingcap/ticdc/pkg/cyclic"
 	"github.com/pingcap/ticdc/pkg/filter"
 	"go.etcd.io/etcd/mvcc/mvccpb"
 	"go.uber.org/zap"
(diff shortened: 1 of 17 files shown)


cdc/changefeed.go
```diff
@@ -188,151 +185,108 @@ func (c *changeFeed) minimumTablesCapture(captures map[string]*model.CaptureInfo
...
-			c.taskStatus[captureID] = newInfo
-		}
-		switch errors.Cause(err) {
-		case model.ErrFindPLockNotCommit:
-			c.restoreTableInfos(statusClone, captureID)
-			log.Info("write table info delay, wait plock resolve",
-				zap.String("changefeed", c.id),
-				zap.String("capture", captureID))
-		case nil:
-			log.Info("dispatch table success",
```
amyangfei, line 314: we can add some logs like this to record `table dispatch` history, which is often useful to diagnosis

cdc/model/owner.go
```diff
@@ -107,59 +80,89 @@ func (tp *TaskPosition) Unmarshal(data []byte) error {
...
+			return true
 		}
 	}
+	return false
+}
 
-	return nil, false
+// AppliedTs returns a Ts which less or equal to the ts boundary of any unapplied operation
+func (ts *TaskStatus) AppliedTs() Ts {
+	log.Info("show status", zap.Stringer("status", ts))
```
amyangfei, line 141: will this log too much?

leoppro: just for debug

amyangfei:
Remember to remove debug log

pkg/notify/notify_test.go
```diff
@@ -44,16 +48,38 @@ func (s *notifySuite) TestNotifyHub(c *check.C) {
 }
 
 func (s *notifySuite) TestContinusStop(c *check.C) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
amyangfei, line 51: why this test case is changed?

leoppro: the 5000 goroutine started in original code is too slow in my computer

PR #583, opened 2020-05-18
schedule: a schedule framework

leoppro:

- implement a schedule framework
- implement a table-number-based scheduler
- add some tests

### Check List 

Tests 

 - Unit test
 - Integration test

test rebanlance manually.


cdc/model/owner.go
```diff
@@ -100,13 +99,31 @@ func (o *TableOperation) Clone() *TableOperation {
 	return &clone
 }
 
+// TaskWorkload records the workloads of a task
+type TaskWorkload map[int64]uint64
```
overvenus, line 103: What do int64 and uint64 mean?

overvenus: Can we use a struct instead of uint64? Struct is more extendable.

leoppro:
@overvenus  PTAL again

cdc/model/owner.go
```diff
@@ -100,13 +99,32 @@ func (o *TableOperation) Clone() *TableOperation {
 	return &clone
 }
 
+// TaskWorkload records the workloads of a task
+// the value of the struct is the workload
+type TaskWorkload map[TableID]uint64
```
overvenus, line 104: What about this? So we can add more fields if needed.
```
type TaskWorkload map[TableID]struct { something uint64 }
```

leoppro:
```
package main

import (
	"fmt"
)

type TaskWorkload map[int]struct{ workload uint64 }

func main() {
	a := make(TaskWorkload)

	// we have to insert value like this ....
	a[1] = struct{ workload uint64 }{workload: 1}
	b := a[1].workload
	fmt.Println(b)
}
```

leoppro:
@overvenus  PTAL again

PR #585, opened 2020-05-19
*: add some fail-fast checks

leoppro:
add some fail-fast checks
fix a bug about the wrong restolved ts in puller

### Release note

- No release note


cdc/kv/client.go
```diff
@@ -1134,6 +1134,14 @@ func (s *eventFeedSession) singleEventFeed(
 							CRTs:    entry.CommitTs,
 						},
 					}
+
+					if atomic.LoadUint32(&initialized) == 1 && entry.CommitTs <= lastResolvedTs {
```
overvenus, line 1138:
```suggestion
					if entry.CommitTs <= lastResolvedTs && atomic.LoadUint32(&initialized) == 1 {
```

Please check resolved ts first so that we can avoid atomic load every time.

PR #587, opened 2020-05-19
tests: enable unit test in puller module

amyangfei:

test case is not registered in `check` in puller module

add `check.TestingT(t)

### Check List 

Tests 

 - Unit test

### Release note

- No release note


leoppro:
LGTM

PR #588, opened 2020-05-20
config: reduce the structure of config file

leoppro:
config: reduce the structure of config file

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cmd/changefeed.toml
+++ b/cmd/changefeed.toml
@@ -1,24 +1,48 @@
 
-filter-case-sensitive = false
-ignore-txn-commit-ts = []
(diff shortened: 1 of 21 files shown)


cmd/changefeed.toml
```diff
@@ -1,24 +1,48 @@
 
-filter-case-sensitive = false
-ignore-txn-commit-ts = []
+# 指定配置文件中涉及的库名、表名是否为大小写敏感的
+# 该配置会同时影响 filter 和 sink 相关配置，默认为 true
```
overvenus, line 3: Please use English.

leoppro: done

PR #591, opened 2020-05-21
http api: reduce status api

leoppro:
reduce status api

### Check List 

Tests 

 - Unit test


leoppro:
@overvenus PTAL

PR #593, opened 2020-05-22
scheduler: support to scheduler manually

leoppro:

1. support to scheduler manually
blocked by #595 
access HTTP API like this:
```
curl 127.0.0.1:8300/capture/owner/move_table -X POST -d 'cf-id=cf060953-036c-4f31-899f-5afa0ad0c2f9&target-cp-id=6f19a6d9-0f8c-4dc9-b299-3ba7c0f216f5&table-id=49'
```
we can move the table(49) from a capture to another.

2. add the config item about `scheduler`

add the follow config item to the config file of changefeed

```
[scheduler]
type = "manual" // support "manual", "table-number" // the default value is manual
polling-time = 5 // the tick time of rebalance 
```

3. Tigger the rebalance when the cdc cluster is expanded

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/http_handler.go
+++ b/cdc/http_handler.go
@@ -125,8 +125,8 @@ func (s *Server) handleRebanlanceTrigger(w http.ResponseWriter, req *http.Reques
 		return
 	}
 	changefeedID := req.Form.Get(APIOpVarChangefeedID)
-	err = s.owner.TriggerRebanlance(changefeedID)
-	handleOwnerResp(w, err)
+	s.owner.TriggerRebanlance(changefeedID)
+	handleOwnerResp(w, nil)
 }
 
 func (s *Server) handleMoveTable(w http.ResponseWriter, req *http.Request) {
--- a/cdc/http_handler.go
+++ b/cdc/http_handler.go
@@ -126,7 +126,7 @@ func (s *Server) handleRebanlanceTrigger(w http.ResponseWriter, req *http.Reques
 	}
 	changefeedID := req.Form.Get(APIOpVarChangefeedID)
 	if !util.IsValidUUIDv4(changefeedID) {
-		writeError(w, http.StatusBadRequest, errors.Errorf("invalid changefeed id: %d", changefeedID))
+		writeError(w, http.StatusBadRequest, errors.Errorf("invalid changefeed id: %s", changefeedID))
 		return
 	}
 	s.owner.TriggerRebanlance(changefeedID)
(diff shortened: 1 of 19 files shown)


cdc/http_handler.go
```diff
@@ -119,6 +123,33 @@ func (s *Server) handleRebanlanceTrigger(w http.ResponseWriter, req *http.Reques
...
+	}
+
+	err := req.ParseForm()
+	if err != nil {
+		writeInternalServerError(w, err)
+		return
+	}
+	changefeedID := req.Form.Get(APIOpVarChangefeedID)
+	to := req.Form.Get(APIOpVarTargetCaptureID)
+	tableIDStr := req.Form.Get(APIOpVarTableID)
```
overvenus, line 147: Better to validate inputs first.

cdc/http_handler.go
```diff
@@ -28,6 +28,10 @@ const (
 	APIOpVarAdminJob = "admin-job"
 	// APIOpVarChangefeedID is the key of changefeed ID in HTTP API
 	APIOpVarChangefeedID = "cf-id"
+	// APIOpVarTargetCaptureID is the key of to-capture ID in HTTP API
+	APIOpVarTargetCaptureID = "target-cp-id"
```
overvenus, line 32: Please do not abbreviate terms in API.
```suggestion
	APIOpVarTargetCaptureID = "target-capture-id"
```

cdc/owner.go
```diff
@@ -335,17 +341,28 @@ func (o *Owner) loadChangeFeeds(ctx context.Context) error {
...
+		}
+		o.rebanlanceMu.Unlock()
+		err := changefeed.tryBalance(ctx, o.captures, rebanlanceNow, scheduleCommands)
 		if err != nil {
 			return errors.Trace(err)
 		}
 	}
+	o.rebanlanceMu.Lock()
+	o.rebanlanceForAllChangefeed = false
+	o.rebanlanceMu.Unlock()
```
overvenus, line 365: Could it miss rebalance if a capture is added right before these lines?

cdc/http_handler.go
leoppro: https://github.com/pingcap/ticdc/pull/593/files#diff-20d40ae92eaf96d272e76a7985938f91R30
Should we be unified with him `cf-id`?

cdc/http_handler.go
leoppro: I validate the inputs here: https://github.com/pingcap/ticdc/pull/593/files#diff-1956ba2b6a217651590eb6f946442117R444
If the user inputs incorrectly, we should provide better error info. but it's hard to validate the inputs here, because we can't access changefeed or capture directly. This will cause the data race problems.
maybe we should add a TODO here?

cdc/http_handler.go
overvenus: Oh yes please!

cdc/http_handler.go
overvenus: Just make sure `from` and `to` are valid UUIDv4 string.

PR #605, opened 2020-05-29
kvclient: advance the max receive msg size of GRPC

leoppro:
advance the max receive msg size of GRPC

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


cdc/kv/client.go
```diff
@@ -53,6 +53,7 @@ const (
 	tikvRequestMaxBackoff     = 20000 // Maximum total sleep time(in ms)
 	grpcInitialWindowSize     = 1 << 30
 	grpcInitialConnWindowSize = 1 << 30
+	grpcInitialMaxRecvMsgSize = 1 << 30
```
overvenus, line 56: Please add a comment about it.

PR #621, opened 2020-06-06
util: fix local timezone parse error

### LINKED ISSUE #619: Using readlink to parse time zone name will fail with legacy time zone names
See https://asktug.com/t/topic/35048 for context.

The user's localtime is linked to `/usr/share/zoneinfo/UTC`, so the extracted name is wrongly determined as `zoneinfo/UTC` rather than just `UTC`.

https://github.com/pingcap/ticdc/blob/2d2c5dbaa01ff59a33dd2cf0d9716e03dc2eee92/pkg/util/tz.go#L42-L46

Instead of `time.LoadLocation` you should just read the file and use [`time.LoadLocationFromTZData`](https://pkg.go.dev/time?tab=doc#LoadLocationFromTZData) 🤷.

amyangfei:

Fix #619 

- if `region` information from zone file is `zoneinfo`, use the last element of zone file only.
- use `path/filepath` to instead `path`

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- No release note

### DIFF
--- a/pkg/util/tz.go
+++ b/pkg/util/tz.go
@@ -15,7 +15,7 @@ package util
 
 import (
 	"os"
-	"path"
+	"path/filepath"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


amyangfei:
@leoppro CI is broken because certificate for download.pingcap.org is expired, still wait SRE to fix it. (or we have to use HTTP or `--no-check-certificate` option)

PR #627, opened 2020-06-08
pkg/util: fix check version in git describe format

overvenus:

PD uses version in git describe format which has a string `-[0-9]+-g[0-9a-f]{8,}`. It breaks semver comparison.

Remove the string `-[0-9]+-g[0-9a-f]{8,}`

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Fix check version in git describe format.

### DIFF
--- a/pkg/util/check.go
+++ b/pkg/util/check.go
@@ -30,9 +30,13 @@ import (
 var minPDVersion *semver.Version = semver.New("4.0.0-rc.1")
 var minTiKVVersion *semver.Version = semver.New("4.0.0-rc.1")
 
-func removeV(v string) string {
+func removeVAndHash(v string) string {
 	if v == "" {
 		return v
+	}
+	hash := regexp.MustCompile("-[0-9]+-g[0-9a-f]{8,}")
+	if hash.Match([]byte(v)) {
+		v = hash.ReplaceAllLiteralString(v, "")
 	}
 	return strings.TrimPrefix(v, "v")
 }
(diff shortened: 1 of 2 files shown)


pkg/util/check.go
```diff
@@ -30,10 +30,14 @@ import (
 var minPDVersion *semver.Version = semver.New("4.0.0-rc.1")
 var minTiKVVersion *semver.Version = semver.New("4.0.0-rc.1")
 
-func removeV(v string) string {
+func removeVAndHash(v string) string {
 	if v == "" {
 		return v
 	}
+	hash := regexp.MustCompile("-[0-9]+-g[0-9a-f]{8,}")
```
amyangfei, line 37: After this, comparing between `4.0.0-8-g30f0b014` and `4.0.0-9-g31dae220` seems not work

overvenus: They are the same. Semver should not include `-8-g30f0b014`.

overvenus:
Looks like file_sort test is unstable.

pkg/util/check.go
amyangfei: According to [semver.org](https://semver.org/#spec-item-9), `4.0.0-8-g30f0b014` is a pre-release version. Well as we don't need to compare `4.0.0-8-g30f0b014` and `4.0.0-9-g31dae220`, it looks all right to treat `4.0.0-8-g30f0b014` as `4.0.0`
(maybe `4.0.0+8-g30f0b014` is better, where `+8-g30f0b014` is used as `build metadata`)

leoppro:
LGTM

PR #628, opened 2020-06-08
Enable support for AutoRandom in tidb sink

liuzix:
When syncing data from tidb to tidb, if the primary key is AutoRandom, insert operations will fail because by default the db session to the sink does not allow writing to the AutoRandom key.

Adding code to set the session veriable allow_auto_random_explicit_insert to true when creating a mysqlSink.

### Check List 

Tests 

### DIFF
--- a/cdc/changefeed.go
+++ b/cdc/changefeed.go
@@ -16,7 +16,6 @@ package cdc
 import (
 	"context"
 	"fmt"
-	"github.com/pingcap/tidb/sessionctx/binloginfo"
 	"math"
 	"time"
 
@@ -30,6 +29,7 @@ import (
 	"github.com/pingcap/ticdc/pkg/cyclic"
 	"github.com/pingcap/ticdc/pkg/filter"
 	"github.com/pingcap/ticdc/pkg/scheduler"
+	"github.com/pingcap/tidb/sessionctx/binloginfo"
 	"go.etcd.io/etcd/mvcc/mvccpb"
 	"go.uber.org/zap"
 )
--- a/cdc/changefeed.go
+++ b/cdc/changefeed.go
@@ -16,6 +16,7 @@ package cdc
 import (
 	"context"
 	"fmt"
+	"github.com/pingcap/tidb/sessionctx/binloginfo"
 	"math"
 	"time"
 
@@ -639,6 +640,8 @@ func (c *changeFeed) handleDDL(ctx context.Context, captures map[string]*model.C
 		return errors.Trace(err)
 	}
 	if !c.cyclicEnabled || c.info.Config.Cyclic.SyncDDL {
+		ddlEvent.Query = binloginfo.AddSpecialComment(ddlEvent.Query)
+		log.Debug("DDL processed to make special features mysql-compatible", zap.String("query", ddlEvent.Query))
 		err = c.sink.EmitDDLEvent(ctx, ddlEvent)
 	}
 	// If DDL executing failed, pause the changefeed and print log, rather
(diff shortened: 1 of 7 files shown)


cdc/sink/mysql.go
```diff
@@ -240,6 +240,17 @@ func (s *mysqlSink) adjustSQLMode(ctx context.Context) error {
 	return errors.Trace(err)
 }
 
+// enableAutoRandom enables writing AutoRandom values to the sink by setting
+// the session variable "allow_auto_random_explicit_insert".
+func (s *mysqlSink) enableAutoRandom(ctx context.Context) error {
+	rows, err := s.db.QueryContext(ctx, fmt.Sprintf("SET allow_auto_random_explicit_insert = true;"))
```
overvenus, line 246:
```suggestion
	rows, err := s.db.QueryContext(ctx, "SET allow_auto_random_explicit_insert = true;")
```

cdc/sink/mysql.go
```diff
@@ -365,6 +377,13 @@ func newMySQLSink(ctx context.Context, sinkURI *url.URL, dsn *dmysql.Config, fil
 		}
 	}
 
+	if scheme == "tidb" {
```
overvenus, line 380: What about other sinks?

cdc/sink/mysql.go
```diff
@@ -254,14 +254,36 @@ var defaultParams = params{
...
+	if err != nil && err != sql.ErrNoRows {
+		return "", errors.Trace(err)
+	}
+
+	if err == nil && autoRandomInsertEnabled == "off" {
+		dsnCfg.Params["allow_auto_random_explicit_insert"] = "1"
+		log.Debug("Set allow_auto_random_explicit_insert to 1")
+	}
+
+	log.Info("Connection string:", zap.String("dsnStr", dsnCfg.FormatDSN()))
```
overvenus, line 286: Could it leak password?

cdc/sink/mysql.go
```diff
@@ -279,9 +301,10 @@ func newMySQLSink(ctx context.Context, sinkURI *url.URL, dsn *dmysql.Config, fil
 	tz := util.TimezoneFromCtx(ctx)
 
 	var dsnStr string
+	var scheme string
 	switch {
 	case sinkURI != nil:
-		scheme := strings.ToLower(sinkURI.Scheme)
+		scheme = strings.ToLower(sinkURI.Scheme)
```
overvenus, line 307:
```suggestion
		scheme := strings.ToLower(sinkURI.Scheme)
```

cdc/sink/mysql.go
```diff
@@ -279,9 +301,10 @@ func newMySQLSink(ctx context.Context, sinkURI *url.URL, dsn *dmysql.Config, fil
 	tz := util.TimezoneFromCtx(ctx)
 
 	var dsnStr string
+	var scheme string
```
overvenus, line 304:
```suggestion
```

cdc/sink/mysql.go
```diff
@@ -132,6 +133,7 @@ func (s *mysqlSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error
 		)
 		return nil
 	}
+	ddl.Query = binloginfo.AddSpecialComment(ddl.Query)
```
leoppro, line 136: I think we should log the specially processed query

cdc/changefeed.go
```diff
@@ -16,6 +16,7 @@ package cdc
 import (
 	"context"
 	"fmt"
+	"github.com/pingcap/tidb/sessionctx/binloginfo"
```
amyangfei, line 19: Please follow the import rules
```
standard lib

third-party lib
```

or try to use tools like https://godoc.org/golang.org/x/tools/cmd/goimports

PR #633, opened 2020-06-09
docker: fix the wrong version in docker images

leoppro:
before this PR, the version showed by binary in docker images:
`ReleaseVersion=v4.0.0-24-g366790f-dev`

after this PR: `ReleaseVersion=v4.0.0-24-g366790f`

### Check List 

Tests 

 - Unit test
 - Integration test


.dockerignore
```diff
@@ -1 +1,46 @@
-.gitignore
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
```
amyangfei, line 1: Just curious is it possible to use a soft link to `.giignore` file

leoppro: good idea

PR #647, opened 2020-06-11
Fix crts maybe less than emit resolved ts in processor

### LINKED ISSUE #644: Fatal error: Commit Ts is less than Resolved Ts
## Bug Report

The CI fails with a fatal error: The CRTs must be greater than the resolvedTs. It means we may mistakenly advance resolved ts, and it could cause data lose.

[fatal.log](https://github.com/pingcap/ticdc/files/4764392/fatal.log)

No fatal error.

3. What did you see instead?

4. Versions of the cluster

        ```
        master
        ```

        ```
        https://github.com/pingcap/ticdc/pull/642/
        ```

amyangfei:

Fix #644 

Found two bugs:

~1. sinkEmittedResolvedTs and real emitted row changed events
    - each sorter sends row changed events to the output channel of a processor.
    - but resolved ts of each table is updated asynchronously, which means if the resolved ts of a table is updated, some row changed events before this resolved ts may be not consumed by processor yet.~ as `processor.output` chan is sequence, when resolved ts event is fetched, row changed events must have been consumed.
2. processor local resolved ts calculation and mark table (Then why the cyclic test often meets this bug makes sense )
    - processor receives and emits row changed events of mark table, but mark table is not maintained in processor table list, and when calculating processor local resolved ts, the mark table is not taken into consideration. https://github.com/pingcap/ticdc/blob/master/cdc/processor.go#L827-L833

~- Update table's resolved ts and consume row changed events in the same routine in processor.~
- Maintain both normal tables and mark tables in processor.

### Check List 

Tests 

 - Integration test

Related changes

### Release note

- No release note

### DIFF
--- a/cdc/processor.go
+++ b/cdc/processor.go
@@ -128,10 +128,6 @@ func (t *tableInfo) loadResolvedTs() uint64 {
 	return tableRts
 }
 
-func (t *tableInfo) storeResolvedTs(ts uint64, rts *uint64) {
-	atomic.StoreUint64(rts, ts)
-}
-
 // newProcessor creates and returns a processor for the specified change feed
 func newProcessor(
 	ctx context.Context,
@@ -815,7 +811,7 @@ func (p *processor) addTable(ctx context.Context, tableID int64, replicaInfo *mo
 						continue
 					}
 					if pEvent.RawKV != nil && pEvent.RawKV.OpType == model.OpTypeResolved {
-						table.storeResolvedTs(pEvent.CRTs, pResolvedTs)
+						atomic.StoreUint64(pResolvedTs, pEvent.CRTs)
 						p.localResolvedNotifier.Notify()
 						resolvedTsGauge.Set(float64(oracle.ExtractPhysical(pEvent.CRTs)))
 						continue
--- a/cdc/processor.go
+++ b/cdc/processor.go
@@ -128,12 +128,8 @@ func (t *tableInfo) loadResolvedTs() uint64 {
 	return tableRts
 }
 
-func (t *tableInfo) storeResolvedTs(ts uint64, isMarkTable bool) {
-	if isMarkTable {
-		atomic.StoreUint64(&t.mResolvedTs, ts)
-	} else {
-		atomic.StoreUint64(&t.resolvedTs, ts)
-	}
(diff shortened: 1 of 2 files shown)


cdc/processor.go
```diff
@@ -109,18 +110,31 @@ type processor struct {
 }
 
 type tableInfo struct {
-	id         int64
-	resolvedTS uint64
-	workload   model.WorkloadInfo
-	cancel     context.CancelFunc
+	id          int64
+	resolvedTs  uint64
+	mid         int64
```
leoppro, line 115:
```suggestion
	markTableId         int64
```

cdc/processor.go
```diff
@@ -741,14 +754,14 @@ func (p *processor) addTable(ctx context.Context, tableID int64, replicaInfo *mo
...
-		resolvedTS: replicaInfo.StartTs,
+		resolvedTs: replicaInfo.StartTs,
 		cancel:     cancel,
 	}
 	// TODO(leoppro) calculate the workload of this table
 	// We temporarily set the value to constant 1
 	table.workload = model.WorkloadInfo{Workload: 1}
 
-	startPuller := func(tableID model.TableID) {
+	startPuller := func(tableID model.TableID, isMarkTable bool) {
```
leoppro, line 764: how about this?
```suggestion
	startPuller := func(tableID model.TableID, resolvedTs *uint64) {
```
and we can remove `storeResolvedTs` function, just using `atomic.StoreUint64(resolvedTs, ts)`

leoppro: we can avoid having to judge in a loop by this way

cdc/processor.go
```diff
@@ -109,18 +109,27 @@ type processor struct {
...
+		if mTableRts < tableRts {
+			return mTableRts
+		}
+	}
+	return tableRts
 }
 
-func (t *tableInfo) storeResolvedTS(ts uint64) {
-	atomic.StoreUint64(&t.resolvedTS, ts)
+func (t *tableInfo) storeResolvedTs(ts uint64, rts *uint64) {
```
leoppro, line 131: I think we can remove this function 😊

amyangfei: fine

PR #648, opened 2020-06-12
mysql sink: fix some bugs and reduce some structures in mysql sink

leoppro:

1. fix the bug about the `txn_size_limiter` may splitting the transactions from upstream and breaking the consistency of transactions
1. Reduced the granularity of conflict detection and improved concurrency
1. Refactored the mysqlSinkWorker
1. Refactored a part of the cyclic replication

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Reduce the granularity of conflict detection and improve the concurrency in MySQL sink.


cdc/sink/mysql.go
```diff
@@ -389,163 +403,153 @@ func newMySQLSink(ctx context.Context, sinkURI *url.URL, dsn *dmysql.Config, fil
...
+			if txn.ReplicaID != replicaID || len(toExecRows)+len(txn.Rows) > w.maxTxnRow {
+				if err := flushRows(); err != nil {
+					return errors.Trace(err)
+				}
+			}
+			replicaID = txn.ReplicaID
+			toExecRows = append(toExecRows, txn.Rows...)
+			txnNum++
+		case <-t.C:
+			if time.Since(lastExecTime) < time.Millisecond*100 {
```
amyangfei, line 545: Why use another time gap checker in a time ticker

pkg/cyclic/mark.go
```diff
@@ -96,98 +95,56 @@ func (m MarkMap) shouldFilterTxn(startTs uint64, filterReplicaIDs []uint64) (*mo
...
+		}
+		if len(filteredTxns) == 0 {
+			delete(txnsMap, table)
+		} else {
+			txnsMap[table] = filteredTxns
 		}
 	}
-	for _, events := range output {
-		// Per table order may lose during map stage, we need to sort events.
-		sort.Slice(events, func(i, j int) bool { return events[i][0].CommitTs < events[j][0].CommitTs })
```
overvenus, line 190: Why can it be removed?

cdc/sink/mysql.go
overvenus, line 545: Why not 50ms?

leoppro: I want to implement the following logic:

If there is no execution for more than 100 ms, then call `flushRows` 

but not:

Call `flushRows` every 100ms, whether or not it is triggered within this 100ms.

leoppro: ditto

pkg/cyclic/mark.go
leoppro: I didn't add `mark table event` into `events`, so we didn't lose the order in my implement

leoppro:
@amyangfei @overvenus  PTAL

cdc/sink/mysql.go
```diff
@@ -597,44 +604,36 @@ func (s *mysqlSink) execDMLWithMaxRetries(
...
 		}
 		sqls = append(sqls, query)
 		values = append(values, args)
 	}
-	if s.cyclic != nil && !markRowFound && len(rows) > 0 {
+	if s.cyclic != nil && len(rows) > 0 {
 		// Write mark table with the current replica ID.
 		updateMark := s.cyclic.UdpateTableCyclicMark(
-			rows[0].Table.Schema, rows[0].Table.Table, uint64(bucket), s.cyclic.ReplicaID())
+			rows[0].Table.Schema, rows[0].Table.Table, uint64(bucket), replicaID)
```
overvenus, line 628: It looks like fail 3 clusters cyclic replication #656 .

pkg/cyclic/mark.go
```diff
@@ -96,98 +95,56 @@ func (m MarkMap) shouldFilterTxn(startTs uint64, filterReplicaIDs []uint64) (*mo
...
-						if event.CommitTs != first.CommitTs || event.RowID != first.RowID {
-							log.Fatal(
-								"there should be at most one mark row for each txn",
-								zap.Uint64("start-ts", event.StartTs),
-								zap.Any("first", first),
-								zap.Any("second", event))
-						}
+// FilterAndReduceTxns filters duplicate txns bases on filterReplicaIDs and
+// set the replicaID into txns
+func FilterAndReduceTxns(txnsMap map[model.TableName][]*model.Txn, filterReplicaIDs []uint64, replicaID uint64) {
```
overvenus, line 100: Please enhance comments, the function only set replicaID if a txn does not have one.

tests/cyclic_abc/run.sh
```diff
@@ -16,7 +16,7 @@ function run() {
 
     rm -rf $WORK_DIR && mkdir -p $WORK_DIR
```
overvenus, line 18: Please also uncomment `set -e` at line 3.

PR #651, opened 2020-06-12
test: kill the process to avoid port occupancy issues

leoppro:

kill the process to avoid port occupancy issues

### Check List 

Tests 

 - Integration test

### DIFF
--- a/tests/_utils/stop_tidb_cluster
+++ b/tests/_utils/stop_tidb_cluster
@@ -6,3 +6,24 @@ killall -w -s 9 tidb-server || true
 killall -w -s 9 cdc || true
 killall -w -s 9 cdc.test || true
 killall -w -s 9 tiflash || true
+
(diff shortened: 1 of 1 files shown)


tests/_utils/stop_tidb_cluster
```diff
@@ -6,3 +6,24 @@ killall -w -s 9 tidb-server || true
 killall -w -s 9 cdc || true
 killall -w -s 9 cdc.test || true
 killall -w -s 9 tiflash || true
+
+kill -9 $(lsof -i tcp:4000 -t) &>/dev/null || true
```
amyangfei, line 10: these ports may be customed by env value, https://github.com/pingcap/ticdc/blob/master/tests/_utils/test_prepare.
Could we use these variables instead of the port number directly?

leoppro: fine

PR #656, opened 2020-06-15
cyclic: fix 3 clusters replication

overvenus:

Fix 3 clusters replication failure due to update a nonexistent mark table.

Update the right mark table.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

### Release note

- Fix the issue that CDC can not replicate data between three or more clusters.


overvenus:
PTAL, thanks!

Issue #663, opened 2020-06-16
Add version to CDC gRPC service


overvenus:
## Feature Request

**Is your feature request related to a problem? Please describe:**

CDC gRPC service needs a version to identify compatibility issues between TiKV and TiCDC.

**Describe the feature you'd like:**

Add version field to CDC request.

https://github.com/pingcap/kvproto/blob/8037ca08f377c8f041ff4a2bf9a7cdbce5479d9a/proto/cdcpb.proto#L18-L20

**Describe alternatives you've considered:**

Instead of adding a version field we could
* Add a new RPC with a version suffix, eg.  `rpc EventFeedV2() {}`.
* Or check TiKV version recorded in PD. (Note: the version bumps after all tikv has been upgraded)

**Teachability, Documentation, Adoption, Migration Strategy:**

N/A

amyangfei:
have supported

PR #672, opened 2020-06-18
kv,util: check TiKV version before event feed and refine version check

overvenus:

Check TiKV version before making EventFeed RPC in order to mitigate compatibility issues.

Cc  #663 

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

Related changes

### Release note

- No release note.

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -291,11 +291,16 @@ func (c *CDCClient) getConn(ctx context.Context, addr string) (*grpc.ClientConn,
 	return ca.Get(), nil
 }
 
-func (c *CDCClient) newStream(ctx context.Context, addr string) (stream cdcpb.ChangeData_EventFeedClient, err error) {
+func (c *CDCClient) newStream(ctx context.Context, addr string, storeID uint64) (stream cdcpb.ChangeData_EventFeedClient, err error) {
 	err = retry.Run(50*time.Millisecond, 3, func() error {
 		conn, err := c.getConn(ctx, addr)
 		if err != nil {
 			log.Info("get connection to store failed, retry later", zap.String("addr", addr), zap.Error(err))
+			return errors.Trace(err)
+		}
+		err = util.CheckStoreVersion(ctx, c.pd, storeID)
+		if err != nil {
+			log.Error("check tikv version failed", zap.Error(err), zap.Uint64("storeID", storeID))
 			return errors.Trace(err)
 		}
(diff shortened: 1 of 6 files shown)


cdc/kv/client.go
```diff
@@ -291,13 +291,18 @@ func (c *CDCClient) getConn(ctx context.Context, addr string) (*grpc.ClientConn,
...
 	err = retry.Run(50*time.Millisecond, 3, func() error {
 		conn, err := c.getConn(ctx, addr)
 		if err != nil {
 			log.Info("get connection to store failed, retry later", zap.String("addr", addr), zap.Error(err))
 			return errors.Trace(err)
 		}
+		err = util.CheckStoreVersion(ctx, c.pd, storeID)
+		if err != nil {
+			log.Error("check tikv version failed", zap.Error(err), zap.Uint64("storeID", storeID))
+			return errors.Trace(err)
```
amyangfei, line 304: can use `backoff.Permanent(err)` here to avoid unnecessary retries

overvenus:
> Firstly we perform rolling update of cdc cluster, but the CheckClusterVersion (which requires tikv-v2) in cdc server will fail the startup of cdc-v2 server.

There is a workaround to break the tie,

* Relaxed version check on startup, warn if TiKV is incompatible, error if PD is incompatible.
* And sends RPC `EventFeed` util TiKV is upgraded.

cdc/kv/client.go
```diff
@@ -621,14 +626,19 @@ MainLoop:
...
-					zap.Uint64("regionID", sri.verID.GetID()), zap.Uint64("requestID", requestID), zap.String("addr", rpcCtx.Addr))
-				stream, err = s.client.newStream(ctx, rpcCtx.Addr)
+					zap.Uint64("regionID", sri.verID.GetID()),
+					zap.Uint64("requestID", requestID),
+					zap.Uint64("storeID", storeID),
+					zap.String("addr", rpcCtx.Addr))
+				stream, err = s.client.newStream(ctx, rpcCtx.Addr, storeID)
 				if err != nil {
 					// if get stream failed, maybe the store is down permanently, we should try to relocate the active store
 					log.Warn("get grpc stream client failed",
```
amyangfei, line 638: A lot of `version is incompatible` errors are logged here, is it better to add some sleep for this error?

overvenus: Updated.

PR #675, opened 2020-06-19
Fix start-ts in row changed event emitted from kafka consumer

amyangfei:

Kafka consumer used in integration test reuses the MySQL sink of TiCDC, MySQL sink groups transaction by `start-ts` of row changed events, but `start-ts` is not contained in TiCDC open protocol.

Just hack to set `start-ts` equals to `commit-ts` of row changed events emitted from Kafka consumer. (We don't change TiCDC open protocol in this PR)

### Check List 

Tests 

 - Integration test

### Release note

- No release note

### DIFF
--- a/kafka_consumer/main.go
+++ b/kafka_consumer/main.go
@@ -378,6 +378,9 @@ ClaimMessages:
 						zap.Int32("partition", partition))
 					break ClaimMessages
 				}
+				// FIXME: hack to set start-ts in row changed event, as start-ts
+				// is not contained in TiCDC open protocol
+				row.StartTs = row.CommitTs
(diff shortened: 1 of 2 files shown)


kafka_consumer/main.go
```diff
@@ -378,6 +378,9 @@ ClaimMessages:
 						zap.Int32("partition", partition))
 					break ClaimMessages
 				}
+				// FIXME: hack to set start-ts in row changed event, as start-ts
+				// is not contained in TiCDC open protocol
+				row.StartTs = row.CommitTs
```
overvenus, line 383: Should we add start ts to open protocol?

PR #686, opened 2020-06-23
*: replace capture ID with address in metrics

overvenus:

Replace capture ID with address in metrics

### Check List 

Tests 

Code changes

 - Has persistent data change

Related changes

### Release note

- No release note

### DIFF
--- a/cdc/processor.go
+++ b/cdc/processor.go
@@ -397,7 +397,7 @@ func (p *processor) ddlPullWorker(ctx context.Context) error {
 }
 
 func (p *processor) workloadWorker(ctx context.Context) error {
-	t := time.NewTicker(1 * time.Minute)
+	t := time.NewTicker(10 * time.Second)
 	err := p.etcdCli.PutTaskWorkload(ctx, p.changefeedID, p.captureID, nil)
 	if err != nil {
 		return errors.Trace(err)
--- a/cdc/processor.go
+++ b/cdc/processor.go
@@ -70,7 +70,7 @@ var (
 
 type processor struct {
 	id           string
-	captureID    string
+	captureInfo  model.CaptureInfo
 	changefeedID string
 	changefeed   model.ChangeFeedInfo
 	limitter     *puller.BlurResourceLimitter
@@ -134,7 +134,8 @@ func newProcessor(
 	session *concurrency.Session,
 	changefeed model.ChangeFeedInfo,
 	sink sink.Sink,
-	changefeedID, captureID string,
+	changefeedID string,
+	captureInfo model.CaptureInfo,
 	checkpointTs uint64,
 	errCh chan error,
 ) (*processor, error) {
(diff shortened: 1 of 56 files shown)


metrics/grafana/ticdc.json
```diff
@@ -1,4 +1,46 @@
...
+    {
+      "name": "DS_TEST-CLUSTER",
+      "label": "test-cluster",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__requires": [
```
overvenus, line 12: I add these two back, are they compatible with tiup? @lichunzhu

metrics/grafana/ticdc.json
```diff
@@ -1,4 +1,46 @@
 {
+  "__inputs": [
+    {
+      "name": "DS_TEST-CLUSTER",
```
lichunzhu, line 4:
```suggestion
      "name": "${DS_TEST-CLUSTER}",
```

overvenus: It is generated from Grafana.

lichunzhu: But tiup use `${DS_TEST-CLUSTER}` to do the replacement.
https://github.com/pingcap/tiup/blob/a33f0971969b5cce3d56784c760af3269368461d/templates/scripts/run_grafana.sh.tpl#L18

lichunzhu: never mind, I think this should work

metrics/grafana/ticdc.json
lichunzhu: I think they are compatible with tiup. @overvenus

scripts/check-copyright.sh
```diff
@@ -1,6 +1,6 @@
 copyright="// Copyright $(date '+%Y') PingCAP, Inc."
 
-result=$(find ./ -name "*.go" ! -path ".//vendor/*" | grep -v '.pb.go' | while read file_path; do
+result=$(find ./ -name "*.go" ! -path ".//vendor/*" | grep -vE '.pb.go|vendor/' | while read file_path; do
```
leoppro, line 3:
```suggestion
result=$(find ./ -name "*.go" | grep -vE '.pb.go|vendor/' | while read file_path; do
```

cdc/processor.go
```diff
@@ -265,7 +266,8 @@ func (p *processor) wait() {
 	err := p.wg.Wait()
 	if err != nil && errors.Cause(err) != context.Canceled {
 		log.Error("processor wait error",
-			zap.String("captureID", p.captureID),
+			zap.Reflect("captureid", p.captureInfo.ID),
```
amyangfei, line 269:
```suggestion
			zap.String("captureid", p.captureInfo.ID),
```

cdc/processor.go
```diff
@@ -265,7 +266,8 @@ func (p *processor) wait() {
...
+			zap.Reflect("captureaddr", p.captureInfo.AdvertiseAddr),
```
amyangfei, line 270:
```suggestion
			zap.String("captureaddr", p.captureInfo.AdvertiseAddr),
```

Issue #660, opened 2020-06-16
Support Avro sink and make TiCDC compatible with Kafka connect


amyangfei:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Apache Kafka provides flexible connectors mechanism, and there exist a lot of third-party connectors implementation which can receive changed logs from upstream in the [Avro](https://avro.apache.org/) protocol, send data to Kafka via `source connector`, and the data in Kafka can be consumed by `sink connector` and replicated to downstream. 

**TODO list**

- [x] Support writing Avro encoded Kafka Connector-compatible data when writing to Kafka.
- [x] Support replicating Avro schema to an open-source `Schema Registry`, such as [Confluent Schema Registry for Kafka](https://github.com/confluentinc/schema-registry)
- [x] Provide a full solution for data replication link, TiDB -> TiCDC -> Kafka -> TiDB, with Avro data from TiCDC to Kafka and using `Kafka sink connector` to consume data from Kafka to TiDB.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

- [TiCDC Avro support research](https://docs.google.com/document/d/1uoJAOiFQTZdEb3jrvqa-B-TAWPrK7PnyqmqrRgEpYKI/edit)

### Value description 

This is an important feature in third-party ecosystem

### Value score 

* (TBD) / 5

### Workload estimation 

* 30 person-day, still need 15 person-day from 7-28

### Time

GanttStart: 2020-07-15
GanttDue: 2020-08-31
GanttProgress: 90%

abbccdda:
Is this still open?

scsldb:
Reason for reprioritization: development complete, testing in process, more important development tasks at the moment.

amyangfei:
- Integration test has been added in https://github.com/pingcap/ticdc/issues/1039
- More test about protocol is still in discussion

PR #687, opened 2020-06-23
Kafka Connect Avro Support Part 1: AvroSchemaManager

liuzix:
### Check List 

Tests 

 - Unit test

Code changes

### Release note

This is the first PR that will lead to support for outputting Avro-serialized data to Kafka. See more in #660. After research and deliberation, we gave up on syncing the Registry Schema ID through Etcd.


cdc/sink/schema_registry.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	log.Info("Avro schema lookup cache miss",
+		zap.String("key", key),
+		zap.Int64("tiSchemaID", tiSchemaID))
+
+	uri := m.registryURL + "/subjects/" + url.QueryEscape(tableNameToSchemaSubject(tableName)) + "/versions/latest"
+	log.Debug("Querying for latest schema", zap.String("uri", uri))
+
+	req, err := http.NewRequest("GET", uri, nil)
```
overvenus, line 168: Please leave a TODO about TLS support.

cdc/sink/schema_registry.go
```diff
@@ -0,0 +1,259 @@
...
+	if err != nil {
+		return errors.Annotate(err, "Failed to read response from Registry")
+	}
+
+	if resp.StatusCode != 200 {
+		log.Warn("Failed to register schema to the Registry, HTTP error",
+			zap.Int("status", resp.StatusCode),
+			zap.String("uri", uri),
+			zap.ByteString("requestBody", payload),
+			zap.ByteString("responseBody", body))
```
overvenus, line 124: Does it contain sensitive data? Eg. password or userdata.

cdc/sink/schema_registry.go
```diff
@@ -0,0 +1,259 @@
...
+
+	var jsonResp registerResponse
+	err = json.Unmarshal(body, &jsonResp)
+
+	if err != nil {
+		return errors.Annotate(err, "Failed to parse result from Registry")
+	}
+
+	if jsonResp.ID == 0 {
+		return errors.New(fmt.Sprintf("Illegal schema ID returned from Registry %d", jsonResp.ID))
```
overvenus, line 136:
```suggestion
		return errors.Errorf("Illegal schema ID returned from Registry %d", jsonResp.ID)
```

cdc/sink/schema_registry.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	log.Info("Registered schema successfully",
+		zap.Int("id", jsonResp.ID),
+		zap.String("uri", uri),
+		zap.ByteString("body", body))
```
overvenus, line 142: ByteString is not readable it produces something like "\u00FF". What about "zap.String"?

cdc/sink/schema_registry_test.go
```diff
@@ -0,0 +1,141 @@
...
+       "name": "test",
+       "fields":
+         [
+           {
+             "type": "string",
+             "name": "field1"
+           },
+           {
+             "type": [
+      			"null",
```
overvenus, line 77:
```suggestion
      	  "null",
```

Please avoid mixing tab and space.

cdc/sink/schema_registry.go
```diff
@@ -0,0 +1,259 @@
...
+	return &AvroSchemaManager{
+		registryURL: registryURL,
+		cache:       make(map[string]*schemaCacheEntry, 1),
+	}, nil
+}
+
+var regexRemoveSpaces = regexp.MustCompile(`\s`)
+
+// Register the latest schema for a table to the Registry, by passing in a Codec
+func (m *AvroSchemaManager) Register(tableName model.TableName, codec *goavro.Codec) error {
```
overvenus, line 92: Is there any go package handles registering schema?

cdc/sink/schema_registry.go
liuzix: If the user configures a registry URL that itself contains sensitive data, e.g. http://username:password@example.com, then it is possible. Other data are strictly about the schema, so they should not contain user data.

cdc/sink/schema_registry.go
liuzix: ByteString is readable when it contains only printable characters, and per my understanding, a JSON body should not contain any unprintable characters. If it does, then something suspicious is definitely happening, so some log output is necessary.

cdc/sink/schema_registry.go
```diff
@@ -0,0 +1,259 @@
...
+		return nil
+	}
+
+	log.Error("Other error when clearing Registry")
+	return err
+}
+
+func tableNameToSchemaSubject(tableName model.TableName) string {
+	// We should guarantee unique names for subjects
+	return url.QueryEscape(tableName.Schema) + "%" + url.QueryEscape(tableName.Table)
```
amyangfei, line 258: `%` could be in schema name or table, then both schema `t%`, table `t` and schema `t`, table `%t` will generate the same subject `t%%t`.
Recommend to use 
`model.QuoteSchema(schema, table)` before query escape

cdc/sink/schema_registry_test.go
```diff
@@ -0,0 +1,262 @@
...
+package sink
+
+import (
+	"encoding/json"
+	"github.com/jarcoal/httpmock"
+	"github.com/linkedin/goavro/v2"
+	"github.com/pingcap/check"
+	"github.com/pingcap/ticdc/cdc/model"
+	"io/ioutil"
+	"net/http"
```
amyangfei, line 23: Please reorg the import library as
```
standard lib

third party lib
```

cdc/sink/schema_registry.go
liuzix: I use QueryEscape first which replaces a plain "%" with "%25", thus avoiding the conflict. I will change tomodel.QuoteSchema() anyways.

cdc/sink/codec/schema_registry.go
```diff
@@ -0,0 +1,293 @@
...
+			continue
+		}
+
+		if resp.StatusCode >= 200 || resp.StatusCode < 300 {
+			break
+		}
+		log.Warn("HTTP server returned with error", zap.Int("status", resp.StatusCode))
+		_ = resp.Body.Close()
+
+		time.Sleep(expBackoff.NextBackOff())
```
amyangfei, line 284: The backoff mechanism doesn't provide a max retry time

amyangfei: Can use `retry.Run` in `github.com/pingcap/ticdc/pkg/retry` for simple

amyangfei:
Please fix data race in the unit test

cdc/sink/codec/schema_registry_test.go
```diff
@@ -0,0 +1,271 @@
...
+	"github.com/pingcap/ticdc/cdc/model"
+)
+
+type AvroSchemaRegistrySuite struct {
+}
+
+var _ = check.SerialSuites(&AvroSchemaRegistrySuite{})
+
+type mockRegistry struct {
+	subjects map[string]*mockRegistrySchema
```
amyangfei, line 35: What about adding a lock to `mockRegistry` and protect the read/write of `subjects` to avoid data race.

PR #689, opened 2020-06-23
Stop changefeed automatically if one of processor exits

amyangfei:

This is part of https://github.com/pingcap/ticdc/issues/664, and tries to release resource if one processor of a changefeed meets error

- If a processor exits with error, upload the error message to `task position`.
- If owner finds error in `task position`, stop the changefeed via admin job. 

### Check List 

Tests 

 - Integration test

### Release note

- Stop changefeed automatically if one of processor exits

### DIFF
--- a/cdc/model/errors.go
+++ b/cdc/model/errors.go
@@ -33,7 +33,7 @@ var (
 
 // RunningError represents some running error from cdc components, such as processor.
 type RunningError struct {
-	ID        string `json:"id"`
-	Error     string `json:"error"`
-	Component string `json:"component"`
+	Addr    string `json:"addr"`
+	Code    string `json:"code"`
+	Message string `json:"message"`
 }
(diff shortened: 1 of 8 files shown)


cdc/model/errors.go
```diff
@@ -30,3 +30,10 @@ var (
 	ErrCaptureNotExist       = errors.New("capture not exists")
 	ErrUnresolved            = errors.New("unresolved")
 )
+
+// RunningError represents some running error from cdc components, such as processor.
+type RunningError struct {
+	ID        string `json:"id"`
```
overvenus, line 36: Could you name it `json:"code"`?

https://github.com/pingcap/tidb/blob/bd28018526397c3ea5e92ee975fea52ec8d6dda7/docs/design/2020-05-08-standardize-error-codes-and-messages.md

amyangfei: good catch, make the `RunningError` compatible with TiDB standard

PR #695, opened 2020-06-28
cdc, pkg: replace table ID with table name in metrics

overvenus:

Replace table ID with table name in metrics.

Cc https://github.com/pingcap/ticdc/issues/677

### Check List 

Tests 

 - No code

Code changes

 - Has persistent data change

Side effects

Related changes

### Release note

- No release note


overvenus:
Base on prometheus spec, label name must match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*`. We need to sanitize table names.

PR #701, opened 2020-06-29
cyclic, cmd: qoute cyclic table name, fatal if loopback and fix a data race

overvenus:

- Correct a few logs
- Fix data race on log
- Fail fast if loopback detected
- Quote mark table name in update SQL.

### Check List 

Tests 

 - Integration test

Related changes

### Release note

- Fix the issue log may cause data race.


overvenus:
Base on prometheus spec, label name must match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*`. We need to sanitize table names.

tests/cyclic_loopback/run.sh
```diff
@@ -1,66 +0,0 @@
-#!/bin/bash
```
amyangfei, line 1: why is this test case removed?

PR #702, opened 2020-06-29
mysql-sink: add more performance related metrics

amyangfei:

Solve https://github.com/pingcap/ticdc/issues/685

- Add bucket size to monitor SQL distribution in each MySQL sink worker.
- Add detect conflict duration metric to monitor wait time for each txn due to conflict.

### Check List 

Tests 

 - Unit test
 - Integration test

Screen Shot 2020-06-29 at 16 28 49

Screen Shot 2020-06-29 at 16 37 01

Related changes

### Release note

- No release note

### DIFF
--- a/cdc/sink/metrics.go
+++ b/cdc/sink/metrics.go
@@ -49,8 +49,8 @@ var (
 			Help:      "Bucketed histogram of conflict detect time (s) for single DML statement",
 			Buckets:   prometheus.ExponentialBuckets(0.00005, 2, 21),
 		}, []string{"capture", "changefeed"})
-	bucketSizeGauge = prometheus.NewGaugeVec(
-		prometheus.GaugeOpts{
+	bucketSizeCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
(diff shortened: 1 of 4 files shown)


cdc/sink/metrics.go
```diff
@@ -41,11 +41,28 @@ var (
...
+			Help:      "Bucketed histogram of conflict detect time (s) for single DML statement",
+			Buckets:   prometheus.ExponentialBuckets(0.00005, 2, 21),
+		}, []string{"capture", "changefeed"})
+	bucketSizeCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "ticdc",
+			Subsystem: "sink",
+			Name:      "bucket_size",
+			Help:      "size of the DML bucket",
+		}, []string{"capture", "changefeed", "bucket"})
```
overvenus, line 58: Looks like gauge is more suitable for size.

amyangfei: Because the sink worker flushes DMLs every 100ms, it is hard to track real bucket size if using gauge

PR #705, opened 2020-06-30
Refine some state transition when handling admin job

amyangfei:

- Fix bug that a paused changefeed can't be removed. item-5 in https://github.com/pingcap/ticdc/issues/542
- Add a `--simple` option in changefeed query command, and output the most common used replication status to users, with more user-friendly output format. part of item-11 in https://github.com/pingcap/ticdc/issues/542
- Make state transition of a changefeed more clearly.

- Expand the feed state, to `normal`, `stopped`, `removed` and `failed`
   - `failed` is an internal state, only exists when the owner creates chagnefeed meets error, in this case, the changefeed can't be resumed, users can re-create a new changefeed.
   - `normal`, `stopped`, `removed` are published to users, and these states can be transformed as follows. 

        | state in column title -> state in row | normal | paused | removed |
        | ---------------------------------- | ------ | ------ | ------- |
        | normal                             | resume | resume | ❌       |
        | paused                             | pause  | pause  | ❌       |
        | removed                            | remove | remove | remove  |

- Minor change to admin job execution model, and don't check whether the job can be executed before job enqueue. Because we can't determine whether the job is valid until the whole jobs before it have been finished.

- sample output of simplified changefeed status query:

      ➜  ./cdc cli changefeed query --simple --changefeed-id=19bdbd27-370e-4022-adc6-b0fb2f838f77
      {
       "state": "normal",
       "tso": 417728968103821313,
       "checkpoint": "2020-06-30 17:32:32.398",
       "error": null
      }

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- No release note

### DIFF
--- a/cdc/http_handler.go
+++ b/cdc/http_handler.go
@@ -17,11 +17,12 @@ import (
 	"context"
 	"net/http"
 	"strconv"
-
-	"github.com/pingcap/ticdc/pkg/util"
+	"time"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/ticdc/cdc/model"
+	"github.com/pingcap/ticdc/pkg/util"
+	"github.com/pingcap/tidb/store/tikv/oracle"
 	"go.etcd.io/etcd/clientv3/concurrency"
 )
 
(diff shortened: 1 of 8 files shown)


cdc/http_handler.go
```diff
@@ -167,3 +175,50 @@ func (s *Server) handleMoveTable(w http.ResponseWriter, req *http.Request) {
...
+	if cf != nil {
+		resp.RunningError = cf.info.Error
+	} else if feedInfo != nil {
+		resp.RunningError = feedInfo.Error
+	}
+	if status != nil {
+		resp.TSO = status.CheckpointTs
+		physical := oracle.ExtractPhysical(status.CheckpointTs)
+		tm := time.Unix(int64(physical/1000), int64(physical)%1000*time.Millisecond.Nanoseconds())
+		resp.Checkpoint = tm.Format("2006-01-02 15:04:05.000")
```
overvenus, line 221: What about 'oracke.GetTimeFromTS()'?

PR #713, opened 2020-07-02
Add rate limit to changefeed error, refine owner rate limit

amyangfei:

Owner could be restarted infinitely due to some changefeed error, such as `newChangeFeed` always fails. We should add more rate limit guard to owner.

- Add error timestamp record in a new `ErrorHis` field in `ChangeFeedInfo`, to pass this data across different captures (as owner can change).
- Delay to start a changefeed if it fails too frequent.
- Add rate limit before a capture campaigns to be owner.
- Re-enable two test cases in `owner_test.go`, update some logic according to current code

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- No release note


cdc/model/changefeed.go
```diff
@@ -58,9 +72,10 @@ type ChangeFeedInfo struct {
 	Engine       SortEngine   `json:"sort-engine"`
 	SortDir      string       `json:"sort-dir"`
 
-	Config *config.ReplicaConfig `json:"config"`
-	State  FeedState             `json:"state"`
-	Error  *RunningError         `json:"error"`
+	Config   *config.ReplicaConfig `json:"config"`
+	State    FeedState             `json:"state"`
+	ErrorHis []int64               `json:"history"`
```
leoppro, line 77: how about using `time.Duration` instead of `int64`?

amyangfei: This filed will be stored into etcd, and may be shared by different captures (as the owner can be transferred), so int64 is used here to store timestamp.

cdc/model/changefeed.go
```diff
@@ -44,6 +45,19 @@ const (
 	StateRemoved FeedState = "removed"
 )
 
+const (
+	// ErrorHistoryGCInterval represents how long we keep error record in changefeed info
+	ErrorHistoryGCInterval = time.Minute * 10
```
overvenus, line 50: Looks like them can be unexported.

PR #720, opened 2020-07-06
puller: fix the encode bytes in span and add a fail-fast

### LINKED ISSUE #684: The listening range of the kvclient may become larger as regions are merged and split.
## Bug Report

We found that the kvclient may have problems when the following steps happened:
1. listen range : [c,e)
2. scan region & listen region: regionCD:[c,d) , region DE:[d,e)
3. regionBC[b,c) and regionCD[c,d) are merged.
4. kvclient rescan [c,e), and start to listion regionBD [b,d)
5. regionBD split into regionBC[b,c) and regionCD[c,d)
6. kvclient listion regionBC[b,c), regionCD[c,d) and region DE[d,e) now

At this point, the range of kvclient listening has been extended.
In extreme cases, the range of kvclient listening may be expanded indefinitely.

leoppro:
fixed #684

- fix the relevant logic for memcomparable key and span
- add a fail-fast for span outputted by kvclient

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/puller/puller.go
+++ b/cdc/puller/puller.go
@@ -50,7 +50,7 @@ type pullerImpl struct {
 	pdCli        pd.Client
 	kvStorage    tikv.Storage
 	checkpointTs uint64
-	spans        []regionspan.Span
+	spans        []regionspan.ComparableSpan
 	buffer       *memBuffer
 	outputCh     chan *model.RawKVEntry
 	tsTracker    frontier.Frontier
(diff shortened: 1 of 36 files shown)


pkg/regionspan/span.go
```diff
@@ -180,3 +174,30 @@ func Intersect(lhs Span, rhs Span) (span Span, err error) {
...
+	return false
+}
+
+// ComparableSpan returns a memcomparable span
+func ComparableSpan(span Span) Span {
+	return Span{
+		Start: codec.EncodeBytes(nil, span.Start),
+		End:   codec.EncodeBytes(nil, span.End),
+	}
+}
```
overvenus, line 198: In order to avoid confusion, what about define a new type span which encodes key in memcomparable format?

cdc/puller/puller.go
```diff
@@ -109,6 +108,7 @@ func (p *pullerImpl) Run(ctx context.Context) error {
 		span := span
 
 		g.Go(func() error {
+			log.Info("start event feed", zap.Reflect("span", span))
```
amyangfei, line 111: seems this message has been logged in kv/client.go, bug in debug levelo

leoppro:
PTAL again @overvenus @amyangfei

amyangfei:
Please fix the unit test

PR #732, opened 2020-07-08
Refine ddl execute log

amyangfei:

In some cases if the DDL is ignored by block list or cyclic replication disabled SyncDDL, the following log will be confused

```
[2020/07/08 16:35:52.721 +08:00] [INFO] [mysql.go:210] ["DDL event ignored"] [sql="create table trule4 (id int primary key auto_increment)"]
[2020/07/08 16:35:52.721 +08:00] [INFO] [changefeed.go:677] ["Execute DDL succeeded"]...
```

Add a result returned by `EmitDDLEvent`, including whether the DDL is ignored.

### Check List 

Tests 

 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/sink/black_hole.go
+++ b/cdc/sink/black_hole.go
@@ -68,9 +68,9 @@ func (b *blackHoleSink) EmitCheckpointTs(ctx context.Context, ts uint64) error {
 	return nil
 }
 
-func (b *blackHoleSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) (EmitEventResult, error) {
+func (b *blackHoleSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
 	log.Info("BlockHoleSink: DDL Event", zap.Any("ddl", ddl))
-	return EmitEventResult{}, nil
+	return nil
 }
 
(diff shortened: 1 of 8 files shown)


cdc/sink/black_hole.go
```diff
@@ -68,9 +68,9 @@ func (b *blackHoleSink) EmitCheckpointTs(ctx context.Context, ts uint64) error {
 	return nil
 }
 
-func (b *blackHoleSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
+func (b *blackHoleSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) (EmitEventResult, error) {
```
leoppro, line 71: We can avoid changing the Sink interface by using a special error to indicate that the DDL is ignored

amyangfei: Use special error is quite a hack way. Changing the export interface is not always a bad thing, but I was considering whether to put the result to each `EmitXXXEvent`, but this may decrease performance

leoppro: Ignoring can also be expressed as an error, I don't think it's hack

amyangfei: addressed

PR #735, opened 2020-07-09
remove incorrect kafka parameter

amyangfei:

Since the values of `Producer.Flush.Frequency`, `Producer.Flush.Bytes` and `Producer.Flush.Messages` are 0, sarama always flushes message as-fast-as-possible 
https://github.com/Shopify/sarama/blob/5933302b8bc4b51c43f1bc2531146d59dc88a91d/produce_set.go#L258-L259
In most cases `Producer.Flush.MaxMessages` will be not used. Besides we can control message size via `Producer.MaxMessageBytes` (https://github.com/Shopify/sarama/blob/master/produce_set.go#L230-L250)

Remove this config

### Check List 

Tests 

 - Integration test

### Release note

- No release note


leoppro:
LGTM

PR #741, opened 2020-07-13
Refine some metrics, remove name from TableReplicaInfo

amyangfei:

Fix https://github.com/pingcap/ticdc/issues/740

- add metric to record processor resolved ts gap and checkpoint ts gap
- query table name from `SchemaStorage` when `addTable`

Screen Shot 2020-07-14 at 10 52 11

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


cdc/processor.go
```diff
@@ -363,7 +366,11 @@ func (p *processor) positionWorker(ctx context.Context) error {
 			}
 
 			p.position.ResolvedTs = minResolvedTs
-			resolvedTsGauge.Set(float64(oracle.ExtractPhysical(minResolvedTs)))
+			phyTs := oracle.ExtractPhysical(minResolvedTs)
+			resolvedTsGauge.Set(float64(phyTs))
+			// It is more accurate to get tso from PD, but in most cases we have
+			// deployed NTP service, a little bias is acceptable here.
+			metricResolvedTsLagGauge.Set(float64(oracle.GetPhysical(time.Now())-phyTs) / 1e3)
```
leoppro, line 373: how about get a ts from PD?

amyangfei: As the comment in L371-L372, it is a trade-off. Get `tso` from PD may fail, timeout or increase the load of PD (although this is not too frequent).

PR #742, opened 2020-07-14
*: support ssl mysql sink

overvenus:

Support sending changes to MySQL sink with SSL enabled.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note

* Support sending changes to MySQL sink with SSL enabled.


cdc/sink/sink.go
```diff
@@ -79,7 +65,7 @@ func NewSink(ctx context.Context, sinkURIStr string, filter *filter.Filter, conf
 	case "blackhole":
 		return newBlackHoleSink(opts), nil
 	case "mysql", "tidb":
```
leoppro, line 67:
```suggestion
	case "mysql", "tidb","mysql+ssl", "tidb+ssl":
```
see: https://github.com/pingcap/ticdc/pull/734/files#diff-d6e2759c88c65cddfd8d333dae3e5b94R85

amyangfei:
Seems some integration tests are always failed

tests/_utils/run_sql
```diff
@@ -2,11 +2,17 @@
...
 # parameter 2: database host
 # parameter 3: database port
+# parameter 4: other mysql client settings
 
 set -e
 
-host=${2:-127.0.0.1}
-port=${3:-4000}
+sql=${1}
+shift
```
amyangfei, line 10: The failure test case may be caused by the shift operation.
some `run_sql sql_statement` doesn't provide `host` and `port`

PR #750, opened 2020-07-15
Add check to sort dir when creating changefeed

amyangfei:

Fix a bug reported by QA. If file sort is enabled and given `SortDir` is not a dir or writable, the changefeed will run with error

Check given `SortDir` is a directory and writable

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


cmd/client_changefeed.go
```diff
@@ -243,6 +243,13 @@ func verifyChangefeedParamers(ctx context.Context, cmd *cobra.Command, isCreate
 		return nil, errors.Annotate(err, "can not load timezone, Please specify the time zone through environment variable `TZ` or command line parameters `--tz`")
 	}
 
+	if sortEngine == string(model.SortInFile) {
+		err = util.IsDirAndWritable(sortDir)
+		if err != nil {
+			return nil, err
+		}
+	}
```
overvenus, line 251: I think it should be checked on CDC server.

amyangfei: addressed, check on cdc server is more suitable.

Issue #697, opened 2020-06-28
Further refinement of the canal protocol


leoppro:
## Feature Request

**Is your feature request related to a problem? Please describe:**

The current canal protocol implementation only supports the replication of row-level changelogs. We would like to support a more complete canal protocol to improve its universality.

**Describe the feature you'd like:**

- [ ] support more field in canal protocol (such as `Column.mysqlType`)
- [ ] support transaction in canal protocol (`TransactionBegin` and `TransactionEnd` message)

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

qinggniq:
@amyangfei  Hello, I have read ticdc code roughly, here is my understand:
1. ticdc/model is the middleware struct between tikv row data change and downstream component protocol.
2. ticdc/sink/codec is the translator for translating from model format to canal format.

and what I should do is:
1. https://github.com/pingcap/ticdc/blob/master/cdc/sink/codec/canal.go#L127, in canalEntryBuilder.buildColumn, I should add the `mysqlType` field to the return value of canal.Column.
2. at https://github.com/pingcap/ticdc/blob/master/cdc/model/sink.go#L110, it seems in ticdc/sink/codec there is no translator  for translating Txn struct to  canal.Entry(canal.TransactionBegin/canal.TransactionEnd/canal.RowData), so I should implement it.
3. To implement demo consumer, it seems that I should add a new file to prepare tikv row change data, and then execute cdc/entry/codec to change it to cdc/model/* struct, and use cdc/sink/codec to change it to canal struct, to check the correction, or just need to add some code at https://github.com/pingcap/ticdc/blob/master/cdc/model/sink.go#L110 ?

Is there anything wrong in my understand ?

liuzix:
2. You should probably add transaction-related logic to FlushRowChangedEvents, which is what the mysql sink does.

3. What you are talking about is Unit Testing. The overall correctness is best tested by Integration tests, which requires a canal client. You should probably read about canal clients on [Canal Wiki](https://github.com/alibaba/canal/wiki).

qinggniq:
@liuzix  what confused me is that: 
2. At https://github.com/pingcap/ticdc/blob/master/cdc/sink/mysql.go#L624, there is a `s.db.BeginTx(ctx, nil)`, so I think it will insert a `begin` sql to sql server, and then seems no need to add transactionBegin or End log.
[...]

PR #757, opened 2020-07-16
Write msyqlType to canal packets and add special handling for Blob & Text field types.

qinggniq:
1. #697 add mysqlType to canal column entry.
2. Fix text/blob type issue.

1. Add `Charset` field to `mode.Column` to help figure out real mysqlType.
2. Now in cdc `TEXT` and `BLOB` mysql types are both mapped to `JavaSQLTypeBLOB`, but in [jdbc](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-type-conversions.html), `TEXT` is mapped to `JavaSQLTypeVARCHAR`, `BLOB` is mapped to `JavaSQLTypeBLOB`, so canal consumer (alibaba canal consumer/adapter) will use different logic to handle `TEXT` and `BLOB` type values, but the encoding logic for `TEXT` and `BLOB` in cdc is same, there might be something wrong. So we should distinguish them. 

### Check List 

Tests 

 - Unit test

Code changes


cdc/entry/mounter.go
```diff
@@ -373,8 +373,9 @@ func (m *mounterImpl) mountRowKVEntry(tableInfo *TableInfo, row *rowKVEntry) (*m
 			return nil, errors.Trace(err)
 		}
 		col := &model.Column{
-			Type:  colInfo.Tp,
-			Value: value,
+			Type:    colInfo.Tp,
+			Value:   value,
+			Charset: colInfo.Charset,
```
liuzix, line 378: Please use a flag to indicate whether it is binary data or not.

cdc/model/sink.go
```diff
@@ -87,6 +88,7 @@ type RowChangedEvent struct {
 type Column struct {
 	Type        byte        `json:"t"`
 	WhereHandle *bool       `json:"h,omitempty"`
+	Charset     string      `json:"c"`
```
liuzix, line 91: Probably should use int32 or int64 for the sake of extensibility. We may need to store other flags in the future.

cdc/sink/codec/canal.go
```diff
@@ -126,6 +129,7 @@ func (b *canalEntryBuilder) buildHeader(commitTs uint64, schema string, table st
 // build the Column in the canal RowData
 func (b *canalEntryBuilder) buildColumn(c *model.Column, colName string, updated bool) (*canal.Column, error) {
 	sqlType := MysqlToJavaType(c.Type)
+	mysqlType := parser_types.TypeToStr(c.Type, c.Charset)
```
liuzix, line 132: You only need to determine whether Charset is 'binary' or 'utf8'.

cdc/sink/codec/canal.go
```diff
@@ -163,12 +173,18 @@ func (b *canalEntryBuilder) buildColumn(c *model.Column, colName string, updated
 		case string:
 			value = v
 		case []byte:
-			decoded, err := b.bytesDecoder.Bytes(v)
-			if err != nil {
-				return nil, errors.Trace(err)
+			// special handle for text and blob
+			// see https://github.com/alibaba/canal/blob/9f6021cf36f78cc8ac853dcf37a1769f359b868b/parse/src/main/java/com/alibaba/otter/canal/parse/inbound/mysql/dbsync/LogEventConvert.java#L801
+			if strings.Contains(mysqlType, "text") {
```
liuzix, line 178: Don't need to test it again. Just use sqlType.

cdc/sink/codec/canal.go
qinggniq: You means that we don't need mysqlType in canal column now?

cdc/model/sink.go
qinggniq: use `0x0` for binary charset,`0x01` for others?

cdc/sink/codec/canal.go
```diff
@@ -148,12 +155,12 @@ func (b *canalEntryBuilder) buildColumn(c *model.Column, colName string, updated
 	}
 	switch sqlType {
 	case JavaSQLTypeBINARY, JavaSQLTypeVARBINARY, JavaSQLTypeLONGVARBINARY:
-		if strings.Contains(mysqlType, "text") {
+		if c.Flag == model.BinaryFlag {
```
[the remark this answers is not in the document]
qinggniq: ok

cdc/entry/mounter.go
```diff
@@ -358,6 +358,13 @@ func (m *mounterImpl) mountRowKVEntry(tableInfo *TableInfo, row *rowKVEntry) (*m
 	if !row.Delete {
 		datumsNum = len(tableInfo.Columns)
 	}
+
+	charsetFlag := func(charset string) uint32 {
```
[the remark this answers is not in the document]
qinggniq: No, just there are duplicate codes.

cdc/model/sink.go
```diff
@@ -85,9 +92,10 @@ type RowChangedEvent struct {
 
 // Column represents a column value in row changed event
 type Column struct {
-	Type        byte        `json:"t"`
-	WhereHandle *bool       `json:"h,omitempty"`
-	Value       interface{} `json:"v"`
+	Type        byte         `json:"t"`
+	WhereHandle *bool        `json:"h,omitempty"`
+	Flag        bitflag.Flag `json:"f"`
```
leoppro, line 97: could you add the unit test for this bitflag?

qinggniq: OK, I will implement it at pkg/util/bitflag.go, and use pingcap check to write some uint tests.

leoppro: and I found that the `bitflag.Flag` is alias of `byte`, so the `bitflag.Flag` can store up to 8 bits of flags. It seems that 8 bits are not enough, could we using `uint64` to store the flags?

qinggniq: Sure!

cdc/model/sink.go
```diff
@@ -34,6 +37,29 @@ const (
 	MqMessageTypeResolved
 )
 
+// ColumnFlagType is for encapsulating the flag operations for different flags.
+type ColumnFlagType util.Flag
+
+const (
+	// BinaryFlag means col charset is binary
+	BinaryFlag ColumnFlagType = 1 << 0
```
liuzix, line 45: Please use iota

cdc/model/sink.go
```diff
@@ -34,6 +37,29 @@ const (
...
+)
+
+//SetIsBinary set BinaryFlag
+func (b *ColumnFlagType) SetIsBinary() {
+	(*util.Flag)(b).Add(util.Flag(BinaryFlag))
+}
+
+//UnSetIsBinary unset BinaryFlag
+func (b *ColumnFlagType) UnSetIsBinary() {
```
liuzix, line 54:
```suggestion
func (b *ColumnFlagType) UnsetIsBinary() {
```

cdc/model/sink.go
```diff
@@ -34,6 +37,29 @@ const (
...
+	(*util.Flag)(b).Remove(util.Flag(BinaryFlag))
+}
+
+//IsSetBinary show whether BinaryFlag is set
+func (b *ColumnFlagType) IsSetBinary() bool {
```
liuzix, line 59:
```suggestion
func (b *ColumnFlagType) IsBinary() bool {
```

cdc/sink/codec/canal.go
```diff
@@ -181,13 +205,15 @@ func (b *canalEntryBuilder) buildColumn(c *model.Column, colName string, updated
...
 		Value:         value,
+		MysqlType:     mysqlType,
 	}
 	return canalColumn, nil
 }
 
 // build the RowData of a canal entry
 func (b *canalEntryBuilder) buildRowData(e *model.RowChangedEvent) (*canal.RowData, error) {
 	var columns []*canal.Column
+
```
liuzix, line 216:
```suggestion
```

cdc/sink/codec/canal_test.go
```diff
@@ -147,7 +149,7 @@ func (s *canalEntrySuite) TestConvertEntry(c *check.C) {
 		},
 		Delete: true,
 		Columns: map[string]*model.Column{
-			"id": {Type: mysql.TypeLong, WhereHandle: &trueVar, Value: 1},
+			"id": {Type: mysql.TypeLong, WhereHandle: &trueVar, Value: 1, Flag: model.BinaryFlag},
```
liuzix, line 152: Why would you want to set the BinaryFlag on a long type?

qinggniq: I just obey the information from `timodel.ColumnInfo.FieldType.Charset`, all type have their charset, and for `bigint` it shows the `Charset : binary` in my `cdc.log`.

qinggniq: Sorry, it's my illusion.

cdc/entry/mounter.go
```diff
@@ -405,9 +411,14 @@ func (m *mounterImpl) mountRowKVEntry(tableInfo *TableInfo, row *rowKVEntry) (*m
 		for _, col := range tableInfo.Columns {
 			_, ok := values[col.Name.O]
 			if !ok && tableInfo.IsColWritable(col) {
+				var flag model.ColumnFlagType
+				if col.Charset == "binary" {
+					flag.SetIsBinary()
+				}
```
amyangfei, line 417: recommend encapsulating this logic into a function like
```golang
func transColumnFlag(col timodel.ColumnInfo) model.ColumnFlagType {
    // some logic
}
```
we can reuse it

qinggniq: OK.

amyangfei:
Good job!

PR #758, opened 2020-07-16
Fix Avro timestamp & duration

liuzix:

Fix replicating timestamp and duration types through Kafka in Avro format.

- Fix discrepancy between Goavro and Registry.
- Re-parse duration before sending it to sink.

### Check List 

Tests 

Related changes

### Release note
- No release note

### DIFF
--- a/cdc/sink/codec/avro.go
+++ b/cdc/sink/codec/avro.go
@@ -347,7 +347,7 @@ func columnToAvroNativeData(col *model.Column) (interface{}, string, error) {
 		fsp := len(frac)
 		fracInt, err := strconv.ParseInt(frac, 10, 32)
 		if err != nil {
-			return "", "", err
+			return nil, "", err
 		}
 		fracInt = int64(float64(fracInt) * math.Pow10(6-fsp))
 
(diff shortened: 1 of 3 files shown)


cdc/sink/codec/avro.go
```diff
@@ -259,9 +267,15 @@ func getAvroDataTypeNameMysql(tp byte) (string, error) {
 	case mysql.TypeVarchar, mysql.TypeString, mysql.TypeVarString:
 		return "string", nil
 	case mysql.TypeDate, mysql.TypeDatetime, mysql.TypeTimestamp:
-		return "long", nil
+		return avroLogicalType{
+			Type:        "long",
+			LogicalType: "timestamp-millis",
```
amyangfei, line 272: Recommend to define `type LogicalType string` and use const variables like
`TimestampMillis LogicalType = "timestamp-millis"`

cdc/sink/codec/avro.go
```diff
@@ -290,31 +304,55 @@ func columnToAvroNativeData(col *model.Column) (interface{}, string, error) {
...
+			if err != nil {
+				return nil, "", err
+			}
+		}
+
+		fsp := len(frac)
+		fracInt, err := strconv.ParseInt(frac, 10, 32)
 		if err != nil {
-			return nil, "error", err
+			return "", "", err
```
amyangfei, line 350:
```suggestion
			return nil, "", err
```

PR #763, opened 2020-07-17
Support stop changefeed after checkpoint reaches target ts

amyangfei:

Fix bug in https://github.com/pingcap/ticdc/issues/745

- Add a periodic health checker in owner, currently check whether the checkpoint-ts of each changefeed reaches target-ts, if reaches, dispatch an admin job.
- Add a new admin job type, `AdminFinish`
- Add target ts verification before chagnefeed is created.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Fix the bug that resources are not released when checkpoint-ts of a changefeed reaches target-ts


cdc/owner.go
```diff
@@ -612,10 +612,30 @@ func (o *Owner) collectChangefeedInfo(ctx context.Context, cid model.ChangeFeedI
...
+		feedState = model.StateFinished
 	}
 	return nil, status, feedState, nil
 }
 
+func (o *Owner) checkClusterHealth(_ context.Context) error {
+	// check whether a changefeed has finished by comparing checkpoint-ts and target-ts
+	for _, cf := range o.changeFeeds {
+		if cf.status.CheckpointTs == cf.info.GetTargetTs() {
+			log.Info("changefeed replication finished", zap.String("changefeed", cf.id), zap.Uint64("ts", cf.status.CheckpointTs))
```
overvenus, line 625:
```suggestion
			log.Info("changefeed replication finished", zap.String("changefeed", cf.id), zap.Uint64("checkpointts", cf.status.CheckpointTs))
```

leoppro: ping @amyangfei

amyangfei: addressed

amyangfei:
PTAL @overvenus

PR #770, opened 2020-07-20
Refactor: Let the encoder control sink's output related operation

liuzix:
It gives encoder writers the flexibility to control if and how to write to the MQ after each encoder operation.
It further decouples the encoders from the mqSink.

Let encoder return an **EncoderResult** which dictates whether and how (sync/async) the sink will output to the MQ.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

### Release note

- This is an experimental change. We should discuss it in detail before merging.


cdc/sink/mq.go
```diff
@@ -322,37 +319,62 @@ func (k *mqSink) runWorker(ctx context.Context, partition int32) error {
...
+		op, err := encoder.AppendRowChangedEvent(e.row)
 		if err != nil {
 			return errors.Trace(err)
 		}
 		batchSize++
 		// This is only a temporary fix so that the Avro encoder does not overflow.
 		// Pending further refactoring
-		if encoder.Size() >= batchSizeLimit || (k.protocol == codec.ProtocolAvro && encoder.Size() >= 1) {
-			if err := flushToProducer(); err != nil {
+		if encoder.Size() >= batchSizeLimit || op == codec.EncoderNeedSyncWrite || op == codec.EncoderNeedAsyncWrite {
```
amyangfei, line 345: Currently, op is 
- `codec.EncoderNeedAsyncWrite` if encoder is `AvroEncoder`
- `codec.EncoderNoOperation` if encoder is other encoder

So when `encoder.Size() >= batchSizeLimit` is true and encoder is not `AvroEncoder`, the `flushToProducer` will output a warning log and does nothing. The real `send message` will have to wait for flush ticker.

cdc/sink/mq.go
```diff
@@ -187,20 +186,16 @@ flushLoop:
...
 	encoder := k.newEncoder()
-	err := encoder.AppendResolvedEvent(ts)
+	op, err := encoder.AppendResolvedEvent(ts)
 	if err != nil {
 		return errors.Trace(err)
 	}
+	if op == codec.EncoderNoOperation {
+		return nil
+	}
 	key, value := encoder.Build()
```
leoppro, line 197: should we skip to write when no message in builder?

leoppro: ignore me

Issue #743, opened 2020-07-14
Some data may be replicated to downstream if table becomes ineligible after DDL


amyangfei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

1. setup a cdc cluster, and start a changefeed
2. in upstream, execute `create table t2 (id int primary key);insert into t2 values (1),(2);`, everything is normal
3. in upstream, execute `alter table t2 drop primary key; insert into t2 values (3),(4);`

2. What did you expect to see?

The replication of table `t2` will be paused and no more data is replicated

3. What did you see instead?

```
t> select * from t2;
+----+
| id |
+----+
| 1  |
| 2  |
| 3  |
| 4  |
+----+
4 rows in set

> show create table t2;
+-------+-------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                |
+-------+-------------------------------------------------------------------------------------------------------------+
| t2    | CREATE TABLE `t2` (\n  `id` int(11) NOT NULL,\n  PRIMARY KEY (`id`)\n) ENGINE=InnoDB DEFAULT CHARSET=latin1 |
+-------+-------------------------------------------------------------------------------------------------------------+
1 row in set
```

4. Versions of the cluster

    - Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

        ```
       Release Version: v4.0.0
        ```

    - TiCDC version (execute `cdc version`):

        ```
        Release Version: v4.0.2-32-ga3d9c7e
        Git Commit Hash: a3d9c7e5cc35ef57afa338f74edeea80ae4dbd59
        Git Branch: master
        UTC Build Time: 2020-07-14 08:00:04
        Go Version: go version go1.14.4 linux/amd64
        ```

PR #773, opened 2020-07-20
mounter: skip the DML of ineligible table

leoppro:

skip the DML of ineligible table

fix the part 2 of #743
### Check List 

Tests 

 - Unit test


leoppro:
> Please add tests for bugfix.

the test is implemented in UTF cases
https://github.com/pingcap/automated-tests/pull/420/files#diff-3af30509c8adc498f016b4d35f74ab44R90

PR #775, opened 2020-07-20
ci: extend the timeout of changefeed_pause_resume test

leoppro:
extend the timeout of changefeed_pause_resume test

### Check List 

Tests 

 - Unit test
 - Integration test


amyangfei:
This test case will be optimized after https://github.com/pingcap/ticdc/pull/760 merged

PR #778, opened 2020-07-20
*: refine the struct about row and ddl event

leoppro:
refine the struct about row and ddl event

### Check List 

Tests 

 - Unit test
 - Integration test


YuJuncen:
/lgtm

for @liuzix

PR #783, opened 2020-07-21
test: integration test for add and drop index

leoppro:
integration test for add and drop index

### Check List 

Tests 

 - Unit test
 - Integration test


tests/multi_source/main.go
```diff
@@ -165,14 +167,35 @@ func truncateDDL(ctx context.Context, db *sql.DB) {
...
 		if err == nil {
-			success++
-			if success%100 == 0 {
-				log.S().Info(id, " success: ", success)
+			insertSuccess++
+			if insertSuccess%100 == 0 {
+				log.S().Info(id, " insert success: ", insertSuccess)
+			}
+		}
+		if err != nil && !strings.HasPrefix(err.Error(), "Error 1146:") {
```
amyangfei, line 182: In which case will `Error 1146:` happen?

leoppro: the `Error 1146:` will happen when the schema or table is dropped in case `createDropSchemaDDL`.

leoppro:
@liuzix  PTAL

lance6716:
LGTM

PR #786, opened 2020-07-22
Fix docker-compose.yml and add support to start local Avro testing environment

liuzix:
Fix local testing facilities, especially for testing Kafka & Avro.

- Fixed bugs in docker-compose.yml
- Added scripts/avro-local-test.sh for the convenient of developers and potential users.
- **tableNameToSchemaSubject now concatenates db name and table name by "_", so as to avoid compatibility problems with older versions of Confluent Schema Registry**.

### Check List 

Tests 
 - Manual test

### Release note

- Testing facilities are changed.

### DIFF
--- a/cdc/sink/codec/schema_registry.go
+++ b/cdc/sink/codec/schema_registry.go
@@ -315,5 +315,5 @@ func httpRetry(ctx context.Context, credential *security.Credential, r *http.Req
 
 func (m *AvroSchemaManager) tableNameToSchemaSubject(tableName model.TableName) string {
 	// We should guarantee unique names for subjects
-	return tableName.Schema + "." + tableName.Table + m.subjectSuffix
+	return tableName.Schema + "_" + tableName.Table + m.subjectSuffix
 }
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,7 @@
           - ./docker/logs:/logs
         entrypoint: "/cdc server"
         command:
+          - --addr=0.0.0.0:8300
           - --pd=http://upstream-pd:2379
           - --log-file=/logs/capturer0.log
         depends_on:
@@ -44,6 +45,7 @@
           - ./docker/logs:/logs
         entrypoint: "/cdc server"
         command:
+          - --addr=0.0.0.0:8300
           - --pd=http://upstream-pd:2379
           - --log-file=/logs/capturer1.log
         depends_on:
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -44,6 +45,7 @@
           - ./docker/logs:/logs
         entrypoint: "/cdc server"
         command:
+          - --addr=0.0.0.0:8300
```
amyangfei, line 48: what about adding --advertise-address, the same as docker-compose.yml

overvenus:
/lgtm

👍

amyangfei:
> /lgtm
> 
> 👍

The bot doesn't recognize the `/lgtm`, would you please type it agian

cdc/sink/codec/schema_registry.go
```diff
@@ -315,5 +315,5 @@ func httpRetry(ctx context.Context, credential *security.Credential, r *http.Req
 
 func (m *AvroSchemaManager) tableNameToSchemaSubject(tableName model.TableName) string {
 	// We should guarantee unique names for subjects
-	return tableName.Schema + "." + tableName.Table + m.subjectSuffix
+	return tableName.Schema + "_" + tableName.Table + m.subjectSuffix
```
leoppro, line 318:
```
schema : t1_t2 table ：t3;
schema ：t1 table：t2_t3;
```
some bad thing happened

amyangfei: https://github.com/pingcap/ticdc/pull/687#discussion_r446751076
This problem has been discussed before, but the quote is removed since some version.
Is it  `` ` `` a valid char in scheme registry? @liuzix

leoppro: @liuzix

liuzix: It is a valid character in Schema Registry 5.5.1, but in tests with older versions I have encountered problems. And it is certainly **not** a valid character for a Kafka topic, but Kafka connect expects the Schema subjects to be like ${kafka topic}-key or ${kafka topic}-value. The expected format of the Schema subjects can be modified in Kafka connect's config files, but I personally feels that requiring all users to do this will increase their cognitive burden. 
Moreover, in Schema Registry 5.5.1, the JDBC sink connector **might** not care about what the Schema subjects are, as it does not try to register new schemas and the schema ID in the wire format is orthogonal to the Subjects. But I don't think we should rely on this type of behavior.
In conclusion, for an experimental version avoiding any special characters is the least problematic way of handling this matter. We should, however, provide table aliasing mechanisms for Kafka sink so that users who could encounter problems due to special db & table names can manually alias the problematic names.

Issue #728, opened 2020-07-08
the events are dispatched to different partitions unexpectedly


leoppro:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

There is a table in upstream TiDB:
```
CREATE TABLE `mars_task_verify_statistics` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `resource_id` varchar(24) NOT NULL,
  `task_id` bigint(20) NOT NULL,
  `source` varchar(36) NOT NULL,
  `category` varchar(26) DEFAULT NULL,
  `priority` smallint(6) DEFAULT 0,
  `type` smallint(6) NOT NULL,
  `task_time` timestamp NOT NULL,
  `start_time` timestamp NULL DEFAULT NULL,
  `end_time` timestamp NULL DEFAULT NULL,
  `reason` varchar(256) DEFAULT NULL,
  PRIMARY KEY (`id`),
  KEY `idx_resource_id` (`resource_id`),
  UNIQUE KEY `u_source_task` (`source`,`task_id`),
  KEY `idx_task_time` (`task_time`),
  KEY `idx_end_time` (`end_time`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
```

the dispatcher of this table is `rowid`, we received the following message from Kafka:

```
[Thread-0][partition-0]
Total Bytes: 159
[CanalEntry]:
header {
  version: 1
  serverenCode: "UTF-8"
  executeTime: 1593749327652
  sourceType: MYSQL
  schemaName: "mars"
  tableName: "mars_task_verify_statistics"
  eventType: DELETE
  props {
    key: "rowsCount"
    value: "1"
  }
}
entryType: ROWDATA
storeValue: "\020\003P\000bB\n$\020\f\032\006source \0010\000B\024user_info_after_pass\n\032\020\003\032\atask_id \0010\000B\t229209952"

[RowChange]:
eventType: DELETE
isDdl: false
rowDatas {
  beforeColumns {
    sqlType: 12
    name: "source"
    isKey: true
    isNull: false
    value: "user_info_after_pass"
  }
  beforeColumns {
    sqlType: 3
    name: "task_id"
    isKey: true
    isNull: false
    value: "229209952"
  }
}

[Thread-0][partition-28]
Total Bytes: 113
[CanalEntry]:
header {
  version: 1
[...]

PR #787, opened 2020-07-22
*: chose one index to output delete events

leoppro:

fix the #728 

- we will select one unique index(including primary key) as the handle index for every table
- the delete events are generated based on the handle index
- the column of handle index will be marked in column flag
- if a unique index includes a virtual generated column, we regard the index as invalid.
- the stored generated columns are captured now and it will be also marked in column flag
- provides a new dispatcher named `index-value`
- the `rwoid` dispatcher is instead by `index-value`
- the `where_handle` field in Column is deprecation

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- select one unique index as the handle index to generate the delete events


amyangfei:
Please add a release note

leoppro:
> Please add a release note

done

cdc/entry/mounter.go
```diff
@@ -369,7 +367,7 @@ func (m *mounterImpl) mountRowKVEntry(tableInfo *TableInfo, row *rowKVEntry) (*m
 		if !exist {
 			return nil, errors.NotFoundf("column info, colID: %d", index)
 		}
-		if !row.Delete && !tableInfo.IsColWritable(colInfo) {
+		if !tableInfo.IsColCDCVisible(colInfo) {
```
amyangfei, line 370: What is `!row.Delete` here used for originally?

cdc/entry/schema_storage.go
```diff
@@ -207,7 +207,15 @@ type TableInfo struct {
...
 	uniqueColumns    map[int64]struct{}
-	handleColID      int64
+
+	// only for new row format decoder
+	handleColID int64
+
+	// the mounter will chose this index to output delete events
+	// special value:
+	// -1 : pk is handle
+	// -2 : the table is not eligible
```
amyangfei, line 217: Can add two constants for these two variables

cdc/entry/schema_storage.go
```diff
@@ -267,10 +277,39 @@ func WrapTableInfo(schemaID int64, schemaName string, version uint64, info *timo
...
+			if len(ti.Indices[handleIndexOffset].Columns) > len(ti.Indices[i].Columns) ||
+				(len(ti.Indices[handleIndexOffset].Columns) == len(ti.Indices[i].Columns) &&
+					ti.Indices[handleIndexOffset].ID > ti.Indices[i].ID) {
+				handleIndexOffset = i
+			}
+		}
+	}
+	if handleIndexOffset >= 0 {
+		ti.HandleIndexID = ti.Indices[handleIndexOffset].ID
+	}
```
amyangfei, line 310: Is it better to add fatal log here, if handle index is not found, we can record more information here.

cdc/sink/dispatcher/table.go
```diff
@@ -14,27 +14,18 @@
...
-		log.Fatal("calculate hash of message key failed, please report a bug", zap.Error(err))
-	}
-	_, err = hash.Write([]byte(row.Table.Table))
-	if err != nil {
-		log.Fatal("calculate hash of message key failed, please report a bug", zap.Error(err))
-	}
-	return int32(hash.Sum32() % uint32(t.partitionNum))
+	h.Write([]byte(row.Table.Schema), []byte(row.Table.Table))
+	h ^= h<<4 | h>>4
+	return int32(byte(h) % byte(t.partitionNum))
```
amyangfei, line 30: Why the hash algorithm is changed here, will this cause compatibility problems as tables will be dispatched to different partitions after cdc server upgrade

cdc/entry/mounter.go
leoppro: The original logic:
if the type of row event here is deletion, the table must be `pk-is-handle`, then we need to output the delete event via rowid, regardless of whether the column is writable or not.

The current logic:
if the table is `pk-is-handle` and the pk column is a virtual generated column, then we won't choose the pk index as handle index;
if the table is `pk-is-handle` and the pk column is a stored generated column, we will output the delete event via rowid, 
so we needn't to judge if the event is delete here

cdc/sink/dispatcher/table.go
leoppro: because the position of column and column value is not stable in columns map, we need a position inertia hash algorithm instead of the crc32

cdc/entry/schema_storage.go
leoppro: I check it here:
https://github.com/pingcap/ticdc/pull/787/files#diff-8cae4fdc1c5f0a4e944025f65917425fR437

cdc/entry/schema_storage.go
```diff
@@ -267,10 +277,39 @@ func WrapTableInfo(schemaID int64, schemaName string, version uint64, info *timo
...
+			continue
+		}
+		if idx.Primary {
+			handleIndexOffset = i
+			break
+		}
+		if handleIndexOffset < 0 {
+			handleIndexOffset = i
+		} else {
+			if len(ti.Indices[handleIndexOffset].Columns) > len(ti.Indices[i].Columns) ||
```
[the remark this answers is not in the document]
leoppro: Yes, We only promise that the selection of handle indexes is stable between two DDLs

leoppro:
@amyangfei PTAL again

cdc/entry/schema_storage.go
amyangfei: But the check in mounter can't log enough debug information

cdc/sink/dispatcher/table.go
amyangfei: This is a table-based dispatcher, the column value is not used

amyangfei:

amyangfei:
/run-kafka-tests tidb=release-4.0 tikv=release-4.0 pd=release-4.0

cdc/model/sink.go
```diff
@@ -109,12 +144,58 @@ type RowChangedEvent struct {
 
 // Column represents a column value in row changed event
 type Column struct {
-	Type        byte           `json:"t"`
+	Type byte `json:"t"`
+	// WhereHandle is deprecation
+	// WhereHandle is instead by HandleKey in Flag
```
liuzix, line 149:
```suggestion
	// WhereHandle is replaced by HandleKey in Flag
```

cdc/model/sink.go
```diff
@@ -109,12 +144,58 @@ type RowChangedEvent struct {
...
 	WhereHandle *bool          `json:"h,omitempty"`
 	Flag        ColumnFlagType `json:"f"`
 	Value       interface{}    `json:"v"`
 }
 
+// ColumnValueString returns a string representation of the column value
```
liuzix, line 155:
```suggestion
// ColumnValueString returns the string representation of the column value
```

cdc/entry/mounter.go
```diff
@@ -480,11 +478,43 @@ func (m *mounterImpl) mountRowKVEntry(tableInfo *model.TableInfo, row *rowKVEntr
 	}, nil
 }
 
+func setHandleKeyFlag(tableInfo *model.TableInfo, colValues map[string]*model.Column) error {
+	switch tableInfo.HandleIndexID {
+	case model.HandleIndexTableIneligible:
+		log.Fatal("this table is not a eligible", zap.Int64("tableID", tableInfo.ID))
```
liuzix, line 484:
```suggestion
		log.Fatal("this table is not eligible", zap.Int64("tableID", tableInfo.ID))
```

cdc/entry/mounter.go
```diff
@@ -480,11 +478,43 @@ func (m *mounterImpl) mountRowKVEntry(tableInfo *model.TableInfo, row *rowKVEntr
...
+	}
+	return nil
+}
+
 func (m *mounterImpl) mountIndexKVEntry(tableInfo *model.TableInfo, idx *indexKVEntry) (*model.RowChangedEvent, error) {
 	// skip set index KV
 	if !idx.Delete || m.enableOldValue {
 		return nil, nil
 	}
+	// skip the index which not handle
```
liuzix, line 514:
```suggestion
	// skip any index that is not the handle
```

cdc/model/schema_storage.go
```diff
@@ -21,6 +21,13 @@ import (
 	"github.com/pingcap/tidb/util/rowcodec"
 )
 
+const (
+	// HandleIndexPKIsHandle represents the handle index is pk and the pk is handle
```
liuzix, line 25:
```suggestion
	// HandleIndexPKIsHandle represents that the handle index is the pk and the pk is the handle
```

cdc/model/schema_storage.go
```diff
@@ -21,6 +21,13 @@ import (
...
+	HandleIndexPKIsHandle = -1
+	// HandleIndexTableIneligible represents the table is ineligible
```
liuzix, line 27:
```suggestion
	// HandleIndexTableIneligible represents that the table is ineligible
```

cdc/model/schema_storage.go
```diff
@@ -30,7 +37,15 @@ type TableInfo struct {
 	columnsOffset    map[int64]int
 	indicesOffset    map[int64]int
 	uniqueColumns    map[int64]struct{}
-	handleColID      int64
+
+	// only for new row format decoder
+	handleColID int64
+
+	// the mounter will chose this index to output delete events
```
liuzix, line 44:
```suggestion
	// the mounter will choose this index to output delete events
```

cdc/model/schema_storage.go
```diff
@@ -121,9 +167,13 @@ func (ti *TableInfo) GetRowColInfos() (int64, []rowcodec.ColInfo) {
 	return ti.handleColID, ti.rowColInfos
 }
 
-// IsColWritable returns is the col is writeable
-func (ti *TableInfo) IsColWritable(col *model.ColumnInfo) bool {
-	return col.State == model.StatePublic && !col.IsGenerated()
+// IsColCDCVisible returns is the col is visible for CDC
```
liuzix, line 170:
```suggestion
// IsColCDCVisible returns whether the col is visible for CDC
```

cdc/model/sink.go
```diff
@@ -40,8 +41,12 @@ const (
 type ColumnFlagType util.Flag
 
 const (
-	// BinaryFlag means col charset is binary
+	// BinaryFlag means the column charset is binary
 	BinaryFlag ColumnFlagType = 1 << ColumnFlagType(iota)
+	// HandleKeyFlag means the column selected as the handle key
```
liuzix, line 46:
```suggestion
	// HandleKeyFlag means the column is selected as the handle key
```

leoppro:
> The rest lgtm

thanks a lot

PR #788, opened 2020-07-23
Make canal consumer can process message as a set of dmls from transaction.

qinggniq:
#697 

1. alibaba consumer can only consume row change events at the granularity of `canal.message`, it will treat a `canal.message` as a set of sql queries from one transaction, and execute them as a transaction. So if we want cdc canal mqSink support transaction, we should collect the sql queries from one transaction, and add them to one `canal.message`, then send it to kafka, at last the canal consumer will  execute it as a transaction.
2. In current implementation of cdc, this situation will happen:
    - Executing at upstream TiDB `ddl, ddl, dml` sql queries
    - cdc call `emitDDLEvent, emitRowChangeEvent, emitDDLEvent`
    - mqSink will send 2nd `ddl` slower than the 3nd `dml`
    so canal consumer will receive `ddl, dml, ddl`, then throw error. Why `mysqlSink` works well is that `mysqlSink` will divide received rowChangeEvents from `resolvedTxns` and `unresolvedTxns` when called `flushRowChangeEvent`, and not execute `dml` which is greater than the `resolvedTs`. But the mqSink doesn't have this guarantee, it will send all `message` to kafka when called `flushRowChangeEvent`.
    I migrate the implementation at mysqlSink, and make mqSink can send `message` to kafka in order.

This version is basically compatible with alibaba canal consumer, some data type like `year, bit(n)` is not support due to canal adapter implementation. 

### Check List 
Code changes

Side effects

 - Unit test


liuzix:
I propose delaying review of this PR until necessary changes in #795 are made.

tests/_utils/run_cdc_server
```diff
@@ -77,9 +77,21 @@ done
...
 pid=$(ps -C run_cdc_server -o pid=|tr -d '[:space:]')
-# Uncomment to turn on grpc versbose log.
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+pid=$(ps -e | grep -m1 run_cdc_server | awk '{print $1}'|tr -d '[:space:]')
+else
+    echo "unkonwn system" && false
+fi
+
 # GRPC_GO_LOG_VERBOSITY_LEVEL=99 GRPC_GO_LOG_SEVERITY_LEVEL=debug \
+<<<<<<< HEAD
```
liuzix, line 90: Please clean up conflict notations

cdc/sink/codec/json.go
```diff
@@ -411,6 +411,15 @@ func (d *JSONEventBatchEncoder) EncodeDDLEvent(e *model.DDLEvent) (*MQMessage, e
 	return ret, nil
 }
 
+// UpdateResolvedTs implements the EventBatchEncoder interface
```
liuzix, line 414: Remove

cdc/sink/codec/canal.go
```diff
@@ -327,53 +336,113 @@ func (d *CanalEventBatchEncoder) EncodeCheckpointEvent(ts uint64) (*MQMessage, e
...
+	}
+	return messages
+}
+
+// MixedBuild implements the EventBatchEncoder interface
+func (d *CanalEventBatchEncoder) MixedBuild(withVersion bool) []byte {
+	panic("Mixed Build only use for JsonEncoder")
+}
+
+//Size implements the EventBatchEncoder interface
```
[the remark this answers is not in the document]
qinggniq: I need some time to consider a good solution for this problem.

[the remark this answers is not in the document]
qinggniq: In previous implementation, Size() can not be determined until events are marshaled. Now events are stored in txnCache, we can not know size until calling Build() which marshals events this time. If we want to know the size before calling Build(), we must marsh the events when call Size(), which means that we might marsh twice, once for Size(), once for Build().

cdc/sink/common/common.go
```diff
@@ -14,21 +14,16 @@
...
-	"github.com/pingcap/log"
 	"github.com/pingcap/ticdc/cdc/model"
-	"github.com/pingcap/ticdc/pkg/filter"
-	"go.uber.org/zap"
 )
 
-// UnresolvedTxnCache caches unresolved txns
+// UnresolvedTxnCache caches unresolved txns, not thread safe
 type UnresolvedTxnCache struct {
-	unresolvedTxnsMu sync.Mutex
```
liuzix, line 29: Why did you remove this?

qinggniq: I don't think UnresolvedTxnCache should be thread safe, it's duty should only be txn generator.

scripts/canal/docker/start.sh
```diff
@@ -0,0 +1,72 @@
...
+outerAdapterKey: mysql1
+concurrent: true
+dbMapping:
+  mirrorDb: true
+  database: ${DB_NAME}
+EOF
+
+bash ./bin/startup.sh
+
+while true; do
```
liuzix, line 70: Why this?

tests/README.md
```diff
@@ -19,7 +19,8 @@
 2. The following programs must be installed:
 
     * `mysql` (the MySQL cli client, currently [mysql client 8.0 is not supported](https://github.com/pingcap/tidb/issues/14021))
-
+    * `bash` # version >= 4.0.0
+    * `zookeepr & kafka` (if you want to run kafka test)
```
liuzix, line 23:
```suggestion
    * `zookeeper & kafka` (if you want to run kafka test)
```

tests/_utils/run_cdc_server
```diff
@@ -77,8 +77,15 @@ done
 
 echo "[$(date)] <<<<<< START cdc server in $TEST_NAME case >>>>>>"
 cd $workdir
+
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
 pid=$(ps -C run_cdc_server -o pid=|tr -d '[:space:]')
-# Uncomment to turn on grpc versbose log.
```
liuzix, line 81: Why delete this?

tests/_utils/start_tidb_cluster_impl
```diff
@@ -221,7 +221,9 @@ run_sql "select tidb_version();set @@global.tidb_enable_clustered_index=0;" ${DO
 sleep 2
 
 run_sql "update mysql.tidb set variable_value='60m' where variable_name='tikv_gc_life_time';" ${UP_TIDB_HOST} ${UP_TIDB_PORT}
+run_sql "update mysql.tidb set variable_value='false' where variable_name='tikv_gc_enable';" ${UP_TIDB_HOST} ${UP_TIDB_PORT}
```
liuzix, line 224: Why this?

scripts/canal/docker/start.sh
qinggniq: @liuzix Canal adapter is running at background, exit shell script will cause container exit. I use this to prevent shell script from exit.

qinggniq:
@liuzix Close. I will decompose this pr into three part.

PR #793, opened 2020-07-24
file-sorter: fix fake resolved ts event breaks CRTs restraint

amyangfei:

Fix bug found in https://github.com/pingcap/ticdc/issues/754

Use `CRTs-1` as fake resolved ts, where `CRTs` is commit ts of a real event. The root cause is sorted events could have the same commit ts, use `CRTs-1` as resolved ts is safe and eliminates this problem.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/puller/file_sorter.go
+++ b/cdc/puller/file_sorter.go
@@ -415,7 +415,7 @@ func (fs *FileSorter) rotate(ctx context.Context, resolvedTs uint64) error {
 			// file sorter outputs all events in this rotate round.
 			rowCount += 1
 			if rowCount%defaultAutoResolvedRows == 0 {
-				fs.output(ctx, model.NewResolvedPolymorphicEvent(item.entry.CRTs))
+				fs.output(ctx, model.NewResolvedPolymorphicEvent(item.entry.CRTs-1))
 			}
 		} else {
 			lastSortedFileUpdated = true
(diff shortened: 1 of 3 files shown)


cdc/puller/file_sorter.go
```diff
@@ -415,7 +415,7 @@ func (fs *FileSorter) rotate(ctx context.Context, resolvedTs uint64) error {
 			// file sorter outputs all events in this rotate round.
 			rowCount += 1
 			if rowCount%defaultAutoResolvedRows == 0 {
-				fs.output(ctx, model.NewResolvedPolymorphicEvent(item.entry.CRTs))
+				fs.output(ctx, model.NewResolvedPolymorphicEvent(item.entry.CRTs-1))
```
leoppro, line 418: what happened if the current `item.entry.CRTs-1` is equal to the `item.entry.CRTs` of the previous event

amyangfei: Events are sorted by commit-ts here, and the previous is guaranteed to be sent to output chan

overvenus, line 418: Please add comments about `-1`.

amyangfei: addressed

PR #794, opened 2020-07-27
make filter support filter column types

qinggniq:
There are some column types at cdc which are not compatible with alibaba consumer, such as `year, bit`, we can use column typs in ddl and dml to filter the queries.

Add `ignoreColumnType` at `config.filter`.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

### DIFF
--- a/cdc/model/sink.go
+++ b/cdc/model/sink.go
@@ -107,6 +107,15 @@ type RowChangedEvent struct {
 	Keys         []string           `json:"keys"`
 }
 
+//GetColumnTypes get column types from dml query
+func (e *RowChangedEvent) GetColumnTypes() []byte {
+	res := make([]byte, 0, len(e.Columns))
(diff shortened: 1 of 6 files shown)


cdc/model/sink.go
```diff
@@ -107,6 +107,15 @@ type RowChangedEvent struct {
 	Keys         []string           `json:"keys"`
 }
 
+//GetColumnTypes get column types from dml query
+func (e *RowChangedEvent) GetColumnTypes() []byte {
+	res := make([]byte, 0, len(e.Columns))
```
qinggniq, line 112: Maybe there are many columns with same type, will cause many duplicate elements
1. return `map[byte]struct{}` 
2. use `map[byte]struct{}`  filter type first, then convert it to `[]byte`.

PR #796, opened 2020-07-27
mounter: support more column flag

leoppro:
blocked by #787 

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/model/sink.go
```diff
@@ -43,6 +43,14 @@ type ColumnFlagType util.Flag
 const (
 	// BinaryFlag means col charset is binary
 	BinaryFlag ColumnFlagType = 1 << ColumnFlagType(iota)
```
overvenus, line 45:
```suggestion
	BinaryFlag ColumnFlagType = uint64(1) << ColumnFlagType(iota)
```

PR #797, opened 2020-07-27
owner: includes stopped changefeed when calculating gc safepoint

amyangfei:

Fix bug found in https://github.com/pingcap/ticdc/issues/791 & https://github.com/pingcap/ticdc/issues/810

- Introduce a `stoppedFeeds` fields in owner
- checkpoint of stopped changefeed is also taken into consideration when calculating safepoint in owner
- clear gc safepoint if no running and stopped changefeed

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Fix the bug that GC safepoint will be forwarded beyond the checkpoint of stopped changefeed.


amyangfei:

PR #799, opened 2020-07-28
mounter: support to output pre table info in ddl events

leoppro:
support to output pre table info in ddl events

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/entry/schema_storage.go
```diff
@@ -59,6 +56,48 @@ func (s *SingleSchemaSnapshot) HandleDDL(job *timodel.Job) error {
...
+	case timodel.ActionRenameTable, timodel.ActionDropTable, timodel.ActionDropView, timodel.ActionTruncateTable:
+		// get the table will be dropped
+		table, ok := s.TableByID(job.TableID)
+		if !ok {
+			return nil, errors.NotFoundf("table %d", job.TableID)
+		}
+		return table, nil
+	case timodel.ActionTruncateTablePartition, timodel.ActionAddTablePartition, timodel.ActionDropTablePartition:
+		tableID := job.BinlogInfo.TableInfo.ID
+		table, ok := s.TableByID(tableID)
```
amyangfei, line 76: Is it sure to return the table that the partition belongs to, if so, seems the logic is similar to the default branch

cdc/entry/schema_storage.go
```diff
@@ -59,6 +56,48 @@ func (s *SingleSchemaSnapshot) HandleDDL(job *timodel.Job) error {
 	return s.handleDDL(job)
 }
 
+// PreTableInfo returns the table info which with overwrite by the specified job
```
liuzix, line 59:
```suggestion
// PreTableInfo returns the table info which will be overwritten by the specified job
```

cdc/model/schema_storage.go
```diff
@@ -0,0 +1,195 @@
...
+	return uniqueKeys
+}
+
+// IsColumnUnique returns whether the column is unique
+func (ti *TableInfo) IsColumnUnique(colID int64) bool {
+	_, exist := ti.uniqueColumns[colID]
+	return exist
+}
+
+// ExistTableUniqueColumn returns whether the table has the unique column
```
liuzix, line 163:
```suggestion
// ExistTableUniqueColumn returns whether the table has a unique column
```

leoppro:
@liuzix @amyangfei PTAL again

Issue #790, opened 2020-07-24
Optimize get tso frequency


amyangfei:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently TiCDC does some TSO query from PD, mainly having two points

- query `CurrentVersion` when resolving lock

```
runtime/debug.Stack(0xc0000b4000, 0x22b7840, 0x469ad30)
        runtime/debug/stack.go:24 +0x9d
github.com/pingcap/tidb/store/tikv/oracle/oracles.(*pdOracle).getTimestamp(0xc0002b3a40, 0x2c35e80, 0xc0000b4000, 0x0, 0x0, 0x437500)
        github.com/pingcap/tidb@v1.1.0-beta.0.20200603101356-552e7709de0d/store/tikv/oracle/oracles/pd.go:109 +0x139
github.com/pingcap/tidb/store/tikv/oracle/oracles.(*pdOracle).GetTimestamp(0xc0002b3a40, 0x2c35e80, 0xc0000b4000, 0x0, 0x4469dc, 0xc00085b9d8)
        github.com/pingcap/tidb@v1.1.0-beta.0.20200603101356-552e7709de0d/store/tikv/oracle/oracles/pd.go:71 +0x43
github.com/pingcap/tidb/store/tikv.(*tikvStore).getTimestampWithRetry(0xc00069f700, 0xc004882ba0, 0x0, 0x0, 0x0)
        github.com/pingcap/tidb@v1.1.0-beta.0.20200603101356-552e7709de0d/store/tikv/kv.go:347 +0xd3
github.com/pingcap/tidb/store/tikv.(*tikvStore).CurrentVersion(0xc00069f700, 0x1176ad857f, 0x466bd40, 0x7ea7b4d)
        github.com/pingcap/tidb@v1.1.0-beta.0.20200603101356-552e7709de0d/store/tikv/kv.go:332 +0xc7
github.com/pingcap/ticdc/cdc/kv.(*eventFeedSession).singleEventFeed(0xc000398600, 0x2c35e40, 0xc0009e70c0, 0x4, 0xc00030f9e0, 0x2d, 0x2d, 0xc00030fa10, 0x2d, 0x2d, ...)
        github.com/pingcap/ticdc/cdc/kv/client.go:1105 +0x35b9
[...]

PR #801, opened 2020-07-28
perf: optimize get tso frequency

liuzix:
Querying CurrentVersion too frequently when resolving lock. #790 

CurrentVersion is now cached and is updated regularly by a separate goroutine.

### Check List 

Tests 

Code changes

### Release note

- No release note

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -1102,7 +1102,7 @@ func (s *eventFeedSession) singleEventFeed(
 				log.Warn("region not receiving event from tikv for too long time",
 					zap.Uint64("regionID", regionID), zap.Reflect("span", span), zap.Duration("duration", sinceLastEvent))
 			}
-			version, err := s.kvStorage.CurrentVersion()
+			version, err := s.kvStorage.(*StorageWithCurVersionCache).GetCachedCurrentVersion()
 			if err != nil {
 				log.Warn("failed to get current version from PD", zap.Error(err))
 				continue
--- a/cdc/kv/store_op.go
+++ b/cdc/kv/store_op.go
@@ -15,8 +15,13 @@ package kv
 
 import (
 	"fmt"
+	"go.uber.org/zap"
+	"sync"
+	"time"
 
 	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
+	"github.com/pingcap/ticdc/cdc/model"
 	"github.com/pingcap/ticdc/pkg/flags"
 	"github.com/pingcap/ticdc/pkg/security"
 	tidbconfig "github.com/pingcap/tidb/config"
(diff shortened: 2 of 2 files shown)


cdc/kv/client.go
```diff
@@ -251,6 +252,8 @@ type CDCClient struct {
 
 	regionCache *tikv.RegionCache
 	kvStorage   tikv.Storage
+
+	currentVersionCache tidbkv.Version
```
leoppro, line 256:
```suggestion
	currentVersionCache model.Ts
```

cdc/kv/client.go
```diff
@@ -270,7 +273,26 @@ func NewCDCClient(ctx context.Context, pd pd.Client, kvStorage tikv.Storage, cre
...
 	}
+
+	go (func() {
+		ticker := time.NewTicker(tsoQueryInterval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
```
leoppro, line 286: if there is no region to resolve lock, we still query the tso from pd every two seconds?

cdc/kv/client.go
```diff
@@ -286,6 +308,14 @@ func (c *CDCClient) Close() error {
 	return nil
 }
 
+// GetCurrentVersion gets a cached version that is regularly updated by a background goroutine
+func (c *CDCClient) GetCurrentVersion() tidbkv.Version {
+	for atomic.LoadUint64(&c.currentVersionCache.Ver) == 0 {
+		time.Sleep(1 * time.Second)
```
leoppro, line 314: It's not graceful

cdc/kv/client.go
```diff
@@ -286,6 +308,14 @@ func (c *CDCClient) Close() error {
 	return nil
 }
 
+// GetCurrentVersion gets a cached version that is regularly updated by a background goroutine
+func (c *CDCClient) GetCurrentVersion() tidbkv.Version {
```
overvenus, line 312: How many CDCClient would CDC create? How about caching the TS in the processor?

cdc/puller/puller.go
```diff
@@ -59,6 +61,101 @@ type pullerImpl struct {
...
+				// the cached value has been updated once
+				ver = atomic.LoadUint64(&s.currentVersionCache)
+				break loop
+			}
+		}
+	}
+	return tidbkv.Version{Ver: ver}, nil
+}
+
+func (s *kvStorageWithVersionCache) Close() error {
```
amyangfei, line 136: Where is the `Close` called

cdc/puller/puller.go
```diff
@@ -15,6 +15,7 @@ package puller
 
 import (
 	"context"
+	"github.com/pingcap/ticdc/pkg/notify"
```
amyangfei, line 18: Please move this import to the group of third-party lib

cdc/puller/puller.go
```diff
@@ -69,6 +166,7 @@ func NewPuller(
 	spans []regionspan.Span,
 	limitter *BlurResourceLimitter,
 ) Puller {
+	kvStorage = newKvStorageWithVersionCache(kvStorage)
```
leoppro, line 169: `KvStorageWithVersionCache` should be globally unique, not one per puller.

PR #804, opened 2020-07-28
Add TicdcEventDataReader to parse ticdc data which consumed from kafka.

dixingxing0:
Add java demo to parse ticdc data which consumed from kafka.
This pr is just for review design, it's not the final version.


java-demo/src/main/java/org/pingcap/ticdc/cdc/TicdcEventFilter.java
```diff
@@ -0,0 +1,42 @@
...
+    private Map<String, Long> tableMaxTSMap = new HashMap<>();
+    private Map<Integer, Long> resolveMaxTSMap = new HashMap<>();
+
+    public boolean filter(long rowChangeTS, int partition, String tableName) {
+        String mapKey = tableName + "_" + partition;
+        Long rowChangeMaxTS = tableMaxTSMap.get(mapKey);
+        if (rowChangeMaxTS == null) {
+            tableMaxTSMap.put(mapKey, rowChangeMaxTS);
+        } else {
+            if (rowChangeTS < rowChangeMaxTS) {
```
amyangfei, line 24:
```suggestion
            if (rowChangeTS <= rowChangeMaxTS) {
```
Is it also a duplicated event If the `rowChangeTS` equals to `rowChangeMaxTS`

java-demo/src/main/java/org/pingcap/ticdc/cdc/TicdcEventFilter.java
```diff
@@ -0,0 +1,42 @@
...
+                return false;
+            }
+            tableMaxTSMap.put(mapKey, rowChangeTS);
+        }
+
+        Long resolveMaxTS = resolveMaxTSMap.get(partition);
+        if (rowChangeMaxTS != null) {
+            if (rowChangeTS < resolveMaxTS) {
```
amyangfei, line 32:
```suggestion
            if (rowChangeTS <= resolveMaxTS) {
```
The same as the comparison of `rowChangeMaxTS`

java-demo/src/main/java/org/pingcap/ticdc/cdc/KafkaMessage.java
```diff
@@ -0,0 +1,57 @@
+package org.pingcap.ticdc.cdc;
```
liuzix, line 1:
```suggestion
package com.pingcap.ticdc.cdc;
```
Probably should use our real domain name?

dixingxing0: OK, I will fix it.

dixingxing0:
> Besides, what about moving the demo dir to `demo/java`?
> Rest lgtm

PR #808, opened 2020-07-29
Support auto create sort dir in file sort mode, fix owner data race

amyangfei:

When sort dir does not exist, try to create it recursively

- check os.IsNotExist
- add a lock to owner access in server

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/http_handler.go
+++ b/cdc/http_handler.go
@@ -65,6 +65,8 @@ func (s *Server) handleResignOwner(w http.ResponseWriter, req *http.Request) {
 		writeError(w, http.StatusBadRequest, errors.New("this api only supports POST method"))
 		return
 	}
+	s.ownerLock.RLock()
+	defer s.ownerLock.RUnlock()
 	if s.owner == nil {
 		handleOwnerResp(w, concurrency.ErrElectionNoLeader)
 		return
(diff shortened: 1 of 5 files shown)


amyangfei:
/run-integration-tests tidb=release-4.0 tikv=release-4.0 pd=release-4.0

cdc/owner.go
```diff
@@ -204,7 +205,14 @@ func (o *Owner) newChangeFeed(
 	if info.Engine == model.SortInFile {
 		err = util.IsDirAndWritable(info.SortDir)
 		if err != nil {
```
leoppro, line 207: Could we try to create folders no matter what errors occur, such as no permissions, strange io errors, etc.?

cdc/http_handler.go
```diff
@@ -65,8 +65,10 @@ func (s *Server) handleResignOwner(w http.ResponseWriter, req *http.Request) {
 		writeError(w, http.StatusBadRequest, errors.New("this api only supports POST method"))
 		return
 	}
+	s.ownerLock.RLock()
 	if s.owner == nil {
 		handleOwnerResp(w, concurrency.ErrElectionNoLeader)
```
leoppro, line 70: why not put the operations of `lock\unlock` and checking nil into handleOwnerResp function

leoppro: bad suggest, ignore me

cdc/owner.go
amyangfei: addressed

amyangfei:
CI failure is caused by https://github.com/pingcap/ticdc/pull/816

PR #809, opened 2020-07-29
Upgrade tidb related go modules to latest

amyangfei:

Upgrade TiDB related dependencies to fix the compatibility problem. ~DNM until release-4.0 has merged current master, because after this PR merged into master, we will maintain the `release-4.0` and `master` separately.~

Will remove the replace of `tidb-tools` after https://github.com/pingcap/tidb-tools/pull/378 merged

update go mod and related code update.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

 - **No need** to cherry-pick to the release branch, but the release branch **should upgrade related dependencies** to the latest release-4.0

### Release note

- No release note


cdc/entry/mounter.go
```diff
@@ -488,7 +488,7 @@ func formatColVal(datum types.Datum, tp byte) (interface{}, error) {
 		return datum.GetMysqlDuration().String(), nil
 	case mysql.TypeJSON:
 		return datum.GetMysqlJSON().String(), nil
-	case mysql.TypeNewDecimal, mysql.TypeDecimal:
```
leoppro, line 491: should we report a fatal when meet a `mysql.TypeUnspecified `

go.mod
```diff
@@ -45,3 +45,5 @@ require (
 	google.golang.org/genproto v0.0.0-20200113173426-e1de0a7b01eb // indirect
 	google.golang.org/grpc v1.26.0
 )
+
+replace github.com/pingcap/tidb-tools => github.com/amyangfei/tidb-tools v3.0.0-rc.2.0.20200729132015-0def89dd24bb+incompatible
```
overvenus, line 49: tidb-tools PR has merged.

cdc/entry/mounter.go
amyangfei: the default branch returns `datum.GetValue()`, I think it is ok if type is unspecified

leoppro:
lgtm

leoppro:
@amyangfei  can we merge this pr now?

PR #814, opened 2020-07-31
update go mod to fix const value compatibilty in parser

amyangfei:

pingcap/parser#947 changes some const value of online DDL state, and pingcap/parser#952 reverts it. So we should  keep parser and tidb dependencies up to date.

the replace of `tidb-tools` will be removed after https://github.com/pingcap/tidb-tools/pull/379 merged

update go mod

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


leoppro:
pingcap/tidb-tools#379 was merged

PR #822, opened 2020-08-05
fix the bug about adding table

leoppro:
fix the bug about adding table. this bug is was introduced by #771 .

add some fail-fast check

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/processor.go
+++ b/cdc/processor.go
@@ -810,14 +810,6 @@ func createSchemaStorage(pdEndpoints []string, credential *security.Credential, 
 }
 
 func (p *processor) addTable(ctx context.Context, tableID int64, replicaInfo *model.TableReplicaInfo) {
-	globalResolvedTs := atomic.LoadUint64(&p.sinkEmittedResolvedTs)
-	/*
-		if replicaInfo.StartTs < globalResolvedTs {
-			log.Fatal("cannot add table which of startTs is less than the global resolved, please report a bug",
-				zap.Int64("tableID", tableID), zap.Any("replicaInfo", replicaInfo),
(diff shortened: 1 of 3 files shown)


cdc/processor.go
```diff
@@ -508,6 +508,10 @@ func (p *processor) updateInfo(ctx context.Context) error {
 			if err != nil {
 				return false, backoff.Permanent(errors.Trace(err))
 			}
+			err = updatePosition()
+			if err != nil {
+				return false, errors.Trace(err)
```
leoppro, line 513:
```suggestion
				return true, errors.Trace(err)
```

cdc/processor.go
```diff
@@ -806,6 +810,14 @@ func createSchemaStorage(pdEndpoints []string, credential *security.Credential,
...
 
 func (p *processor) addTable(ctx context.Context, tableID int64, replicaInfo *model.TableReplicaInfo) {
+	globalResolvedTs := atomic.LoadUint64(&p.sinkEmittedResolvedTs)
+	/*
+		if replicaInfo.StartTs < globalResolvedTs {
+			log.Fatal("cannot add table which of startTs is less than the global resolved, please report a bug",
+				zap.Int64("tableID", tableID), zap.Any("replicaInfo", replicaInfo),
+				zap.Uint64("globalResolvedTs", globalResolvedTs))
+		}
+	*/
```
overvenus, line 820: How about removing '/*' and '*/'?

leoppro: There's a bug that's causing this place to fatal. I'm working on it.

leoppro: I will fix it in next PR

Issue #768, opened 2020-07-20
Add a new sink to external storage


3pointer:
## Feature Request

**Is your feature request related to a problem? Please describe:**
https://github.com/pingcap/br/issues/325 need TiCDC has the ability to record KV changes to external storage like S3 and Localfile

**Describe the feature you'd like:**
Add new sink to external storage. and according to some users' requirements, S3 is needed. so we should implement new sink to S3 first.

**Describe alternatives you've considered:**

1. Consider Rollback DML, we need old value feature.
2. Define specify message proto, for example, we can define bucket or file name as `DB`.`Table`, the detail content may looks like

|  start_ts  | commit_ts  | event_type  | redo KV change(SQL) | undo KV change | 
|  ----  | ----  |  ----  | ----  | ---- |
| t1  | t2 | DML  | add `key1` `value1` ( insert into `table1`) | remove `key1` |
| t3  | t4 | DDL  | alter table `table1` add column | not support|

3. To keep consistency, only event `commit_ts` < `resolve_ts` should be record.

**Teachability, Documentation, Adoption, Migration Strategy:**

## Workload estimation(person-day)

* Design: 2
* Develop: 4
* QA: 3
* Document: 3

IANTHEREAL:
duplicate with https://github.com/pingcap/ticdc/issues/655

DanielZhangQD:
@3pointer Justo to confirm some info:

1. No plan for gcs now, right?
2. This only supports from v5.0 or will it be cherry-picked to v4.0?

3pointer:
> @3pointer Justo to confirm some info:
> 
> 1. No plan for gcs now, right?
> 2. This only supports from v5.0 or will it be cherry-picked to v4.0?

1. No, Currently we only support s3 and local shared file system.
2. It will be merge into release-4.0, but it will GA in 5.0

PR #826, opened 2020-08-05
Support write TiCDC log to S3.

3pointer:

part of #768, support incremental log and write log data to s3 directly.
 
1.  Define cdclog with three kind of events.
    - row changed event: translate to protobuf then write data into `cdclog`
    - ddl event: translate to protobuf then write data into `ddlog`
    - resolved ts event: remain global resolved ts to `log.meta`

2.  Since s3 doesn't allow append-like operation. there are three different treatments for different events
     - row changed event: flush data to cdclog, if flushed data size is less than 5M. directly write to s3 with putObject request. if flushed data size is greater than 5M, use multi-upload request to upload data as one chunk in s3. if total upload chunks size is greater than 100M, then complete and rotate it.  use the last row event's commit ts as the suffix of each cdclog name.
    - ddl event: consider ddl is not frequent in most cases. when ddllog size less than 10M, I choose to read content, append new ddl event then write back to ddllog.
    - resolved ts event: update in every EmitCheckpoinTs inplace.

### Check List 

Tests 

屏幕快照 2020-08-06 下午2 04 55
屏幕快照 2020-08-06 下午2 05 04

Side effects

 - Increased code complexity

Related changes

 - Need to update the `tidb-cdc/cdc-ansible`


cdc/model/sink.go
```diff
@@ -189,6 +191,11 @@ func (t *TableName) GetTable() string {
 	return t.Table
 }
 
+// GetTableID returns table name.
```
amyangfei, line 194:
```suggestion
// GetTableID returns table ID.
```

proto/LogProtocol.proto
```diff
@@ -0,0 +1,63 @@
+syntax = "proto3";
+option optimize_for = SPEED;
+
+enum ColumnFlagType {
```
amyangfei, line 4: We have more column flag now, ref: https://github.com/pingcap/ticdc/blob/e0dcfb707c531292111bf5a4d3a12e8d52e8c189/cdc/model/sink.go#L43-L58

proto/LogProtocol.proto
```diff
@@ -0,0 +1,63 @@
...
+	int64 tableID = 3;
+	int64 partition = 4;
+}
+
+message Column {
+	bytes type = 1;
+	bool whereHandle = 2;
+	oneof columnType_present{
+    	ColumnFlagType flag= 3;
+    }
```
amyangfei, line 20: We use a uint64 to store the column flag, and it can contain multiple ColumnFlag via `or` operation.
Besides the `whereHandle` can be removed now.

cdc/processor.go
```diff
@@ -1024,6 +1024,11 @@ func runProcessor(
 		cancel()
 		return nil, errors.Trace(err)
 	}
+	err = sink.Initialize(ctx, nil)
```
amyangfei, line 1027: Seems `initialize` does nothing in processor, is it better to remove it?

cdc/sink/cdclog/file.go
```diff
@@ -0,0 +1,330 @@
...
+			rowDatas := make([]byte, 0, flushedSize)
+			for event := int64(0); event < flushedEvents; event++ {
+				row := <-ts.dataCh
+				if event == flushedEvents-1 {
+					// the last event
+					fileName = makeTableFileName(row.CommitTs)
+				}
+				data, err := row.ToProtoBuf().Marshal()
+				if err != nil {
+					errCh <- err
```
amyangfei, line 95: Do we need to return when an error happens?

cdc/sink/cdclog/file.go
```diff
@@ -0,0 +1,330 @@
...
+				}
+				rowDatas = append(rowDatas, append(data, '\n')...)
```
amyangfei, line 97: is `\n` used as row delimiter, what happens when data contains `\n`

cdc/sink/cdclog/file.go
```diff
@@ -0,0 +1,330 @@
...
+			}
+
+			tableDir := path.Join(f.logPath.root, makeTableDirectoryName(ts.tableID))
```
amyangfei, line 100: It is better to use `path/filepath`

cdc/sink/cdclog/file.go
```diff
@@ -0,0 +1,330 @@
...
+		return err
+	}
+	_, err = file.Write(data)
+	return err
+}
+
+func (f *fileSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
+	switch ddl.Type {
+	case parsemodel.ActionCreateTable:
+		f.logMeta.Names[ddl.TableInfo.TableID] = makeSpecifyTableName(ddl.TableInfo.Schema, ddl.TableInfo.Table)
```
amyangfei, line 220: Where will `f.logMeta.Names` be used, is it only recorded in log meta file?
`makeSpecifyTableName` is not table unified, since `.` could be used in schema name and table name, recommend to use `QuoteSchema` in `pkg/quote/quotes.go`

cdc/sink/cdclog/s3.go
```diff
@@ -0,0 +1,599 @@
...
+		s.logMeta.Names[ddl.TableInfo.TableID] = makeSpecifyTableName(ddl.TableInfo.Schema, ddl.TableInfo.Table)
+	case parsemodel.ActionRenameTable:
+		delete(s.logMeta.Names, ddl.PreTableInfo.TableID)
+		s.logMeta.Names[ddl.TableInfo.TableID] = makeSpecifyTableName(ddl.TableInfo.Schema, ddl.TableInfo.Table)
+	}
+	data, err := ddl.ToProtoBuf().Marshal()
+	if err != nil {
+		return err
+	}
+	listing, err := s.ListObject(ddlEventsDir, 1)
```
amyangfei, line 482: `MaxKeys means the response of `ListObjectsV2Input` returns at most one key, do we need to query the latest file?

3pointer: According to [Document](https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysUsingAPIs.html),` List results are always returned in UTF-8 binary order.`

So I choose `max.Long - commitTs` as suffix of ddl log file. this guarantee the ListObject will return the latest file.

cdc/sink/cdclog/s3.go
```diff
@@ -0,0 +1,592 @@
...
+				shouldFlush = true
+			}
+			s.tableBuffers[hash].sendEvents.Inc()
+		}
+	}
+	if shouldFlush {
+		// should not block here
+		go func() {
+			s.notifyChan <- struct{}{}
+		}()
```
overvenus, line 401:
```suggestion
		select {
		case s.notifyChan <- struct{}{}:
		default:
		}
```

cdc/sink/cdclog/s3.go
```diff
@@ -0,0 +1,592 @@
...
+	}
+	return nil
+}
+
+func (s *s3Sink) flushTableBuffers(ctx context.Context) error {
+	var wg sync.WaitGroup
```
overvenus, line 407: What about `golang.org/x/sync/errgroup`?

cdc/sink/cdclog/s3.go
```diff
@@ -0,0 +1,592 @@
...
+		client:  s3client,
+		logMeta: newLogMeta(),
+
+		tableBuffers: tableBuffers,
+		notifyChan:   notifyChan,
+	}, nil
+}
+
+// checkBucket checks if a path exists
+var checkBucket = func(svc *s3.S3, bucket string) error {
```
overvenus, line 586: Prefer to declare a function.
```suggestion
func checkBucket(svc *s3.S3, bucket string) error {
```

proto/LogProtocol.proto
```diff
@@ -0,0 +1,57 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
3pointer: After discussed with @leoppro, both TiCDC Open protocol and protobuf has the same information. 
but for these reasons we are not choose TiCDC Open protocol.
1.  TiCDC Open protocol is key/value format, this format is more suitable for MQ.
2.  Protobuf is easy to control compatibility than Json.

cdc/entry/mounter.go
```diff
@@ -442,6 +442,7 @@ func (m *mounterImpl) mountRowKVEntry(tableInfo *model.TableInfo, row *rowKVEntr
 		Table: &model.TableName{
 			Schema:    schemaName,
 			Table:     tableName,
+			TableID:   row.PhysicalTableID,
 			Partition: partitionID,
```
leoppro, line 446: we can remove the `Partition` now

3pointer: How about remove the Partition in another PR?

leoppro: ok

cdc/sink/cdclog/file.go
```diff
@@ -0,0 +1,338 @@
...
+				return err
+			}
+
+			if stat.Size() > maxRowFileSize {
+				// rotate file
+				err := tsReplica.rowFile.Close()
+				if err != nil {
+					return err
+				}
+				file, err := os.OpenFile(filepath.Join(tableDir, fileName), os.O_CREATE|os.O_WRONLY|os.O_APPEND, defaultFileMode)
```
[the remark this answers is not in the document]
3pointer: No, when every batch flush, it will create new file name [here](https://github.com/pingcap/ticdc/pull/826/files/9f72844e79dcb885735803fd8a03e252333acaca#diff-2662504f1f959169dbcc1c8723c19361R93), but not every new file name will be used.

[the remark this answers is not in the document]
3pointer: eh... maybe we should use the first event' commit ts as the filename

PR #827, opened 2020-08-06
Use batch replace in mysql sink to increase performance

amyangfei:

Implement https://github.com/pingcap/ticdc/issues/807

- [x] Add a batchReplace option, if enabled, group replace SQLs by `` `schema`.`table` ``
- [x] Add unit test about map/reduce replace
- [x] Make the batchReplace option configurable
- [ ] Benchmark for this optimization
- [x] ~~Pass necessary information to sink to avoid column name sorting, has the following solutions, choose 3 now.~~ Column sequence is stable now
     1. pass `SchemaStorage` from processor
     2. pass table info in each RowChangedEvent
     3. pass a stable coumn name list in each RowChangedEvent

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Add a batch replace way to optimize download write throughput in MySQL sink.

### DIFF
--- a/cdc/sink/mysql.go
+++ b/cdc/sink/mysql.go
@@ -718,8 +718,6 @@ func (s *mysqlSink) prepareDMLs(rows []*model.RowChangedEvent, replicaID uint64,
 				replaces = make(map[string][][]interface{})
 			}
 			query, args = prepareDelete(quoteTable, row.PreColumns)
-			sqls = append(sqls, query)
-			values = append(values, args)
 			if query != "" {
 				sqls = append(sqls, query)
 				values = append(values, args)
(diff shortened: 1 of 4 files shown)


cdc/sink/mysql.go
```diff
@@ -52,12 +52,14 @@ import (
...
-	defaultDDLMaxRetryTime = 20
-	defaultTiDBTxnMode     = "optimistic"
-	defaultFlushInterval   = time.Millisecond * 50
+	defaultWorkerCount         = 16
+	defaultMaxTxnRow           = 256
+	defaultDMLMaxRetryTime     = 8
+	defaultDDLMaxRetryTime     = 20
+	defaultTiDBTxnMode         = "optimistic"
+	defaultFlushInterval       = time.Millisecond * 50
+	defaultBatchReplaceEnabled = true
```
amyangfei, line 61: The default value need to discuss and confirm

PR #829, opened 2020-08-10
filter: using tidb-tools.filter to filter the system schemas

leoppro:

using tidb-tools.filter to filter the system schemas

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/pkg/filter/filter.go
+++ b/pkg/filter/filter.go
@@ -14,12 +14,11 @@
 package filter
 
 import (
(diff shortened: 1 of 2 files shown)


leoppro:
> Please fix the unit test. the `metrics_schema` has changed name since [pingcap/tidb#14874](https://github.com/pingcap/tidb/pull/14874)

done

PR #830, opened 2020-08-10
sink & mounter: refine the struct of  row changed events

leoppro:
#### the new struct of row change events

```
// RowChangedEvent represents a row changed event
type RowChangedEvent struct {
	StartTs  uint64 `json:"start-ts"`
	CommitTs uint64 `json:"commit-ts"`

	RowID int64 `json:"row-id"`

	Table *TableName `json:"table"`

	// TODO: remove it
	Delete bool `json:"delete"`

	TableInfoVersion uint64 `json:"table-info-version,omitempty"`

	// TODO: remove it
	// if the table of this row only has one unique index(includes primary key),
	// IndieMarkCol will be set to the name of the unique index
	IndieMarkCol string `json:"indie-mark-col"`

	Columns    []*Column `json:"columns"`
	PreColumns []*Column `json:"pre-columns"`

	IndexColumns [][]int

	// TODO: remove it
	Keys []string `json:"keys"`
}
```

 + `Delete`, `IndieMarkCol`, `Keys` will be removed in next PR
 + To maintain column order stability, use slices to store `Columns` and `PreColumns`
     + the indices of the `Columns` are fixed. If a row event has only partial columns, the missing columns are filled using `nil`. for example, a complete event like :`RowChangedEvent{Columns:[col1,col2,col3]}`, in the event of deletion, cdc only gets column 3, like:`RowChangedEvent{Columns:[nil,nil,col3]}`.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/cdc/model/schema_storage.go
+++ b/cdc/model/schema_storage.go
@@ -274,6 +274,7 @@ func (ti *TableInfo) IsIndexUnique(indexInfo *model.IndexInfo, allowColumnNullab
 	return false
 }
 
+// IsHandleColumn returns whether the column is a handle column
 func (ti *TableInfo) IsHandleColumn(colInfo *model.ColumnInfo) bool {
 	switch ti.HandleIndexID {
 	case HandleIndexTableIneligible:
--- a/cdc/model/schema_storage.go
+++ b/cdc/model/schema_storage.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 
 	"github.com/pingcap/log"
+
 	"go.uber.org/zap"
 
 	"github.com/pingcap/parser/model"
@@ -137,8 +138,8 @@ func WrapTableInfo(schemaID int64, schemaName string, version uint64, info *mode
 		}
 	}
 	ti.findHandleIndex()
-	log.Debug("select the handle index", zap.Int64("tableID", ti.ID), zap.Int64("HandleIndexID", ti.HandleIndexID))
 	ti.initColumnsFlag()
+	log.Debug("warp table info", zap.Any("tableInfo", ti))
 	return ti
 }
 
--- a/cdc/model/schema_storage.go
+++ b/cdc/model/schema_storage.go
@@ -94,6 +94,9 @@ func WrapTableInfo(schemaID int64, schemaName string, version uint64, info *mode
 
 	for i, idx := range ti.Indices {
 		ti.indicesOffset[idx.ID] = i
+		if idx.Unique {
+
+		}
 		if ti.IsIndexUnique(idx) {
 			for _, col := range idx.Columns {
 				ti.uniqueColumns[ti.Columns[col.Offset].ID] = struct{}{}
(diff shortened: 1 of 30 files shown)


cdc/model/schema_storage.go
```diff
@@ -140,6 +172,59 @@ func (ti *TableInfo) findHandleIndex() {
...
+		if !mysql.HasNotNullFlag(colInfo.Flag) {
+			flag.SetIsNullable()
+		}
+		if mysql.HasMultipleKeyFlag(colInfo.Flag) {
+			flag.SetIsMultipleKey()
+		}
+		ti.ColumnsFlag[colInfo.ID] = flag
+	}
+
+	// In TiDB, Only the first column can be set multiple key,
```
liuzix, line 202: Suggested:
In TiDB, just as in MySql, only the first column of an index can be marked as "multiple key" or "unique key", and only the first column of a unique index may be marked as "unique key". See https://dev.mysql.com/doc/refman/5.7/en/show-columns.html.
Yet if an index has multiple columns, we would like to easily determine that all those columns are indexed, which is crucial for the completeness of the information we pass to the downstream. Therefore, instead of using the MySql standard, we made our own decision to mark all columns in an index with the appropriate flag(s).

leoppro:
@liuzix  PTAL again

cdc/model/schema_storage.go
```diff
@@ -88,12 +108,22 @@ func WrapTableInfo(schemaID int64, schemaName string, version uint64, info *mode
...
 	for i, idx := range ti.Indices {
 		ti.indicesOffset[idx.ID] = i
-		if ti.IsIndexUnique(idx) {
+		if ti.IsIndexUnique(idx, false) {
 			for _, col := range idx.Columns {
 				ti.uniqueColumns[ti.Columns[col.Offset].ID] = struct{}{}
 			}
 		}
-		if idx.Primary || idx.Unique {
+		if ti.IsIndexUnique(idx, true) {
```
amyangfei, line 116: suggest to use another function for the condition of `allowColumnNullable=true`

cdc/model/schema_storage.go
```diff
@@ -140,6 +172,61 @@ func (ti *TableInfo) findHandleIndex() {
...
+			}
+			if idxInfo.Unique {
+				flag.SetIsUniqueKey()
+			}
+			if len(idxInfo.Columns) > 1 {
+				flag.SetIsMultipleKey()
+			}
+			if idxInfo.ID == ti.HandleIndexID && ti.HandleIndexID >= 0 {
+				flag.SetIsHandleKey()
+			}
```
amyangfei, line 224: Seems some flags including `PrimaryKey`, `UniqueKey` and `IsHandleKey` have been set in L184-L195

leoppro: yes, some flags will be set twice, but the final result will not be affected.

cdc/model/schema_storage.go
leoppro: fine

PR #835, opened 2020-08-11
cmd: fix cmd TLS usage

overvenus:

Fix `changefeed list` and `changefeed query` do not work when TLS is enabled.

### Check List 

Tests 

 - Integration test

Related changes

### Release note

- Fix `changefeed list` and `changefeed query` do not work when TLS is enabled.

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"net/http"
 	"net/url"
 	"os"
 	"os/signal"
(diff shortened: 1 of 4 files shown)


cmd/util.go
```diff
@@ -133,13 +133,21 @@ func getOwnerCapture(ctx context.Context) (*capture, error) {
...
 
-func applyAdminChangefeed(ctx context.Context, job model.AdminJob) error {
+func applyAdminChangefeed(ctx context.Context, job model.AdminJob, credential *security.Credential) error {
 	owner, err := getOwnerCapture(ctx)
 	if err != nil {
 		return err
 	}
-	addr := fmt.Sprintf("http://%s/capture/owner/admin", owner.AdvertiseAddr)
-	resp, err := http.PostForm(addr, url.Values(map[string][]string{
+	scheme := "http"
```
liuzix, line 141: Need a new helper function for creating the Client?

overvenus: We have a helper function already, `httputil.NewClient`.

PR #840, opened 2020-08-12
fix unstable test in sink hang

amyangfei:

Fix case-3 in https://github.com/pingcap/ticdc/issues/838

The transactions are sent to two sink workers, we must ensure the sink worker with `MySQLSinkHangLongTime` injected is executed before the sink worker with `MySQLSinkExecDMLError` injected returns.

Add a delay in sink worker before injected `MySQLSinkExecDMLError` returns.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/sink/mysql.go
+++ b/cdc/sink/mysql.go
@@ -716,6 +716,9 @@ func (s *mysqlSink) prepareDMLs(rows []*model.RowChangedEvent, replicaID uint64,
 
 func (s *mysqlSink) execDMLs(ctx context.Context, rows []*model.RowChangedEvent, replicaID uint64, bucket int) error {
 	failpoint.Inject("MySQLSinkExecDMLError", func() {
+		// Add a delay to ensure the sink worker with `MySQLSinkHangLongTime`
+		// failpoint injected is executed first.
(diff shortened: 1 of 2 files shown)


cdc/sink/mysql.go
```diff
@@ -716,6 +716,9 @@ func (s *mysqlSink) prepareDMLs(rows []*model.RowChangedEvent, replicaID uint64,
 
 func (s *mysqlSink) execDMLs(ctx context.Context, rows []*model.RowChangedEvent, replicaID uint64, bucket int) error {
 	failpoint.Inject("MySQLSinkExecDMLError", func() {
+		// Add a delay to ensure the sink worker with `MySQLSinkHangLongTime`
+		// failpoint injected is executed first.
+		time.Sleep(time.Second * 2)
```
overvenus, line 721: What about use the failpoint expr "1*2s -> return(true)"?

amyangfei: `1*sleep(2000)->1*return(true)`
should be triggered two times, but the first time DMLs will be executed successfully

PR #842, opened 2020-08-12
remove the indie-mark-col field in row change event

leoppro:
remove the indie-mark-col field in row change event


cdc/sink/dispatcher/interface.go
```diff
@@ -103,13 +103,18 @@ func NewDispatcher(cfg *config.ReplicaConfig, partitionNum int32) (Dispatcher, e
 		rule.fromString(ruleConfig.Dispatcher)
 		switch rule {
 		case dispatchRuleRowID, dispatchRuleIndexValue:
+			if cfg.EnableOldValue {
+				log.Warn("This index-value distribution mode " +
+					"does not guarantee row-level orderliness when " +
+					"switching on the old value, so please use caution!")
+			}
```
overvenus, line 110: Could you add the warning to cli too?

leoppro: fine

PR #845, opened 2020-08-13
revoke all leases when delete meta in cli

amyangfei:

If some CDC servers are running and we execute `cdc cli meta delete`, the CDC servers will be running but meta data is incorrect, the cluster is abnormal.

Revoke lease of all captures after meta data is cleared.

- If no cdc server exists when executing `cdc cli meta delete`, all meta will be deleted and leases revoked.
- If some cdc servers are running when executing `cdc cli meta delete`, all cdc servers will be restarted, and all changefeeds are expired. (NOTE in this case some key could be overwritten, because the processor exists after capture suicides, including `task status`, `task position`, `task workload` and `changefeed status`, but `changefeed info` must be deleted and the residual keys make no sense to replication. User can re-execute `cdc cli meta delete` to clean these residual keys)

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


cmd/client_meta.go
```diff
@@ -44,15 +44,30 @@ func newDeleteMetaCommand() *cobra.Command {
...
+			if err != nil {
+				return errors.Trace(err)
+			}
+
+			err = cdcEtcdCli.ClearAllCDCInfo(ctx)
+			if err != nil {
+				return errors.Trace(err)
+			}
+
+			err = cdcEtcdCli.RevokeAllLeases(ctx, leases)
```
leoppro, line 58: how about revoking all leases before clear all cdc info?

amyangfei: If revoking leases first, the capture will suicide, recover, and reconstruct existing changefeeds.

leoppro:
lgtm

PR #848, opened 2020-08-14
fix panic when a columns that are not null by default are null

leoppro:
fix panic when a columns that are not null by default are null

### Check List 

Tests 

 - Unit test
 - Integration test


lance6716:
LGTM

PR #850, opened 2020-08-14
refine file gc in file sorter

amyangfei:

If a puller is canceled, some files may be remained in the `toRemoveFiles` list and not fully removed.

- gc file once before the gc routine exits because of context is Done.
- Add control gc running time restriction.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


leoppro:
@amyangfei please resolve the conflicts

PR #853, opened 2020-08-14
cdc/kv: limit EventFeed retry

overvenus:

Set a rate limit for eventfeed retry to a region that is currently unavailable.
 
### Check List 

Tests 

Related changes

### Release note

- No release note.


cdc/kv/client.go
```diff
@@ -270,6 +298,7 @@ func NewCDCClient(ctx context.Context, pd pd.Client, kvStorage tikv.Storage, cre
 		}{
 			conns: make(map[string]*connArray),
 		},
+		regionLimiters: defaultRegionEventFeedLimiters,
```
MyonKeminta, line 301: May a CDC note have more than one eventfeed to a single table? Looks like DDL meta are usually listened multiple times. May it be a problem to share a single limiter between them?

overvenus: > May a CDC note have more than one eventfeed to a single table? Looks like DDL meta are usually listened multiple times.

This is exactly the reason I choose to share limiters, so we can avoid unnecessary retry (blocks on `onRegionError`) at much as possible.

Issue #860, opened 2020-08-19
Global ResolvedTs should be reset to CheckpointTs when a capture fails


liuzix:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
I made a capture panic by adding some unrelated code. (The first RowChangedEvent of the changefeed made the sink panic)

2. What did you expect to see?
The capture should be recreated on some node and try to sync the RowChangedEvent again. 

3. What did you see instead?
Any restarted capture will panic because it tries to pull starting from the global CheckpointTs, which is smaller than the global ResolvedTs. Error info: "**The CRTs of event is not expected, please report a bug**".

4. Versions of the cluster
    - TiCDC version (execute `cdc version`):

        ```
        Forked from master 9f21207123d03de3ee7c056f92397a5b2bd0a1cc
        ```

amyangfei:
Could you please add detail panic log

amyangfei:
fixed by #875

PR #875, opened 2020-08-23
Processor: fix Fatals and other potential problems when localResolvedTs < globalResolvedTs

liuzix:
- #860
- If a table is recovered immediately after it is dropped, CDC may not output data after the recovery.

- GlobalResolvedTs update events are now masked and not sent to sink if globalResolvedTs > localResolvedTs (happens when an orphan table is just dispatched to this capture).
- AddTable operations are now not marked as done until localResolvedTs has caught up with globalResolvedTs.
- When addtable encounters an existing tableID, it checks if the table is in the process of being removed, if it is, addtable cancels the previous task immediately and start a new one.

### Check List 

Tests 
 - Integration test

### Release note

- No release note


cdc/processor.go
```diff
@@ -937,6 +946,23 @@ func (p *processor) addTable(ctx context.Context, tableID int64, replicaInfo *mo
 						lastResolvedTs = pEvent.CRTs
 						p.localResolvedNotifier.Notify()
 						resolvedTsGauge.Set(float64(oracle.ExtractPhysical(pEvent.CRTs)))
+
+						for {
+							if gTs := atomic.LoadUint64(&p.globalResolvedTs); gTs > lastResolvedTs {
```
overvenus, line 951: To save indentation, you could try if it needs break first.
```suggestion
							if gTs := atomic.LoadUint64(&p.globalResolvedTs); gTs <= lastResolvedTs {
								break
							}
```

cdc/processor.go
```diff
@@ -937,6 +946,23 @@ func (p *processor) addTable(ctx context.Context, tableID int64, replicaInfo *mo
 						lastResolvedTs = pEvent.CRTs
 						p.localResolvedNotifier.Notify()
 						resolvedTsGauge.Set(float64(oracle.ExtractPhysical(pEvent.CRTs)))
+
+						for {
```
overvenus, line 950: Please add comments about why it needs to wait.

cdc/processor.go
```diff
@@ -745,6 +750,9 @@ func (p *processor) syncResolved(ctx context.Context) error {
 			}
 			rows = append(rows, ev.Row)
 		}
+		failpoint.Inject("processorPanic", func() {
+			panic("processorPanic failpoint")
+		})
```
overvenus, line 755: We don't need to write panic explicitly, instead, please use `processorPanic ="1*panic"`  https://github.com/pingcap/failpoint#key-concepts

overvenus:
Please follow [PingCAP PR title style](https://github.com/pingcap/community/blob/master/contributors/commit-message-pr-style.md#pull-request-title-style).

leoppro (approved):

ti-srebot (approved):

cdc/processor.go
```diff
@@ -937,6 +948,29 @@ func (p *processor) addTable(ctx context.Context, tableID int64, replicaInfo *mo
...
+
+						// We need to wait for the globalResolvedTs to be updated,
+						// because we may be a newly recovered table and the localResolvedTs
+						// has just been reset to the previous checkpoint.
+						// Waiting will ensure that any premature flush sink requests have been sent
+						// and that whatever we will send will not be interrupted by sink flush requests
+						// driven by an outdated globalResolvedTs.
+						for {
+							gTs := atomic.LoadUint64(&p.globalResolvedTs)
+							if gTs <= lastResolvedTs {
```
amyangfei, line 960: I don't quite understand the logic, if `gTs > lastResolvedTs` (which means the table is newly added and resovled reset to previous checkpoint), should we wait the `lastResolvedTs` reaches the `gTs`? but it seems the for loop will prevent the update of `lastResolvedTs`

cdc/processor.go
```diff
@@ -123,6 +125,7 @@ type tableInfo struct {
 	sorter      *puller.Rectifier
 	workload    model.WorkloadInfo
 	cancel      context.CancelFunc
+	isDying     uint32
```
amyangfei, line 128: Please add some comments to describe why we introduce this flag

amyangfei (commented):
Please add some details about what this PR fix and how to fix it in PR description.

amyangfei (commented):
LGTM
Good job!

PR #877, opened 2020-08-24
sink: rename mq producer package

leoppro:

Modify the package name to comply with the following conventions:

> Another convention is that the package name is the base name of its source directory; the package in src/encoding/base64 is imported as "encoding/base64" but has name base64, not encoding_base64 and not encodingBase64.

detail: https://golang.org/doc/effective_go.html#package-names


amyangfei (commented):
LGTM

ti-srebot (approved):

PR #881, opened 2020-08-24
mounter: fix three bugs about old value

leoppro:
- fix the old value is lost in the committed event
- fix the pre-row is wrong when pk is handle
- fix the row and pre-row may be ignored by mistake when this row does not contain any `Datums`

### Check List 

Tests 

 - Unit test
 - Integration test


overvenus (commented):
Please add a test and follow [PingCAP PR title style](https://github.com/pingcap/community/blob/master/contributors/commit-message-pr-style.md#pull-request-title-style), rest LGTM

amyangfei (commented):
LGTM

cdc/entry/mounter.go
```diff
@@ -50,8 +50,10 @@ type baseKVEntry struct {
 
 type rowKVEntry struct {
 	baseKVEntry
-	Row    map[int64]types.Datum
-	PreRow map[int64]types.Datum
+	Row         map[int64]types.Datum
+	PreRow      map[int64]types.Datum
+	RowExist    bool
+	PreRowExist bool
```
amyangfei, line 56: recommend adding some comments about why we add these two variables

leoppro: done

liuzix (approved):

ti-srebot (approved):

PR #892, opened 2020-08-27
sink/kafka: zero partition is not valid

leoppro:
fix the bug when create a zero partition topic in kafka 

### Check List 

Tests 

 - Unit test
 - Integration test


amyangfei (commented):
LGTM

ti-srebot (approved):

PR #903, opened 2020-09-01
fix: sync vendor directory in Dockerfile

liuzix:
- Failure to build docker image when a local `vendor` directory does not exist or is not up-to-date.

- Added `RUN go mod vendor` to Dockerfile.

### Check List 

Tests 

### Release note
- No release note


Dockerfile
```diff
@@ -3,6 +3,7 @@ RUN apk add --no-cache git make bash
 WORKDIR /go/src/github.com/pingcap/ticdc
 COPY . .
 ENV CDC_ENABLE_VENDOR=1
+RUN go mod vendor
```
amyangfei, line 6: Should add `vendor` to `.gitignore`?

PR #909, opened 2020-09-01
*: fix clone function of some structs

leoppro:
fix default params could be changed by mistake

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/sink/mysql.go
```diff
@@ -271,6 +271,11 @@ type sinkParams struct {
 	batchReplaceSize    int
 }
 
+func (s *sinkParams) Clone() *sinkParams {
```
overvenus, line 274: I think the easiest way to clone is using value instead of pointer.

overvenus (commented):
Rest LGTM

amyangfei (commented):
LGTM

cdc/sink/mysql.go
leoppro: yes, but this is a more uniform implementation.

ti-srebot (approved):

amyangfei:
Please resolve conflicts

PR #920, opened 2020-09-03
kv/client: Lock region ID as well as range  in the range lock.

MyonKeminta:

Previously, it's possible that a region's range become totally different and disjointed to the old range after a series of splitting and merging, and the range lock will not keep the disjointed ranges mutually exclusive. So in very rare cases, it's possible that CDC sends multiple requests to the same region ID but with different ranges, and leading TiKV to reporting "duplicated request" error.
This PR makes the client able to avoid sending double request to the same region ID in this case.

* In range lock, when locking ranges, the region ID is also considered, which means, two locking attempt with different range but the same region ID will be mutually exclusive.
* Panics when receives duplicated request error.
* Assertions on region ID when unlocking
* Some other changes to support the behavior.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

  - None

### Release note

 - Fix a bug that in some rare cases CDC may encounter "duplicated request" error and possibly stops syncing.


amyangfei:

MyonKeminta:

amyangfei:
Please resolve conflicts

pkg/regionspan/region_range_lock.go
```diff
@@ -286,6 +323,18 @@ func (l *RegionRangeLock) UnlockRange(startKey, endKey []byte, version uint64, c
 	}
 
 	entry := item.(*rangeLockEntry)
+	if entry.regionID != regionID {
+		panic(fmt.Sprintf("unlocked a range but regionID mismatch. expected: %v, found: %v",
```
leoppro, line 327: please use `log.Fatal`

pkg/regionspan/region_range_lock.go
```diff
@@ -286,6 +323,18 @@ func (l *RegionRangeLock) UnlockRange(startKey, endKey []byte, version uint64, c
...
+			regionID, entry.regionID))
+	}
+	if entry != l.regionIDLock[regionID] {
+		panic(fmt.Sprintf("range lock and region id lock mismatch when trying to unlock region %v"+
```
leoppro, line 331: ditto

leoppro (approved):

amyangfei (commented):
LGTM

ti-srebot (approved):

PR #949, opened 2020-09-11
refactor:sink make UnresolvedTxnCache just a txn generator.

qinggniq:
In current implementation, the `cdc/sink/common/UnresolvedTxnCache` contains `mutex, resolvedTs, checkpoint`, which make this struct have more duties (thread safe, maintain resolvedTs and checkpointTs), and know much about the other component(like filter), break SRP. To keep it clean, I removed unnecessary attributes, make it just a "transaction generator".

Move `mutex, resolvedTs, checkpoint` from `UnresolvedTxnCache` to `mysqlSink`.

### Check List 

Code changes


qinggniq:
@5kbpers Hello, what's your opinion? TxnCache will be used in CanalBatchEncoder to generate a transaction, and the encoder is always running in single goroutine, it's wired to make a function thread safe which is always running in single thread.

PR #965, opened 2020-09-22
changefeed: simplify debug log, put task position only value changes

amyangfei:

Some debug logs are too noisy, remove some debug logs that don't carry more debug information.

- In the `checkUpdateTs` function, the `prevMinCheckpointTs` and `prevMinResolvedTs` are initialized with `targetTs`, so the first time check is always logged, we can remove this log.
- The processor always performs an etcd write operation when `flushTaskPosition` is called and has one line log. The `PutTaskPosition` operation can be optimized to `PutTaskPositionOnChange` operation, which means the processor only writes to etcd when the value of task position is changed (or etcd doesn't have the initialized key value). After this optimization etcd write ops in `flushTaskPosition` decreases from 30 to 2. 

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/changefeed.go
+++ b/cdc/changefeed.go
@@ -695,6 +695,7 @@ func (c *changeFeed) calcResolvedTs(ctx context.Context) error {
 	minResolvedTs := c.targetTs
 	minCheckpointTs := c.targetTs
 
+	// prevMinResolvedTs and prevMinCheckpointTs are used for debug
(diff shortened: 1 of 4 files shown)


overvenus:
LGTM

leoppro:
LGTM

ti-srebot (approved):

PR #967, opened 2020-09-23
Part II & Part III: Integration test of canal.

qinggniq:
#697 

Use integration test framework write integration tests for canal.

### Check List 

Tests 

 - Integration test

Code changes

Side effects

 - Increased code complexity


qinggniq:
@liuzix PTAL.

PR #969, opened 2020-09-24
feat: make origin canal encoder can pass canal integration test.

qinggniq:
1. Judge pre columns to figure out `update` and `insert`
2. When type is `char`, it's value should not be encode with ISO charset.
3. Commented the test cases which could not pass due to cdc's and canal adapter's implementation.

### Check List 

Tests 

 - Integration test


qinggniq:
@liuzix PTAL.

liuzix:
Please fix Avro.

PR #976, opened 2020-09-29
sink: support canal old value & fix bug

liuzix:
- Incomplete support for old value in Canal.
- A string type encoding bug in Canal.

### Check List 

Tests 

 - Integration test (the code for integration test is not yet suitable for merging, but it proves correctness of our implementation)

Side effects

 **- Breaking backward compatibility**

### Release note

- No release note

### DIFF
--- a/cdc/sink/mq.go
+++ b/cdc/sink/mq.go
@@ -107,9 +107,7 @@ func newMqSink(
 			return avroEncoder
 		}
 	} else if protocol == codec.ProtocolCanal && !config.EnableOldValue {
-		log.Warn("Old value not enabled when using Canal protocol")
-		return nil, cerror.WrapError(cerror.ErrKafkaInvalidConfig,
-			errors.New("Canal protocol requires old value support"))
+		log.Fatal("Old value not enabled when using Canal protocol")
 	}
 
 	k := &mqSink{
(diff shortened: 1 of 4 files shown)


cdc/sink/mq.go
```diff
@@ -106,6 +106,10 @@ func newMqSink(
 			avroEncoder.SetValueSchemaManager(valueSchemaManager)
 			return avroEncoder
 		}
+	} else if protocol == codec.ProtocolCanal && !config.EnableOldValue {
+		log.Warn("Old value not enabled when using Canal protocol")
```
overvenus, line 110: Looks like it should be log.Error.

overvenus (approved):
LGTM

amyangfei (commented):
Please fix the unit test

leoppro (approved):

ti-srebot (approved):

cdc/sink/mq.go
leoppro: I think it shoule be `FATAL`

amyangfei: What's the purpose of using `log.Fatal` here.

When user creates a new changefeed, `newMqSink` will be called in `verifySink`, IMO return error is enough, we don't need to use fatal everywhrere

overvenus: +1 for `log.Error`

amyangfei:
LGTM

PR #989, opened 2020-10-12
model: fix json name typo

amyangfei:

fix json name typo

use unified key name between json and toml key. `TableName` is not persistent in etcd, so this is a compatible fix

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


overvenus:
LGTM

overvenus:
Will it be a breaking change?

amyangfei:
> Will it be a breaking change?

no, this data is not persistent in etcd @overvenus

leoppro (approved):

ti-srebot (approved):

PR #990, opened 2020-10-13
tests: fix session variable is not reload in row format case

amyangfei:

After global level variable is changed, the existing session doesn't reload this variable.

Re-create a new tidb session after variable `tidb_row_format_version` value is changed.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


overvenus:
LGTM

leoppro (approved):

ti-srebot (approved):

PR #991, opened 2020-10-13
integration: Move canal state checker into healthchecker

qinggniq:
`HealthChecker` should check canal adapter state before start integration test.

Move `checkCanalAdapterState` to `HealthChecker`.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


qinggniq:
@liuzix PTAL.

Issue #985, opened 2020-10-09
Support Canal JSON data format in mq sink


LittleFall:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Flink supports reading Canal JSON data source. To integrate with Flink, TiCDC should support outputting Canal JSON format.

https://ci.apache.org/projects/flink/flink-docs-stable/dev/table/connectors/formats/canal.html#canal-format

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #1013, opened 2020-10-21
sink: support Canal Json protocol

liuzix:
#985 
- Lack of support for Canal Json format, which is supported by Alibaba's Canal Adapter, and usable with Flink as a table data source.

- Implemented `CanalFlatEventBatchEncoder`.
- Added support for protocol `canal-json`.
- Fixed bugs in mqSink.
- Added integration tests.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Related changes

### Release note

- No release note

### Review guidance
- **Most debug logs will be removed prior to merging.**


cdc/sink/codec/canal_flat.go
```diff
@@ -0,0 +1,238 @@
...
+	)
+
+	if len(rowData.BeforeColumns) > 0 {
+		oldData = make(map[string]interface{}, len(rowData.BeforeColumns))
+		for i := range rowData.BeforeColumns {
+			if !rowData.BeforeColumns[i].GetIsNull() {
+				oldData[rowData.BeforeColumns[i].Name] = rowData.BeforeColumns[i].Value
+			} else {
+				oldData[rowData.BeforeColumns[i].Name] = nil
+			}
```
amyangfei, line 99: If the column is nil, what should be output in Canal rule, `null` or empty string(`""`)

If using string is valid in data and oldData, we can save some CPU cost of interface convert 

The Canal Column provides the following method 
https://github.com/pingcap/ticdc/blob/fc58c002bb6efec14b8c4e88ab8325e7e43e1c06/proto/canal/EntryProtocol.pb.go#L622-L627

cdc/sink/codec/canal_flat.go
```diff
@@ -0,0 +1,238 @@
...
+
+	ret := &canalFlatMessage{
+		ID:            0,
+		Schema:        header.SchemaName,
+		Table:         header.TableName,
+		PKNames:       pkNames,
+		IsDDL:         false,
+		EventType:     header.GetEventType().String(),
+		ExecutionTime: header.ExecuteTime,
+		BuildTime:     0,
```
amyangfei, line 126: what does this field mean and why it is zero

liuzix: It means the time when the Canal message is built by TiCDC (or binlog parser or any similar facility). It is not used in the Canal Protobuf implementation, and Alibaba's consumer does not care about it.

cdc/sink/codec/canal_flat.go
liuzix: Empty string in not supported in Canal Json because there would be no place in the Json to indicate whether the value is really null or just an empty string if the column is a string type. Canal Protobuf supports this because of the `IsNullPresent` field.

cdc/sink/codec/canal_flat.go
```diff
@@ -0,0 +1,239 @@
...
+
+// EncodeDDLEvent encodes DDL events
+func (c *CanalFlatEventBatchEncoder) EncodeDDLEvent(e *model.DDLEvent) (*MQMessage, error) {
+	msg, err := c.newFlatMessageForDDL(e)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	value, err := json.Marshal(msg)
+	if err != nil {
+		return nil, errors.Trace(err)
```
amyangfei, line 202:
```suggestion
		return nil, cerrors.WrapError(cerrors.ErrCanalEncodeFailed, err)
```

cdc/sink/codec/canal_flat.go
```diff
@@ -0,0 +1,239 @@
...
+// Size implements the EventBatchEncoder interface
+func (c *CanalFlatEventBatchEncoder) Size() int {
+	return -1
+}
+
+// Reset resets the internal state of the encoder
+func (c *CanalFlatEventBatchEncoder) Reset() {
+	c.unresolvedBuf = make([]*canalFlatMessage, 0)
+	c.resolvedBuf = make([]*canalFlatMessage, 0)
+}
```
amyangfei, line 239: `Reset` is only used in json encoder, is it better to leave a panic("not supported") only

amyangfei (commented):
LGTM

cdc/sink/codec/canal_flat.go
leoppro: and why `ID` is zero? Is the `ID` cared by consumer?

leoppro (approved):

ti-srebot (approved):

cdc/sink/codec/canal_flat.go
```diff
@@ -0,0 +1,243 @@
...
+			break
+		}
+		nextIdx++
+	}
+	c.unresolvedBuf = c.unresolvedBuf[nextIdx:]
+	if len(c.resolvedBuf) > 0 {
+		return EncoderNeedAsyncWrite, nil
+	}
+	return EncoderNoOperation, nil
+}
```
qinggniq, line 197:
```suggestion
func (c *CanalFlatEventBatchEncoder) AppendResolvedEvent(ts uint64) (EncoderResult, error) {
	c.unresolvedIdx := 	i := sort.Search(len(c.unresolvedBuf), func(i int) bool { return c.unresolvedBuf[i].tikvTs > ts })
	if c.unresolvedIdx > 0 {
		return EncoderNeedAsyncWrite, nil
	}
	return EncoderNoOperation, nil
}
```
In `Build`
```go
func (c *CanalFlatEventBatchEncoder) Build() []*MQMessage {
	if c.unresolvedIdx == 0 {
		return nil
	}
	ret := make([]*MQMessage, len(c.unresolvedIdx))
	for i := range ret {
		value, err := json.Marshal(c.unresolvedBuf[i])
		if err != nil {
			log.Fatal("CanalFlatEventBatchEncoder", zap.Error(err))
			return nil
		}
		ret[i] = NewMQMessage(nil, value, c.unresolvedBuf[i].tikvTs)
	}
	c.unresolvedBuf = c.unresolvedBuf[c.unresolvedIdx:]
	return ret
}
```
Maybe this could be more efficient.

PR #1021, opened 2020-10-22
sink: ignore duplicated syncpoint

amyangfei:

Duplicated syncpoint could happen in following scenarios

- ticker and ddl trigger at the same time
- ddl trigger happens first, and in the next round ticker happens
- owner records by ticker once, but crashes, another owner meets the same syncpoint

For each changefeed and primary_ts pair, keep the first secondary_ts is enough, so change the syncpoint SQL to `insert ignore` pattern.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


overvenus:
LGTM

liuzix:
LGTM

ti-srebot (approved):

PR #1030, opened 2020-10-26
sink/mysql: support to pass a timezone dsn parameter

amyangfei:

In some Cloud MySQL service, only timezone SYSTEM and CST is available, we need to allow user to pass a string represented timezone

- cdc server keeps using the `timezone` parameter that is configured in `cdc server` command.
- The `time_zone` query string in sink DSN will be determined by the `time-zone` parameter in sink uri and cdc server's timezone. The decision sequence is as follows:
    1. if `time-zone` is provided in sink uri, but value is empty, the downstream sink DSN will not contain any `time_zone` parameter.
    2. if `time-zone` is provided in sink uri with any value, the downstream sink DSN will contain the same value in `time_zone`
    3. if `time-zone` is not provided in sink uri, query context for timezone and use timezone string value in DSN.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Support to pass a time-zone variable from sink uri, which will be used in sink DSN directly.

### DIFF
--- a/cdc/sink/mysql.go
+++ b/cdc/sink/mysql.go
@@ -286,6 +286,7 @@ type sinkParams struct {
 	enableOldValue      bool
 	safeMode            bool
 	timezone            string
+	tls                 string
 }
 
 func (s *sinkParams) Clone() *sinkParams {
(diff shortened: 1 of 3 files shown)


amyangfei:
PTAL @overvenus @leoppro

overvenus (approved):
LGTM

leoppro (approved):

ti-srebot (approved):

PR #1031, opened 2020-10-27
*: fix many goroutine leak

### LINKED ISSUE #1025: The memory and goroutine leak when PD leader switched
## Bug Report

A user report that the CPU usage of CDC owner periodic increase
image

After some initial investigation, I found that the cpu is occupied by gc. And there is a memory and goroutine leak in TiCDC owner

leoppro:
 fix 6 goroutine leaks in TiCDC fix #1025

a goroutine leak in tidb client is found by this test pingcap/tidb#20654

before this PR:

企业微信截图_10ec311a-66fd-4bf5-abf5-efd2e4f2903d

after this PR:
image

### Release note

- Fix many goroutine leaks.

### DIFF
--- a/cdc/changefeed.go
+++ b/cdc/changefeed.go
@@ -20,8 +20,6 @@ import (
 	"sync"
 	"time"
 
-	tidbkv "github.com/pingcap/tidb/kv"
-
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	timodel "github.com/pingcap/parser/model"
@@ -116,7 +114,6 @@ type changeFeed struct {
 	lastRebalanceTime time.Time
 
 	etcdCli kv.CDCEtcdClient
-	kvStore tidbkv.Storage
 }
 
 // String implements fmt.Stringer interface.
@@ -958,10 +955,6 @@ func (c *changeFeed) Close() {
 		if err != nil {
 			log.Warn("failed to close owner sink", zap.Error(err))
 		}
-	}
-	err = c.kvStore.Close()
-	if err != nil {
-		log.Warn("failed to close kv store", zap.Error(err))
 	}
 	log.Info("changefeed closed", zap.String("id", c.id))
 }
--- a/cdc/changefeed.go
+++ b/cdc/changefeed.go
@@ -20,6 +20,8 @@ import (
 	"sync"
 	"time"
 
+	tidbkv "github.com/pingcap/tidb/kv"
+
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	timodel "github.com/pingcap/parser/model"
@@ -114,6 +116,7 @@ type changeFeed struct {
 	lastRebalanceTime time.Time
 
 	etcdCli kv.CDCEtcdClient
+	kvStore tidbkv.Storage
 }
 
 // String implements fmt.Stringer interface.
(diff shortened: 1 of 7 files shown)


cdc/changefeed.go
```diff
@@ -20,6 +20,8 @@ import (
 	"sync"
 	"time"
 
+	tidbkv "github.com/pingcap/tidb/kv"
+
```
overvenus, line 24:
```suggestion
	tidbkv "github.com/pingcap/tidb/kv"
```

pkg/notify/notify.go
```diff
@@ -60,7 +60,12 @@ func (r *Receiver) signalNonBlocking() bool {
 func (r *Receiver) signalTickLoop() {
 	go func() {
 	loop:
-		for range r.ticker.C {
+		for {
+			select {
+			case <-r.closeCh:
+				break
+			case <-r.ticker.C:
+			}
```
overvenus, line 68: Does it cause performance regression?

overvenus (commented):
Could you use tidb/util/testleak to check leaks?

https://github.com/pingcap/tidb/tree/master/util/testleak

pkg/notify/notify.go
leoppro: This won't be a performance bottleneck here theoretically.

cdc/owner.go
```diff
@@ -343,7 +338,11 @@ func (o *Owner) newChangeFeed(
...
 	go func() {
-		err := <-errCh
+		var err error
+		select {
+		case <-ctx.Done():
+		case err = <-errCh:
+		}
 		if errors.Cause(err) != context.Canceled {
 			log.Error("error on running owner", zap.Error(err))
 		} else {
```
amyangfei, line 348: If the branch `<-ctx.Done()` is fired, `err` is nil in L347, the error log is not as wished.

cdc/owner.go
```diff
@@ -205,11 +204,7 @@ func (o *Owner) newChangeFeed(
 		failpoint.Return(nil, errors.New("failpoint injected retriable error"))
 	})
 
-	// TODO here we create another pb client,we should reuse them
-	kvStore, err := kv.CreateTiStore(strings.Join(o.pdEndpoints, ","), o.credential)
-	if err != nil {
-		return nil, err
-	}
+	kvStore := util.KVStorageFromCtx(ctx)
```
overvenus, line 207: Is kvstorage thread-safe?

leoppro: yes, and the recommended usage is to use him as a singleto.

cdc/owner.go
leoppro: Yes, but I haven't any better solution. And when the ctx is done, the error from the sink must be canceled.

amyangfei: I mean if `ctx.Done` is triggered first, the errCh is not fetched and `err` is nil, we should log as normal exit then.

leoppro: updated

amyangfei (commented):
LGTM

cdc/kv/client.go
```diff
@@ -351,12 +351,14 @@ func (c *CDCClient) newStream(ctx context.Context, addr string, storeID uint64)
 		}
 		err = version.CheckStoreVersion(ctx, c.pd, storeID)
 		if err != nil {
+			conn.Close()
```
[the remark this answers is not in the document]
leoppro: emmm. I have not idea...

overvenus:
LGTM

ti-srebot (approved):

cdc/owner.go
overvenus: So what about a global variable?

PR #1045, opened 2020-10-29
owner: check if the checkpoint ts is less then gc safe time when create the changefeed

### LINKED ISSUE #1014: CDC should not work if changefeed's checkpoint is earlier than current gc safepoint
## Bug Report

* Create a changefeed
* Stop CDC server for more than 24h so that the service safepoint is expired and invalidated, and the data after that time is gc-ed.
* Start CDC server

The changefeed is broken and should not be able to continue

3. What did you see instead?

CDC still tries to continue the changefeed and sets a service safepoint at the checkpoint, which is before the current actual gc safepoint.

4. Versions of the cluster

        ```
        4.0.4
        ```

        ```
        4.0.4
        ```

    I didn't confirm if this bug still exist in current version, but I suspect it does.

leoppro:
check if the checkpoint ts is less then gc safe time when create the changefeed
fix #1014 

### Check List 

Tests 

 - Integration test
the `changefeed_error` test in the current integration test is already do good.
try to follow the step in #1014, we can see the owner throw an error immediately for now.

### Release note

- No release note


cdc/kv/etcd.go
```diff
@@ -853,3 +855,22 @@ func (c CDCEtcdClient) GetOwnerID(ctx context.Context, key string) (string, erro
 	}
 	return string(resp.Kvs[0].Value), nil
 }
+
+// GetGCSafePoint get tikv GC safe point
+func (c CDCEtcdClient) GetGCSafePoint(ctx context.Context) (uint64, error) {
+	resp, err := c.Client.Get(ctx, tikv.GcSavedSafePoint)
```
overvenus, line 861: Could you get gc safepoint by `pd_client.UpdateGCSafePoint(0)`?

cdc/owner.go
```diff
@@ -186,16 +186,16 @@ func (o *Owner) addOrphanTable(cid model.CaptureID, tableID model.TableID, start
...
 	log.Info("Find new changefeed", zap.Stringer("info", info),
 		zap.String("id", id), zap.Uint64("checkpoint ts", checkpointTs))
-
+	safePoint, err := cli.GetGCSafePoint(ctx)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if checkpointTs < safePoint {
+		return nil, tikv.ErrGCTooEarly.GenWithStackByArgs(checkpointTs, safePoint)
+	}
```
overvenus, line 198: Please add a **unit** test. (Test newChangeFeed is enough)

overvenus:
Could you move this forward? tikv/pd#3136 does not block this fix I think.

leoppro:
> Could you move this forward? [tikv/pd#3136](https://github.com/tikv/pd/issues/3136) does not block this fix I think.

If tikv/pd#3136 is not fixed, cdc will throw an error when create a changefeed, because cdc can't get gc safetime before the gc worker running first time.

pkg/util/gc_service.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/pingcap/errors"
+	"github.com/pingcap/tidb/store/tikv"
+	pd "github.com/tikv/pd/client"
+)
+
+const (
+	// CDCChangefeedCreatingServiceGCSafePointID is service GC safe point ID
+	CDCChangefeedCreatingServiceGCSafePointID = "ticdc-changefeed-creating"
+	// CDCChangefeedCreatingServiceGCSafePointTTL is service GC safe point TTL
+	CDCChangefeedCreatingServiceGCSafePointTTL = 10 * 60 // 10 mins
```
amyangfei, line 28: Not necessary to export these two variables

cmd/util.go
```diff
@@ -22,10 +22,11 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
-	"strconv"
 	"strings"
 	"syscall"
 
+	"github.com/pingcap/ticdc/pkg/util"
+
```
overvenus, line 29:
```suggestion
	"github.com/pingcap/ticdc/pkg/util"
```

leoppro:
@amyangfei PTAL @liuzix

amyangfei (commented):
LGTM

pkg/util/gc_service_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"context"
+	"math"
+
+	"github.com/pingcap/check"
+	"github.com/pingcap/kvproto/pkg/metapb"
+	"github.com/pingcap/kvproto/pkg/pdpb"
+	pd "github.com/tikv/pd/client"
+)
+
+type gcServiceuite struct {
```
overvenus, line 26:
```suggestion
type gcServiceSuite struct {
```

pkg/util/gc_service_test.go
```diff
@@ -0,0 +1,171 @@
...
+	s.pdCli.UpdateServiceGCSafePoint(ctx, "service2", 10, 80) //nolint:errcheck
+	s.pdCli.UpdateServiceGCSafePoint(ctx, "service3", 10, 70) //nolint:errcheck
+	err = CheckSafetyOfStartTs(ctx, s.pdCli, 65)
+	c.Assert(err, check.IsNil)
+	c.Assert(s.pdCli.serviceSafePoint, check.DeepEquals, map[string]uint64{"service1": 60, "service2": 80, "service3": 70, "ticdc-changefeed-creating": 65})
+
+}
+
+type mockPdClientForServiceGCSafePoint struct {
+	baseMockPdClient
```
overvenus, line 48:
```suggestion
	pd.Client
```

How about this? So that we do not need to impl all methods.

pkg/config/config.go
```diff
@@ -25,8 +25,9 @@ import (
 )
 
 var defaultReplicaConfig = &ReplicaConfig{
-	CaseSensitive:  true,
-	EnableOldValue: false,
+	CaseSensitive:    true,
+	EnableOldValue:   false,
+	CheckGCSafePoint: true,
```
overvenus, line 30: Why add flag? I think it must be checked all the time.

leoppro: cause of the bug in pd (tikv/pd#3128), if TiCDC(4.0.9+) try to connect to a TiKV(which of the version is less then 4.0.8), the gc safe point check will failed. so we add a flag to skip the check

overvenus: What about checking GC safepoint by cluster version automatically?

leoppro: It's not strict I think. the version check is not correct all the time(for example : the version of a binary built in the master branch may be 4.0.x-xxx-dirty)

leoppro: and multi-version components in one cluster are not recommended for production. I think it's enough that we just make sure that the cdc is able to connect to the old version tikv

overvenus:
LGTM

ti-srebot (approved):

PR #1052, opened 2020-11-03
tests: fix syncpoint check in ddl

amyangfei:

Fix https://github.com/pingcap/ticdc/issues/1050

Currently TiCDC only subscribes key value change of `DDLJobList` and `DDLJobAddIdxList` key, so we can't find a strong schema version barrier in TiCDC.

See the following timeline, before T3 the snapshot reads the old schema, after T3 the snapshot reads the new schema.

TiCDC could record a syncpoint with primary ts (T5) between T3 and T4,  and at T5, the schema in upstream is new vesrion, while the schema in downstream is old version, TiCDC will execute DDL at T4

tso2

We have two choices

- Monit more DDL metadata in TiKV, and be aware of the table info commit transaction at T3, and then skip to record syncpoint when primary ts happens in a DDL. 
- Just leave the `inaccurate` syncpoint record (Only one DDL missed compared to the upstream) and fix test case.

Solution 1 is overheight. Since the missing DDL doesn't introduce data loss, maybe keep the `inaccurate` syncpoint record is acceptable.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/tests/syncpoint/run.sh
+++ b/tests/syncpoint/run.sh
@@ -81,12 +81,34 @@ function deployConfig() {
     echo "snapshot = \"$2\"" >> $CUR/conf/diff_config.toml
 }
 
+# check wheter the given tso happens in a DDL job, if true returns the start_ts
+# and commit_ts of the DDL job
+function checkPrimaryTsNotInDDL() {
+    primary_ts=$1
+    tsos=$2
+    count=$(( ${#tsos[@]} / 2 ))
+    for i in $(seq 1 $count); do
+        start_ts=${tsos[(( 2 * $i - 2 ))]}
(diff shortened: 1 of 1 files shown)


overvenus:
If there are multiple TiDB in upstream, is it possible that some TiDBs do not have the latest schema version between T3 and T4?

amyangfei:
> If there are multiple TiDB in upstream, is it possible that some TiDBs do not have the latest schema version between T3 and T4?

Yes, this is acceptable and as expected. Before DDL is synced, both TiDB and TiCDC has the logic to fillback default value for missing column @overvenus

leoppro (approved):

overvenus (approved):
LGTM

ti-srebot (approved):

PR #1072, opened 2020-11-12
owner: fix multiple owners exist when owner campaign key is deleted

amyangfei:

Fix https://github.com/pingcap/ticdc/issues/1044

In some cases, if the campaign owner key is deleted, the owner itself doesn't aware of it and continues running the owner routines. While at the same time, another capture will campaign to be the new owner. In this scenario, multiple owners could exist.

Add a background campaign key watcher in the owner background.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Fix the bug that multiple owners could exist when owner campaign key is deleted


cdc/owner.go
```diff
@@ -1019,6 +1028,36 @@ loop:
...
+	}
+restart:
+	wch := o.etcdClient.Client.Watch(ctx, key, clientv3.WithRev(resp.Header.Revision+1))
+	for resp := range wch {
+		err := resp.Err()
+		if err != nil {
+			if err != mvcc.ErrCompacted {
+				log.Error("watch owner campaign key failed, restart the watcher", zap.Error(err))
+			}
+			goto restart
```
overvenus, line 1049: Does resp.Header.Revision change if goto retstart?

cdc/owner.go
```diff
@@ -1019,6 +1028,36 @@ loop:
...
+	key := fmt.Sprintf("%s/%x", kv.CaptureOwnerKey, o.session.Lease())
+	resp, err := o.etcdClient.Client.Get(ctx, key)
+	if err != nil {
+		return cerror.WrapError(cerror.ErrPDEtcdAPIError, err)
+	}
+	if resp.Count == 0 {
+		return cerror.ErrOwnerCampaignKeyDeleted.GenWithStackByArgs()
+	}
+restart:
+	wch := o.etcdClient.Client.Watch(ctx, key, clientv3.WithRev(resp.Header.Revision+1))
```
overvenus, line 1042: Why `+1`?

cdc/owner.go
amyangfei: fixed

cdc/owner.go
amyangfei: The change with revision `resp.Header.Revision` is not needed, so `+1`

leoppro (approved):

liuzix:
LGTM

ti-srebot (approved):

cdc/owner.go
overvenus: Could you add some comments?

cdc/owner.go
```diff
@@ -1020,6 +1029,36 @@ loop:
...
+	key := fmt.Sprintf("%s/%x", kv.CaptureOwnerKey, o.session.Lease())
+restart:
+	resp, err := o.etcdClient.Client.Get(ctx, key)
+	if err != nil {
+		return cerror.WrapError(cerror.ErrPDEtcdAPIError, err)
+	}
+	if resp.Count == 0 {
+		return cerror.ErrOwnerCampaignKeyDeleted.GenWithStackByArgs()
+	}
+	wch := o.etcdClient.Client.Watch(ctx, key, clientv3.WithRev(resp.Header.Revision+1))
```
amyangfei, line 1043:
```suggestion
	// watch key change from the next revision relatived to the current 
	wch := o.etcdClient.Client.Watch(ctx, key, clientv3.WithRev(resp.Header.Revision+1))
```

PR #1078, opened 2020-11-14
tests: add leak test in all unit test cases

amyangfei:

Add a leak test 

- Use TiDB leak test library
- Add leak test in every unit test entry
- Fix several goroutine leaks in unit test

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


pkg/util/testleak/add-leaktest.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
```
overvenus, line 1: Could you collect all scripts in a dedicated folder?

pkg/util/testleak/add-leaktest.sh
```diff
@@ -0,0 +1,34 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Usage: add-leaktest.sh pkg/*_test.go
+
+set -eu
+
+sed -i'~' -e '
+  /^func (.*Suite) Test.*(c \*check.C) {/ {
```
overvenus, line 20:
```suggestion
  /^func (.*) Test.*(c \*check.C) {/ {
```

What about removing "Suite"?

cdc/kv/etcd.go
```diff
@@ -126,6 +126,11 @@ func NewCDCEtcdClient(ctx context.Context, cli *clientv3.Client) CDCEtcdClient {
 	return CDCEtcdClient{Client: etcd.Wrap(cli, metrics)}
 }
 
+// Close releases resources in CDCEtcdClient
+func (c CDCEtcdClient) Close() error {
```
overvenus, line 130: Do we leak the client in non-testing code?

overvenus (commented):
Rest LGTM

pkg/util/testleak/add-leaktest.sh
amyangfei: addressed

pkg/util/testleak/add-leaktest.sh
amyangfei: addressed

cdc/kv/etcd.go
amyangfei: I will investigate it

amyangfei: But I suggest merging this PR first

leoppro (approved):

overvenus:
LGTM

ti-srebot (approved):

PR #1079, opened 2020-11-16
sink: Kafka message length and batch size limit for Open Protocol

liuzix:
- Open Protocol messages risk exceeding the message length limit for Kafka.
- Users cannot control the number of records in a single message produced by Open Protocol.

- The Open Protocol Encoder reads the `max-message-bytes` parameter from the `sink-uri` and respects this limit.
- The Open Protocol Encoder now reads `max-batch-size` from `opts` and respects this limit.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note

- No release note


cdc/sink/codec/json.go
```diff
@@ -481,6 +505,31 @@ func (d *JSONEventBatchEncoder) Reset() {
...
+func (d *JSONEventBatchEncoder) SetParams(params map[string]string) error {
+	var err error
+	if maxMessageBytes, ok := params["max-message-bytes"]; ok {
+		d.maxKafkaMessageSize, err = strconv.Atoi(maxMessageBytes)
+		if err != nil {
+			// TODO add error code
+			return errors.Trace(err)
+		}
+	} else {
+		d.maxKafkaMessageSize = 64 * 1024 * 1024 // 64M
```
amyangfei, line 518: Please add a const variable for the default value of `max-message-bytes`. Ditto for the `max-batch-size`

kafka_consumer/main.go
```diff
@@ -370,6 +381,10 @@ func (c *Consumer) ConsumeClaim(session sarama.ConsumerGroupSession, claim saram
 ClaimMessages:
 	for message := range claim.Messages() {
 		log.Info("Message claimed", zap.Int32("partition", message.Partition), zap.ByteString("key", message.Key), zap.ByteString("value", message.Value))
+		if len(message.Key)+len(message.Value) > kafkaMaxMessageBytes {
+			log.Fatal("kafka max-messages-bytes exceeded", zap.Int("max-message-bytes", kafkaMaxMessageBytes),
+				zap.Int("recevied-bytes", len(message.Key)+len(message.Value)))
+		}
```
amyangfei, line 387: Is it possible that the message size of only one row changed exceeds the `kafkaMaxMessageBytes`? (Although the ycsb workload doesn't contain a wide table) In this test `max-message-bytes` is set to 4096

kafka_consumer/main.go
```diff
@@ -71,6 +73,7 @@ func init() {
 	flag.StringVar(&ca, "ca", "", "CA certificate path for Kafka SSL connection")
 	flag.StringVar(&cert, "cert", "", "Certificate path for Kafka SSL connection")
 	flag.StringVar(&key, "key", "", "Private key path for Kafka SSL connection")
+	flag.IntVar(&maxBatchSize, "max-batch-size", 4096, "The limit of acceptable batch size in Open Protocol")
```
leoppro, line 76: should this flag set in sink-uri?

kafka_consumer/main.go
```diff
@@ -122,6 +125,16 @@ func init() {
 		}
 		kafkaPartitionNum = int32(c)
 	}
+
+	s = upstreamURI.Query().Get("max-message-bytes")
```
leoppro, line 129: what difference between max-batch-size and max-message-bytes

amyangfei (commented):
LGTM

kafka_consumer/main.go
liuzix: max-batch-size concerns how many row changes can be packed into a single Kafka message, while max-message-bytes limits how many bytes can be in a single Kafka message.

kafka_consumer/main.go
liuzix: We can put this into the sink-uri.

leoppro: I think this param should not exist in the cmd flag

kafka_consumer/main.go
leoppro: what happened when the MQ is plusar?

leoppro (approved):

ti-srebot (approved):

PR #1097, opened 2020-11-19
owner: check if the checkpoint ts is less then gc safe time when create the changefeed (#1045)

### LINKED ISSUE #1014: CDC should not work if changefeed's checkpoint is earlier than current gc safepoint

ti-srebot:
cherry-pick #1045 to release-4.0

---

check if the checkpoint ts is less then gc safe time when create the changefeed
fix #1014 

### Check List 

Tests 

 - Integration test
the `changefeed_error` test in the current integration test is already do good.
try to follow the step in #1014, we can see the owner throw an error immediately for now.

### Release note

- No release note


overvenus:
LGTM

PR #1115, opened 2020-11-24
schemaStorage: reduce memory usage and add unit tests

### LINKED ISSUE #1098: TiCDC costs too much memory if table count is large or DDL is frequent
## Bug Report

   In one user's scenario:

   - there exists about 1000 tables in upstream
   - DDL operation is frequent
   - Each DDL could involve hundreds of tables.

    Replication is normal

3. What did you see instead?

    - TiCDC costs too much memory and could be OOM, for some reasons

      - Although changefeed only replicates part of tables, all tables' information is cached in memory.
      - TiCDC owner executes DDL one by one, so the schema versions could be cached too many in `processor`, and GC is delayed because of the DDL execution latency. Such as upstream executes 100 DDLs in sequence, TiCDC could cache 101 copies of schema version at most.

    - memory profile: [heap.log](https://github.com/pingcap/ticdc/files/5564822/heap.log)

4. Versions of the cluster

        ```
        v4.0.8
        ```

        ```
        v4.0.8
        ```

leoppro:
fix #1098

* reduce memory usage in schema storage by avoiding table info replication
* add many unit tests

500 table created:
before this PR:
image

after this PR:
image

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/entry/schema_storage.go
```diff
@@ -223,13 +227,21 @@ func (s *schemaSnapshot) Clone() *schemaSnapshot {
 
 	tables := make(map[int64]*model.TableInfo, len(s.tables))
 	for k, v := range s.tables {
-		tables[k] = v.Clone()
+		tables[k] = v
```
amyangfei, line 230: Is it safe to change L224 to `schemas[k] = v`.

The behavior is to change `TableInfo` when applying DDL changes, better to leave some comments here about the principle and comments the function is not a totally `Clone`.

cdc/entry/schema_storage_test.go
```diff
@@ -770,3 +777,195 @@ func (t *schemaSuite) TestExplicitTables(c *check.C) {
...
+			snapFromMeta, err := newSchemaSnapshotFromMeta(meta, ts, false)
+			c.Assert(err, check.IsNil)
+			snapFromSchemaStore, err := scheamStorage.GetSnapshot(ctx, ts)
+			c.Assert(err, check.IsNil)
+
+			tidySchemaSnapshot(snapFromMeta)
+			tidySchemaSnapshot(snapFromSchemaStore)
+			// check if the two snapshot are equal.
+			checkSnapsEquals(snapFromMeta, snapFromSchemaStore)
+		}
```
amyangfei, line 907: Could we also check the historical snapshot from the `schemaStorage`, such as querying by `ts-1` or last job FinishedTs

zhangjinpeng1987:
How about 10K tables? Maybe we should load the table schema lazily? Using the LRU strategy?

leoppro:
> How about 10K tables? Maybe we should load the table schema lazily? Using the LRU strategy?

The way to reduce memory usage in this PR is cheap and effective.
To support 10K tables, only reduce the memory usage is not enough. We need to consider memory, ETCD access frequency, and so on. We have a plan to solve the problem by optimizing the replication model.

cdc/entry/schema_storage.go
```diff
@@ -376,18 +384,20 @@ func (s *schemaSnapshot) createSchema(db *timodel.DBInfo) error {
 		return cerror.ErrSnapshotSchemaExists.GenWithStackByArgs(db.Name, db.ID)
 	}
 
-	s.schemas[db.ID] = db.Clone()
```
amyangfei, line 379: Is it better to keep this `Clone`, and the `DDL job` can be recycled

cdc/entry/schema_storage.go
```diff
@@ -376,18 +384,20 @@ func (s *schemaSnapshot) createSchema(db *timodel.DBInfo) error {
...
 	return nil
 }
 
 func (s *schemaSnapshot) replaceSchema(db *timodel.DBInfo) error {
-	if _, ok := s.schemas[db.ID]; !ok {
+	_, ok := s.schemas[db.ID]
+	if !ok {
 		return cerror.ErrSnapshotSchemaNotFound.GenWithStack("schema %s(%d) not found", db.Name, db.ID)
 	}
-	s.schemas[db.ID] = db.Clone()
```
amyangfei, line 390: ditto

cdc/entry/schema_storage.go
```diff
@@ -376,18 +384,20 @@ func (s *schemaSnapshot) createSchema(db *timodel.DBInfo) error {
 		return cerror.ErrSnapshotSchemaExists.GenWithStackByArgs(db.Name, db.ID)
 	}
 
-	s.schemas[db.ID] = db.Clone()
+	s.schemas[db.ID] = db
```
leoppro, line 387:
```suggestion
	s.schemas[db.ID] = db.Clone()
```

cdc/entry/schema_storage.go
```diff
@@ -376,18 +384,20 @@ func (s *schemaSnapshot) createSchema(db *timodel.DBInfo) error {
...
 }
 
 func (s *schemaSnapshot) replaceSchema(db *timodel.DBInfo) error {
-	if _, ok := s.schemas[db.ID]; !ok {
+	_, ok := s.schemas[db.ID]
+	if !ok {
 		return cerror.ErrSnapshotSchemaNotFound.GenWithStack("schema %s(%d) not found", db.Name, db.ID)
 	}
-	s.schemas[db.ID] = db.Clone()
+	s.schemas[db.ID] = db
```
leoppro, line 400:
```suggestion
	s.schemas[db.ID] = db.Clone()
```

amyangfei (commented):
LGTM

liuzix:
LGTM

ti-srebot (approved):

PR #1119, opened 2020-11-25
server: disable tikv batch client in tidb library

amyangfei:

TiCDC uses TiDB library to query necessary information from TiKV, such using `meta.ListDatabase` or `meta.ListTables`.

The gRPC connection manager in TiDB will start a background routine(`batchSendLoop`) when batch is enabled, which is not used in TiCDC.

https://github.com/pingcap/tidb/blob/f9cd7ed69dc07d4a83e78cda3c70528aec3d53e8/store/tikv/client.go#L193-L195

Force set `TiKVClient.MaxBatchSize` of tidb global config to `0` to disable batch client.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


cmd/server.go
```diff
@@ -46,7 +47,15 @@ var (
 	}
 )
 
+func patchTiDBConf() {
+	conf := config.NewConfig()
```
leoppro, line 51: we can update the tidb config like this:
https://github.com/pingcap/ticdc/pull/1115/files#diff-fc79b7274e95fe44af88567f9dbe6e2c49c386a923698cc7541157b5747ff079R870-R872

amyangfei: addressed

liuzix:
LGTM

leoppro (approved):

ti-srebot (approved):

PR #1124, opened 2020-11-25
server: disable tikv batch client in tidb library (#1119)

ti-srebot:
cherry-pick #1119 to release-4.0

---

TiCDC uses TiDB library to query necessary information from TiKV, such using `meta.ListDatabase` or `meta.ListTables`.

The gRPC connection manager in TiDB will start a background routine(`batchSendLoop`) when batch is enabled, which is not used in TiCDC.

https://github.com/pingcap/tidb/blob/f9cd7ed69dc07d4a83e78cda3c70528aec3d53e8/store/tikv/client.go#L193-L195

Force set `TiKVClient.MaxBatchSize` of tidb global config to `0` to disable batch client.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


cmd/server.go
```diff
@@ -20,9 +20,11 @@ import (
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	"github.com/pingcap/ticdc/cdc"
+	"github.com/pingcap/ticdc/pkg/config"
```
amyangfei, line 23:
```suggestion
```

amyangfei (commented):
LGTM

PR #1127, opened 2020-11-26
schemaStorage: reduce memory usage and add unit tests (#1115)

### LINKED ISSUE #1098: TiCDC costs too much memory if table count is large or DDL is frequent

ti-srebot:
cherry-pick #1115 to release-4.0

---

fix #1098

* reduce memory usage in schema storage by avoiding table info replication
* add many unit tests

500 table created:
before this PR:
image

after this PR:
image

### Check List 

Tests 

 - Unit test
 - Integration test


leoppro:
lgtm

PR #1156, opened 2020-12-01
tests: add unit tests for cdc/puller/puller.go

amyangfei:

Add unit tests for cdc/puller/puller.go

- Extract kv client into a new interface `CDCKVClient`
- Add a simple injected mock puller, which can be injected with `RegionFeedEvent`

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -271,6 +271,28 @@ func (rl *regionEventFeedLimiters) getLimiter(regionID uint64) *rate.Limiter {
 	return limiter
 }
 
+// CDCKVClient is an interface to receives kv changed logs from TiKV
+type CDCKVClient interface {
+	EventFeed(
+		ctx context.Context,
+		span regionspan.ComparableSpan,
+		ts uint64,
+		enableOldValue bool,
+		lockResolver txnutil.LockResolver,
+		isPullerInit PullerInitialization,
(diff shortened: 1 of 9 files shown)


leoppro (approved):

cdc/kv/client.go
```diff
@@ -271,6 +271,28 @@ func (rl *regionEventFeedLimiters) getLimiter(regionID uint64) *rate.Limiter {
...
+// CDCKVClient is an interface to receives kv changed logs from TiKV
+type CDCKVClient interface {
+	EventFeed(
+		ctx context.Context,
+		span regionspan.ComparableSpan,
+		ts uint64,
+		enableOldValue bool,
+		lockResolver txnutil.LockResolver,
+		isPullerInit PullerInitialization,
+		eventCh chan<- *model.RegionFeedEvent,
```
liuzix, line 283: Would it be better if the writer "owned" the channel?

amyangfei: Do you mean the kv client creates a channel and passes it back to the outer caller? I prefer to keep current way, maybe in the future the eventCh could be shared by multiple KV writer

liuzix: Okay, seems fair.

liuzix:
LGTM

ti-srebot (approved):

PR #1161, opened 2020-12-02
refactor: EtcdWorker

liuzix:
- To provide Owner and Processor a unified way to process Etcd changes, with snapshot isolation with respect to the Etcd storage.

- Added EtcdWorker.

### Check List 

Tests 

 - Unit test

Code changes

### Release note

- No release note

### Design notes

- The main goal of designing `EtcdWorker` is to (1) decouple Etcd operations from the implementation of the replication model, and (2) provide a mechanism to write data-race-free Owner and Processor.
- `EtcdWorker` models an "owner-driver" or a "processor-driver", which runs in a loop, updating global statuses for the use of the driven component. 
- `ReactorState` models the global statuses that have been managed directly by the Owner and the Processor until now. `ReactorState` is an interface that can be implemented according to the internals of the Owner and the Processor, and it acts as a bridge between the Owner/Processor and the `EtcdWorker`.
- `Reactor` is an interface that should be implemented by the Owner/Processor. The design of the `Reactor` requires the core logic of the Owner/Processor to operate the "ticks". The ticks are essentially a next-state logic, which should ideally be as pure (i.e. without side effects) as possible, save for necessary management of pipeline statuses in a Processor.


pkg/orchestrator/etcd_worker_test.go
```diff
@@ -0,0 +1,266 @@
...
+	dir := c.MkDir()
+	url, etcdServer, err := etcd.SetupEmbedEtcd(dir)
+	c.Assert(err, check.IsNil)
+	s.etcdServer = etcdServer
+	s.endpoints = []string{url.String()}
+
+	log.Debug("Set up embedded etcd", zap.Strings("endpoints", s.endpoints))
+}
+
+func (s *etcdWorkerSuite) TestEtcdSum(c *check.C) {
```
overvenus, line 207: Could you test some anomalies against to etcd guarantee? To make sure CDC is comfortable with  some corner cases.

pkg/orchestrator/etcd_worker.go
```diff
@@ -0,0 +1,260 @@
...
+	client  *etcd.Client
+	reactor Reactor
+	state   ReactorState
+	// rawState is the local cache of the latest Etcd state.
+	rawState       map[string][]byte
+	pendingUpdates []*etcdUpdate
+	// revision is the Etcd revision of the latest event received from Etcd
+	// (which has not necessarily been applied to the ReactorState)
+	revision int64
+	prefix   string
```
amyangfei, line 41: I have several problems
- How many etcd workers will be maintained, will they be classified by `prefix`
- What happens when an EtcdWorker manages a prefix, while some `reactors` only care about a sub-prefix, such as EtcdWorker manages `/tidb/cdc/job`, and each reactor only updates `/tidb/cdc/job/<reactor-id>`
- Will all etcd key-value operations be executed in serialize

pkg/orchestrator/etcd_worker.go
```diff
@@ -0,0 +1,260 @@
...
+		} else {
+			// We are safe to update the ReactorState only if there is no pending patch.
+			for _, update := range worker.pendingUpdates {
+				worker.state.Update(worker.removePrefixBytes(update.key), update.value)
+			}
+
+			worker.pendingUpdates = worker.pendingUpdates[:0]
+			nextState, err := worker.reactor.Tick(ctx, worker.state)
+			if err != nil {
+				if errors.Cause(err) == ErrReactorFinished {
```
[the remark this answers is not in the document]
liuzix: Yes, `context.Canceled` should also be treated as a signal for normal exits. I will adjust the code.

pkg/orchestrator/etcd_worker.go
liuzix: - Ideally each instance of owner or processor should maintain one `EtcdWorker`, and they should watch the prefix that is the highest level in the CDC key space.
- We can either maintain separate EtcdWorkers for each job, or create a super-reactor that dispatches events according to the subprefix.
- Yes, reactors see changes in the linearized order determined by the Etcd server. And reactors can see all changes it has made in previous ticks.

pkg/orchestrator/interfaces.go
```diff
@@ -0,0 +1,54 @@
...
+	Tick(ctx context.Context, state ReactorState) (nextState ReactorState, err error)
+}
+
+var (
+	// ErrEtcdTryAgain is used by a PatchFunc to force a transaction abort.
+	ErrEtcdTryAgain = errors.New("ErrEtcdTryAgain")
+	// ErrEtcdIgnore is used by a PatchFunc to signal that the reactor no longer wishes to update Etcd.
+	ErrEtcdIgnore = errors.New("ErrEtcdIgnore")
+	// ErrReactorFinished is used by reactor to signal a **normal** exit.
+	ErrReactorFinished = errors.New("ErrReactorFinished")
```
overvenus, line 33: What about moving them to pkg/errors?

pkg/orchestrator/etcd_worker_test.go
```diff
@@ -0,0 +1,357 @@
...
+		val:       0,
+		isUpdated: false,
+	})
+	c.Assert(err, check.IsNil)
+	errg := &errgroup.Group{}
+	errg.Go(func() error {
+		return reactor.Run(ctx, 10*time.Millisecond)
+	})
+
+	time.Sleep(500 * time.Millisecond)
```
overvenus, line 344: Why sleep 500ms?

overvenus:
PTAL @leoppro

pkg/orchestrator/etcd_worker_test.go
liuzix: The test case requires the reactor to run at least once before putting more data into Etcd.

pkg/orchestrator/interfaces.go
liuzix: My thoughts are that they are not real errors, but are status codes and they do not signal that something has gone wrong. Maybe I can use some other data type to express these status codes if there is confusion.

leoppro (approved):

liuzix:
@amyangfei  PTAL

amyangfei (commented):
LGTM

ti-srebot (approved):

PR #1163, opened 2020-12-03
owner: fix owner doesn't clean up stale tasks

amyangfei:

Fix https://github.com/pingcap/ticdc/issues/1158

- TiCDC should clean the metadata of a changefeed as thoroughly as possible when removing a changefeed, but it is still possible to leave redundant metadata.
- Owner should choose whether to filter metadata when using it or clean up redundant data regularly, this PR chooses the second way.

Owner clean up the stale tasks at the beginning of each run round.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Fix a bug that outdated metadata could cause the newly created changefeed abnormal.

### DIFF
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -1100,7 +1100,14 @@ func (o *Owner) run(ctx context.Context) error {
 	o.l.Lock()
 	defer o.l.Unlock()
 
-	err := o.loadChangeFeeds(ctx)
+	var err error
+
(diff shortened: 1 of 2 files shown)


overvenus:
LGTM

leoppro:
LGTM

ti-srebot (approved):

PR #1166, opened 2020-12-03
tests: add leak test check in Makefile, complete leak tests

amyangfei:

Some unit tests missed leak test

- Add a check in `make check` to ensure leak test is added in all unit tests
- Fix some routine leak in unit tests

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/scripts/add-leaktest.sh
+++ b/scripts/add-leaktest.sh
@@ -29,6 +29,11 @@ for i in $@; do
     goimports -w $i
     gofmt -w $i
   fi
-echo $i
   rm -f $i~
 done
+
+if [ `git --no-pager diff | wc -c` -ne 0 ]; then
+  echo "Please ensure leak test is added in all unit tests"
+  git --no-pager diff
+  exit 1
+fi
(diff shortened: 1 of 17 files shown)


scripts/add-leaktest.sh
```diff
@@ -29,6 +29,11 @@ for i in $@; do
     goimports -w $i
     gofmt -w $i
   fi
-echo $i
   rm -f $i~
 done
+
+if [ `git --no-pager diff | wc -c` -ne 0 ]; then
```
overvenus, line 35: What about `git diff --exit-code`?

overvenus (commented):
Rest LGTM

scripts/add-leaktest.sh
amyangfei: addressed

overvenus:
LGTM

leoppro (approved):

ti-srebot (approved):

PR #1172, opened 2020-12-07
pipeline: refine pipeline, refactor processor part one

leoppro:
This PR is the part one of refactor processor.
refine pipeline framework

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


pkg/context/context.go
```diff
@@ -0,0 +1,90 @@
...
+	Config        *config.ReplicaConfig
+}
+
+// Context contains Vars(), Done(), Throw(error) and StdContext() context.Context
+// Context is used to instead of standard context
+type Context interface {
+	Vars() *Vars
+
+	Done() <-chan struct{}
+	Throw(error)
```
amyangfei, line 38: Please add some comments about what this function does, whether it is reentrant.

leoppro, line 38:
```suggestion
	// Vars return the `Vars` store by the root context created by `NewContext`
	// Note that the `Vars` should be READ-ONLY
	// The root node and all its children node share one point of `Vars`
	// So any modification of `Vars` will cause all other family nodes to change.
	Vars() *Vars

	Done() <-chan struct{}
	Throw(error)
```

pkg/context/context.go
```diff
@@ -0,0 +1,90 @@
...
+	StdContext() context.Context
```
leoppro, line 39:
```suggestion
	// StdContext return a simple struct implement the stdcontext.Context interface
	// The Context in this package and the StdContext returned by this function have the same life cycle
	StdContext() context.Context
```

pkg/context/context.go
```diff
@@ -0,0 +1,105 @@
...
+	Done() <-chan struct{}
+
+	// Throw an error to parents nodes
+	// we can using `WatchThrow` to listen the errors thrown by children nodes
+	Throw(error)
+
+	// StdContext return a simple struct implement the stdcontext.Context interface
+	// The Context in this package and the StdContext returned by this function have the same life cycle
+	// It means the `StdContext.Done()` will done when the `Context` is done.
+	StdContext() context.Context
```
overvenus, line 54: Why not  inherit `context.Context`?

```go
type Context interface {
  context.Context
  // ...
}
```

pkg/context/context.go
```diff
@@ -0,0 +1,105 @@
...
+	pd "github.com/tikv/pd/client"
+)
+
+// Vars contains some vars which can be used to anywhere in a pipeline
+type Vars struct {
+	// TODO add more vars
+	PDClient      pd.Client
+	SchemaStorage *entry.SchemaStorage
+	Config        *config.ReplicaConfig
+}
```
overvenus, line 30: So they all must be thread-safe?

pkg/context/context.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Context contains Vars(), Done(), Throw(error) and StdContext() context.Context
+// Context is used to instead of standard context
```
overvenus, line 33: Please add some comments about thread-safety.

pkg/pipeline/node.go
```diff
@@ -15,32 +15,21 @@ package pipeline
 
 // Node represents a handle unit for the message stream in pipeline
 type Node interface {
```
overvenus, line 17: Please add some comments about thread-safety.

pkg/context/context.go
leoppro: because I don't want to export the `Err()`, `Deadline()` and `Value()` function.

pkg/context/context.go
leoppro: yes, READ-ONLY and THREAD-SAFE

pkg/context/context.go
```diff
@@ -0,0 +1,105 @@
...
+type Context interface {
+
+	// Vars return the `Vars` store by the root context created by `NewContext`
+	// Note that the `Vars` should be READ-ONLY
```
leoppro, line 37:
```suggestion
	// Note that the `Vars` should be READ-ONLY and THREAD-SAFE
```

pkg/context/context.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"context"
+
+	"github.com/pingcap/ticdc/cdc/entry"
+	"github.com/pingcap/ticdc/pkg/config"
+	pd "github.com/tikv/pd/client"
+)
+
+// Vars contains some vars which can be used to anywhere in a pipeline
```
leoppro, line 24:
```suggestion
// Vars contains some vars which can be used to anywhere in a pipeline
// All field in Vars should be READ-ONLY and THREAD-SAFE
```

pkg/pipeline/node.go
```diff
@@ -15,32 +15,21 @@ package pipeline
 
 // Node represents a handle unit for the message stream in pipeline
```
leoppro, line 16:
```suggestion
// Node represents a handle unit for the message stream in the pipeline
// The following function in this interface will be called in one goroutine.
// It's NO NEED to consider concurrency issues
```

pkg/pipeline/node.go
```diff
@@ -13,34 +13,25 @@
 
 package pipeline
 
-// Node represents a handle unit for the message stream in pipeline
+// Node represents a handle unit for the message stream in the pipeline
+// The following function in this interface will be called in one goroutine.
```
leoppro, line 17:
```suggestion
// The following functions in this interface will be called in one goroutine.
```

pkg/context/context.go
```diff
@@ -0,0 +1,106 @@
...
+	f func(error)
+}
+
+// WatchThrow creates a new context that can watch the Throw function
+func WatchThrow(ctx Context, f func(error)) Context {
+	return &throwContext{
+		Context: ctx,
+		f:       f,
+	}
+}
```
amyangfei, line 98: Is the behavior more like `register callback function for thrown error`?

pkg/pipeline/message.go
```diff
@@ -21,62 +21,52 @@ type MessageType int
...
 	// Tp is the type of Message
 	Tp MessageType
-	// Lifecycle represents the message about the lifecycle
-	Lifecycle *LifecycleMessage
-	// RowChangedEvent represents the row change event
-	RowChangedEvent *model.RowChangedEvent
+	// Command is the command in this message
+	Command *Command
+	// PolymorphicEvent represents the row change event
+	PolymorphicEvent *model.PolymorphicEvent
```
amyangfei, line 38: I noticed `*model.PolymorphicEvent` type is used here, which may cause the library less universal. If we use an interface here, more assertation overhead may be introduced. Just curious about the design mode, not need to address.

pkg/context/context.go
leoppro: yes

pkg/pipeline/message.go
leoppro: the pipeline framework is designed for processor mainly, so I didn't  using `interface{}` here

amyangfei: If it is used for processor only, I prefer moving it to another package

leoppro: Where do you think better?

pkg/context/context.go
amyangfei: Will `WatchThrow` be changed?

pkg/pipeline/message.go
amyangfei: some path like `github.com/pingcap/ticdc/cdc/xxx`

amyangfei: Not a big problem

pkg/context/context.go
leoppro: rename it to `WithErrorHandler`

amyangfei (commented):
LGTM

liuzix:
LGTM

ti-srebot (approved):

Issue #1169, opened 2020-12-04
TiCDC pkg/notifier can create a new receiver even after the notifier has been closed


amyangfei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

   Add unit test cases as following

```golang
func (s *notifySuite) TestNewReceiverWithClosedNotifier(c *check.C) {
	defer testleak.AfterTest(c)()
	notifier := new(Notifier)
	notifier.Close()
	notifier.NewReceiver(50 * time.Millisecond)
}
```

This scenario could happen in a concurrent environment, where we close the notifier and create a new receiver from the notifier at the same time.

2. What did you expect to see?

    test runs successfully. The notifier should refuse to create a new receiver or return an error when it has been closed.

3. What did you see instead?

```
=== RUN   Test

----------------------------------------------------------------------
FAIL: notify_test.go:104: notifySuite.TestNewReceiverWithClosedNotifier

notify_test.go:110:
    ...
/home/apple/work/code/pingcap/ticdc/pkg/util/testleak/leaktest.go:154:
    c.Errorf("Test %s check-count %d appears to have leaked: %v", c.TestName(), cnt, g)
... Error: Test notifySuite.TestNewReceiverWithClosedNotifier check-count 50 appears to have leaked: github.com/pingcap/ticdc/pkg/notify.(*Receiver).signalTickLoop.func1(0xc00009d140)
        /home/apple/work/code/pingcap/ticdc/pkg/notify/notify.go:64 +0xb5
created by github.com/pingcap/ticdc/pkg/notify.(*Receiver).signalTickLoop
        /home/apple/work/code/pingcap/ticdc/pkg/notify/notify.go:61 +0x3f

OOPS: 0 passed, 1 FAILED
--- FAIL: Test (3.29s)
FAIL
exit status 1
FAIL    github.com/pingcap/ticdc/pkg/notify     3.290s
```

4. Versions of the cluster

    - TiCDC version (execute `cdc version`):

        ```
        0f71b071fa4cb7473984809e33ca94bc8b8b1981
        ```

PR #1177, opened 2020-12-07
tests: add unit tests for mysql sink

amyangfei:

Add unit tests for MySQL sink

- Fix a bug that cyclic replication doesn't relax SQL mode.
- use sql mock to test MySQL sink

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/sink/mysql.go
+++ b/cdc/sink/mysql.go
@@ -252,7 +252,7 @@ func (s *mysqlSink) execDDL(ctx context.Context, ddl *model.DDLEvent) error {
 func (s *mysqlSink) adjustSQLMode(ctx context.Context) error {
 	// Must relax sql mode to support cyclic replication, as downstream may have
 	// extra columns (not null and no default value).
-	if s.cyclic == nil {
+	if s.cyclic == nil || !s.cyclic.Enabled() {
 		return nil
 	}
 	var oldMode, newMode string
--- a/cdc/sink/mysql.go
+++ b/cdc/sink/mysql.go
@@ -252,7 +252,7 @@ func (s *mysqlSink) execDDL(ctx context.Context, ddl *model.DDLEvent) error {
 func (s *mysqlSink) adjustSQLMode(ctx context.Context) error {
 	// Must relax sql mode to support cyclic replication, as downstream may have
 	// extra columns (not null and no default value).
-	if s.cyclic != nil {
+	if s.cyclic == nil {
 		return nil
 	}
 	var oldMode, newMode string
--- a/cdc/sink/mysql.go
+++ b/cdc/sink/mysql.go
@@ -467,6 +467,22 @@ func parseSinkURI(ctx context.Context, sinkURI *url.URL, opts map[string]string)
 	return params, nil
 }
 
+var getDBConnImpl = getDBConn
+
+func getDBConn(ctx context.Context, dsnStr string) (*sql.DB, error) {
+	db, err := sql.Open("mysql", dsnStr)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


cdc/sink/mysql.go
```diff
@@ -511,26 +524,19 @@ func newMySQLSink(
 	if params.timezone != "" {
 		dsn.Params["time_zone"] = params.timezone
 	}
-	testDB, err := sql.Open("mysql", dsn.FormatDSN())
+	testDB, err := getDBConnImpl(ctx, dsn.FormatDSN())
 	if err != nil {
-		return nil, errors.Annotate(
-			cerror.WrapError(cerror.ErrMySQLConnectionError, err), "fail to open MySQL connection when configuring sink")
+		return nil, err
```
overvenus, line 529:
```suggestion
		return nil, errors.Trace(err)
```

cdc/sink/mysql.go
```diff
@@ -511,26 +524,19 @@ func newMySQLSink(
...
+	db, err := getDBConnImpl(ctx, dsnStr)
 	if err != nil {
-		return nil, errors.Annotate(
-			cerror.WrapError(cerror.ErrMySQLConnectionError, err), "Open database connection failed")
-	}
-	err = db.PingContext(ctx)
-	if err != nil {
-		return nil, errors.Annotate(
-			cerror.WrapError(cerror.ErrMySQLConnectionError, err), "fail to open MySQL connection")
+		return nil, err
```
overvenus, line 539:
```suggestion
		return nil, errors.Trace(err)
```

cdc/sink/mysql_test.go
```diff
@@ -960,6 +966,289 @@ func (s MySQLSinkSuite) TestCheckTiDBVariable(c *check.C) {
...
+	c.Assert(err, check.IsNil)
+	cyclicConfig, err := rc.Cyclic.Marshal()
+	c.Assert(err, check.IsNil)
+	opts := map[string]string{
+		mark.OptCyclicConfig: cyclicConfig,
+	}
+	sink, err := newMySQLSink(ctx, changefeed, sinkURI, f, rc, opts)
+	c.Assert(err, check.IsNil)
+
+	// TODO: remove this line after https://github.com/pingcap/ticdc/issues/1169 is fixed
```
overvenus, line 1039: Cc #1169, a TODO is left.

overvenus (commented):
Rest LGTM

cdc/sink/mysql.go
amyangfei: err has contained error trace, `errors.Trace` is not necessary here

cdc/sink/mysql.go
amyangfei: ditto

cdc/sink/mysql.go
overvenus: Yes, though I think it would idiomatic to wrap errors as suggests from wrapcheck linter https://golangci-lint.run/usage/linters/#:~:text=wrapcheck

overvenus:
LGTM

cdc/sink/mysql.go
amyangfei: Maybe not suitable for the current error mechanism. Currently all TiCDC errors should be under the TiDB error RFC, which means
- error from an external package should be wrapped by an RFC error generator
- error generated by TiCDC internal logic must be generated from an RFC error generator

leoppro (approved):

ti-srebot (approved):

PR #1190, opened 2020-12-09
processor: fix task status flushed too many before table is initialized

amyangfei:

Fix https://github.com/pingcap/ticdc/issues/1188. This is a hotfix for current codebase, we should process with the `Dirty` field of task status carefully.

change flush check condition

### Check List 

Tests 

 - Unit test
 - Integration test

   - hack puller `p.tsTracker.Frontier()` to force no forward to simulate long time incremental scan.
   - watch the task status key version in etcd doesn't change.

### Release note

- No release note

### DIFF
--- a/cdc/model/owner.go
+++ b/cdc/model/owner.go
@@ -209,6 +209,8 @@ type TaskStatus struct {
 	Operation    map[TableID]*TableOperation   `json:"operation"`
 	AdminJobType AdminJobType                  `json:"admin-job-type"`
 	ModRevision  int64                         `json:"-"`
+	// true means Operation record has been changed
+	Dirty bool `json:"-"`
 }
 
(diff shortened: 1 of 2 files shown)


amyangfei:
https://github.com/pingcap/ticdc/pull/1190

overvenus:
LGTM

liuzix:
LGTM

ti-srebot (approved):

liuzix:
Just for your reference, once we adapt the Owner to use the EtcdWorker, such problems are guaranteed to no longer exist.

amyangfei:
> Just for your reference, once we adapt the Owner to use the EtcdWorker, such problems are guaranteed to no longer exist.

I don’t think so, this is not an owner’s problem, it is the way how task status is observation and updating

PR #1194, opened 2020-12-10
tests: turn on the tidb_enable_change_multi_schema switch in some tests

amyangfei:

https://github.com/pingcap/tidb/pull/21533 disables multi schema change feature by default, which leads to some tests failure caused by `ERROR 8200 (HY000) at line 12: Unsupported multi schema change`

Turn on the `tidb_enable_change_multi_schema` switch in tests where multi schema change happens

Note cherry-pick just wants to keep the test code same as far as possible

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


leoppro:
LGTM

liuzix:
LGTM

ti-srebot (approved):

PR #1218, opened 2020-12-16
cpature: fix session done is not considered during handling task event

amyangfei:

Fix https://github.com/pingcap/ticdc/issues/1170

If a capture's etcd session is disconnected, it should suicide and recover asap. However we have a selection of two channels, one is checking `etcd session done` and the other one is `handle task event`.
When an error happens after handling one task event, check whether the etcd session is already invalid, if it is disconnected, just ignore the error and suicide and recover the capture internal, instead of the process level exit.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Fix a bug that CDC could exit unexpected when etcd session is disconnected

### DIFF
--- a/cdc/capture.go
+++ b/cdc/capture.go
@@ -167,9 +167,7 @@ func (c *Capture) Run(ctx context.Context) (err error) {
 			if ev.Err != nil {
 				return errors.Trace(ev.Err)
 			}
-			log.Info("before capture handle task")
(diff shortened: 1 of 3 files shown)


overvenus:
LGTM

leoppro:
LGTM

ti-srebot (approved):

PR #1247, opened 2020-12-29
*: fix the output in changefeed out of order

leoppro:

fix the output in changefeed out of order

add a middleware between sorter and sink, make the life cycle of sink managed by processor.

before this pr:
image

after this pr:
image

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- fix the output in changefeed out of order


cdc/processor.go
```diff
@@ -1170,24 +1030,57 @@ func (p *processor) sorterConsume(
...
+				if errors.Cause(err) != context.Canceled {
+					p.errCh <- errors.Trace(err)
+				}
+				return
+			}
+		case <-globalResolvedTsReceiver.C:
+			localResolvedTs := atomic.LoadUint64(&p.localResolvedTs)
+			globalResolvedTs := atomic.LoadUint64(&p.globalResolvedTs)
+			var minTs uint64
+			if localResolvedTs < globalResolvedTs {
```
amyangfei, line 1057: In which scenario will `localResolvedTs < globalResolvedTs` happen, do we need to add a warning log here

cdc/sink/manager.go
```diff
@@ -0,0 +1,160 @@
...
+		atomic.StoreUint64(&t.emittedTs, resolvedTs)
+		t.manager.tableSinksMu.Lock()
+		defer t.manager.tableSinksMu.Unlock()
+		return t.manager.flushBackendSink(ctx)
+	}
+	resolvedRows := t.buffer[:i]
+	t.buffer = t.buffer[i:]
+
+	t.manager.tableSinksMu.Lock()
+	defer t.manager.tableSinksMu.Unlock()
```
amyangfei, line 139: Seems the `EmitRowChangedEvents` and `FlushRowChangedEvents` of `backendSink` are not thread-safe, will there be too much lock contention if there are many tables?

cdc/processor.go
```diff
@@ -613,19 +589,20 @@ func (p *processor) handleTables(ctx context.Context, status *model.TaskStatus)
...
-				if stopped {
-					opt.BoundaryTs = checkpointTs
-					if checkpointTs <= p.position.CheckPointTs {
-						tablesToRemove = append(tablesToRemove, tableID)
-						opt.Done = true
-						opt.Status = model.OperFinished
-					}
-					status.Dirty = true
+				if opt.BoundaryTs != p.position.CheckPointTs {
+					log.Panic("panic")
```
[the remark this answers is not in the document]
leoppro: It's a debug code, I will remove it soon

cdc/sink/manager.go
leoppro: In the original implementation, the output streams of multiple tables were pushed into a single channel(processor.outputCh), the current implementation is essentially the same as the original, with sinks being single-threaded

cdc/processor.go
leoppro: added a warn log

cdc/sink/manager.go
amyangfei: Locked time is different.
- The channel way 
- comparing to the current lock mechanism

cdc/changefeed.go
```diff
@@ -512,28 +518,27 @@ func (c *changeFeed) handleMoveTableJobs(ctx context.Context, captures map[model
...
 			}
-			replicaInfo, exist := status.RemoveTable(tableID, c.status.CheckpointTs)
+			replicaInfo, exist := status.RemoveTable(tableID, c.status.ResolvedTs)
 			if !exist {
 				delete(c.moveTableJobs, tableID)
 				log.Warn("ignored the move job, the table is not exist in the source capture", zap.Reflect("job", job))
 				continue
 			}
-			replicaInfo.StartTs = c.status.CheckpointTs
+			replicaInfo.StartTs = c.status.ResolvedTs
```
amyangfei, line 527: Please add comments about why using `c.status.ResolvedTs`

overvenus:
Could you update `docs/data-flow.dot` too?

cdc/sink/manager.go
```diff
@@ -0,0 +1,242 @@
...
+	return nil
+}
+
+type bufferSink struct {
+	Sink
+	buffer chan struct {
+		rows       []*model.RowChangedEvent
+		resolved   bool
+		resolvedTs model.Ts
+	}
```
overvenus, line 174: Could you define the type?

cdc/sink/manager.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (m *Manager) getCheckpointTs() uint64 {
+	return atomic.LoadUint64(&m.checkpointTs)
+}
+
+type tableSink struct {
+	tableID   model.TableID
+	manager   *Manager
+	buffer    []*model.RowChangedEvent
+	emittedTs model.Ts
```
overvenus, line 117: Could you add some comments? Before update `emittedTs`, does it acquire receiving an ack from downstream?

cdc/sink/mq.go
```diff
@@ -166,6 +166,7 @@ func (k *mqSink) EmitRowChangedEvents(ctx context.Context, rows ...*model.RowCha
 			continue
 		}
 		partition := k.dispatcher.Dispatch(row)
+		log.Debug("LEOPPRO show row in mq sink", zap.Int32("partition", partition), zap.Reflect("row", row))
```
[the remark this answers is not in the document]
leoppro: removed

cdc/sink/manager.go
leoppro: I think we needn'd define a independent type for it
This coding style is clearer, easier to understand, and I think not very wordy. It's like `Result<Rows,Ts>` in rust.

overvenus:
Please update ticdc-grafana-dashboard.md

cdc/sink/metrics.go
```diff
@@ -70,6 +70,21 @@ var (
...
+			Help:      "Bucketed histogram of processing time (s) of flushing events in processor",
+			Buckets:   prometheus.ExponentialBuckets(0.002 /* 2ms */, 2, 20),
+		}, []string{"capture", "changefeed"})
+	bufferChanSizeGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "ticdc",
+			Subsystem: "sink",
+			Name:      "buffer_chan_size",
+			Help:      "size of row changed event buffer channel in sink manager",
+		}, []string{"capture", "changefeed"})
```
overvenus, line 87: Is it added to dashboard?

cdc/metrics_processor.go
```diff
@@ -75,28 +75,13 @@ var (
 			Help:      "The time it took to update sub changefeed info.",
 			Buckets:   prometheus.ExponentialBuckets(0.001 /* 1 ms */, 2, 18),
 		}, []string{"capture"})
-	tableOutputChanSizeGauge = prometheus.NewGaugeVec(
-		prometheus.GaugeOpts{
-			Namespace: "ticdc",
-			Subsystem: "processor",
-			Name:      "txn_output_chan_size",
-			Help:      "size of row changed event output channel from table to processor",
-		}, []string{"changefeed", "capture"})
```
overvenus, line 84: Have you remove it from dashboard?

overvenus (commented):
Why does ticdc.json change so much?

cdc/sink/metrics.go
leoppro: yes

cdc/metrics_processor.go
leoppro: yes, I renamed txn_output_chan_size to buffer_chan_size

amyangfei:
There could exist an optimization in data flow dot, the role count of upstream and downstream in each data flow link is not displayed.
For example

`Table Sink (N)` -> `Buffer Sink (1)` -> `Backend Sink(1)`

cdc/sink/metrics.go
overvenus: Do you add it to ticdc-grafana-dashboard.md?

cdc/metrics_processor.go
overvenus: Does ticdc-grafana-dashboard.md reflect the change?

leoppro: yes

amyangfei:
Some unchanged fields are changed in `metrics/grafana/ticdc.json`, can we just update the contents of changed panels

cdc/sink/manager.go
```diff
@@ -0,0 +1,267 @@
...
+				errCh <- err
+			}
+			return
+		case e := <-b.buffer:
+			if e.resolved {
+				start := time.Now()
+				checkpointTs, err := b.Sink.FlushRowChangedEvents(ctx, e.resolvedTs)
+				if err != nil {
+					if errors.Cause(err) != context.Canceled {
+						errCh <- err
```
amyangfei, line 216: `SinkManager` shares the same `errCh` with backend sink, in some extreme scenarios, both of them may try to send error to this channel, then one of them will hang

cdc/sink/manager.go
```diff
@@ -0,0 +1,267 @@
...
+func (t *tableSink) Close() error {
+	t.manager.destroyTableSink(t.tableID)
+	return nil
+}
+
+type bufferSink struct {
+	Sink
+	buffer chan struct {
+		rows       []*model.RowChangedEvent
+		resolved   bool
```
amyangfei, line 177: We can remove the `resolved` field, and check whether it is a resolved event by `rows == nil`

cdc/sink/manager.go
```diff
@@ -0,0 +1,267 @@
...
+
+				dur := time.Since(start)
+				metricFlushDuration.Observe(dur.Seconds())
+				if dur > 3*time.Second {
+					log.Warn("flush row changed events too slow",
+						zap.Duration("duration", dur), util.ZapFieldChangefeed(ctx))
+				}
+				continue
+			}
+			err := b.Sink.EmitRowChangedEvents(ctx, e.rows...)
```
amyangfei, line 230: Do we need to monitor the duration of `EmitRowChangedEvents` too? If so, may be we can resue the metric `flushRowChangedDuration` by adding one more label

amyangfei:
Recommend adding more unit tests to cover two secnarios
- cancel a table puller or remove a table sink from sink manager
- error happens in `EmitRowChangedEvents` or `FlushRowChangedEvents` of the backend sink and sink manager can handle error properly

amyangfei (commented):
LGTM

liuzix:
LGTM

ti-srebot (approved):

Issue #1292, opened 2021-01-11
Improve unit tests of kv client


amyangfei:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Since the stability and performance of the kv client have a great impact on TiCDC service, we need to add more tests to make it robust and efficient enough.

**Describe the feature you'd like:**

- Improve the unit test coverage of kv client to 85%
- Correctness guarantee: The kv client must handle all kinds of region behaviors correctly, which must be tested with unit tests, currently test coverage is still low and several key logic paths are not covered.
  - [ ] Region split
  - [ ] Region random merge
  - [ ] Region leader transfer
  - Scenarios with region cache.
    - [ ] Request for cache missed region
    - [ ] Request for unreachable store, region cache automatically rebalance
    - [ ] Meets RegionNotFound error, which means region cache is outdated
    - [ ] Meets EpochNotMatch error
    - [ ] GetTiKVRPCContext failed
  - Other kv event error
    - [ ] DuplicateRequest
    - [ ] Empty or unknown error
  - gRPC stream error
    - [ ] Send error
    - [ ] Recv error
  - [ ] Lock resolver work logic
  - Some code logic that will raise a warning
    - [ ] drop event due to event belongs to a stale request
    - [ ] drop event due to region feed stopped
    - [ ] drop resolved ts due to region feed stopped
  - [ ] Normal kv events handling

- Performance measurement: repeatable benchmark is not available with kv client now. We need a detailed performance measurement to evaluate the optimization.

amyangfei:
The feature has been GA, and expected unit tests have been added.

PR #1298, opened 2021-01-12
tests: test TiKV is not compatible with TiCDC to simulate rolling update

amyangfei:

Test TiCDC new request to TiKV meets `ErrVersionIncompatible` error (in fact this error is raised before EventFeed API is really called), TiCDC will wait for 20s and then retry. This is a common scenario when rolling upgrade a cluster and the new version is not compatible with the old version

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


cdc/kv/client_test.go
```diff
@@ -749,26 +763,106 @@ func (s *etcdSuite) TestHandleFeedEvent(c *check.C) {
...
 }
 
-// TODO enable the test
-func (s *etcdSuite) TodoTestIncompatibleTiKV(c *check.C) {
+// TestIncompatibleTiKV tests TiCDC new request to TiKV meets `ErrVersionIncompatible`
+// error (in fact this error is raised before EventFeed API is really called),
+// TiCDC will wait 20s and then retry. This is a common scenario when rolling
+// upgrade a cluster and the new version is not compatible with the old version
+// (upgrade TiCDC before TiKV, since upgrade TiKV often takes much longer).
+func (s *etcdSuite) TestIncompatibleTiKV(c *check.C) {
```
overvenus, line 771: Thank you for enabling the test!

cdc/kv/client_test.go
```diff
@@ -749,26 +763,106 @@ func (s *etcdSuite) TestHandleFeedEvent(c *check.C) {
...
+
+	call := int32(0)
+	// 20 here not too much, since check version itself has 3 time retry, and
+	// region cache could also call get store API, which will trigger version
+	// generator too.
+	versionGenCallBoundary := int32(20)
+	gen := func() string {
+		atomic.AddInt32(&call, 1)
+		if atomic.LoadInt32(&call) < versionGenCallBoundary {
+			return "v2.1.0" /* CDC is not compatible with 2.1.0 */
```
overvenus, line 785: Could you also test v3.0.x and v3.1.x?

overvenus (commented):
Rest LGTM

liuzix:
LGTM

cdc/kv/client_test.go
amyangfei: addressed, ptal

leoppro (approved):

ti-srebot (approved):

PR #1325, opened 2021-01-21
tests: kv client should handle event with invalid request id

amyangfei:

- Test kv client should drop an event if its request-id is outdated
- Test kv client should return an error when receiving a new subscription but the corresponding region is not found in pending regions

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


leoppro:
LGTM

liuzix:
LGTM

ti-srebot (approved):

Issue #1258, opened 2020-12-31
checkpointTs invariant violation after CREATE TABLE


liuzix:
We generally assume that the global checkpointTs is less than or equal to the local checkpointTs of each capture. This invariant can be violated due to the fact that the owner delays flushing changefeed status because of `flushChangefeedInterval`. Consider the following scenario:

- Owner recalculates global checkpointTs to `B`. Previously the global checkpointTs was `A`. Note that `A < B`.
- Owner tries to flush changefeed status, but because of `flushChangefeedInterval`, this operation is skipped. The global checkpointTs in Etcd remains `A`.
- Owner executes DDL and dispatches table to a capture where the processor has not yet been run.
- The processor is launched, and the startTs is the global checkpointTs in Etcd, i.e. `A`.
- Owner recalculates global checkpointTs. It is still `B`, because checkpoint regression is not permitted.
- Owner flushes changefeed status. Now the Etcd status has global checkpointTs as `B`, but a local checkpointTs is `A`. Note that `A < B`. This is an invariant violation.

The consequences of this violation is under investigation.

PR #1236, opened 2020-12-24
owner: prevent table start-ts regressions

liuzix:
### What problem does this PR solve? 
- The start-ts of a table could regress if the table is added to a capture whose current checkpoint-ts is less than the start-ts of said table, and before the checkpoint-ts catches up, the capture exits abnormally. This regression, in rare cases, could cause the start-ts to regress so far that it is earlier than the global checkpoint-ts. When this happens, SchemaStorage related errors occur. #1258 #1235 

### What is changed and how it works?
- Fixed several bugs in Owner so that never will a local checkpoint be less than the global checkpoint.
- Cleaned up checkpoint ts semantics with respect to DDL operations, so that a DDL is finished **if and only if** `global checkpoint >= FinishTs`.
- Fixed a bug in processor introduced by #1247 which causes local checkpoint regression.
- Added Etcd state checker to integration test.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

 - Need to cherry-pick to the release branch

### Release note

- Fix bugs in the owner that could lead to abnormal CDC server exits during table migrations.

Issue #1235, opened 2020-12-24
meet ErrSchemaSnapshotNotFound error in TiCDC 4.0.9


amyangfei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

    changefeed meets error

2. What did you expect to see?

3. What did you see instead?

```
[2020/12/24 08:32:44.792 +00:00] [ERROR] [processor.go:1266] ["error on running processor"] [captureid=e42540a4-6bcf-4b8e-bc53-80bf02a6f71e] [captureaddr=tidb-cluster01-ticdc-5.tidb-cluster01-ticdc-peer.tidb-cluster01.svc:8301] [changefeedid=user-common-offline-5] [processorid=7deca114-d1da-4036-b20a-36b3dd75cb41] [error="[CDC:ErrSchemaSnapshotNotFound]can not found schema snapshot, ts: 421736913222762923"] [errorVerbose="[CDC:ErrSchemaSnapshotNotFound]can not found schema snapshot, ts: 421736913222762923
github.com/pingcap/errors.AddStack
        github.com/pingcap/errors@v0.11.5-0.20201029093017-5a7df2af2ac7/errors.go:174
github.com/pingcap/errors.(*Error).GenWithStackByArgs
        github.com/pingcap/errors@v0.11.5-0.20201029093017-5a7df2af2ac7/normalize.go:156
github.com/pingcap/ticdc/cdc/entry.(*SchemaStorage).getSnapshot
        github.com/pingcap/ticdc/cdc/entry/schema_storage.go:684
github.com/pingcap/ticdc/cdc/entry.(*SchemaStorage).GetSnapshot.func1
        github.com/pingcap/ticdc/cdc/entry/schema_storage.go:702
github.com/pingcap/ticdc/pkg/retry.RunWithInfiniteRetry.func1
        github.com/pingcap/ticdc/pkg/retry/retry.go:50
github.com/cenkalti/backoff.RetryNotify
        github.com/cenkalti/backoff@v2.2.1+incompatible/retry.go:37
github.com/cenkalti/backoff.Retry
        github.com/cenkalti/backoff@v2.2.1+incompatible/retry.go:24
github.com/pingcap/ticdc/pkg/retry.RunWithInfiniteRetry
        github.com/pingcap/ticdc/pkg/retry/retry.go:49
[...]

PR #1323, opened 2021-01-21
*: fix the output in changefeed out of order (#1247)

ti-srebot:
cherry-pick #1247 to release-4.0
You can switch your code base to this Pull Request by using [git-extras](https://github.com/tj/git-extras):
```bash
# In ticdc repo:
git pr https://github.com/pingcap/ticdc/pull/1323
```

After apply modifications, you can push your change to this PR via:
```bash
git push git@github.com:ti-srebot/ticdc.git pr/1323:release-4.0-5c62e67750c0
```

---

### What problem does this PR solve? 

fix the output in changefeed out of order

### What is changed and how it works?
add a middleware between sorter and sink, make the life cycle of sink managed by processor.

before this pr:
image

after this pr:
image

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- fix the output in changefeed out of order

PR #1331, opened 2021-01-22
owner: prevent table start-ts regressions (#1236)

ti-srebot:
cherry-pick #1236 to release-4.0
You can switch your code base to this Pull Request by using [git-extras](https://github.com/tj/git-extras):
```bash
# In ticdc repo:
git pr https://github.com/pingcap/ticdc/pull/1331
```

After apply modifications, you can push your change to this PR via:
```bash
git push git@github.com:ti-srebot/ticdc.git pr/1331:release-4.0-78de95e14e1f
```

---

- The start-ts of a table could regress if the table is added to a capture whose current checkpoint-ts is less than the start-ts of said table, and before the checkpoint-ts catches up, the capture exits abnormally. This regression, in rare cases, could cause the start-ts to regress so far that it is earlier than the global checkpoint-ts. When this happens, SchemaStorage related errors occur. #1258 #1235 

- Fixed several bugs in Owner so that never will a local checkpoint be less than the global checkpoint.
- Cleaned up checkpoint ts semantics with respect to DDL operations, so that a DDL is finished **if and only if** `global checkpoint >= FinishTs`.
- Fixed a bug in processor introduced by #1247 which causes local checkpoint regression.
- Added Etcd state checker to integration test.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Fix bugs in the owner that could lead to abnormal CDC server exits during table migrations.


liuzix:
This cherry-pick depends on #1323

amyangfei:
> This cherry-pick depends on #1323

PTAL @liuzix

liuzix:
Will cherry pick manually

PR #1336, opened 2021-01-25
kv-client: fix stream is not recycled when meeting recv error

amyangfei:

- Find a but that gRPC stream is not recycled correctly in some scenarios
   - if `stream.Recv` with error, but the underlying stream is normal actually (or the next round of `stream.Send` is successfully, then the `receiveFromStream` goroutine will be never created, and corresponding events will not be consumed by kv client.
- Improve unit test of kv client, part of #1292

- Close stream before `receiveFromStream` routine exits, and remove stream from stream cache
- Add unit test to cover this scenario

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Fix a bug that kv client may skip to recreate event feed receiving routine by accident.


leoppro (approved):

amyangfei:
PTAL @liuzix

liuzix:
LGTM

ti-srebot (approved):

PR #1352, opened 2021-01-27
owner: prevent table start-ts regressions (#1236) 

liuzix:
- The start-ts of a table could regress if the table is added to a capture whose current checkpoint-ts is less than the start-ts of said table, and before the checkpoint-ts catches up, the capture exits abnormally. This regression, in rare cases, could cause the start-ts to regress so far that it is earlier than the global checkpoint-ts. When this happens, SchemaStorage related errors occur. #1258 #1235 

- Fixed several bugs in Owner so that never will a local checkpoint be less than the global checkpoint.
- Cleaned up checkpoint ts semantics with respect to DDL operations, so that a DDL is finished **if and only if** `global checkpoint >= FinishTs`.
- Fixed a bug in processor introduced by #1247 which causes local checkpoint regression.
- Added Etcd state checker to integration test.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Fix bugs in the owner that could lead to abnormal CDC server exits during table migrations.


tests/clustered_index/run.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
amyangfei, line 1: `clustered index` is in TiDB 5.0 only, please remove this case

amyangfei (approved):

PR #1354, opened 2021-01-27
processor: a processor state based etcd worker, refactor processor part three

leoppro:

a processor state based etcd worker

### Check List 

Tests 

 - Unit test

### Release note

- No release note


cdc/processor/etcdkey.go
```diff
@@ -0,0 +1,147 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package processor
```
amyangfei, line 14: Is it better to put this file into another package, since the ower also uses these keys

cdc/processor/etcdkey.go
```diff
@@ -0,0 +1,147 @@
...
+func (k *CDCEtcdKey) Parse(key string) error {
+	if !strings.HasPrefix(key, etcdKeyBase) {
+		return cerror.ErrInvalidEtcdKey.GenWithStackByArgs(key)
+	}
+	key = key[len(etcdKeyBase):]
+	switch {
+	case strings.HasPrefix(key, ownerKey):
+		k.Tp = CDCEtcdKeyTypeOnwer
+		k.CaptureID = ""
+		k.ChangefeedID = ""
```
amyangfei, line 71: The empty value can be omitted

cdc/processor/etcdkey.go
```diff
@@ -0,0 +1,147 @@
...
+	CDCEtcdKeyTypeCapture
+	CDCEtcdKeyTypeChangefeedInfo
+	CDCEtcdKeyTypeChangeFeedStatus
+	CDCEtcdKeyTypeTaskPosition
+	CDCEtcdKeyTypeTaskStatus
+	CDCEtcdKeyTypeTaskWorkload
+)
+
+// CDCEtcdKey represents a etcd key which is defined by TiCDC
+type CDCEtcdKey struct {
```
amyangfei, line 54: How will this struct be used, supposing capture wants to write a key `"/tidb/cdc/capture/6bbc01c8-0605-4f86-a0f9-b3119109b225"`, should it be constructed from a raw string or generate a `CDCEtcdKey` struct. Please add some regulations about it.

cdc/processor/state.go
```diff
@@ -0,0 +1,262 @@
...
+)
+
+func (s *changefeedState) PatchTaskPosition(fn func(*model.TaskPosition) (*model.TaskPosition, error)) {
+	key := &CDCEtcdKey{
+		Tp:           CDCEtcdKeyTypeTaskPosition,
+		CaptureID:    s.CaptureID,
+		ChangefeedID: s.ID,
+	}
+	s.patchAny(key.String(), taskPositionTPI, func(e interface{}) (interface{}, error) {
+		if e == nil {
```
amyangfei, line 203: what does `e == nil` mean here, does it mean key has been deleted or something else

cdc/processor/state.go
```diff
@@ -0,0 +1,262 @@
...
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
+	"github.com/pingcap/ticdc/cdc/model"
+	"github.com/pingcap/ticdc/pkg/orchestrator"
+	"github.com/pingcap/ticdc/pkg/orchestrator/util"
+	"go.uber.org/zap"
+)
+
+type globalState struct {
```
amyangfei, line 28: Will this state be stored in etcd or in memory only

leoppro: in etcd.
state is a snapshot of etcd

cdc/processor/state.go
leoppro: e == nil means that the key is not exist before this patch

cdc/processor/etcdkey.go
leoppro: I am afraid that I can't agree with you.
We can't ensure whether the `k.CaptureID` is empty when parsing the raw key

amyangfei: It depends on how a `CDCEtcdKey` is initialized, and whether it can be changed from one type to another one?

Better to add some comments about how to use it

amyangfei (commented):
rest LGTM

cdc/processor/etcdkey.go
leoppro: I hope that the `Parse` function will set `CDCKey` to a deterministic state, no matter how the `CDCKey` initialized

amyangfei (commented):
LGTM

liuzix:
LGTM

ti-srebot (approved):

PR #1357, opened 2021-01-27
tests: check kv client can drop events after region feed is stopped

amyangfei:

- Test kv client should drop event due to region feed stopped
- Test kv client should drop resolved ts due to region feed stopped

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


leoppro (approved):

liuzix:
LGTM

ti-srebot (approved):

Issue #1392, opened 2021-02-02
log output optimization


amyangfei:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Some logs in cdc server are not diagnose oriented, such as 

- `[statistics.go:126] ["sink replication status"]` may log too many
- stream connect/disconnect log too many in kv client
- `warn` or `error` logs maybe not helpful to problem diagnosis

**Describe the feature you'd like:**

- [ ] remove unnecesary logs in cdc
- [ ] make `warn` and `error` log can expose more diagnose information

PR #1395, opened 2021-02-02
sink: increase statistics interval

amyangfei:

related to #1392 

Statistics is **changefeed independent**, it works more like a sink keepalive sign.

If TiCDC supports to replicate thousands of changefeeds in the future, the statistic way should be optimized. But currently, it is all right to keep the periodic statistics

Increase sink statistics interval to `10 minutes`, which will not be logged too much.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/sink/statistics.go
+++ b/cdc/sink/statistics.go
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	printStatusInterval  = 30 * time.Second
+	printStatusInterval  = 10 * time.Minute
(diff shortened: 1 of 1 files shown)


amyangfei:
PTAL @overvenus @leoppro @liuzix

overvenus (approved):
LGTM, thanks!

liuzix:
LGTM

ti-srebot (approved):

PR #1396, opened 2021-02-02
owner: cleanup checkpoint lag metric when owner exits

amyangfei:

If owner transfers, the metric `ticdc_owner_checkpoint_ts_lag` is still kept in the old owner, it is safe to remove them since new owner will maintain these metrics.

If we don't remove the outdated metric, we will see two values for the same metric, where the `max` operator will give an unexpected result in Grafana

cleanup checkpoint lag metric when owner exits

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -1143,6 +1143,8 @@ loop:
 	}
 	for _, cf := range o.changeFeeds {
 		cf.Close()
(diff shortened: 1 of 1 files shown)


liuzix:
LGTM

leoppro (approved):

ti-srebot (approved):

PR #1416, opened 2021-02-18
table_pipeline: add cyclic mark node and fix some bugs 

leoppro:
- add cyclic mark node for table pipeline
- fix some bugs
- add some metrics

Why I think that the cyclic mark filter should be in table piepline:

we refer to the logical about match cyclic mark rows and normal rows as match-process, refer to the logical about filter the marked rows as filter-process, refer to the logical about generate mark-sql and insert it into down stream as generate-mark-process.

In the current implementation, all three processes are in the same mysql sink module, which is bloated and inflexible. In fact, only the generate-mark-process is related to MySQL sink.

There is a difference between old and new processor: when many tables using the same mark table, the old processor will create only one mark table puller, but new processor will create more than one mark table pullers in the same case. this difference lead to the `outputCh` of the processor contains more than one mark table stream. and it's hard to match and filter in MySQL sink module

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/processor/pipeline/cyclic_mark.go
+++ b/cdc/processor/pipeline/cyclic_mark.go
@@ -23,6 +23,8 @@ import (
 	"go.uber.org/zap"
 )
 
+// cyclicMarkNode match the mark rows and normal rows, set the ReplicaID of normal rows and filter the mark rows
+// and filter the normal rows by the FilterReplicaID config item.
 type cyclicMarkNode struct {
 	localReplicaID  uint64
 	filterReplicaID map[uint64]struct{}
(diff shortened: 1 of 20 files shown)


cdc/processor/pipeline/puller.go
```diff
@@ -41,17 +44,19 @@ type pullerNode struct {
 }
 
 func newPullerNode(
+	changefeedID model.ChangeFeedID,
```
amyangfei, line 47: Can we put `changefeedID` into context Vars?

cdc/processor/pipeline/cyclic_mark.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
+	"github.com/pingcap/ticdc/cdc/entry"
+	"github.com/pingcap/ticdc/cdc/model"
+	"github.com/pingcap/ticdc/pkg/cyclic/mark"
+	"github.com/pingcap/ticdc/pkg/pipeline"
+	"go.uber.org/zap"
+)
+
+type cyclicMarkNode struct {
```
amyangfei, line 26: Please add some comments about what does `cyclicMarkNode` do and how it works

cdc/processor/pipeline/puller.go
leoppro: yes, I will refine ctx.Vars in next PR

cdc/processor/pipeline/cyclic_mark.go
leoppro: done

amyangfei (commented):
LGTM

liuzix:
LGTM

ti-srebot (approved):

PR #1418, opened 2021-02-18
kv-client: don't maintain need reload info for region

amyangfei:

TiCDC always interacts with region leader, every time something goes wrong, failed region will be reloaded via `BatchLoadRegionsWithKeyRange` API. So we don't need to force reload region any more.

**NOTE: current logic will always reload region information when meeting failure, in some extreme scenarios, such as TiKV region leader information is not uploaded to PD in time, the recover time with leader change will increase. This can be an optimization point.**

- `scheduleReload` is always false
- Related test cases such as change region leader, store failure etc. are already in unit tests

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -63,6 +63,11 @@ const (
 
 	// The threshold of warning a message is too large. TiKV split events into 6MB per-message.
 	warnRecvMsgSizeThreshold = 12 * 1024 * 1024
+
+	// TiCDC always interacts with region leader, every time something goes wrong,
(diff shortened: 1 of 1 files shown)


leoppro:
LGTM

overvenus (commented):
Can it be tested?

amyangfei:
> Can it be tested?

unit test covers the code logic

overvenus:
LGTM

ti-srebot (approved):

PR #1423, opened 2021-02-19
Restore cdc log failed due to excess version part in ddl file when using s3 sink

yuqi1129:
see https://github.com/pingcap/ticdc/issues/1422

only when `firstCreated` is true and previous log size is 0 will write `version` to ddl cdc log

### Check List 

Tests 

 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


cdc/sink/cdclog/s3.go
```diff
@@ -289,6 +284,26 @@ func (s *s3Sink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
 	if err != nil {
 		return cerror.WrapError(cerror.ErrS3SinkStorageAPI, err)
 	}
+
+	// only reboot and size = 0 should we add version to s3
+	withVersion := firstCreated && size == 0
```
amyangfei, line 289: If `size > maxDDLFlushSize`, should the `withVersion` be true?

yuqi1129: Oh, you are right, maxDDLFlushSize should also be considered

amyangfei (commented):
LGTM

amyangfei:
PTAL @leoppro @liuzix

liuzix:
@3pointer PTAL. BTW is it possible to add some tests for this case?

yuqi1129:
> @3pointer PTAL. BTW is it possible to add some tests for this case?

What's your opinion?

yuqi1129:
Similar test case, please see PR https://github.com/pingcap/ticdc/pull/1515/files

overvenus:
LGTM, thanks!

yuqi1129:
Can anyone help to merge this PR if there is not problem~

PR #1429, opened 2021-02-20
kv-client: filter key overflow region before sending to puller

amyangfei:

Fix duplicated DDL output from puller.

DDL puller watches two region range, [mDDLJobListl, mDDLJobListm) and [mDDLJobAddIdxListl, mDDLJobAddIdxListm), two ranges belong to the same region. If we filter region-uncover event in puller, we will get duplicated DDL events.

In most cases, regions should not be overlapped, such as each table owns one or more independent regions. However DDL update subscription is a particular case.

There is no bug exposed to user, so we don't need to update release note.

Filter key overflow region in kv-client before sending to the puller. The CPU cost for checking region coverage is almost same between checking in kv-client and in puller, while check in kv-client can save some time for event entry processing and channel send operation.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -1387,6 +1387,14 @@ func (s *eventFeedSession) singleEventFeed(
 			switch x := event.changeEvent.Event.(type) {
 			case *cdcpb.Event_Entries_:
 				for _, entry := range x.Entries.GetEntries() {
+					// if a region with kv range [a, z)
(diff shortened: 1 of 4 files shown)


leoppro:
LGTM

liuzix:
LGTM

ti-srebot (approved):

PR #1426, opened 2021-02-19
kv-client: add resolved ts heap for quick resolve lock

amyangfei:
### What problem does this PR solve? 

part 2 of #1393, please review this PR after #1397 is merged

### What is changed and how it works?

- Add a heap to manage region resolved ts, so we don't need to iterate each region every time, and this can reduce lock contention for region state.
- Use a channel to pass resolved ts update event, so resolved ts update will not block event processing

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

Issue #1393, opened 2021-02-02
kv client optimization


amyangfei:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently TiCDC subscribes to the KV changed logs based on region, and allocates an independent goroutine for each region to handle KV changed logs. When the region count is large, there exists huge lock contention in Golang runtime and CPU cost is too high, which will reduce replication performance.  Since the stability and performance of kv client have a great impact on TiCDC service, we want to refactor the implement mechanism of kv client in TiCDC, in order to increase stability and performance.

**Describe the feature you'd like:**

- [x] Implement a better way to handle region’s kv changed logs. #1397, #1439
- [x] Separate lock resolve logic from each region. #1397, #1426
- [x] Add batch mechanism to region event feeds dispatcher. #1720
- [x] Add more unit tests for kv client, ref: #1292 
- [x] Add benchmark for kv client. #1448 
- [x] Real scenario benchmark with new kv client [ga matrix](https://docs.google.com/document/d/1PAbTvIdLVCItwvDTSGxNTSF4Jj5wmK1U37ST5uu0q44/edit#heading=h.wvfxqs3i10q6)
- [x] Support kv client reconnect when resolved ts in TiKV is blocked #1586 
    - [x] cherry pick reconnect logic of client v2 to 5.0

## New bugs found in 5.0 test

- [x] region range lock blocking issue #1599. Both in v1 and v2
- [x] performance bottleneck in error handling and lock resolver #1612. Both in v1 and v2
- [x] The CRTs must be greater than the resolvedTs, #1548 . v2 only
- [x] kv client may raise error when stream.Recv meets io.EOF in some scenarios. Both in v1 and v2. #1641

PR #1439, opened 2021-02-22
kv-client: use multiple sync.Map to reduce lock contention

amyangfei:

part 3 of #1393, please review this PR after #1426 is merged

- In write less, read more scenario, `sync.Map` is faster than RWMutex and Mutex. Note this is concluded from some benchmarks, such as https://github.com/golang/go/issues/17973, https://www.jianshu.com/p/cffffa914381, but maybe not quite suitable for our scene, we need more tests for this point, the most important thing here is to simulate the real world workload.
- Separate into multiple `sync.Map` can accelerate map access

- Use multiple `sync.Map`s for regionFeedState get/put operation.
- The bucket size of `sync.Map` pool is determined by cpu number, min threshold, and max threshold

### Check List 

Tests 

 - Unit test
 - Integration test
 
### Release note

- No release note

### DIFF
--- a/cdc/kv/region_worker_test.go
+++ b/cdc/kv/region_worker_test.go
@@ -1,0 +1,106 @@
+// Copyright 2021 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kv
+
+import (
+	"math/rand"
+	"sync"
+
+	"github.com/pingcap/check"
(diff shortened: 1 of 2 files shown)


cdc/kv/region_worker_test.go
```diff
@@ -0,0 +1,106 @@
...
+	regionIDs := make([]uint64, regionCount)
+	for i := 0; i < regionCount; i++ {
+		regionID := uint64(1000 + i)
+		regionIDs[i] = regionID
+		rsm.setState(regionID, &regionFeedState{requestID: uint64(i + 1), lastResolvedTs: uint64(1000)})
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
```
leoppro, line 55: could you please add more goroutine to make sure the thread-safety?

amyangfei: addressed

amyangfei:
PTAL @leoppro @liuzix

leoppro:
LGTM

liuzix:
LGTM

ti-srebot (approved):

PR #1447, opened 2021-02-24
sink: fix float support in Avro

liuzix:
- Avro not supporting FLOAT type.

- Setting the internal data type representing FLOAT to float64 instead of float32.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- sink: fix float support in Avro


leoppro (approved):

overvenus:
Please add a release note.

overvenus:
What about other mysql types? E.g., `mysql.TypeDecimal`?

cdc/sink/codec/avro.go
```diff
@@ -466,6 +466,8 @@ func columnToAvroNativeData(col *model.Column) (interface{}, string, error) {
 			return handleUnsignedInt64()
 		}
 		return col.Value.(int64), "long", nil
+	case mysql.TypeFloat:
+		return col.Value.(float64), "float", nil
 	default:
 		avroType, err := getAvroDataTypeFallback(col.Value)
```
overvenus, line 472: Could log the value in line avro.go:299?

```
299     log.Warn("getAvroDataTypeFallback: unknown type")
```

overvenus (commented):
LGTM

ti-srebot (approved):

cdc/sink/codec/avro.go
```diff
@@ -466,6 +466,8 @@ func columnToAvroNativeData(col *model.Column) (interface{}, string, error) {
 			return handleUnsignedInt64()
 		}
 		return col.Value.(int64), "long", nil
+	case mysql.TypeFloat:
+		return col.Value.(float64), "float", nil
```
amyangfei, line 470: Is it ok to use `float32` here, since float in TiDB/MySQL is 4 bytes

amyangfei: OH, just ignore this comment, float is 8 bytes in TiDB

liuzix: If it were `float32`, the original implementation would have worked.

cdc/sink/codec/avro.go
liuzix: Could logging the value violate customer privacy?

PR #1459, opened 2021-02-26
*: refine the vars in context.Context

leoppro:

- define a GlobalVars to store some global instance 
- define a ChangefeedVars to store id and info in changefeed

### Check List 

Tests 

 - Unit test
 - Integration test


cdc/server.go
```diff
@@ -370,9 +370,13 @@ func (s *Server) etcdHealthChecker(ctx context.Context) error {
...
 	ctx = util.PutCaptureAddrInCtx(ctx, s.opts.advertiseAddr)
 	ctx = util.PutTimezoneInCtx(ctx, s.opts.timezone)
+	kvStorage, err := util.KVStorageFromCtx(ctx)
+	if err != nil {
+		return errors.Trace(err)
+	}
 
 	procOpts := &processorOpts{flushCheckpointInterval: s.opts.processorFlushInterval}
-	capture, err := NewCapture(ctx, s.pdEndpoints, s.pdClient, s.opts.credential, s.opts.advertiseAddr, procOpts)
+	capture, err := NewCapture(ctx, s.pdEndpoints, s.pdClient, s.opts.credential, kvStorage, s.opts.advertiseAddr, procOpts)
```
amyangfei, line 379: We still have a mixture use of `cdc context` and std context, `globalVars` and std context value, will we unify to use one context later, or when should we use `cdc context` and when using `std context` is better.
For example, can we use `cdc context` when calling `NewCapture`

leoppro: I hope we can unify the context step by step

liuzix:
be careful with the formatting of imports. Rest LGTM

amyangfei (approved):

cdc/processor/manager.go
```diff
@@ -14,7 +14,7 @@
 package processor
 
 import (
-	"context"
+	stdContext "context"
```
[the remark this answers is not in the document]
leoppro: fine, but I will rename it in next pr

PR #1468, opened 2021-03-02
context: refine error handler

leoppro:
- Allow the error handler to throw the errors to the parent context
- listen to error in table pipeline with the error handler

### Check List 

Tests 

 - Unit test

### Release note

- No release note


amyangfei:
rest lgtm

PR #1481, opened 2021-03-03
kv-client: use worker pool in region worker

amyangfei:

Part of #1393 

To increase the throughput of kv client, we should use multiple goroutines in region worker

- Reuse `github.com/pingcap/ticdc/pkg/workerpool`

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


amyangfei:
PTAL @liuzix @leoppro @overvenus

cdc/kv/client_test.go
```diff
@@ -45,7 +46,15 @@ import (
 	"google.golang.org/grpc"
 )
 
-func Test(t *testing.T) { check.TestingT(t) }
+func Test(t *testing.T) {
+	conf := config.GetDefaultServerConfig()
+	config.StoreGlobalServerConfig(conf)
```
leoppro, line 51: should we set a default config in `init` function?

cdc/server.go
```diff
@@ -263,6 +264,10 @@ func (s *Server) run(ctx context.Context) (err error) {
 		return sorter.RunWorkerPool(cctx)
 	})
 
+	wg.Go(func() error {
```
leoppro, line 267: maybe we can run worker pool in `init` function?

leoppro (approved):

cdc/server.go
amyangfei: yep both run it in `init` function and explicit call work, but imho the `init` function is a little dangerous in some scenarios, which may make initialization step unclear (such as init sequence, unexpected init when we import a package but don't want it to init, etc)
such as the discussion in this article: https://medium.com/higher-order-functions/golang-how-to-sensibly-use-init-function-8077a4ec15e5

pkg/config/config.go
```diff
@@ -161,6 +161,10 @@ var defaultServerConfig = &ServerConfig{
 		SortDir:                "/tmp/cdc_sort",
 	},
 	Security: &SecurityConfig{},
+	KVClient: &KVClientConfig{
+		WorkerConcurrent: 8,
+		WorkerPoolSize:   0, // 0 will use NumCPU() * 2
```
overvenus, line 166: Do we need an upper limit of pool size?

cdc/kv/region_worker.go
```diff
@@ -617,3 +785,35 @@ func (w *regionWorker) evictAllRegions(ctx context.Context) error {
...
+		return
+	}
+	workerPoolOnce.Do(func() {
+		cfg := config.GetGlobalServerConfig().KVClient
+		var size int
+		if cfg.WorkerPoolSize > 0 {
+			size = cfg.WorkerPoolSize
+		} else {
+			size = runtime.NumCPU() * 2
+		}
```
overvenus, line 802: Is there any test that covers these lines?

amyangfei: added unit test

pkg/config/config.go
amyangfei: add an upper limit 64

amyangfei:
ptal @overvenus @liuzix

cdc/kv/region_worker.go
```diff
@@ -15,28 +15,43 @@ package kv
...
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/time/rate"
 )
 
+var (
+	regionWorkerPool workerpool.WorkerPool
+	workerPoolOnce   sync.Once
+	// TODO: unified channel buffer mechanism
+	regionWorkerInputChanSize = 128000
```
JinLingChristopher, line 45: can you add a comment about this magic 128000?

amyangfei: added some comments

PR #1483, opened 2021-03-06
WIP: Upgrade pulsar-client-go to use OrderingKey

sunxiaoguang:
Upgrade pulsar-client-go to use OrderingKey to choose partition which is
the recommended way for routing message in pulsar.

Signed-off-by: Xiaoguang Sun <sunxiaoguang@zhihu.com>

Pulsar recommend using OrderingKey for routing if key shouldn't be used instead.  

Upgrade pulsar to 0.4.0 which implemented OrderingKey feature that existed in clients of other languages. By encoding partition into OrderingKey, there is no need to use magic name to store partition in Properties.  

### Check List 

Tests 

 - Unit test

Code changes
NA

Side effects
NA

Related changes

### Release note

- No release note


cdc/sink/producer/pulsar/option.go
```diff
@@ -30,8 +30,6 @@ type Option struct {
 	producerOptions *pulsar.ProducerOptions
 }
 
-const route = "$route"
```
sunxiaoguang, line 33: Magic property name is no longer needed.

sunxiaoguang:
Looks like there are runtime conflict introduced by the new version of pulsar-client-go need to be fixed.

hi-rustin:
PR currently has a large number of conflicts and I will close it for now. Please feel free to reopen it if you want to continue working on it subsequently. Thank you for your contribution!

Issue #1482, opened 2021-03-05
Use proxy in all http/http2 requests when http_proxy/https_proxy is set in env


amyangfei:
## Feature Request

**Is your feature request related to a problem? Please describe:**

When `http_proxy` and `https_proxy` are set in env, TiCDC ignores it and won't use proxy when making http/http2(gRPC) request.

**Describe the feature you'd like:**

- Add a check for env variables and log when `http_proxy` or `https_proxy` is enabled
- Use proxy when `http_proxy` or `https_proxy` is enabled

YuJuncen:
I'm wondering whether currently connections has been using those environment variables. Seems both connection to PD and TiKV are created by standard gRPC connection creating, which uses `HTTP_PROXY` and `HTTPS_PROXY`.

See [here](https://github.com/grpc/grpc/blob/da8c26f302bcb854b6f6ca65ecc51fa791b40e8e/doc/environment_variables.md#:~:text=%20gRPC%20environment%20variables%20%201%20http_proxy%20The,12%20resolver_refcount%2013%20lb_policy_refcount%2014%20chttp2_refcount%20More%20).

I have tried creating a TiCDC server with `HTTP_PROXY`. And the log of proxy shows that TiCDC to PD connection did uses this proxy.
![image](https://user-images.githubusercontent.com/36239017/110272132-6b19a780-8004-11eb-9c4c-2fdd6df38d66.png)

YuJuncen:
Finally, I think I got what is happing.
1. gRPC DID use `HTTPS_PROXY`.
2. Not as what documented, `HTTP_PROXY` would be ignored at most time, because `mapAddress` made a https URL even for plain text HTML connections:

https://github.com/grpc/grpc-go/blob/d5b628860d4e17d106f1964097c0a297659e8110/internal/transport/proxy.go#L40-L46

3. After setting `HTTPS_PROXY` to a local proxy, CDC uses it, and we can capture those gRPC packets:

![image](https://user-images.githubusercontent.com/36239017/110589252-18312300-81b1-11eb-8582-349ff056d753.png)

amyangfei:
This issue has been fixed by #1492

PR #1492, opened 2021-03-10
cmd: add log to http proxy environment variables

YuJuncen:
slove #1482

Since gRPC uses `https_proxy` already(see [here](https://github.com/grpc/grpc-go/blob/master/Documentation/proxy.md), even with a [caveat](https://github.com/pingcap/ticdc/issues/1482#issuecomment-794988747)), we only need to hint the user that those variables was used.

### Check List 

Tests 

#### cli
image

#### server
image

### Release note

- cmd: commands would environment variables relative to proxies now.


YuJuncen:
PD disconnected...?

/run_integration_test

cmd/util.go
```diff
@@ -311,3 +313,21 @@ func strictDecodeFile(path, component string, cfg interface{}) error {
 
 	return errors.Trace(err)
 }
+
+// logHTTPProxies logs HTTP proxy relative environment variables.
+func logHTTPProxies() {
```
amyangfei, line 318: Please add a unit test for this function

YuJuncen: Done.

YuJuncen:
Can I run it?

/run-integration-tests

YuJuncen:
Finally... PTAL @leoppro @amyangfei

YuJuncen:
/run-all-tests

It fails consistently? 🧐

PR #1500, opened 2021-03-12
mounter: fix a compatibility issue after tidb#23092

leoppro:

before https://github.com/pingcap/tidb/pull/23092, the Datum of a BLOB/TEXT column is bytes type.
after https://github.com/pingcap/tidb/pull/23092, the Datum of a BLOB/TEXT column is string type.

so we need to convert string to bytes to keep compatibility 

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


amyangfei (commented):
LGTM

amyangfei (approved):

PR #1507, opened 2021-03-16
schema storage: get the schema snap of  `commit ts -1`  to support async commit

leoppro:
after https://github.com/pingcap/tidb/pull/23274 , tidb enable async commit by default, and it leads that the commit ts of a DML may be equal with the finished ts of a DDL, and the DML using the schema info before the DDL.

so we should get schema snapshot by `CRTs - 1`

### Check List 

Tests 

 - Integration test

### Release note
- No release note


amyangfei:
Have we tested the compatibility, with TiDB 4.0, TiDB 5.0 async commit off, TiDB 5.0 async commit on

leoppro:
/run-integration-tests tidb=release-4.0 tikv=release-4.0 pd=release-4.0

PR #1512, opened 2021-03-16
Fix run make test error

yuqi1129:
see https://github.com/pingcap/ticdc/issues/1511

### Check List 

Tests 

 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


scripts/fix_lib_zstd.sh
```diff
@@ -7,4 +7,10 @@ GO111MODULE=on go mod download ${module}
 # In CI environment, the gopath contains multiple dirs, choose the first one
 cd $(echo ${GOPATH}|awk -F':' '{print $1}')/pkg/mod/${module}/pulsar/internal/compression
 sudo rm zstd_cgo.go
-sudo sed -i '/build !cgo/d' zstd.go
+
```
hi-rustin, line 10: Tested on a Mac. There still seems to be some errors.

```log
➜  ticdc git:(master) ✗ make test
# use -mod=mod to avoid error: missing go.sum entry for module providing package
# ref: https://github.com/golang/go/issues/44129
which bin/failpoint-ctl >/dev/null 2>&1 || CGO_ENABLED=0 GO111MODULE=on go build  -trimpath -mod=mod -o bin/failpoint-ctl github.com/pingcap/failpoint/failpoint-ctl && go mod tidy
./scripts/fix_lib_zstd.sh
rm: zstd_cgo.go: No such file or directory
sed: -e expression #1, char 2: extra characters after command
make: *** [unit_test] Error 1
➜  ticdc git:(master) ✗ 

```

amyangfei: It was caused by a static link in `https://github.com/apache/pulsar-client-go`, once we upgrade this library, this patch can be removed. However if we want to upgrade this library, since we have to upgrade gRPC-go at the same time.

maxshuang: go version 1.17 + macOS,  TiCDC branch release_5.1 +  this patch,  run "make unit_test"，will cause SIGTRAP directly.

hi-rustin:
/close 

by https://github.com/pingcap/ticdc/pull/2753

hi-rustin:
Thanks for your contribution!   💚 💙 💜 💛 ❤️

PR #1524, opened 2021-03-18
sink/mysql: fix txn wg counter is not updated correctly

### LINKED ISSUE #1486: ErrMySQLConnectionError: 1040: Too many connections
## Bug Report

- Ran TiCDC with a MySQL downstream. Added a pressure of 20000 qps to the upstream TiDB.

- That the downstream remains available to other queries.

3. What did you see instead?
- Although replication proceeds, the downstream does not accept more connections, and reports `1040: Too many connections`

4. Versions of the cluster

        ```
        4.0.11
        ```

        ```
        4.0.11
        ```

amyangfei:

Fix #1486

There exist two scenarios that `txnWg` is not updated correctly

- When a worker receives a new txn from txnCh, and triggers `flushRows` function, `flushRows` returns an error and the routine terminates. There could be txns remained in `txnCh`
- When a worker exits and fetches all txns from `txnCh` and update `txnWg`, there could be new txns sent to `txnCh`.

~So when the worker exits, fill in the `txnCh` to prevent new txns sent to `txnCh`, and update the `txnWg` with correct txn number.~

- Add a `*sync.WaitGroup` field in `SingleTableTxn` struct.
- Before wait for all workers flushed txns, broadcast a control txn, which contains a `*sync.Waitgroup` field
- Wait the wait group until all txns in workers are flushed.
- If error happens in worker, consume all txns and mark wait group done once.

There are two candidate solutions, one is to use txnNum to control, the other one is to introduce the control txn

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Fix a bug that txn counter is not correctly updated which may cause database connection leak.


amyangfei:
PTAL @leoppro @liuzix

cdc/sink/mysql.go
```diff
@@ -765,28 +785,33 @@ func (w *mysqlSinkWorker) run(ctx context.Context) (err error) {
 		copy(rows, toExecRows)
 		err := w.execDMLs(ctx, rows, replicaID, w.bucket)
 		if err != nil {
-			w.txnWg.Add(-1 * txnNum)
 			txnNum = 0
```
leoppro, line 788:
```suggestion
			txnNum = 0
```
could we remove the `txnNum`

amyangfei: used in metrics

cdc/sink/mysql.go
```diff
@@ -765,28 +785,33 @@ func (w *mysqlSinkWorker) run(ctx context.Context) (err error) {
...
 			if txn == nil {
 				return errors.Trace(flushRows())
 			}
+			if txn.FinishWg != nil {
+				if err := flushRows(); err != nil {
+					return errors.Trace(err)
+				}
+				txn.FinishWg.Done()
+			}
 			if txn.ReplicaID != replicaID || len(toExecRows)+len(txn.Rows) > w.maxTxnRow {
```
leoppro, line 812: wu should check if the len(row) is zero

amyangfei: add a continue, no need to check here

leoppro (commented):
LGTM

PR #1525, opened 2021-03-18
tests: use global tidb_enable_clustered_index variable

amyangfei:

Fix error in integration test

`[2021-03-18T10:59:15.078Z] ERROR 1105 (HY000) at line 5: Variable 'tidb_enable_clustered_index' is a GLOBAL variable and should be set with SET GLOBAL`

ref: https://github.com/pingcap/tidb/pull/23270

- ~`alter-primary-key` is deprecated, use `nonclustered` keyword instead~, to keep backward compatibility (tidb 5.0.0-rc), we don't remove this configuration. Besides, the `nonclustered` keyword can't be used either, but since int primary key is using `nonclustered` by default, we can omit it.
- `tidb_enable_clustered_index` is global variable only

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/tests/clustered_index/data/test.sql
+++ b/tests/clustered_index/data/test.sql
@@ -2,7 +2,8 @@ drop database if exists `clustered_index_test`;
 create database `clustered_index_test`;
 use `clustered_index_test`;
 
-set @@tidb_enable_clustered_index=1;
+set global tidb_enable_clustered_index=1;
+select sleep(2);
 
 CREATE TABLE t0 (
 	id VARCHAR(255),
--- a/tests/clustered_index/data/test.sql
+++ b/tests/clustered_index/data/test.sql
@@ -2,9 +2,6 @@ drop database if exists `clustered_index_test`;
 create database `clustered_index_test`;
 use `clustered_index_test`;
 
-set global tidb_enable_clustered_index=1;
-select sleep(2);
-
 CREATE TABLE t0 (
 	id VARCHAR(255),
 	data INT,
--- a/tests/clustered_index/data/test.sql
+++ b/tests/clustered_index/data/test.sql
@@ -2,8 +2,7 @@ drop database if exists `clustered_index_test`;
 create database `clustered_index_test`;
 use `clustered_index_test`;
 
-set global tidb_enable_clustered_index=1;
-select sleep(2);
+set @@tidb_enable_clustered_index=1;
 
 CREATE TABLE t0 (
 	id VARCHAR(255),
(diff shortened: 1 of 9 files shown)


liuzix:
LGTM

tests/clustered_index/data/test.sql
```diff
@@ -2,7 +2,8 @@ drop database if exists `clustered_index_test`;
 create database `clustered_index_test`;
 use `clustered_index_test`;
 
-set @@tidb_enable_clustered_index=1;
+set global tidb_enable_clustered_index=1;
+select sleep(2);
```
tangenta, line 6: > Similar to MySQL, changes to GLOBAL variables do not apply to either existing connections, or the local connection. Only new sessions reflect the changes to the value.

https://docs.pingcap.com/tidb/stable/sql-statement-set-variable#set-globalsession-variable :cry:

amyangfei: So is this variable enabled by default in tidb master?

amyangfei: Besides I found we can't query `tidb_enable_clustered_index` by show variables.

```
mysql root@127.0.0.1:(none)> select @@global.tidb_enable_clustered_index
+--------------------------------------+
| @@global.tidb_enable_clustered_index |
+--------------------------------------+
| 1                                    |
+--------------------------------------+
1 row in set
Time: 0.009s
mysql root@127.0.0.1:(none)> show variables like '%clustered%';
+---------------+-------+
| Variable_name | Value |
+---------------+-------+
0 rows in set
Time: 0.009s
mysql root@127.0.0.1:(none)> select tidb_version()
+-------------------------------------------------------------------+
| tidb_version()                                                    |
+-------------------------------------------------------------------+
| Release Version: v4.0.0-beta.2-2410-g915dde1f9                    |
| Edition: Community                                                |
| Git Commit Hash: 915dde1f93df4e542bff1eb9a27147a3eede1954         |
| Git Branch: master                                                |
| UTC Build Time: 2021-03-18 15:16:08                               |
| GoVersion: go1.15                                                 |
| Race Enabled: false                                               |
| TiKV Min Version: v3.0.0-60965b006877ca7234adaced7890d7b029ed1306 |
| Check Table Before Drop: false                                    |
+-------------------------------------------------------------------+
1 row in set
Time: 0.007s

```

amyangfei: ptal @tangenta , thanks!

tangenta (commented):
/lgtm

leoppro:
LGTM

ti-srebot (approved):

PR #1530, opened 2021-03-22
kv/client: fix lock range may block when context is cancelled

amyangfei:

Fix region lock blocked if the kv client has been canceled and some of the regions are not unlocked.

Add context cancel check in region range lock

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Fix lock range may block when context is cancelled in kv client

### DIFF
--- a/pkg/regionspan/region_range_lock.go
+++ b/pkg/regionspan/region_range_lock.go
@@ -15,6 +15,7 @@ package regionspan
 
 import (
 	"bytes"
+	"context"
 	"encoding/hex"
 	"fmt"
 	"math"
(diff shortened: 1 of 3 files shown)


amyangfei:
PTAL @leoppro @liuzix

pkg/regionspan/region_range_lock.go
```diff
@@ -298,7 +299,11 @@ func (l *RegionRangeLock) LockRange(startKey, endKey []byte, regionID, version u
 		var res1 LockRangeResult
 		for {
 			for _, ch := range signalChs1 {
```
liuzix, line 301: Could reading from signalChs1 block when the kvclient is being cancelled?

amyangfei: Do you mean range for `signalChs1`, it is a chan slice

leoppro:
LGTM

ti-srebot (approved):

PR #1554, opened 2021-03-26
cmd: make update config conservative, only update listed parameter

### LINKED ISSUE #1545: Unexpected behavior for cdc cli changefeed update
## Bug Report

   1. create a changefeed with `sort-engine=unified`
   2. update changefeed config via `cdc cli changefeed update -c xxx --sink-uri="blackhole://"

   Only `sink-uri` is updated

3. What did you see instead?

   `sort-engine` is set to `memory`

4. Versions of the cluster

        ```
        v4.0.11
        ```

        ```
        v4.0.11
        ```

amyangfei:

Fix #1545

Change the behavior of `cdc cli changefeed update`, will only update parameter that provided in command.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- cli changefeed update will update parameter provided in command only.


kennytm:
what the heck, bot 😓.

amyangfei:
> what the heck, bot 😓.

Make me confused too. What's the rule that tichi-bot chooses review candidates? @hi-rustin

hi-rustin:
> > what the heck, bot 😓.
> 
> Make me confused too. What's the rule that tichi-bot chooses review candidates? @hi-rustin



> > what the heck, bot 😓.
> 
> Make me confused too. What's the rule that tichi-bot chooses review candidates? @hi-rustin

See: https://book.prow.tidb.io/#/en/plugins/blunderbuss?id=design
kennytm is a member of migrate SIG and may be picked because it is weighted random. If you don't think you want this, you need to ignore kennytm in the configuration, which is not in the current configuration. 

https://github.com/ti-community-infra/configs/blob/49e8130252ba166e5de5f880b1fc7a6f9c61d361/prow/config/external_plugins_config.yaml#L727

amyangfei:
PTAL @leoppro @liuzix

kennytm:
@hi-rustin is it possible to make that an "include" list instead of an "exclude" list?

hi-rustin:
> @hi-rustin is it possible to make that an "include" list instead of an "exclude" list?

For some large repos, this may be more difficult to list everyone.

cdc/model/changefeed_test.go
```diff
@@ -190,6 +190,30 @@ func (s *configSuite) TestVerifyAndFix(c *check.C) {
 	c.Assert(marshalConfig1, check.Equals, marshalConfig2)
 }
 
+func (s *configSuite) TestChangeFeedInfoClone(c *check.C) {
+	defer testleak.AfterTest(c)()
+	info := &ChangeFeedInfo{
+		SinkURI: "blackhole://",
+		Opts:    map[string]string{},
```
lance6716, line 197: i remember a empty map will become nil in some case. will it break something?

amyangfei: Not the same thing, `map[string]string{}` or `new(map[string]string)` has initialized the map, it is not nil anymore

leoppro (approved):

cdc/model/changefeed_test.go
lance6716: I checked that only for Unmarshal of `protoc-gen-gogo`, empty map will become nil 😂

PR #1558, opened 2021-03-26
sorter: fix Unified Sorter resource release

liuzix:
- Unified Sorter recycles temp files prematurely, causing file corruption.

- Added a mechanism to cancel asynchronous IO operations.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes

### Release note

- No release note

### DIFF
--- a/cdc/puller/sorter/heap_sorter.go
+++ b/cdc/puller/sorter/heap_sorter.go
@@ -1,4 +1,4 @@
-// Copyright 2020 PingCAP, Inc.
+// Copyright 2021 PingCAP, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import (
 	"github.com/pingcap/log"
 	"github.com/pingcap/ticdc/cdc/model"
 	"github.com/pingcap/ticdc/pkg/config"
+	cerrors "github.com/pingcap/ticdc/pkg/errors"
 	"github.com/pingcap/ticdc/pkg/util"
 	"github.com/pingcap/ticdc/pkg/workerpool"
 	"go.uber.org/zap"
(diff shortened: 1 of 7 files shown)


cdc/puller/sorter/heap_sorter.go
```diff
@@ -284,6 +319,35 @@ func (h *heapSorter) init(ctx context.Context, onError func(err error)) {
...
+	c.exitRWLock.RLock()
+}
+
+func (c *asyncCanceller) FinishAsyncOp() {
+	c.exitRWLock.RUnlock()
+}
+
+func (c *asyncCanceller) IsCanceled() bool {
+	return c.hasExited
+}
```
overvenus, line 342: Can we remove `EnterAsyncOp` and `FinishAsyncOp` by rewriting `IsCanceled` to the following?

```suggestion
func (c *asyncCanceller) IsCanceled() bool {
	c.exitRWLock.Lock()
	defer c.exitRWLock.Unlock()

	return c.hasExited
}
```

liuzix: The whole IO operation needs to be protected so that we don't try to release the files when they are being written to. So it's advisable to wrap the entire execution in the read lock.

cdc/puller/sorter/heap_sorter.go
```diff
@@ -146,6 +155,21 @@ func (h *heapSorter) flush(ctx context.Context, maxResolvedTs uint64) error {
 	if !isEmptyFlush {
 		backEndFinal := backEnd
 		err := heapSorterIOPool.Go(ctx, func() {
+			failpoint.Inject("asyncFlushStartDelay", func() {
+				log.Debug("asyncFlushStartDelay")
+			})
+
+			h.canceller.EnterAsyncOp()
+			defer h.canceller.FinishAsyncOp()
```
amyangfei, line 163: If read lock is acquired, the `lock` operation will wait, does it mean the flush can't be cancelled during it is running, is it by design?

liuzix: Will fix

liuzix:
PTAL @amyangfei  @leoppro

PR #1574, opened 2021-03-29
sorter: avoid blocking in thread pool

liuzix:
- The Unified Sorter makes use of a thread pool internally, and inside the thread pool, the task of one table might be blocked by the task of another one. It was fine until we introduce flow control, which will create cyclic waiting because only the progress of the slowest table will cause other tables' output to be unblocked.

- Decouple Merger into two parts, which prevent the pressure from the blockage to reach the heapSorter. Since only the heapSorter makes use of the thread pool, preventing heapSorter blockage will prevent deadlocking.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test 

Side effects

 - Increased code complexity

Related changes

### Release note

- Make Unified Sorter compatible with sink flow control.


liuzix:
I will finalize this PR after #1558 is merged

cdc/puller/sorter/unified_sorter.go
```diff
@@ -198,6 +201,15 @@ func (s *UnifiedSorter) Run(ctx context.Context) error {
 
 		nextSorterID := 0
 		for {
+			// tentative value 1280000
+			for atomic.LoadInt64(&mergerBufLen) > 1280000 {
```
amyangfei, line 205: When the merge buffer exceeds max length, could the deadlock also happen?

liuzix: No, it couldn't. Because blocking here does not block any other table from proceeding. @leoppro Can you confirm this?

leoppro: i think is fine, the blocking will not passing to other tables

cdc/puller/sorter/merger.go
```diff
@@ -417,3 +482,89 @@ func printError(err error) error {
...
+				return nil, nil
+			}
+
+			if time.Since(startTime) > time.Second*5 {
+				log.Debug("taskBuffer reading blocked for too long", zap.Duration("duration", time.Since(startTime)))
+			}
+		}
+	}
+
+	postCount := atomic.AddInt64(b.len, -1)
```
siddontang, line 547: for above `put`, you handle `b.len` and queue in the mutex together, but here you handle `b.len` outside of mutex, is it correct?

liuzix: The mutex is used to protect the queue but not `b.len`, so there is no problem. Moreover, since `atomic` operations in golang's concurrency model enforces a happens-before relation, the enqueuing and bumping the `len` operations have a strictly defined order, so it is okay to assume we see the bumped `len` only after enqueuing has been successful even without a mutex.

cdc/puller/sorter/merger.go
```diff
@@ -417,3 +482,89 @@ func printError(err error) error {
 	}
 	return err
 }
+
+// taskBuffer is used to store pending flushTasks.
+// The design purpose is to reduce the backpressure caused by a congested output chan of the merger,
+// so that heapSorter does not block.
+type taskBuffer struct {
+	mu       sync.Mutex
+	queue    deque.Deque
```
amyangfei, line 491: curious about, is `deque.Deque` faster than generical slice

cdc/puller/sorter/merger.go
```diff
@@ -417,3 +482,89 @@ func printError(err error) error {
...
+		for atomic.LoadInt64(b.len) == 0 {
+			select {
+			case <-ctx.Done():
+				return nil, errors.Trace(ctx.Err())
+			case <-recv.C:
+			}
+
+			if atomic.LoadInt32(&b.isClosed) == 1 {
+				return nil, nil
+			}
```
amyangfei, line 539: The behavior of `get` when `taskBuffer` is closed is not unified.
- In L518, if `b.isClosed == 1` && `b.len == 0`, return nil
- In L518, if `b.isClosed == 1` && `b.len > 0`, will return a task
- In L518, `b.isClosed == 0`, `b.len == 0`, in L537, if `recv.C` bumps and `b.len > 0` and `b.isClosed == 1`, still return nil

amyangfei: Besides, can we simplify the check condition, seems the check of `b.len > 0` and `recv.C` do the same work

liuzix: Note that there might be spurious wake-ups because the receiver has a timer.

amyangfei: Yep, but receiver can be configured with non-ticker

PR #1553, opened 2021-03-26
kvclientv2: fix resolvedts go back

leoppro:
### What problem does this PR solve? 

fix #1548

when a stream to kvstore is closed, kvclientv2 sent all region info under this stream with `onRegionFail`.
but the checkpoint is not updated in this time.

### Check List 

Tests 

 - Unit test
### Release note

- No release note (this bug is never released)

PR #1591, opened 2021-03-31
kv/client: support reconnect regions when no event received too long

amyangfei:

Solve https://github.com/pingcap/ticdc/issues/1586 (This is the master PR, release-5.0 PR is https://github.com/pingcap/ticdc/pull/1589)

NOTE, **need discussion**: 

1. The check only workers after region is initialized, is this reasonable.
2. The reconnect interval (15m) is hard coded, we need to discuss whether to add a config in server config file

After discussion

1. Keep check after the region is initialized, because the incremental scan phase could cost long.
2. We don't need to add it in server config file, since the check is only in real time data phase.

- Reuse kv event aliveness check in lock resolver tick
- If a region doesn't receive any event for more than `15 minutes`, force the gRPC stream to reconnect
- Only update the kv client v1, kv client v2 will be updated later

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Force the gRPC stream to reconnect if no event (including resolved ts event) is received for more than 15 minutes.


amyangfei:
This is master PR, PTAL too, thx @leoppro @liuzix

leoppro (approved):

amyangfei:
The integration test relys on #1553

PR #1603, opened 2021-04-02
sink: Decrease the default value of `MaxBatchSize` to avoid `Kafka Msg too large`

leoppro:
Decrease the default value of MaxBatchSize to avoid Kafka Msg too large

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


cdc/sink/codec/json.go
```diff
@@ -38,7 +38,7 @@ const (
 	// DefaultMaxMessageBytes sets the default value for max-message-bytes
 	DefaultMaxMessageBytes int = 64 * 1024 * 1024 // 64M
 	// DefaultMaxBatchSize sets the default value for max-batch-size
-	DefaultMaxBatchSize int = 4096
+	DefaultMaxBatchSize int = 16
```
amyangfei, line 41: Better to add some comments, about why `16` is used

cdc/sink/codec/json.go
```diff
@@ -38,7 +38,7 @@ const (
 	// DefaultMaxMessageBytes sets the default value for max-message-bytes
 	DefaultMaxMessageBytes int = 64 * 1024 * 1024 // 64M
```
overvenus, line 39: What about 1MB? 64MB exceeds Kafka's default `max.message.bytes` (1MB) https://kafka.apache.org/documentation/#brokerconfigs

leoppro: `1MB` is not a suitable default value, the max size of a value in tikv is 6MB, so the max size of a row event approximately equal to 12MB (with old value), the default value should make sure one row event at lastest can be sent to kafka

liuzix: The current implementation tries to send out the Kafka message with only one event, **no matter what**. The codec will print a warning if the message is larger than the configured value.

liuzix:
LGTM

cdc/sink/codec/json.go
leoppro: I don't know if this is a suitable value, but `4096` is too long

amyangfei:
unit test is broken

PR #1643, opened 2021-04-12
file: add a O_TRUNC flag when create a new log.meta

dengqee:
When create a new log.meta file, the old file was not  cleared. If the new file is shorter the old file, it will like this:
image
There is an extra string in log.meta

This is a feature of Golang, when create a new file, old file will not be cleared, So I add a **O_TRUNC** flag when create a new log.meta to cover old file.

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes


cdc/sink/cdclog/file.go
```diff
@@ -197,7 +197,7 @@ func (f *fileSink) flushLogMeta() error {
 		return cerror.WrapError(cerror.ErrMarshalFailed, err)
 	}
 	// FIXME: if initialize succeed, O_WRONLY is enough, but sometimes it will failed
-	file, err := os.OpenFile(f.logPath.meta, os.O_CREATE|os.O_WRONLY, defaultFileMode)
+	file, err := os.OpenFile(f.logPath.meta, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, defaultFileMode)
```
overvenus, line 200: `O_TRUNC` may discard all contents of a file, if TiCDC is killed between line 200 and line 201, the log.meta is lost.
I suggest using a temp file eg "log.meta.tmp" for saving the data, after that, we can rename the temp file to log.meta.

Note: rename is atomic, https://man7.org/linux/man-pages/man2/rename.2.html#:~:text=atomically

lance6716:
example

https://github.com/pingcap/br/blob/2c82efc277110219ee4b9efb82c7f86a933aecf2/pkg/lightning/checkpoints/checkpoints.go#L967-L975

cdc/sink/cdclog/file.go
dengqee: Thanks for your advice. I'll do it.

amyangfei (approved):

leoppro (approved):

PR #1680, opened 2021-04-18
*: sink flow control 

liuzix:
- Lack of flow control between the sorter and the outlet of the sink.

- Memory quota based flow control below the sorter.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test

1. Single table test

Insert data by
`sysbench --test=oltp_insert --tables=1 --table-size=100000000 --threads=1 --create_secondary=off prepare`.

Waiting for about 150G incremental data to be scanned, and verify that TiCDC does not OOM.

2. Multiple table test (pending)

Side effects

 - Increased code complexity

Related changes

### Release note

- Sink flow control based on memory quota

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -56,9 +56,10 @@ const (
 	dialTimeout           = 10 * time.Second
 	maxRetry              = 100
 	tikvRequestMaxBackoff = 20000 // Maximum total sleep time(in ms)
-	// TODO find optimal values here
-	grpcInitialWindowSize     = 1 << 28 // 128 MB The value for initial window size on a stream
-	grpcInitialConnWindowSize = 1 << 29 // 128 MB The value for initial window size on a connection
+	// TODO find optimal values and test extensively before releasing
+	// The old values cause the gRPC stream to stall for some unknown reason.
+	grpcInitialWindowSize     = 1 << 28 // 256 MB The value for initial window size on a stream
+	grpcInitialConnWindowSize = 1 << 29 // 512 MB The value for initial window size on a connection
 	grpcMaxCallRecvMsgSize    = 1 << 30 // 1024 MB The maximum message size the client can receive
 	grpcConnCount             = 10
 
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -70,10 +70,10 @@ const (
 	// failed region will be reloaded via `BatchLoadRegionsWithKeyRange` API. So we
 	// don't need to force reload region any more.
 	regionScheduleReload = false
-
-	// time interval to force kv client to terminate gRPC stream and reconnect
-	reconnectInterval = 15 * time.Minute
 )
+
+// time interval to force kv client to terminate gRPC stream and reconnect
+var reconnectInterval = 15 * time.Minute
 
 // hard code switch
 // true: use kv client v2, which has a region worker for each stream
@@ -168,6 +168,24 @@ func (s *regionFeedState) markStopped() {
 
 func (s *regionFeedState) isStopped() bool {
 	return atomic.LoadInt32(&s.stopped) > 0
+}
+
+func (s *regionFeedState) isInitialized() bool {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	return s.initialized
+}
+
+func (s *regionFeedState) getLastResolvedTs() uint64 {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	return s.lastResolvedTs
+}
+
+func (s *regionFeedState) getRegionSpan() regionspan.ComparableSpan {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	return s.sri.span
 }
 
 type syncRegionFeedStateMap struct {
@@ -1397,6 +1415,9 @@ func (s *eventFeedSession) singleEventFeed(
 	failpoint.Inject("kvClientResolveLockInterval", func(val failpoint.Value) {
 		resolveLockInterval = time.Duration(val.(int)) * time.Second
 	})
+	failpoint.Inject("kvClientReconnectInterval", func(val failpoint.Value) {
+		reconnectInterval = time.Duration(val.(int)) * time.Second
+	})
 
 	for {
 		var event *regionEvent
@@ -1405,10 +1426,6 @@ func (s *eventFeedSession) singleEventFeed(
 		case <-ctx.Done():
 			return lastResolvedTs, ctx.Err()
 		case <-advanceCheckTicker.C:
-			failpoint.Inject("kvClientForceReconnect", func() {
-				log.Warn("kv client reconnect triggered by failpoint")
-				failpoint.Return(lastResolvedTs, errReconnect)
-			})
 			if time.Since(startFeedTime) < resolveLockInterval {
 				continue
 			}
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -53,11 +53,12 @@ import (
 )
 
 const (
-	dialTimeout               = 10 * time.Second
-	maxRetry                  = 100
-	tikvRequestMaxBackoff     = 20000   // Maximum total sleep time(in ms)
-	grpcInitialWindowSize     = 1 << 27 // 128 MB The value for initial window size on a stream
-	grpcInitialConnWindowSize = 1 << 27 // 128 MB The value for initial window size on a connection
+	dialTimeout           = 10 * time.Second
+	maxRetry              = 100
+	tikvRequestMaxBackoff = 20000 // Maximum total sleep time(in ms)
+	// TODO find optimal values here
+	grpcInitialWindowSize     = 1 << 28 // 128 MB The value for initial window size on a stream
+	grpcInitialConnWindowSize = 1 << 29 // 128 MB The value for initial window size on a connection
 	grpcMaxCallRecvMsgSize    = 1 << 30 // 1024 MB The maximum message size the client can receive
 	grpcConnCount             = 10
 
--- a/cdc/processor/pipeline/sink.go
+++ b/cdc/processor/pipeline/sink.go
@@ -72,9 +72,11 @@ type sinkNode struct {
 
 	eventBuffer []*model.PolymorphicEvent
 	rowBuffer   []*model.RowChangedEvent
+
+	flowController tableFlowController
 }
 
-func newSinkNode(sink sink.Sink, startTs model.Ts, targetTs model.Ts) *sinkNode {
+func newSinkNode(sink sink.Sink, startTs model.Ts, targetTs model.Ts, flowController tableFlowController) *sinkNode {
 	return &sinkNode{
 		sink:         sink,
 		status:       TableStatusInitializing,
@@ -82,6 +84,8 @@ func newSinkNode(sink sink.Sink, startTs model.Ts, targetTs model.Ts) *sinkNode 
 		resolvedTs:   startTs,
 		checkpointTs: startTs,
 		barrierTs:    startTs,
+
+		flowController: flowController,
 	}
 }
 
@@ -130,6 +134,8 @@ func (n *sinkNode) flushSink(ctx pipeline.NodeContext, resolvedTs model.Ts) (err
 		return nil
 	}
 	atomic.StoreUint64(&n.checkpointTs, checkpointTs)
+
+	n.flowController.Release(checkpointTs)
 	return nil
 }
 
--- a/cdc/processor/pipeline/sink.go
+++ b/cdc/processor/pipeline/sink.go
@@ -70,8 +70,6 @@ type sinkNode struct {
 	targetTs     model.Ts
 	barrierTs    model.Ts
 
-	lastFlushedTs model.Ts
-
 	eventBuffer []*model.PolymorphicEvent
 	rowBuffer   []*model.RowChangedEvent
 
@@ -128,7 +126,6 @@ func (n *sinkNode) flushSink(ctx pipeline.NodeContext, resolvedTs model.Ts) (err
 	if err := n.flushRow2Sink(ctx); err != nil {
 		return errors.Trace(err)
 	}
-	atomic.StoreUint64(&n.lastFlushedTs, resolvedTs)
 	checkpointTs, err := n.sink.FlushRowChangedEvents(ctx.StdContext(), resolvedTs)
 	if err != nil {
 		return errors.Trace(err)
@@ -197,11 +194,6 @@ func (n *sinkNode) Receive(ctx pipeline.NodeContext) error {
 			}
 			atomic.StoreUint64(&n.resolvedTs, msg.PolymorphicEvent.CRTs)
 			return nil
-		}
-		lastFlushedTs := atomic.LoadUint64(&n.lastFlushedTs)
-		if event.CRTs < lastFlushedTs {
-			log.Panic("CRTs too small", zap.Uint64("CRTs", event.CRTs),
-				zap.Uint64("lastFlushedTs", lastFlushedTs))
 		}
 		if err := n.emitEvent(ctx, event); err != nil {
 			return errors.Trace(err)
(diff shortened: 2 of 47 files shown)


amyangfei:
PTAL @JinLingChristopher

cdc/kv/client.go
```diff
@@ -53,11 +53,12 @@ import (
...
-	maxRetry                  = 100
-	tikvRequestMaxBackoff     = 20000   // Maximum total sleep time(in ms)
-	grpcInitialWindowSize     = 1 << 27 // 128 MB The value for initial window size on a stream
-	grpcInitialConnWindowSize = 1 << 27 // 128 MB The value for initial window size on a connection
+	dialTimeout           = 10 * time.Second
+	maxRetry              = 100
+	tikvRequestMaxBackoff = 20000 // Maximum total sleep time(in ms)
+	// TODO find optimal values here
+	grpcInitialWindowSize     = 1 << 28 // 128 MB The value for initial window size on a stream
+	grpcInitialConnWindowSize = 1 << 29 // 128 MB The value for initial window size on a connection
```
JinLingChristopher, line 61: comment need to be update, 1 << 28 should be 256, 1 << 29 is 512

liuzix: Yes. These are just experimental values, will need to be finalized before releasing.

cdc/processor/pipeline/sorter.go
```diff
@@ -102,6 +114,45 @@ func (n *sorterNode) Init(ctx pipeline.NodeContext) error {
 				if msg == nil {
 					continue
 				}
+				if msg.RawKV != nil && msg.RawKV.OpType != model.OpTypeResolved {
```
JinLingChristopher, line 117: `if msg.RawKV != nil && msg.RawKV.OpType != model.OpTypeResolved` and `if msg.RawKV != nil && msg.RawKV.OpType == model.OpTypeResolved` is quite verbose, and only one would be true, maybe this logic can be simplified ? how about check `msg.RawKv == nil` just continue, and check `OpType` further ?

cdc/processor/pipeline/sorter.go
```diff
@@ -102,6 +114,45 @@ func (n *sorterNode) Init(ctx pipeline.NodeContext) error {
...
+					size := uint64(msg.RawKV.ApproximateSize() * 2)
+					commitTs := msg.CRTs
+
+					if time.Since(lastSendResolvedTsTime) > time.Millisecond*200 {
```
amyangfei, line 121: Could you please add some comments about how these three variables (`lastCRTs`, `lastSentResolvedTs`, `commitTs`) work together.

cdc/sink/common/flow_control.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	c.Consumed += nBytes
+	return nil
+}
+
+// Release is called when a chuck of memory is done being used.
+func (c *TableMemorySizeController) Release(nBytes uint64) {
+	c.mu.Lock()
+
```
JinLingChristopher, line 106: how about `defer c.mu.Unlock()` at this place ?

cdc/processor/pipeline/table.go
```diff
@@ -154,15 +169,17 @@ func NewTablePipeline(ctx context.Context,
 		cancel:      cancel,
 	}
 
+	flowController := common.NewTableFlowController(perTableMemoryQuota)
```
amyangfei, line 172: The memory controller is processor level, which means each changefeed has its own memory quota, is this by design?

cdc/sink/common/flow_control.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	if commitTs > lastCommitTs {
+		atomic.StoreUint64(&c.lastCommitTs, commitTs)
+		err := c.memoryController.ConsumeWithBlocking(size, blockCallBack)
+		if err != nil {
+			return errors.Trace(err)
+		}
+	} else {
+		// commitTs == lastCommitTs
```
amyangfei, line 171: Does this mean large transaction is not restricted by the memory quote

cdc/sink/common/flow_control.go
```diff
@@ -0,0 +1,209 @@
...
+
+	if commitTs < lastCommitTs {
+		log.Panic("commitTs regressed, report a bug",
+			zap.Uint64("commitTs", commitTs),
+			zap.Uint64("lastCommitTs", c.lastCommitTs))
+	}
+
+	if commitTs > lastCommitTs {
+		atomic.StoreUint64(&c.lastCommitTs, commitTs)
+		err := c.memoryController.ConsumeWithBlocking(size, blockCallBack)
```
amyangfei, line 166: Memory allocation is one by one, could this becomes bottleneck, is there any plan to introduce batch consume mechanism or pre-memory allocation?
Besides we should add some benchmark for the flow controller in later PRs.

cdc/processor/pipeline/sorter.go
```diff
@@ -94,6 +103,9 @@ func (n *sorterNode) Init(ctx pipeline.NodeContext) error {
 		return nil
 	})
 	n.wg.Go(func() error {
+		lastSentResolvedTs := uint64(0)
+		lastSendResolvedTsTime := time.Now()
```
JinLingChristopher, line 107: Is there any relation between `lastSentResolvedTs` and `lastSendResolvedTsTime`, why not  `lastSentResolvedTs := lastSendResolvedTsTime.Unix()`? should timezone be considered ?

cdc/sink/common/flow_control.go
liuzix: It would be correct implementation, but it will likely cause a block consumer to be waken up only to be blocked again immediately by `mu` that is still locked.

cdc/processor/pipeline/sorter.go
```diff
@@ -102,6 +114,45 @@ func (n *sorterNode) Init(ctx pipeline.NodeContext) error {
 				if msg == nil {
 					continue
 				}
+				if msg.RawKV != nil && msg.RawKV.OpType != model.OpTypeResolved {
+					size := uint64(msg.RawKV.ApproximateSize() * 2)
```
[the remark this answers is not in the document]
liuzix: Will fix. This is a tentative value for the "data expansion factor".

cdc/processor/pipeline/sorter.go
liuzix: Ok

cdc/processor/pipeline/sorter.go
```diff
@@ -102,6 +114,45 @@ func (n *sorterNode) Init(ctx pipeline.NodeContext) error {
...
+								zap.String("tableName", n.tableName))
+						} else {
+							ctx.Throw(err)
+						}
+						return nil
+					}
+					lastCRTs = commitTs
+				}
+
+				if msg.RawKV != nil && msg.RawKV.OpType == model.OpTypeResolved {
```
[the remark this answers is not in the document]
liuzix: We can count it, but since the quota is not strictly obeyed in this implementation, counting it is not a priority. We cannot strictly obey to quota because blocking in the middle of a Txn can deadlock the whole pipeline.

cdc/processor/pipeline/table.go
liuzix: This is by design. It is actually at the table level. A global solution is not of high priority for now.

cdc/sink/common/flow_control.go
liuzix: Yes.

cdc/sink/common/flow_control.go
liuzix: This is not a bottleneck. I have pinpointed the bottleneck to the sink manager.

cdc/processor/pipeline/table.go
amyangfei: why it is table level, does all tables in a processor share the same flow controller?

cdc/processor/pipeline/sorter.go
liuzix: 1. `lastSentResolvedTs` must be a valid TiDB timestamp, which is `(physicalTs << 18 + logicalTs)`.
2. `lastSendResolvedTsTime` is the time at which this pipeline staged has sent a Resolved Event. It is a clock time, not directly related to the TiDB timestamp.
3. There is no need to consider timezone, because 1) TiDB timestamps are timezone-less 2) Golang's `time.Time` takes care of timezone internally.

cdc/processor/pipeline/sorter.go
liuzix: Yes, sure

cdc/processor/pipeline/table.go
liuzix: The flow controller is actually per-table. Each table has its own quota. This is enough for our purpose now because we have yet to encounter any user who experienced OOM due to large amount of data on a large number of tables.

Sharing the quota among the tables will create very complex deadlock hazards, whose safety would require extensive proofs to demonstrate.

cdc/kv/client.go
leoppro: ping @liuzix

cdc/sink/common/flow_control.go
```diff
@@ -0,0 +1,222 @@
...
+	c.Consumed += nBytes
+	return nil
+}
+
+// Release is called when a chuck of memory is done being used.
+func (c *TableMemorySizeController) Release(nBytes uint64) {
+	c.mu.Lock()
+
+	if c.Consumed < nBytes {
+		defer c.mu.Unlock()
```
amyangfei, line 108:
```suggestion
		c.mu.Unlock()
```

pkg/config/config.go
```diff
@@ -160,7 +160,8 @@ var defaultServerConfig = &ServerConfig{
 		NumWorkerPoolGoroutine: 16,
 		SortDir:                "/tmp/cdc_sort",
 	},
-	Security: &SecurityConfig{},
+	Security:            &SecurityConfig{},
+	PerTableMemoryQuota: 20 * 1024 * 1024, // 20MB
```
amyangfei, line 164: Is `20MB` too small?

amyangfei:
Since the memory quota is table based, is there any plan to add two level memory quota, where the top level is a global resource manager, the second level is table based resource manager

pkg/config/config.go
liuzix: The `20M` is for RawKV, not taking into consideration the data structures generated by the mounter.

liuzix:
> Since the memory quota is table based, is there any plan to add two level memory quota, where the top level is a global resource manager, the second level is table based resource manager

Given our current design, we need at least have table-level checkpoint before introducing any kind of global flow control.

pkg/config/config.go
amyangfei: Could you add benchmark about the throughput of the flow controller with `20MB` as memory quota

liuzix: Done

cdc/sink/common/flow_control.go
overvenus: Could you add some comments?

cdc/kv/client.go
```diff
@@ -53,11 +53,13 @@ import (
...
-	tikvRequestMaxBackoff     = 20000   // Maximum total sleep time(in ms)
-	grpcInitialWindowSize     = 1 << 27 // 128 MB The value for initial window size on a stream
-	grpcInitialConnWindowSize = 1 << 27 // 128 MB The value for initial window size on a connection
+	dialTimeout           = 10 * time.Second
+	maxRetry              = 100
+	tikvRequestMaxBackoff = 20000 // Maximum total sleep time(in ms)
+	// TODO find optimal values and test extensively before releasing
+	// The old values cause the gRPC stream to stall for some unknown reason.
+	grpcInitialWindowSize     = 1 << 28 // 256 MB The value for initial window size on a stream
+	grpcInitialConnWindowSize = 1 << 29 // 512 MB The value for initial window size on a connection
```
overvenus, line 62: I'm afraid gRPC may consume too much memory when there are multiple TiKV (N). Total memory = N * 512MB.

cdc/processor/pipeline/sorter.go
```diff
@@ -94,14 +107,80 @@ func (n *sorterNode) Init(ctx pipeline.NodeContext) error {
...
+
 		for {
 			select {
 			case <-stdCtx.Done():
 				return nil
+			case <-metricsTicker.C:
+				metricsTableMemoryGauge.Set(float64(n.flowController.GetConsumption()))
 			case msg := <-sorter.Output():
-				if msg == nil {
+				if msg == nil || msg.RawKV == nil {
```
JinLingChristopher, line 133: Shall call `ctx.SendToNextNode(pipeline.PolymorphicEventMessage(msg))` before continue ?

cdc/sink/common/flow_control.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/edwingeng/deque"
+	"github.com/pingcap/errors"
+	cerrors "github.com/pingcap/ticdc/pkg/errors"
+	"go.uber.org/zap"
+)
+
+// TableMemorySizeController is designed to curb the total memory consumption of processing
+// the event streams in a table.
+// A higher-level controller more suitable for direct use by the processor is TableFlowController.
+type TableMemorySizeController struct {
```
overvenus, line 30:
```suggestion
type TableMemoryQuota struct {
```

It feels like a memory quota.

cdc/processor/pipeline/table.go
```diff
@@ -31,6 +32,12 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	// TODO determine a reasonable default value
+	// This is part of sink performance optimization
+	resolvedTsInterpolateInterval = 200 * time.Millisecond
```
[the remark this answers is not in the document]
liuzix: No. This only inserts extra resolved events. It does not delay the event stream or remove any thing from the stream.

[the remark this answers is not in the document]
liuzix: This actually improves the sink's performance, for unknown reason though.

cdc/kv/client.go
liuzix: I might be a better idea to sort this out in a future PR, because when I ran tests, a small gRPC window size is more likely to cause problems such as sudden decrease in throughput.

amyangfei:
please fix ci

cdc/processor/pipeline/table.go
```diff
@@ -154,15 +171,23 @@ func NewTablePipeline(ctx context.Context,
...
+	log.Debug("creating table flow controller",
+		zap.String("changefeed-id", changefeedID),
+		zap.String("table-name", tableName),
+		zap.Int64("table-id", tableID),
+		zap.Uint64("quota", perTableMemoryQuota))
+	flowController := common.NewTableFlowController(perTableMemoryQuota)
 	p := pipeline.NewPipeline(ctx, 500*time.Millisecond)
-	p.AppendNode(ctx, "puller", newPullerNode(changefeedID, credential, kvStorage, limitter, tableID, replicaInfo, tableName))
-	p.AppendNode(ctx, "sorter", newSorterNode(sortEngine, sortDir, changefeedID, tableName, tableID))
+	tablePipeline.pullerNode = newPullerNode(changefeedID, credential, kvStorage, limitter, tableID, replicaInfo, tableName)
```
leoppro, line 182: noone call `pullerNode.ResolvedTs`

cdc/processor/pipeline/sink.go
```diff
@@ -204,6 +210,7 @@ func (n *sinkNode) Receive(ctx pipeline.NodeContext) error {
 					zap.Uint64("stoppedTs", msg.Command.StoppedTs), zap.Uint64("checkpointTs", n.checkpointTs))
 			}
 			n.targetTs = msg.Command.StoppedTs
+			n.flowController.Abort()
```
leoppro, line 213: maybe we can call `n.flowController.Abort()` in `Destroy` function?

leoppro (commented):
rest LGTM

PR #1682, opened 2021-04-20
kv/client: fix force reconnect in client v2

amyangfei:

- Fix a potential deadlock when a reconnect is performed
- Fix un initialized regions are not reconnected.

note release-5.0 should be picked manually.

- Reuse the `resolvedTsManager` to manage the time of last event in each un-initialized region, rename `resolvedTsManager` to `regionTsManager`
- Separate the reconnect check of un-initialized regions from the lock resolver routine to reduce lock contention.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/kv/region_worker.go
+++ b/cdc/kv/region_worker.go
@@ -208,9 +208,6 @@ func (w *regionWorker) handleSingleRegionError(ctx context.Context, err error, s
 func (w *regionWorker) checkUnInitRegions(ctx context.Context) error {
 	checkInterval := time.Minute
 
-	failpoint.Inject("kvClientReconnectInterval", func(val failpoint.Value) {
-		reconnectInterval = time.Duration(val.(int)) * time.Second
-	})
 	failpoint.Inject("kvClientCheckUnInitRegionInterval", func(val failpoint.Value) {
 		checkInterval = time.Duration(val.(int)) * time.Second
 	})
@@ -255,9 +252,6 @@ func (w *regionWorker) resolveLock(ctx context.Context) error {
 	resolveLockInterval := 20 * time.Second
 	failpoint.Inject("kvClientResolveLockInterval", func(val failpoint.Value) {
 		resolveLockInterval = time.Duration(val.(int)) * time.Second
-	})
-	failpoint.Inject("kvClientReconnectInterval", func(val failpoint.Value) {
-		reconnectInterval = time.Duration(val.(int)) * time.Second
 	})
 	advanceCheckTicker := time.NewTicker(time.Second * 5)
 	defer advanceCheckTicker.Stop()
--- a/cdc/kv/region_worker.go
+++ b/cdc/kv/region_worker.go
@@ -237,7 +237,7 @@ func (w *regionWorker) checkUnInitRegions(ctx context.Context) error {
 					break
 				}
 				state, ok := w.getRegionState(item.regionID)
-				if !ok || state.isStopped() || state.IsInitialized() {
+				if !ok || state.isStopped() || state.isInitialized() {
 					// check state is deleted, stopped, or initialized, if
 					// so just ignore this region, and don't need to push the
 					// eventTimeItem back to heap.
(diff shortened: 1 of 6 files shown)


cdc/kv/region_worker.go
```diff
@@ -390,6 +446,17 @@ func (w *regionWorker) handleEventEntry(
 			}
 			metricPullEventInitializedCounter.Inc()
 			state.initialized = true
+			select {
+			case w.rtsUpdateCh <- &regionResolvedTs{regionID: regionID, resolvedTs: state.sri.ts}:
+			default:
+				// rtsUpdateCh block often means too many regions are suffering
+				// lock resolve, the kv client status is not very healthy.
+				log.Warn("region is not upsert into rts manager", zap.Uint64("region-id", regionID))
```
overvenus, line 454: What is "upsert"?

amyangfei: It is a special syntax in some RDBMS, which means `insert or update`, ref: https://en.wikipedia.org/wiki/Merge_(SQL)

liuzix:
What will happen if TiKV pushes an old resolved ts (the resolved ts for the incremental scan) immediately after the Initialize event? This will be the case if you use TiKV master branch.

amyangfei:
> What will happen if TiKV pushes an old resolved ts (the resolved ts for the incremental scan) immediately after the Initialize event? This will be the case if you use TiKV master branch.

I'm afraid there exist some corner cases

1. TiKV keeps sending unchanged resolved ts, kv client v1 can't detect it now, since kv client v1 checks `last receive event time` only. But kv client v2 can detect it because kv client checks `lastResovledTs` of initialized region
2. TiKV pushes an old resolved ts after the initialized event. kv client v1 has not false positive, yet kv client v2 could reconnect region 

From 1 we know use the `last receive event time` is not enough, while use `lastResolvedTs` check mechanism could get false positive.

amyangfei:
> What will happen if TiKV pushes an old resolved ts (the resolved ts for the incremental scan) immediately after the Initialize event? This will be the case if you use TiKV master branch.

Besides, Does the `old resolved ts` equal to the `start ts` of this region? @liuzix

amyangfei:
> > What will happen if TiKV pushes an old resolved ts (the resolved ts for the incremental scan) immediately after the Initialize event? This will be the case if you use TiKV master branch.
> 
> I'm afraid there do exist some corner cases
> 
> 1. TiKV keeps sending unchanged resolved ts, kv client v1 can't detect it now, since kv client v1 checks `last receive event time` only. But kv client v2 can detect it because kv client checks `lastResovledTs` of initialized region
> 2. TiKV pushes an old resolved ts after the initialized event. kv client v1 has not false positive, yet kv client v2 could reconnect region
> 
> From 1 we know use the `last receive event time` is not enough, while use `lastResolvedTs` check mechanism could get false positive.

Also add event time in resovled ts item and solve problem 2, the problem 1 exists now. This PR will not solve unchanged resolved ts problem

liuzix:
> > What will happen if TiKV pushes an old resolved ts (the resolved ts for the incremental scan) immediately after the Initialize event? This will be the case if you use TiKV master branch.
> 
> Besides, Does the `old resolved ts` equal to the `start ts` of this region? @liuzix

No. It equals the end of the range of incremental scan.

cdc/kv/region_worker.go
```diff
@@ -195,10 +263,6 @@ func (w *regionWorker) resolveLock(ctx context.Context) error {
 		case rtsUpdate := <-w.rtsUpdateCh:
 			w.rtsManager.Upsert(rtsUpdate)
 		case <-advanceCheckTicker.C:
-			failpoint.Inject("kvClientForceReconnect", func() {
-				log.Warn("kv client reconnect triggered by failpoint")
-				failpoint.Return(errReconnect)
-			})
 			if !w.session.isPullerInit.IsInitialized() {
 				// Initializing a puller may take a long time, skip resolved lock to save unnecessary overhead.
 				continue
```
liuzix, line 268: Even if the puller has not been initialized, if TiKV is buggy enough, it is possible for an already initialized region to stop sending events. Are we handling this possibility?

amyangfei: - If puller is not initialized, there must be uninitialized regions remaining, some of which must have no more events and will be reconnected
- Yet this check can be removed, since client v2 has separated initialized regions and uninitialized regions.

amyangfei: For the scenario-a above, there could be one or more uninitialized regions keep uninitialized and reconnect (caused by epoch not match or something else), which will lead to initialized regions can't be reconnected. (However in this scenario, even reconnect for the initialized regions may not work)

But still IMHO we don't need to check whether puller is initialized any more before checking the initialized regions.

PR #1684, opened 2021-04-21
cdc/owner.go: add a TiCDC ttl strategy

asddongmen:
Fix the problem that TiKV GC safe time is blocked indefinitely due to TiCDC changefeed checkpoint stagnation.

Set a threshold Y: Regardless of the status of the changefeed, TiCDC is allowed to block TiKV GC safe time for Y hours at most.

The threshold Y is set in the config/config.go file as a variable named gc-TTL and the default value is 84000 seconds (24 hours).

### Check List 

Tests 
 - Unit test

step1: set  `defaultServerConfig.gcTTL = 20` in /pkg/config.go 

step2: add ` log.Warn("try to update gcSafePoint", zap.Time("gcSafePoint", oracle.GetTimeFromTS(gcSafePoint)))`  at line 780 in cdc/owner.go to print log info

step3: In a shell window type `tiup playground` to deploy a tidb cluster.

step4: Move to ticdc directory, and type `./bin/cdc server` to to start the cdc component。

step5: Type `./cdc cli changefeed create --pd=http://127.0.0.1:2379 --sink-uri="mysql://root@127.0.0.1:4000/" --changefeed-id="ttl-test-10"` to create a changefeed.

step6: pause above changefeed, you would see gcSafePoint stop update:
image

step7: gcSafePoint update again after stop 20s:
image

Code changes

Side effects

Related changes

### Release note
 - No release note

### DIFF
--- a/cdc/http_handler.go
+++ b/cdc/http_handler.go
@@ -41,6 +41,8 @@ const (
 	APIOpVarTableID = "table-id"
 	// APIOpForceRemoveChangefeed is used when remove a changefeed
 	APIOpForceRemoveChangefeed = "force-remove"
+	// APIOpVarGcBLT is the key of gcTTL in HTTP API
+	APIOpVarGcBLT = "gc-blt"
 )
 
 type commonResp struct {
@@ -284,4 +286,34 @@ func handleAdminLogLevel(w http.ResponseWriter, r *http.Request) {
 	log.Warn("log level changed", zap.String("level", level))
 
 	writeData(w, struct{}{})
+}
+
+// Use to change gcBLT
+func (s *Server) handleConfigGCTimeToBlock(w http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodPost {
+		writeError(w, http.StatusBadRequest, cerror.ErrSupportPostOnly.GenWithStackByArgs())
+		return
+	}
+	s.ownerLock.Lock()
+	defer s.ownerLock.Unlock()
+	if s.owner == nil {
+		handleOwnerResp(w, concurrency.ErrElectionNotLeader)
+		return
+	}
+
+	err := req.ParseForm()
+	if err != nil {
+		writeInternalServerError(w, cerror.WrapError(cerror.ErrInternalServerError, err))
+		return
+	}
+
+	gcBLTStr := req.Form.Get(APIOpVarGcBLT)
+	gcBLT, err := strconv.ParseInt(gcBLTStr, 10, 64)
+	if err != nil {
+		writeError(w, http.StatusBadRequest,
+			cerror.ErrAPIInvalidParam.GenWithStack("invalid gcBLT: %s", gcBLTStr))
+		return
+	}
+	s.owner.setGcBLT(gcBLT)
+	handleOwnerResp(w, nil)
 }
--- a/cdc/http_handler.go
+++ b/cdc/http_handler.go
@@ -41,8 +41,6 @@ const (
 	APIOpVarTableID = "table-id"
 	// APIOpForceRemoveChangefeed is used when remove a changefeed
 	APIOpForceRemoveChangefeed = "force-remove"
-	// APIOpVarGcTTL is the key of gcTTL in HTTP API
-	APIOpVarGcTTL = "gc-ttl"
 )
 
 type commonResp struct {
@@ -286,34 +284,4 @@ func handleAdminLogLevel(w http.ResponseWriter, r *http.Request) {
 	log.Warn("log level changed", zap.String("level", level))
 
 	writeData(w, struct{}{})
-}
-
-// Use to change gcTTL
-func (s *Server) handleConfigGCTimeToBlock(w http.ResponseWriter, req *http.Request) {
-	if req.Method != http.MethodPost {
-		writeError(w, http.StatusBadRequest, cerror.ErrSupportPostOnly.GenWithStackByArgs())
-		return
-	}
-	s.ownerLock.Lock()
-	defer s.ownerLock.Unlock()
-	if s.owner == nil {
-		handleOwnerResp(w, concurrency.ErrElectionNotLeader)
-		return
-	}
-
-	err := req.ParseForm()
-	if err != nil {
-		writeInternalServerError(w, cerror.WrapError(cerror.ErrInternalServerError, err))
-		return
-	}
-
-	gcTTLStr := req.Form.Get(APIOpVarGcTTL)
-	gcTTL, err := strconv.ParseInt(gcTTLStr, 10, 64)
-	if err != nil {
-		writeError(w, http.StatusBadRequest,
-			cerror.ErrAPIInvalidParam.GenWithStack("invalid gcTTL: %s", gcTTLStr))
-		return
-	}
-	s.owner.setGcTTL(gcTTL)
-	handleOwnerResp(w, nil)
 }
--- a/cdc/http_handler.go
+++ b/cdc/http_handler.go
@@ -41,8 +41,8 @@ const (
 	APIOpVarTableID = "table-id"
 	// APIOpForceRemoveChangefeed is used when remove a changefeed
 	APIOpForceRemoveChangefeed = "force-remove"
-	// APIOpVarGcBLT is the key of gcTTL in HTTP API
-	APIOpVarGcBLT = "gc-blt"
+	// APIOpVarGcTTL is the key of gcTTL in HTTP API
+	APIOpVarGcTTL = "gc-ttl"
 )
 
 type commonResp struct {
@@ -288,7 +288,7 @@ func handleAdminLogLevel(w http.ResponseWriter, r *http.Request) {
 	writeData(w, struct{}{})
 }
 
-// Use to change gcBLT
+// Use to change gcTTL
 func (s *Server) handleConfigGCTimeToBlock(w http.ResponseWriter, req *http.Request) {
 	if req.Method != http.MethodPost {
 		writeError(w, http.StatusBadRequest, cerror.ErrSupportPostOnly.GenWithStackByArgs())
@@ -307,13 +307,13 @@ func (s *Server) handleConfigGCTimeToBlock(w http.ResponseWriter, req *http.Requ
 		return
 	}
 
-	gcBLTStr := req.Form.Get(APIOpVarGcBLT)
-	gcBLT, err := strconv.ParseInt(gcBLTStr, 10, 64)
+	gcTTLStr := req.Form.Get(APIOpVarGcTTL)
+	gcTTL, err := strconv.ParseInt(gcTTLStr, 10, 64)
 	if err != nil {
 		writeError(w, http.StatusBadRequest,
-			cerror.ErrAPIInvalidParam.GenWithStack("invalid gcBLT: %s", gcBLTStr))
+			cerror.ErrAPIInvalidParam.GenWithStack("invalid gcTTL: %s", gcTTLStr))
 		return
 	}
-	s.owner.setGcBLT(gcBLT)
+	s.owner.setGcTTL(gcTTL)
 	handleOwnerResp(w, nil)
 }
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -734,13 +734,12 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 		snapshot := make(map[model.ChangeFeedID]*model.ChangeFeedStatus, len(o.changeFeeds))
 		for id, changefeed := range o.changeFeeds {
 			snapshot[id] = changefeed.status
-			if changefeed.appliedCheckpointTs < gcSafePoint {
-				gcSafePoint = changefeed.appliedCheckpointTs
+			if changefeed.status.CheckpointTs < gcSafePoint {
+				gcSafePoint = changefeed.status.CheckpointTs
 			}
-
 			// if changefeed's appliedCheckpoinTs < minGcSafePoint, means this changefeed is stagnant
 			// They are collected into this queue, and then a function is called to deal with these stagnant changefeed.
-			if changefeed.appliedCheckpointTs < minGcSafePoint {
+			if changefeed.status.CheckpointTs < minGcSafePoint {
 				staleChangeFeedId = append(staleChangeFeedId, id)
 			}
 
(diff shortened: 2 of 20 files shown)


cdc/http_handler.go
```diff
@@ -41,6 +41,8 @@ const (
 	APIOpVarTableID = "table-id"
 	// APIOpForceRemoveChangefeed is used when remove a changefeed
 	APIOpForceRemoveChangefeed = "force-remove"
+	// APIOpVarGcBLT is the key of gcTTL in HTTP API
```
leoppro, line 44:
```suggestion
	// APIOpVarGcBLT is the key of gcBLT in HTTP API
```

cdc/owner.go
```diff
@@ -750,6 +754,9 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 		}
 	}
 	if time.Since(o.gcSafepointLastUpdate) > GCSafepointUpdateInterval {
+		if (time.Now().Unix() - int64(minCheckpointTs)) > o.gcBLT {
+			minCheckpointTs = uint64(time.Now().Unix() - o.gcBLT)
```
leoppro, line 758: the name of `minCheckpointTs` is not accurate any more, may be we can call it `gcSafePoint`

cdc/owner.go
```diff
@@ -1626,3 +1633,9 @@ func (o *Owner) startCaptureWatcher(ctx context.Context) {
 		}
 	}()
 }
+
+// reset gcBLT
+func (o *Owner) setGcBLT(gcBLT int64) {
+	o.gcBLT = gcBLT
```
leoppro, line 1639: Is there any data race here?

cdc/owner.go
```diff
@@ -151,6 +154,7 @@ func NewOwner(
 		cfRWriter:               cli,
 		etcdClient:              cli,
 		gcTTL:                   gcTTL,
+		gcBLT:                   gcBLT,
```
leoppro, line 157: may be `maxGCSafeTimeDelay` is a better name?

leoppro:
please check and fix the failed CI

cdc/owner.go
leoppro: or `maxGCBlockTime`

leoppro: We do not promote the use of abbreviations that are not conventional

cdc/owner.go
```diff
@@ -1626,3 +1633,9 @@ func (o *Owner) startCaptureWatcher(ctx context.Context) {
 		}
 	}()
 }
+
+// reset gcBLT
+func (o *Owner) setGcBLT(gcBLT int64) {
+	o.gcBLT = gcBLT
+	log.Warn("change gc-ttl", zap.Int64("gc-ttl", o.gcBLT))
```
leoppro, line 1640: `gc-ttl`? type?

cdc/http_handler.go
```diff
@@ -285,3 +287,33 @@ func handleAdminLogLevel(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	err := req.ParseForm()
+	if err != nil {
+		writeInternalServerError(w, cerror.WrapError(cerror.ErrInternalServerError, err))
+		return
+	}
+
+	gcBLTStr := req.Form.Get(APIOpVarGcBLT)
```
leoppro, line 310: In design, the input of `/server/config` should be a JSON, but not an HTTPRequestForm

cdc/owner.go
asddongmen: In `http_handler.go` the caller function `handleConfigGCTimeToBlock` must acquire  a write lock before calling the function, thus avoiding data race here.

cdc/owner.go
asddongmen: Thanks, I will change it name to `maxGCBlockTime`. It is a better name.

pkg/config/config.go
```diff
@@ -144,6 +144,7 @@ var defaultServerConfig = &ServerConfig{
 	LogFile:       "",
 	LogLevel:      "info",
 	GcTTL:         24 * 60 * 60, // 24H
+	GcBLT:         24 * 60 * 60, // 24H
```
leoppro, line 147: Should we just reuse the `GcTTL` parameter? If we reuse it, the GC safe time delay thresholds will be uniform regardless of whether the CDC server is running or not.

cdc/owner.go
amyangfei, line 758: `time.Now()` could have clock drift, it is not safe to operate `time.Now().Unix()` with a tso. We can query tso from PD here.

cdc/owner.go
asddongmen: You are right... I will correct it.  thankssssss!

cdc/owner.go
```diff
@@ -750,6 +754,9 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 		}
 	}
 	if time.Since(o.gcSafepointLastUpdate) > GCSafepointUpdateInterval {
+		if (time.Now().Unix() - int64(minCheckpointTs)) > o.gcBLT {
```
leoppro, line 757: what happens when the checkpoint delay exceeds `gcBLT` in a running changefeed?

cdc/owner.go
asddongmen: ok, I will change it.

cdc/http_handler.go
asddongmen: ok.

cdc/owner.go
JinLingChristopher: `gcSafePoint` or `gcSafepoint`, keep all the same...

cdc/owner.go
```diff
@@ -750,6 +754,9 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
...
+			minCheckpointTs = uint64(time.Now().Unix() - o.gcBLT)
+		}
 		actual, err := o.pdClient.UpdateServiceGCSafePoint(ctx, CDCServiceSafePointID, o.gcTTL, minCheckpointTs)
```
amyangfei, line 760: Please add comments about how the `service gc safepoint` is decided in every single scenario, and add unit tests for each of these scenarios

pkg/config/config.go
leoppro: @amyangfei @liuzix

asddongmen: I think it is a good idea.

cdc/owner.go
asddongmen: Thanks for your comment, I will correct it.

pkg/config/config.go
liuzix: Agreed

cdc/owner.go
asddongmen: ok~

cdc/http_handler.go
liuzix: Will this be an official public API?

cdc/owner.go
liuzix: Note that our tso is `milliseconds-since-epoch << 18 + logical-ts`

pkg/config/config.go
```diff
@@ -172,6 +173,7 @@ type ServerConfig struct {
 	LogLevel string `toml:"log-level" json:"log-level"`
 
 	GcTTL int64  `toml:"gc-ttl" json:"gc-ttl"`
+	GcBLT int64  `toml:"gc-blt" json:"gc-blt"`
```
liuzix, line 176: Actually, is it better for this to be a global config item? It should be associated with the entire TiCDC cluster, but not any single node, nor any single changefeed. Just a thought.

cdc/owner.go
liuzix: Note that you need to fail any running changefeed whose checkpoint is behind your newly generated `minCheckpointTs`, before you can safely update the safepoint in PD. Otherwise there's a chance that data loss will occur.

overvenus:
Is there any test covers your changes?

cdc/owner.go
```diff
@@ -717,14 +717,31 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
...
+	gcSafePoint := uint64(math.MaxUint64)
+
+	// init the lower bound of gcSafePoint, the logical part is zero.
+	minGcSafePoint := oracle.EncodeTSO(oracle.GetPhysical(time.Now()) - (o.gcTTL * 1000))
+	// Try to get physical and logical timestamp from pd
+	pyTs, lgTs, err := o.pdClient.GetTS(ctx)
+	if err != nil {
+		log.Warn("failed to acquire time from pd, will use this machine time", zap.Error(err))
+	} else {
+		minGcSafePoint = oracle.ComposeTS(pyTs-(o.gcTTL*1000), lgTs)
```
overvenus, line 730: Could you add some comments about the magic number 1000?

cdc/owner.go
```diff
@@ -717,14 +717,31 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
...
+	}
 
-	minCheckpointTs := uint64(math.MaxUint64)
 	if len(o.changeFeeds) > 0 {
 		snapshot := make(map[model.ChangeFeedID]*model.ChangeFeedStatus, len(o.changeFeeds))
 		for id, changefeed := range o.changeFeeds {
 			snapshot[id] = changefeed.status
-			if changefeed.appliedCheckpointTs < minCheckpointTs {
-				minCheckpointTs = changefeed.appliedCheckpointTs
+			if changefeed.appliedCheckpointTs < gcSafePoint {
```
overvenus, line 737: What if `appliedCheckpointTs` is zero?

cdc/owner.go
```diff
@@ -1626,3 +1658,27 @@ func (o *Owner) startCaptureWatcher(ctx context.Context) {
...
 }
+
+// handle the StaleChangeFeed
+// By setting the AdminJob type to AdminStop and the Error code to indicate that the changeFeed is stagnant
+func (o *Owner) handleStaleChangeFeed(ctx context.Context, staleChangeFeedId []model.ChangeFeedID, minGcSafePoint uint64) error {
+	for _, id := range staleChangeFeedId {
+
+		runningError := &model.RunningError{
+			Addr:    util.CaptureAddrFromCtx(ctx),
+			Code:    string(cerror.ErrStartTsBeforeGC.RFCCode()), // changFeed is stagnant
```
overvenus, line 1669:
```suggestion
			Code:    string(cerror.ErrStartTsBeforeGC.RFCCode()), // changefeed is stagnant
```

cdc/owner.go
```diff
@@ -1626,3 +1658,27 @@ func (o *Owner) startCaptureWatcher(ctx context.Context) {
 		}
 	}()
 }
+
+// handle the StaleChangeFeed
+// By setting the AdminJob type to AdminStop and the Error code to indicate that the changeFeed is stagnant
```
overvenus, line 1663:
```suggestion
// By setting the AdminJob type to AdminStop and the Error code to indicate that the changefeed is stagnant
```

leoppro:
please fix lint

cdc/processor/doc.go
```diff
@@ -30,7 +30,5 @@ One Capture(with processor role)  -> Processr Manager -> Processor(changefeed1)
                                                           -> Processor(changefeed2) -> TablePipeline(tableC)
                                                                                     ╲
                                                                                      -> TablePipeline(tableD)
-
-
```
leoppro, line 34: restore this file please

asddongmen (commented):
> Is there any test covers your changes?

I added a unit test to test these changes.

cdc/owner.go
```diff
@@ -717,14 +717,31 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 		}
 		return nil
 	}
+	staleChangeFeedID := make([]model.ChangeFeedID, 0)
+	gcSafePoint := uint64(math.MaxUint64)
+
+	// init the lower bound of gcSafePoint, the logical part is zero.
+	minGcSafePoint := oracle.EncodeTSO(oracle.GetPhysical(time.Now()) - (o.gcTTL * 1000))
+	// Try to get physical and logical timestamp from pd
+	pyTs, lgTs, err := o.pdClient.GetTS(ctx)
```
amyangfei, line 726: Since it is safe to use a smaller TSO, can we add a TSO cache such as https://github.com/pingcap/ticdc/blob/109257825f7805bfc0f87b68d9ef7e969aa19161/cdc/kv/store_op.go#L83

Otherwise the `GetTS` could lead to heavy requests to PD, because `flushChangeFeedInfos` could be called frequently.

cdc/owner.go
```diff
@@ -717,14 +717,31 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 		}
 		return nil
 	}
+	staleChangeFeedID := make([]model.ChangeFeedID, 0)
+	gcSafePoint := uint64(math.MaxUint64)
+
+	// init the lower bound of gcSafePoint, the logical part is zero.
+	minGcSafePoint := oracle.EncodeTSO(oracle.GetPhysical(time.Now()) - (o.gcTTL * 1000))
+	// Try to get physical and logical timestamp from pd
```
JinLingChristopher, line 725: pyTs -> physicalTs, lgTs -> logicTs.

then you can delete comment.

cdc/owner.go
asddongmen: Very reasonable suggestion. I've added a cache.

cdc/owner.go
```diff
@@ -766,11 +818,10 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
...
 			actual = uint64(val.(int))
 		})
-
-		if actual > minCheckpointTs {
+		if actual > gcSafePoint {
 			// UpdateServiceGCSafePoint has failed.
-			log.Warn("updating an outdated service safe point", zap.Uint64("checkpoint-ts", minCheckpointTs), zap.Uint64("actual-safepoint", actual))
-
+			// log.Warn("updating an outdated service safe point", zap.Uint64("checkpoint-ts", gcSafePoint), zap.Uint64("actual-safepoint", actual))
+			log.Warn("updating an outdated service safe point", zap.Time("checkpoint-ts", oracle.GetTimeFromTS(gcSafePoint)), zap.Time("actual-safepoint", oracle.GetTimeFromTS(actual)))
```
amyangfei, line 824: Please remove the comment line. Besides use `zap.Uint64("checkpoint-ts", gcSafePoint), zap.Uint64("actual-safepoint", actual)` is more accurate

amyangfei (commented):
rest lgtm

cdc/owner.go
asddongmen: ok, I've restored it.

Issue #1599, opened 2021-04-01
region not receiving event from tikv for too long time cause the batch resolvedts event is dropped


leoppro:
## Bug Report
```
[2021/04/01 15:19:42.978 +08:00] [WARN] [client.go:1299] ["region not found"] [regionID=1069410]
[2021/04/01 15:19:43.440 +08:00] [WARN] [client.go:1400] ["region not receiving event from tikv for too long time"] [regionID=1069410] [span="[7480000000000000ff4b5f728000000001ff6e32d50000000000fa, 7480000000000000ff4b5f728000000001ff6e4cdf0000000000fa)"] [duration=2m47.71997726s]
[2021/04/01 15:19:43.443 +08:00] [WARN] [client.go:1411] ["region not receiving resolved event from tikv or resolved ts is not pushing for too long time, try to resolve lock"] [regionID=1069410] [span="[7480000000000000ff4b5f728000000001ff6e32d50000000000fa, 7480000000000000ff4b5f728000000001ff6e4cdf0000000000fa)"] [duration=21m50.8s] [resolvedTs=423955076771741710]
[2021/04/01 15:19:43.445 +08:00] [INFO] [lock_resolver.go:122] ["resolve lock successfully"] [regionID=1069410] [maxVersion=423955417768656896]
[2021/04/01 15:19:48.439 +08:00] [WARN] [client.go:1400] ["region not receiving event from tikv for too long time"] [regionID=1069410] [span="[7480000000000000ff4b5f728000000001ff6e32d50000000000fa, 7480000000000000ff4b5f728000000001ff6e4cdf0000000000fa)"] [duration=2m52.719338532s]
[2021/04/01 15:19:48.439 +08:00] [WARN] [client.go:1411] ["region not receiving resolved event from tikv or resolved ts is not pushing for too long time, try to resolve lock"] [regionID=1069410] [span="[7480000000000000ff4b5f728000000001ff6e32d50000000000fa, 7480000000000000ff4b5f728000000001ff6e4cdf0000000000fa)"] [duration=21m54.8s] [resolvedTs=423955076771741710]
[2021/04/01 15:19:48.441 +08:00] [INFO] [lock_resolver.go:122] ["resolve lock successfully"] [regionID=1069410] [maxVersion=423955418817232896]
[...]

PR #1720, opened 2021-04-28
kv/client: use ephemeral goroutine to handle region range request

amyangfei:

Fix part bug found in #1599

- The region request is processed in a block way

https://github.com/pingcap/ticdc/blob/109257825f7805bfc0f87b68d9ef7e969aa19161/cdc/kv/client.go#L578-L591

- The `divideAndSendEventFeedToRegions` function scans range into several regions, and dispatch each region via `scheduleRegionRequest`. There could be stale regions, and these stale regions will be resent to `s.requestRangeCh` in L646. If the `s.requestRangeCh` is full, deadlock happens.

https://github.com/pingcap/ticdc/blob/109257825f7805bfc0f87b68d9ef7e969aa19161/cdc/kv/client.go#L627-L648

https://github.com/pingcap/ticdc/blob/109257825f7805bfc0f87b68d9ef7e969aa19161/cdc/kv/client.go#L616-L623

In order to consume region range request from `requestRangeCh` as soon as possible, we create a new goroutine to handle it. The sequence of region range we process is not matter, the region lock keeps the region access sequence. Besides the count or frequency of range request is limitted, we use ephemeral goroutine instead of permanent gourotine.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Fix a deadlock bug in the kv client that deadlock happens when large stale regions exist.


overvenus (commented):
/lgtm

amyangfei:
ptal @liuzix

amyangfei:
ptal @leoppro

leoppro (approved):

PR #1731, opened 2021-04-30
owner: Modified the update strategy of  gcSafePoint

asddongmen:
TiKV GC safe point will be blocked indefinitely due to TiCDC changefeed checkpoint stagnation.

Set a threshold Y: Regardless of the status of the changefeed, TiCDC is allowed to block TiKV GC safe time for Y hours at most.

The threshold Y is set in the config/config.go file as a variable named gc-TTL and the default value is 84000 seconds (24 hours).

### Check List 

Tests 
 - Unit test

step1: set  `defaultServerConfig.gcTTL = 20` in /pkg/config.go 

step2: add ` log.Warn("try to update gcSafePoint", zap.Time("gcSafePoint", oracle.GetTimeFromTS(gcSafePoint)))`  below line 802 in cdc/owner.go to print log info

step3: In a shell window type `tiup playground` to deploy a tidb cluster.

step4: Move to ticdc directory, and type `./bin/cdc server` to to start the cdc component。

step5: Type `./cdc cli changefeed create --pd=http://127.0.0.1:2379 --sink-uri="mysql://root@127.0.0.1:4000/" --changefeed-id="ttl-test-10"` to create a changefeed.

step6: pause above changefeed, you would see gcSafePoint stop update:
image

step7: gcSafePoint update again after stop 20s:
image

Code changes

Side effects

Related changes

### Release note

- Modified the update strategy of gcSafePoint.  Fix the problem that TiKV GC safe point is blocked indefinitely due to TiCDC changefeed checkpoint stagnation.

### DIFF
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -756,8 +756,11 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 			if changefeed.status.CheckpointTs < gcSafePoint {
 				gcSafePoint = changefeed.status.CheckpointTs
 			}
-			// if changefeed's appliedCheckpoinTs < minGCSafePoint, it means this changefeed is stagnant
-			// They are collected into this map, and then a function is called to deal with these stagnant changefeed.
+			// If changefeed's appliedCheckpoinTs < minGCSafePoint, it means this changefeed is stagnant.
+			// They are collected into this map, and then handleStaleChangeFeed() is called to deal with these stagnant changefeed.
+			// A changefeed will not enter the map twice, because in run(),
+			// handleAdminJob() will always be executed before flushChangeFeedInfos(),
+			// ensuring that the previous changefeed in staleChangeFeeds has been stopped and removed from o.changeFeeds.
 			if changefeed.status.CheckpointTs < minGCSafePoint {
 				staleChangeFeeds[id] = changefeed.status
 			}
@@ -781,8 +784,9 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 	}
 
 	for _, status := range o.stoppedFeeds {
-		// if a stopped changefeed's CheckpoinTs < minGCSafePoint, means this changefeed is stagnant
-		// it should never be resume, so here we can skip it
+		// If a stopped changefeed's CheckpoinTs < minGCSafePoint, means this changefeed is stagnant.
+		// It should never be resumed. This part of the logic is in newChangeFeed()
+		// So here we can skip it.
 		if status.CheckpointTs < minGCSafePoint {
 			continue
 		}
@@ -1138,11 +1142,6 @@ func (o *Owner) handleAdminJob(ctx context.Context) error {
 			cfInfo, err := o.etcdClient.GetChangeFeedInfo(ctx, job.CfID)
 			if err != nil {
 				return errors.Trace(err)
-			}
-			// refuse to resume adminJob that checkPointTs lag behind gcSafePoint
-			if o.minGCSafePointCache.ts != 0 && cfInfo.StartTs < o.minGCSafePointCache.ts {
-				log.Warn("invalid admin job, changefeed checkpoint lag behind gcSafePoint", zap.String("changefeed", job.CfID))
-				continue
 			}
 
 			// set admin job in changefeed status to tell owner resume changefeed
@@ -1683,15 +1682,15 @@ func (o *Owner) startCaptureWatcher(ctx context.Context) {
 }
 
 // handle the StaleChangeFeed
-// By setting the AdminJob type to AdminStop and the Error code to indicate that the changefeed is stagnant
+// By setting the AdminJob type to AdminStop and the Error code to indicate that the changefeed is stagnant.
 func (o *Owner) handleStaleChangeFeed(ctx context.Context, staleChangeFeeds map[model.ChangeFeedID]*model.ChangeFeedStatus, minGCSafePoint uint64) error {
 	for id, status := range staleChangeFeeds {
-		message := cerror.ErrSnapshotLostCauseByGC.GenWithStackByArgs(status.CheckpointTs, minGCSafePoint).Error()
+		message := cerror.ErrSnapshotLostByGC.GenWithStackByArgs(status.CheckpointTs, minGCSafePoint).Error()
 		log.Warn("handle staleChangeFeed ", zap.String("changefeed", id), zap.String("Error message", message))
 
 		runningError := &model.RunningError{
 			Addr:    util.CaptureAddrFromCtx(ctx),
-			Code:    string(cerror.ErrSnapshotLostCauseByGC.RFCCode()), // changfeed is stagnant
+			Code:    string(cerror.ErrSnapshotLostByGC.RFCCode()), // changfeed is stagnant
 			Message: message,
 		}
 
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -77,22 +77,20 @@ type minGCSafePointCacheEntry struct {
 	lastUpdated time.Time
 }
 
-func getGcMinSafePointCache(ctx context.Context, pdClient pd.Client, gcTTL int64) (model.Ts, error) {
-	if time.Now().After(minGCSafePointCache.lastUpdated.Add(MinGCSafePointCacheUpdateInterval)) {
-		physicalTs, logicalTs, err := pdClient.GetTS(ctx)
+func (o *Owner) getGcMinSafePointCache(ctx context.Context) (model.Ts, error) {
+	log.Warn("Time", zap.Time("lastUpdated", o.minGCSafePointCache.lastUpdated))
+	log.Warn("After", zap.Bool("lastupdate time after 2 second", time.Now().After(o.minGCSafePointCache.lastUpdated.Add(MinGCSafePointCacheUpdateInterval))))
+	if time.Now().After(o.minGCSafePointCache.lastUpdated.Add(MinGCSafePointCacheUpdateInterval)) {
+		log.Warn("get ts from pd client")
+		physicalTs, logicalTs, err := o.pdClient.GetTS(ctx)
 		if err != nil {
 			return 0, err
 		}
-		minGCSafePointCache.ts = oracle.ComposeTS(physicalTs-(gcTTL*1000), logicalTs)
-		minGCSafePointCache.lastUpdated = time.Now()
+		o.minGCSafePointCache.ts = oracle.ComposeTS(physicalTs-(o.gcTTL*1000), logicalTs)
+		o.minGCSafePointCache.lastUpdated = time.Now()
 	}
-	return minGCSafePointCache.ts, nil
+	return o.minGCSafePointCache.ts, nil
 }
-
-// This value stores the ts obtained from PD and is updated every MinGCSafePointCacheUpdateInterval.
-var (
-	minGCSafePointCache minGCSafePointCacheEntry
-)
 
 // Owner manages the cdc cluster
 type Owner struct {
@@ -129,6 +127,8 @@ type Owner struct {
 	gcTTL int64
 	// last update gc safepoint time. zero time means has not updated or cleared
 	gcSafepointLastUpdate time.Time
+	// stores the ts obtained from PD and is updated every MinGCSafePointCacheUpdateInterval.
+	minGCSafePointCache minGCSafePointCacheEntry
 	// record last time that flushes all changefeeds' replication status
 	lastFlushChangefeeds    time.Time
 	flushChangefeedInterval time.Duration
@@ -746,7 +746,7 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 	gcSafePoint := uint64(math.MaxUint64)
 
 	// get the lower bound of gcSafePoint
-	minGCSafePoint, err := getGcMinSafePointCache(ctx, o.pdClient, o.gcTTL)
+	minGCSafePoint, err := o.getGcMinSafePointCache(ctx)
 	if err != nil {
 		log.Warn("failed to acquire minGCSafePoint from Cache, will use this machine time", zap.Error(err))
 		minGCSafePoint = oracle.EncodeTSO(oracle.GetPhysical(time.Now()) - (o.gcTTL * 1000))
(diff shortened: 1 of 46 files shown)


cdc/owner_test.go
```diff
@@ -175,6 +184,89 @@ func (s *ownerSuite) TestOwnerFlushChangeFeedInfosFailed(c *check.C) {
...
 
+// Test whether the owner handles the stagnant task correctly, so that it can't block the update of gcSafePoint.
+// If a changefeed is put into the stop queue due to stagnation, it can no longer affect the update of gcSafePoint.
+// So we just need to test whether the stagnant changefeed is put into the stop queue.
+func (s *ownerSuite) TestOwnerHandleStaleChangeFeed(c *check.C) {
+	defer testleak.AfterTest(c)()
+	mockPDCli := &mockPDClient{}
+
+	changeFeeds := map[model.ChangeFeedID]*changeFeed{
+		"test_change_feed_1": {
```
overvenus, line 195: Could you add a changefeed whose checkpoint is 0?

overvenus (commented):
rest LGTM

cdc/owner_test.go
asddongmen: ok, I will try.

cdc/owner.go
```diff
@@ -72,6 +72,28 @@ func (o *ownership) inc() {
 	}
 }
 
+type minGcSafePointCacheEntry struct {
```
liuzix, line 75:
```suggestion
type minGCSafePointCacheEntry struct {
```

cdc/owner.go
```diff
@@ -1626,3 +1684,28 @@ func (o *Owner) startCaptureWatcher(ctx context.Context) {
...
+// handle the StaleChangeFeed
+// By setting the AdminJob type to AdminStop and the Error code to indicate that the changefeed is stagnant
+func (o *Owner) handleStaleChangeFeed(ctx context.Context, staleChangeFeedID []model.ChangeFeedID, minGcSafePoint, gcSafePoint uint64) error {
+	for _, id := range staleChangeFeedID {
+		message := cerror.ErrStartTsBeforeGC.GenWithStackByArgs(minGcSafePoint, gcSafePoint).Error()
+		log.Info("handle staleChangeFeed ", zap.String("changefeed", id), zap.String("Error message", message))
+
+		runningError := &model.RunningError{
+			Addr:    util.CaptureAddrFromCtx(ctx),
+			Code:    string(cerror.ErrStartTsBeforeGC.RFCCode()), // changfeed is stagnant
```
liuzix, line 1697: `ErrStartTsBeforeGC` might not be the most appropriate error code. We would like the user to get more useful information.

cdc/owner.go
```diff
@@ -72,6 +72,26 @@ func (o *ownership) inc() {
...
 }
 
+type minGCSafePointCacheEntry struct {
+	ts          model.Ts
+	lastUpdated time.Time
+}
+
+func (o *Owner) getGcMinSafePointCache(ctx context.Context) (model.Ts, error) {
+	log.Warn("Time", zap.Time("lastUpdated", o.minGCSafePointCache.lastUpdated))
+	log.Warn("After", zap.Bool("lastupdate time after 2 second", time.Now().After(o.minGCSafePointCache.lastUpdated.Add(MinGCSafePointCacheUpdateInterval))))
```
amyangfei, line 82: useless log?

cdc/owner.go
```diff
@@ -72,6 +72,26 @@ func (o *ownership) inc() {
...
+	if time.Now().After(o.minGCSafePointCache.lastUpdated.Add(MinGCSafePointCacheUpdateInterval)) {
+		log.Warn("get ts from pd client")
```
amyangfei, line 84:
```suggestion
```

cdc/owner_test.go
```diff
@@ -22,6 +22,11 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
+
+	"github.com/pingcap/tidb/store/tikv/oracle"
+
```
lonng, line 29: Please optimize the imports.

cdc/owner_test.go
```diff
@@ -124,6 +139,7 @@ func (s *ownerSuite) TestOwnerFlushChangeFeedInfos(c *check.C) {
 }
 
 func (s *ownerSuite) TestOwnerFlushChangeFeedInfosFailed(c *check.C) {
+	log.Warn("Begin", zap.String("Test", "TestOwnerFlushChangeFeedInfosFailed"))
```
lonng, line 142: Don't use the `log` package in the test cases. The `check.C` has an `Log/Logf` interface to output some information. BTW, I don't think we need this log.

cdc/owner.go
ben1009, line 82: the logs are used for debug? if not, if get the warn log what is the next move.  **zap.Bool("lastupdate time after 2 second"** , may be should not hardcode to 2, MinGCSafePointCacheUpdateInterval will not always be 2 second. btw why set MinGCSafePointCacheUpdateInterval to 2 instead of 3/4

cdc/owner_test.go
asddongmen: Okay, I print the log here to troubleshoot why the unit test failed, and then forgot to delete these lines.

cdc/owner.go
asddongmen: Yes, I forgot to delete it.

asddongmen: Yes, it is for debug. I forgot to delete it.

cdc/owner_test.go
asddongmen: ok

cdc/owner.go
```diff
@@ -874,6 +926,7 @@ func (o *Owner) dispatchJob(ctx context.Context, job model.AdminJob) error {
 	// For `AdminResume`, we remove stopped feed in changefeed initialization phase.
 	// For `AdminRemove`, we need to update stoppedFeeds when removing a stopped changefeed.
 	if job.Type == model.AdminStop {
+		log.Info("put changfeed into stoppedFeeds queue", zap.String("changefeed", job.CfID))
```
liuzix, line 926: maybe we should use `log.Debug`?

cdc/owner.go
```diff
@@ -1089,6 +1142,11 @@ func (o *Owner) handleAdminJob(ctx context.Context) error {
 			if err != nil {
 				return errors.Trace(err)
 			}
+			// refuse to resume adminJob that checkPointTs lag behind gcSafePoint
+			if cfInfo.Error != nil && cfInfo.Error.Code == string(cerror.ErrSnapshotLostCauseByGC.RFCCode()) {
```
liuzix, line 1143: This `cfInfo.Error` can be cleared. We should not rely on it.

cdc/owner.go
```diff
@@ -744,13 +782,27 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
...
+			continue
+		}
+
+		if status.CheckpointTs < gcSafePoint {
+			gcSafePoint = status.CheckpointTs
 		}
 	}
+
+	// handle stagnant changefeed collected above
+	err = o.handleStaleChangeFeed(ctx, staleChangeFeeds, minGCSafePoint)
```
liuzix, line 796: It is possible for a changefeed to be added to `staleChangeFeeds` multiple times before it stops?

liuzix (commented):
Could you add some integration tests to see if a stale changefeed can be safely stopped without data loss?

cdc/owner_test.go
asddongmen: The package `github.com/pingcap/tidb/store/tikv/oracle`  needs to be imported. Otherwise I have to write a method to get the physical time from the TS.

liuzix: We should avoid empty lines among the imports, unless it's the one separating standard lib imports and custom/third-party imports.

asddongmen: ok, I get it.

cdc/owner.go
asddongmen: This problem may occur, and I am working on it.

cdc/owner.go
asddongmen: You are right.

cdc/owner.go
asddongmen: I re-checked the entire process and found that this problem did not occur.

asddongmen: I re-checked the entire process and found that this problem will not occur.

cdc/owner.go
```diff
@@ -1089,6 +1139,11 @@ func (o *Owner) handleAdminJob(ctx context.Context) error {
 			if err != nil {
 				return errors.Trace(err)
 			}
+			// refuse to resume adminJob that checkPointTs lag behind gcSafePoint
+			if o.minGCSafePointCache.ts != 0 && cfInfo.StartTs < o.minGCSafePointCache.ts {
```
liuzix, line 1143: refer to `GetCheckpointTs` in `model/changefeed.go` for how to get the correct checkpoint

cdc/owner.go
```diff
@@ -1089,6 +1139,11 @@ func (o *Owner) handleAdminJob(ctx context.Context) error {
...
+				log.Warn("invalid admin job, changefeed checkpoint lag behind gcSafePoint", zap.String("changefeed", job.CfID))
```
liuzix, line 1144: It would be more user-friendly if the error message in changefeed status were updated. Otherwise, the user might not understand what is wrong.

pkg/errors/errors.go
```diff
@@ -178,6 +178,7 @@ var (
 	ErrServiceSafepointLost         = errors.Normalize("service safepoint lost. current safepoint is %d, please remove all changefeed(s) whose checkpoints are behind the current safepoint", errors.RFCCodeText("CDC:ErrServiceSafepointLost"))
 	ErrUpdateServiceSafepointFailed = errors.Normalize("updating service safepoint failed", errors.RFCCodeText("CDC:ErrUpdateServiceSafepointFailed"))
 	ErrStartTsBeforeGC              = errors.Normalize("fail to create changefeed because start-ts %d is earlier than GC safepoint at %d", errors.RFCCodeText("CDC:ErrStartTsBeforeGC"))
+	ErrSnapshotLostCauseByGC        = errors.Normalize("fail to create or maintain changefeed because snapshot lost cause by GC, checkpoint-ts %d is earlier than GC safepoint at %d", errors.RFCCodeText("CDC:ErrSnapshotLostCauseByGC"))
```
liuzix, line 181:
```suggestion
	ErrSnapshotLostByGC        = errors.Normalize("fail to create or maintain changefeed due to snapshot loss caused by GC. checkpoint-ts %d is earlier than GC safepoint at %d", errors.RFCCodeText("CDC:ErrSnapshotLostCauseByGC"))
```

cdc/owner.go
```diff
@@ -1626,3 +1680,29 @@ func (o *Owner) startCaptureWatcher(ctx context.Context) {
 		}
 	}()
 }
+
+// handle the StaleChangeFeed
+// By setting the AdminJob type to AdminStop and the Error code to indicate that the changefeed is stagnant.
+func (o *Owner) handleStaleChangeFeed(ctx context.Context, staleChangeFeeds map[model.ChangeFeedID]*model.ChangeFeedStatus, minGCSafePoint uint64) error {
+	for id, status := range staleChangeFeeds {
+		message := cerror.ErrSnapshotLostByGC.GenWithStackByArgs(status.CheckpointTs, minGCSafePoint).Error()
+		log.Warn("handle staleChangeFeed ", zap.String("changefeed", id), zap.String("Error message", message))
```
liuzix, line 1689: A more user-friendly warning would be better. For example, `changefeed checkpoint is lagging too much, so that ...`.

cdc/owner.go
```diff
@@ -718,13 +739,30 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
...
+				gcSafePoint = changefeed.status.CheckpointTs
+			}
+			// If changefeed's appliedCheckpoinTs < minGCSafePoint, it means this changefeed is stagnant.
+			// They are collected into this map, and then handleStaleChangeFeed() is called to deal with these stagnant changefeed.
+			// A changefeed will not enter the map twice, because in run(),
+			// handleAdminJob() will always be executed before flushChangeFeedInfos(),
+			// ensuring that the previous changefeed in staleChangeFeeds has been stopped and removed from o.changeFeeds.
+			if changefeed.status.CheckpointTs < minGCSafePoint {
+				staleChangeFeeds[id] = changefeed.status
 			}
```
amyangfei, line 766: If the `CheckpointTs` of changefeed is less than `minGCSafePoint`, `gcSafePoint` is still set to a value less than `minGCSafePoint`?

asddongmen: Yes. If this happens, in order to prevent the running `changefeed` data loss , the `gcSafePoint` will be set to `CheckpointTs` in first. 
In the same time,  the `changefeed` will be put into `staleChangeFeeds`.  
The update time of `gcSafePoint` will actually be delayed until the next `flushChangeFeedInfos()` excution.
In the next time `flushChangeFeedInfos()` is excuted the `changefeed` is stopped, the value of `gcSafePoint` will really push forward.
For summary, `minGCSafePoints` just a lower bound for judging whether a `changefeed` is stagnant, and the value of `gcSafePoint` does not directly depend on it.

pkg/errors/errors.go
```diff
@@ -178,7 +178,7 @@ var (
 	ErrServiceSafepointLost         = errors.Normalize("service safepoint lost. current safepoint is %d, please remove all changefeed(s) whose checkpoints are behind the current safepoint", errors.RFCCodeText("CDC:ErrServiceSafepointLost"))
 	ErrUpdateServiceSafepointFailed = errors.Normalize("updating service safepoint failed", errors.RFCCodeText("CDC:ErrUpdateServiceSafepointFailed"))
 	ErrStartTsBeforeGC              = errors.Normalize("fail to create changefeed because start-ts %d is earlier than GC safepoint at %d", errors.RFCCodeText("CDC:ErrStartTsBeforeGC"))
```
leoppro, line 180: could we remove this error?

cdc/owner.go
```diff
@@ -718,13 +739,30 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
...
 	}
 
-	minCheckpointTs := uint64(math.MaxUint64)
+	staleChangeFeeds := make(map[model.ChangeFeedID]*model.ChangeFeedStatus, len(o.changeFeeds))
+	gcSafePoint := uint64(math.MaxUint64)
+
+	// get the lower bound of gcSafePoint
+	minGCSafePoint, err := o.getGcMinSafePointCache(ctx)
+	if err != nil {
+		log.Warn("failed to acquire minGCSafePoint from Cache, will use this machine time", zap.Error(err))
```
leoppro, line 748: if we get tso failed from pd, we can just skip this tick

overvenus:
Could you add a release note in the PR description?

asddongmen:
> Could you add a release note in the PR description?

done.

PR #1638, opened 2021-04-12
sink/mysql: explicit_defaults_for_timestamp compatibility with mysql

amyangfei:
### What problem does this PR solve? 

Fix #1585 

### What is changed and how it works?

- Set session variable `explicit_defaults_for_timestamp` to `ON` in MySQL sink, which will keep the same behavior between upstream TiDB and downstream MySQL 5.7.
- Note only set this session variable when the value `explicit_defaults_for_timestamp`  in downstream is `OFF`, because in TiDB variable `explicit_defaults_for_timestamp` is readonly.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation

### Release note

- Set session variable `explicit_defaults_for_timestamp` to `ON` to make downstream MySQL5.7 keeps the same behavior with upstream TiDB.

PR #1743, opened 2021-05-08
Revert "sink/mysql: explicit_defaults_for_timestamp compatibility with mysql (#1638)"

amyangfei:

This reverts commit 6d4d4e4faead09c404a00b7a29ea94bb1133ce67.

Since even set session variable `explicit_defaults_for_timestamp` in MySQL 5.7 requires the `SUPER` privilege, which blocks many usage scenarios.

revert #1638, user should pay attention to the `explicit_defaults_for_timestamp` config when they replicate data from to MySQL 5.7. 

### Release note

- Revert the update for explicit_defaults_for_timestamp which requires `SUPER` privilege when replicating to MySQL.

### DIFF
--- a/cdc/sink/mysql.go
+++ b/cdc/sink/mysql.go
@@ -69,8 +69,6 @@ const (
 // SyncpointTableName is the name of table where all syncpoint maps sit
 const syncpointTableName string = "syncpoint_v1"
 
-const tidbVersionString string = "TiDB"
-
 var validSchemes = map[string]bool{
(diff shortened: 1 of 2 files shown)


leoppro (approved):

amyangfei:
> Could you add a test that checks TiCDC does not send SUPER privilege SQL?

@overvenus  This PR only reverts changes, more tests and document update is tracked in https://github.com/pingcap/ticdc/issues/1748

PR #1747, opened 2021-05-08
etcd_worker: support update multi-keys in a DataPatch

leoppro:

- add a package doc
- support update multi-keys in a DatePatch
  - before this pr, a DataPatch only can update one key-value pair
  - after this pr, a DataPatch can read/update more than one key-value pairs
    - In some cases, the behavior of a DataPatch depends on multiple key-values pairs, so we need to read more than one key-value pairs in one DataPatch
- remove useless files： files in package `jsonstate`

### Check List 

Tests 

 - Unit test
 - Integration test
### Release note

- No release note


pkg/orchestrator/etcd_worker.go
```diff
@@ -223,81 +222,50 @@ func (worker *EtcdWorker) syncRawState(ctx context.Context) error {
...
-				return old, nil
-			},
-		})
+func (worker *EtcdWorker) cloneRawState() map[util.EtcdKey][]byte {
+	ret := make(map[util.EtcdKey][]byte)
+	for k, v := range worker.rawState {
+		ret[k] = v
 	}
-	return result
+	return ret
```
amyangfei, line 230: What is the purpose of this clone function, the underlying byte slice is not deep cloned.

pkg/orchestrator/etcd_worker.go
```diff
@@ -223,81 +222,52 @@ func (worker *EtcdWorker) syncRawState(ctx context.Context) error {
...
-	return bytes.Equal(left, right)
-}
-
-func (worker *EtcdWorker) applyPatches(ctx context.Context, patches []*DataPatch) error {
-	patches = mergePatch(patches)
-	cmps := make([]clientv3.Cmp, 0, len(patches))
-	ops := make([]clientv3.Op, 0, len(patches))
-
+func (worker *EtcdWorker) applyPatches(ctx context.Context, patches []DataPatch, session *concurrency.Session) error {
+	stete := worker.cloneRawState()
```
JinLingChristopher, line 236: `stete` might be a typo - -···

leoppro: yes

leoppro: fixed

ben1009:
could you add some context about why need the change, thx

lichunzhu (commented):
/lgtm

PR #1766, opened 2021-05-12
new_owner: a schema wrapper for owner

leoppro:
implement a schema wrapper for owner

the new owner push DDL job into this schema wrapper, and get table info from the schema wrapper 

### Check List 

Tests 

 - Unit test

### Release note

- No release note


cdc/entry/schema_test_helper.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2021 PingCAP, Inc.
```
amyangfei, line 1: Can exclude this file from coverage calculation

cdc/owner/schema.go
```diff
@@ -0,0 +1,158 @@
...
+		config:         config,
+	}, nil
+}
+
+// AllPhysicalTables returns the table IDs of all tables and partition tables.
+func (s *schemaWrap4Owner) AllPhysicalTables() []model.TableID {
+	if s.allPhysicalTablesCache != nil {
+		return s.allPhysicalTablesCache
+	}
+	tables := s.schemaSnapshot.CloneTables()
```
amyangfei, line 68: Can we expose `tables` from `schemaSnapshot` directly, there is no need to copy it.
Besides, `CloneTables` only returns table IDs from `s.tables`, it doesn't count the `partitionTable`, is this a bug?

amyangfei (approved):

PR #1772, opened 2021-05-13
old_processor: fix processor cancelation

liuzix:
- The processor may fail to exit if canceled from outside by `context.Context`.

- Added an error type `ErrProcessorTableCanceled`.
- Report all errors other than `ErrProcessorTableCanceled` in the processor's subcomponents

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

### Release note

- Fix bugs during processor cancellations.


cdc/processor.go
```diff
@@ -290,7 +291,7 @@ func (p *oldProcessor) Run(ctx context.Context) {
 // wait blocks until all routines in processor are returned
 func (p *oldProcessor) wait() {
 	err := p.wg.Wait()
-	if err != nil && errors.Cause(err) != context.Canceled {
+	if err != nil && cerror.ErrProcessorTableCanceled.NotEqual(err) {
```
amyangfei, line 294: should we also filter `context.Canceled`, otherwise the log bumps too much `context canceled` error when cdc exits normally

PR #1773, opened 2021-05-13
context: uniform the import naming of context, part 1

leoppro:
all context of std lib should be named as `context`
all context of cdc package should be named as `cdcContext`

this pr modified the code in `ticdc/cdc/processor/pipeline` package.

### Check List 

Tests 

 - Unit test

### Release note

- No release note

### DIFF
--- a/cdc/processor/pipeline/sink.go
+++ b/cdc/processor/pipeline/sink.go
@@ -53,11 +53,13 @@ func (s TableStatus) String() string {
 	return "Unknown"
 }
 
-func (s *TableStatus) load() TableStatus {
+// Load TableStatus with THREAD-SAFE
+func (s *TableStatus) Load() TableStatus {
 	return TableStatus(atomic.LoadInt32((*int32)(s)))
 }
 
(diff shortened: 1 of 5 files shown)


cdc/processor/pipeline/sink.go
```diff
@@ -53,11 +53,13 @@ func (s TableStatus) String() string {
...
+// Load TableStatus with THREAD-SAFE
+func (s *TableStatus) Load() TableStatus {
 	return TableStatus(atomic.LoadInt32((*int32)(s)))
 }
 
-func (s *TableStatus) store(new TableStatus) {
+// Store TableStatus with THREAD-SAFE
+func (s *TableStatus) Store(new TableStatus) {
 	atomic.StoreInt32((*int32)(s), int32(new))
 }
```
amyangfei, line 64: Will `Load` and `Store` be used by other packages?

leoppro: yes, in new owner

PR #1784, opened 2021-05-17
cmd: hide --sort-dir in changefeed command. (deprecated warning exists)

amyangfei:

Hide `--sort-dir` in changefeed command

- Hide `--sort-dir` in changefeed command
- Set `SortDir` to `` `json:"-"` `` in changefeed, so the `changefeed query` won't print any `sort-dir` which may confuse user. Note only set it to an ignore field is **backward compatibility**

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


leoppro (approved):

lonng (commented):
/lgtm

PR #1787, opened 2021-05-17
config: hidden the ddl-allow-list config entry

leoppro:
hidden the ddl-allow-list config entry

Tests 

 - Unit test
### Release note

- No release note


amyangfei (approved):

pkg/config/filter.go
```diff
@@ -23,5 +23,5 @@ type FilterConfig struct {
 	Rules []string `toml:"rules" json:"rules"`
 	*filter.MySQLReplicationRules
 	IgnoreTxnStartTs []uint64           `toml:"ignore-txn-start-ts" json:"ignore-txn-start-ts"`
-	DDLAllowlist     []model.ActionType `toml:"ddl-allow-list" json:"ddl-allow-list"`
+	DDLAllowlist     []model.ActionType `toml:"ddl-allow-list" json:"ddl-allow-list,omitempty"`
```
lonng, line 26: I think we should omit the field always not just when `ddl-allow-list` empty.

leoppro: but if the we set `json:"-"` for `ddl-allow-list`, the `DDLAllowlist` will be not persistent in ETCD, so this will make `ddl-allow-list` ineffective and break the compatibility

lonng: Can we implement the fmt.Stringer interface to hide this field when output logs.

leoppro: If a user already sets the `ddl-allow-list`, it's no need to hide `ddl-allow-list` for him, because he has already known it. On the other hand, we need to observe `ddl-allow-list` in the log for debugging.
If a user doesn't know `ddl-allow-list`, the config entry will be default value (nil), so he can't know ``ddl-allow-list``

PR #1819, opened 2021-05-24
new_owner: implement gc manager

leoppro:
GC manager: a module of the new owner, which reports the min checkpoint ts to PD GC service and checks the changefeed which of checkpoint ts is behind too far to stop 

### Check List 

Tests 

 - Unit test
 
### Release note

- No release note

### DIFF
--- a/cdc/model/changefeed.go
+++ b/cdc/model/changefeed.go
@@ -45,6 +45,7 @@ type FeedState string
 // All FeedStates
 const (
 	StateNormal   FeedState = "normal"
+	StateError    FeedState = "error"
 	StateFailed   FeedState = "failed"
 	StateStopped  FeedState = "stopped"
 	StateRemoved  FeedState = "removed"
--- a/cdc/owner/gc_manager.go
+++ b/cdc/owner/gc_manager.go
@@ -95,6 +95,9 @@ func (m *gcManager) updateGCSafePoint(ctx cdcContext.Context, state *model.Globa
 	failpoint.Inject("InjectActualGCSafePoint", func(val failpoint.Value) {
 		actual = uint64(val.(int))
 	})
+	if actual > minCheckpointTs {
+		log.Warn("update gc safe point failed, the gc safe point is larger than checkpointTs", zap.Uint64("actual", actual), zap.Uint64("checkpointTs", minCheckpointTs))
+	}
 	m.lastSafePointTs = actual
 	m.lastSucceededTime = time.Now()
 	return nil
--- a/cdc/owner/gc_manager.go
+++ b/cdc/owner/gc_manager.go
@@ -1,0 +1,125 @@
+// Copyright 2021 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package owner
+
+import (
+	"math"
+	"time"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/failpoint"
+	"github.com/pingcap/log"
+	"github.com/pingcap/ticdc/cdc/model"
+	"github.com/pingcap/ticdc/pkg/config"
+	cdcContext "github.com/pingcap/ticdc/pkg/context"
+	cerror "github.com/pingcap/ticdc/pkg/errors"
+	"github.com/pingcap/tidb/store/tikv/oracle"
+	"go.uber.org/zap"
+)
+
+const (
+	// cdcServiceSafePointID is the ID of CDC service in pd.UpdateServiceGCSafePoint.
+	cdcServiceSafePointID = "ticdc"
+	pdTimeUpdateInterval  = 10 * time.Minute
+)
+
+// gcSafepointUpdateInterval is the minimual interval that CDC can update gc safepoint
+var gcSafepointUpdateInterval = 1 * time.Minute
+
+type gcManager struct {
+	gcTTL int64
+
+	lastUpdatedTime time.Time
+	lastSucceedTime time.Time
+	lastSafePointTs uint64
+
+	pdTimeCached      time.Time
+	lastUpdatedPdTime time.Time
+}
+
+func newGCManager() *gcManager {
+	serverConfig := config.GetGlobalServerConfig()
+	failpoint.Inject("InjectGcSafepointUpdateInterval", func(val failpoint.Value) {
+		gcSafepointUpdateInterval = time.Duration(val.(int) * int(time.Millisecond))
+	})
+	return &gcManager{
+		gcTTL: serverConfig.GcTTL,
+	}
+}
+
+func (m *gcManager) updateGCSafePoint(ctx cdcContext.Context, state *model.GlobalReactorState) error {
+	if time.Since(m.lastUpdatedTime) < gcSafepointUpdateInterval {
+		return nil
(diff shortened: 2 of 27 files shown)


cdc/owner/gc_manager.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// gcSafepointUpdateInterval is the minimual interval that CDC can update gc safepoint
+var gcSafepointUpdateInterval = 1 * time.Minute
+
+type gcManager struct {
+	gcTTL int64
+
+	lastUpdatedTime time.Time
+	lastSucceedTime time.Time
```
liuzix, line 44:
```suggestion
	lastSucceededTime time.Time
```

cdc/owner/gc_manager.go
```diff
@@ -0,0 +1,125 @@
...
+	lastSafePointTs uint64
+
+	pdTimeCached      time.Time
```
liuzix, line 47:
```suggestion
	pdPhysicalTimeCache      time.Time
```

cdc/owner/gc_manager.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	minCheckpointTs := uint64(math.MaxUint64)
+	for _, cfState := range state.Changefeeds {
+		if cfState.Info == nil {
+			continue
+		}
+		switch cfState.Info.State {
+		case model.StateNormal, model.StateStopped, model.StateError:
+		default:
+			continue
```
liuzix, line 73: Maybe add a debug log here?

cdc/owner/gc_manager.go
```diff
@@ -0,0 +1,125 @@
...
+			zap.Error(err))
+		if time.Since(m.lastSucceedTime) >= time.Second*time.Duration(m.gcTTL) {
+			return cerror.ErrUpdateServiceSafepointFailed.Wrap(err)
+		}
+		return nil
+	}
+	failpoint.Inject("InjectActualGCSafePoint", func(val failpoint.Value) {
+		actual = uint64(val.(int))
+	})
+	m.lastSafePointTs = actual
```
liuzix, line 98: Should we report a failure if `actual > minCheckpointTs`?

cdc/owner/gc_manager.go
```diff
@@ -0,0 +1,125 @@
...
+	physical, logical, err := ctx.GlobalVars().PDClient.GetTS(ctx)
+	if err != nil {
+		return time.Now(), errors.Trace(err)
+	}
+	m.pdTimeCached = oracle.GetTimeFromTS(oracle.ComposeTS(physical, logical))
+	m.lastUpdatedPdTime = time.Now()
+	return m.pdTimeCached, nil
+}
+
+func (m *gcManager) CheckTsTooFarBehindToStop(ctx cdcContext.Context, checkpointTs model.Ts) error {
```
liuzix, line 116:
```suggestion
func (m *gcManager) CheckStaleCheckpointTs(ctx cdcContext.Context, checkpointTs model.Ts) error {
```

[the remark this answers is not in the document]
leoppro: yes, it used by changefeed

cdc/owner/gc_manager.go
leoppro: no need to log this, i think

cdc/owner/gc_manager.go
leoppro: if `actual > minCheckpointTs`, the `CheckTsTooFarBehindToStop` will return false(the check will be not pass), it will lead the changefeed to report an error.

cdc/owner/gc_manager_test.go
```diff
@@ -0,0 +1,164 @@
...
+	// assume that the gc safe point updated one hour ago
+	gcManager.lastUpdatedPdTime = time.Now().Add(-time.Hour)
+	t3, err := gcManager.currentTimeFromPDCached(ctx)
+	c.Assert(err, check.IsNil)
+	c.Assert(t3, check.Equals, gcManager.pdPhysicalTimeCache)
+	// should return new time
+	c.Assert(t3, check.Not(check.Equals), t2)
+}
+
+func (s *gcManagerSuite) TestCheckTsTooFarBehindToStop(c *check.C) {
```
leoppro, line 153:
```suggestion
func (s *gcManagerSuite) TestCheckStaleCheckpointTs(c *check.C) {
```

cdc/owner/gc_manager.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+func (m *gcManager) updateGCSafePoint(ctx cdcContext.Context, state *model.GlobalReactorState) error {
+	if time.Since(m.lastUpdatedTime) < gcSafepointUpdateInterval {
+		return nil
+	}
+	minCheckpointTs := uint64(math.MaxUint64)
+	for _, cfState := range state.Changefeeds {
+		if cfState.Info == nil {
```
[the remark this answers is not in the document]
leoppro: no, it not a valid state

cdc/owner/gc_manager.go
overvenus: Why do we need a log? @liuzix 
And why not? @leoppro

cdc/model/changefeed.go
```diff
@@ -45,6 +45,7 @@ type FeedState string
 // All FeedStates
 const (
 	StateNormal   FeedState = "normal"
+	StateError    FeedState = "error"
```
overvenus, line 48: By adding a new state, does it cause compatibility issues?

cdc/owner/gc_manager.go
```diff
@@ -0,0 +1,125 @@
...
+
+	pdPhysicalTimeCache time.Time
+	lastUpdatedPdTime   time.Time
+}
+
+func newGCManager() *gcManager {
+	serverConfig := config.GetGlobalServerConfig()
+	failpoint.Inject("InjectGcSafepointUpdateInterval", func(val failpoint.Value) {
+		gcSafepointUpdateInterval = time.Duration(val.(int) * int(time.Millisecond))
+	})
```
overvenus, line 55: Why do we need a failpoint? And why does it change a global variable inside a function?

cdc/owner/gc_manager.go
```diff
@@ -0,0 +1,125 @@
...
+	m.lastUpdatedPdTime = time.Now()
+	return m.pdPhysicalTimeCache, nil
+}
+
+func (m *gcManager) CheckStaleCheckpointTs(ctx cdcContext.Context, checkpointTs model.Ts) error {
+	pdTime, err := m.currentTimeFromPDCached(ctx)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if checkpointTs < m.lastSafePointTs || pdTime.Sub(oracle.GetTimeFromTS(checkpointTs)) > time.Duration(m.gcTTL)*time.Second {
```
overvenus, line 121: What will happen when we create a changefeed with a start ts _x_ where gc safepoint is _x-48h_ and gcTTL is _24h_?

leoppro: the owner will stop the changefeed and set the feed state to failed

cdc/owner/gc_manager.go
leoppro: it's port from the old owner, and this failpoint is already used by some test cases

cdc/model/changefeed.go
leoppro: no, it will not cause compatibility issues.

cdc/owner/gc_manager.go
leoppro: Because there is no need to keep a log here, for no particular reason.

liuzix: My thought was that we could use some log to debug the calculation of gcSafePoint in case of unexpected behavior. But on second thought, the log might produce too many lines and it might not be that useful. So NVM.

PR #1847, opened 2021-05-26
kv/client: fix time is not updated when receiving committed event

amyangfei:

Fix a bug in kv client v2 that committed events are not counted in event ts manager

- Always send ts update notification to event ts manager when region receives any event and it is not initialized.
- Remove a failpoint in test case, and we can update global variable directly.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/kv/client_test.go
+++ b/cdc/kv/client_test.go
@@ -2803,11 +2803,11 @@ func (s *etcdSuite) testKVClientForceReconnect(c *check.C) {
 
 	err = failpoint.Enable("github.com/pingcap/ticdc/cdc/kv/kvClientResolveLockInterval", "return(1)")
 	c.Assert(err, check.IsNil)
-	err = failpoint.Enable("github.com/pingcap/ticdc/cdc/kv/kvClientReconnectInterval", "return(3)")
-	c.Assert(err, check.IsNil)
+	originalReconnectInterval := reconnectInterval
+	reconnectInterval = 3 * time.Second
 	defer func() {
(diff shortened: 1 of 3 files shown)


leoppro (approved):

cdc/kv/client_test.go
```diff
@@ -2803,11 +2803,11 @@ func (s *etcdSuite) testKVClientForceReconnect(c *check.C) {
 
 	err = failpoint.Enable("github.com/pingcap/ticdc/cdc/kv/kvClientResolveLockInterval", "return(1)")
 	c.Assert(err, check.IsNil)
-	err = failpoint.Enable("github.com/pingcap/ticdc/cdc/kv/kvClientReconnectInterval", "return(3)")
-	c.Assert(err, check.IsNil)
+	originalReconnectInterval := reconnectInterval
+	reconnectInterval = 3 * time.Second
```
liuzix, line 2807: Is it possible for the race detector to complain about this?

leoppro:
any command to cancel LGTM? /unlgtm

leoppro:
> any command to cancel LGTM? /unlgtm

fine, I have to review this PR ASAP

cdc/kv/client_test.go
amyangfei: No, the test cases run in sequence. (The original code also changes the global value, but triggered by the failpoint)

amyangfei:
> any command to cancel LGTM? /unlgtm

@leoppro try `/lgtm cancel`

Issue #1724, opened 2021-04-29
Fail-fast for unrecoverable DML errors


liuzix:
## Feature Request

**Is your feature request related to a problem? Please describe:**
Users can be frustrated if an unrecoverable error such as `tableNotExists` is not reported in a timely fashion.

**Describe the feature you'd like:**
Instead of retrying, the sink should fail fast when a DML error is likely to be unrecoverable without manual intervention.

**Describe alternatives you've considered:**
Report currently on-going retries.

**Teachability, Documentation, Adoption, Migration Strategy:**
We should take care to ensure that failed changefeeds are periodically retried, in case the user completed the manual intervention and expects the changefeed to continue on its own.

amyangfei:
duplicated with https://github.com/pingcap/ticdc/issues/525

ben1009:
close this one, all prs got merged

PR #1848, opened 2021-05-26
pkg: add a new retry to support the error isRetryable or not

ben1009:
-  intention is  try to solve Fail-fast for unrecoverable DML errors #1724, for now add a new retry pkg, will try to replace the current retry usage later
- add functional options feature to better support extend the retry ability
- return `ctx.Error()` when got canceled instead of just return err, for more detail check https://github.com/cenkalti/backoff/issues/97
- support RunWithInfiniteRetry and RunSpecificTimesRetry in one func with `WithInfiniteTries, WithMaxTries` Option
- change the param from WithMaxRetries to WithMaxTries for better understanding 	
```
        just paste some comments from test

        err := Run(500*time.Millisecond, 3, f)
	c.Assert(err, check.ErrorMatches, "test")
	// It's weird that backoff may retry one more time than maxTries.
	// Because the steps in backoff.Retry is:
	// 1. Call function
	// 2. Compare numTries and maxTries
	// 3. Increment numTries
	c.Assert(callCount, check.Equals, 3+1)
```
add a new implementation of retry to support config error retry or not

### Check List 

Tests 

 - Unit test

Code changes

 - Has new function added

Related changes


pkg/retry/options.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2020 PingCAP, Inc.
```
lonng, line 1:
```suggestion
// Copyright 2021 PingCAP, Inc.
```

pkg/retry/options.go
```diff
@@ -0,0 +1,90 @@
...
+	return &retryOptions{
+		maxTries:    defaultMaxTries,
+		backoffBase: defaultBackoffBaseInMs,
+		backoffCap:  defaultBackoffCapInMs,
+		isRetryable: func(err error) bool { return true },
+	}
+}
+
+// WithBackoffBaseDelay configures the initial delay
+func WithBackoffBaseDelay(delayInMs int64) Option {
```
lonng, line 50: Can we use `time.Duration` to represent the `delayInMs` instead of `int64`?

pkg/retry/retry_with_opt.go
```diff
@@ -0,0 +1,94 @@
+// Copyright 2020 PingCAP, Inc.
```
lonng, line 1:
```suggestion
// Copyright 2021 PingCAP, Inc.
```

pkg/retry/retry_test.go
```diff
@@ -121,3 +126,96 @@ func (s *runSuite) TestInfiniteRetry(c *check.C) {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+		callCount++
+		return errors.New("test")
+	}
+
+	err := Do(ctx, f, WithInfiniteTries(), WithBackoffBaseDelay(2), WithBackoffMaxDelay(10))
+	log.Info("Test_DoCancelInfiniteRetry run f times", zap.Int("tries", callCount))
```
lonng, line 201: It's better to delete the log in the unit tests and use `Commentf` if you want print some extra information.

pkg/retry/retry_with_opt.go
```diff
@@ -0,0 +1,94 @@
...
+
+func run(ctx context.Context, op Operation, retryOption *retryOptions) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+	}
+
+	var t *time.Timer
+	try := 0.0
```
lonng, line 50: It's weird to use a float type number to represent the tried counter.

pkg/retry/retry_with_opt.go
```diff
@@ -0,0 +1,94 @@
...
+
+		try++
+		if try >= retryOption.maxTries {
+			return errors.Wrapf(err, "reach maximum try:%f", retryOption.maxTries)
+		}
+
+		backOff = getBackoffInMs(retryOption.backoffBase, retryOption.backoffCap, try)
+		if t == nil {
+			t = time.NewTimer(backOff)
+			defer t.Stop()
```
lonng, line 70: It's a bad idea to use `defer` in the loop. I think we can use `time.After` function to construct a one-off ticker.

liuzix:
Can we make use of `github.com/cenkalti/backoff`?

pkg/retry/options.go
ben1009: we could, but normally the delay should be covered at ms or s level, so just a int64 seems all right and more convenient  IMO

pkg/retry/retry_with_opt.go
ben1009: i agree with you, but for this case the defer only get called once `if t=nil`, so should be fine. could use `time.After` instead, let gc taking care of `NewTimer`, but i prefer not use time.After in select{}

ben1009:
> Can we make use of `github.com/cenkalti/backoff`?

i also want reuse `github.com/cenkalti/backoff` at the first place, but after check the code, seems rewrite could be  easier than add  the feature i want in the pkg, for instance check the issues https://github.com/cenkalti/backoff/issues

pkg/retry/options.go
```diff
@@ -0,0 +1,90 @@
...
+	defaultBackoffCapInMs = 100.0
+	defaultMaxTries       = 3
+)
+
+// Option ...
+type Option func(*retryOptions)
+
+// retryOptions ...
+type retryOptions struct {
+	maxTries    float64
```
lonng, line 33: Why the `maxTries` is a float type?

pkg/retry/retry_test.go
```diff
@@ -121,3 +124,94 @@ func (s *runSuite) TestInfiniteRetry(c *check.C) {
 	c.Assert(reportedElapsed, check.Greater, time.Second)
 	c.Assert(reportedElapsed, check.LessEqual, 3*time.Second)
 }
+
+func (s *runSuite) Test_DoShouldRetryAtMostSpecifiedTimes(c *check.C) {
```
overvenus, line 128:
```suggestion
func (s *runSuite) TestDoShouldRetryAtMostSpecifiedTimes(c *check.C) {
```

Tips about golang naming, https://golang.org/doc/effective_go#mixed-caps

pkg/retry/retry_test.go
```diff
@@ -121,3 +124,94 @@ func (s *runSuite) TestInfiniteRetry(c *check.C) {
...
+		callCount++
+		return errors.New("test")
+	}
+
+	err := Do(context.Background(), f, WithMaxTries(3))
+	c.Assert(err, check.ErrorMatches, "*test")
+	c.Assert(callCount, check.Equals, 3)
+}
+
+func (s *runSuite) Test_DoShouldStopOnSuccess(c *check.C) {
```
overvenus, line 141:
```suggestion
func (s *runSuite) TestDoShouldStopOnSuccess(c *check.C) {
```

pkg/retry/retry_test.go
```diff
@@ -121,3 +124,94 @@ func (s *runSuite) TestInfiniteRetry(c *check.C) {
...
+		}
+		return errors.New("test")
+	}
+
+	err := Do(context.Background(), f, WithMaxTries(3))
+	c.Assert(err, check.IsNil)
+	c.Assert(callCount, check.Equals, 2)
+}
+
+func (s *runSuite) Test_IsRetryable(c *check.C) {
```
overvenus, line 157:
```suggestion
func (s *runSuite) TestIsRetryable(c *check.C) {
```

pkg/retry/retry_test.go
```diff
@@ -121,3 +124,94 @@ func (s *runSuite) TestInfiniteRetry(c *check.C) {
...
+	c.Assert(callCount, check.Equals, 1)
+
+	callCount = 0
+	err = Do(context.Background(), f, WithMaxTries(3))
+
+	c.Assert(errors.Cause(err), check.Equals, context.Canceled)
+	c.Assert(callCount, check.Equals, 3)
+}
+
+func (s *runSuite) Test_DoCancelInfiniteRetry(c *check.C) {
```
overvenus, line 183:
```suggestion
func (s *runSuite) TestDoCancelInfiniteRetry(c *check.C) {
```

pkg/retry/retry_test.go
```diff
@@ -121,3 +124,94 @@ func (s *runSuite) TestInfiniteRetry(c *check.C) {
...
+		return errors.New("test")
+	}
+
+	err := Do(ctx, f, WithInfiniteTries(), WithBackoffBaseDelay(2), WithBackoffMaxDelay(10))
+	c.Assert(err, check.Equals, context.DeadlineExceeded)
+	c.Assert(callCount, check.GreaterEqual, 1, check.Commentf("tries:%d", callCount))
+	c.Assert(float64(callCount), check.Less, math.Inf(1))
+}
+
+func (s *runSuite) Test_DoCancelAtBeginning(c *check.C) {
```
overvenus, line 204:
```suggestion
func (s *runSuite) TestDoCancelAtBeginning(c *check.C) {
```

pkg/retry/retry_with_opt.go
```diff
@@ -0,0 +1,94 @@
...
+			return nil
+		}
+
+		if !retryOption.isRetryable(err) {
+			return err
+		}
+
+		try++
+		if int64(try) >= retryOption.maxTries {
+			return errors.Wrapf(err, "reach maximum try:%d", retryOption.maxTries)
```
overvenus, line 64:
```suggestion
			return errors.Annotatef(err, "reach maximum try:%d", retryOption.maxTries)
```

TiCDC uses `errors.Annotatef`.

pkg/retry/retry_with_opt.go
```diff
@@ -0,0 +1,94 @@
...
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-t.C:
+		}
+	}
+}
+
+// getBackoffInMs returns the duration to wait before next try
+// See https://www.awsarchitectureblog.com/2015/03/backoff.html
+func getBackoffInMs(backoffBaseInMs, backoffCapInMs, try float64) time.Duration {
```
overvenus, line 85: Could you add tests for try `math.MinInt64`, `-1`, `[0-9]`, `math.MaxInt64`?

overvenus:
What's your next step in solving #1724?

ben1009:
> What's your next step in solving #1724?

will make the change in another pr, since have 10+ usage exclude tests, do not want involve too many changes in one place.
- replace current retry pkg with this one
- config WithIsRetryableErr for differenet cases
-  add more tests

pkg/retry/retry_test.go
ben1009: my intention is to use `_` to separate new pkg tests from old one, i will rm `_` to make you happy 🤡

pkg/retry/options.go
```diff
@@ -0,0 +1,89 @@
...
+	defaultBackoffCapInMs = 100.0
+	defaultMaxTries       = 3
+)
+
+// Option ...
+type Option func(*retryOptions)
+
+type retryOptions struct {
+	maxTries    int64
+	backoffBase float64
```
liuzix, line 33: Specify the unit here? Milliseconds?

pkg/retry/options.go
```diff
@@ -0,0 +1,89 @@
...
+	backoffCap  float64
```
liuzix, line 34: ditto

pkg/retry/options.go
```diff
@@ -0,0 +1,89 @@
...
+	return &retryOptions{
+		maxTries:    defaultMaxTries,
+		backoffBase: defaultBackoffBaseInMs,
+		backoffCap:  defaultBackoffCapInMs,
+		isRetryable: func(err error) bool { return true },
+	}
+}
+
+// WithBackoffBaseDelay configures the initial delay
+func WithBackoffBaseDelay(delayInMs int64) Option {
```
liuzix, line 49: Is zero an acceptable input here? Comment on it?

pkg/retry/options.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Option ...
+type Option func(*retryOptions)
+
+type retryOptions struct {
+	maxTries    int64
+	backoffBase float64
+	backoffCap  float64
+	// isRetryable checks the error is safe to retry or not, eg. "context.Canceled" better not retry
+	isRetryable func(error) bool
```
liuzix, line 36: Maybe we can add a type alias `Pred` for `func(error) bool`?

pkg/retry/options.go
```diff
@@ -0,0 +1,89 @@
...
+func WithMaxTries(tries int64) Option {
+	return func(o *retryOptions) {
+		if tries > 0 {
+			o.maxTries = tries
+		}
+	}
+}
+
+// WithInfiniteTries configures to retry forever (math.MaxInt64 times) till success or got canceled
+func WithInfiniteTries() Option {
```
liuzix, line 76:
```suggestion
func WithInfiniteRetries() Option {
```

pkg/retry/retry_with_opt.go
```diff
@@ -0,0 +1,94 @@
...
+			return nil
+		}
+
+		if !retryOption.isRetryable(err) {
+			return err
+		}
+
+		try++
+		if int64(try) >= retryOption.maxTries {
+			return errors.Annotatef(err, "reach maximum try:%d", retryOption.maxTries)
```
liuzix, line 64: Please add a new error in `pkg/errors`

pkg/retry/retry_with_opt.go
```diff
@@ -0,0 +1,95 @@
...
+// See https://www.awsarchitectureblog.com/2015/03/backoff.html
+func getBackoffInMs(backoffBaseInMs, backoffCapInMs, try float64) time.Duration {
+	temp := int64(math.Min(backoffCapInMs, backoffBaseInMs*math.Exp2(try)) / 2)
+	if temp <= 0 {
+		temp = 1
+	}
+	sleep := temp + rand.Int63n(temp)
+
+	backOff := math.Min(backoffCapInMs, float64(rand.Int63n(sleep*3))+backoffBaseInMs)
+	return time.Duration(math.Min(backoffCapInMs, backOff)) * time.Millisecond
```
amyangfei, line 94: Seems `backOff` has counted `math.Min(backoffCapInMs, xx)` in L93, we don't need to `math.Min` again in L94?

pkg/retry/options.go
```diff
@@ -0,0 +1,89 @@
...
+			o.backoffBaseInMs = float64(delayInMs)
+		}
+	}
+}
+
+// WithBackoffMaxDelay configures the maximum delay, if delayInMs <= 0 "defaultBackoffCapInMs" will be used
+func WithBackoffMaxDelay(delayInMs int64) Option {
+	return func(o *retryOptions) {
+		if delayInMs > 0 {
+			o.backoffCapInMs = float64(delayInMs)
```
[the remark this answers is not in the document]
ben1009: i think what u mean is `backoffCapInMs < backoffBaseInMs`, 
`temp := int64(math.Min(backoffCapInMs, backoffBaseInMs*math.Exp2(try)) / 2))`
will always use `backoffCapInMs` to calculate, kind like after try for 100 times, always return `backoffCapInMs` for `math.Min(backoffCapInMs, backoffBaseInMs*math.Exp2(try)) / 2)`

amyangfei (approved):

ben1009:
/run-integration-tests /tidb=release-5.0 /tikv=release-5.0 /pd=release-5.0 /tiflash=release-5.0

ben1009:
/run-integration-tests /tidb=release-5.0 /tikv=release-5.0 /pd=release-5.0 /tiflash=release-5.0

overvenus (approved):
LGTM

Issue #1853, opened 2021-05-26
High cpu usage with slice malloc and mallocGC


amyangfei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

   1. running a TiCDC cluster, with 2 changefeeds, one is 32 tables, the other is 105 tables.
   2. use sysbench workload in upstream, with 250tps/1500qps of each changefeed
   3. observe cpu usage

2. What did you expect to see?

   normal cpu profile

3. What did you see instead?

    From cpu profile, we found high cpu usage in the slice malloc/gc of `sorter.(*heapSorter).flush`

<img width="1171" alt="Screen Shot 2021-05-26 at 22 27 51" src="https://user-images.githubusercontent.com/1527315/119677821-9e383e00-be71-11eb-9587-0fdacf1ec58b.png">

profile files:

before: 
[cpu.profile.zip](https://github.com/pingcap/ticdc/files/6547436/cpu.profile.zip)

after: 
[cpu.profile.2.zip](https://github.com/pingcap/ticdc/files/6547437/cpu.profile.2.zip)

4. Versions of the cluster

    - TiCDC version (execute `cdc version`):

        master@43e6752791f8f09706a4b29a54a5b2250bd9d2fa

5. workaround

   - After simple change the default slice capacity from 65536 to 16/32, the cpu usage decreases largely

<img width="1139" alt="Screen Shot 2021-05-26 at 22 33 00" src="https://user-images.githubusercontent.com/1527315/119678816-6d0c3d80-be72-11eb-9148-ef11fe556556.png">

   - Besides, we can use a double buffer slice to save memory alloc and dealloc, but we should process with memory more careful.

PR #1854, opened 2021-05-26
sorter: reduce memory malloc to avoid too much CPU overhead

amyangfei:

ref: #1853

~Since the access of `h.heap` is not concurrent, we can use the same `slice`~

The heap will be used in an async pool. We choose a conservative initialize size for sort heap slice.

- the heap is flushed every 100ms or accumulated data size execeeds `1GB`?

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Reduce memory malloc in sort heap to avoid too much CPU overhead.

### DIFF
--- a/cdc/puller/sorter/heap_sorter.go
+++ b/cdc/puller/sorter/heap_sorter.go
@@ -33,7 +33,7 @@ import (
 
 const (
 	flushRateLimitPerSecond = 10
-	sortHeapCapacity        = 65536
+	sortHeapCapacity        = 32
 )
 
 type flushTask struct {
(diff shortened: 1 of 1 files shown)


cdc/puller/sorter/heap_sorter.go
```diff
@@ -159,7 +160,7 @@ func (h *heapSorter) flush(ctx context.Context, maxResolvedTs uint64) error {
 			return nil
 		}
 		oldHeap = h.heap
-		h.heap = make(sortHeap, 0, 65536)
+		h.heap = make(sortHeap, 0, sortHeapCapacity)
```
lonng, line 163: Do we need to allocate a new slice for `heap`? Can we reuse the previous one, eg: `h.heap = h.heap[:0]`

liuzix: No. Flushing the heap to disk is done in a separate goroutine, so we need to keep the original underlying array intact until it's been flushed.

Issue #1839, opened 2021-05-26
Can not create changefeed when start ts is less than current ts - gcttl


overvenus:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

Create a changefeed with a start ts that is close to gc safepoint. 

```
# Change gc lift time to 2400h
mysql -h 172.16.5.33 -P 47904 -u root -E -e 'update mysql.tidb set VARIABLE_VALUE="2400h" where VARIABLE_NAME="tikv_gc_life_time";'

# Wait 24 hours (the default gc-ttl)

tiup cdc:nightly cli --pd="http://172.16.5.33:47902" changefeed create --sink-uri="blackhole://" -c "bl2" \
  --start-ts=$(( $(pd-ctl -u http://0.0.0.0:47902 service-gc-safepoint | jq '.service_gc_safe_points[] | select(.service_id == "gc_worker") | .safe_point') + 1000 )) --no-confirm --sort-engine=unified

tiup cdc:v5.0.0-nightly cli --pd="http://172.16.5.33:47902" changefeed list
The component `cdc` version v5.0.0-nightly is not installed; downloading from repository.
component cdc version v5.0.0-nightly is already installed
Starting component `cdc`: /home/stn/.tiup/components/cdc/v5.0.0-nightly/cdc cli --pd=http://172.16.5.33:47902 changefeed list
[
  {
    "id": "bl1",
    "summary": {
      "state": "normal",
      "tso": 425199630964752529,
      "checkpoint": "2021-05-26 13:44:29.586",
      "error": null
    }
  },
  {
    "id": "bl2",
    "summary": {
      "state": "stopped",
      "tso": 425043878316868600,
      "checkpoint": "2021-05-19 16:42:00.386",
      "error": {
        "addr": "172.16.5.33:47913",
        "code": "CDC:ErrSnapshotLostByGC",
        "message": "[CDC:ErrSnapshotLostByGC]fail to create or maintain changefeed due to snapshot loss caused by GC. checkpoint-ts 425043878316868600 is earlier than GC safepoint at 425177025946845484"
      }
    }
  }
]

pd-ctl tso 425043878316868600
system:  2021-05-19 16:42:00.386 +0800 CST
logic:  1016

pd-ctl tso 425177025946845484
system:  2021-05-25 13:47:18.286 +0800 CST
logic:  300
```

2. What did you expect to see?

Create a changefeed successfully.

[...]

PR #1861, opened 2021-05-27
Owner: Fix gcttl bug

asddongmen:
#1839 

Modify the calculation method of MinGCSafePoint （the lower bound of CDC's gcSafePoint）

Before：
MinGCSafePoint = curentTs - gcTTL

After：
MinGCSafePoint = min(curentTs - gcTTL,  pdSafePoint) 

### Check List 

Tests 

 - Unit test

Side effects

 - Increased code complexity

Related changes

### Release note

- owner:  When tikv_gc_life_time is greater than gcttl, use tikv_gc_life_time to calculate the lower bound of gcSafePoint.

### DIFF
--- a/cdc/owner.go
+++ b/cdc/owner.go
@@ -85,13 +85,9 @@ func (o *Owner) getMinGCSafePointCache(ctx context.Context) model.Ts {
 			return o.minGCSafePointCache.ts
 		}
 		o.minGCSafePointCache.ts = oracle.ComposeTS(physicalTs-(o.gcTTL*1000), logicalTs)
-		// get minimum safepoint across all services
-		actualGCSafePoint, err := o.pdClient.UpdateServiceGCSafePoint(ctx, CDCMinGCSafePointCacheServiceID, CDCMinGCSafePointCacheServiceTTL, o.minGCSafePointCache.ts)
-		if err != nil {
-			log.Warn("Fail to get actualGCSafePoint from PD.", zap.Error(err))
-		}
-		if actualGCSafePoint < o.minGCSafePointCache.ts {
-			o.minGCSafePointCache.ts = actualGCSafePoint
+
+		if o.pdGCSafePoint < o.minGCSafePointCache.ts {
+			o.minGCSafePointCache.ts = o.pdGCSafePoint
 		}
 
 		o.minGCSafePointCache.lastUpdated = time.Now()
@@ -136,6 +132,8 @@ type Owner struct {
 	gcSafepointLastUpdate time.Time
 	// stores the ts obtained from PD and is updated every MinGCSafePointCacheUpdateInterval.
 	minGCSafePointCache minGCSafePointCacheEntry
+	// stores the actual gcSafePoint store in pd
+	pdGCSafePoint model.Ts
 	// record last time that flushes all changefeeds' replication status
 	lastFlushChangefeeds    time.Time
 	flushChangefeedInterval time.Duration
@@ -145,10 +143,6 @@ type Owner struct {
 const (
 	// CDCServiceSafePointID is the ID of CDC service in pd.UpdateServiceGCSafePoint.
 	CDCServiceSafePointID = "ticdc"
-	// CDCMinGCSafePointCacheServiceID is this service GC safe point ID
-	CDCMinGCSafePointCacheServiceID = "ticdc-minGCSafePoint-cache"
-	// CDCMinGCSafePointCacheServiceTTL is this service GC safe point TTL
-	CDCMinGCSafePointCacheServiceTTL = 10 * 60 // 10mins
 	// GCSafepointUpdateInterval is the minimual interval that CDC can update gc safepoint
 	GCSafepointUpdateInterval = time.Duration(2 * time.Second)
 	// MinGCSafePointCacheUpdateInterval is the interval that update minGCSafePointCache
@@ -771,7 +765,7 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 			// A changefeed will not enter the map twice, because in run(),
 			// handleAdminJob() will always be executed before flushChangeFeedInfos(),
 			// ensuring that the previous changefeed in staleChangeFeeds has been stopped and removed from o.changeFeeds.
-			if changefeed.status.CheckpointTs <= minGCSafePoint {
+			if minGCSafePoint != 0 && changefeed.status.CheckpointTs <= minGCSafePoint {
 				staleChangeFeeds[id] = changefeed.status
 			}
 
@@ -797,7 +791,7 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 		// If a stopped changefeed's CheckpoinTs < minGCSafePoint, means this changefeed is stagnant.
 		// It should never be resumed. This part of the logic is in newChangeFeed()
 		// So here we can skip it.
-		if status.CheckpointTs <= minGCSafePoint {
+		if minGCSafePoint != 0 && status.CheckpointTs <= minGCSafePoint {
 			continue
 		}
 
@@ -823,6 +817,7 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 				return cerror.ErrUpdateServiceSafepointFailed.Wrap(err)
 			}
 		} else {
+			o.pdGCSafePoint = actual
 			o.gcSafepointLastUpdate = time.Now()
 		}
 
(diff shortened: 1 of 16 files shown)


cdc/owner.go
```diff
@@ -85,6 +85,15 @@ func (o *Owner) getMinGCSafePointCache(ctx context.Context) model.Ts {
 			return o.minGCSafePointCache.ts
 		}
 		o.minGCSafePointCache.ts = oracle.ComposeTS(physicalTs-(o.gcTTL*1000), logicalTs)
+		// get minimum safepoint across all services
+		actualGCSafePoint, err := o.pdClient.UpdateServiceGCSafePoint(ctx, CDCMinGCSafePointCacheServiceID, CDCMinGCSafePointCacheServiceTTL, o.minGCSafePointCache.ts)
+		if err != nil {
+			log.Warn("Fail to get actualGCSafePoint from PD.", zap.Error(err))
```
liuzix, line 91: Do we need handling the error? Note that `actualGCSafePoint` would be `0` if `err != nil`.

cdc/owner.go
```diff
@@ -136,6 +145,10 @@ type Owner struct {
 const (
 	// CDCServiceSafePointID is the ID of CDC service in pd.UpdateServiceGCSafePoint.
 	CDCServiceSafePointID = "ticdc"
+	// CDCMinGCSafePointCacheServiceID is this service GC safe point ID
+	CDCMinGCSafePointCacheServiceID = "ticdc-minGCSafePoint-cache"
```
liuzix, line 149: Is adding another safe-point the only way to get the current min safe-point?

cdc/owner.go
```diff
@@ -85,6 +85,15 @@ func (o *Owner) getMinGCSafePointCache(ctx context.Context) model.Ts {
 			return o.minGCSafePointCache.ts
 		}
 		o.minGCSafePointCache.ts = oracle.ComposeTS(physicalTs-(o.gcTTL*1000), logicalTs)
+		// get minimum safepoint across all services
+		actualGCSafePoint, err := o.pdClient.UpdateServiceGCSafePoint(ctx, CDCMinGCSafePointCacheServiceID, CDCMinGCSafePointCacheServiceTTL, o.minGCSafePointCache.ts)
```
amyangfei, line 89: - It is not a good semantic to call an `updateXX` API in a `getXX` interface
- Rather than to call `UpdateServiceGCSafePoint` to get min GC safepoint, I would rather to query the gc_safe_point from PD, but seems PD doesn't expose this API directly

asddongmen: Do you mean to add an interface in `cdc/kv/etcd.go` that can get the `gc_safe_point` directly from etcd?

cdc/owner.go
asddongmen: There is no direct way to get it now.

cdc/owner.go
amyangfei: Yep, is using `gc_safe_point` all right here?

asddongmen: I changed the way to get `actualGCSafePoint`.  Please review it again.

cdc/owner.go
```diff
@@ -758,7 +765,7 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 			// A changefeed will not enter the map twice, because in run(),
 			// handleAdminJob() will always be executed before flushChangeFeedInfos(),
 			// ensuring that the previous changefeed in staleChangeFeeds has been stopped and removed from o.changeFeeds.
-			if changefeed.status.CheckpointTs < minGCSafePoint {
+			if minGCSafePoint != 0 && changefeed.status.CheckpointTs <= minGCSafePoint {
```
amyangfei, line 768: We don't need to check `minGCSafePoint != 0` since `changefeed.status.CheckpointTs` is always larger than 0

cdc/owner.go
```diff
@@ -784,7 +791,7 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 		// If a stopped changefeed's CheckpoinTs < minGCSafePoint, means this changefeed is stagnant.
 		// It should never be resumed. This part of the logic is in newChangeFeed()
 		// So here we can skip it.
-		if status.CheckpointTs < minGCSafePoint {
+		if minGCSafePoint != 0 && status.CheckpointTs <= minGCSafePoint {
```
amyangfei, line 794: ditto

cdc/owner.go
asddongmen: - We need to check `minGCSafePoint != 0` here because `o.pdGCSafePoint` may be `0` at the beginning, resulting in `minGCSafePoint` being `0` and will stop all changefeeds.  
- When the `minGCSafePoint` value is `0`, just skip the action to avoid the situation described above.

cdc/owner.go
asddongmen: - We need the `<=` check here is because when a `changefeed` is stagnant, its `checkpointTs` will be updated to `pd`, and it would be the minimum `gcSafePoint` of `cdc`. In order to avoid circular dependence, it is necessary to check for equality.  
- Checking whether they are equal will make the loop assignment to happen only once, and then it will be skipped in next time.
- The situation here is a bit complicated. Since it is not possible to get `tikv_gc_life_time` directly from pd, I can only take this dirty hack.

cdc/owner.go
leoppro: please add a comment for checking `minGCSafePoint != 0` and `changefeed.status.CheckpointTs <= minGCSafePoint`

cdc/owner.go
leoppro: ditto

asddongmen: done

cdc/owner.go
asddongmen: done

asddongmen: These checks here and below (line:799) are for almost the same reason.

cdc/owner.go
```diff
@@ -127,6 +132,8 @@ type Owner struct {
 	gcSafepointLastUpdate time.Time
 	// stores the ts obtained from PD and is updated every MinGCSafePointCacheUpdateInterval.
 	minGCSafePointCache minGCSafePointCacheEntry
+	// stores the actual gcSafePoint store in pd
```
liuzix, line 135:
```suggestion
	// stores the actual gcSafePoint stored in pd
```

cdc/owner.go
```diff
@@ -753,12 +760,15 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
...
-			// If changefeed's appliedCheckpoinTs < minGCSafePoint, it means this changefeed is stagnant.
+			// 1. If changefeed's appliedCheckpoinTs <= minGCSafePoint, it means this changefeed is stagnant.
 			// They are collected into this map, and then handleStaleChangeFeed() is called to deal with these stagnant changefeed.
 			// A changefeed will not enter the map twice, because in run(),
 			// handleAdminJob() will always be executed before flushChangeFeedInfos(),
 			// ensuring that the previous changefeed in staleChangeFeeds has been stopped and removed from o.changeFeeds.
-			if changefeed.status.CheckpointTs < minGCSafePoint {
+			// 2. We need the <= check here is because when a changefeed is stagnant, its checkpointTs will be updated to pd,
+			// and it would be the minimum gcSafePoint of cdc. In order to avoid circular dependence, it is necessary to check for equality.
+			// Checking whether they are equal will make the loop assignment to happen only once, and then it will be skipped in next time.
```
liuzix, line 770: What is `loop assignment`? Can you be more specific?

cdc/owner.go
```diff
@@ -85,6 +85,11 @@ func (o *Owner) getMinGCSafePointCache(ctx context.Context) model.Ts {
 			return o.minGCSafePointCache.ts
 		}
 		o.minGCSafePointCache.ts = oracle.ComposeTS(physicalTs-(o.gcTTL*1000), logicalTs)
+
+		if o.pdGCSafePoint < o.minGCSafePointCache.ts {
```
liuzix, line 89: Please comment on the design here

liuzix (commented):
rest /lgtm

cdc/owner.go
asddongmen: done

cdc/owner.go
asddongmen: done

cdc/owner.go
asddongmen: done

leoppro (approved):

PR #1879, opened 2021-05-29
CDC Server support data-dir

JinLingChristopher:

This PR makes TiCDC support `--data-dir` at the server level.

* `--data-dir` is supported at the server level, default to `""`, if not set, it will be set by conditions:
  * if there is no other changefeed, `/tmp/cdc_data` will be set as default data-dir
  * if there is any changefeed with the `--sort-dir` set, choose the valid one with the largest available space as data-dir.
    * if there are no candidates who meet the requirement, server start failed
* `--sort-dir` set at both server level and changefeed level won't take effect, and default to `/tmp/sorter`, this means it can not be changed by both command flag or config file
  * sort dir of the server and changefeed will be set as `/{data-dir}/tmp/sorter`
* during server running, if the available space is lower than 10% of the disk, serve running will fail, due to trying to consume disk space (create new sort temporary file)
* warning user when set data dir and space unavailable.

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


cdc/server.go
```diff
@@ -47,6 +49,9 @@ import (
 
 const (
 	ownerRunInterval = time.Millisecond * 500
+	defaultDataDir   = "/tmp/cdc_data"
+	// dataDirThreshold is used to warn if the free space of the specified data-dir is lower than it, unit is byte
+	dataDirThreshold = 500
```
liuzix, line 54: Is 500 bytes a reasonable threshold?

cdc/server.go
```diff
@@ -350,3 +360,119 @@ func (s *Server) Close() {
...
+func (s *Server) setUpDataDir(ctx context.Context) error {
+	conf := config.GetGlobalServerConfig()
+	if conf.DataDir != "" {
+		conf.Sorter.SortDir = filepath.Join(conf.DataDir, config.DefaultSortDir)
+		config.StoreGlobalServerConfig(conf)
+
+		return nil
+	}
+
+	// data-dir will be decide by exist changefeed for backward compatibility
```
liuzix, line 373: Do we have a design doc for this? And what if we don't have a changefeed initially, but the user plans to create one soon, whose `sort-dir` is what the user intends for TiCDC to use?

cdc/server.go
```diff
@@ -350,3 +360,119 @@ func (s *Server) Close() {
...
+		if info.Avail > low {
+			result = dir
+			low = info.Avail
+			find = true
+		}
+	}
+
+	if !find {
+		log.Error("try to find directory for data-dir failed", zap.Strings("candidates", dirs))
+		return "", errors.Errorf("data-dir is not set, can not find a valid directory")
```
liuzix, line 458: Can you define a new error in `pkg/errors`?

cmd/client_changefeed.go
```diff
@@ -359,10 +355,12 @@ func verifyChangefeedParamers(ctx context.Context, cmd *cobra.Command, isCreate
 		CreatorVersion:    version.ReleaseVersion,
 	}
 
+	// user is not allowed to set sort-dir at changefeed level
 	if sortDir != "" {
-		cmd.Printf("[WARN] --sort-dir is deprecated in changefeed settings. " +
-			"Please use `cdc server --sort-dir` if possible. " +
-			"The sort-dir here will be no-op\n")
+		cmd.Printf(color.HiYellowString("[WARN] --sort-dir is deprecated in changefeed settings. " +
```
liuzix, line 360: Can you test whether `color.HiYellowString` will cause any unexpected output when stdout is something OTHER THAN an ANSI terminal?

cmd/client_changefeed.go
```diff
@@ -359,10 +355,12 @@ func verifyChangefeedParamers(ctx context.Context, cmd *cobra.Command, isCreate
...
+			"Please use `cdc server --data-dir` to start the cdc server if possible, sort-dir will be set automatically. " +
+			"The --sort-dir here will be no-op\n"))
+		return nil, errors.New("Creating changefeed with `--sort-dir`, it's invalid")
```
liuzix, line 363: Is it expected that the user CANNOT create a changefeed with `sort-dir`? Is it not contradictory to the code in `setUpDataDir`?

cmd/server.go
```diff
@@ -69,6 +70,7 @@ func initServerCmd(cmd *cobra.Command) {
 	cmd.Flags().Int64Var(&serverConfig.GcTTL, "gc-ttl", defaultServerConfig.GcTTL, "CDC GC safepoint TTL duration, specified in seconds")
 	cmd.Flags().StringVar(&serverConfig.LogFile, "log-file", defaultServerConfig.LogFile, "log file path")
 	cmd.Flags().StringVar(&serverConfig.LogLevel, "log-level", defaultServerConfig.LogLevel, "log level (etc: debug|info|warn|error)")
+	cmd.Flags().StringVar(&serverConfig.DataDir, "data-dir", defaultServerConfig.DataDir, "Set the dir to store CDC generated data")
```
liuzix, line 73:
```suggestion
	cmd.Flags().StringVar(&serverConfig.DataDir, "data-dir", defaultServerConfig.DataDir, "the path to the directory used to store TiCDC-generated data")
```

cmd/server.go
```diff
@@ -179,7 +184,8 @@ func loadAndVerifyServerConfig(cmd *cobra.Command) (*config.ServerConfig, error)
 		case "cert-allowed-cn":
 			conf.Security.CertAllowedCN = serverConfig.Security.CertAllowedCN
 		case "sort-dir":
-			conf.Sorter.SortDir = serverConfig.Sorter.SortDir
+			// user specified sorter dir not work
+			conf.Sorter.SortDir = config.DefaultSortDir
```
liuzix, line 188: Is there a design doc for this?

cmd/server.go
```diff
@@ -179,7 +184,8 @@ func loadAndVerifyServerConfig(cmd *cobra.Command) (*config.ServerConfig, error)
 		case "cert-allowed-cn":
 			conf.Security.CertAllowedCN = serverConfig.Security.CertAllowedCN
 		case "sort-dir":
-			conf.Sorter.SortDir = serverConfig.Sorter.SortDir
+			// user specified sorter dir not work
```
liuzix, line 187:
```suggestion
			// user specified sorter dir should not take effect
```

cmd/server.go
```diff
@@ -202,5 +208,15 @@ func loadAndVerifyServerConfig(cmd *cobra.Command) (*config.ServerConfig, error)
 		}
 	}
 
+	if conf.DataDir == "" {
+		cmd.Printf(color.HiYellowString("[WARN] TiCDC server data-dir is not set. " +
+			"Please use `cdc server --data-dir` to start the cdc server if possible.\n"))
+	}
+
+	if serverConfig.Sorter.SortDir != config.DefaultSortDir {
+		cmd.Printf(color.HiYellowString("[WARN] --sort-dir is deprecated in server settings. " +
```
liuzix, line 217: You wrote `conf.Sorter.SortDir = config.DefaultSortDir`. So when will this be run?

cdc/server.go
JinLingChristopher: it's GB, I will fix it.

cmd/client_changefeed.go
JinLingChristopher: it's documented

cdc/puller/sorter/sorter_test.go
```diff
@@ -17,12 +17,15 @@ import (
...
 	"os"
+	"path/filepath"
 	"sync/atomic"
 	"testing"
 	"time"
 
 	"go.uber.org/zap/zapcore"
 
+	_ "net/http/pprof"
+
```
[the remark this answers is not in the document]
JinLingChristopher: it's auto formated by IDE, with gofmt enabled.

cmd/client_changefeed.go
JinLingChristopher: In `setUpDataDir`, we check changefeed set `sort-dir` or not, is for old changefeeds, which may already exist from the older version of the TiCDC. In this, we ban the `sort-dir` in changefeed level.

cmd/server.go
JinLingChristopher: we should put it when trying to set conf.Sort.SortDir

cdc/server.go
```diff
@@ -350,3 +360,110 @@ func (s *Server) Close() {
...
+	}
+
+	if !ok && len(dirs) != 0 {
+		log.Warn("try to find directory for data-dir failed, use `/tmp/cdc_data` as data-dir", zap.Strings("candidates", dirs))
+	}
+
+	return result, ok
+}
+
+func getDataDirCandidates(dirs []string) (result map[string]struct{}) {
```
amyangfei, line 455: Does this function try to find a `data-dir` based on given `sort-dir`?

cmd/server.go
```diff
@@ -179,7 +184,12 @@ func loadAndVerifyServerConfig(cmd *cobra.Command) (*config.ServerConfig, error)
 		case "cert-allowed-cn":
 			conf.Security.CertAllowedCN = serverConfig.Security.CertAllowedCN
 		case "sort-dir":
-			conf.Sorter.SortDir = serverConfig.Sorter.SortDir
+			// user specified sorter dir should not take effect
+			if serverConfig.Sorter.SortDir != config.DefaultSortDir {
+				cmd.Printf(color.HiYellowString("[WARN] --sort-dir is deprecated in server settings. " +
+					"sort-dir will be set to `{data-dir}/tmp/sorter`. The sort-dir here will be no-op\n"))
+			}
+			conf.Sorter.SortDir = config.DefaultSortDir
```
amyangfei, line 192: The `sorter.SortDir` should be `{data-dir}/tmp/sorter`, but it is set to `/tmp/sorter` here, will it be set later?

JinLingChristopher: yes, it will be set at `initDataDir`

cdc/server.go
JinLingChristopher: Use the given `sort-dir`directly as `data-dir` for server is not a good choice. Consider `/tmp/cdc_sort` was an old changefeed `sort-dir`, it's not good as server `data-dir`

amyangfei: IMO the reason we should use one of the old `sort-dir` is to keep the changefeed can run normally after upgrading.

Considering user has set `sort-dir` to `/data/ticdc/sort-dir`, which is mounted with a 1T SSD, the `/` path could only has 100GB, which is used to store binaries and libraries. In this scenario using `/` as data-dir is not a good choice.

JinLingChristopher: I got your point, your idea is more considerable. We have to mention user how to start server with data-dir and the logic behind data-dir chosen.

overvenus (approved):
LGTM

amyangfei (approved):

PR #1899, opened 2021-05-31
kv/client: add incremental scan region count limit

amyangfei:

This PR adds a token limiter to kv client, in order to control the concurrency of incremental scan region, for some reasons:

- A large scale incremental scan tasks could cause pressure in TiKV server, which could lead to OOM in TiKV.
- TiKV could also add concurrency limit of incremental scan regions and prevents too much memory consumption, however it will cause too many uninitialized pending regions in TiCDC kv clients. Region reconnect could happen before the TiKV can serve them.

- Add a token based region router.
    - When the used token doesn't reach upper limit, redirect the region request as soon as possible.
    - Otherwise buffer the region request, and check 100ms to see whether there exists new token opportunity.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Add concurrency limit to the region incremental scan in kv client.


cdc/kv/metrics.go
```diff
@@ -67,6 +67,13 @@ var (
 			Name:      "channel_size",
 			Help:      "size of each channel in kv client",
 		}, []string{"id", "channel"})
+	clientRegionTokenSize = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "ticdc",
```
ben1009, line 72: maybe extract the `ticdc`  `kvclient` in
```
Namespace: "ticdc",
Subsystem: "kvclient",
```
 to const ?

cdc/kv/token_region.go
```diff
@@ -0,0 +1,120 @@
...
+	"context"
+	"sync"
+	"time"
+
+	"github.com/pingcap/errors"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	regionOutputChanSize = 16
```
ben1009, line 26: could u add more comment explain why 16, thx

cdc/kv/token_region.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (r *sizedRegionRouter) RevokeToken() {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	r.tokenUsed--
+	r.metricTokenSize.Dec()
+}
+
+func (r *sizedRegionRouter) Run(ctx context.Context) error {
+	ticker := time.NewTicker(time.Millisecond * 100)
```
ben1009, line 85: maybe defined as const also?

cdc/kv/client.go
```diff
@@ -516,7 +519,7 @@ type eventFeedSession struct {
 	// The channel to send the processed events.
 	eventCh chan<- *model.RegionFeedEvent
 	// The channel to put the region that will be sent requests.
-	regionCh chan singleRegionInfo
+	regionRouter LimitRegionRouter
```
overvenus, line 522: I think it's not a channel anymore, could you update the comment?

cdc/kv/client.go
```diff
@@ -70,6 +70,9 @@ const (
 	// failed region will be reloaded via `BatchLoadRegionsWithKeyRange` API. So we
 	// don't need to force reload region any more.
 	regionScheduleReload = false
+
+	// defines the scan region limit for each table
+	regionScanLimitPerTable = 40
```
overvenus, line 75: Is it possible to limit scan region for each store?

cdc/kv/token_region.go
```diff
@@ -0,0 +1,120 @@
...
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.Trace(ctx.Err())
+		case <-ticker.C:
```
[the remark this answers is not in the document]
amyangfei: No, `AddRegion` will try to send region if token is still available. So when the consumer can consume as soon as possible, the buffer here will be empty, and background worker `Run` does nothing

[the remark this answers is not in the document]
amyangfei: We don't need the notification mechanism here

cdc/kv/token_region.go
amyangfei: addressed

cdc/kv/token_region.go
amyangfei: addressed

cdc/kv/metrics.go
amyangfei: Yep they are constant values, but I will not change them to constants in this PR, since too many literal constants are used in metric names, maybe in another PR we can address it

cdc/kv/client.go
amyangfei: addressed

amyangfei:
> Can you comment on how your design is similar or different form a "cancellable semaphore"?

The semaphore counter will be used for fast check and metric counter

- the fast check means we will check whether we have token available, which can be achieved by `semaphore.TryAcquire(n)` and `semaphore.Release(n)`, but it is simpler with counter
- golang native semaphore doesn't expose available counter

@liuzix

amyangfei:
/run-all-tests /tidb=release-5.0 /tikv=release-5.0 /pd=release-5.0 /tiflash=release-5.0

liuzix (approved):

overvenus (approved):
LGTM

PR #1914, opened 2021-06-02
new_owner: a capture implement, the capture manages and runs the owner and processor

leoppro:
a capture implement, the capture manages and runs the owner and processor

the capture starts and runs the owner and processor, restart the owner and processor when they throw an error

#### why the capture_test.go is not exist?
the unit test of capture is a unit test framework for the hole capture(including the owner and the processor), the framework is WIP.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

### DIFF
--- a/cdc/capture/capture.go
+++ b/cdc/capture/capture.go
@@ -1,0 +1,332 @@
+// Copyright 2021 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package capture
+
+import (
+	"context"
+	"io"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/pingcap/errors"
+	"github.com/pingcap/failpoint"
+	"github.com/pingcap/log"
+	"github.com/pingcap/ticdc/cdc/kv"
+	"github.com/pingcap/ticdc/cdc/model"
+	"github.com/pingcap/ticdc/cdc/owner"
+	"github.com/pingcap/ticdc/cdc/processor"
+	"github.com/pingcap/ticdc/pkg/config"
+	cdcContext "github.com/pingcap/ticdc/pkg/context"
+	cerror "github.com/pingcap/ticdc/pkg/errors"
+	"github.com/pingcap/ticdc/pkg/orchestrator"
+	"github.com/pingcap/ticdc/pkg/version"
+	tidbkv "github.com/pingcap/tidb/kv"
+	pd "github.com/tikv/pd/client"
+	"go.etcd.io/etcd/clientv3/concurrency"
+	"go.etcd.io/etcd/mvcc"
+	"go.uber.org/zap"
+	"golang.org/x/time/rate"
+)
+
+// Capture represents a Capture server, it monitors the changefeed information in etcd and schedules Task on it.
+type Capture struct {
+	captureMu sync.Mutex
+	info      *model.CaptureInfo
+
+	owner            *owner.Owner
+	ownerMu          sync.Mutex
+	processorManager *processor.Manager
+
+	// session keeps alive between the capture and etcd
+	session  *concurrency.Session
+	election *concurrency.Election
+
+	pdClient   pd.Client
+	kvStorage  tidbkv.Storage
+	etcdClient *kv.CDCEtcdClient
+
+	newProcessorManager func() *processor.Manager
+	newOwner            func() *owner.Owner
+	stopCampaign        int32
+}
+
+// NewCapture returns a new Capture instance
+func NewCapture(pdClient pd.Client, kvStorage tidbkv.Storage, etcdClient *kv.CDCEtcdClient) *Capture {
+	return &Capture{
+		pdClient:   pdClient,
+		kvStorage:  kvStorage,
+		etcdClient: etcdClient,
+
+		newProcessorManager: processor.NewManager,
+		newOwner:            owner.NewOwner,
+	}
+}
+
+func (c *Capture) reset() error {
+	c.captureMu.Lock()
+	defer c.captureMu.Unlock()
+	conf := config.GetGlobalServerConfig()
+	c.info = &model.CaptureInfo{
+		ID:            uuid.New().String(),
+		AdvertiseAddr: conf.AdvertiseAddr,
+		Version:       version.ReleaseVersion,
+	}
+	c.processorManager = c.newProcessorManager()
+	if c.session != nil {
+		c.session.Close() //nolint:errcheck
+	}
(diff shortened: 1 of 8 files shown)


cdc/capture/capture.go
```diff
@@ -0,0 +1,332 @@
...
+			return nil
+		default:
+		}
+		err := c.reset()
+		if err != nil {
+			return errors.Trace(err)
+		}
+		err = c.run(ctx)
+		// if no error returned or capture suicide, reset the capture and run again
+		if err == nil || cerror.ErrCaptureSuicide.Equal(errors.Cause(err)) {
```
amyangfei, line 117:
```suggestion
		if err == nil || cerror.ErrCaptureSuicide.Equal(err) {
```
errors.Cause is not needed here, ditto for other error Equal check

cdc/owner/changefeed.go
```diff
@@ -258,8 +258,8 @@ func (c *changefeed) preflightCheck(captures map[model.CaptureID]*model.CaptureI
 		c.state.PatchStatus(func(status *model.ChangeFeedStatus) (*model.ChangeFeedStatus, bool, error) {
 			if status == nil {
 				status = &model.ChangeFeedStatus{
-					ResolvedTs:   c.state.Info.StartTs,
-					CheckpointTs: c.state.Info.StartTs,
+					ResolvedTs:   c.state.Info.StartTs - 1,
+					CheckpointTs: c.state.Info.StartTs - 1,
```
amyangfei, line 262: why `-1` here

cdc/capture/capture.go
amyangfei: In which scenario `capture.run` returns a nil error

cdc/capture/capture.go
```diff
@@ -0,0 +1,332 @@
...
+		// no matter why the processor is exited, return to let capture restart or exit
+		processorErr = c.runEtcdWorker(ctx, c.processorManager, model.NewGlobalState(), processorFlushInterval)
+	}()
+	wg.Wait()
+	if ownerErr != nil {
+		return errors.Annotate(ownerErr, "owner exited with error")
+	}
+	if processorErr != nil {
+		return errors.Annotate(processorErr, "processor exited with error")
+	}
```
amyangfei, line 167: There are two independent goroutines above, the `owner` routine and the `processorManager` routine, should the error in one of the goroutines terminate the running of the other one?
To see it in another way, could it happen that one of the goroutine exits and the other one is still running?

cdc/capture/capture.go
```diff
@@ -0,0 +1,340 @@
...
+	session  *concurrency.Session
+	election *concurrency.Election
+
+	pdClient   pd.Client
+	kvStorage  tidbkv.Storage
+	etcdClient *kv.CDCEtcdClient
+
+	newProcessorManager func() *processor.Manager
+	newOwner            func() *owner.Owner
+	campaignStopped     int32
```
liuzix, line 64: Can you add comment for `campaignStopped`?

cdc/capture/capture.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+func (c *Capture) setOwner(owner *owner.Owner) {
+	c.ownerMu.Lock()
+	defer c.ownerMu.Unlock()
+	c.owner = owner
+}
+
+// OperateOwnerUnderLock operates the owner with lock
+func (c *Capture) OperateOwnerUnderLock(fn func(*owner.Owner) error) error {
```
liuzix, line 272: Why is such a function necessary?

cdc/capture/capture.go
```diff
@@ -0,0 +1,340 @@
...
+	defer c.ownerMu.Unlock()
+	if c.owner == nil {
+		return cerror.ErrNotOwner.GenWithStackByArgs()
+	}
+	return fn(c.owner)
+}
+
+// Campaign to be an owner
+func (c *Capture) campaign(ctx cdcContext.Context) error {
+	failpoint.Inject("capture-campaign-compacted-error", func() {
```
liuzix, line 283: Can you add comment on why we should test handling `ErrCompacted`? Under what circumstances can `ErrCompacted` happen in reality?

cdc/capture/capture.go
```diff
@@ -0,0 +1,340 @@
...
+	if err := etcdWorker.Run(ctx, c.session, timerInterval); err != nil {
+		// We check ttl of lease instead of check `session.Done`, because
+		// `session.Done` is only notified when etcd client establish a
+		// new keepalive request, there could be a time window as long as
+		// 1/3 of session ttl that `session.Done` can't be triggered even
+		// the lease is already revoked.
+		switch {
+		case cerror.ErrEtcdSessionDone.Equal(err),
+			cerror.ErrLeaseExpired.Equal(err):
+			return cerror.ErrCaptureSuicide.GenWithStackByArgs()
```
liuzix, line 250: Do we need to log here?

cdc/capture/capture.go
leoppro: because There are plenty of use cases in http_status.go and http_handler.go (the next pr will show you )

cdc/capture/capture.go
```diff
@@ -0,0 +1,349 @@
...
+
+		log.Info("campaign owner successfully", zap.String("capture-id", c.info.ID))
+		owner := c.newOwner()
+		c.setOwner(owner)
+		err = c.runEtcdWorker(ctx, owner, model.NewGlobalState(), ownerFlushInterval)
+		c.setOwner(nil)
+		log.Info("run owner exited", zap.Error(err))
+		// if owner exits, resign the owner key
+		if resignErr := c.resign(ctx); resignErr != nil {
+			// if regisn owner failed, return error to let capture exits
```
liuzix, line 234:
```suggestion
			// if resigning owner failed, return error to let capture exits
```

liuzix (approved):

cdc/capture/capture.go
```diff
@@ -0,0 +1,349 @@
...
+			return errors.Trace(err)
+		}
+		ctx, cancel := context.WithCancel(ctx)
+		c.cancel = cancel
+		err = c.run(ctx)
+		// if capture suicided, reset the capture and run again.
+		// if the canceled error throw, there are two possible scenarios:
+		//   1. the internal context canceled, it means some error happened in the internal, and the routine is exited, we should restart the capture
+		//   2. the parent context canceled, it means that the caller of the capture hope the capture to exit, and this loop will return in the above `select` block
+		if cerror.ErrCaptureSuicide.Equal(err) || context.Canceled == errors.Cause(err) {
```
amyangfei, line 124: if err is `context.Canceled`, the following `capture recovered` log is not needed?

leoppro: if the internal context canceled, the capture recovered log is needed;
if the parent context canceled, the capture recovered log is not needed;
What suggestions do you have to refine it?

amyangfei: IMHO the internal cancel should return the real error instead of context.Canceled

amyangfei (approved):

cdc/capture/capture.go
leoppro, line 124:
```suggestion
		// TODO: make sure the internal cancel should return the real error instead of context.Canceled
		if cerror.ErrCaptureSuicide.Equal(err) || context.Canceled == errors.Cause(err) {
```

PR #1923, opened 2021-06-03
changefeed: filter out the temporary table schema change

tiancaiamao:

We are implementing [temporary table](https://github.com/pingcap/tidb/blob/master/docs/design/2021-04-20-temporary-table.md) 

Write on the temporary table do not generate any key-value to TiKV,
but for the global temporary table, the table meta data is generated.

Our local temporary table will be MySQL compatible.
Our global temporary table is an extension of MySQL, so it's not supposed to sync to a MySQL downstream.
This extension should make the CDC processing fail.

Filter the create global temporary table DDL changes.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Seems the integration test environment is so ... complex.
This is just one line change, could we skip the test?

Side effects

Update the parser, the `model.Table` is updated, this should not affect the compatibility because the lack of the field would be set the default value 

### Release note

- No release note

### DIFF
--- a/cdc/changefeed.go
+++ b/cdc/changefeed.go
@@ -174,6 +174,10 @@ func (c *changeFeed) dropSchema(schemaID model.SchemaID, targetTs model.Ts) {
 }
 
 func (c *changeFeed) addTable(tblInfo *model.TableInfo, targetTs model.Ts) {
+	// Ignore temporary table.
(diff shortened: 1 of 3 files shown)


tiancaiamao:
Without changes, the behaviour is the same with other feature(like auto random etc):
Sync TiDB to MySQL with tidb-specific feature result in error.

PR #1929, opened 2021-06-03
sink/mysql: adjust producer/consumer exit sequence to avoid goroutine leak

amyangfei:

Fix a goroutine leak bug, where the sink worker has exited, but more txns are sent to txn channel of sink worker.
And the producer will wait for the txn to be consumed forever.

Add a cleanup mechanism for sink worker.

1. producer(sink.flushRowChangedEvents goroutine) of txnCh exits
2. goroutine in 1 sends notification to closedCh of each sink worker
3. each sink worker receives the notification from closedCh and mark FinishWg as Done

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- Fix a bug that some MySQL connection could leak after MySQL sink meets error and pauses.


liuzix (approved):

cdc/sink/mysql.go
```diff
@@ -134,6 +134,11 @@ func (s *mysqlSink) FlushRowChangedEvents(ctx context.Context, resolvedTs uint64
 }
 
 func (s *mysqlSink) flushRowChangedEvents(ctx context.Context, receiver *notify.Receiver) {
+	defer func() {
+		for _, worker := range s.workers {
+			worker.closedCh <- struct{}{}
```
leoppro, line 139: will the goroutine hang at L139 when the flushRowChangedEvents function return

leoppro (approved):

PR #1949, opened 2021-06-05
kv/client: refine region reconnect and resolve lock strategy

amyangfei:

Fix some bugs and adjust kv client strategies according to some internal test results.

- Fix a bug in kv client v1/v2 that some region token is not released correctly
- Decrease `grpcMaxCallRecvMsgSize` to `256MB` (From test this doesn't affect throughtput)
- Remove un-initialized region reconnect mechanism
- Increase initialized region reconnect threshold from `15min` to `60min`
- ~Check resolve lock with last event received time, instead of the `lastResolvedTs`, resolve lock optimization will be in another PR~

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


liuzix (approved):

overvenus (approved):
LGTM

ben1009 (approved):

Issue #1902, opened 2021-06-01
Owner exits unexpectedly due to emit checkpoint failure


overvenus:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

Owner exits unexpectedly due to emit checkpoint failure.

```
[2021/05/31 11:55:02.228 +00:00] [WARN] [server.go:188] ["run owner failed"] [error="[CDC:ErrKafkaSendMessage]kafka: Failed to deliver 1 messages."] [errorVerbose="[CDC:ErrKafkaSendMessage]kafka: Failed to deliver 1 messages.
github.com/pingcap/errors.AddStack
	github.com/pingcap/errors@v0.11.5-0.20201126102027-b0a155152ca3/errors.go:174
github.com/pingcap/errors.(*Error).GenWithStackByCause
	github.com/pingcap/errors@v0.11.5-0.20201126102027-b0a155152ca3/normalize.go:279
github.com/pingcap/ticdc/pkg/errors.WrapError
	github.com/pingcap/ticdc@/pkg/errors/helper.go:28
github.com/pingcap/ticdc/cdc/sink/producer/kafka.(*kafkaSaramaProducer).SyncBroadcastMessage
	github.com/pingcap/ticdc@/cdc/sink/producer/kafka/kafka.go:142
github.com/pingcap/ticdc/cdc/sink.(*mqSink).writeToProducer
	github.com/pingcap/ticdc@/cdc/sink/mq.go:381
github.com/pingcap/ticdc/cdc/sink.(*mqSink).EmitCheckpointTs
	github.com/pingcap/ticdc@/cdc/sink/mq.go:234
github.com/pingcap/ticdc/cdc.(*changeFeed).calcResolvedTs
	github.com/pingcap/ticdc@/cdc/changefeed.go:939
github.com/pingcap/ticdc/cdc.(*Owner).calcResolvedTs
	github.com/pingcap/ticdc@/cdc/owner.go:800
github.com/pingcap/ticdc/cdc.(*Owner).run
	github.com/pingcap/ticdc@/cdc/owner.go:1334
github.com/pingcap/ticdc/cdc.(*Owner).Run
	github.com/pingcap/ticdc@/cdc/owner.go:1204
github.com/pingcap/ticdc/cdc.(*Server).campaignOwnerLoop
	github.com/pingcap/ticdc@/cdc/server.go:172
[...]

PR #1985, opened 2021-06-08
owner: Fix owner exit due to emit checkpoint failure

JinLingChristopher:
#1902 

* do not return error if send checkpoint message to sink failed, stop the  change feed that cause the error instead.
* fix some typo during walking the code.

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes

### Release note

No release note


cdc/owner.go
```diff
@@ -866,7 +866,29 @@ func (o *Owner) flushChangeFeedInfos(ctx context.Context) error {
 func (o *Owner) calcResolvedTs(ctx context.Context) error {
 	for _, cf := range o.changeFeeds {
 		if err := cf.calcResolvedTs(ctx); err != nil {
-			return errors.Trace(err)
+			log.Error("fail to calculate checkpoint ts, so it will be stopped", zap.String("changefeed", cf.id), zap.String("Error message", err.Error()))
```
overvenus, line 869:
```suggestion
			log.Error("fail to calculate checkpoint ts, so it will be stopped", zap.String("changefeed", cf.id), zap.Error(err))
```

overvenus (commented):
Could you add a unit test to cover your changes? 

BTW, nice catch on typos. 😜

cdc/owner.go
```diff
@@ -1395,12 +1417,12 @@ func (o *Owner) run(ctx context.Context) error {
...
 	}
 
-	err = o.handleAdminJob(ctx)
+	err = o.calcResolvedTs(ctx)
 	if err != nil {
 		return errors.Trace(err)
 	}
 
-	err = o.calcResolvedTs(ctx)
+	err = o.handleAdminJob(ctx)
```
amyangfei, line 1425: Is it necessary to change the order of these two operations?

asddongmen: I think even without changing the order, it will work correctly.

amyangfei (approved):

overvenus (approved):

PR #1990, opened 2021-06-09
owner: deprecate the removed changefeed state

leoppro:
before this pr, the changefeed state is very complex, now we have simplified the state representation, remove the `Removed` state.

the changefeed state transfer:
ticdc-state-transfer

 ① execute changefeed pause   
 ② execute changefeed resume   
 ③ changefeed throw a recoverable errors  
 ④ execute changefeed resume   
 ⑤ changefeed throw a unrecoverable errors  
 ⑥ The sync task has reached the preset TargetTs and the sync is automatically stopped  

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note


amyangfei (commented):
Have you check the `removed` state in integration tests.
Besides would you please follow up the related change in utf cases:
- Whether we should update test cases utf according to the state change
- How can we make utf cases compatible with different release versions.

ben1009:
is there any other tests beside ut could verify the compatibility of the change, thx

amyangfei (approved):

liuzix (approved):

PR #1999, opened 2021-06-09
*: reduce memory consumption by reducing channel sizes

liuzix:
- TiCDC can OOM frequently on machines with memory less than 16GB.

- Decreased the channel sizes in two places.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test 

Side effects

Related changes

### Release note

- Reduce unnecessary memory consumption

### DIFF
--- a/cdc/kv/region_worker.go
+++ b/cdc/kv/region_worker.go
@@ -45,7 +45,7 @@ var (
 	// other components in TiCDC, including worker pool task chan size, mounter
 	// chan size etc.
 	// TODO: unified channel buffer mechanism
-	regionWorkerInputChanSize = 12800
+	regionWorkerInputChanSize = 128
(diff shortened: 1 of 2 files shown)


amyangfei (approved):

cdc/kv/region_worker.go
```diff
@@ -45,7 +45,7 @@ var (
 	// other components in TiCDC, including worker pool task chan size, mounter
 	// chan size etc.
 	// TODO: unified channel buffer mechanism
-	regionWorkerInputChanSize = 12800
+	regionWorkerInputChanSize = 128
```
amyangfei, line 48: The smaller `regionWorkerInputChanSize` will tend to use region worker pool

ben1009 (approved):

PR #2033, opened 2021-06-10
mysql_sink: fix premature checkpoint advances during resolved-ts regressions

liuzix:
- mysqlSink might report checkpoints prematurely if `EmitRowChangedEvents` is called with events whose `commit-ts` is less than what has been last used to call `FlushRowChangedEvents`. This scenario would not have happened before #1247 

- We dial back the internal `resolved-ts` in mysqlSink if an earlier event is received.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes

### Release note

- Fix a bug that could cause data losses if a TiCDC node is kills immediate after a table migration


cdc/sink/mysql.go
```diff
@@ -106,6 +106,25 @@ type mysqlSink struct {
 }
 
 func (s *mysqlSink) EmitRowChangedEvents(ctx context.Context, rows ...*model.RowChangedEvent) error {
+	if len(rows) > 0 {
+		resolvedTs := rows[0].CommitTs - 1
+		for {
+			oldResolvedTs := atomic.LoadUint64(&s.resolvedTs)
+			if oldResolvedTs <= resolvedTs || atomic.CompareAndSwapUint64(&s.resolvedTs, oldResolvedTs, resolvedTs) {
+				break
+			}
```
amyangfei, line 115: After this change, the upper caller of sink must have a level trigger mechanism (current processor/sink manager has  ensured), would you please add some comments here about it, in case that we change the behavior by accident in the future.

amyangfei (commented):
By the way, please complete the pr description.

PR #2078, opened 2021-06-17
table_pipeline: Fix blocking entire pipeline when DDL puller is lagging behind

liuzix:
- If DDL puller is behind, mounter could block the entire tablePipeline, causing deadlocks together with TiKV concurrency control or other components.

- Events waiting to be mounted are buffered in a Deque, which has no explicit size limit (the flowController ensures there will be no OOM).

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (WIP)

Side effects

 - Increased code complexity

Related changes

### Release note

- No release note

### DIFF
--- a/cdc/processor/pipeline/mounter.go
+++ b/cdc/processor/pipeline/mounter.go
@@ -1,4 +1,4 @@
-// Copyright 2020 PingCAP, Inc.
+// Copyright 2021 PingCAP, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cdc/processor/pipeline/mounter.go
+++ b/cdc/processor/pipeline/mounter.go
@@ -29,8 +29,8 @@ import (
 )
 
 const (
-	waitEventMountedBatchSize = 1024
-	maxNotificationsPerSecond = 20.0
+	waitEventMountedBatchSize = 32 * 1024
+	maxNotificationsPerSecond = 10.0
 )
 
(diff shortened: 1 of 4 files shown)


cdc/processor/pipeline/mounter.go
```diff
@@ -14,42 +14,120 @@
...
+	stdCtx, cancel := context.WithCancel(ctx)
+	n.cancel = cancel
+
+	receiver, err := n.notifier.NewReceiver(time.Millisecond * 100)
+	if err != nil {
+		log.Panic("unexpected error", zap.Error(err))
+	}
+	defer receiver.Stop()
+
+	n.wg.Go(func() error {
```
amyangfei, line 68: Do we have unit test to cover logic in this goroutine

cdc/processor/pipeline/mounter.go
```diff
@@ -14,42 +14,119 @@
...
+	"github.com/pingcap/ticdc/pkg/notify"
 	"github.com/pingcap/ticdc/pkg/pipeline"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+	"golang.org/x/time/rate"
+)
+
+const (
+	waitEventMountedBatchSize = 128 * 1024
+	maxNotificationsPerSecond = 2.0
```
amyangfei, line 34: The message batch size, notification rate and check interval (100ms), how can these parameters affect the cpu usage and processing throughput

amyangfei (approved):

leoppro (approved):

PR #2085, opened 2021-06-17
sorter: add close check to AddEntry

liuzix:
- Cancelling Unified Sorter's `Run` does not cancel `AddEntry`. This can cause unexpected channel congestion.

- Added close detection in `AddEntry`.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes

### Release note

- No release note

### DIFF
--- a/cdc/puller/sorter/unified_sorter.go
+++ b/cdc/puller/sorter/unified_sorter.go
@@ -41,6 +41,8 @@ type UnifiedSorter struct {
 	dir         string
 	pool        *backEndPool
 	metricsInfo *metricsInfo
+
+	closeCh chan struct{}
 }
 
 type metricsInfo struct {
@@ -116,6 +118,7 @@ func NewUnifiedSorter(
 			tableID:      tableID,
 			captureAddr:  captureAddr,
 		},
+		closeCh: make(chan struct{}, 1),
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


leoppro (approved):

cdc/puller/sorter/unified_sorter.go
```diff
@@ -41,6 +41,8 @@ type UnifiedSorter struct {
 	dir         string
 	pool        *backEndPool
 	metricsInfo *metricsInfo
+
+	closeCh chan struct{}
```
amyangfei, line 45: Could you add some comments to describe the exit workflow of a unified sorter and why cancelling Unified Sorter's Run does not cancel AddEntry, such as the hierarchy of context used in `Run` and `AddEntry`

amyangfei (approved):

cdc/puller/sorter/unified_sorter.go
liuzix: This is not a problem of the sorter itself, but of the caller(s). There is no guarantee that the contexts used to call `Run` and `AddEntry` are the same one.

amyangfei: Yes, what I mean is there is no limit of the context hierarchy here, please add comments to address it

PR #2110, opened 2021-06-20
Fix unfied sort test conflict dir

JinLingChristopher:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/tests/unified_sorter_sort_dir_conflict/run.sh
+++ b/tests/unified_sorter_sort_dir_conflict/run.sh
@@ -40,7 +40,7 @@ function prepare() {
     start_ts=$(run_cdc_cli tso query --pd=http://$UP_PD_HOST_1:$UP_PD_PORT_1)
 
     # starts the first cdc server instance. It will lock the sort-dir first.
-    run_cdc_server --workdir $WORK_DIR --binary $CDC_BINARY --addr "127.0.0.1:8300" --logsuffix 1 --sort-dir /tmp/cdc_sort_1
+    run_cdc_server --workdir $WORK_DIR --binary $CDC_BINARY --addr "127.0.0.1:8300" --logsuffix 1
     capture_pid=$(ps -C $CDC_BINARY -o pid= | awk '{print $1}')
 
     TOPIC_NAME="ticdc-simple-test-$RANDOM"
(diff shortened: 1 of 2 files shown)


tests/unified_sorter_sort_dir_conflict/run.sh
```diff
@@ -60,7 +60,7 @@ function prepare() {
 
     sleep 10
     # starts the first second server instance. It should fail, and bring down the changefeed
-    run_cdc_server --workdir $WORK_DIR --binary $CDC_BINARY --addr "127.0.0.1:8301" --logsuffix 1 --sort-dir /tmp/cdc_sort_1
+    run_cdc_server --workdir $WORK_DIR --binary $CDC_BINARY --addr "127.0.0.1:8301" --logsuffix 1
```
liuzix, line 63: Can you try using a different logsuffix here?

JinLingChristopher: it will fail, because `".*ErrConflictingFileLocks.*"` is preferred, so that set the same logsuffix.

liuzix: Can you override `data-dir`?

JinLingChristopher: By using `--data-dir` flag, data-dir can be override.

tests/unified_sorter_sort_dir_conflict/run.sh
```diff
@@ -59,8 +59,8 @@ function prepare() {
     run_sql "CREATE table test.simple4(id int primary key, val int);"
 
     sleep 10
-    # starts the first second server instance. It should fail, and bring down the changefeed
-    run_cdc_server --workdir $WORK_DIR --binary $CDC_BINARY --addr "127.0.0.1:8301" --logsuffix 1 --sort-dir /tmp/cdc_sort_1
+    # starts the first second server instance. It should fail, and bring down the changefeed， due to sort-dir conflict
```
overvenus, line 62:
```suggestion
    # starts the first second server instance. It should fail, and bring down the changefeed, due to sort-dir conflict
```

overvenus (commented):
Rest LGTM

Issue #1625, opened 2021-04-09
Changfeed rules filter is updated to be invalid format, changfeed status becomes to failed


Tammyxia:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

- " tiup cdc cli changefeed update --config changfeed.toml -c replication-task-1 --pd=http://172.31.10.94:2379", update rules filter sucessfully.
- query this changfeed, check the filter rules is : 
"filter": {
        "rules": [
          "rtest1",
          "rtest2"
        ],

- Then the changfeed status become to failed, and cannot be resumed.
![image](https://user-images.githubusercontent.com/78345569/114131058-6c583000-9934-11eb-9d67-3214fe991c96.png)

- There's no way to update filter rules to be correct format. Failed changfeed cannot be recovered.

2. What did you expect to see?
If config file changfeed.toml has invalid content, the update cli should be failed to execute,  to avoid the changfeed become failed status and unrecoverable.

3. What did you see instead?

4. Versions of the cluster

    - Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

        ```
        v5.0.0-nightly, Git Commit Hash: 09a4c57d00175ae3ff9c900d5fb67f6757c58181
        ```

    - TiCDC version (execute `cdc version`):

        ```
       release-version=v5.0.0-master] [git-hash=b33b7fa15b4986db069cfc031bcd7fb827207b0b]
        ```

ben1009:
fixed by add a validation verifyReplicaConfig when load config

test result
![image](https://user-images.githubusercontent.com/1741864/123067754-d8323b00-d443-11eb-8930-6dc0482f786e.png)

config
![image](https://user-images.githubusercontent.com/1741864/123067820-e718ed80-d443-11eb-9b37-fd1ebf9a6114.png)

PR #2117, opened 2021-06-21
cli: prevent invalid filter rules from being applied, which would lead to an irrecoverable changefeed failure

ben1009:
fix Changfeed rules filter is updated to be invalid format, changfeed status becomes to failed #1625

add a validation `verifyReplicaConfig` when load config

### Check List 

Tests 

 - Unit test

Code changes

Related changes


overvenus (approved):
LGTM

amyangfei (approved):

Issue #1384, opened 2021-01-30
Server log rotation configuration


LINKIWI:
## Feature Request

**Is your feature request related to a problem? Please describe:**

TiCDC server currently inherits log rotation configuration defaults when output to a log file is enabled.

**Describe the feature you'd like:**

The ability to configure log rotation parameters on the command line alongside `--log-file` and `--log-level`.

Currently, the log configuration struct is only partially populated: call trace:

https://github.com/pingcap/ticdc/blob/8ba8df75bd4a1fa9a0ffa847eabb2136fffe658f/cmd/server.go#L92

https://github.com/pingcap/ticdc/blob/8ba8df75bd4a1fa9a0ffa847eabb2136fffe658f/cmd/util.go#L83

https://github.com/pingcap/ticdc/blob/8ba8df75bd4a1fa9a0ffa847eabb2136fffe658f/pkg/logutil/log.go#L88

https://github.com/pingcap/log/blob/8f1e84a3abc8f75ae33a8bee86d561f21057393e/log.go#L41

https://github.com/pingcap/log/blob/8f1e84a3abc8f75ae33a8bee86d561f21057393e/log.go#L86

I'm proposing the addition of new server CLI flags `--log-max-size`, `--log-max-days` and `--log-max-backups` and propagating these values to the underlying logger per the stack trace above.

If this is something the team would be interested in I'm happy to submit a patch.

**Describe alternatives you've considered:**

N/A, feature is not critically needed.

**Teachability, Documentation, Adoption, Migration Strategy:**

If the proposal is accepted I can file some follow up documentation PRs as well as needed.

amyangfei:
@LINKIWI  Thanks for your proposal. TiCDC does lack the configuration for log rotation.

In [TiDB](https://github.com/pingcap/tidb), there exist [three configuration items](https://github.com/pingcap/tidb/blob/109ad453031aa69124230cea4a5e0bb565d7783a/config/config.toml.example#L174-L181 ) corresponding to log rotation

```toml
# Max log file size in MB (upper limit to 4096MB).
max-size = 300

# Max log file keep days. No clean up by default.
max-days = 0

# Maximum number of old log files to retain. No clean up by default.
max-backups = 0
```

We would like to provide the same configuration in TiCDC.

[...]

PR #2137, opened 2021-06-23
Add log configuration items to the config file

SuperSupeng:
Add log configuration items to the config file. #1384 

Add logFileMaxSize, logFileMaxDays and logFileMaxBackups  items when load config file.

### Check List 

Tests 

 - Unit test

Code changes

Related changes

### Release note
- no release note


cmd/server.go
```diff
@@ -70,6 +70,9 @@ func initServerCmd(cmd *cobra.Command) {
 	cmd.Flags().Int64Var(&serverConfig.GcTTL, "gc-ttl", defaultServerConfig.GcTTL, "CDC GC safepoint TTL duration, specified in seconds")
 	cmd.Flags().StringVar(&serverConfig.LogFile, "log-file", defaultServerConfig.LogFile, "log file path")
 	cmd.Flags().StringVar(&serverConfig.LogLevel, "log-level", defaultServerConfig.LogLevel, "log level (etc: debug|info|warn|error)")
+	cmd.Flags().IntVar(&serverConfig.LogFileMaxSize, "log-file-max-size", defaultServerConfig.LogFileMaxSize, "max log file size in MB")
+	cmd.Flags().IntVar(&serverConfig.LogFileMaxDays, "log-file-max-days", defaultServerConfig.LogFileMaxDays, "max log file keep days")
+	cmd.Flags().IntVar(&serverConfig.LogFileMaxBackups, "log-file-max-backups", defaultServerConfig.LogFileMaxBackups, "maximum number of old log files to retain")
```
overvenus, line 75: I don't think we need to expose these settings in command flags.

It is enough to support these settings in TiCDC's config file, just like TiDB https://github.com/pingcap/tidb/blob/516584aaa47bec5a4dc131be95b413b48500f5fa/config/config.toml.example#L168-L179

overvenus (commented):
Rest LGTM

cmd/server.go
SuperSupeng: ok, I have modified it. PTAL

cmd/ticdc.toml
```diff
@@ -15,6 +15,15 @@ log-file = "/tmp/ticdc/ticdc.log"
...
 log-level = "info"
 
+# Max log file size in MB (upper limit to 4096MB).
+log-file-max-size = 300
+
+# Max log file keep days. No clean up by default.
+log-file-max-days = 0
+
+# Maximum number of old log files to retain. No clean up by default.
+log-file-max-backups = 0
```
amyangfei, line 25: The log related configuration in TiDB/PD is different, should TiCDC keep the same config style?

/cc @leoppro 

```
[log]
# Log level: debug, info, warn, error, fatal.
level = "info"

[log.file]
# Log file name.
filename = ""

# Max log file size in MB (upper limit to 4096MB).
max-size = 300

# Max log file keep days. No clean up by default.
max-days = 0

# Maximum number of old log files to retain. No clean up by default.
max-backups = 0
```

ben1009: agree, rest lgtm

SuperSupeng: I have modified it. PTAL

overvenus: For the most cases, tiup sets log level and path via flags, so the compatibility issue is negligible.

overvenus (approved):
LGTM

cmd/server.go
```diff
@@ -15,6 +15,7 @@ package cmd
 
 import (
 	"context"
+	"github.com/pingcap/ticdc/pkg/logutil"
```
overvenus, line 18: BTW, could you move the import to line 29?

SuperSupeng: get!

PR #2138, opened 2021-06-23
client,server: better err msg when PD endpoint missing certificate

hi-rustin:
close https://github.com/pingcap/ticdc/issues/1973

Add better err msg.
### Check List 

Tests 

 - Unit test

Code changes

Side effects

None

Related changes

### Release note

- Better err msg when PD endpoint missing certificate


hi-rustin:
I'm not sure we need a more detailed error message (e.g. prompting the user to use parameters to specify a certificate), suggestions welcome!

hi-rustin:
Also, should we check that it is a valid URL first?

leoppro (approved):

cmd/server.go
```diff
@@ -204,11 +210,15 @@ func loadAndVerifyServerConfig(cmd *cobra.Command) (*config.ServerConfig, error)
...
 				return nil, cerror.ErrInvalidServerOption.GenWithStack("PD endpoint scheme should be https")
 			}
-		} else if strings.Index(ep, "http://") != 0 {
-			return nil, cerror.ErrInvalidServerOption.GenWithStack("PD endpoint scheme should be http")
+		} else {
+			if strings.HasPrefix(ep, HTTPS) {
+				return nil, errors.New("PD endpoint scheme is https, please provide certificate")
+			} else if !strings.HasPrefix(ep, HTTP) {
+				return nil, cerror.ErrInvalidServerOption.GenWithStack("PD endpoint scheme should be http")
+			}
```
lonng, line 221: Seems the same code was duplicated, and one has the error code but the other hasn't.

cmd/client.go
```diff
@@ -138,11 +138,15 @@ func newCliCommand() *cobra.Command {
 				return errors.Annotate(err, "fail to validate TLS settings")
 			}
 			if tlsConfig != nil {
-				if strings.Contains(cliPdAddr, "http://") {
+				if strings.HasPrefix(cliPdAddr, HTTP) {
```
lonng, line 141: Why not parse the URL to check the scheme information?

hi-rustin: Yes, it is best to do a URL validate first: https://github.com/pingcap/ticdc/pull/2138#issuecomment-867303878

I will add that step.

cmd/server.go
hi-rustin: Good catch! Fixed.

cmd/client.go
hi-rustin: Added. Could you please take again?

hi-rustin:
/hold 

Code can be better.

hi-rustin:
/unhold

Ready for review.

overvenus (approved):

Issue #1984, opened 2021-06-08
remove changefeed need a comfirm step


Tammyxia:
## Feature Request

**Is your feature request related to a problem? Please describe:**

> Now remove changefeed without comfirm hint, it's not safe if there's  wrong remove operation.
![image](https://user-images.githubusercontent.com/78345569/121171523-2d3d5100-c889-11eb-87ce-b915a4d69b76.png)

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

SuperSupeng:
I will implement this feature.

PR #2173, opened 2021-06-28
Add confirm when remove changefeed

SuperSupeng:
 Add confirm when remove changefeed. #1984 

image

### Check List 

Tests 

 - Unit test

Code changes

Related changes

### Release note
- No release note

### DIFF
--- a/cmd/client_changefeed.go
+++ b/cmd/client_changefeed.go
@@ -112,6 +112,9 @@ func newAdminChangefeedCommand() []*cobra.Command {
 			Use:   "remove",
 			Short: "Remove a replication task (changefeed)",
 			RunE: func(cmd *cobra.Command, args []string) error {
+				if err := confirmChangefeedRemove(cmd); err != nil {
+					return err
+				}
 				ctx := defaultContext
 				job := model.AdminJob{
 					CfID: changefeedID,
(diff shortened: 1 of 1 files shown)


ben1009:
about the title, could refer this part `Contribution flow`
https://github.com/pingcap/ticdc/blob/master/CONTRIBUTING.md
some of the content maybe outdated, we could update the doc according, i think, thanks

cmd/client_changefeed.go
```diff
@@ -739,3 +742,19 @@ func getUpstreamCredential() *security.Credential {
 		KeyPath:  upstreamSslKeyPath,
 	}
 }
+
+func confirmChangefeedRemove(cmd *cobra.Command) error {
+	if noConfirm {
+		return nil
+	}
+	cmd.Printf("Confirm that you want to remove this changefeed [Y/N]\n")
+	var yOrN string
```
ben1009, line 751: see this code block several times, can we unify all the cases, thx?

cmd/client_changefeed.go
```diff
@@ -739,3 +742,19 @@ func getUpstreamCredential() *security.Credential {
...
+	_, err := fmt.Scan(&yOrN)
+	if err != nil {
+		return err
+	}
+	if strings.ToLower(strings.TrimSpace(yOrN)) != "y" {
+		return errors.NewNoStackError("Abort remove changefeed!")
```
ben1009, line 757: if i type `N` to cancel the operation,  seems should not return error?

cmd/client_changefeed.go
SuperSupeng: Sorry, I don't quite know what you mean.

cmd/client_changefeed.go
SuperSupeng: I did this by referring to the [unsafe operation](https://github.com/pingcap/ticdc/blob/dfc046fd5dbe973a49eba4d31c376ed7634c3f82/cmd/client_unsafe.go#L116). But I think you are right, I will fix it.

cmd/client_changefeed.go
ben1009: sorry, what i mean is the check `y/n` logic been used several times in this file, maybe could extract the same logic to a func instead.

SuperSupeng: Ok, sure. It's a good suggestion.

amyangfei (commented):
Have you checked whether this change breaks the integration test, as well as utf test

PR #2197, opened 2021-06-30
cdc_*: delete some code that use old owner

asddongmen:

Delete some code that that relied on the old leader.

### Check List 

Tests 

 - No code

Code changes

 - Has persistent data change

Side effects

 - None

Related changes

### Release note

- No release note


amyangfei:
Should we cherry-pick to release-4.0

ben1009:
seems we should have the new owner tested in 5.1 first, then pick to 5.0, could u help confirm @overvenus ?

asddongmen:
> Should we cherry-pick to release-4.0
It seems that the new owner is not opened in 4.0 & 5.0 branch. 
So we can cheery-pick it to the 5.1 branch first, and we can manually cherry-pick it to 4.0 & 5.0 branches later.

cdc/server.go
```diff
@@ -56,11 +53,7 @@ const (
 
 // Server is the capture server
 type Server struct {
-	captureV2 *capture.Capture
-
-	capture      *Capture
-	owner        *Owner
-	ownerLock    sync.RWMutex
+	captureV2    *capture.Capture
```
overvenus, line 56: Could you rename 'captureV2' to 'capture'?

overvenus (commented):
Rest LGTM

cdc/server.go
asddongmen: done

overvenus (approved):

amyangfei (approved):

PR #2200, opened 2021-07-01
cmd(refactoring): move applications into cmd

hi-rustin:

part of https://github.com/pingcap/ticdc/issues/2199

move applications into cmd

### Check List 

Tests 

 - No code

Code changes

None

Side effects

 None

Related changes

### Release note

- No release note


overvenus (approved):
LGTM

amyangfei (approved):

Makefile
```diff
@@ -79,14 +79,14 @@ build: cdc
...
+	$(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/cdc ./cmd/cdc/main.go
 	$(FAILPOINT_DISABLE)
 
 cdc:
-	$(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/cdc ./main.go
+	$(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/cdc ./cmd/cdc/main.go
 
 kafka_consumer:
-	$(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/cdc_kafka_consumer ./kafka_consumer/main.go
+	$(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/cdc_kafka_consumer ./cmd/kafka-consumer/main.go
```
amyangfei, line 89: oops, I didn't notice the change here, since the `kafka-consumer` is only used for integration test and will not be published, do we still need to move it to the `cmd` directory? @hi-rustin

hi-rustin: Thanks for your review!

I think it's appropriate. Because it is also an executable program. I don't see a problem with putting it into cmd.

hi-rustin: For example: https://github.com/kubernetes/kubernetes/blob/master/cmd/linkcheck/links.go  may also be used only internally. But it is also placed in the cmd as an executable program.

amyangfei: Fine, thank you

PR #2201, opened 2021-07-01
workerpool: remove mergeContext

### LINKED ISSUE #2211: workerpool.(*mergedContext).Done creates too many goroutines
## Bug Report

Capturing 6k tables, go runtime trace shows workerpool.(*mergedContext).Done creates 105k goroutines, this is clearly not expected, it should be optimized.

```
Goroutines:
golang.org/x/sync/errgroup.(*Group).Go.func1 N=81848
runtime.gcBgMarkWorker N=38google.golang.org/grpc/internal/transport.(*http2Client).reader N=27802
github.com/pingcap/ticdc/cdc.(*oldProcessor).addTable.func2.5 N=6672
github.com/pingcap/ticdc/pkg/workerpool.(*mergedContext).Done.func1 N=1056480
github.com/pingcap/ticdc/cdc.(*oldProcessor).addTable.func2.4 N=6672
runtime.bgsweep N=1google.golang.org/grpc/internal/transport.(*http2Client).keepalive N=27798
github.com/pingcap/ticdc/pkg/notify.(*Receiver).signalTickLoop.func1 N=13381
google.golang.org/grpc/internal/transport.newHTTP2Client.func3 N=27802
net/http.(*conn).serve N=2
```

No 105k goroutines.

3. What did you see instead?

105k goroutines.

4. Versions of the cluster

v5.0.3

liuzix:
- `mergeContext` spawns a goroutine when `Done()` is called for the first time. Since we generate a `mergeContext` for each event before calling its handler, when concurrency is high, we will be running a large number of goroutines concurrently, and these goroutines do little that's useful.

Close #2211 

- Remove `mergeContext` completely. Event handler calls can only be cancelled with the `ctx` used to add that event. They will no longer be cancelled when the workerpool is cancelled.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Optimize workerpool for fewer goroutines when concurrency is high.


pkg/workerpool/pool.go
```diff
@@ -38,11 +38,14 @@ type EventHandle interface {
 	// AddEvent adds an `event` object to the internal queue, so that the `f` used to register the handle can be called.
 	// Note: events are always processed in the order they are added.
 	// Unregistering the EventHandle MAY CAUSE EVENT LOSSES. But for an event lost, any event after it is guaranteed to be lost too.
+	// Cancelling `ctx` here will cancel the on-going or next execution of the event.
 	AddEvent(ctx context.Context, event interface{}) error
 
 	// SetTimer is used to provide a function that is periodic called, as long as the EventHandle has not been unregistered.
 	// The current implementation uses as the base clock source a ticker whose interval is the const workerPoolDefaultClockSourceInterval.
 	// DO NOT set an interval less than workerPoolDefaultClockSourceInterval.
+	// DO NOT set an interval less than workerPoolDefaultClockSourceInterval.
```
ben1009, line 47: duplicated ?

pkg/workerpool/pool_test.go
```diff
@@ -368,3 +368,52 @@ func (s *workerPoolSuite) TestBasics(c *check.C) {
...
+		errg.Go(func() error {
+			select {
+			case <-ctx.Done():
+			case <-handler.ErrCh():
+			}
+			return nil
+		})
+	}
+
+	time.Sleep(5 * time.Second)
```
amyangfei, line 414: Is it necessary to sleep 5s here?

pkg/workerpool/pool_test.go
```diff
@@ -368,3 +368,52 @@ func (s *workerPoolSuite) TestBasics(c *check.C) {
...
+			<-ctx.Done()
+			return ctx.Err()
+		})
+
+		errg.Go(func() error {
+			for j := 0; j < 64; j++ {
+				err := handler.AddEvent(ctx, j)
+				if err != nil {
+					return nil
+				}
```
amyangfei, line 400: if return `nil` here, should add an assertion to the error?

liuzix: No. `AddEvent` is allowed to return an error if `ctx` has already been cancelled, which will happen here.

pkg/workerpool/pool_test.go
liuzix: I just wanted to make sure that the internal queue in the pool has been fully congested. I can reduce it to 2s.

overvenus: We can loop and check.

```
for {
  sleep(100ms)
  if check_full? {
    break
  }
  if wait_time > 5s {
    panic("too long")
  }
} 
```

overvenus (commented):
Rest LGTM

pkg/workerpool/pool_test.go
amyangfei: Yes, `AddEvent` could return `context.Canceled` or `ErrWorkerPoolHandleCancelled`, both these two errors are expected?

liuzix: We are not test cancelling the handles, so `ErrWorkerPoolHandleCancelled` is not expected in this test case.

overvenus:
Could you move this issue forward? I have been test it for a while not issue found.

amyangfei (approved):

overvenus (approved):

PR #2205, opened 2021-07-01
changefeed: fix add partition in old owner

amyangfei:

When TiCDC meets partition table update, the owner diffs the latest partition table information with the memory information, the new added partition and dropped partition should be calculated by

new added = Set(partition table IDs in latest table info) - Set(partition table IDs that TiCDC maintains)
dropped = Set(partition table IDs that TiCDC maintains) - Set(partition table IDs in latest table info)

The new added partition calculation is wrong, the affected version includes >=4.0.2, >=5.0.0.

5.1.0 and master are not affected.

Should use `oldIDs` as the partition table IDs that TiCDC maintains

### Check List 

Tests 

 - Unit test
 - Integration test
 
### Release note

- Fix extra partition dispatching when adding new table partition.


overvenus (approved):
LGTM

overvenus (approved):

lonng (approved):

PR #2209, opened 2021-07-02
replication_model: split keys for task status

liuzix:
- TaskStatus keys stored in Etcd is too large, easily exceeding the default limit of 1.5MB, and even if this limit can be overridden, it would cause great latency for the whole Etcd cluster.

- Split the TaskStatus keys into a AdminJob key (still called TaskStatus for compatibility), and numerous per-table TableStatus keys.
- Kept the interface of the TiCDC Etcd ReactorState unchanged, achieving full (or nearly full) semantic compatibility with the old implementation, circumventing the need for the owner & processor to be modified.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes

 - Need discuss whether to cherry pick to master

### Release note

- Optimize Etcd data model.


ben1009:
FYI, currently add `wip` or `dnm` in the title, the PTAL bot will skip the pr

PR #2218, opened 2021-07-04
filter: support for synchronizing user permission-related tables

hi-rustin:
close https://github.com/pingcap/ticdc/issues/2192

Support for synchronizing user permission-related tables.

Please see the code comments for details.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

N/A

### Release note

- Support for synchronizing user permission-related tables

### DIFF
--- a/pkg/filter/filter.go
+++ b/pkg/filter/filter.go
@@ -14,6 +14,8 @@
 package filter
 
 import (
+	"strings"
+
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/ticdc/pkg/config"
 	"github.com/pingcap/ticdc/pkg/cyclic/mark"
@@ -22,7 +24,19 @@ import (
 	filterV2 "github.com/pingcap/tidb-tools/pkg/table-filter"
 )
 
-// Filter is a event filter implementation
+// userRelatedTables specifies the tables related to user permissions.
+var userRelatedTables = map[string]struct{}{
+	"columns_priv":  {},
+	"db":            {},
+	"default_roles": {},
+	"global_grants": {},
+	"global_priv":   {},
+	"role_edges":    {},
+	"tables_priv":   {},
+	"user":          {},
+}
+
+// Filter is a event filter implementation.
 type Filter struct {
 	filter           filterV2.Filter
 	ignoreTxnStartTs []uint64
@@ -30,8 +44,9 @@ type Filter struct {
 	isCyclicEnabled  bool
 }
 
-// NewFilter creates a filter
-func NewFilter(cfg *config.ReplicaConfig) (*Filter, error) {
+// VerifyRules checks the filter rules in the configuration
+// and returns an invalid rule error if the verification fails, otherwise it will return the parsed filter.
+func VerifyRules(cfg *config.ReplicaConfig) (filterV2.Filter, error) {
 	var f filterV2.Filter
 	var err error
 	if len(cfg.Filter.Rules) == 0 && cfg.Filter.MySQLReplicationRules != nil {
@@ -46,6 +61,17 @@ func NewFilter(cfg *config.ReplicaConfig) (*Filter, error) {
 	if err != nil {
 		return nil, cerror.WrapError(cerror.ErrFilterRuleInvalid, err)
 	}
+
+	return f, nil
+}
+
+// NewFilter creates a filter.
+func NewFilter(cfg *config.ReplicaConfig) (*Filter, error) {
+	f, err := VerifyRules(cfg)
+	if err != nil {
+		return nil, cerror.WrapError(cerror.ErrFilterRuleInvalid, err)
+	}
+
 	if !cfg.CaseSensitive {
 		f = filterV2.CaseInsensitive(f)
 	}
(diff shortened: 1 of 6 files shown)


hi-rustin:
/hold

I currently have some questions:
1. should we determine if both upstream and downstream are TiDB?
2. should these related system tables still follow the filtering rules? (Right now related tables will follow and other unrelated tables will not, which I think brings complexity, but I'm not sure if we should keep it.)

hi-rustin:
I've done some simple integration tests locally and they work well. But we should need more tests. If you have suggestions on how to test, feel free to comment. Thanks!

amyangfei (commented):
Is there any privilege requirement(in downstream) to write to these system tables?

hi-rustin:
> Is there any privilege requirement(in downstream) to write to these system tables?

Because we need to perform a `flush privileges` operation after synchronizing the user privilege information table, which requires `RELOAD` permissions.


I've done the following tests locally:

1. start the upstream and downstream tidb clusters
2. Create `pig` user in the downstream cluster and : `GRANT SELECT, INSERT, UPDATE, DELETE, Index, Create, Drop, Alter, Create View, RELOAD ON *. * TO 'pig'@'%';`
3. flush downstream cluster privileges: `flush privileges`
4. Start ticdc: `. /cdc cli changefeed create --pd http://upstream-pd:2379 --sink-uri mysql://pig:123456@downstream-tidb:4000/`
5. Create the dog user upstream, use sysbench to create the `sbtest1` table and add data, grant select permission to the dog user only: `GRANT SELECT ON test.sbtest1 TO 'dog'@'%';`
6. Refresh the upstream cluster permissions: `flush privileges` and test that the permissions are valid
7. Login with `pig` user in the downstream cluster and refresh privileges: `flush privileges`
8. Log in with the `dog `user in the downstream cluster
9. can select data correctly, but cannot drop table...  `show grants for 'dog'@'%';` see if the privileges are well.

hi-rustin:
Also about dynamic permissions: https://docs.pingcap.com/zh/tidb/dev/privilege-management#%E5%8A%A8%E6%80%81%E6%9D%83%E9%99%90  

I have done the following tests locally:

1. try with `dog` user in downstream cluster: `set global sort_buffer_size=1024*1024*4;` execution failed, permission missing.
2. grant for dog user in upstream cluster: `grant system_variables_admin on *. * TO 'dog'@'%';`
3. Refresh the upstream cluster permissions and check `dog`'s permissions
4. Log in to the downstream tidb cluster using the `pig` user, refresh the permissions, and check the permissions of the `dog` user.
5. Set again with `dog` user: set global sort_buffer_size=1024*1024*4; Execute successfully.

amyangfei:
Please fix the unit test

pkg/filter/filter.go
```diff
@@ -180,7 +195,27 @@ func (f *Filter) shouldDiscardByBuiltInDDLAllowlist(ddlType model.ActionType) bo
...
+// user table fields due to inconsistent upstream and downstream TiDB versions.
+// Also we did not consider the cooperation with other filtering rules and
+// will copy all the permissions directly.
+// See also: https://github.com/pingcap/ticdc/issues/2192#issuecomment-872096744.
+// NOTICE: If you filter the user-related tables in the filter rule it will still take effect.
+func isUserRelatedSysTable(db, table string) bool {
+	// First we need to check if it is a MYSQL system table.
+	// See also: https://dev.mysql.com/doc/refman/8.0/en/system-schema.html.
+	if strings.ToUpper(db) == "MYSQL" {
+		if _, ok := userRelatedTables[table]; ok {
```
overvenus, line 215: Do we need `strings.ToLower(table)`?

overvenus (commented):
Rest LGTM

pkg/filter/filter.go
hi-rustin: Thanks for your review!
I don't think it's necessary, because in general these tables should always be lowercase. I'm using upper case here to align with the logic in our filter package.

pkg/filter/filter_test.go
```diff
@@ -39,13 +39,16 @@ func (s *filterSuite) TestShouldUseDefaultRules(c *check.C) {
...
+	c.Assert(filter.ShouldIgnoreTable("mysql", "global_grants"), check.IsFalse)
+	c.Assert(filter.ShouldIgnoreTable("mysql", "tidb"), check.IsTrue)
 }
 
 func (s *filterSuite) TestShouldUseCustomRules(c *check.C) {
 	defer testleak.AfterTest(c)()
 	filter, err := NewFilter(&config.ReplicaConfig{
 		Filter: &config.FilterConfig{
-			Rules: []string{"sns.*", "ecom.*", "!sns.log", "!ecom.test"},
+			Rules: []string{"sns.*", "ecom.*", "!sns.log", "!ecom.test", "mysql.*", "mysql.user", "!mysql.global_grants"},
```
amyangfei, line 51: "mysql.user" does not take effect in test, since `mysql.*` is used

hi-rustin: Fixed. Thanks!

amyangfei (approved):

PR #2223, opened 2021-07-05
etcd_worker: big txn

liuzix:
- A native Etcd transactions can only hold up to 128 operations. Even if the parameter can be adjusted, it would cause performance degradation in the whole Etcd cluster.
- We need Etcd transactions to make sure that the owner knows, at every instance, which tables are taken into consideration when each capture calculates its own checkpoint-ts, to prevent data loss.

- Emulate transactions in EtcdWorker with a prewrite-commit protocol.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

### Release note

- No release note

### DIFF
--- a/pkg/orchestrator/etcd_txn.go
+++ b/pkg/orchestrator/etcd_txn.go
@@ -1,0 +1,335 @@
+// Copyright 2021 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package orchestrator
+
+import (
+	"bytes"
+	"container/heap"
+	"context"
+	"fmt"
+	"math"
+	"sort"
+	"time"
+
+	"github.com/pingcap/failpoint"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
+	"github.com/pingcap/ticdc/pkg/orchestrator/util"
+	"go.etcd.io/etcd/clientv3"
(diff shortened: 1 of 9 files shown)


pkg/orchestrator/etcd_worker.go
```diff
@@ -291,6 +436,156 @@ func (worker *EtcdWorker) applyPatches(ctx context.Context, patches []DataPatch)
...
+			}
+			if !resp.Succeeded {
+				log.Warn("Undo Txn failed", zap.Reflect("undo-log", undoLogEntry))
+			}
+		}
+	}()
+
+	changeSet.sort()
+
+	for _, change := range changeSet {
```
amyangfei, line 487: Will we add batch execution to the `changeSet`?

pkg/orchestrator/etcd_worker.go
```diff
@@ -291,6 +436,156 @@ func (worker *EtcdWorker) applyPatches(ctx context.Context, patches []DataPatch)
...
+
+	txnResp, err := worker.client.Txn(ctx).
+		If(clientv3.Compare(clientv3.ModRevision(newTxnKey.String()), "<", txnRevision+1)).
+		Then(clientv3.OpPut(newTxnKey.String(), string(newTxnBytes), clientv3.WithLease(clientv3.NoLease))).
+		Commit()
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if txnResp.Succeeded {
+		log.Info("Committing big txn successful")
```
amyangfei, line 558: Do we need to add more information in the log?

pkg/orchestrator/etcd_txn.go
```diff
@@ -0,0 +1,334 @@
...
+		})
+	}
+	return ret
+}
+
+func (s txnChangeSet) sort() {
+	sort.Slice(s, func(i, j int) bool {
+		a := s[i].key.Bytes()
+		b := s[j].key.Bytes()
+		return bytes.Compare(a, b) == -1
```
amyangfei, line 261: Compare `key.String()` directly can save []bytes cast.

PR #2249, opened 2021-07-08
sink: fix concurrent map access in sink manager

liuzix:
- There is an unsynchronized access to a shared map, which can cause runtime panic.

- Lock it.

### Check List 

Tests 

 - Unit test

Related changes

### Release note

- Fix minor runtime panic risk

### DIFF
--- a/cdc/sink/manager_test.go
+++ b/cdc/sink/manager_test.go
@@ -100,7 +100,6 @@ func (s *managerSuite) TestManagerRandom(c *check.C) {
 		}()
 	}
 	wg.Wait()
-	wg = sync.WaitGroup{}
 	for i := 0; i < goroutineNum; i++ {
 		i := i
 		tableSink := tableSinks[i]
--- a/cdc/sink/manager_test.go
+++ b/cdc/sink/manager_test.go
@@ -92,8 +92,15 @@ func (s *managerSuite) TestManagerRandom(c *check.C) {
 	var wg sync.WaitGroup
 	tableSinks := make([]Sink, goroutineNum)
 	for i := 0; i < goroutineNum; i++ {
-		tableSinks[i] = manager.CreateTableSink(model.TableID(i), 0)
+		i := i
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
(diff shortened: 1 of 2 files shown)


cdc/sink/manager_test.go
```diff
@@ -92,8 +92,15 @@ func (s *managerSuite) TestManagerRandom(c *check.C) {
...
-		tableSinks[i] = manager.CreateTableSink(model.TableID(i), 0)
+		i := i
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			tableSinks[i] = manager.CreateTableSink(model.TableID(i), 0)
+		}()
 	}
+	wg.Wait()
+	wg = sync.WaitGroup{}
```
3AceShowHand, line 103: Is L103 necessary? I think wg can be reused.

amyangfei (approved):

cdc/sink/manager_test.go
```diff
@@ -92,8 +92,14 @@ func (s *managerSuite) TestManagerRandom(c *check.C) {
 	var wg sync.WaitGroup
 	tableSinks := make([]Sink, goroutineNum)
 	for i := 0; i < goroutineNum; i++ {
-		tableSinks[i] = manager.CreateTableSink(model.TableID(i), 0)
+		i := i
```
ben1009, line 95: `i := i`， maybe` j :=i`

ben1009 (approved):

cdc/sink/manager_test.go
liuzix: The variable `i` is deliberately shadowed to avoid programmer mistakes

ben1009: Because of closure.? But with the same name kind of strange, imo

amyangfei:
```
It may seem odd to write

req := req

but it's legal and idiomatic in Go to do this. You get a fresh version of the variable with the same name, deliberately shadowing the loop variable locally but unique to each goroutine.
```

This way is widely used in Go, ref: https://golang.org/doc/effective_go

overvenus (approved):

Issue #1260, opened 2020-12-31
DDL lost after owner is killed (kill_owner_with_ddl fails)


liuzix:
DDL can be lost after adjusting order of owner operations. The owner now tries to flush changefeed status right after recalculating the ts's. Consider the following scenario:
- Owner recalculates the global checkpointTs to `A`, which is the finishTs of a DDL.
- Owner successfully upload the new the global checkpointTs, which gets persisted in Etcd.
- Owner gets killed before successfully executing DDL.
- New Owner thinks this DDL has already been executed.

This scenario is a conjecture, please comment if you have more thoughts on this.

amyangfei:
Have been fixed by https://github.com/pingcap/ticdc/pull/1236.

the checkpoint won't be updated to DDL `FinishedTs` until the DDL has ben executed successfully

PR #2252, opened 2021-07-08
owner: fix new owner updating checkpoint too early with pending DDL

liuzix:
- #1260 was reintroduced into the new owner (released in v5.1.0)

- Changed the start-ts of DDL puller to `checkpoint - 1`, to satisfy the open-interval assumption by TiKV.
- Fetch the snapshot of meta data at `checkpoint` instead of `checkpoint-1` avoid overlap between listened DDL events and fetched schema snapshot.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Fix potential DDL loss when owner crashes while executing DDL

### DIFF
--- a/tests/kill_owner_with_ddl/run.sh
+++ b/tests/kill_owner_with_ddl/run.sh
@@ -72,7 +72,7 @@ function run() {
 
     for i in $(seq 1 3); do
         kill_cdc_and_restart $pd_addr $WORK_DIR $CDC_BINARY
-        sleep 15
+        sleep 8
     done
 
     export GO_FAILPOINTS=''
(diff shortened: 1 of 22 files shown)


tests/kill_owner_with_ddl/run.sh
```diff
@@ -72,7 +72,7 @@ function run() {
 
     for i in $(seq 1 3); do
         kill_cdc_and_restart $pd_addr $WORK_DIR $CDC_BINARY
-        sleep 2
+        sleep 15
     done
```
amyangfei, line 76: this test will sleep `45s` totally, it costs too much time

amyangfei (approved):

ben1009:
just curious, do we have test cover the case or how do we detect the issue, thanks

liuzix:
> just curious, do we have test cover the case or how do we detect the issue, thanks

kill_owner_with_ddl was supposed to detect this issue, but since it did not sleep long enough for the owner to actually start processing the DDL, the test would pass.

overvenus (approved):

PR #2268, opened 2021-07-12
new_collation: enable old_value

SuperSupeng:
By setting enableOldValue to true, the data can be correctly synchronized when open NEW_COLLATION regardless of configuration.
This method is only a temporary solution.

- set enableOldValue to true
- modify sink manager to remove excess data

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note

- No release note

### DIFF
--- a/cdc/sink/manager.go
+++ b/cdc/sink/manager.go
@@ -15,6 +15,7 @@ package sink
 
 import (
 	"context"
+	"github.com/pingcap/ticdc/pkg/config"
 	"math"
 	"sort"
 	"sync"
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -161,8 +161,8 @@ type LogConfig struct {
 var defaultServerConfig = &ServerConfig{
 	Addr:          "127.0.0.1:8300",
 	AdvertiseAddr: "",
-	LogFile:       "",
-	LogLevel:      "info",
+	LogFile:       "./ticdc.log",
+	LogLevel:      "debug",
```
overvenus, line 165: Why change these two default settings?

cdc/sink/manager.go
```diff
@@ -15,6 +15,7 @@ package sink
 
 import (
 	"context"
+	"github.com/pingcap/ticdc/pkg/config"
```
ben1009, line 18: fmt

cdc/sink/manager.go
```diff
@@ -230,6 +231,15 @@ func (b *bufferSink) run(ctx context.Context, errCh chan error) {
 				continue
 			}
 			start := time.Now()
+			for i := 0; i<len(e.rows); i++{
+				if !config.GetDefaultReplicaConfig().EnableOldValue{
```
ben1009, line 235: could you add some comments, kind of confusing about the logic

PR #2282, opened 2021-07-14
pipeline: Move mounting before flow control

liuzix:
- We want to save goroutines consumed by pipeline by reducing the number of nodes.
- We want to use the sorter to absorb backpressure caused by insufficient mounter performance.
- The previous solution of using a deque in a separate node might have introduced CPU and memory overhead.

- Move mounting entirely in front of flow controlling
- Make sending to mounter and reading from sorter share the same goroutine.

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note

<!-- bugfixes or new feature need a release note, must in the form of a list, such as

- Combine the `sorter` and `mounter` pipeline nodes to optimize for performance.


cdc/processor/pipeline/sorter.go
```diff
@@ -120,20 +123,85 @@ func (n *sorterNode) Init(ctx pipeline.NodeContext) error {
...
+				}
+				eventToMount = msg
+			case writeNext() <- eventToMount:
+				eventToMount = nil
+			case msg, ok := <-loadNext():
+				if !ok {
+					// mounter wait channel closed
+					return nil
+				}
+				nextEvent = msg
```
overvenus, line 201: What about inlining `nextReady` after line 201?

cdc/processor/pipeline/sorter.go
```diff
@@ -120,20 +123,85 @@ func (n *sorterNode) Init(ctx pipeline.NodeContext) error {
...
+					select {
+					case <-ctx.Done():
+						return nil
+					case n.mounter.Input() <- msg:
+					}
+				}
+				eventToMount = msg
+			case writeNext() <- eventToMount:
+				eventToMount = nil
+			case msg, ok := <-loadNext():
```
overvenus, line 196: Could you select the two cases in another `select` statement (with `default` if no-blocking is needed)?

PR #2344, opened 2021-07-21
cmd/root: Add a check to check whether the cli parameter contains non-ascii code

### LINKED ISSUE #2303: cdc cli changefeed create --start-ts does not work properly and may lead to data loss
## Bug Report

TiCDC cli creates a changefeed that does not use the start-ts specified in command line. It leads to data loss.

image

Note that the start-ts in command line is invalid and it points to 1975-02-25 15:14:19.597000.

```
mysql> select tidb_parse_tso(426318359552065537); # start-ts in cli output

+------------------------------------+
| tidb_parse_tso(426318359552065537) |
+------------------------------------+
| 2021-07-14 23:11:20.469000         |
+------------------------------------+

mysql> select tidb_parse_tso(42610054816776198); # start-ts in command line

+-----------------------------------+
| tidb_parse_tso(42610054816776198) |
+-----------------------------------+
| 1975-02-25 15:14:19.597000        |
+-----------------------------------+
```

  a. create a changefeed with a start-ts specified in command line.
  b. report warnings about invalid start-ts. 

3. What did you see instead?

A wrong changefeed is created and it leads to data loss. 

4. Versions of the cluster

v5.0.1

asddongmen:
Add a check to check whether the parameter contains non-ascii code

close #2303  

Add a check to check whether the parameter contains non-ascii code

### Check List 

Tests 

Code changes

 - None

Side effects

 - None

Related changes

### Release note

- No release note

### DIFF
--- a/pkg/httputil/httputil.go
+++ b/pkg/httputil/httputil.go
@@ -14,9 +14,10 @@
 package httputil
 
 import (
+	"net/http"
+
 	"github.com/pingcap/ticdc/cdc/model"
 	"github.com/pingcap/ticdc/pkg/security"
-	"net/http"
 )
 
 // Client wraps an HTTP client and support TLS requests.
--- a/pkg/httputil/httputil.go
+++ b/pkg/httputil/httputil.go
@@ -14,10 +14,9 @@
 package httputil
 
 import (
-	"net/http"
-
 	"github.com/pingcap/ticdc/cdc/model"
 	"github.com/pingcap/ticdc/pkg/security"
+	"net/http"
 )
 
 // Client wraps an HTTP client and support TLS requests.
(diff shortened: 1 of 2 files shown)


hi-rustin (commented):
Doesn't cobra have a more elegant solution?

pkg/httputil/httputil.go
```diff
@@ -62,3 +61,13 @@ func IsFiltered(whiteList string, feedState model.FeedState) bool {
 	}
 	return whiteList == string(feedState)
 }
+
+// IsContainNonASCII return true if a string contain non-ascii char
+func IsContainNonASCII(s string) bool {
```
hi-rustin, line 66: Why do we need to put that method in httputil and maybe we need some tests?

asddongmen: ok, i will add some tests.
put it into httputil is because it may be use in http api.

pkg/httputil/httputil.go
```diff
@@ -62,3 +62,13 @@ func IsFiltered(whiteList string, feedState model.FeedState) bool {
...
+	for _, r := range s {
+		if r < 0 || r > 127 {
```
hi-rustin, line 69: Maybe we can use unicode.MaxASCII?

Also, will this value be less than 0?

asddongmen: good
it seems will not less than 0.

PR #2262, opened 2021-07-10
refactor(cli): rewrite all cmds

hi-rustin:
### What problem does this PR solve? 

cc: https://github.com/pingcap/ticdc/issues/2168

### What is changed and how it works?

Rewrite all cmds.

The adjustments are as follows:
1. 28 global variables were cleared
2. abstracted out the client construction factory for future refactoring
3. cleared all init methods
4. more standard flag collection
5. more standard parameter validation
6. go project compliant directory
7. better comment
8. cleaned up a lot of code that shouldn't appear in util

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)

Code changes

None

Side effects

None

Related changes

 - Need to cherry-pick to the release branch

### Release note

- No release note

Issue #2199, opened 2021-07-01
Standardized project structure


hi-rustin:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently our project file structure has some confusion, which is not very convenient and clear when reading the code and finding it.

**Describe the feature you'd like:**

Aligning the standard structure of other go projects:

- https://github.com/golang-standards/project-layout
- https://github.com/kubernetes/kubernetes

**Describe alternatives you've considered:**

N/A

**Teachability, Documentation, Adoption, Migration Strategy:**

N/A

hi-rustin:
/assign 

I am working on this.

hi-rustin:
I have spent a lot of time in the past improving it. I think it's mostly done. Maybe the next effort will be to merge the code of ticdc and dm.

Issue #2168, opened 2021-06-28
Unified and standardized ticdc parameter validation


hi-rustin:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently ticdc has three parameters that need to be verified:
- [x] cli parameters [WIP] [I am writing a detailed design proposal](https://docs.google.com/document/d/12vbelCSsKhbGFY9HyBlEmDzEBGtBsd6aOeSytvJiwgc/edit?usp=sharing).
- [ ] API parameters
- [ ] configuration file validation

But currently these checks are not uniform enough, and there is no documented rule to standardize this part of the check logic. So we have a lot of issues about parameter validation.

1. https://github.com/pingcap/ticdc/issues/1718
2. https://github.com/pingcap/ticdc/issues/1716
3. https://github.com/pingcap/ticdc/issues/1706
4. https://github.com/pingcap/ticdc/issues/1704

**Describe the feature you'd like:**

Unify validation parameters by category and provide standardized documentation to guide the validation of parameters.

**Describe alternatives you've considered:**

N/A

hi-rustin:
/assign 

I am working on this.

hi-rustin:
It's been a long time since I pushed it forward again, and most of the work has been done.

PR #2365, opened 2021-07-25
cli(refactor): add version cmd

hi-rustin:
part of #2168

cc: #2199

split from #2262

add test and version cmds.

### Check List 

Tests 

None

Code changes

Side effects

None

Related changes

None

### Release note

- No release note


pkg/cmd/cmd.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/pingcap/ticdc/pkg/cmd/version"
+	"github.com/spf13/cobra"
+)
+
+// NewCmd creates the root command.
+func NewCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "cdc",
+		Short: "CDC",
+		Long:  `Change Data capture`,
```
overvenus, line 29:
```suggestion
		Long:  `Change Data Capture`,
```

pkg/cmd/test/test.go
```diff
@@ -0,0 +1,108 @@
+// Copyright 2021 PingCAP, Inc.
```
overvenus, line 1: Seem like we do not need the test cmd any more. Cc @amyangfei

overvenus (commented):
Rest LGTM

pkg/cmd/test/test.go
amyangfei: Yes, it can be removed

hi-rustin: Removed! Thanks for your review!

overvenus (approved):

amyangfei (approved):

hi-rustin:
💚 💙 💜 💛 ❤️
Thanks!

PR #2366, opened 2021-07-25
cli(feat): add metrics

hi-rustin:
part of #2168

Completed to-do items.
add metrics.

### Check List 

Tests 

None

Code changes

Side effects

None

Related changes

None

### Release note

- No release note

/cc @overvenus  @amyangfei

### DIFF
--- a/pkg/cmd/factory/factory_impl.go
+++ b/pkg/cmd/factory/factory_impl.go
@@ -18,13 +18,11 @@ import (
 	"strings"
 	"time"
 
-	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/ticdc/cdc/kv"
 	cmdconetxt "github.com/pingcap/ticdc/pkg/cmd/context"
 	"github.com/pingcap/ticdc/pkg/security"
 	"github.com/pingcap/ticdc/pkg/version"
-	"github.com/prometheus/client_golang/prometheus"
 	pd "github.com/tikv/pd/client"
 	"go.etcd.io/etcd/clientv3"
 	etcdlogutil "go.etcd.io/etcd/pkg/logutil"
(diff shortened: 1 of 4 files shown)


hi-rustin:
@overvenus We should not be able to register grpc metric more than once, should I expose the metric in kv and use it in cmd?

overvenus (commented):
Could you add a panel to ticdc.json?

pkg/cmd/factory/factory_impl.go
```diff
@@ -95,10 +101,10 @@ func (f *factoryImpl) EtcdClient() (*kv.CDCEtcdClient, error) {
 		TLS:         tlsConfig,
 		LogConfig:   &logConfig,
 		DialTimeout: 30 * time.Second,
-		// TODO(hi-rustin): add gRPC metrics to Options.
-		// See also: https://github.com/pingcap/ticdc/pull/2341#discussion_r673018537.
 		DialOptions: []grpc.DialOption{
 			grpcTLSOption,
+			grpc.WithUnaryInterceptor(kv.GRPCMetrics.UnaryClientInterceptor()),
+			grpc.WithStreamInterceptor(kv.GRPCMetrics.StreamClientInterceptor()),
```
amyangfei, line 107: These metrics are initialized in command line tool, which has nothing to do with cdc server, will they be pulled by the prometheus server?

overvenus: I think 'cdc server' also use the pd client.

amyangfei: Yes, `cdc server` does use pd client, but this factory is not used in `cdc server`.

hi-rustin: Yes, it is not used in the server. But I understand they will use the same metric and it will be pulled as long as it is registered to the same registry? I'm not sure, I need to try it.

hi-rustin: I read the server code carefully and we should be able to use the construction factory on the server as well, with some minor changes.

hi-rustin:
/close

Since they are different processes, they should not take effect.

So let's leave the todos for now and see how we can monitor it after the cli refactoring work is done.

cc: @amyangfei  @overvenus

Issue #2301, opened 2021-07-16
always pull the old value internally


hi-rustin:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Because our new collation feature needs the old value to work properly. However, some users will not use old value, but may use new collation, so it needs to be configured by the user. (Users don't even know if they have new collation on)

To simplify this configuration and to make it easier for the user, we decided to:

    1. always use old value internally.
    2. the data output to the user needs to be in the correct format (without the old value configuration on, the old value cannot be output)

**Describe the feature you'd like:**
ticdc always pull the old value internally.

**Describe alternatives you've considered:**
None

hi-rustin:
A more detailed document: https://github.com/SuperSupeng/TIDB-tutorial/blob/main/content/TiCDC/%E9%9C%80%E6%B1%82/oldValue.md

hi-rustin:
cc: https://github.com/pingcap/ticdc/issues/2073

This seems to be the same requirement?  @pepezzzz

amyangfei:
Closed by #2367

PR #2367, opened 2021-07-25
sink(refactor): Refine functions and improve event splitting code

hi-rustin:
ref: #2301

Refine functions and improve event splitting code.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

None

Side effects

None

Related changes

### Release note

- No release note


hi-rustin:
Marking the change requires only one 👍 because it is simple and does not modify any logic.

cdc/processor/pipeline/sink.go
```diff
@@ -215,6 +177,75 @@ func (n *sinkNode) emitEvent(ctx pipeline.NodeContext, event *model.PolymorphicE
...
+	return !(handleKeyCount == equivalentHandleKeyCount)
+}
+
+// splitUpdateEvent splits an update event into a delete and an insert event.
+func splitUpdateEvent(updateEvent *model.PolymorphicEvent) ([]*model.PolymorphicEvent, error) {
+	if updateEvent == nil {
+		return nil, errors.New("nil event cannot be split")
+	}
+
+	var result []*model.PolymorphicEvent
```
overvenus, line 213:
```suggestion
	result := make([]*model.PolymorphicEvent, 0, 2)
```

overvenus (commented):
Rest LGTM

cdc/processor/pipeline/sink.go
```diff
@@ -215,6 +177,75 @@ func (n *sinkNode) emitEvent(ctx pipeline.NodeContext, event *model.PolymorphicE
...
+			}
+		}
+	}
+
+	// If the handle key columns are not updated, so we do **not** need to split the event row.
+	return !(handleKeyCount == equivalentHandleKeyCount)
+}
+
+// splitUpdateEvent splits an update event into a delete and an insert event.
+func splitUpdateEvent(updateEvent *model.PolymorphicEvent) ([]*model.PolymorphicEvent, error) {
```
overvenus, line 208: Or maybe we can retrun an array of 2 model.PolymorphicEvent to save alloctions.
```suggestion
func splitUpdateEvent(updateEvent *model.PolymorphicEvent) ([2]*model.PolymorphicEvent, error) {
```

cdc/processor/pipeline/sink.go
hi-rustin: Good catch!

cdc/processor/pipeline/sink.go
hi-rustin: Solved by your previous comment.

overvenus: Returning an array is better than pre-allocation, as it does not need alloction in heap, see more: https://godbolt.org/z/K1Mdde5Kd

(I should mark https://github.com/pingcap/ticdc/pull/2367#discussion_r676615140 outdated before making this comment. 😅)

hi-rustin: Yes, but does this perhaps affect the readability of the code? Because we need to access it via index both for writing and reading.

```diff
-n.eventBuffer = append(n.eventBuffer, splitEvents...)
+n.eventBuffer = append(n.eventBuffer, splitEvents[0],splitEvents[1]) or n.eventBuffer = append(n.eventBuffer, splitEvents[:]...)
-result = append(result, &deleteEvent)
+result[0] = &deleteEvent
```

Also, if updateEvent is nil, what does our parameter detection need to return? 
```go
	var result [2]*model.PolymorphicEvent
	if updateEvent == nil {
		return result, errors.New("nil event cannot be split")
	}

```
Wouldn't that be a bit strange? Maybe we can keep it to improve code readability?

ben1009: return `[2]*model.PolymorphicEvent` also kind of strange to me 🤓

amyangfei: I prefer `func splitUpdateEvent(updateEvent *model.PolymorphicEvent) (delEv, insertIv, *model.PolymorphicEvent, error)`

hi-rustin: Cool! Let me address that.

hi-rustin: Addressed. Thank you all for the review!

amyangfei (approved):

overvenus (approved):

PR #2384, opened 2021-07-27
cli: Fix cli using memory sorter when working on 4.0.x cluster

liuzix:
- Unified Sorter has become the default on 4.0.13 for the 4.0.X branch, but the current cli treats all 4.0.X versions as a whole and disables Unified Sorter by default.

Close https://github.com/pingcap/ticdc/issues/2373

- Better version discrimination.

### Check List 

Tests 

 - Unit test

Side effects

Related changes

### Release note

```release-note
Fix CLI back-compatibility
```


overvenus (approved):
LGTM

overvenus:
Does it need to be cherry-picked to release-4.0?

pkg/version/check.go
```diff
@@ -165,7 +191,7 @@ func GetTiCDCClusterVersion(captureInfos []*model.CaptureInfo) (TiCDCClusterVers
 		if captureInfo.Version != "" {
 			ver, err = semver.NewVersion(removeVAndHash(captureInfo.Version))
 		} else {
-			ver = semver.New("4.0.1")
+			ver = MinTiCDCVersion
```
overvenus, line 194: Can it be TiCDCClusterVersionUnknown?

cmd/client_changefeed.go
```diff
@@ -270,14 +270,19 @@ func verifyChangefeedParameters(ctx context.Context, cmd *cobra.Command, isCreat
...
-		}
-		log.Warn("The TiCDC cluster is built from 4.0-release branch, the old-value and unified-sorter are disabled by default.")
+		log.Warn("The TiCDC cluster is built from an older version, disabling old value by default.",
+			zap.String("version", cdcClusterVer.String()))
+	}
+
+	sortEngineFlag := cmd.Flag("sort-engine")
+	if !sortEngineFlag.Changed && !cdcClusterVer.ShouldEnableUnifiedSorterByDefault() {
+		sortEngine = model.SortInMemory
+		log.Warn("The TiCDC cluster is built form an older version, disabling Unified Sorter by default",
```
amyangfei, line 282:
```suggestion
		log.Warn("The TiCDC cluster is built from an older version, disabling Unified Sorter by default",
```

pkg/version/check.go
```diff
@@ -144,14 +147,37 @@ func CheckStoreVersion(ctx context.Context, client pd.Client, storeID uint64) er
...
+// ShouldEnableUnifiedSorterByDefault returns whether Unified Sorter should be enabled by default
+func (v *TiCDCClusterVersion) ShouldEnableUnifiedSorterByDefault() bool {
+	if v.isUnknown {
+		// we assume the unknown version to be the latest version
+		return true
+	}
+	// x >= 4.0.13 AND x != 5.0.0-rc
+	if v.String() == "5.0.0-rc" {
+		return false
+	}
```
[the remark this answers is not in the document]
liuzix: There has been examples of users trying to use 5.0.0-rc in production, so I suggest that we be careful here.

amyangfei (approved):

PR #2406, opened 2021-07-29
pipeline: skip stack backtrace in hot path

overvenus:

Save CPU by skipping stack backtrace.

image

Skips stack backtrace.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/pipeline/pipeline.go
+++ b/pkg/pipeline/pipeline.go
@@ -100,6 +100,8 @@ func (p *Pipeline) driveRunner(ctx context.Context, previousRunner, runner runne
 	}
 }
 
+var pipelineTryAgainError error = cerror.ErrPipelineTryAgain.FastGenByArgs()
+
 // SendToFirstNode sends the message to the first node
 func (p *Pipeline) SendToFirstNode(msg *Message) error {
 	p.closeMu.Lock()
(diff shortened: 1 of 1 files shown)


pkg/pipeline/pipeline.go
```diff
@@ -115,7 +115,9 @@ func (p *Pipeline) SendToFirstNode(msg *Message) error {
 	select {
 	case p.header <- msg:
 	default:
-		return cerror.ErrPipelineTryAgain.GenWithStackByArgs()
+		// Do not call `GenWithStackByArgs` in the hot path,
+		// it consumes lots of CPU.
+		return cerror.ErrPipelineTryAgain
```
amyangfei, line 120: What about creating a global error using `cerror.ErrPipelineTryAgain.FastGenByArgs()` and returning with it

overvenus: Do we have to call `GenXXX` or `FastGenXXX`? Is there any error handling idiom of TiCDC?

amyangfei: `cerror.ErrPipelineTryAgain` is a global error instance, which is used to generate an error. To use a global rfc error is dangerous because we may change it by accident.

amyangfei (approved):

liuzix (approved):

PR #2430, opened 2021-08-01
cmd(refactor): add capture cmd

hi-rustin:

part of #2168

cc: #2199

split from #2262

add capture cmd

### Check List 

Tests 

 - Unit test

Code changes

None

Side effects

None

Related changes

None

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/cmd/cli/cli.go
+++ b/pkg/cmd/cli/cli.go
@@ -69,7 +69,7 @@ func NewCmdCli() *cobra.Command {
 		},
 	}
 
-	// Binding the `cmd` command flags.
+	// Binding the `cli` command flags.
 	o.addFlags(cmds)
 	cf.AddFlags(cmds)
 
--- a/pkg/cmd/cli/cli.go
+++ b/pkg/cmd/cli/cli.go
@@ -1,0 +1,126 @@
+// Copyright 2021 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


hi-rustin:
/hold

- [ ] tests
- [ ] refactoring

pkg/cmd/cli/cli.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/chzyer/readline"
+	"github.com/mattn/go-shellwords"
+	"github.com/pingcap/ticdc/pkg/cmd/factory"
+	"github.com/pingcap/ticdc/pkg/cmd/util"
+	"github.com/pingcap/ticdc/pkg/logutil"
+	"github.com/spf13/cobra"
+)
+
+// options defines flags for the `cli` command.
+type options struct {
```
3AceShowHand, line 30: I would prefer `option`

hi-rustin: Nice catch! Thanks. Addressed.

hi-rustin:
/unhold

Ready for review.

/run-all-tests

/uncc lonng leoppro

/cc @amyangfei  @overvenus

pkg/cmd/cli/cli.go
hi-rustin: Although I addressed it, I don't know if it really pays off. Because it's not uniform with other commands anymore. Also the singular and plural need to be taken care of when maintaining the code. Maybe we shouldn't obsess about singular and plural here? I don't know, what do you think? cc: @amyangfei @overvenus

overvenus (approved):
LGTM

pkg/cmd/cli/cli.go
amyangfei: I notice kubectl uses `xxxOptions`.
Is it because there is only one option in this struct, what if we add another option? Maybe use `options` is enough. Not a big problem

amyangfei (approved):

Issue #2351, opened 2021-07-22
Subtask tracing issue for redo log based replication to implement eventually consistency


amyangfei:
## Basic Design

This solution will introduce a redo log mechanism, the redo logs are persisted before sink module, the new workflow is as follows:
- Before each sink replicates row data to downstream, saves the sorted and mounted row data to persistent storage.
- Each sink uploads the resolved ts of persistent row data as a watermark and TiCDC owner aggregates the watermarks to generate a global watermark.
- Then each sink can replicate row data with commit-ts less than the global watermark.

When a disaster happens in the upstream, TiCDC can switch to an offline mode, which means it only consumes data in local persistent storage.

## Task breakdown

19 weeks totally, without optional and won't do features.

Note we will separate the sub tasks into several stages

1. Implement basic feature, and provide a basic available version with eventually consistent replication feature. In this stage we will finish subtasks including 3, 4, 5.
2. Test stage, in which we will finish subtask 6.
3. Improve stage, in which we will finish subtask 1, 2.

---

Detail subtasks as follows

1. Separate etcd from PD, adding standalone meta store. 3weeks (**Won't do**)
   - [ ] Add start-up parameters, add embedded etcd server in TiCDC. 1week
   - [ ] Add `legacy-mode` to provide backward compatibility, refine upgrade process. 1week
   - [ ] Refactor internal pd/etcd client usage, refine session management. 1week
   - [ ] Implement changefeed migration from PD to the embedded Etcd cluster. (optional)

2. TiCDC offline mode. 4weeks  (**Won't do**)
   - [ ] Add a new upstream resource manager, including pd client and tikv store.
       - [ ] When upstream is abnormal, the upstream resource manager can report error. 1week
       - [ ] Add upstream liveness checker in upstream resource management. 1week
       - [ ] TiCDC server maintains the upstream manager, add offline mode switch feature (also works during rolling upgrade or network jitter). 1week
   - [ ] Support to consume redo log in offline mode. 1week 
   - [x] #2581

3. Binary based redo log consumer. 2weeks
   - [x] Support to consume redo log with cdc binary only (via command such as `cdc cli redo apply`). 2weeks #2814
   
4. Redo log mechanism. 4weeks
    - [x] #2372
       - [x] #2435
       - [ ] On-premise: high availability for redo log storage. (Won't do currently)
       - [x] #2538
[...]

PR #2462, opened 2021-08-04
redo-log: add basic framework for redo log manager and writer

amyangfei:

This PR adds basic logic for #2351 to complete subtasks 4-2, 5-1 and 5-2, and implements the following workflows.

- Add interface of redo log writer and basic redo log manager.
- Add necessary changefeed config, currently it is simple, containing consistent level and consistent storage.
- Implement basic watermark mechanism, adding `RedoLogTs` in each table sink.
- Add a mock blackhole redo log writer and verify the replication can run normally with consistent replication feature enabled.

There still exists some remaining works, including

   - [ ] Add admin command, such as remove a table from redo log writer
   - [ ] Better interface abstraction
   - [x] Unit test for new code

   As described in PR description

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/redo/manager.go
+++ b/cdc/redo/manager.go
@@ -1,0 +1,254 @@
+// Copyright 2021 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing pemissions and
+// limitations under the License.
+
+package redo
+
+import (
+	"context"
+	"math"
+	"sort"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/pingcap/log"
+	"github.com/pingcap/ticdc/cdc/model"
+	"github.com/pingcap/ticdc/pkg/config"
+	cerror "github.com/pingcap/ticdc/pkg/errors"
+	"go.uber.org/zap"
+)
+
+var updateRtsInterval = time.Second
+
+type consistentLevelType string
+
+const (
+	consistentLevelNormal   consistentLevelType = "normal"
+	consistentLevelEventual consistentLevelType = "eventual"
+)
+
+type consistentStorage string
+
+const (
+	consistentStorageLocal     consistentStorage = "local"
+	consistentStorageS3        consistentStorage = "s3"
+	consistentStorageBlackhole consistentStorage = "blackhole"
+)
+
+// IsValidConsistentLevel checks whether a give consistent level is valid
+func IsValidConsistentLevel(level string) bool {
+	switch consistentLevelType(level) {
+	case consistentLevelNormal, consistentLevelEventual:
+		return true
+	default:
+		return false
+	}
+}
+
+// IsValidConsistentStorage checks whether a give consistent storage is valid
+func IsValidConsistentStorage(storage string) bool {
+	switch consistentStorage(storage) {
+	case consistentStorageLocal, consistentStorageS3, consistentStorageBlackhole:
+		return true
+	default:
+		return false
+	}
+}
+
+// LogManager defines an interface that is used to manage redo log
+type LogManager interface {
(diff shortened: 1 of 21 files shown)


cdc/redo/storage.go
```diff
@@ -0,0 +1,42 @@
...
+	Index [][]int                `json:"index"`
+}
+
+// LogWriter is a writer abstraction for redo log storage layer
+type LogWriter interface {
+	WriteLog(ctx context.Context, tableID model.TableID, log []*RowRedoLog) (offset uint64, err error)
+	// FlushLog sends resolved-ts from table pipeline to log writer, it is
+	// essential to flush when a table doesn't have any row change event for
+	// sometime, and the resolved ts of this table should be moved forward.
+	FlushLog(ctx context.Context, tableID model.TableID, ts uint64) error
```
overvenus, line 34:
```suggestion
	FlushLog(ctx context.Context, tableID model.TableID, resolvedTs uint64) error
```

cdc/redo/storage.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"context"
+
+	"github.com/pingcap/ticdc/cdc/model"
+)
+
+// RowRedoLog stores one RowChangedEvent and its index column
+type RowRedoLog struct {
+	Row   *model.RowChangedEvent `json:"row"`
+	Index [][]int                `json:"index"`
```
overvenus, line 25: Could you add some comments about why `Index [][]int` is needed?

cdc/redo/manager.go
```diff
@@ -0,0 +1,224 @@
...
+const (
+	consistentLevelNomal    consistentLevelType = "normal"
+	consistentLevelEventual consistentLevelType = "eventual"
+)
+
+type consistentStorage string
+
+const (
+	consistentStorageLocal     consistentStorage = "local"
+	consistentStorageS3        consistentStorage = "s3"
```
overvenus, line 44:
```suggestion
	// consistentStorageS3        consistentStorage = "s3"
```

Please do not add it until TiCDC actually supports it.

cdc/redo/manager.go
```diff
@@ -0,0 +1,224 @@
...
+)
+
+var updateOffsetInterval = time.Second
+
+type consistentLevelType string
+
+const (
+	consistentLevelNomal    consistentLevelType = "normal"
+	consistentLevelEventual consistentLevelType = "eventual"
+)
```
overvenus, line 38: Why do we need level? What's the difference between normal and eventual?

cdc/redo/manager.go
```diff
@@ -0,0 +1,224 @@
...
+		return m.tableIDs[i] >= tableID
+	})
+	if i < len(m.tableIDs) && m.tableIDs[i] == tableID {
+		copy(m.tableIDs[i:], m.tableIDs[i+1:])
+		m.tableIDs = m.tableIDs[:len(m.tableIDs)-1]
+		delete(m.offsets, tableID)
+	} else {
+		log.Warn("remove a table not maintained in redo log manager", zap.Int64("table-id", tableID))
+	}
+	// TODO: send remove table command to redo log writer
```
overvenus, line 186: Do we need a new method in `LogWriter`?

cdc/redo/storage.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// LogWriter is a writer abstraction for redo log storage layer
+type LogWriter interface {
+	WriteLog(ctx context.Context, tableID model.TableID, log []*RowRedoLog) (offset uint64, err error)
```
overvenus, line 30: What is the offset?

cdc/redo/manager.go
amyangfei: Add it when we really need

cdc/redo/manager.go
amyangfei: normal is default mode, without eventually consistent guarantee

cdc/redo/manager.go
amyangfei: It is a prototype PR, the first redo log writer we will add is s3, IMHO it is ok to leave it here

cdc/redo/storage.go
amyangfei: will be changed to resolved ts

PR #2474, opened 2021-08-06
*: use values to reduce GC pressure 

overvenus:

Use prefer using values for channel types to reduce GC pressure. Same as https://github.com/pingcap/ticdc/pull/2415

Left: master
Right: this patch

Max (master): ~80ms
Max (this patch): ~35ms
image
image
image
image

### Check List 

Tests 

### Release note 

```release-note
None
```


pkg/workerpool/pool_impl.go
```diff
@@ -280,9 +280,6 @@ func (w *worker) run(ctx context.Context) error {
 		case <-ctx.Done():
 			return errors.Trace(ctx.Err())
 		case task := <-w.taskCh:
-			if task == nil {
-				return cerrors.ErrWorkerPoolEmptyTask.GenWithStackByArgs()
-			}
```
amyangfei, line 285: Should we check empty task here?

pkg/workerpool/pool_impl.go
```diff
@@ -246,7 +246,7 @@ type task struct {
 }
 
 type worker struct {
-	taskCh       chan *task
+	taskCh       chan task
```
amyangfei, line 249: What about adding a benchmark to the workerpool after this change

amyangfei (commented):
Curious about whether CPU usage has any change after this memory optimization?

overvenus:
> Curious about whether CPU usage has any change after this memory optimization?

@amyangfei No significant change on CPU usage, see PR description.

amyangfei:
> > Curious about whether CPU usage has any change after this memory optimization?
> 
> @amyangfei No significant change on CPU usage, see PR description.

From the monitor can we draw the following conclusions, after using values

- Average CPU usage increases a little(a **rough** observation), but not significant compared to the whole CPU usage
- CPU usage is more stable, fewer fluctuation in monitor

pkg/workerpool/pool_impl.go
overvenus: I don't think it necessary as task is used only in this file, and it never be empty.

overvenus:
```
# Master
goos: linux
goarch: amd64
pkg: github.com/pingcap/ticdc/pkg/workerpool
cpu: Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz
BenchmarkWorkerpool-40            476822              2585 ns/op             272 B/op          5 allocs/op

# This patch
goos: linux
goarch: amd64
pkg: github.com/pingcap/ticdc/pkg/workerpool
cpu: Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz
BenchmarkWorkerpool-40            480866              2453 ns/op             256 B/op          4 allocs/op
```

amyangfei (approved):

ben1009 (approved):

overvenus:
#2201 reduces 3 allocations, well done @liuzix !

```
goos: linux
goarch: amd64
pkg: github.com/pingcap/ticdc/pkg/workerpool
cpu: Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz
BenchmarkWorkerpool-40            837417              1579 ns/op              56 B/op          1 allocs/op
PASS
ok      github.com/pingcap/ticdc/pkg/workerpool 2.343s
```

PR #2481, opened 2021-08-06
version: prohibit operating TiCDC clusters across major and minor versions

overvenus:

Prohibit operating TiCDC clusters across major versions

Add version check for both TiDB cluster and TiCDC cluster.

### Check List 

Tests 

```
tiup playground v5.1.0 --tiflash 0 --ticdc 1
cdc cli --pd http://127.0.0.1:2379 capture list // ok

tiup playground v5.0.3 --tiflash 0 --ticdc 1
cdc cli --pd http://127.0.0.1:2379 capture list // error "CDC:ErrVersionIncompatible"

tiup playground v4.0.14 --tiflash 0 --ticdc 1
cdc cli --pd http://127.0.0.1:2379 capture list // error "CDC:ErrVersionIncompatible"

// Or there is no TiCDC cluster
cdc cli --pd http://127.0.0.1:2379 capture list // error TiCDC cluster is unknown, please start TiCDC cluster
```

Related changes

### Release note 

```release-note
Prohibit operating TiCDC clusters across major and minor versions
```

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -196,6 +196,21 @@ func newCliCommand() *cobra.Command {
 				return errors.Annotatef(err, "fail to open PD client, pd=\"%s\"", cliPdAddr)
 			}
 			ctx := defaultContext
+			_, captureInfos, err := cdcEtcdCli.GetCaptures(ctx)
+			if err != nil {
+				return err
+			}
+			cdcClusterVer, err := version.GetTiCDCClusterVersion(captureInfos)
+			if err != nil {
+				return err
+			}
+			isUnknownVersion, err := version.CheckTiCDCClusterVersion(cdcClusterVer)
+			if err != nil {
+				return err
+			}
+			if isUnknownVersion {
+				return errors.NewNoStackError("TiCDC cluster is unknown, please start TiCDC cluster")
+			}
 			errorTiKVIncompatible := true // Error if TiKV is incompatible.
 			err = version.CheckClusterVersion(ctx, pdCli, pdEndpoints[0], credential, errorTiKVIncompatible)
 			if err != nil {
--- a/cmd/client_changefeed.go
+++ b/cmd/client_changefeed.go
@@ -242,7 +242,10 @@ func newQueryChangefeedCommand() *cobra.Command {
 	return command
 }
 
-func verifyChangefeedParameters(ctx context.Context, cmd *cobra.Command, isCreate bool, credential *security.Credential, captureInfos []*model.CaptureInfo) (*model.ChangeFeedInfo, error) {
+func verifyChangefeedParameters(
+	ctx context.Context, cmd *cobra.Command, isCreate bool,
+	credential *security.Credential, cdcClusterVer version.TiCDCClusterVersion,
+) (*model.ChangeFeedInfo, error) {
 	if isCreate {
 		if sinkURI == "" {
 			return nil, errors.New("Creating changefeed without a sink-uri")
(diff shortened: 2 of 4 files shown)


hi-rustin:
It looks like we are having some problems with our integration test environment.

cmd/client.go
```diff
@@ -196,6 +196,21 @@ func newCliCommand() *cobra.Command {
...
+			if err != nil {
+				return err
+			}
+			isUnknownVersion, err := version.CheckTiCDCClusterVersion(cdcClusterVer)
+			if err != nil {
+				return err
+			}
+			if isUnknownVersion {
+				return errors.NewNoStackError("TiCDC cluster is unknown, please start TiCDC cluster")
+			}
```
amyangfei, line 213: Note this change means when we execute cli commands, TiCDC server must be alive, some points should be payed attention to
1. Some commands don't require TiCDC server alive or we need to execute some commands if TiCDC server can't start up(because of some bugs), such as `cdc cli unsafe`, what should we do in this scenario?
2. Whether some test cases are broken after this change. (Considering the scenario we create a changefeed first, and then start up TiCDC servers)

overvenus: Thanks for the remaindering, I will put the restriction to the changefeed sub-command only.

overvenus: > Whether some test cases are broken after this change. (Considering the scenario we create a changefeed first, and then start up TiCDC servers)

I think creating a changefeed is meaningless without a running TiCDC cluster. Also if the TiCDC starts too late, it is possible that TiDB GC safepoint exceeds changefeed start-ts.

overvenus: @leoppro PTAL

.github/workflows/integration.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         run: docker-compose -f docker-compose-avro.yml pull --ignore-pull-failures
 
       - name: TiKV version
-        run: docker run pingcap/tikv:release-4.0-nightly -V
+        run: docker run pingcap/tikv:nightly -V
```
hi-rustin, line 48: Perhaps a conflict with https://github.com/pingcap/ticdc/issues/2442? We need to correspond with the branch.

overvenus: Yes, we need to change versions in cherry-picks.

cmd/util_test.go
```diff
@@ -92,3 +93,41 @@ func (s *utilsSuite) TestVerifyPdEndpoint(c *check.C) {
...
 	c.Assert(verifyPdEndpoint("https://aa", true), check.IsNil)
 }
+
+func (s *utilsSuite) TestNeedVerifyCmd(c *check.C) {
+	defer testleak.AfterTest(c)()
+
+	// Test command tree:
+	//      root
+	//   l        r
+	//     ll   rr  rl
```
overvenus, line 103:
```suggestion
	//   l        r
	// ll       rr  rl
```

.github/workflows/integration.yml
hi-rustin: Only the actions in the master branch take effect, so we might need to add it now if we really want to run these tests in the release branch. But it looks like you don't have it enabled in the release branch yet.

integration/integration.go
```diff
@@ -113,9 +113,9 @@ func testMySQLWithCheckingOldvValue() {
 
 func runTests(cases []framework.Task, env framework.Environment) {
 	log.SetLevel(zapcore.DebugLevel)
-	env.Setup()
 
 	for i := range cases {
+		env.Setup()
```
amyangfei, line 118: Some behaviors duplicated in `env.Reset` and `env.Setup`, and part of them are not idempotent, such as `createConnector` in avro test. I have changed `createConnector` to idempotent (ignore the 409 error returned from JdbcSinkConnector

amyangfei: Is there any special reason to change the location of `env.Setup()` @overvenus

overvenus: `env.Reset` resets all TiCDC data in PD, cli may think TiCDC cluster is not up, so we need `env.Setup` makes sure TiCDC cluster is ready to serve.

hi-rustin (approved):
LGTM 👍 

Perhaps we should put this check within each command, rather than collecting a list each time and then determining whether a check is needed. But that's okay I think we can merge it now and I'll refactor it later.(in new cli)

amyangfei (approved):

PR #2510, opened 2021-08-12
EtcdWorker: fix EtcdWorker snapshot isolation

liuzix:
- EtcdWorker's snapshot isolation is violated if there are two concurrent transactions one deleting and the other putting to the same key.
- EtcdWorker's snapshot isolation is violated if while `A` is running and trying to create a key `k`, `B` and `C` creates and deletes `k` sequentially. 

- For the first case, we use equality of `mod_revision` to check for write-set intersection instead of inequality.
- For the second case, we use a global `deleteCounter` as a key to make deletions impossible to race with other operations.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Fix a bug in metadata management
```


pkg/orchestrator/etcd_worker.go
```diff
@@ -45,6 +47,8 @@ type EtcdWorker struct {
 	barrierRev int64
 	// prefix is the scope of Etcd watch
 	prefix util.EtcdPrefix
+
+	deleteCounter int64
```
3AceShowHand, line 51: can you add a comment about `deleteCounter`, it maybe hard for other to get the reason why we need it.

pkg/orchestrator/etcd_worker.go
```diff
@@ -53,19 +57,27 @@ type etcdUpdate struct {
 	revision int64
 }
 
+type rawStateEntry struct {
```
3AceShowHand, line 60: can you add a comment about this?

pkg/orchestrator/etcd_worker.go
```diff
@@ -201,10 +235,25 @@ func (worker *EtcdWorker) syncRawState(ctx context.Context) error {
 		return errors.Trace(err)
 	}
 
-	worker.rawState = make(map[util.EtcdKey][]byte)
+	worker.rawState = make(map[util.EtcdKey]rawStateEntry)
 	for _, kv := range resp.Kvs {
+		if string(kv.Key) == worker.prefix.String()+deletionCounterKey {
+			var err error
```
3AceShowHand, line 241: this logic look the same as above in `handleEvent.Put`, maybe add a function ?

pkg/orchestrator/etcd_worker.go
```diff
@@ -177,6 +189,25 @@ func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session,
 }
 
 func (worker *EtcdWorker) handleEvent(_ context.Context, event *clientv3.Event) {
+	if string(event.Kv.Key) == worker.prefix.String()+deletionCounterKey {
+		switch event.Type {
+		case mvccpb.PUT:
+			var err error
+			worker.deleteCounter, err = strconv.ParseInt(string(event.Kv.Value), 10, 64)
+			if worker.deleteCounter <= 0 {
+				log.Panic("unexpected delete counter", zap.Int64("value", worker.deleteCounter))
```
overvenus, line 198: Could you log key and value?

overvenus (commented):
Could you add some tests?

pkg/orchestrator/etcd_worker.go
liuzix: The key is actually fixed. There is only one possible value if this occurs in running the cdc server.

pkg/orchestrator/etcd_worker.go
```diff
@@ -45,6 +47,13 @@ type EtcdWorker struct {
 	barrierRev int64
 	// prefix is the scope of Etcd watch
 	prefix util.EtcdPrefix
+	// deleteCounter maintains a local copy of a value stored in Etcd used to keep track of
+	// how many deletes have been committed by an EtcdWorker watching this key prefix.
+	// This mechanism is necessary as a workaround to correctly detect write-write conflicts when at least a transaction
+	// commits a delete, because deletes in Etcd resets the mod-revision of a key, making it difficult to use
+	// it as a unique version identifier to implement a `compare-and-swap` semantics, which is essential for implementing
+	// snapshot isolation for Reactor ticks.
```
overvenus, line 55: Nit: try wrapping lines, keep them within 80 chars width.

pkg/orchestrator/etcd_worker.go
```diff
@@ -269,10 +312,22 @@ func (worker *EtcdWorker) applyPatches(ctx context.Context, patches []DataPatch)
...
+			hasDelete = true
 		}
 		ops = append(ops, op)
 	}
 
+	if hasDelete {
+		ops = append(ops, clientv3.OpPut(worker.prefix.String()+deletionCounterKey, fmt.Sprint(worker.deleteCounter+1)))
+	}
+	if worker.deleteCounter > 0 {
+		cmps = append(cmps, clientv3.Compare(clientv3.Value(worker.prefix.String()+deletionCounterKey), "=", fmt.Sprint(worker.deleteCounter)))
```
overvenus, line 324: What about `path.Join(worker.prefix.String(), deletionCounterKey)` ?

overvenus (commented):
Rest LGTM

pkg/orchestrator/etcd_worker.go
```diff
@@ -319,3 +374,25 @@ func (worker *EtcdWorker) cleanUp() {
...
+
+	var err error
+	worker.deleteCounter, err = strconv.ParseInt(string(value), 10, 64)
+	if worker.deleteCounter <= 0 {
+		log.Panic("unexpected delete counter", zap.Int64("value", worker.deleteCounter))
+	}
+	if err != nil {
+		// This should never happen unless Etcd server has been tampered with.
+		log.Panic("strconv failed. Unexpected Etcd state.", zap.Error(err))
+	}
```
amyangfei, line 397: Better to check `err != nil` first, when err is not nil, `worker.deleteCounter` is 0

pkg/orchestrator/etcd_worker.go
```diff
@@ -269,10 +312,22 @@ func (worker *EtcdWorker) applyPatches(ctx context.Context, patches []DataPatch)
 			op = clientv3.OpPut(key.String(), string(value))
 		} else {
 			op = clientv3.OpDelete(key.String())
+			hasDelete = true
 		}
 		ops = append(ops, op)
 	}
 
+	if hasDelete {
+		ops = append(ops, clientv3.OpPut(worker.prefix.String()+deletionCounterKey, fmt.Sprint(worker.deleteCounter+1)))
```
amyangfei, line 321: Does this key exist in etcd permanently and always increase, do we need any gc mechanism

liuzix: Yes, it will always increase and it is not expected to be deleted unless the CDC cluster is reset. Developing a GC mechanism is unnecessary because the benefit is small.

pkg/orchestrator/etcd_worker.go
liuzix: We can do this. But the Etcd prefixes are all normalized here, so it must not end with `/`.

amyangfei (approved):

overvenus (approved):

PR #2523, opened 2021-08-13
http: enable all sub paths matching in pprof

amyangfei:

Fix some pprof requests (`/debug/pprof/xxx`) return 404

```
➜  curl http://172.16.6.135:8301/debug/pprof/goroutine\?debug=1
404 page not found%
```

Should match path `/debug/pprof/:any` and handled by `pprof.Index`

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/http_router.go
+++ b/cdc/http_router.go
@@ -84,6 +84,7 @@ func newRouter(capture2 *capture.Capture) *gin.Engine {
 	pprofGroup := router.Group("/debug/pprof")
 	{
 		pprofGroup.GET("", gin.WrapF(pprof.Index))
+		pprofGroup.GET("/:any", gin.WrapF(pprof.Index))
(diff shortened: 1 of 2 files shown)


cdc/http_router.go
```diff
@@ -84,6 +84,7 @@ func newRouter(capture2 *capture.Capture) *gin.Engine {
 	pprofGroup := router.Group("/debug/pprof")
 	{
 		pprofGroup.GET("", gin.WrapF(pprof.Index))
+		pprofGroup.GET("/:any", gin.WrapF(pprof.Index))
```
overvenus, line 87: @asddongmen Please add tests for pprof endpoints later.

overvenus (approved):
LGTM

asddongmen (approved):

PR #2530, opened 2021-08-15
actor: add actor framework

overvenus:

Introduce actor framework.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/actor/actor.go
+++ b/pkg/actor/actor.go
@@ -1,0 +1,433 @@
+package actor
+
+import (
+	"container/list"
+	"context"
+	"runtime"
+	"sync"
+
+	"github.com/pingcap/log"
+	cerrors "github.com/pingcap/ticdc/pkg/errors"
+	"github.com/pingcap/ticdc/pkg/pipeline"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+type Actor interface {
+	// Receive handles messages that are sent to actor's mailbox.
+	//
+	// The ctx is only for cancellation, and an actor must be aware of
+	// the cancellation.
+	//
+	// If it returns true, then the actor will be rescheduled and polled later.
+	// If it returns false, then the actor will be removed from Router and
+	// polled if there are still messages in its mailbox.
+	//
+	// Once it returns false, it must always return false.
+	Receive(ctx context.Context, msgs []pipeline.Message) (closed bool)
+}
+
+// ID is ID of actors.
+type ID int
+
+// Mailbox sends messages to an actor.
+// Mailbox is threadsafe.
+type Mailbox interface {
+	ID() ID
+	// Send message to its actor.
+	// It's a non-blocking send, returns ErrMailboxFull when it's full.
+	Send(msg pipeline.Message) error
+	// SendB sends message to its actor, blocks when it's full.
+	// It may return context.Canceled or context.DeadlineExceeded.
+	SendB(ctx context.Context, msg pipeline.Message) error
+	tryReceive() (pipeline.Message, bool)
+	len() int
+}
+
+func NewMailbox(id ID, cap int) Mailbox {
+	return &mailbox{
+		id: id,
+		ch: make(chan pipeline.Message, cap),
+	}
+}
+
+var _ Mailbox = (*mailbox)(nil)
+
+type mailbox struct {
+	id ID
+	ch chan pipeline.Message
+}
+
+func (m *mailbox) ID() ID {
+	return m.id
+}
+
+var errMailboxFull = cerrors.ErrMailboxFull.FastGenByArgs()
+
+func (m *mailbox) Send(msg pipeline.Message) error {
+	select {
+	case m.ch <- msg:
+		return nil
+	default:
+		return errMailboxFull
+	}
+}
+
+func (m *mailbox) SendB(ctx context.Context, msg pipeline.Message) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case m.ch <- msg:
+		return nil
+	}
+}
+
+func (m *mailbox) tryReceive() (msg pipeline.Message, ok bool) {
+	select {
+	case msg, ok = <-m.ch:
+		return
+	default:
+	}
+	return
+}
+
+func (m *mailbox) len() int {
+	return len(m.ch)
+}
+
+type proc struct {
+	mb     Mailbox
+	actor  Actor
+	closed bool
+}
+
+type ready struct {
+	sync.Mutex
+	cond *sync.Cond
+
+	// TODO: replace with a memory efficient queue.
+	queue   list.List
+	procs   map[ID]struct{}
+	stopped bool
+}
+
+func (rd *ready) stop() {
+	rd.Lock()
+	rd.stopped = true
+	rd.Unlock()
+	rd.cond.Broadcast()
+}
+
+func (rd *ready) enqueueLocked(p *proc, force bool) {
+	id := p.mb.ID()
+	if _, ok := rd.procs[id]; !ok || force {
+		rd.queue.PushBack(p)
+		rd.procs[id] = struct{}{}
+	}
+}
+
+func (rd *ready) schedule(p *proc) {
+	rd.Lock()
+	rd.enqueueLocked(p, false)
+	rd.Unlock()
+	rd.cond.Signal()
+}
+
+func (rd *ready) scheduleN(procs []*proc) {
+	rd.Lock()
+	for _, p := range procs {
+		rd.enqueueLocked(p, false)
+	}
+	rd.Unlock()
+	rd.cond.Broadcast()
+}
+
+type Router struct {
+	rd *ready
+
+	// Map of ID to proc
+	procs sync.Map
+}
+
+func newRouter() *Router {
+	r := &Router{
+		rd: &ready{},
+	}
+	r.rd.cond = sync.NewCond(&r.rd.Mutex)
+	r.rd.procs = make(map[ID]struct{})
+	r.rd.queue.Init()
+	return r
+}
+
+var errActorNotFound = cerrors.ErrActorNotFound.FastGenByArgs()
+
+func (r *Router) Send(id ID, msg pipeline.Message) error {
+	value, ok := r.procs.Load(id)
+	if !ok {
+		return errActorNotFound
+	}
+	p := value.(*proc)
+	err := p.mb.Send(msg)
+	if err != nil {
+		return err
+	}
+	r.rd.schedule(p)
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+const maxWorkerNum = 64
+
+// The default size of polled actor batch.
+const defaultActorBatchSize = 1
+
+// The default size of receive message batch.
+const defaultMsgBatchSizePerActor = 64
+
+// NewSystemBuilder returns a new system builder.
+func NewSystemBuilder(name string) *SystemBuilder {
```
sdojjy, line 286: name is not used

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+	// If it returns false, then the actor will be removed from Router and
+	// polled if there are still messages in its mailbox.
+	// Once it returns false, it must always return false.
+	//
+	// We choose message to have a concrete type instead of an interface to save
+	// memory allocation.
+	Poll(ctx context.Context, msgs []pipeline.Message) (closed bool)
+}
+
+// ID is ID of actors.
```
3AceShowHand, line 49: this comment is a little mouthful. how about changing it to `ID for actors.`

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+
+func (s *System) handleFatal(msg string, id ID) {
+	handler := defautlFatalhandler
+	if s.fatalHandler != nil {
+		handler = s.fatalHandler
+	}
+	handler(msg, id)
+}
+
+func defautlFatalhandler(msg string, id ID) {
```
sdojjy, line 516: typo, change to defaultFatalHandler

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+				}
+				batchMsg = append(batchMsg, msg)
+			}
+			if len(batchMsg) == 0 {
+				continue
+			}
+			s.metricMsgBatch.Observe(float64(len(batchMsg)))
+
+			pollStartTime := time.Now()
+			close := !p.actor.Poll(ctx, batchMsg)
```
sdojjy, line 470: close is the builtin function name, use another variable name

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+		return err
+	}
+	r.rd.schedule(p)
+	return nil
+}
+
+// SendB sends a message to an actor, blocks when it's full.
+// ErrActorNotFound when the actor not found.
+// Canceled or DeadlineExceeded when the context is canceled or done.
+func (r *Router) SendB(ctx context.Context, id ID, msg pipeline.Message) error {
```
sdojjy, line 214: too many duplicate codes with Send

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+	ID() ID
+	// Send a message to its actor.
+	// It's a non-blocking send, returns ErrMailboxFull when it's full.
+	Send(msg pipeline.Message) error
+	// SendB sends a message to its actor, blocks when it's full.
+	// It may return context.Canceled or context.DeadlineExceeded.
+	SendB(ctx context.Context, msg pipeline.Message) error
+
+	// Try to receive a message.
+	// It is must not block and should only be called by System.
```
3AceShowHand, line 64: not block -> nonblocking

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+func (m *mailbox) SendB(ctx context.Context, msg pipeline.Message) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case m.ch <- msg:
+		return nil
+	}
+}
+
+func (m *mailbox) tryReceive() (msg pipeline.Message, ok bool) {
```
3AceShowHand, line 110: how about use explicit return, to keep the code format the same, implicit is a little hard to reason about.

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+	r := &Router{
+		rd: &ready{},
+	}
+	r.rd.cond = sync.NewCond(&r.rd.Mutex)
+	r.rd.procs = make(map[ID]struct{})
+	r.rd.queue.Init()
+	return r
+}
+
+var errActorNotFound = cerrors.ErrActorNotFound.FastGenByArgs()
```
3AceShowHand, line 192: how about move to the top of the file, group with `errMailboxFull`

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+	}
+	r.rd.schedule(p)
+	return nil
+}
+
+// Broadcast a message to all actors in the router.
+// The message may be dropped when a actor is full.
+func (r *Router) Broadcast(msg pipeline.Message) {
+	batchSize := 128
+	ps := make([]*proc, 0, batchSize)
```
3AceShowHand, line 232: can we set this `ps` as a struct field, to avoid reallocation?

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+type SystemBuilder struct {
+	name                 string
+	numWorker            int
+	actorBatchSize       int
+	msgBatchSizePerActor int
+
+	fatalHandler func(string, ID)
+}
+
+// The max number of workers of a system.
```
3AceShowHand, line 276: how about group related const together and put at the top of the file.

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+// The main poll of actor system.
+func (s *System) poll(ctx context.Context) {
+	batchP := make([]*proc, 0, s.actorBatchSize)
+	batchMsg := make([]pipeline.Message, 0, s.msgBatchSizePerActor)
+	s.metricProcBatch.Observe(float64(len(batchP)))
+	rd := s.rd
+	rd.Lock()
+
+	startTime := time.Now()
+	s.metricWorkingWorkers.Add(1)
```
3AceShowHand, line 426: `inc`

pkg/actor/actor.go
3AceShowHand: this pieces of code is quite long and makes others feel complicate, extract some method to make it easier to read.

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package actor
```
3AceShowHand, line 14: I found several different structs defined in this file, we can split them into different files

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+// ready is a centralize notification struct, shared by a router and a system.
+// It schedules notification and actors.
+type ready struct {
+	sync.Mutex
+	cond *sync.Cond
+
+	// TODO: replace with a memory efficient queue,
+	// e.g., an array based queue to save allocation.
+	queue   list.List
+	procs   map[ID]struct{}
```
asddongmen, line 139: Maybe we can add a comment here to explain the use of this field `procs`

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+}
+
+// Broadcast a message to all actors in the router.
+// The message may be dropped when a actor is full.
+func (r *Router) Broadcast(msg pipeline.Message) {
+	batchSize := 128
+	ps := make([]*proc, 0, batchSize)
+	r.procs.Range(func(key, value interface{}) bool {
+		p := value.(*proc)
+		if err := p.mb.Send(msg); err != nil {
```
asddongmen, line 235: should we skip this `p` when `err != nil` ?

pkg/actor/actor.go
```diff
@@ -0,0 +1,518 @@
...
+			pollStartTime := time.Now()
+			close := !p.actor.Poll(ctx, batchMsg)
+			if close {
+				if !p.closed {
+					p.closed = true
+					present := s.router.remove(p.mb.ID())
+					if !present {
+						s.handleFatal(
+							"try to remove non-existent actor", p.mb.ID())
+					}
```
asddongmen, line 478: maybe `running := p.actor.Poll(ctx, batchMsg)` if better?

pkg/actor/system.go
```diff
@@ -0,0 +1,457 @@
...
+	rd.Lock()
+	for _, p := range procs {
+		rd.enqueueLocked(p, false)
+	}
+	rd.Unlock()
+	rd.cond.Broadcast()
+}
+
+// batchReceiveProcs receives ready procs into batchP.
+func (rd *ready) batchReceiveProcs(batchP []*proc, batchSize int) []*proc {
```
amyangfei, line 115: ditto

pkg/actor/system.go
```diff
@@ -0,0 +1,457 @@
...
+}
+
+// Send a message to an actor. It's a non-blocking send.
+// ErrMailboxFull when the actor full.
+// ErrActorNotFound when the actor not found.
+func (r *Router) Send(id ID, msg message.Message) error {
+	value, ok := r.procs.Load(id)
+	if !ok {
+		return errActorNotFound
+	}
```
amyangfei, line 154: In which scenario could `actor not found` happen

docs/actor-system.wsd
```diff
@@ -0,0 +1,37 @@
...
+class Router {
+- ready
+- procs
+}
+
+class System {
+- ready
+}
+
+class ready {
```
nongfushanquan, line 27: those class name seem too popular

pkg/actor/system.go
```diff
@@ -0,0 +1,457 @@
...
+
+// proc is wrapper of a running actor.
+type proc struct {
+	mb     Mailbox
+	actor  Actor
+	closed bool
+}
+
+// batchReceiveMsgs receives messages into batchMsg.
+func (p *proc) batchReceiveMsgs(batchMsg []message.Message, batchSize int) []message.Message {
```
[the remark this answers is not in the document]
amyangfei: Then why pass the `batchMsg` here, just to save memory allocation?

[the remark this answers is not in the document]
amyangfei: IMHO this code snippet should be carefully used, for two reasons
- Pass a slice would copy underlying value. In this scenario it is zero
- If the slice update beyonds the original capacity underlying array will be reallocated. In this scenario it is controlled by `batchSize`

Another way is to use `func (p *proc) batchReceiveMsgs(batchMsg *[]message.Message, batchSize int)`

PR #2563, opened 2021-08-18
tests: Add integration test for http api

asddongmen:
#2412 

Add integration tests for all Open API.

### Check List 

Tests 

 - No code

### Release note 

```release-note
None
```


asddongmen:
I triger `/run-all-tests` times to check if the added tests is stable.

tests/http_api/util/test_case.py
```diff
@@ -0,0 +1,326 @@
+import sys
+import requests as rq
+import time
+import json
+
+# http status code
+OK = 200
```
hi-rustin, line 7: Could we use the existing http library?

tests/http_api/util/test_case.py
```diff
@@ -0,0 +1,326 @@
...
+    url = BASE_URL+"/changefeeds"
+    # create changefeed
+    for i in range(1, 4):
+        data = {
+            "changefeed_id": "changefeed-test"+str(i),
+            "sink_uri": "blackhole://",
+            "ignore_ineligible_table": True
+        }
+        # set sink_uri
+        if i == 1 and sink_uri != "":
```
hi-rustin, line 29: Why do we need to determine if it is 1?

tests/http_api/util/test_case.py
```diff
@@ -0,0 +1,326 @@
...
+
+    print("pass test: create changefeed")
+
+
+def list_changefeed():
+    # test state: all
+    url = BASE_URL+"/changefeeds?state=all"
+    resp = rq.get(url)
+    assert resp.status_code == OK
+    data = resp.json()
```
hi-rustin, line 55: It seems we don't use this value?

tests/http_api/run.sh
```diff
@@ -0,0 +1,61 @@
...
+    python $CUR/util/test_case.py get_status
+
+    python $CUR/util/test_case.py create_changefeed "$SINK_URI"
+
+    run_sql "CREATE table test.simple(id int primary key, val int);"
+    run_sql "CREATE table test.\`simple-dash\`(id int primary key, val int);"
+    # wait for above sql done in the up source
+    sleep 1
+
+    python $CUR/util/test_case.py list_changefeed
```
hi-rustin, line 42: It looks like these tests have sequential dependencies, maybe we need to add comments to suggest not to modify these sequences?

hi-rustin (approved):
rest LGTM 👍

tests/http_api/run.sh
amyangfei: What about using an array for sequential cases, such as
```
sequential_cases=("list_changefeed" "get_changefeed" ...)
for case in $sequential_cases; do
    python $CUR/util/test_case.py $case
done
```

asddongmen: good idea ! thanks!

tests/http_api/util/test_case.py
asddongmen: done

tests/http_api/util/test_case.py
asddongmen: done

amyangfei (approved):

PR #2604, opened 2021-08-23
owner: fix schema snapshot read from TiKV

### LINKED ISSUE #2603: DDL handling fails when changefeed starts at FinishTs of a schema change
## Bug Report

- Truncate some tables, and while the DDLs are still being executed, kill the owner.

- The cluster can recover and replication continues

3. What did you see instead?
- `CDC:ErrSchemaStorageTableMiss`.

4. Versions of the cluster

        ```
        v5.2.0
        ```

5. Cause

The TS used for snapshot read from TiKV is not correctly set. https://github.com/pingcap/ticdc/blob/04fa75b2ae194e481e8a82c0cd12bf7d8ca2a249/cdc/owner/schema.go#L44

```go
        var meta *timeta.Meta
	if kvStorage != nil {
		var err error
		meta, err = kv.GetSnapshotMeta(kvStorage, startTs)  // wrong! should be startTs-1
		if err != nil {
			return nil, errors.Trace(err)
		}
	}
	// We do a snapshot read of the metadata from TiKV at (startTs-1) instead of startTs,
	// because the DDL puller might send a DDL at startTs, which would cause schema conflicts if
	// the DDL's result is already contained in the snapshot.
	schemaSnap, err := entry.NewSingleSchemaSnapshotFromMeta(meta, startTs-1, config.ForceReplicate)
	if err != nil {
		return nil, errors.Trace(err)
	}
```

liuzix:

Close #2603 

- Fixed the timestamp with which to read TiKV snapshot for schema.
- Improved integration test case `ddl_reentrant` so that similar bugs will not be reintroduced in the future.

### Check List 

Tests 

 - Integration test

Related changes

### Release note 

```release-note
Fixed a bug in DDL handling when the owner restarts.
```

### DIFF
--- a/cdc/owner/schema.go
+++ b/cdc/owner/schema.go
@@ -41,7 +41,7 @@ func newSchemaWrap4Owner(kvStorage tidbkv.Storage, startTs model.Ts, config *con
 	var meta *timeta.Meta
 	if kvStorage != nil {
 		var err error
-		meta, err = kv.GetSnapshotMeta(kvStorage, startTs)
+		meta, err = kv.GetSnapshotMeta(kvStorage, startTs-1)
 		if err != nil {
 			return nil, errors.Trace(err)
 		}
--- a/cdc/owner/schema.go
+++ b/cdc/owner/schema.go
@@ -61,7 +61,7 @@ func newSchemaWrap4Owner(kvStorage tidbkv.Storage, startTs model.Ts, config *con
 		schemaSnapshot: schemaSnap,
 		filter:         f,
 		config:         config,
-		ddlHandledTs:   startTs,
+		ddlHandledTs:   startTs - 1,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


cdc/owner/schema.go
```diff
@@ -61,7 +61,7 @@ func newSchemaWrap4Owner(kvStorage tidbkv.Storage, startTs model.Ts, config *con
 		schemaSnapshot: schemaSnap,
 		filter:         f,
 		config:         config,
-		ddlHandledTs:   startTs,
+		ddlHandledTs:   startTs - 1,
```
amyangfei, line 64: Can we keep using `startTs` in this function, L44, L52 and L 64, but change the caller to pass `checkpointTs - 1`, from `cdc/owner/changefeed.go`, and add comment the `startTs` is `[startTs, ...)` in this function

nongfushanquan: sound reasonable

amyangfei (approved):

asddongmen (approved):

Issue #2612, opened 2021-08-24
Some transaction could generate empty values in open protocol


amyangfei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

   1. Setup a tidb and ticdc cluster, create changefeed to replicate all tables.
   2. Execute the following SQLs in upstream TiDB

```sql
mysql root@127.0.0.1:test> create table t2 (id varchar(32) not null primary key, a int);
Query OK, 0 rows affected

mysql root@127.0.0.1:test> begin;
Query OK, 0 rows affected

mysql root@127.0.0.1:test> insert into t2 values (2,2);
Query OK, 1 row affected

mysql root@127.0.0.1:test> delete from t2 where id = 2;
Query OK, 1 row affected

mysql root@127.0.0.1:test> commit;
Query OK, 0 rows affected
```

2. What did you expect to see?

   The mq sink should not output any row changes.

3. What did you see instead?
    
   The following output using `kafka-console-consumer.sh`
   `?{"ts":427235938124431362,"scm":"test","tbl":"t2","rid":2,"t":1}        {}`

    The log of MySQL sink, note both `columns` and `pre-columns` are empty

   ```
   [2021/08/24 11:36:52.871 +08:00] [DEBUG] [mysql.go:1111] ["prepare DMLs"] [rows="[{\"start-ts\":427236053953019905,\"commit-ts\":427236055407656962,\"row-id\":2,\"table\":{\"db-name\":\"test\",\"tbl-name\":\"t2\",\"tbl-id\":59,\"is-partition\":false},\"table-info-version\":427235957064073226,\"replica-id\":0,\"columns\":null,\"pre-columns\":null}]"] [sqls="[]"] [values="[]"]
   [2021/08/24 11:36:52.871 +08:00] [DEBUG] [mysql.go:994] ["Exec Rows succeeded"] [changefeed=test-cf] ["num of Rows"=0] [bucket=0]
   ```

4. Versions of the cluster

    - Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

        All versions since v4.0.x

    - TiCDC version (execute `cdc version`):

        All versions since v4.0.x

PR #2615, opened 2021-08-24
sink/mq: fix empty value in open protocol

amyangfei:

Fix the open protocol part of #2612 

Note we still need to confirm what is the right behavior in protocol canal, canal-json, avro, maxwell for empty row change.

Ignore empty row change events in open protocol

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
Fix open protocol, don't output an empty value when there is no change in one transaction. 
```

### DIFF
--- a/cdc/sink/codec/json.go
+++ b/cdc/sink/codec/json.go
@@ -395,6 +395,12 @@ func (d *JSONEventBatchEncoder) EncodeCheckpointEvent(ts uint64) (*MQMessage, er
 
 // AppendRowChangedEvent implements the EventBatchEncoder interface
 func (d *JSONEventBatchEncoder) AppendRowChangedEvent(e *model.RowChangedEvent) (EncoderResult, error) {
+	// Some transactions could generate empty row change event, such as
+	// begin; insert into t (id) values (1); delete from t where id=1; commit;
+	// Just ignore these row changed events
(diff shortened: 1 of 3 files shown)


cdc/sink/codec/json.go
```diff
@@ -395,6 +395,12 @@ func (d *JSONEventBatchEncoder) EncodeCheckpointEvent(ts uint64) (*MQMessage, er
 
 // AppendRowChangedEvent implements the EventBatchEncoder interface
 func (d *JSONEventBatchEncoder) AppendRowChangedEvent(e *model.RowChangedEvent) (EncoderResult, error) {
+	// Some transactions could generate empty row change event, such as
+	// begin; insert into t (id) values (1); delete from t where id=1; commit;
+	// Just ignore these row changed events
+	if len(e.Columns) == 0 && len(e.PreColumns) == 0 {
```
hi-rustin, line 401: Do we need to add the issue link here?

hi-rustin (approved):
LGTM

liuzix (approved):

cdc/sink/codec/json.go
amyangfei: IMHO we can trace the issue from commit message and pr description.

PR #2695, opened 2021-08-31
server, version: check and wait if cluster is incompatible

overvenus:

Check and wait if cluster incompatible before starting CDC.

### Check List 

Tests 

 - Unit test

Related changes

### Release note 

```release-note
Check and wait if TiDB cluster is incompatible
```

### DIFF
--- a/cdc/server.go
+++ b/cdc/server.go
@@ -362,13 +362,10 @@ func checkAndWaitClusterVersion(
 	ctx context.Context, pdClient pd.Client, pdEndpoints []string, security *config.SecurityConfig,
 ) {
 	const backoffOneSecond = 1000
-	const errorTiKVIncompatible = true
-
 	retry.Do(ctx, func() error {
 		var err error
 		for _, pdEndpoint := range pdEndpoints {
-			err = version.CheckClusterVersion(
-				ctx, pdClient, pdEndpoint, security, errorTiKVIncompatible)
+			err = version.CheckClusterVersion(ctx, pdClient, pdEndpoint, security)
 			if err == nil {
 				break
 			}
--- a/cdc/server.go
+++ b/cdc/server.go
@@ -30,6 +30,7 @@ import (
 	"github.com/pingcap/ticdc/pkg/config"
 	cerror "github.com/pingcap/ticdc/pkg/errors"
 	"github.com/pingcap/ticdc/pkg/httputil"
+	"github.com/pingcap/ticdc/pkg/retry"
 	"github.com/pingcap/ticdc/pkg/util"
 	"github.com/pingcap/ticdc/pkg/version"
 	tidbkv "github.com/pingcap/tidb/kv"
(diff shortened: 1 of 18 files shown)


pkg/version/check.go
```diff
@@ -35,9 +35,9 @@ import (
 )
 
 var (
-	// minPDVersion is the version of the minimal compatible PD.
+	// MinPDVersion is the version of the minimal compatible PD.
 	// TODO bump 5.2.0-alpha once PD releases.
-	minPDVersion *semver.Version = semver.New("5.1.0-alpha")
+	MinPDVersion *semver.Version = semver.New("5.1.0-alpha")
```
amyangfei, line 40: Should this version be upgraded to 5.2.0-alpha, ditto for `MinTiKVVersion`

cdc/server.go
```diff
@@ -367,3 +357,25 @@ func findBestDataDir(candidates []string) (result string, ok bool) {
...
+	ctx context.Context, pdClient pd.Client, pdEndpoints []string, security *config.SecurityConfig,
+) {
+	const backoffOneSecond = 1000
+	const errorTiKVIncompatible = true
+
+	retry.Do(ctx, func() error {
+		var err error
+		for _, pdEndpoint := range pdEndpoints {
+			err = version.CheckClusterVersion(
+				ctx, pdClient, pdEndpoint, security, errorTiKVIncompatible)
```
amyangfei, line 371: Seems the parameter `errorTiKVIncompatible` can be removed from function `CheckClusterVersion`

pkg/cmd/factory/factory_impl.go
```diff
@@ -161,7 +161,7 @@ func (f factoryImpl) PdClient() (pd.Client, error) {
 
 	// TODO: we need to check all pd endpoint and make sure they belong to the same cluster.
 	// See also: https://github.com/pingcap/ticdc/pull/2341#discussion_r673021305.
-	err = version.CheckClusterVersion(ctx, pdClient, pdEndpoints[0], credential, true)
+	err = version.CheckClusterVersion(ctx, pdClient, pdEndpoints[0], credential)
```
lonng, line 164: I think we need to check all PD endpoints to avoid failure when the first PD endpoint is unreachable.

pkg/version/check.go
```diff
@@ -26,34 +26,29 @@ import (
...
 	pd "github.com/tikv/pd/client"
-	"go.uber.org/zap"
 )
 
 var (
-	// minPDVersion is the version of the minimal compatible PD.
-	// TODO bump 5.2.0-alpha once PD releases.
-	minPDVersion *semver.Version = semver.New("5.1.0-alpha")
+	// MinPDVersion is the version of the minimal compatible PD.
+	MinPDVersion *semver.Version = semver.New("5.2.0-alpha")
```
lonng, line 37: Can we remove all the version-related global variables? I think we can check whether the major/minor version matched TiKV/PD/TiCDC.

cdc/server.go
```diff
@@ -177,6 +165,10 @@ func (s *Server) Run(ctx context.Context) error {
 		return err
 	}
 
+	// Check cluster version and wait if it's incompatible.
+	// We start status server first to not block tiup cluster upgrading.
+	checkAndWaitClusterVersion(ctx, s.pdClient, s.pdEndpoints, conf.Security)
+
 	return s.run(ctx)
```
amyangfei, line 172: Currently `/status` API is available after starts server starts, however https://github.com/pingcap/ticdc/issues/2691 is considering change the behavior, which means the `/status` API will be available after capture info is persisted to etcd.

cdc/server.go
```diff
@@ -356,3 +348,22 @@ func findBestDataDir(candidates []string) (result string, ok bool) {
...
+	ctx context.Context, pdClient pd.Client, pdEndpoints []string, security *config.SecurityConfig,
+) {
+	const backoffOneSecond = 1000
+	retry.Do(ctx, func() error {
+		var err error
+		for _, pdEndpoint := range pdEndpoints {
+			err = version.CheckClusterVersion(ctx, pdClient, pdEndpoint, security)
+			if err == nil {
+				break
+			}
```
amyangfei, line 362: Two scenarios in this block
- More than 3 PDs in upstream, one of them is down, TiCDC can start up, which is the HA feature, in this scenario TiCDC only check one PD.
- Upstream upgrade, should TiCDC waits all PDs upgrade to the required version, then TiCDC should check all PDs.

PR #2696, opened 2021-08-31
cli(refactor): enable new cmds and cleanup old cmds

hi-rustin:
part of #2168

cc: #2199

split from #2262

close https://github.com/pingcap/ticdc/issues/1704

enable new cmds and cleanup old cmds.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

None

Related changes

None

### Release note 

```release-note
None
```


hi-rustin:
/hold

- [ ] tests
- [ ] Version compatibility check

hi-rustin:
/unhold

Ready to review.

/cc @asddongmen 

https://github.com/pingcap/ticdc/pull/2696/commits/accc80e504a0bf8c4d7be89e4bed0c19345bf0bc 

Maybe this check code is a bit repetitive, but probably in terms of maintainability it is better than if we go through a list and then check it globally probably.
Also, I didn't pass the etcd client used as a common option for two reasons.
1. it might not be created, because it might turn off the check.
2. We will later use the http API to refactor the cli, and passing the etcd cli may cause higher refactoring costs. So simply create and use it here.

pkg/cmd/util/helper.go
```diff
@@ -22,6 +22,9 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/pingcap/ticdc/cdc/kv"
+	"github.com/pingcap/ticdc/pkg/version"
+
```
amyangfei, line 27: can remove the blank line

hi-rustin: Done. Thanks for your review!

pkg/cmd/cli/cli_changefeed_helper_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	err = confirmLargeDataGap(cmd, currentTs, startTs)
+	c.Assert(err, check.ErrorMatches, "abort changefeed create or resume")
+
+	// check start ts more than 1 day before current ts, and type Y when confirming
+	err = ioutil.WriteFile(path, []byte("Y"), 0o644)
+	c.Assert(err, check.IsNil)
+	f, err = os.Open(path)
+	c.Assert(err, check.IsNil)
+	os.Stdin = f
```
overvenus, line 58: What about `io.Pipe()`?

hi-rustin: Sorry, I didn't understand what you meant. The method uses standard input, and here a file is needed, so we passed the parameters by writing a file.

pkg/cmd/util/helper.go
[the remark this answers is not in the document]
hi-rustin: Already done!

asddongmen (approved):

ben1009 (approved):

pkg/cmd/cli/cli_changefeed_helper_test.go
hi-rustin: Oh, I see what you mean. We don't need to reset stdin, we just need to copy the reader into stdin.

hi-rustin: @overvenus Let me merge it first (for fear of conflicts 😨) and I'll follow up with a small patch to improve it. Because maybe it might be a little faster not to create the file?

hi-rustin:
The refactoring of our cli is finally over! Thanks for your review!  💚 💙 💜 💛 ❤️

pkg/cmd/cli/cli_changefeed_helper_test.go
hi-rustin: @overvenus It looks like we need to use os.Pipe() to do this, but then the file is created again. So maybe we don't need to modify it.

PR #2713, opened 2021-09-01
check: refine CheckClusterVersion 

asddongmen:
make  func `CheckClusterVersion` more  intuitive to use

Separate the logic of checking pd version into `CheckPDVersion` .

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - None 

Related changes

 - None 

### Release note 

```release-note
None
```


pkg/cmd/factory/factory_impl.go
```diff
@@ -161,7 +161,7 @@ func (f factoryImpl) PdClient() (pd.Client, error) {
 
 	// TODO: we need to check all pd endpoint and make sure they belong to the same cluster.
 	// See also: https://github.com/pingcap/ticdc/pull/2341#discussion_r673021305.
-	err = version.CheckClusterVersion(ctx, pdClient, pdEndpoints[0], credential, true)
+	err = version.CheckClusterVersion(ctx, pdClient, pdEndpoints[0:1], credential, true, false)
```
overvenus, line 164:
```suggestion
	err = version.CheckClusterVersion(ctx, pdClient, pdEndpoints, credential, true, false)
```

pkg/version/check.go
```diff
@@ -75,8 +74,9 @@ func removeVAndHash(v string) string {
 }
 
 // CheckClusterVersion check TiKV and PD version.
+// If checkAllPD is false, need only one PD alive and match the cdc version.
 func CheckClusterVersion(
-	ctx context.Context, client pd.Client, pdHTTP string, credential *security.Credential, errorTiKVIncompat bool,
+	ctx context.Context, client pd.Client, pdAddrs []string, credential *security.Credential, errorTiKVIncompat bool, checkAllPD bool,
```
overvenus, line 79: Who checks all PD?

asddongmen: It is for possible needs in the future.

overvenus: So we can add it in the future.

asddongmen: done

pkg/cmd/factory/factory_impl.go
```diff
@@ -161,7 +161,7 @@ func (f factoryImpl) PdClient() (pd.Client, error) {
 
 	// TODO: we need to check all pd endpoint and make sure they belong to the same cluster.
```
hi-rustin, line 162: Maybe this comment became strange after this change?

pkg/version/check.go
```diff
@@ -75,9 +74,9 @@ func removeVAndHash(v string) string {
 }
 
 // CheckClusterVersion check TiKV and PD version.
+// need only one PD alive and match the cdc version.
 func CheckClusterVersion(
-	ctx context.Context, client pd.Client, pdHTTP string, credential *security.Credential, errorTiKVIncompat bool,
-) error {
+	ctx context.Context, client pd.Client, pdAddrs []string, credential *security.Credential, errorTiKVIncompat bool) error {
```
hi-rustin, line 79: Only one pd is needed to be valid, but we pass in all the addresses, which is a bit strange for me. But I think it's acceptable.

What do you think?

pkg/version/check.go
```diff
@@ -86,17 +85,30 @@ func CheckClusterVersion(
...
+}
+
+// CheckPDVersion check PD version.
+func CheckPDVersion(ctx context.Context, pdAddr string, credential *security.Credential) error {
 	// See more: https://github.com/pingcap/pd/blob/v4.0.0-rc.1/server/api/version.go
 	pdVer := struct {
 		Version string `json:"version"`
 	}{}
 
+	httpCli, err := httputil.NewClient(credential)
```
hi-rustin, line 105: Maybe httpClient is better?

hi-rustin (commented):
reset LGTM.

hi-rustin:
Also, maybe we need to change the pr title since it will be used as the commit msg, but it looks like we are mostly modifying the version checking of the pd.

pkg/version/check.go
asddongmen: Because we need to find a suitable one from all PDs, we can only pass all PDs in.
Do you have a better way? If so, please tell me, thanks.

pkg/version/check.go
asddongmen: good~

pkg/cmd/factory/factory_impl.go
asddongmen: I still think it is necessary to leave this comment, because it means that we need to ensure that all PD are belong to a same cluster, but this function does not guarantee it now.

hi-rustin: I mean it might be weird to leave it here?

asddongmen: done

pkg/version/check.go
```diff
@@ -86,22 +85,35 @@ func CheckClusterVersion(
...
+}
+
+// CheckPDVersion check PD version.
+func CheckPDVersion(ctx context.Context, pdAddr string, credential *security.Credential) error {
 	// See more: https://github.com/pingcap/pd/blob/v4.0.0-rc.1/server/api/version.go
 	pdVer := struct {
 		Version string `json:"version"`
 	}{}
 
+	httpClient, err := httputil.NewClient(credential)
```
3AceShowHand, line 105: I would prefer to name it `cli`, we know that it's an HTTP client by the right-hand side.

3AceShowHand (approved):

pkg/version/check.go
hi-rustin: cli is a very confusing acronym (first thought is command line interface), and it is not a recognized acronym for the client, it does not have a consensus acronym. So I prefer to change it to client to reduce confusion.

overvenus (approved):
LGTM

PR #2417, opened 2021-07-29
sink: add barrier in underlying sink to support accruate stop

amyangfei:
### What problem does this PR solve? 

Fix #2230 

This PR introduces a new `Barrier` API in sink interface, which is a synchronous function to wait all events to be flushed in underlying sink, note once `Barrier` is called, the resolved ts can't be pushed until the Barrier call returns.

We call `Barrier` when table sink is closing (via the `Close` method of table sink), and before `Barrier` is called, we draw back all existing events in sink buffer.

Note the cherry-pick cannot be merged into release-4.0, release-5.0 until new owner is enabled in these branches.

### What is changed and how it works?

- The `Barrier` implementation
   - MySQL sink: wait checkpoint-ts to reach max received resolved-ts
   - MQ sink: TODO
- Drawback mechanism in buffer sink
   - Buffer sink stores rows by table in memory, and emit rows to underlying sink only resolved ts is forwarded.
   - Hook a channel from sink manager to notify buffer sink to draw back rows of a table, and wait callback

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
Fix a bug that multiple processors could write the same table when this table is re-scheduling
```

PR #2716, opened 2021-09-01
pipeline: set table status to stopped after sink is closed

amyangfei:

Fix a data inconsistency issue caused by conflict write from two captures, it was a remaining bug after #2417

- The table status of a pipeline should not set to stopped until the underlying sink is closed.
- Remove unused `StoppedTs` parameter in pipeline command.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
Fix a bug that multiple processors could write the same table when this table is re-scheduling
```

### DIFF
--- a/cdc/processor/pipeline/table.go
+++ b/cdc/processor/pipeline/table.go
@@ -100,8 +100,7 @@ func (t *tablePipelineImpl) UpdateBarrierTs(ts model.Ts) {
 // AsyncStop tells the pipeline to stop, and returns true is the pipeline is already stopped.
 func (t *tablePipelineImpl) AsyncStop(targetTs model.Ts) bool {
 	err := t.p.SendToFirstNode(pipeline.CommandMessage(&pipeline.Command{
-		Tp:        pipeline.CommandTypeStopAtTs,
-		StoppedTs: targetTs,
+		Tp: pipeline.CommandTypeStopAtTs,
 	}))
 	log.Info("send async stop signal to table", zap.Int64("tableID", t.tableID), zap.Uint64("targetTs", targetTs))
 	if err != nil {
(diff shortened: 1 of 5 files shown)


cdc/processor/processor.go
```diff
@@ -367,7 +367,7 @@ func (p *processor) handleTableOperation(ctx cdcContext.Context) error {
 				})
 			case model.OperProcessed:
 				if table.Status() != tablepipeline.TableStatusStopped {
-					log.Debug("the table is still not stopped", zap.Uint64("checkpointTs", table.CheckpointTs()), zap.Int64("tableID", tableID))
+					log.Info("the table is still not stopped", zap.Uint64("checkpointTs", table.CheckpointTs()), zap.Int64("tableID", tableID))
```
overvenus, line 370: Could you also change Debug to Info for the log at line 734?

https://github.com/pingcap/ticdc/blob/4f7c0b9606bc2c97f793caac42970aecc8f0a6c1/cdc/processor/processor.go#L734

cdc/processor/pipeline/table.go
```diff
@@ -100,8 +100,7 @@ func (t *tablePipelineImpl) UpdateBarrierTs(ts model.Ts) {
 // AsyncStop tells the pipeline to stop, and returns true is the pipeline is already stopped.
 func (t *tablePipelineImpl) AsyncStop(targetTs model.Ts) bool {
 	err := t.p.SendToFirstNode(pipeline.CommandMessage(&pipeline.Command{
-		Tp:        pipeline.CommandTypeStopAtTs,
-		StoppedTs: targetTs,
+		Tp: pipeline.CommandTypeStopAtTs,
```
overvenus, line 103:
```suggestion
		Tp: pipeline.CommandTypeStopAtTs,
```

Cloud you rename `pipeline.CommandTypeStopAtTs` to `pipeline.CommandTypeStop` ?

overvenus (approved):
Rest LGTM

liuzix (approved):

cdc/processor/processor.go
amyangfei: I'm afraid there are too many logs if table count is large, such as 10k tables.

amyangfei: I also revert this log level change, since
1. we can deduce lazy stop from sink barrier block log
2. It is better to refine scheduling logs in a specific PR

overvenus (approved):

PR #2773, opened 2021-09-08
[DNM]: Add log for debug mysql sink

3AceShowHand:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/cdc/sink/mysql.go
+++ b/cdc/sink/mysql.go
@@ -904,7 +904,7 @@ func (s *mysqlSink) execDMLWithMaxRetries(
 				}
 				for i, query := range dmls.sqls {
 					args := dmls.values[i]
-					log.Debug("exec row", zap.String("sql", query), zap.Any("args", args), zap.Int("bucket", bucket))
+					log.Debug("exec row", zap.String("sql", query), zap.Any("args", args))
 					if _, err := tx.ExecContext(ctx, query, args...); err != nil {
 						if rbErr := tx.Rollback(); rbErr != nil {
 							log.Warn("failed to rollback txn", zap.Error(err))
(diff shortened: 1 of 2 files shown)


cdc/sink/mysql.go
```diff
@@ -915,6 +917,9 @@ func (s *mysqlSink) execDMLWithMaxRetries(
 					}
 				}
 				if err = tx.Commit(); err != nil {
+					if strings.Contains(err.Error(), "Error 9007: Write conflict") {
+						log.Info("write conflict happened", zap.Strings("query", dmls.sqls), zap.Any("args", dmls.values), zap.Int("bucket", bucket))
```
amyangfei, line 921: We should pay attention to partition table id, but seems we can use row values to find data belongs to which partition.

3AceShowHand: At the moment, `RowChangedEvent` does not record Partition-ID. maybe we should support this field

3AceShowHand: For this case, the row belongs to which partition can be distinguished by row values.

3AceShowHand: For this case, the row belongs to which partition can be distinguished by row values.

cdc/sink/mysql.go
```diff
@@ -897,7 +899,7 @@ func (s *mysqlSink) execDMLWithMaxRetries(
 				}
 				for i, query := range dmls.sqls {
 					args := dmls.values[i]
-					log.Debug("exec row", zap.String("sql", query), zap.Any("args", args))
+					log.Info("exec row", zap.String("sql", query), zap.Any("args", args), zap.Int("bucket", bucket))
```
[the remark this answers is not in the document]
3AceShowHand: yes, it would be many logs, but it may give us some useful information.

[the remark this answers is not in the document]
3AceShowHand: I think we can convert this to `Debug` level, and when the problem occurred, change log-level to debug.

PR #2779, opened 2021-09-09
test: fix  TestTaskError and TestMultiError flaky tests

hi-rustin:
close https://github.com/pingcap/ticdc/issues/2469

In one case, it did happen that isCancelled was used to the next event before it was updated, and ErrCh immediately received an error, just as cancel was called during the execution of AddEvent, which is why I encountered this unstable test. I was able to reproduce the problem occasionally after increasing the number of events and limiting resources locally.

### Check List 

Tests 

 - Unit test
 

Code changes

None

Side effects

None

Related changes

### Release note 

```release-note
None
```


overvenus (approved):
LGTM, could you check if we have handle `context canceled` error properly when we call `AddEvent` in other files?

hi-rustin:
> LGTM, could you check if we have handle `context canceled` error properly when we call `AddEvent` in other files?

@overvenus It looks like calls elsewhere will return this error directly.

hi-rustin:
@amyangfei @ben1009  I think it ready for review.
Could you please take a look? Thanks!

hi-rustin:
I have tested this many times by limiting resources to 1C locally and it works.

pkg/workerpool/pool_test.go
```diff
@@ -55,7 +54,10 @@ func (s *workerPoolSuite) TestTaskError(c *check.C) {
 		c.Assert(err, check.ErrorMatches, "test error")
 	})
 
+	var wg sync.WaitGroup
+	wg.Add(1)
 	errg.Go(func() error {
```
amyangfei, line 59: `errg.Go` is not necessary for this goroutine

pkg/workerpool/pool_test.go
```diff
@@ -130,7 +134,10 @@ func (s *workerPoolSuite) TestMultiError(c *check.C) {
 		return nil
 	})
 
+	var wg sync.WaitGroup
+	wg.Add(1)
 	errg.Go(func() error {
```
amyangfei, line 139: ditto

amyangfei (approved):

pkg/workerpool/pool_test.go
hi-rustin: Thanks for your review!  💚 💙 💜 💛 ❤️

hi-rustin:
/assign @amyangfei 

Could you please help me trigger the merge process? Thanks!

amyangfei:
needs to be cherry picked to release branches

PR #2809, opened 2021-09-14
redolog : redolog storage layer data model & common util

ben1009:
### What problem does this PR solve? 
https://github.com/pingcap/ticdc/issues/2351  redo log based replication to implement eventually consistency
https://github.com/pingcap/ticdc/pull/2685 contains all of the change for the requirement, will split into small ones for review, this is one of the prs, split from the https://github.com/pingcap/ticdc/pull/2685

### What is changed and how it works?
this one contains basic data model, common util used in the fellowing PRs 

### Check List 

Tests 

 - Unit test

Code changes

 - Has exported function/method change
 - Has exported variable/fields change

Side effects

Related changes

### Release note 

```release-note
None
```

PR #2685, opened 2021-08-31
[WIP]Redo log: just used for integration test only

ben1009:
### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to update key monitor metrics in both TiCDC document and official document

### Release note 

```release-note
Please add a release note.
If you don't think this PR needs a release note then fill it with `None`.
```

PR #2811, opened 2021-09-14
redolog : redolog storage layer reader implementation

ben1009:
#2351 redo log based replication to implement eventually consistency
#2685 contains all of the change for the requirement, will split into small ones for review, this is one of the prs, split from the #2685, this one contains #2809,https://github.com/pingcap/ticdc/pull/2810, review #2809,https://github.com/pingcap/ticdc/pull/2810 first

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

Related changes

### Release note 

```release-note
None
```


amyangfei:
Could you please add docs about the relationship of files with different extensions(.tmp, .log, .meta, .mtmp, .sort etc.) and how they change from one extension to another.

cdc/redo/reader/reader.go
```diff
@@ -0,0 +1,439 @@
...
+			}
+
+			if !haveMeta {
+				haveMeta = true
+			}
+			metaList[l.meta.CheckPointTs] = l.meta
+			// since checkPointTs is guaranteed to increase always
+			if l.meta.CheckPointTs > maxCheckPointTs {
+				maxCheckPointTs = l.meta.CheckPointTs
+			}
```
amyangfei, line 361: Does this mean there could exist multiple meta files?

cdc/redo/reader/reader.go
```diff
@@ -0,0 +1,439 @@
...
+		}
+	}
+	if !haveMeta {
+		return 0, 0, errors.Errorf("no redo meta file found in dir:%s", l.cfg.Dir)
```
amyangfei, line 365: Better to create an RFC error

cdc/redo/reader/reader.go
ben1009: yes, if fail over happened between cdc clusters, could end up with multiple meta files

PR #2810, opened 2021-09-14
redolog : redolog storage layer writer implementation

ben1009:
### What problem does this PR solve? 
#2351 redo log based replication to implement eventually consistency
#2685 contains all of the change for the requirement, will split into small ones for review, this is one of the prs, split from the #2685, this one contains https://github.com/pingcap/ticdc/pull/2809, review https://github.com/pingcap/ticdc/pull/2809 first

### Check List 

Tests 

 - Unit test

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change

Side effects

Related changes

### Release note 

```release-note
None
```

PR #2812, opened 2021-09-14
redolog: add redo log manager and config

amyangfei:

part of #2351 redo log based replication to implement eventually consistency

#2685 contains all of the change for the requirement, will split into small ones for review, this is one of the prs, split from the #2685, this one contains #2809,#2810, review #2809,#2810,#2811 first. Or review commit https://github.com/amyangfei/ticdc/commit/6a5fe6bd52f68bf39c35cc699c7d54152b3e0e96 directly

NOTE: This PR is separated

- Add redo log manager, maintaining resolved ts of redo log writer of each table.
   - redo log manager interacts with upper caller and redo log writer.
   - In current design, the upper caller is table sink, table sink can emit DMLs and DDLs to redo log manager, redo log manager converts these TiCDC internal events to `redo logs` and pass the redo logs to underlying redo log writer.
      - Note the redo logs are kept in memory, the redo log manager doesn't implement back pressure currently, since the upper caller table sink/sink manager has a built in flow control mechanism.
   - The resolved ts of each table is maintained independently, in order we can implement table based replication in the future.
- Add `consistent` config in changefeed

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


cdc/redo/convert.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	pmodel "github.com/pingcap/parser/model"
+	"github.com/pingcap/ticdc/cdc/model"
+)
+
+// RowToRedo converts row changed event to redo log row
+func RowToRedo(row *model.RowChangedEvent) *model.RedoRowChangedEvent {
+	redoLog := &model.RedoRowChangedEvent{
+		Row:        row,
+		Columns:    make([]*model.RedoColumn, 0, len(row.PreColumns)),
```
ben1009, line 25: should be row.Columns instead ?

cdc/redo/manager_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"testing"
+	"time"
+
+	"github.com/pingcap/check"
+	"github.com/pingcap/ticdc/cdc/model"
+	"github.com/pingcap/ticdc/pkg/config"
+	"github.com/pingcap/ticdc/pkg/util/testleak"
+)
+
+func TestSuite(t *testing.T) { check.TestingT(t) }
```
ben1009, line 27: switch to testify if possible ?

amyangfei: changed

cdc/redo/convert.go
amyangfei: addressed

Issue #2826, opened 2021-09-15
migrate test-infra to testify for `cdc/entry` pkg


maxshuang:

tisonkun:
@maxshuang close by #2833?

maxshuang:
> @maxshuang close by #2833?

Sure

PR #2833, opened 2021-09-15
test(entry): migrate test-infra to testify for cdc/entry pkg

maxshuang:
migrate test-infra to testify for cdc/entry pkg   #2826 

1. migrate gocheck to testify,  migrate testleak to goleak
2. change tidb dependency to commit 0854595a6992 for new testkit 
3. move tidb/util/testkit pkg to tidb/testkit

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
migrate test-infra to testify for cdc/entry pkg
```


cdc/entry/main_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+package entry
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+func TestMain(m *testing.M) {
```
overvenus, line 22: I see `TestMain` in other PRs, can it be extracted into a function and reused?

overvenus (approved):
LGTM

cdc/entry/main_test.go
maxshuang: It is the package level function.  Before all other test-case in any *_test.go,  TestMain will be called firstly to do some initialization work.   

I don't think it can be wrapped.

ben1009 (approved):

cdc/entry/main_test.go
overvenus: Can we extract following code?

```go
package somepkg

func SetupGoLeak(m *testing.M) {
	opts := []goleak.Option{
		goleak.IgnoreTopFunction("go.etcd.io/etcd/pkg/logutil.(*MergeLogger).outputLoop"),
		goleak.IgnoreTopFunction("go.opencensus.io/stats/view.(*worker).start"),
	}

	goleak.VerifyTestMain(m, opts...)
}

// -------

package otherpkg

import somepkg

func TestMain(m *testing.M) {
    somepkg.SetupGoLeak(m)
}
```

maxshuang: But i am not sure  if SetupGoLeak  is so common that need to be wrapped. Some pkgs will add other stack functions.

I will check later.

PR #2877, opened 2021-09-23
kv: add cdc cluster id into grpc metadata

hi-rustin:
part of https://github.com/tikv/tikv/issues/10182

add cdc cluster id into grpc metadata

### Check List 

Tests 

 - Integration test

Code changes

None

Side effects

None

Related changes

None

### Release note 

```release-note
None
```


hi-rustin:
/hold

Needs test.

lonng:
Is there compatibility issue? The older version will not carry the cluster id metadata.

hi-rustin:
> Is there compatibility issue? The older version will not carry the cluster id metadata.

This should be a bug fix, I think we may need to bring it to older versions? cc: @overvenus

hi-rustin:
/close 
by https://github.com/tikv/tikv/pull/10907#discussion_r716166816

overvenus:
> This should be a bug fix, I think we may need to bring it to older versions? cc: @overvenus

Let's fix it only in future release. It's a minor bug as it rarely happens.

PR #2881, opened 2021-09-23
test(pkg/leakutil): extract leak helper func to a new pkg

maxshuang:
Extract leak helper func to a new pkg to avoid the circle deps in pkg/util

Extract leak helper func to a new pkg, avoid adding other unnecessary deps.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Add leakutil.SetUpLeakTest() to ignore some common unexpected stack top functions

### Release note 

```release-note
Extract leak helper func to a new pkg to avoid the circle deps in pkg/util
```

### DIFF
--- a/pkg/leakutil/leak_helper.go
+++ b/pkg/leakutil/leak_helper.go
@@ -19,7 +19,8 @@ import (
 	"go.uber.org/goleak"
 )
 
-//ignore common etcd and opencensus stack function for goleak
+//SetUpLeakTest ignore unexpected common etcd and opencensus stack functions for goleak
+//options can be used to implement other ignore items
 func SetUpLeakTest(m *testing.M, options ...goleak.Option) {
 	opts := []goleak.Option{
 		goleak.IgnoreTopFunction("go.etcd.io/etcd/pkg/logutil.(*MergeLogger).outputLoop"),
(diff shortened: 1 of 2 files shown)


overvenus (approved):

ben1009 (approved):

pkg/leakutil/leak_helper.go
```diff
@@ -0,0 +1,31 @@
...
+
+package leakutil
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+func SetUpLeakTest(m *testing.M, options ...goleak.Option) {
```
amyangfei, line 22: I'm wondering why ci lint doesn't report `exported function SetUpLeakTest should have comment or be unexported`

maxshuang: add comment for exported function SetUpLeakTest ---->done

amyangfei: The comment style is `SetUpLeakTest ...`, should be started with function name
ref: https://golang.org/doc/effective_go#commentary
In old days the lint could check this, seems it does not work now.

amyangfei: Seems the behavior changed after https://github.com/pingcap/ticdc/pull/2756 @hi-rustin would you please take a look?

PR #2939, opened 2021-09-26
sink: add a method to validate sink

3AceShowHand:

When walking the sink code, I found some duplicate code for validating the health of the sink, this PR try to make it dry.

add a new method to the sink package.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`.
```


amyangfei (approved):

ben1009 (approved):

cdc/sink/sink.go
```diff
@@ -118,3 +118,29 @@ func NewSink(ctx context.Context, changefeedID model.ChangeFeedID, sinkURIStr st
 	}
 	return nil, cerror.ErrSinkURIInvalid.GenWithStack("the sink scheme (%s) is not supported", sinkURI.Scheme)
 }
+
+// Validate the health of the sink
```
asddongmen, line 122: In fact, what we are checking is whether sinkuri and config are valid.
So this comment may be better?  
```
Validate whether the sinkuri and config are valid.
```

3AceShowHand:
@amyangfei I think we should also cherry-pick this to other branches, what do you think?

PR #2946, opened 2021-09-27
Turn on new replicate model on release-5.0

asddongmen:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```


pkg/config/config.go
```diff
@@ -32,7 +32,7 @@ import (
 const (
 	// NewReplicaImpl is true if we using new processor
 	// new owner should be also switched on after it implemented
-	NewReplicaImpl = false
+	NewReplicaImpl = true
```
ben1009, line 35: we keep the flag mean it is possible will switch back to old one ? if not could delete the flag, make it as the default behaviour ?

asddongmen: i wil clean all related codes in another pr after this pr merge.

amyangfei:
Please add release note

overvenus (approved):

PR #2955, opened 2021-09-28
mq_sink: report metadata when kafka client fails

liuzix:
- When Kafka client fails, it does not report what message has caused the error, making it difficult for the user to debug the situation.

- The `codec` can now pass an optional `Metadata` field to the `Producer`, which in turn will print the `Metadata` on errors.

### Check List 

Tests 

 - Unit test (WIP)
 - Integration test (WIP)

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Report details on the message causing the error when TiCDC fails to write message to Kafka.
```


cdc/sink/producer/kafka/kafka.go
```diff
@@ -130,6 +132,7 @@ func (k *kafkaSaramaProducer) SyncBroadcastMessage(ctx context.Context, message
 			Key:       sarama.ByteEncoder(message.Key),
 			Value:     sarama.ByteEncoder(message.Value),
 			Partition: int32(i),
+			Metadata:  message.Metadata,
```
amyangfei, line 135: Is it necessary to record partition number in metadata

cdc/sink/codec/json.go
```diff
@@ -451,6 +457,14 @@ func (d *JSONEventBatchEncoder) AppendRowChangedEvent(e *model.RowChangedEvent)
 		message.Schema = &e.Table.Schema
 		message.Table = &e.Table.Table
 
+		if message.Metadata == nil {
+			message.Metadata = &messageMeta{}
+		}
+
+		messageMeta := message.Metadata.(*messageMeta)
+		messageMeta.tables = append(messageMeta.tables, fmt.Sprintf("`%s`.`%s`", e.Table.Schema, e.Table.Table))
```
amyangfei, line 465: what about using a map, since table names could be duplicated

cdc/sink/producer/kafka/kafka.go
liuzix: Yes.

cdc/sink/codec/json.go
liuzix: I think it would be better if the user knows which table each record comes from and how long each record is. So we need to store an ordered list instead of a set of table names,.

PR #2977, opened 2021-10-04
server: add go gc metrics

overvenus:

Add TiCDC go gc metric.

| Before | After |
|--|--|
| image | image |

### Check List 

Run TiCDC

Related changes

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/metrics_server.go
+++ b/cdc/metrics_server.go
@@ -14,19 +14,42 @@
 package cdc
 
 import (
+	"os"
+	"strconv"
+
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-var etcdHealthCheckDuration = prometheus.NewHistogramVec(
-	prometheus.HistogramOpts{
-		Namespace: "ticdc",
-		Subsystem: "server",
-		Name:      "etcd_health_check_duration",
-		Help:      "Bucketed histogram of processing time (s) of flushing events in processor",
-		Buckets:   prometheus.ExponentialBuckets(0.0001 /* 0.1ms */, 2, 18),
-	}, []string{"capture", "pd"})
+var (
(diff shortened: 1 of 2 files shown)


cdc/metrics_server.go
```diff
@@ -14,19 +14,42 @@
...
+	goGC = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: "ticdc",
+			Subsystem: "server",
+			Name:      "gogc",
+			Help:      "The value of GOGC",
+		})
+)
+
+func init() {
```
amyangfei, line 42: Find a similar `init` function in tidb util package, I prefer to check `GOGC` env and set it with an explicit function.
Besides seems go runtime has processed `GOGC` env variable

https://github.com/golang/go/blob/170a72e58bd128b421f4b3974fe2a37fd035efdf/src/runtime/mgc.go#L190-L208

amyangfei (approved):

cdc/metrics_server.go
```diff
@@ -14,19 +14,55 @@
...
+	goGC.Set(float64(gogcValue))
+
+	maxProcs := runtime.GOMAXPROCS(0)
+	goMaxProcs.Set(float64(maxProcs))
+}
 
 // initServerMetrics registers all metrics used in processor
 func initServerMetrics(registry *prometheus.Registry) {
 	registry.MustRegister(etcdHealthCheckDuration)
+	registry.MustRegister(goGC)
```
CharlesCheung96, line 67: Should `goMaxProcs` be registered here?

CharlesCheung96 (approved):

asddongmen (approved):

PR #3002, opened 2021-10-09
sink: fix kafka max message size inaccurate issue.

### LINKED ISSUE #2962: mq_sink: unexpected behavior of `max-message-bytes`
### What did you do?

Assume the Kafka broker has 1M as `message.max.bytes`. I use TiCDC with Kafka sink and Open Protocol, where I set `max-message-bytes=10K`. 

If there is no row whose serialized size is more than 1M then everything works normally.  If there is, I would need a detailed error message helping me identify the problematic row (and table).

The expected behavior is that: as long as CDC has `max-message-bytes <= 1M` and there is no row larger than `1M`, everything should work.

### What did you see instead?

Any message larger than 10K fails to be delivered.

The root cause is that (1) `max-message-bytes` is used both as an advisory to the Open Protocol encoder AND as a hard-limit for the kafka producer (sarama) (2) we have failed to take into account the producer overhead by the kafka client.

### Problematic Scenarios

a.

|<------ CDC's limit = Broker's Limit = 10M ----->|<-**overflow**->| 
| row A = 5M | row B = 4M | row C = 1000K| overhead = 100K|

The batch should have been reduced, but was not.

b. 
|<--------------------------- Broker's Limit = 10M ------------------------>|
|<---- CDC's Limit = 5M --->|<--**overflow**-->|
|<------------- row A = 6M ----------------------->|<--still space here->|

The message should have been sent, but was not.

### Versions of the cluster

v5.0.1

3AceShowHand:

try to close #2962 

* When calculating the length of encoded MQMessage, also consider the `maximumRecordOverhead`, which was introduced by the sarama client.
* Ignoring `max-message-size` at the producer level, default to `512m`. `max-message-size` only works for the encoder.
* Change log level to `Debug` when `AppendRowChangedEvent` in json encoder.

### Check List 

Tests 

 - Unit test

Code changes
Side effects

Related changes

### Release note 

```release-note
fix kafka sink can not send message due to constraint by `max-message-size` option.
```

### DIFF
--- a/cdc/owner/ddl_puller.go
+++ b/cdc/owner/ddl_puller.go
@@ -77,21 +77,19 @@ func newDDLPuller(ctx cdcContext.Context, startTs uint64) (DDLPuller, error) {
 	}, nil
 }
 
-const ddlPullerName = "DDL_PULLER"
-
 func (h *ddlPullerImpl) Run(ctx cdcContext.Context) error {
 	ctx, cancel := cdcContext.WithCancel(ctx)
 	h.cancel = cancel
 	log.Debug("DDL puller started", zap.String("changefeed-id", ctx.ChangefeedVars().ID))
-	stdCtx := util.PutTableInfoInCtx(ctx, -1, ddlPullerName)
+	stdCtx := util.PutTableInfoInCtx(ctx, -1, puller.DDLPullerTableName)
+	stdCtx = util.PutChangefeedIDInCtx(stdCtx, ctx.ChangefeedVars().ID)
 	errg, stdCtx := errgroup.WithContext(stdCtx)
-	ctx = cdcContext.WithStd(ctx, stdCtx)
 
 	errg.Go(func() error {
-		return h.puller.Run(ctx)
+		return h.puller.Run(stdCtx)
 	})
 
-	rawDDLCh := puller.SortOutput(ctx, h.puller.Output())
+	rawDDLCh := puller.SortOutput(stdCtx, h.puller.Output())
 
 	receiveDDL := func(rawDDL *model.RawKVEntry) error {
 		if rawDDL == nil {
@@ -130,8 +128,8 @@ func (h *ddlPullerImpl) Run(ctx cdcContext.Context) error {
 	errg.Go(func() error {
 		for {
 			select {
-			case <-ctx.Done():
-				return ctx.Err()
+			case <-stdCtx.Done():
+				return stdCtx.Err()
 			case e := <-rawDDLCh:
 				if err := receiveDDL(e); err != nil {
 					return errors.Trace(err)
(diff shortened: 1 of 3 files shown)


cdc/sink/codec/interface.go
```diff
@@ -63,9 +63,13 @@ type MQMessage struct {
 	Protocol Protocol            // protocol
 }
 
+// producerMessageOverhead is used to calculate ProducerMessage's byteSize by sarama kafka client.
+// reference: https://github.com/Shopify/sarama/blob/main/async_producer.go#L233
```
overvenus, line 67:
```suggestion
// reference: https://github.com/Shopify/sarama/blob/66521126c71c522c15a36663ae9cddc2b024c799/async_producer.go#L231
```

cdc/sink/codec/interface.go
```diff
@@ -63,9 +63,13 @@ type MQMessage struct {
 	Protocol Protocol            // protocol
 }
 
+// producerMessageOverhead is used to calculate ProducerMessage's byteSize by sarama kafka client.
+// reference: https://github.com/Shopify/sarama/blob/66521126c71c522c15a36663ae9cddc2b024c799/async_producer.go#L231
```
lance6716, line 67: can we add a unit test to ensure this value of sarama won't be changed?

cdc/sink/codec/interface.go
```diff
@@ -63,9 +63,13 @@ type MQMessage struct {
...
+const producerMessageOverhead = 26 // the metadata overhead of CRC, flags, etc.
```
amyangfei, line 68: Is this constant related to Kafka version? when `version >= 2`, maximumRecordOverhead is 36
https://github.com/Shopify/sarama/blob/66521126c71c522c15a36663ae9cddc2b024c799/async_producer.go#L235-L242

3AceShowHand: yes, we use `2.4.0` as default, is it possible to be `version < 2`?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -64,6 +67,101 @@ func NewKafkaConfig() Config {
...
+
+	s = sinkURI.Query().Get("max-message-bytes")
+	if s != "" {
+		a, err := strconv.Atoi(s)
+		if err != nil {
+			return err
+		}
+		if a > c.MaxMessageBytes {
+			c.MaxMessageBytes = a
+		}
```
amyangfei, line 102: Please add some comments here

cdc/sink/codec/interface.go
3AceShowHand: it must be version `2`.

cdc/sink/codec/interface.go
```diff
@@ -63,9 +64,14 @@ type MQMessage struct {
 	Protocol Protocol            // protocol
 }
 
+// maximumRecordOverhead is used to calculate ProducerMessage's byteSize by sarama kafka client.
+// reference: https://github.com/Shopify/sarama/blob/66521126c71c522c15a36663ae9cddc2b024c799/async_producer.go#L233
+// for TiCDC, minimum supported kafka version is `0.11.0.2`, which will be treated as `version = 2` by sarama producer.
+const maximumRecordOverhead = 5*binary.MaxVarintLen32 + binary.MaxVarintLen64 + 1
+
```
liuzix, line 71: Can you comment on why the Headers do not matter? https://github.com/Shopify/sarama/blob/66521126c71c522c15a36663ae9cddc2b024c799/async_producer.go#L237

PR #3003, opened 2021-10-09
kafka_producer: fix deadlock when an error occurs in asyncClient

liuzix:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```


cdc/sink/producer/kafka/kafka.go
```diff
@@ -197,26 +213,27 @@ func (k *kafkaSaramaProducer) GetPartitionNum() int32 {
 }
 
 // stop closes the closeCh to signal other routines to exit
+// It SHOULD NOT be called under `clientLock`.
 func (k *kafkaSaramaProducer) stop() {
-	k.clientLock.Lock()
-	defer k.clientLock.Unlock()
-	select {
-	case <-k.closeCh:
+	if atomic.SwapInt32(&k.closing, kafkaProducerClosing) == kafkaProducerClosing {
```
asddongmen, line 218: The logic here has a double meaning, will it be easier to understand if you separate them?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -116,8 +133,7 @@ func (k *kafkaSaramaProducer) SendMessage(ctx context.Context, message *codec.MQ
 		return ctx.Err()
 	case <-k.closeCh:
 		return nil
-	default:
-		k.asyncClient.Input() <- msg
+	case k.asyncClient.Input() <- msg:
```
amyangfei, line 136: `SyncBroadcastMessage` has a similar default branch, please check whether we need to update in `SyncBroadcastMessage`

cdc/sink/producer/kafka/kafka.go
asddongmen: This logic cannot be divided into two steps, I am wrong.

amyangfei (approved):

PR #3022, opened 2021-10-12
config: remove per-table-memory-quota min value check

overvenus:

Remove per-table-memory-quota min value check, because it is
possible to capture thousands of tables, and in this case, 6MB is
too large.

Also, it changes the default quota from 20MB to 10MB. Tests show
that no obvious bottleneck for 10MB.

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-0713f176-7fff-9d9e-d26f-c9e5e829a698"><div dir="ltr" style="margin-left:0pt;" align="left">

Quota = 1MB | Quota = 10MB | Quota = 20MB
-- | -- | --
 
 
4.5s vs 49s vs 58s
41k vs 361k vs 474k
 

</div></b>

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```


pkg/config/config_test.go
```diff
@@ -117,5 +117,5 @@ func TestServerConfigValidateAndAdjust(t *testing.T) {
 	require.Regexp(t, ".*does not contain a port", conf.ValidateAndAdjust())
 	conf.AdvertiseAddr = "advertise:1234"
 	conf.PerTableMemoryQuota = 1
```
amyangfei, line 119: Can add a test, when `conf. PerTableMemoryQuota` is zero, `ValidateAndAdjust` will reset it to default quota

Issue #2867, opened 2021-09-22
TiCDC does not resolve lock in time


overvenus:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

* What did you do? If possible, provide a recipe for reproducing the error.

1. Create table A and B. Create a changefeed that captures A.
2. Kill TiDB instances and leave some locks in B. 
3. Wait 10 minutes.
4. Merge regions a (belongs table A) and region b (belongs table B).

* What did you expect to see?

Changefeed is not blocked and locks in B can be resolved by TiCDC.

* What did you see instead?

Changefeed is blocked. And lots of fallen back logs.

```
[2021/09/17 09:57:25.081 +08:00] [WARN] [region_worker.go:690] ["The resolvedTs is fallen back in kvclient"] ["Event Type"=RESOLVED] [resolvedTs=427759786830332014] [lastResolvedTs=427767830928687109] [regionID=478066658]
```

Also, we see TiKV resolved ts jump, because of region merge.

![image](https://user-images.githubusercontent.com/2150711/134284397-36e9f62c-cbf9-4b4b-9328-95adb307237b.png)

* Versions of the cluster

  v5.1.1

amyangfei:
Whether there exists `resolve lock` log for `regionID=478066658`? @overvenus

amyangfei:
Besides, whether `kv client reconnect triggered` happens?

PR #3062, opened 2021-10-14
fix(kvclient): fix region merging would block the progress of resolveLock

### LINKED ISSUE #3061: fallback resolvedTs event skips the  lock resolving unexpectly
This bug is caused by following reason:
(1) when region merge occur,  a new stream with new region_worker will be created.  Then region_worker.rtsManager will be empty;
(2) if merged region has some fallback prewrite lock in it, the resolvedTs event will fall back last checkpointTs and will be ignored. Then resolve lock logic can't trigger for  region_worker.rtsManager has no record for that region.

maxshuang:
fix the bug that region merging would block the progress of resolveLock when tidb tx end abnormally and remain some prewrite lock in region. Close  #3061 #2867 

(1) When meet a fallback resolvedTs event, we send it to resolveLock thread instead of ignoring it  so that resolve lock logic can trigger;
(2) Fallback resolvedTs event will also increase the penalty which will trigger resolve lock eventually.

### Check List 

Tests 

 - Unit test
 - Integration test
 Ref to the step in issue #2867 

Related changes

### Release note 

```release-note
fix the bug that fallback resolvedTs event  will block the progress of resolve lock when occur region merging
  ```

### DIFF
--- a/.github/workflows/check_and_build.yaml
+++ b/.github/workflows/check_and_build.yaml
@@ -49,7 +49,10 @@ jobs:
       uses: actions/cache@v2
       with:
         path: tools/bin
-        key: ${{ runner.os }}-cdc-lint-tools-${{ hashFiles('tools/check/go.sum') }}
+        # See: https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows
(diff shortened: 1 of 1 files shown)


maxshuang:
@amyangfei  @overvenus  PTAL if available. Thanks.

PR #3071, opened 2021-10-15
workerpool: support GracefulUnregister for handles

liuzix:
- The peer messaging system needs a graceful unregister mechanism to avoid message losses after the messages are passed to `pool`.

- `GracefulUnregister(ctx context.Context, timeout time.Duration) error`, which will wait until all pending messages have done being processed.

### Check List 

Tests 

 - Unit test

Code changes

### Release note 

```release-note
None
```


asddongmen:
rest LGTM

PR #3076, opened 2021-10-18
sink: canal json support encode checkpoint event.

### LINKED ISSUE #3096: Canal-JSON Support encode checkpoint event.

Make canal-json encode support encode checkpoint event, so that downstream consumer can use it to reform the original transaction.

Make canal-json encode support encode checkpoint event

_No response_

_No response_

3AceShowHand:

close #3096, #3097

This PR is part of the work try to make Kafka Sink with Canal-JSON GA.

* Canal-JSON `CheckpointEvent` with the EventType `WaterMark`, and control by option `waterMark`, which default to false.
* Canal-JSON format will output `BuildTime`, Identical to Official implementation.
* Send DDL to partition 0 for Canal-JSON / Canal-PB
* Add some comments about Canal-JSON fields.

Future work: add a Canal-JSON Decoder example.

### Different between TiCDC Canal-JSON and Official Canal-JSON

* For official Canal-JSON, `ID` start from 1, and monotonically increase, messages generated by the same event, have the same ID. For TiCDC, it's always 0.  
  * https://github.com/alibaba/canal/blob/master/meta/src/main/java/com/alibaba/otter/canal/meta/CanalMetaManager.java#L65
  * https://github.com/alibaba/canal/blob/master/meta/src/main/java/com/alibaba/otter/canal/meta/MemoryMetaManager.java#L122

1. add a `WATERMARK` type in Canal-Proto
2. send DDL to partition 0 for Canal-JSON / Canal-PB
3. fill the `BuildTime` field with a millisecond timestamp.

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes
update documents about compatibility with this one:
* `Protocol` become a required parameter
* more detailed description about how to use Kafka sink with the Canal-JSON protocol.
* Update Document about how to handle `WATERMARK` EventType in `Canal-JSON` format.

### Release note 

```release-note
`None`.
```


proto/canal/EntryProtocol.pb.go
```diff
@@ -84,6 +84,8 @@ const (
 	EventType_XAROLLBACK EventType = 14
 	//* MASTER HEARTBEAT *
 	EventType_MHEARTBEAT EventType = 15
+	//* TiCDC CUSTOMIZED *
+	EventType_WATERMARK EventType = 16
```
amyangfei, line 88: What if `canal` add another filed = 16 in the future?
Besides is there any side effect to add one more field, such as will it break the usage of existing consumer.

cdc/sink/producer/kafka/kafka.go
```diff
@@ -116,9 +119,10 @@ func (c *Config) Initialize(sinkURI *url.URL, replicaConfig *config.ReplicaConfi
 	c.ClientID = sinkURI.Query().Get("kafka-client-id")
 
 	s = sinkURI.Query().Get("protocol")
-	if s != "" {
-		replicaConfig.Sink.Protocol = s
+	if s == "" {
+		return cerror.ErrKafkaInvalidConfig.GenWithStack("encoding protocol is not set")
 	}
```
amyangfei, line 124: Is this a breaking change, if TiCDC upgrades and an old changefeed whose protocol is not set in sink-uri, the changefeed can't be resumed without changing it sink-uri.

cdc/sink/codec/canal_flat.go
```diff
@@ -30,6 +32,8 @@ type CanalFlatEventBatchEncoder struct {
 	builder       *canalEntryBuilder
 	unresolvedBuf []*canalFlatMessage
 	resolvedBuf   []*canalFlatMessage
+	// output checkpoint event is true
```
liuzix, line 35:
```suggestion
	// When it is true, checkpoint events are generated.
```

liuzix: If it is read only, please add a note.

cdc/sink/codec/canal_flat.go
```diff
@@ -62,21 +67,29 @@ func newCanalFlatEventBatchEncoderBuilder(opts map[string]string) EncoderBuilder
...
-	BuildTime uint64            `json:"ts"`
-	Query     string            `json:"sql"`
-	SQLType   map[string]int32  `json:"sqlType"`
+	ID        int64    `json:"id"`
+	Schema    string   `json:"database"`
+	Table     string   `json:"table"`
+	PKNames   []string `json:"pkNames"`
+	IsDDL     bool     `json:"isDdl"`
+	EventType string   `json:"type"`
+	// officially the timestamp of the event which generate the message executed, in milliseconds
```
liuzix, line 76:
```suggestion
	// officially the timestamp of the event-time of the message, in milliseconds since Epoch.
```

cdc/sink/codec/canal_flat.go
```diff
@@ -62,21 +67,29 @@ func newCanalFlatEventBatchEncoderBuilder(opts map[string]string) EncoderBuilder
...
+	ExecutionTime int64 `json:"es"`
+	// officially the timestamp of building the MQ message, in milliseconds
```
liuzix, line 78:
```suggestion
	// officially the timestamp of building the MQ message, in milliseconds in Epoch
```

cdc/sink/codec/canal_flat.go
```diff
@@ -62,21 +67,29 @@ func newCanalFlatEventBatchEncoderBuilder(opts map[string]string) EncoderBuilder
...
+	BuildTime int64 `json:"ts"`
+	// SQL that generate the change event, DDL or Query
```
liuzix, line 80:
```suggestion
	// SQL that generated the change event, DDL or Query
```
Is this used only for DDLs?

cdc/sink/codec/canal_flat.go
```diff
@@ -62,21 +67,29 @@ func newCanalFlatEventBatchEncoderBuilder(opts map[string]string) EncoderBuilder
...
+	Query string `json:"sql"`
+	// only works for INSERT / UPDATE / DELETE event, record each column's java representation type.
```
liuzix, line 82:
```suggestion
	// only works for INSERT / UPDATE / DELETE events, records each column's java representation type.
```

cdc/sink/codec/canal_flat.go
```diff
@@ -62,21 +67,29 @@ func newCanalFlatEventBatchEncoderBuilder(opts map[string]string) EncoderBuilder
...
+	SQLType map[string]int32 `json:"sqlType"`
+	// only works for INSERT / UPDATE / DELETE event, record each column's mysql representation type.
```
liuzix, line 84:
```suggestion
	// only works for INSERT / UPDATE / DELETE events, records each column's mysql representation type.
```

cdc/sink/codec/canal_flat.go
```diff
@@ -62,21 +67,29 @@ func newCanalFlatEventBatchEncoderBuilder(opts map[string]string) EncoderBuilder
...
 	MySQLType map[string]string `json:"mysqlType"`
 	// A Datum should be a string or nil
 	Data []map[string]interface{} `json:"data"`
 	Old  []map[string]interface{} `json:"old"`
+
+	// CheckpointTs is a TiCDC custom field, which does not support by original canal protocol
```
liuzix, line 90:
```suggestion
	// CheckpointTs is a TiCDC custom field, which is not supported by original Canal protocol
```

cdc/sink/codec/canal_flat.go
```diff
@@ -62,21 +67,29 @@ func newCanalFlatEventBatchEncoderBuilder(opts map[string]string) EncoderBuilder
...
+	// this is useful for message consumer to reform the original transactions.
```
liuzix, line 91:
```suggestion
	// It is useful if the message consumer needs to restore the original transactions.
```

cdc/sink/codec/canal_flat.go
```diff
@@ -148,7 +161,7 @@ func (c *CanalFlatEventBatchEncoder) newFlatMessageForDML(e *model.RowChangedEve
 		IsDDL:         false,
 		EventType:     header.GetEventType().String(),
 		ExecutionTime: header.ExecuteTime,
-		BuildTime:     0, // ignored by both Canal Adapter and Flink
+		BuildTime:     time.Now().UnixNano() / 1e6,
```
liuzix, line 164: Please add a comment here on the calculation.

cdc/sink/codec/canal_flat.go
```diff
@@ -163,9 +167,22 @@ func (c *CanalFlatEventBatchEncoder) newFlatMessageForDDL(e *model.DDLEvent) *ca
 	return ret
 }
 
-// EncodeCheckpointEvent is no-op
+func (c *CanalFlatEventBatchEncoder) newFlatMessage4CheckpointEvent(ts uint64) *canalFlatMessage {
```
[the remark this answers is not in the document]
liuzix: Could a consumer not recognizing this event complain?

cdc/sink/mq.go
```diff
@@ -230,9 +230,19 @@ func (k *mqSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
 		return nil
 	}
 
+	// -1 means broadcast to all partition, it's the default for the default open protocol.
+	var partition int32 = -1
```
liuzix, line 234: Could you add a constant for `-1`

cdc/sink/sink.go
```diff
@@ -44,7 +44,7 @@ type Sink interface {
 	EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error
 
 	// FlushRowChangedEvents flushes each row which of commitTs less than or equal to `resolvedTs` into downstream.
-	// TiCDC guarantees that all of Event which of commitTs less than or equal to `resolvedTs` are sent to Sink through `EmitRowChangedEvents`
+	// TiCDC guarantees that all the Event which of commitTs less than or equal to `resolvedTs` are sent to Sink through `EmitRowChangedEvents`
```
liuzix, line 47:
```suggestion
	// TiCDC guarantees that all the Events whose commitTs is less than or equal to `resolvedTs` are sent to Sink through `EmitRowChangedEvents`
```

cdc/sink/codec/canal_flat.go
3AceShowHand: from canal source code comment, it may be DDL / Query.

But we have a manually test for this, only for DDL.

cdc/sink/codec/canal_flat.go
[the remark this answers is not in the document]
3AceShowHand: We would have to update the document to describe how to handle the `WATERMARK` event.

I think consumers who are not expecting this, will not complain. Those who want this should be know how to enable this and handle the `WATERMARK` message by reading the documentation.

cdc/sink/producer/kafka/kafka.go
3AceShowHand: got it, we have to handle it in another way.

proto/canal/EntryProtocol.pb.go
3AceShowHand: Can you give me any hint about how to handle the situation mentioned above ?

Maybe we can define `WATERMARK` as a constant, without altering the protobuf related files?

For the existing consumers, it will not receive any message containing this event type, because we controlled by a parameter `waterMark` default to false. 

For any user that wants to use this event type, we should provide concrete documentation and consumer example, which will be done in the near future.

cdc/sink/codec/canal_flat.go
```diff
@@ -30,6 +32,8 @@ type CanalFlatEventBatchEncoder struct {
 	builder       *canalEntryBuilder
 	unresolvedBuf []*canalFlatMessage
 	resolvedBuf   []*canalFlatMessage
+	// When it is true, checkpoint events are generated.
+	waterMark bool
```
overvenus, line 36:
```suggestion
	watermark bool
```

Just _watermark_, the term is widely used.

proto/EntryProtocol.proto
```diff
@@ -217,30 +217,32 @@ enum EntryType{
...
+	CINDEX		= 		10;
+	DINDEX 		= 		11;
+	GTID        =       12;
+	/** XA **/
+	XACOMMIT    =       13;
+	XAROLLBACK  =		14;
+	/** MASTER HEARTBEAT **/
+	MHEARTBEAT  =       15;
+	/** TiCDC CUSTOMIZED **/
+	WATERMARK = 16;
```
overvenus, line 239: Please choose a larger number, leave some space for changes from upstream.

overvenus: Also, please add a prefix to prevent name collision in the future.

.github/workflows/ticdc_integration.yaml
```diff
@@ -205,3 +205,65 @@ jobs:
         if: ${{ always() }}
         run: |
           $GITHUB_WORKSPACE/scripts/canal/canal-local-test.sh down
+
+  CanalJson-with-watermark-integration:
```
liuzix, line 209: Is it necessary to make `CanalJson-with-watermark-integration` a new Action, in addition to the existing one? Will enabling watermarks on the existing Canal Action suffice?

cdc/sink/codec/canal_flat.go
```diff
@@ -59,20 +67,30 @@ func newCanalFlatEventBatchEncoderBuilder(opts map[string]string) EncoderBuilder
 	return &canalFlatEventBatchEncoderBuilder{opts: opts}
 }
 
+type canalFlatMessageInterface interface {
```
liuzix, line 70: Could you comment on the motivation to make it an interface?

cdc/sink/codec/canal_flat.go
```diff
@@ -28,16 +30,22 @@ import (
 // CanalFlatEventBatchEncoder encodes Canal flat messages in JSON format
 type CanalFlatEventBatchEncoder struct {
 	builder       *canalEntryBuilder
-	unresolvedBuf []*canalFlatMessage
-	resolvedBuf   []*canalFlatMessage
+	unresolvedBuf []canalFlatMessageInterface
+	resolvedBuf   []canalFlatMessageInterface
+	// When it is true, canal-json would generate TiDB extended information
```
liuzix, line 35:
```suggestion
	// When it is true, canal-json would generate TiDB extension information,
```

cdc/sink/codec/canal_flat.go
```diff
@@ -28,16 +30,22 @@ import (
...
+	// At the moment, only `tidbWaterMarkType` and `_tidb` fields.
```
liuzix, line 36:
```suggestion
	// which, at the moment, only includes `tidbWaterMarkType` and `_tidb` fields.
```

cdc/sink/codec/canal_flat.go
```diff
@@ -159,49 +214,85 @@ func (c *CanalFlatEventBatchEncoder) newFlatMessageForDML(e *model.RowChangedEve
...
+}
+
+func (c *CanalFlatEventBatchEncoder) newFlatMessage4CheckpointEvent(ts uint64) *canalFlatMessageWithTiDBExtension {
+	return &canalFlatMessageWithTiDBExtension{
+		canalFlatMessage: &canalFlatMessage{
+			ID:            0,
+			IsDDL:         false,
+			EventType:     tidbWaterMarkType,
+			ExecutionTime: convertToCanalTs(ts),
+			BuildTime:     time.Now().UnixNano() / 1e6,
```
liuzix, line 261:
```suggestion
			BuildTime:     time.Now().UnixNano() / 1e6, // converts to milliseconds
```

cdc/sink/codec/canal_flat.go
```diff
@@ -258,8 +349,13 @@ func (c *CanalFlatEventBatchEncoder) Reset() {
 	panic("not supported")
 }
 
-// SetParams is no-op for now
 func (c *CanalFlatEventBatchEncoder) SetParams(params map[string]string) error {
-	// no op
+	if s, ok := params["enable-tidb-extension"]; ok {
+		a, err := strconv.ParseBool(s)
+		if err != nil {
+			return err
```
liuzix, line 356: Maybe we can wrap this error so that an RFC code can be provided for the user to identify the problem?

.github/workflows/ticdc_integration.yaml
3AceShowHand: I will add another action for flink test, I would create a new workflow for extension test.

.github/workflows/ticdc_canal_json_extension.yaml
```diff
@@ -0,0 +1,83 @@
+name: CDC Canal-JSON-Extension Tests
```
hi-rustin, line 1: Can we integrate it into the original test, we no longer have the resources to support the new actions.

cdc/sink/codec/canal_flat.go
```diff
@@ -28,16 +30,22 @@ import (
...
-	unresolvedBuf []*canalFlatMessage
-	resolvedBuf   []*canalFlatMessage
+	unresolvedBuf []canalFlatMessageInterface
+	resolvedBuf   []canalFlatMessageInterface
+	// When it is true, canal-json would generate TiDB extension information
+	// which, at the moment, only includes `tidbWaterMarkType` and `_tidb` fields.
+	enableTiDBExtension bool
 }
 
+const tidbWaterMarkType = "TiDB_WATERMARK"
```
overvenus, line 40: Can it be "TIDB_WATERMARK"? Similar to "MYSQL" (for MySQL).

proto/EntryProtocol.proto
```diff
@@ -217,30 +217,30 @@ enum EntryType{
...
-    /**CREATE INDEX**/
-    CINDEX		= 		10;
-    DINDEX 		= 		11;
-    GTID        =       12;
-    /** XA **/
-    XACOMMIT    =       13;
-    XAROLLBACK  =		14;
-    /** MASTER HEARTBEAT **/
-    MHEARTBEAT  =       15;
+	INSERT 		= 		1;
```
overvenus, line 220: Could you revert changes to this file?

overvenus (commented):
Rest LGTM

cdc/sink/mq.go
```diff
@@ -42,7 +42,11 @@ type mqEvent struct {
 	resolvedTs uint64
 }
 
-const defaultPartitionInputChSize = 12800
+const (
+	defaultPartitionInputChSize = 12800
+	// -1 means broadcast to all partition, it's the default for the default open protocol.
+	defaultDDLDispatchPartition = -1
```
hi-rustin, line 48: We also use this value in EmitCheckpointTs, can you change it together?

hi-rustin (commented):
rest LGTM

cdc/sink/codec/canal_flat.go
```diff
@@ -159,49 +217,85 @@ func (c *CanalFlatEventBatchEncoder) newFlatMessageForDML(e *model.RowChangedEve
...
+}
+
+func (c *CanalFlatEventBatchEncoder) newFlatMessage4CheckpointEvent(ts uint64) *canalFlatMessageWithTiDBExtension {
+	return &canalFlatMessageWithTiDBExtension{
+		canalFlatMessage: &canalFlatMessage{
+			ID:            0,
+			IsDDL:         false,
+			EventType:     tidbWaterMarkType,
+			ExecutionTime: convertToCanalTs(ts),
+			BuildTime:     time.Now().UnixNano() / 1e6, // converts to milliseconds
```
amyangfei, line 264:
```suggestion
			BuildTime:     time.Now().UnixNano() / int64(time.Millisecond), // converts to milliseconds
```
can update too for other places using `1e6`.

cdc/sink/mq.go
```diff
@@ -42,7 +42,11 @@ type mqEvent struct {
 	resolvedTs uint64
 }
 
-const defaultPartitionInputChSize = 12800
+const (
+	defaultPartitionInputChSize = 12800
+	// -1 means broadcast to all partition, it's the default for the default open protocol.
```
amyangfei, line 47:
```suggestion
	// -1 means broadcast to all partitions, it's the default for the default open protocol.
```

cdc/sink/mq.go
3AceShowHand: this is only for ddl, I think we should have a further discussion about how to dispatch checkpointTs.

overvenus (approved):

amyangfei (approved):

cdc/sink/mq.go
hi-rustin: I understand that in the current implementation, we also broadcast the checkpoint. I mean let it represent the broadcast, it doesn't change the current implementation. It just treats it as a constant. For example it could be named: defaultBrodcastPartitionNum.

PR #3163, opened 2021-10-28
scheduler_v2: implement the basic logic for agent (the schedule receiver in Processor)

liuzix:
- This PR contains the basic logic of an "agent", which is designed to be part of the processor.
- This PR **does not** contain, nor does it require a specific messaging system.

- Implements the logic of receiving schedules from the owner.
- Implements the protocol for the processor to interact with the owner.

### Check List 

Tests 

 - Unit test

Code changes

### Release note 

```release-note
None
```


lonng:
@liuzix Please split this PR into some smalls.

lonng:
@liuzix Can you split the large PR into some smalls?

cdc/scheduler/agent.go
```diff
@@ -0,0 +1,448 @@
...
+		checkpointSender: newCheckpointSender(messenger, logger, config.SendCheckpointTsInterval),
+		ownerHasChanged:  atomic.NewBool(false),
+		config:           config,
+	}
+}
+
+type agentOperationStatus int32
+
+const (
+	operationReceived = agentOperationStatus(iota)
```
amyangfei, line 141: Better to start from `iota + 1`

cdc/scheduler/agent.go
```diff
@@ -0,0 +1,448 @@
...
+		// This is necessary because the communicator might be waiting for
+		// messages to be received by the previous owner.
+		a.communicator.OnOwnerChanged(ctx, a.currentOwner())
+	}
+
+	if err := a.sendCheckpoint(ctx); err != nil {
+		return errors.Trace(err)
+	}
+
+	if a.needSyncNow.Load() {
```
amyangfei, line 176: Can processor send checkpoint before a needed `sync`?

cdc/scheduler/agent.go
```diff
@@ -0,0 +1,448 @@
...
+	util.SorterTableIDs(adding)
+	util.SorterTableIDs(removing)
+	done, err := a.communicator.SyncTaskStatuses(ctx, running, adding, removing)
+	if err != nil {
+		return false, errors.Trace(err)
+	}
+	if !done {
+		return false, nil
+	}
+	return true, nil
```
amyangfei, line 252: Seems can use `return done, err` for short

cdc/scheduler/agent.go
```diff
@@ -0,0 +1,448 @@
...
+				if err != nil {
+					return errors.Trace(err)
+				}
+				if done {
+					op.status = operationProcessed
+				} else {
+					break
+				}
+			}
+			fallthrough
```
amyangfei, line 279: What about
- Remove L266
- Change L273-L277 to 
```go
if !done {
    break
}
```
- move `op.status = operationProcessed` before L279, then the state transition will be more clear.

cdc/scheduler/agent.go
```diff
@@ -0,0 +1,448 @@
...
+}
+
+// TableExecutor is an abstraction for "Processor".
+type TableExecutor interface {
+	AddTable(ctx context.Context, tableID model.TableID) error
+	RemoveTable(ctx context.Context, tableID model.TableID) (done bool, err error)
+	IsAddTableFinished(ctx context.Context, tableID model.TableID) (done bool)
+	IsRemoveTableFinished(ctx context.Context, tableID model.TableID) (done bool)
+
+	GetAllCurrentTables() []model.TableID
```
amyangfei, line 49: Better to add comments to this function, seems it returns running, adding, removing tables.

cdc/scheduler/agent.go
liuzix: The Protocol does not forbid this behavior. There are two scenarios where we need `sync`:
- Processor initialization: Since we forbids the processor from sending checkpoints when no table is running, we cannot send checkpoint in this situation.
- In response to owner's announce: The checkpoint sent after receiving an announce would be the expected correct checkpoint, and the owner will take it into consideration if the checkpoint is received after `sync`.

Although there is no obvious correctness problem, I will modify the code a little bit so there will be less confusion.

amyangfei (approved):

cdc/scheduler/agent.go
```diff
@@ -0,0 +1,460 @@
...
+				if !done {
+					break
+				}
+			}
+			op.status = operationProcessed
+			fallthrough
+		case operationProcessed:
+			var (
+				done bool
+				err  error
```
asddongmen, line 295: why we need a err here?

asddongmen (approved):

PR #3164, opened 2021-10-28
dm/relay: fix intermediate meta info written to meta file

D3Hunter:
fix a bug that may cause upstream switch fail, it happens when master/slave switches:
1. upstream switch, relay pause and resume, relay.process will reset meta, since it's a new server, we don't know binlog filename, it's set to `minCheckpoint`(filename is empty), create new uuid dir, write binlog file. flush meta with empty filename
2. reader found new uuid dir, read event from that dir and reach the end of last relay log file, and that event is also in an already processed transaction, replaceWithHeartbeat: true.
3. reader wait for new event and found meta(empty) is different with current binlog filename, reparse current relay log file, but lost the replaceWithHeartbeat variable, so already processed transaction will be sync again.
```
[2021/10/27 07:33:21.156 +00:00] [INFO] [file.go:247] ["newer relay log file is already generated"] ["now file path"=relay-dir/038f4cc2-36f8-11ec-839d-0242ac1c8007.000002/mysql-bin.000003] ["new file path"=relay-dir/038f4cc2-36f8-11ec-839d-0242ac1c8007.000002]
```
it an old bug that is hard to reproduce before, for when `binlog reader` waits for new events, it checks every 100ms, but now we check it immediately after notified of new events, so more chance to trigger this bug in `dm/tests/upstream_switch`

- save and flush meta when we receive first event and before write

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
dm/relay: fix intermediate meta info written to meta file
```


GMHDBJD (approved):
LGTM
By the way, I think we can add a log for switch new uuid dir in https://github.com/pingcap/ticdc/blob/0a8390ae8e3f6d590b4ed836f960a0332efc6391/dm/pkg/streamer/file.go#L274 as we do in https://github.com/pingcap/ticdc/blob/0a8390ae8e3f6d590b4ed836f960a0332efc6391/dm/pkg/streamer/file.go#L247 for debug.

dm/relay/relay.go
```diff
@@ -558,6 +559,17 @@ func (r *Relay) handleEvents(
 			}
 		}
 
+		if firstEvent {
```
Ehco1996, line 562: seems don't need `firstEvent` var, we can  move this logic to L549 it already check the event is `fakerotateEvent`  with `if _, ok := e.Event.(*replication.RotateEvent); ok && utils.IsFakeRotateEvent(e.Header)`

dm/relay/relay_test.go
```diff
@@ -453,14 +455,44 @@ func (t *testRelaySuite) TestHandleEvent(c *C) {
...
+	reader2.err = nil
+	reader2.result.Event = fakeRotateEv
+	// writer return error to force handleEvents return
+	writer2.err = errors.New("writer error for testing")
+	// return with the annotated writer error
+	_, err = r.handleEvents(ctx, reader2, transformer2, writer2)
+	c.Assert(errors.Cause(err), Equals, writer2.err)
+	// after handle rotate event, we save and flush the meta immediately
+	c.Assert(r.meta.Dirty(), Equals, false)
+	{
```
Ehco1996, line 468: just a question, why ues new blocks? seems this is not a common usage, cau you tell me under what circumstances should I use `local code block`😬

dm/relay/relay_test.go
```diff
@@ -453,14 +455,44 @@ func (t *testRelaySuite) TestHandleEvent(c *C) {
...
+		lm := r.meta.(*LocalMeta)
+		c.Assert(lm.BinLogName, Equals, "mysql-bin.666888")
```
Ehco1996, line 470:
```suggestion
		c.Assert(lm.BinLogName, Equals, binlogPos.Name)
```
ditto in L476
and i can not understand why `lm.BinLogPos, Equals, uint32(1234)` , what is `uint32(1234)`?

dm/relay/relay_test.go
```diff
@@ -453,14 +455,44 @@ func (t *testRelaySuite) TestHandleEvent(c *C) {
...
+
 	// reader return valid event
 	reader2.err = nil
 	reader2.result.Event = rotateEv
 
 	// writer return error
 	writer2.err = errors.New("writer error for testing")
 	// return with the annotated writer error
-	_, err = r.handleEvents(ctx, reader2, transformer2, writer2)
+	_, err = r.handleEvents(context.Background(), reader2, transformer2, writer2)
```
Ehco1996, line 495: why here need to use backgroud ctx?

dm/relay/relay.go
D3Hunter: fake rorate events are sent in many cases, not just on connect.

dm/relay/relay_test.go
D3Hunter: isolate cases to make it easier to read as we put many cases in a single test method

can also be used to isolate local variables, although not used here

dm/relay/relay_test.go
D3Hunter: see line 427

dm/relay/relay_test.go
D3Hunter: it's unstable to use a timeout context which'll expire in 10ms and this case is not expecting to test it, it causes this case fail sometimes.

dm/relay/relay.go
```diff
@@ -558,6 +559,17 @@ func (r *Relay) handleEvents(
...
+			// on the first event we got, which must be a fake rotate event, save and flush meta once to make sure
```
lance6716, line 563: > which must be a fake rotate event

seems we can move this `saveAndFlushMeta` to above `IsFakeRotateEvent` check

dm/relay/meta.go
```diff
@@ -234,6 +234,13 @@ func (lm *LocalMeta) doFlush() error {
 		return terror.ErrRelayNoCurrentUUID.Generate()
 	}
 
+	// it happens when switching to a new server, and meta is reset,
```
lance6716, line 237: this is not needed now? since we can make sure Meta file is written a right filename before reading, by your modification of `relay.go`

dm/relay/relay.go
lance6716: https://github.com/mysql/mysql-server/blob/3290a66c89eb1625a7058e0ef732432b6952b435/sql/rpl_binlog_sender.cc#L395-L402

seems there's no other cases of sending fake rotate events, so i also agree we move this code inside above `IsFakeRotateEvent` check

D3Hunter: see https://github.com/mysql/mysql-server/blob/3290a66c89eb1625a7058e0ef732432b6952b435/sql/rpl_binlog_sender.h#L247-L273

dm/relay/meta.go
D3Hunter: good point

lance6716 (approved):

lance6716:
some DM lint CI failed @D3Hunter 

https://github.com/pingcap/ticdc/runs/4064829003?check_suite_focus=true

Ehco1996:
```
/home/jenkins/agent/workspace/dm_ghpr_integration_test/go/src/github.com/pingcap/ticdc/dm/tests/shardddl1/conf/single-source-no-sharding.yaml --remove-meta"

[2021-11-01T08:04:11.588Z] dmctl test cmd: "query-status test"

[2021-11-01T08:04:11.588Z] got=1 expected=1

[2021-11-01T08:04:12.152Z] check diff successfully

[2021-11-01T08:04:12.152Z] merge dmls less than 5, insertMergeCnt: 0, updateMergeCnt: 0, deleteMergeCnt: 0

[2021-11-01T08:04:12.152Z] ...

[2021-11-01T08:04:12.152Z] make: *** [dm_integration_test] Error 1
```

seems a unsteable test? cc @D3Hunter @GMHDBJD

PR #3169, opened 2021-10-28
[DNM] extract consumer related thing to package.

3AceShowHand:
This PR tries to extract Kafka consumer-related things to `pkg`, then can be reused by other project, such as  the `test-infra`

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes

### Release note 

```release-note
Please add a release note.
```


cdc/sink/producer/kafka/kafka.go
```diff
@@ -70,6 +74,8 @@ func NewConfig() *Config {
```
nongfushanquan, line 74: some test cases are needed

PR #3190, opened 2021-10-29
dm: turn on/off relay by source config

lance6716:

we re-support old behavour. if user set `enable-relay: true` in source config, we'll use relay replication for the bound worker.

note that `start-relay -s source workername` (whether executed now or upgraded from metadata of old version) on a `enable-relay: true` source is undefined behaviour.

⬆️

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### Release note 

```release-note
`enable-relay: true` in source config takes effect again
```


dm/dm/worker/server.go
```diff
@@ -340,6 +340,10 @@ func (s *Server) observeRelayConfig(ctx context.Context, rev int64) error {
 					}
 					rev = rev1
 					if relaySource == nil {
+						if w := s.getWorker(false); w != nil && w.startedRelayBySourceCfg {
```
lichunzhu, line 343: `s` is not locked here. Should we use lock?
```suggestion
						if w := s.getWorker(true); w != nil && w.startedRelayBySourceCfg {
```

dm/dm/worker/server.go
```diff
@@ -340,6 +340,10 @@ func (s *Server) observeRelayConfig(ctx context.Context, rev int64) error {
...
+							break
+						}
+						log.L().Info("didn't found relay config after etcd retryable error")
```
lichunzhu, line 346: Just suggestion
```suggestion
						log.L().Info("didn't found relay config after etcd retryable error. Will start to stop relay now")
```

lichunzhu (approved):
Do we have an integration test for this? If not, we can add that in a later PR

lance6716:
> Do we have an integration test for this? If not, we can add that in a later PR

Our old CI has set many `enable-relay: true`. I changed many "Running" number check in test scripts because now "Running" relay unit is started, so I think the current integration test should be enough.

dm/dm/worker/server.go
lance6716: above comments should be fixed in https://github.com/pingcap/ticdc/pull/3190/commits/4d349c7e3dd1ec5592b55849e75a35a9ae8f7c77

dm/dm/worker/server.go
```diff
@@ -340,6 +340,10 @@ func (s *Server) observeRelayConfig(ctx context.Context, rev int64) error {
 					}
 					rev = rev1
 					if relaySource == nil {
+						if w := s.getWorker(true); w != nil && w.startedRelayBySourceCfg {
```
Ehco1996, line 343: maybe some retry logic can move in to `w.handleRelayStage`, for retry, jus call handleRelayStage again?

Ehco1996: don't have to change it in this pr, I'm just making a slight suggestion

lance6716: what's the relation of this function and handleRelayStage?

Ehco1996: sorry, it's `handleRelayConfig` not `handleRelayStage`
i think from L324 to L377 is retry logic for `handleRelayConfig`, so maybe it's beterr handle retry in `handleRelayConfig`

Ehco1996 (approved):
LGTM 🉑 please fix the confilcts

GMHDBJD:
Need update upstream-switch case https://github.com/pingcap/ticdc/runs/4209093582?check_suite_focus=true

GMHDBJD:
need update https://github.com/pingcap/ticdc/runs/4209093582?check_suite_focus=true @lance6716

PR #3192, opened 2021-10-29
cdc/sink: adjust kafka initialization logic

### LINKED ISSUE #3337: [ticdc 5.3.0] kafka changefeed error: kafka server: Message was too large
### What did you do?

-here's kafka topic server.log: image

- upgrade tidb cluster from 5.2.1 to 5.3.0

- remove all changefeed, than create 2 new changefeeds: tiup ctl:v5.3.0 cdc changefeed create --pd=http://172.16.4.158:2379 --sink-uri="mysql://root:123456@172.16.6.35:35002/"  --changefeed-id="mysql-task-1" ;tiup ctl:v5.3.0 cdc changefeed create --pd=http://172.16.4.158:2379 --sink-uri='kafka://172.16.6.24:9092/cdc-events?'  --changefeed-id="kafka-task-1"
- Run workload:go-tpc tpcc --db workload --warehouses 20 -T 20 --host 172.16.4.158 --port 4000 --parts 1 prepare
- kafka topic cdc-events has 6 partitions, max.message.bytes -> 1000012

_No response_

### What did you see instead?

 "id": "kafka-task-1",
    "summary": {
      "state": "error",
      "tso": 428963091419234305,
      "checkpoint": "2021-11-08 17:38:51.891",
      "error": {
        "addr": "172.16.4.158:8300",
        "code": "CDC:ErrProcessorUnknown",
        "message": "[CDC:ErrKafkaAsyncSendMessage]kafka: Failed to produce message to topic cdc-events: kafka server: Message was too large, server rejected it to avoid allocation error.: kafka: Failed to produce message to topic cdc-events: kafka server: Message was too large, server rejected it to avoid allocation error."
      }
    }
  }
]

- kafka changefeed error and won't sync data.

### Versions of the cluster

```
elease Version: v5.3.0
Edition: Community
Git Commit Hash: e401a0606bf84214bd8e765295ec6cfadd9ba24e
Git Branch: heads/refs/tags/v5.3.0
UTC Build Time: 2021-11-08 04:40:23
GoVersion: go1.16.4
Race Enabled: false
TiKV Min Version: v3.0.0-60965b006877ca7234adaced7890d7b029ed1306
Check Table Before Drop: false |```

```
./cdc version
Release Version: v5.3.0
Git Commit Hash: 7dc044edd1a894c4e6846c27673069ccfc920d56
Git Branch: heads/refs/tags/v5.3.0
UTC Build Time: 2021-11-08 04:37:11
Go Version: go version go1.16.4 linux/amd64
Failpoint Build: false```

3AceShowHand:

close #3337

This is part work of making Kafka GA. Contain the following:
* Refine the logic of creating topic
* Make some variable become a field of the `Config`, which would be helpful if we would support one producer per topic in the future since this group producer parameters together.
* Bugfix
  * If `auto-create-topic` disable by the user, the partition count will not be checked.
  * before create the topic, make sure broker's `message.max.bytes` not less than `max-message-bytes`

### Check List 

Tests 

 - Unit test
Screen Shot 2021-11-09 at 13 24 36

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/cdc/sink/mq.go
+++ b/cdc/sink/mq.go
@@ -397,7 +397,9 @@ func newKafkaSaramaSink(ctx context.Context, sinkURI *url.URL, filter *filter.Fi
 		return nil, cerror.ErrKafkaInvalidConfig.GenWithStack("topic name not found")
 	}
 
-	producer, err := kafka.NewKafkaSaramaProducer(ctx, topic, config, errCh)
+	var protocol codec.Protocol
+	protocol.FromString(replicaConfig.Sink.Protocol)
+	producer, err := kafka.NewKafkaSaramaProducer(ctx, topic, protocol, config, errCh)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
--- a/cdc/sink/mq.go
+++ b/cdc/sink/mq.go
@@ -16,6 +16,7 @@ package sink
 import (
 	"context"
 	"net/url"
+	"strings"
 	"sync/atomic"
 	"time"
 
@@ -389,7 +390,14 @@ func newKafkaSaramaSink(ctx context.Context, sinkURI *url.URL, filter *filter.Fi
 		return nil, cerror.WrapError(cerror.ErrKafkaInvalidConfig, err)
 	}
 
-	producer, err := kafka.NewKafkaSaramaProducer(ctx, config, errCh)
+	topic := strings.TrimFunc(sinkURI.Path, func(r rune) bool {
+		return r == '/'
+	})
+	if topic == "" {
+		return nil, cerror.ErrKafkaInvalidConfig.GenWithStack("topic name not found")
+	}
+
+	producer, err := kafka.NewKafkaSaramaProducer(ctx, topic, config, errCh)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
--- a/cdc/sink/mq.go
+++ b/cdc/sink/mq.go
@@ -394,7 +394,7 @@ func newKafkaSaramaSink(ctx context.Context, sinkURI *url.URL, filter *filter.Fi
 		return r == '/'
 	})
 	if topic == "" {
-		return nil, cerror.ErrKafkaInvalidConfig.GenWithStack("topic name not found")
+		return nil, cerror.ErrKafkaInvalidConfig.GenWithStack("no topic is specified in sink-uri")
 	}
 
 	var protocol codec.Protocol
--- a/cdc/sink/producer/kafka/kafka.go
+++ b/cdc/sink/producer/kafka/kafka.go
@@ -480,6 +480,16 @@ func topicPreProcess(topic string, config *Config, saramaConfig *sarama.Config) 
 			zap.String("topic", topic), zap.Int32("partitions", config.PartitionNum))
 		config.PartitionNum = defaultPartitionNum
 	}
+
+	brokerConfigEntry, err := admin.DescribeConfig(sarama.ConfigResource{
+		Type:        sarama.BrokerResource,
+		Name:        "message.max.bytes",
+		ConfigNames: nil,
+	})
+	if err != nil {
+		return errors.Trace(err)
+	}
+
 	maxMessageBytes := strconv.Itoa(config.MaxMessageBytes)
 	err = admin.CreateTopic(topic, &sarama.TopicDetail{
 		NumPartitions:     config.PartitionNum,
--- a/cdc/sink/producer/kafka/kafka.go
+++ b/cdc/sink/producer/kafka/kafka.go
@@ -449,7 +449,7 @@ func topicPreProcess(config *Config, saramaConfig *sarama.Config) error {
 			NumPartitions:     config.PartitionNum,
 			ReplicationFactor: config.ReplicationFactor,
 		}, false)
-		// TODO idenfity the cause of "Topic with this name already exists"
+		// TODO identify the cause of "Topic with this name already exists"
 		if err != nil && !strings.Contains(err.Error(), "already exists") {
 			return err
 		}
--- a/cdc/sink/producer/kafka/kafka.go
+++ b/cdc/sink/producer/kafka/kafka.go
@@ -420,83 +420,72 @@ func topicPreProcess(topic string, config *Config, saramaConfig *sarama.Config) 
 		return err
 	}
 
-	var realPartitionCount int32 = 0
-	info, ok := topics[topic]
-	if ok {
-		realPartitionCount = info.NumPartitions
-	}
-
-	// if user specify cdc to create the topic, we must make sure
-	// that `partition-num` and `replication-factor` is given
-	if config.AutoCreate {
-		// if the specified topic name already exist, we should make sure that
-		// topic's `max.message.bytes` is not less than given `max-message-size`
-		if ok {
-			if a, ok := info.ConfigEntries["max.message.bytes"]; ok {
-				topicMaxMessageBytes, err := strconv.Atoi(*a)
-				if err != nil {
-					return cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
-				}
-				if topicMaxMessageBytes < config.MaxMessageBytes {
-					return cerror.ErrKafkaInvalidConfig.GenWithStack(
-						"topic already exist, and max.message.size(%d) less than max-message-size(%d)."+
-							"Please make sure max-message-size not greater than topic max.message.size",
-						topicMaxMessageBytes, config.MaxMessageBytes)
-				}
+	info, created := topics[topic]
+	// once we have found the topic, no matter `auto-create-topic`, make sure all user input parameters are valid.
+	if created {
+		// make sure that topic's `max.message.bytes` is not less than given `max-message-size`
+		// else the producer will send message that too large to make topic reject, then changefeed would error.
+		if a, found := info.ConfigEntries["max.message.bytes"]; found {
+			topicMaxMessageBytes, err := strconv.Atoi(*a)
+			if err != nil {
+				return cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
 			}
+			if topicMaxMessageBytes < config.MaxMessageBytes {
+				return cerror.ErrKafkaInvalidConfig.GenWithStack(
+					"topic already exist, and max.message.size(%d) less than max-message-size(%d)."+
+						"Please make sure max-message-size not greater than topic max.message.size",
+					topicMaxMessageBytes, config.MaxMessageBytes)
+			}
+		}
 
-			config.PartitionNum = realPartitionCount
-			log.Warn("topic already exist", zap.String("topic", topic))
+		// no need to create the topic, but we would have to log user if they found enter wrong topic name later
+		if config.AutoCreate {
+			log.Warn("topic already exist, TiCDC will not create the topic",
+				zap.String("topic", topic), zap.Any("detail", info))
+		}
+
+		realPartitionCount := info.NumPartitions
+		if config.PartitionNum < realPartitionCount {
+			log.Warn("number of partition specified in sink-uri is less than that of the actual topic. "+
+				"Some partitions will not have messages dispatched to",
+				zap.Int32("sink-uri partitions", config.PartitionNum),
+				zap.Int32("topic partitions", realPartitionCount))
 			return nil
 		}
-		// this could happen if user does not specify the `partition-num` in the sink uri.
-		if config.PartitionNum == 0 {
-			config.PartitionNum = defaultPartitionNum
-			log.Warn("partition-num is not set, use the default partition count",
-				zap.String("topic", topic), zap.Int32("partitions", config.PartitionNum))
+
+		// Make sure that the user-specified partition count is not greater than the real partition count,
+		// since messages would be dispatched to different partitions, this could prevent potential correctness problems.
+		if config.PartitionNum > realPartitionCount {
+			return cerror.ErrKafkaInvalidPartitionNum.GenWithStack(
+				"the number of partition (%d) specified in sink-uri is more than that of actual topic (%d)",
+				config.PartitionNum, realPartitionCount)
 		}
-		maxMessageBytes := strconv.Itoa(config.MaxMessageBytes)
-		err := admin.CreateTopic(topic, &sarama.TopicDetail{
-			NumPartitions:     config.PartitionNum,
-			ReplicationFactor: config.ReplicationFactor,
-			ConfigEntries: map[string]*string{
-				"max.message.bytes": &maxMessageBytes,
-			},
-		}, false)
-		// TODO identify the cause of "Topic with this name already exists"
-		if err != nil && !strings.Contains(err.Error(), "already exists") {
-			return cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
-		}
+
 		return nil
 	}
 
-	// if `auto-create-topic` is disabled by user, we would assume topic should have already created
-	if !ok {
-		return cerror.ErrKafkaInvalidConfig.GenWithStack("auto-create-topic is false, and topic not found")
+	if !config.AutoCreate {
+		return cerror.ErrKafkaInvalidConfig.GenWithStack("`auto-create-topic` is false, and topic not found")
 	}
 
+	// topic not created yet, and user does not specify the `partition-num` in the sink uri.
 	if config.PartitionNum == 0 {
-		config.PartitionNum = realPartitionCount
-		return nil
+		log.Warn("partition-num is not set, use the default partition count",
+			zap.String("topic", topic), zap.Int32("partitions", config.PartitionNum))
+		config.PartitionNum = defaultPartitionNum
 	}
-
-	if config.PartitionNum < realPartitionCount {
-		log.Warn("number of partition specified in sink-uri is less than that of the actual topic. "+
-			"Some partitions will not have messages dispatched to",
-			zap.Int32("sink-uri partitions", config.PartitionNum),
-			zap.Int32("topic partitions", realPartitionCount))
-		return nil
+	maxMessageBytes := strconv.Itoa(config.MaxMessageBytes)
+	err = admin.CreateTopic(topic, &sarama.TopicDetail{
+		NumPartitions:     config.PartitionNum,
+		ReplicationFactor: config.ReplicationFactor,
+		ConfigEntries: map[string]*string{
+			"max.message.bytes": &maxMessageBytes,
+		},
+	}, false)
+	// TODO identify the cause of "Topic with this name already exists"
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
+		return cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
 	}
-
-	// For a topic already existing, we just make sure that the user-specified
-	// partition count is not greater than the real partition count, since
-	// messages would be dispatched to different partitions, this could prevent potential correctness problems.
-	if config.PartitionNum > realPartitionCount {
-		return cerror.ErrKafkaInvalidPartitionNum.GenWithStack(
-			"the number of partition (%d) specified in sink-uri is more than that of actual topic (%d)",
-			config.PartitionNum, realPartitionCount)
-	}
-
 	return nil
 }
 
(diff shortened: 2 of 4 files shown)


cdc/sink/producer/kafka/kafka.go
```diff
@@ -64,12 +69,21 @@ func NewConfig() *Config {
...
+		AutoCreate:        true,
 	}
 }
 
 // Initialize the kafka configuration
 func (c *Config) Initialize(sinkURI *url.URL, replicaConfig *config.ReplicaConfig, opts map[string]string) error {
+	c.BrokerEndpoints = strings.Split(sinkURI.Host, ",")
+	topic := strings.TrimFunc(sinkURI.Path, func(r rune) bool {
+		return r == '/'
+	})
```
overvenus, line 81: Could you add some comments? Is it a bugfix?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +409,120 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
 	}
 }
 
-// kafkaTopicPreProcess gets partition number from existing topic, if topic doesn't
-// exit, creates it automatically.
-func kafkaTopicPreProcess(topic, address string, config *Config, cfg *sarama.Config) (int32, error) {
-	admin, err := sarama.NewClusterAdmin(strings.Split(address, ","), cfg)
+func topicPreProcess(config *Config, saramaConfig *sarama.Config) error {
+	failpoint.Inject("workaround4Test", func() {
```
overvenus, line 413: Could you avoid failpoints in unit tests? Can it be replaced by mocking `topicPreProcess`?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -550,7 +604,9 @@ func newSaramaConfig(ctx context.Context, c *Config) (*sarama.Config, error) {
 	}
 	config.Version = version
 	// See: https://kafka.apache.org/documentation/#replication
-	// When one of the brokers in a Kafka cluster is down, the partition leaders in this broker is broken, Kafka will election a new partition leader and replication logs, this process will last from a few seconds to a few minutes. Kafka cluster will not provide a writing service in this process.
+	// When one of the brokers in a Kafka cluster is down, the partition leaders in this broker is broken,
+	// Kafka will election a new partition leader and replication logs, this process will last from a few seconds to a few minutes.
```
overvenus, line 608: Please keep line width within 80 chars.
  
```suggestion
	// When one of the brokers in a Kafka cluster is down, the partition leaders
	// in this broker is broken, Kafka will election a new partition leader and
	// replication logs, this process will last from a few seconds to a few
	// minutes.
```

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +409,120 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+		return err
+	}
+
+	var realPartitionCount int32 = 0
+	info, ok := topics[config.TopicName]
+	if ok {
+		realPartitionCount = info.NumPartitions
+	}
+
+	// if user specify cdc to create the topic, we must make sure that `partition-num` and `replication-factor` is given
```
overvenus, line 437: Please keep line width within 80 chars.

cdc/sink/producer/kafka/kafka.go
3AceShowHand: no, this only move the topic name into the config.

If we would support multiple producers in the future, this is helpful, since we can construct different configs for different topics.

cdc/sink/producer/kafka/kafka.go
3AceShowHand: It was quite hard to mock this method, since this method is defined in the package `sink/kafka`, but will also be called by unit test in the `sink` package.

If try to mock the method, have to make this method public, I think it's unreasonable.

failpoint would be the easiest way to workaround this.

liuzix: Maybe we can leave a `FIXME` here for now?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +409,121 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+		realPartitionCount = info.NumPartitions
+	}
+
+	// if user specify cdc to create the topic, we must make sure
+	// that `partition-num` and `replication-factor` is given
+	if config.AutoCreate {
+		// if the specified topic name already exist, we should make sure that
+		// topic's `max.message.bytes` is not less than given `max-message-size`
+		if ok {
+			if a, ok := info.ConfigEntries["max.message.bytes"]; ok {
```
liuzix, line 443: Do we need to check `max.request.size`?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +409,121 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+				topicMaxMessageBytes, err := strconv.Atoi(*a)
+				if err != nil {
+					return err
```
liuzix, line 446: Please wrap the error so that the user has a basic idea how it happens.

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +409,121 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+				}
+				if topicMaxMessageBytes < config.MaxMessageBytes {
+					return cerror.ErrKafkaInvalidConfig.GenWithStack(
+						"topic already exist, and max.message.size(%d) less than max-message-size(%d)",
```
liuzix, line 450: Leave a short hint for user to fix?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +409,121 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+			config.PartitionNum = realPartitionCount
+			log.Warn("topic already exist", zap.String("topic", config.TopicName))
+			return nil
 		}
-	} else {
-		if partitionNum == 0 {
-			partitionNum = defaultPartitionNum
-			log.Warn("topic not found and partition number is not specified, using default partition number", zap.String("topic", topic), zap.Int32("partition_num", partitionNum))
+		// this could happen if user does not specify the `partition-num` in the sink uri.
+		if config.PartitionNum == 0 {
```
liuzix, line 460: Is it acceptable if the user specifies `partition-num=0`?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +409,121 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+		return cerror.ErrKafkaInvalidConfig.GenWithStack("auto-create-topic is false, and topic not found")
+	}
+
+	if config.PartitionNum == 0 {
+		config.PartitionNum = realPartitionCount
+		return nil
+	}
+
+	if config.PartitionNum < realPartitionCount {
+		log.Warn("partition count assigned in sink-uri is less than that of topic, some partitions would not have messages to be dispatched",
```
liuzix, line 491:
```suggestion
		log.Warn("number of partition specified in sink-uri is less than that of the actual topic. Some partitions will not have messages dispatched to",
```

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +409,121 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+			zap.Int32("sink-uri partitions", config.PartitionNum), zap.Int32("topic partitions", realPartitionCount))
+		return nil
+	}
+
+	// for topic already exist, we just make sure that the user specified partition count is not greater than the real
```
liuzix, line 496:
```suggestion
	// For topic already existing, we just make sure that the user-specified partition count is not greater than the real
```

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +409,121 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+	// partition count, since we would dispatch messages to different partitions, this could prevent potential message loss.
```
liuzix, line 497: Please reword this sentence. `since we would dispatch messages to different partitions` does not have a clear meaning.

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +409,121 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+	if config.PartitionNum > realPartitionCount {
+		return cerror.ErrKafkaInvalidPartitionNum.GenWithStack(
+			"partition number(%d) assigned in sink-uri is more than that of topic(%d)", config.PartitionNum, realPartitionCount)
```
liuzix, line 500:
```suggestion
			"the number of partition (%d) specified in sink-uri is more than that of actual topic (%d)", config.PartitionNum, realPartitionCount)
```

cdc/sink/producer/kafka/kafka.go
3AceShowHand: this could happen if the user does not specify the `partition-num`, which would be 0.

liuzix: Is it checked that the user does not specified `partition-num=0`?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -37,11 +37,16 @@ import (
 	"go.uber.org/zap"
 )
 
-const defaultPartitionNum = 4
+const defaultPartitionNum = 3
```
amyangfei, line 40: Why is this value changed?

3AceShowHand: 1. we would use partition-num as the mod for partition dispatch, and `3` is a prime number, much better for hash
2. `3` is a much more reasonable number for replication, and as I know, Kafka would use `KRaft` for consistency protocol, it's good for `raft`
3. `3` would consume less memory usage.

cdc/sink/producer/kafka/kafka.go
3AceShowHand: `max.request.size` is a producer level configuration.
I think when creating the sarama producer, it's `MaxMessageBytes` is for this.

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +400,127 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
 	}
 }
 
-// kafkaTopicPreProcess gets partition number from existing topic, if topic doesn't
-// exit, creates it automatically.
-func kafkaTopicPreProcess(topic, address string, config *Config, cfg *sarama.Config) (int32, error) {
-	admin, err := sarama.NewClusterAdmin(strings.Split(address, ","), cfg)
+func topicPreProcess(topic string, config *Config, saramaConfig *sarama.Config) error {
+	// FIXME: find a way to remove this failpoint for workload the unit test
+	failpoint.Inject("workaround4Test", func() {
```
amyangfei, line 405: Better to rename failpoint injection to `SkipTopicAutoCreate`

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +400,127 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+		failpoint.Return(nil)
+	})
+	admin, err := sarama.NewClusterAdmin(config.BrokerEndpoints, saramaConfig)
 	if err != nil {
-		return 0, cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
+		return err
```
amyangfei, line 410: Should wrap `err` with an RFC error

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +400,127 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+				}
+				if topicMaxMessageBytes < config.MaxMessageBytes {
+					return cerror.ErrKafkaInvalidConfig.GenWithStack(
+						"topic already exist, and max.message.size(%d) less than max-message-size(%d)."+
+							"Please make sure max-message-size not greater than topic max.message.size",
+						topicMaxMessageBytes, config.MaxMessageBytes)
+				}
+			}
+
+			config.PartitionNum = realPartitionCount
```
amyangfei, line 448: Is `realPartitionCount` zero when no topic exists before?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +400,127 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+		realPartitionCount = info.NumPartitions
+	}
+
+	// if user specify cdc to create the topic, we must make sure
+	// that `partition-num` and `replication-factor` is given
+	if config.AutoCreate {
+		// if the specified topic name already exist, we should make sure that
+		// topic's `max.message.bytes` is not less than given `max-message-size`
+		if ok {
+			if a, ok := info.ConfigEntries["max.message.bytes"]; ok {
```
amyangfei, line 435: `info` could be nil

cdc/sink/producer/kafka/kafka.go
amyangfei: ignore this

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +400,127 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+		return nil
+	}
+
+	// if `auto-create-topic` is disabled by user, we would assume topic should have already created
+	if !ok {
+		return cerror.ErrKafkaInvalidConfig.GenWithStack("auto-create-topic is false, and topic not found")
+	}
+
+	if config.PartitionNum == 0 {
+		config.PartitionNum = realPartitionCount
```
amyangfei, line 479: `realPartitionCount` could be zero here?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +400,127 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+	if config.AutoCreate {
+		// if the specified topic name already exist, we should make sure that
+		// topic's `max.message.bytes` is not less than given `max-message-size`
+		if ok {
+			if a, ok := info.ConfigEntries["max.message.bytes"]; ok {
+				topicMaxMessageBytes, err := strconv.Atoi(*a)
+				if err != nil {
+					return cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
+				}
+				if topicMaxMessageBytes < config.MaxMessageBytes {
```
amyangfei, line 440: Do we also need to check max message bytes when topic exists and `AutoCreate` is false?

cdc/sink/producer/kafka/kafka.go
amyangfei, line 448: The original logic, if `partition-num` in sink-uri is larger than 0, and less than the partition number of existing topic, will set `config.PartitionNum = partition num in sink uri`, seems the behavior is changed here?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +400,127 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
 			ReplicationFactor: config.ReplicationFactor,
+			ConfigEntries: map[string]*string{
+				"max.message.bytes": &maxMessageBytes,
+			},
 		}, false)
-		// TODO idenfity the cause of "Topic with this name already exists"
+		// TODO identify the cause of "Topic with this name already exists"
 		if err != nil && !strings.Contains(err.Error(), "already exists") {
-			return 0, cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
+			return err
```
amyangfei, line 468: Should use `cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)`

cdc/sink/producer/kafka/kafka.go
3AceShowHand: If the user does not specify `partition-num` in the sink uri, it will be 0, which means uninitialized.

amyangfei: Well, it must be `AutoCreate == false` and topic exists when running at L479.
Can we extract several independent part, such as
- determine and check config.PartitionNum
- create topic if needed
- check max message bytes if needed

cdc/sink/producer/kafka/kafka.go
3AceShowHand: we have check `partition-num` in https://github.com/pingcap/ticdc/blob/master/cdc/sink/producer/kafka/kafka.go#L74

But if they does not pass `partition-num` in the sink-uri, it would be 0.

overvenus (approved):
LGTM

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +399,122 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
-		err := admin.CreateTopic(topic, &sarama.TopicDetail{
-			NumPartitions:     partitionNum,
-			ReplicationFactor: config.ReplicationFactor,
-		}, false)
-		// TODO idenfity the cause of "Topic with this name already exists"
-		if err != nil && !strings.Contains(err.Error(), "already exists") {
-			return 0, cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
+
+		if err := config.adjustPartitionNum(info.NumPartitions); err != nil {
+			return cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
```
amyangfei, line 448: Better to `return err` directly, since the `err` returned from `config.adjustPartitionNum` is already an RFC error

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +399,122 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+		return cerror.ErrKafkaInvalidConfig.GenWithStack("`auto-create-topic` is false, and topic not found")
+	}
+
+	// when try to create the topic, we don't know how to set the `max.message.bytes` for the topic.
+	// Kafka would create the topic with broker's `message.max.bytes`,
+	// we have to make sure it's not greater than `max-message-bytes`
+	brokerMessageMaxBytes, err := getBrokerMessageMaxBytes(admin)
+	if err != nil {
+		log.Warn("TiCDC cannot find `message.max.bytes` from broker's configuration")
+		return cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
```
amyangfei, line 464: ditto, it is better to return RFC error from function `getBrokerMessageMaxBytes`

cdc/sink/producer/kafka/kafka.go
```diff
@@ -395,80 +399,122 @@ func (k *kafkaSaramaProducer) run(ctx context.Context) error {
...
+	if config.PartitionNum == 0 {
+		config.PartitionNum = defaultPartitionNum
+		log.Warn("partition-num is not set, use the default partition count",
+			zap.String("topic", topic), zap.Int32("partitions", config.PartitionNum))
+	}
+
+	err = admin.CreateTopic(topic, &sarama.TopicDetail{
+		NumPartitions:     config.PartitionNum,
+		ReplicationFactor: config.ReplicationFactor,
+	}, false)
```
amyangfei, line 484: Do we need to pass the `max.message.bytes` config here?

3AceShowHand: no, let broker decide the value.

amyangfei (approved):

cdc/sink/mq.go
```diff
@@ -393,7 +393,13 @@ func newKafkaSaramaSink(ctx context.Context, sinkURI *url.URL, filter *filter.Fi
 	topic := strings.TrimFunc(sinkURI.Path, func(r rune) bool {
 		return r == '/'
 	})
-	producer, err := kafka.NewKafkaSaramaProducer(ctx, sinkURI.Host, topic, config, errCh)
+	if topic == "" {
+		return nil, cerror.ErrKafkaInvalidConfig.GenWithStack("topic name not found")
```
liuzix, line 397:
```suggestion
		return nil, cerror.ErrKafkaInvalidConfig.GenWithStack("no topic is specified in sink-uri")
```

liuzix (commented):
rest LGTM

liuzix (approved):

PR #3203, opened 2021-10-31
codec: canal-json support batch decoder

3AceShowHand:

This is a partwork of making Kafka Sink GA, support batch decoder interface for Canal-JSON format.

* implement the `EventBatchDecoder` for Canal-JSON protocol.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
Please add a release note.
```


cdc/sink/producer/kafka/kafka.go
```diff
@@ -457,18 +457,20 @@ func topicPreProcess(topic string, protocol codec.Protocol, config *Config, sara
...
 	// when try to create the topic, we don't know how to set the `max.message.bytes` for the topic.
 	// Kafka would create the topic with broker's `message.max.bytes`,
-	// we have to make sure it's not greater than `max-message-bytes`
-	brokerMessageMaxBytes, err := getBrokerMessageMaxBytes(admin)
-	if err != nil {
-		log.Warn("TiCDC cannot find `message.max.bytes` from broker's configuration")
-		return errors.Trace(err)
-	}
+	// we have to make sure it's not greater than `max-message-bytes` for the default open protocol & craft protocol.
+	if protocol == codec.ProtocolDefault || protocol == codec.ProtocolCraft {
```
amyangfei, line 461: Is this change related to https://github.com/pingcap/ticdc/pull/3192?

3AceShowHand: Yes. I was thinking about should we check `max-message-bytes` < `message.max.bytes` for protocol which not batching messages, such as Canal-JSON, and I think it's unnecessary, because most of the time, the Canal-JSON message should be successfully sent, and once meet size too large problem, it's unrelated to `max-message-bytes`, so only check for open protocol and craft.

cdc/sink/codec/canal_flat.go
```diff
@@ -362,3 +441,146 @@ func (c *CanalFlatEventBatchEncoder) SetParams(params map[string]string) error {
...
 }
+
+// CanalFlatEventBatchDecoder decodes the byte into the original message.
+type CanalFlatEventBatchDecoder struct {
+	data                []byte
+	msg                 *MQMessage
+	enableTiDBExtension bool
+}
+
+func NewCanalFlatEventBatchDecoder(data []byte, enableTiDBExtension bool) (EventBatchDecoder, error) {
```
lonng, line 452: Seems the function `NewCanalFlatEventBatchDecoder` never returns an error, but why do you define your signature like this?

cdc/sink/codec/canal_flat_test.go
```diff
@@ -90,6 +98,85 @@ func (s *canalFlatSuite) TestNewCanalFlatMessageFromDML(c *check.C) {
...
+	encodedBytes, err := charmap.ISO8859_1.NewDecoder().Bytes([]byte("测试blob"))
+	c.Assert(err, check.IsNil)
+	expected := map[string]interface{}{
+		"id":         "1",
+		"name":       "Bob",
+		"tiny":       "255",
+		"comment":    "测试",
+		"blob":       string(encodedBytes),
+		"journalist": "Sharon run so fast",
+		"elder":      "图样图森破",
```
lonng, line 113: Prefer to use more common non-English characters.

cdc/sink/codec/canal_flat.go
```diff
@@ -362,3 +441,146 @@ func (c *CanalFlatEventBatchEncoder) SetParams(params map[string]string) error {
...
+
+	return result
+}
+
+func canalFlatJSONColumnMap2SinkColumns(cols map[string]interface{}, mysqlType map[string]string) []*model.Column {
+	result := make([]*model.Column, 0, len(cols))
+	for name, value := range cols {
+		typeStr, ok := mysqlType[name]
+		if !ok {
+			log.Panic("mysql type does not found", zap.String("column name", name), zap.Any("mysqlType", mysqlType))
```
amyangfei, line 551: Can this return error, instead of panic.
Supposing TiCDC really meets panic here, the replication will pause, TiCDC nodes will panic one by one (given that the table will be rescheduled.

cdc/sink/codec/canal_flat.go
```diff
@@ -39,6 +42,36 @@ type CanalFlatEventBatchEncoder struct {
 
 const tidbWaterMarkType = "TIDB_WATERMARK"
 
+var str2MySQLType = map[string]byte{
```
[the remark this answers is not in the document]
3AceShowHand: At the moment, it's a reverse map of the `https://github.com/pingcap/tidb/blob/master/parser/types/etc.go#L43`.

But it would be hard to track TiDB type changes. If we would update the TiDB dependency in the go mod file, we have to manually check for this.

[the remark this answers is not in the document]
overvenus: How about exporting `type2Str` in parser and reversing map at runtime?

overvenus (commented):
Left one comment https://github.com/pingcap/ticdc/pull/3203/files#r750851191, rest LGTM

cdc/sink/codec/canal_flat.go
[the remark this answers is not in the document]
3AceShowHand: If we can find a way to only reversing the map once, I think exporting `type2Str` is a good choice, otherwise, I would prefer move `str2MySQLType` to the parser directly.

amyangfei (approved):

cdc/sink/producer/kafka/kafka.go
```diff
@@ -457,18 +457,20 @@ func topicPreProcess(topic string, protocol codec.Protocol, config *Config, sara
 
 	// when try to create the topic, we don't know how to set the `max.message.bytes` for the topic.
 	// Kafka would create the topic with broker's `message.max.bytes`,
-	// we have to make sure it's not greater than `max-message-bytes`
-	brokerMessageMaxBytes, err := getBrokerMessageMaxBytes(admin)
-	if err != nil {
-		log.Warn("TiCDC cannot find `message.max.bytes` from broker's configuration")
-		return errors.Trace(err)
-	}
+	// we have to make sure it's not greater than `max-message-bytes` for the default open protocol & craft protocol.
```
liuzix, line 460: Can you move these to a separate PR? Rest LGTM

overvenus (approved):
LGTM

PR #3217, opened 2021-11-01
scheduler_v2: implement a data structure for storing table status

liuzix:
- To implement a peer-message based scheduler, we need a data structure to store and look up information of all the replicated tables.

- Implemented `TableSet`.
- Methods include:
```
AddTableRecord(record *TableRecord) (successful bool)
GetTableRecord(tableID model.TableID) (*TableRecord, bool)
RemoveTableRecord(tableID model.TableID) bool
RemoveTableRecordByCaptureID(captureID model.CaptureID)
CountTableByCaptureID(captureID model.CaptureID) int
GetDistinctCaptures() []model.CaptureID
GetAllTables() map[model.TableID]*TableRecord
GetAllTablesGroupedByCaptures() map[model.CaptureID]map[model.TableID]*TableRecord
CountTableByStatus(status TableStatus) (count int)
```

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/scheduler/util/table_set.go
```diff
@@ -0,0 +1,137 @@
...
+// TableRecord is a record to be inserted into TableSet.
+type TableRecord struct {
+	TableID   model.TableID
+	CaptureID model.CaptureID
+	Status    TableStatus
+}
+
+// TableStatus is a type representing the table's replication status.
+// The meaning of its value can be defined by the user of TableSet.
+type TableStatus int32
```
overvenus, line 36: What about move the type definition to the user package or define its value here? I think it's better to put type definition and its values together.

overvenus (commented):
Rest LGTM

cdc/scheduler/util/table_set.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"github.com/pingcap/ticdc/cdc/model"
+)
+
+// TableSet provides a data structure to store the tables' states for the
+// scheduler.
+type TableSet struct {
+	tableIDMap   map[model.TableID]*TableRecord
+	captureIndex map[model.CaptureID]map[model.TableID]*TableRecord
```
3AceShowHand, line 24: how about extract `map[model.TableID]*TableRecord` to a `type`.

It looks that `captureIndex` and `TableRecord` are mutual indexing each other.

cdc/scheduler/util/table_set.go
```diff
@@ -0,0 +1,137 @@
...
+func NewTableSet() *TableSet {
+	return &TableSet{
+		tableIDMap:   map[model.TableID]*TableRecord{},
+		captureIndex: map[model.CaptureID]map[model.TableID]*TableRecord{},
+	}
+}
+
+// AddTableRecord inserts a new TableRecord.
+// It returns true if it succeeds. Returns false if there is a duplicate.
+func (s *TableSet) AddTableRecord(record *TableRecord) (successful bool) {
```
3AceShowHand, line 48: Does it necessary to check whether success or not?

cdc/scheduler/util/table_set.go
```diff
@@ -0,0 +1,137 @@
...
+	delete(s.captureIndex, captureID)
+}
+
+// CountTableByCaptureID counts the number of tables associated with the captureID.
+func (s *TableSet) CountTableByCaptureID(captureID model.CaptureID) int {
+	return len(s.captureIndex[captureID])
+}
+
+// GetDistinctCaptures counts distinct captures with tables.
+func (s *TableSet) GetDistinctCaptures() []model.CaptureID {
```
3AceShowHand, line 111: `TableSet` already indicate that there is no duplicate captures, so I would prefer `GetCaptures` directly.

cdc/scheduler/util/table_set.go
```diff
@@ -0,0 +1,137 @@
...
+	var ret []model.CaptureID
+	for captureID := range s.captureIndex {
+		ret = append(ret, captureID)
+	}
+	return ret
+}
+
+// GetAllTables returns all stored information on all tables.
+func (s *TableSet) GetAllTables() map[model.TableID]*TableRecord {
```
3AceShowHand, line 120: I am curious about `the definition of table` in this context, it may be `table = tableID + tableStatus`. 

It looks that `GetAllTables` can return a slice of `TableRecord` directly.

cdc/scheduler/util/table_set.go
liuzix: I don't think they are mutually indexing each other. Can you given an example?

cdc/scheduler/util/table_set.go
liuzix: If there is a duplicate, usually there is a fatal bug. We would like to detect it so that the program can fail fast instead of introducing a correctness problem silently.

cdc/scheduler/util/table_set.go
liuzix: It would be difficult to look up a slice of `TableRecord`.

cdc/scheduler/util/table_set.go
liuzix: `TableSet` only indicates that there are no duplicate tables, but not captures.

3AceShowHand:
LGTM

cdc/scheduler/util/table_set.go
```diff
@@ -35,6 +35,12 @@ type TableRecord struct {
 // The meaning of its value can be defined by the user of TableSet.
 type TableStatus int32
 
+const (
+	AddingTable = TableStatus(iota)
```
overvenus, line 39: Could you start enums at 1?

See https://github.com/uber-go/guide/blob/master/style.md#start-enums-at-one

cdc/scheduler/util/table_set.go
```diff
@@ -0,0 +1,143 @@
...
+	var ret []model.CaptureID
+	for captureID := range s.captureIndex {
+		ret = append(ret, captureID)
+	}
+	return ret
+}
+
+// GetAllTables returns all stored information on all tables.
+func (s *TableSet) GetAllTables() map[model.TableID]*TableRecord {
+	return s.tableIDMap
```
overvenus, line 127: What about return a copied map?

https://github.com/uber-go/guide/blob/master/style.md#returning-slices-and-maps

cdc/scheduler/util/table_set.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// GetAllTablesGroupedByCaptures returns all stored information grouped by associated CaptureID.
+func (s *TableSet) GetAllTablesGroupedByCaptures() map[model.CaptureID]map[model.TableID]*TableRecord {
```
nongfushanquan, line 131: it is unnecessary to  provide  the Tableset class if the user can access the inner struct

cdc/scheduler/util/table_set.go
overvenus: Does the comment at line 35 need update?

cdc/scheduler/util/table_set.go
liuzix: No, the user can't

cdc/scheduler/util/table_set.go
```diff
@@ -0,0 +1,165 @@
...
+package util
+
+import (
+	"github.com/pingcap/ticdc/cdc/model"
+)
+
+// TableSet provides a data structure to store the tables' states for the
+// scheduler.
+type TableSet struct {
+	tableIDMap   map[model.TableID]*TableRecord
```
asddongmen, line 23: may be we can add a comment about here.

asddongmen (approved):

3AceShowHand (approved):

PR #3218, opened 2021-11-01
bug-fix: fix duplicate key for insert with compactor is enabled

GMHDBJD:
In the original design, I imagined that insert, delete, and update would be executed with the order of delete, replace, and update. But now we use causality after compactor to execute in the original order and insert will not become replace(insert on duplicate update) unless safemode is true. So we need to explicitly set the safemode to true for delete + insert = insert.

- set safemode for delete + insert = insert
- fix the originnal unstable test

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


GMHDBJD:
cc @lance6716 @Ehco1996 @glorv

glorv (approved):

lichunzhu (approved):

lance6716:
/hold

wait dm-integration-test to pass

lance6716:
```
[2021-11-01T08:41:04.886Z]     import requests
[2021-11-01T08:41:04.886Z] ImportError: No module named requests
```

😵 

also there's make fmt needed @GMHDBJD

Issue #3112, opened 2021-10-20
EtcdWorker ticks too often


liuzix:
### What did you do?

Create 32 changefeeds corresponding to 32 tables. Run the CDC cluster.

### What did you expect to see?

The Etcd server (embedded in PD) would experience a reasonable amount of QPS pressure.

### What did you see instead?

The total QPS to Etcd can be as high as `1000`. It can be higher. The exact quantitative relation with the number of changefeeds is still under investigation.

### Versions of the cluster

v5.1.x, v5.2.x

overvenus:
This issue has been mitigated, change severity to major.

asddongmen:
It Seems that this bug has been solved by #3219 and #3217 , if it happends again, feel free to reopen this issue and @ me.

PR #3219, opened 2021-11-01
CDC/etcd_worker: add rate limiter to limit EtcdWorker tick frequency

asddongmen:
#3112 
Too frequent etcd worker tick will cause etcd to be overburdened, and the current etcd qps will increase exponentially with the number of tables that need to be replicated.

Add rate limiter to limit EtcdWorker ticks frequency.

### Check List 

Tests 

**Summary:** Limiting the tick frequency of `EtcdWoker` can reduce etcd qps by about 50%, but it will reduce the replication speed by about 20%.

**Test Environment:**  
1. 2 Machines: 8-core 16G memory 500G high-speed disk
2. Sink Type: mysql 
3. Cluster topo: 
```yaml
pd_servers:
  - host: 172.x.x.x
  - host: 172.x.x.x

tidb_servers:
  - host: 172.x.x.x
  - host: 172.x.x.x
    
tikv_servers:
  - host: 172.x.x.x
  - host: 172.x.x.x

cdc_servers:
  - host: 172.x.x.x
  - host: 172.x.x.x

monitoring_servers:
  - host: 172.x.x.x

grafana_servers:
  - host: 172.x.x.x

alertmanager_servers:
  - host: 172.x.x.x
```

**Test1** (EtcdWorker ticks limit 10 times / s)
Create **16** changefeeds, synchronize a table for each, and write 50w rows of data to each table in upstream.

<!DOCTYPE html><br class="Apple-interchange-newline">

  | Max checkpoint lag | Average checkpoint lag in peak | Max etcd qps | CPU usage | Total synchronization time
-- | -- | -- | -- | -- | --
owner | 3.3 min | 1.1min | 320 | 164% | 4 min
processor | - | - | 160 | 82% | -

**Test2** (EtcdWorker ticks limit 10 times / s)
Create **30** changefeeds, synchronize a table for each, and write 50w rows of data to each table in upstream.

<!DOCTYPE html><br class="Apple-interchange-newline">

  | Max checkpoint lag | Average checkpoint lag in peak | Max etcd qps | CPU usage | Total synchronization time
-- | -- | -- | -- | -- | --
owner | 6.8 min | 5 min | 600 | 189% | 10 min
processor | - | - | 300 | 95% | -

**Test3** (EtcdWorker ticks without limit )
Create **16** changefeeds, synchronize a table for each, and write 50w rows of data to each table in upstream.
<!DOCTYPE html><br class="Apple-interchange-newline">

  | Max checkpoint lag | Average checkpoint lag in peak | Max etcd qps | CPU usage | Total synchronization time
-- | -- | -- | -- | -- | --
owner | 2.74 min | 50 s | 630 | 193% | 3 min
processor | - | - | 260 | 92% | -

**Test4** (EtcdWorker ticks without limit )
Create **30** changefeeds, synchronize a table for each, and write 50w rows of data to each table in upstream.

<!DOCTYPE html><br class="Apple-interchange-newline">

  | Max checkpoint lag | Average checkpoint lag in peak | Max etcd qps | CPU usage | Total synchronization time
-- | -- | -- | -- | -- | --
owner | 6.9 min | 3 min | 1180 | 199% | 8 min
processor | - | - | 600 | 100% | -

**Test5** (Incremental scan、EtcdWorker ticks limit 10 times / s) 
1. Create 30 changefeed and pause.
2. 30 tables are created upstream, and 50w data is written into each table.
3. After waiting for the upstream data to be written, resume the above 30 changefeeds.

|           | Max checkpoint lag | Average checkpoint lag in peak | **Max etcd qps** | CPU usage | Total synchronization time |
| --------- | ------------------ | ------------------------------ | ---------------- | --------- | -------------------------- |
| owner     | 7 min              | 4 min                          | 600              | 400%      | 9 min                      |
| processor | -                  | -                              | 300              | 200%      | -                          |

**Test6** (Incremental scan、EtcdWorker ticks limit 10 times / s) 
1. Create 30 changefeed and pause.
2. 30 tables are created upstream, and 50w data is written into each table.
3. After waiting for the upstream data to be written, resume the above 30 changefeeds.

When resume changefeed, the owner was down instantly.

  | Max checkpoint lag | Average checkpoint lag in peak | Max etcd qps | CPU usage | Total synchronization time
-- | -- | -- | -- | -- | --
owner | 10 min | 7 min | 871 | 400% | 10 min
processor | - | - | 658 | 200% | -

Side effects
 - Has exported variable changed.

Related changes

### Release note 

```release-note
Please add a release note.
```


pkg/orchestrator/etcd_worker.go
```diff
@@ -120,7 +121,8 @@ func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session,
 		sessionDone = make(chan struct{})
 	}
 	lastReceivedEventTime := time.Now()
-
+	// limit the frequency of EtcdWorker ticks
+	rl := rate.NewLimiter(10, 30)
```
liuzix, line 125: Can you add some comments on the meaning of the numbers? And ideally they should be configurable.

asddongmen: don

maxshuang:
the current etcd qps will increase exponentially with the number of tables that need to be replicated.

----- @asddongmen  could u explain why?

maxshuang:
I still can't get the point of your tests.  "write 50w rows of data to each table in upstream" means a heavy continuous flow in upstream,  so it is reasonable to increase the etcdworker ticks to speed up.  

Am I right?   Or merge some resolvedTs or checkpointTs report will be a better solution? instead of just limit the rate.

liuzix:
> I still can't get the point of your tests. "write 50w rows of data to each table in upstream" means a heavy continuous flow in upstream, so it is reasonable to increase the etcdworker ticks to speed up.
> 
> Am I right? Or merge some resolvedTs or checkpointTs report will be a better solution? instead of just limit the rate.

From what I understand, the point of this PR is to address the problem that the frequency of ticks in one node must be greater than or equal to the frequency of updates by all nodes. This behavior is not optimal from a very low level.

If you mean batching updates for different changefeeds by "merge some resolvedTs or checkpointTs report", I think this will be a different optimization. The behavior of the business logic is orthogonal to what we are trying to fix here. We can investigate this and address it in another PR.

pkg/orchestrator/etcd_worker.go
```diff
@@ -121,6 +122,10 @@ func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session,
 	}
 	lastReceivedEventTime := time.Now()
 
+	// tickRate represents the number of times EtcdWorker can tick
+	// the reactor per second
+	tickRate := time.Second / timerInterval
+	rl := rate.NewLimiter(rate.Limit(tickRate), int(tickRate)*3)
```
liuzix, line 128: Could you explain the choice of `int(tickRate)*3`, and add a comment here?

overvenus (approved):

pkg/orchestrator/etcd_worker.go
```diff
@@ -184,6 +188,13 @@ func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session,
 			if err := worker.applyUpdates(); err != nil {
 				return errors.Trace(err)
 			}
+
+			// if !rl.Allow(), skip this Tick to avoid etcd worker tick reactor too frequency
+			// It make etcdWorker to batch etcd changed event in worker.state
```
liuzix, line 193:
```
The semantics of `ReactorState` requires that any implementation can batch updates internally.
```
Can you add this in comments?

liuzix (approved):

hi-rustin:
@asddongmen You don't seem to have filled out the release note correctly, so please fill out the release note for the next fix.

PR #3222, opened 2021-11-01
syncer(dm): support async checkpoint flush

glorv:
[design doc](https://github.com/pingcap/ticdc/blob/16a5038d9fd1d54189234177cb35bfb1284ab916/dm/docs/RFCS/20211012_async_checkpoint_flush.md)

- make checkpoint flush aync for dml triggered flush, leave other flush sync
  - When triggers a flush, first make a snapshot of the current checkpoint change, and then use the snapshot data to do the actual flush.
- change the `causality` map to make it compatible with async-flush

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Make periodically triggered checkpoint flush async and do not block DML sync.
```


dm/syncer/syncer.go
```diff
@@ -931,10 +940,15 @@ func (s *Syncer) addJob(job *job) error {
...
 		s.addCount(false, adminQueueName, job.tp, 1, job.targetTable)
 		s.jobWg.Add(1)
+		job.wg.Add(s.cfg.WorkerCount)
 		s.dmlJobCh <- job
-		s.jobWg.Wait()
-		s.addCount(true, adminQueueName, job.tp, 1, job.targetTable)
-		return s.flushCheckPoints()
+		var err error
+		if job.sync {
+			err = s.flushCheckPoints()
```
GMHDBJD, line 947: If my review is correct, job.sync means we should wait for all jobs flushed, but here s.flushCheckpoints will call `doFlushCheckPointsAsync` with empty waitgroup, how do we ensure that all jobs are flushed?

dm/syncer/syncer.go
```diff
@@ -402,6 +405,11 @@ func (s *Syncer) Init(ctx context.Context) (err error) {
 			}
 		}
 	}
+	s.flushCpWorker = &checkpointFlushWorker{
+		input:        make(chan *flushCpTask, 16),
```
Ehco1996, line 409: why chansize is 16

dm/syncer/syncer.go
```diff
@@ -1954,6 +2070,11 @@ func (s *Syncer) handleRotateEvent(ev *replication.RotateEvent, ec eventContext)
 	return nil
 }
 
+func (s *Syncer) getSeq() int64 {
```
Ehco1996, line 2073: please add some somments to explain how to use this `seq`

dm/syncer/syncer.go
```diff
@@ -963,8 +977,11 @@ func (s *Syncer) addJob(job *job) error {
 	})
 	if needFlush {
 		s.jobWg.Add(1)
-		s.dmlJobCh <- newFlushJob()
-		s.jobWg.Wait()
+		j := newFlushJob()
+		j.wg.Add(s.cfg.WorkerCount)
+		j.seq = s.getSeq()
```
Ehco1996, line 982: why incr job.seq by 1 again ? seems job.seq is already init in L931

dm/syncer/syncer.go
Ehco1996: +1

dm/syncer/syncer.go
```diff
@@ -931,10 +940,15 @@ func (s *Syncer) addJob(job *job) error {
 	case flush:
 		s.addCount(false, adminQueueName, job.tp, 1, job.targetTable)
 		s.jobWg.Add(1)
+		job.wg.Add(s.cfg.WorkerCount)
 		s.dmlJobCh <- job
-		s.jobWg.Wait()
```
Ehco1996, line 935: if we don't wait s.jobWg, seems no need to `s.jobWg.Add(1)` in L942?

dm/syncer/syncer.go
glorv: It's set by intuition. We just need to ensure that event the checkpoint flush is too slow, once there are too many flush jobs, it will back pressure the sync phase to avoid OOM.

dm/syncer/syncer.go
glorv: Other jobs like DDL still need to wait all previous done.

dm/syncer/syncer.go
```diff
@@ -1083,14 +1169,31 @@ func (s *Syncer) flushCheckPoints() error {
...
+		shardMetaSQLs: shardMetaSQLs,
+		shardMetaArgs: shardMetaArgs,
+		resChan:       outCh,
+		version:       version,
+	}
+	s.flushCpWorker.Add(task)
+}
 
+func (s *Syncer) afterFlushCheckpoint(task *flushCpTask) error {
+	s.dmlJobCh <- &job{
```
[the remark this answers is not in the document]
glorv: I think we don't need to statistics the gc metrics, it's just an inner implement logic for the `causality` module. And there operation is so fast, this kind of metrics is useless.

dm/syncer/syncer.go
```diff
@@ -1042,10 +1063,79 @@ func (s *Syncer) resetShardingGroup(table *filter.Table) {
...
+		}
+
+		err := w.execError.Load()
+		// TODO: for now, if any error occurred (including user canceled), checkpoint won't be updated. But if we have put
+		// optimistic shard info, DM-master may resolved the optimistic lock and let other worker execute DDL. So after this
+		// worker resume, it can not execute the DML/DDL in old binlog because of downstream table structure mismatching.
+		// We should find a way to (compensating) implement a transaction containing interaction with both etcd and SQL.
+		if err != nil && (terror.ErrDBExecuteFailed.Equal(err) || terror.ErrDBUnExpect.Equal(err)) {
+			ctx.L().Warn("error detected when executing SQL job, skip flush checkpoint",
+				zap.Stringer("checkpoint", s.checkpoint),
```
Ehco1996, line 1106:
```suggestion
				zap.Stringer("checkpoint", w.cp),
```

dm/syncer/syncer.go
```diff
@@ -2930,9 +3058,11 @@ func (s *Syncer) recordSkipSQLsLocation(ec *eventContext) error {
 }
 
 // flushJobs add a flush job and wait for all jobs finished.
+// NOTE: currently, flush job is alwasy sync operation.
```
Ehco1996, line 3061:
```suggestion
// NOTE: currently, flush job is always sync operation.
```

dm/syncer/syncer.go
```diff
@@ -1042,10 +1063,79 @@ func (s *Syncer) resetShardingGroup(table *filter.Table) {
...
+}
+
+// Add add a new flush checkpoint job.
+func (w *checkpointFlushWorker) Add(msg *flushCpTask) {
+	w.input <- msg
+}
+
+// Run read flush tasks from input and execute one by one
+func (w *checkpointFlushWorker) Run(ctx *tcontext.Context) {
+	for msg := range w.input {
```
Ehco1996, line 1093:
```suggestion
	for task := range w.input {
```
maybe use task is better for read here

dm/syncer/checkpoint.go
```diff
@@ -497,11 +583,23 @@ func (cp *RemoteCheckPoint) SaveGlobalPoint(location binlog.Location) {
...
+// FlushPointsExcept implements CheckPoint.FlushSnapshotPointsExcept.
+func (cp *RemoteCheckPoint) FlushPointsExcept(
+	tctx *tcontext.Context,
+	snapshot int,
+	exceptTables []*filter.Table,
+	extraSQLs []string,
+	extraArgs [][]interface{},
+) error {
 	cp.RLock()
 	defer cp.RUnlock()
```
Ehco1996, line 595: seems need user cp.Lock() beacue we update `cp.globalPointCheckOrSaveTime` in L676

dm/syncer/syncer.go
```diff
@@ -1054,18 +1144,19 @@ func (s *Syncer) resetShardingGroup(table *filter.Table) {
...
-			zap.Error(err))
-		return nil
-	}
+	errCh := make(chan error, 1)
+	// use math.MaxInt64 to clear all the data in causality component.
+	s.doFlushCheckPointsAsync(nil, errCh, math.MaxInt64)
+	return <-errCh
+}
+
+// flushCheckPointsAsync flushes previous saved checkpoint asyncronously after all worker finsh flush.
```
Ehco1996, line 1153:
```suggestion
// flushCheckPointsAsync flushes previous saved checkpoint asynchronously after all worker finish flush.
```

dm/syncer/syncer.go
```diff
@@ -932,9 +942,17 @@ func (s *Syncer) addJob(job *job) error {
...
-		s.jobWg.Wait()
-		s.addCount(true, adminQueueName, job.tp, 1, job.targetTable)
-		return s.flushCheckPoints()
+		var err error
+		// if flush is sync, should wait for all the previous job done,
+		// else don't block and let the flush async
+		if job.sync {
+			s.jobWg.Wait()
+			err = s.flushCheckPoints()
+		} else {
```
lance6716, line 951: seems addJob will only receive flushJob that sync is true

dm/syncer/syncer.go
lance6716: by `s.jobWg.Wait()` in line 949?

dm/syncer/syncer.go
```diff
@@ -963,8 +981,11 @@ func (s *Syncer) addJob(job *job) error {
 	})
 	if needFlush {
 		s.jobWg.Add(1)
-		s.dmlJobCh <- newFlushJob()
-		s.jobWg.Wait()
+		j := newFlushJob()
+		j.wg.Add(s.cfg.WorkerCount)
```
lance6716, line 985: DMLWorker.run will also `j.wg.Add(len(jobChs))`

dm/syncer/syncer.go
```diff
@@ -1042,10 +1063,79 @@ func (s *Syncer) resetShardingGroup(table *filter.Table) {
 	}
 }
 
+type flushCpTask struct {
+	snapshot SnapshotID
+	wg       *sync.WaitGroup
+	// job version, causality gc max version
+	version int64
```
lance6716, line 1070: (checking version is higher than any maxVer)

dm/syncer/syncer.go
```diff
@@ -1054,18 +1144,19 @@ func (s *Syncer) resetShardingGroup(table *filter.Table) {
 //
 // we may need to refactor the concurrency model to make the work-flow more clearer later.
 func (s *Syncer) flushCheckPoints() error {
```
lance6716, line 1146: it's no harm to put a duplicated s.wg.Wait() inside this function, to make some logic clear 😂

dm/syncer/causality.go
```diff
@@ -134,3 +135,79 @@ func (c *causality) detectConflict(keys []string) bool {
...
 	return false
 }
+
+type versionedMap struct {
+	data   map[string]string
+	maxVer int64
+}
+
+// rollingMap is a map this contains multi map instances.
+type rollingMap struct {
```
lonng, line 145: We need to add some unit tests for `rollingMap`.

dm/syncer/syncer.go
```diff
@@ -2930,9 +3058,11 @@ func (s *Syncer) recordSkipSQLsLocation(ec *eventContext) error {
 }
 
 // flushJobs add a flush job and wait for all jobs finished.
+// NOTE: currently, flush job is always sync operation.
 func (s *Syncer) flushJobs() error {
```
lance6716, line 3062: there's an invocation at line 1980, which is triggered by binlog heartbeat and last CheckOrSaveTime is longer than 30s by default. This flushing should also be async 🤔

lance6716 (commented):
seems if we want a synchronized flushing checkpoint, we should first wait the job waitgroup and then call `flushCheckPointsAsync` without waitgroup argument?

dm/syncer/syncer.go
db-will: for `flushCheckPoints` function, it will wait until errCh return, and this errCh will emit data out when `w.cp.FlushPointsExcept` return.
```
func (s *Syncer) flushCheckPoints() error {
	errCh := make(chan error, 1)
	// use math.MaxInt64 to clear all the data in causality component.
	s.doFlushCheckPointsAsync(nil, errCh, math.MaxInt64)
	return <-errCh
}
```

dm/syncer/syncer.go
db-will: basically, the existing DM behavoir is sync flush checkpoint when it receive heartbeat. basically, we are not changing the behavoir over here, and is there any case would require us to make such change here?

dm/syncer/dml_worker.go
```diff
@@ -172,8 +191,8 @@ func (w *DMLWorker) executeJobs(queueID int, jobCh chan *job) {
 		})
 
 		w.executeBatchJobs(queueID, jobs)
-		if j.tp == conflict || j.tp == flush {
-			w.wg.Done()
+		if j.tp == conflict || j.tp == flush || j.tp == asyncFlush {
```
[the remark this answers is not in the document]
db-will: decided keep it as what it is for better readability, with `j.wg != nil`, it's hard to tell why would we check this value here without further looking into `job.go` to find out which job may have such value and so we block it here.

dm/syncer/checkpoint.go
```diff
@@ -561,12 +677,14 @@ func (cp *RemoteCheckPoint) FlushPointsExcept(tctx *tcontext.Context, exceptTabl
 		return err
 	}
 
-	cp.globalPoint.flush()
+	cp.Lock()
```
lance6716, line 680: seems `binlogPoint` has its own lock so doesn't need the checkpoint lock.

And `needFlushSafeModeExitPoint` will be written by `checkpointFlushWorker.Run` and `SaveSafeModeExitPoint`, `SaveSafeModeExitPoint` is called in `Syncer.Run` so should be protected by checkpoint lock

lance6716 (commented):
will review soon

dm/syncer/causality.go
```diff
@@ -137,3 +140,81 @@ func (c *causality) detectConflict(keys []string) bool {
...
+	maps []*versionedMap
+	// current map fro write
+	cur *versionedMap
+}
+
+func newRollingMap() *rollingMap {
+	m := &rollingMap{
+		maps: make([]*versionedMap, 0),
```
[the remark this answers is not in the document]
db-will: Do you suggest to remove `m.rotate()` here? If so, I think it's used for creating a version map pointed by cur versionedMap. It's actually needed for `set`/`get` ect. @glorv feel free to add any comments here as well.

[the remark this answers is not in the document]
lance6716: No, I mean we keep `m.rotate()` but remove `maps: make([]*versionedMap, 0),`, which means we left `rollingMap.maps` as a nil slice.

lance6716 (commented):
will review later

dm/syncer/syncer.go
```diff
@@ -1051,15 +1065,100 @@ func (s *Syncer) resetShardingGroup(table *filter.Table) {
...
+
+// Run read flush tasks from input and execute one by one.
+func (w *checkpointFlushWorker) Run(ctx *tcontext.Context) {
+	for task := range w.input {
+		isAsyncFlush := task.asyncflushJob != nil
+
+		// for async checkpoint flush, it waits all worker finish execute flush job
+		// for sync checkpoint flush, it skips waiting here because
+		// we waits all worker finish execute flush job before adding flushCPTask into flush worker
+		if task.asyncflushJob != nil {
```
lance6716, line 1100:
```suggestion
		if isAsyncFlush {
```

lance6716 (commented):
rest lgtm

dm/syncer/checkpoint.go
```diff
@@ -517,34 +622,43 @@ func (cp *RemoteCheckPoint) FlushPointsExcept(tctx *tcontext.Context, exceptTabl
 	sqls := make([]string, 0, 100)
 	args := make([][]interface{}, 0, 100)
 
-	if cp.globalPoint.outOfDate() || cp.globalPointSaveTime.IsZero() || cp.needFlushSafeModeExitPoint {
-		locationG := cp.GlobalPoint()
+	if cp.globalPoint.outOfDateBy(snapshotCp.globalPoint.location) || cp.globalPointSaveTime.IsZero() || snapshotCp.needFlushSafeModeExitPoint {
+		locationG := snapshotCp.globalPoint.location
 		sqlG, argG := cp.genUpdateSQL(globalCpSchema, globalCpTable, locationG, cp.safeModeExitPoint, nil, true)
```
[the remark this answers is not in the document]
db-will: Plan to fix this issue, let me know if sounds good to you:
- use cp.needFlushSafeModeExitPoint instead of snapshotCp.needFlushSafeModeExitPoint here
- since flush safe mode exit point only happen for sync flush method, we only check if the current flush is sync flush
- avoid to pass `needFlushSafeModeExitPoint `, instead we only add global point as part of snapshot when it pass check in `Snapshot()` creation method

Basically, these actions could simplify the mental model of flushing safe mode exit point here. safe mode exit point will only be flushed synchronously, and it only happen sequentially, so the worst case is we flush it twice(still doubt if that could ever happen), and we are also safe to flush it twice as it's done that incrementally.

dm/syncer/syncer.go
```diff
@@ -961,18 +967,16 @@ func (s *Syncer) addJob(job *job) error {
 		})
 	}
 
-	// nolint:ifshort
-	needFlush := s.checkFlush()
 	failpoint.Inject("flushFirstJob", func() {
```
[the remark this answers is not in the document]
db-will: actually we still have sync flush, the async flush only applies for those flush happened on every flush time interval. But we still support sync flush used by ddl, heardbeat event. (can be found by search usage of `flushJobs()`)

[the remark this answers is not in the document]
lance6716: yes, sync flush is still used. 

The original bug is not about sync or async flush, it is caused by the time of triggering a flush: when we trigger a flush because of `CheckGlobalPoint` in the middle of a multi-row event, we may save a location forwarded too much. Now the real code logic of "trigger a flush in the middle of a multi-row event" use async flush, we should test async flush.

dm/syncer/checkpoint.go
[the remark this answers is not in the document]
lance6716: > since flush safe mode exit point only happen for sync flush method, we only check if the current flush is sync flush

our old logic abut "safe mode exit point" is when the replication progress pass the point, we are going to clear "safe mode exit point"  in downstream checkpoint. Now replication progress will fire two types of flush: sync flush by binlog heartbeat event and async flush when `CheckGlobalPoint && CheckLastSnapshotCreationTime`. I'm double checking that you mean we only let sync flush update "safe mode exit point"?

If so I think it's OK, since global checkpoint is forwarded, an outdated "safe mode exit point" should not be hurt.

> use cp.needFlushSafeModeExitPoint instead of snapshotCp.needFlushSafeModeExitPoint here ....

I agree we use current checkpint rather than snapshot checkpoint when dealing with "safe mode exit point".

dm/syncer/syncer.go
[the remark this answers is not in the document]
db-will: got it!

db-will:
@glorv looking for reviewing causality part as i refactored it only create seq for flush job instead of all job type

db-will:
@lance6716 @GMHDBJD looking for review/suggestions about metrics usage of current pr, and i feel there could be some issues with the metrics update mechanism.

dm/syncer/dml_worker.go
```diff
@@ -117,24 +115,43 @@ func (w *DMLWorker) run() {
 
 	for j := range w.inCh {
 		metrics.QueueSizeGauge.WithLabelValues(w.task, "dml_worker_input", w.source).Set(float64(len(w.inCh)))
-		if j.tp == flush || j.tp == conflict {
-			if j.tp == conflict {
-				w.addCountFunc(false, adminQueueName, j.tp, 1, j.targetTable)
+		switch j.tp {
+		case flush:
```
Ehco1996, line 119: how about merge aysnc/syncflush and conflict like this ? it seems that most of the codes are same for both of them

```golang
		case flush, asyncFlush,conflict:
			w.addCountFunc(false, adminQueueName, j.tp, 1, j.targetTable)
			// flush for every DML queue
			for i, jobCh := range jobChs {
				startTime := time.Now()
				jobCh <- j
				metrics.AddJobDurationHistogram.WithLabelValues(j.tp.String(), w.task, queueBucketMapping[i], w.source).Observe(time.Since(startTime).Seconds())
			}
			if j.tp == asyncFlush  || j.tp == conflict{
				j.wg.Wait()
			}
			w.addCountFunc(true, adminQueueName, j.tp, 1, j.targetTable)
			w.flushCh <- j
```

dm/syncer/syncer.go
```diff
@@ -217,6 +219,7 @@ type Syncer struct {
 	}
 
 	addJobFunc func(*job) error
+	seq        int64
```
Ehco1996, line 222: do we need reset `seq` in `syncer.reset()`

dm/syncer/syncer.go
```diff
@@ -413,6 +418,12 @@ func (s *Syncer) Init(ctx context.Context) (err error) {
 			}
 		}
 	}
+	s.flushCpWorker = &checkpointFlushWorker{
+		input:        nil, // will be created in s.reset()
```
Ehco1996, line 422:
```suggestion
		input:        nil, // will be created in s.newJobChans()
```
btw, how about move the logic of `flushCpWorker ` to another file? `syncer.go` is already too large 😂

Ehco1996 (commented):
will review checkpoint logic later

dm/syncer/causality.go
```diff
@@ -100,25 +110,20 @@ func (c *causality) add(keys []string) string {
...
 	// set causal relations for those non-exist keys
 	for _, key := range nonExistKeys {
-		c.relations[key] = selectedRelation
+		c.relations.set(key, selectedRelation)
 	}
 
 	return selectedRelation
 }
 
-// reset resets relations.
```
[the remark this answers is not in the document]
db-will: could refactor in the future if needed, since currently the extra reset() method won't be able to bring much extra value, and could cause a bit confusion as we sometimes call operation of rollingmap directly, but sometimes we call this reset() method...compared to `c.relations.clear`,  it adds extra unnecessary abstraction layer

dm/syncer/compactor.go
```diff
@@ -76,12 +76,17 @@ func (c *compactor) run() {
 			}
 			metrics.QueueSizeGauge.WithLabelValues(c.task, "compactor_input", c.source).Set(float64(len(c.inCh)))
 
-			if j.tp == flush {
+			if j.tp == flush || j.tp == asyncFlush {
```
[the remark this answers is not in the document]
db-will: we can add following up refactor pr here...just want to keep this PR clean, not containing other refactoring code, which could lead to an increasing size of PR, hard for others to review.

dm/syncer/dml_worker.go
db-will: instead of merging it together, i have separate out the flush dml queue logic out as separate function to call.

the issue of merging all together is causing us to hard build the mental model about each case execution and hence raise the probabilities to introduce bugs. As presented on above code, it's easily to miss the point that we don't need to do `w.flushCh <- j` for conflict type job.

db-will: btw, one more question here, how does this `w.addCountFunc(false, adminQueueName, j.tp, 1, j.targetTable)` work here? do we use it correctly here?

Ehco1996: >  it's easily to miss  the point that we don't need to do w.flushCh <- j for conflict type job.

sorry, mybad 😬

`addCountFunc` is used to update metrics, i think the current usage is mostly correct , but there is one point may  bother the user `j.targetTable` will always be empty string when jobType in [flush/asyncflush/confilct]

Ehco1996 (commented):
GreatWork ! btw,do we need to add some new metric for Causality and Checkpoint ? and i think we need performance test for this new feature

dm/syncer/checkpoint.go
```diff
@@ -683,6 +817,13 @@ func (cp *RemoteCheckPoint) CheckGlobalPoint() bool {
 	return time.Since(cp.globalPointSaveTime) >= time.Duration(cp.cfg.CheckpointFlushInterval)*time.Second
 }
 
+// CheckLastSnapshotCreationTime implements CheckPoint.CheckLastSnapshotCreationTime.
+func (cp *RemoteCheckPoint) CheckLastSnapshotCreationTime() bool {
+	cp.RLock()
+	defer cp.RUnlock()
+	return time.Since(cp.lastSnapshotCreationTime) >= time.Duration(cp.cfg.CheckpointFlushInterval)*time.Second
```
GMHDBJD, line 824: After supporting async flush checkpoint, can we reduce the default value of CheckpointFlushInterval?

db-will: that sounds good, will do a further testing to see if such interval could impact perf

dm/syncer/syncer.go
lance6716: in fact it's not possible for int64 to overflow

dm/syncer/checkpoint.go
```diff
@@ -304,11 +351,64 @@ func NewRemoteCheckPoint(tctx *tcontext.Context, cfg *config.SubTaskConfig, id s
...
+			if point.outOfDate() {
+				tableCpSnapshots[tbl] = point.savedPoint
+			}
+		}
+		if len(tableCpSnapshots) > 0 {
+			tableCheckPoints[s] = tableCpSnapshots
+		}
+	}
+
+	flushGlobalPoint := cp.globalPoint.outOfDate() || cp.globalPointSaveTime.IsZero() || (isSyncFlush && cp.needFlushSafeModeExitPoint)
```
lance6716, line 383: can you explain why we need `isSyncFlush && cp.needFlushSafeModeExitPoint` rather than `cp.needFlushSafeModeExitPoint` 😵‍💫

dm/syncer/checkpoint.go
```diff
@@ -561,12 +685,14 @@ func (cp *RemoteCheckPoint) FlushPointsExcept(tctx *tcontext.Context, exceptTabl
...
+	if snapshotCp.globalPoint != nil {
+		cp.globalPoint.flushBy(*snapshotCp.globalPoint)
+		cp.globalPointSaveTime = snapshotCp.globalPointSaveTime
 	}
 
-	cp.globalPointSaveTime = time.Now()
+	for _, point := range points {
+		point.tableCp.flushBy(point.snapshotTableCP)
+	}
 	cp.needFlushSafeModeExitPoint = false
```
lance6716, line 696: golang's race detector will report `needFlushSafeModeExitPoint` is written by event loop of syncer 
https://github.com/pingcap/ticdc/blob/6760414c14168950fda9525e1f299adb3699a235/dm/syncer/syncer.go#L1806
and checkpointFlushWorker's goroutine (though a bool should be ok to race). We might use Atomic.Bool

lance6716 (commented):
will review metrics soon

dm/syncer/dml_worker.go
```diff
@@ -117,24 +115,24 @@ func (w *DMLWorker) run() {
...
-			} else {
-				w.flushCh <- j
-			}
-		} else {
+		switch j.tp {
+		case flush:
+			w.addCountFunc(false, adminQueueName, j.tp, 1, j.targetTable)
+			w.sendJobToAllDmlQueue(j, jobChs, queueBucketMapping)
+			j.wg.Wait()
+			w.addCountFunc(true, adminQueueName, j.tp, 1, j.targetTable)
```
lance6716, line 123: for sync flush jobs, from this metrics we see that on one point in Grafana there's a job in adminQueue, and on another point the job is finished. The interval of points in Grafana is at lease 5s, maybe that's not accurate enough for observing the blocking time of flush?

maybe you're expected it a histogram like `metrics.AddJobDurationHistogram` in line 141?

dm/syncer/dml_worker.go
```diff
@@ -117,24 +115,24 @@ func (w *DMLWorker) run() {
...
+			w.flushCh <- j
+		case asyncFlush:
+			w.addCountFunc(false, adminQueueName, j.tp, 1, j.targetTable)
+			w.sendJobToAllDmlQueue(j, jobChs, queueBucketMapping)
+			w.addCountFunc(true, adminQueueName, j.tp, 1, j.targetTable)
```
lance6716, line 128: because async doesn't need `j.wg.Wait()`, this is very quickly and should almost be same with `metrics.AddJobDurationHistogram`.

Also for DML jobs, the `w.addCountFunc(true, ...)` is postponed to Syncer.successFunc, so for async flush it's more consistent that we also `w.addCountFunc(true, ...)` when async flush is finished?

lance6716 (commented):
will review soon

dm/syncer/syncer.go
```diff
@@ -1056,15 +1072,15 @@ func (s *Syncer) resetShardingGroup(table *filter.Table) {
 	}
 }
 
-// flushCheckPoints flushes previous saved checkpoint in memory to persistent storage, like TiDB
-// we flush checkpoints in four cases:
+// flushCheckPoints synchronously flushes previous saved checkpoint in memory to persistent storage, like TiDB
+// we flush checkpoints in three cases:
```
lance6716, line 1076:
```suggestion
// we flush checkpoints in four cases:
```

lance6716 (commented):
rest lgtm

dm/syncer/causality.go
```diff
@@ -137,3 +142,72 @@ func (c *causality) detectConflict(keys []string) bool {
 
 	return false
 }
+
+type versionedMap struct {
+	data            map[string]string
+	prevFlushJobSeq int64
```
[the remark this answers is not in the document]
db-will: what about rename versionedMap itself? Since each `versionedMap` means a group of keys that committed after `prevFlushJobSeq`, i am thinking of a name like `DmlJobKeysGroup` or similar ones.

dm/syncer/checkpoint.go
db-will: yeah, as we discussed before, we only deal with SafeModeExitPoint for sync flush, while async flush only being used for routinely flush checkpoint. 

Quote from our prev discussion:
```
Plan to fix this issue, let me know if sounds good to you:

use cp.needFlushSafeModeExitPoint instead of snapshotCp.needFlushSafeModeExitPoint here
since flush safe mode exit point only happen for sync flush method, we only check if the current flush is sync flush
avoid to pass needFlushSafeModeExitPoint , instead we only add global point as part of snapshot when it pass check in Snapshot() creation method
Basically, these actions could simplify the mental model of flushing safe mode exit point here. safe mode exit point will only be flushed synchronously, and it only happen sequentially, so the worst case is we flush it twice(still doubt if that could ever happen), and we are also safe to flush it twice as it's done that incrementally.
```

dm/syncer/dml_worker.go
lance6716: > record the total number of conlict/sync flush job have been committed?

`metrics.AddedJobsTotal` and `metrics.FinishedJobsTotal`?

> why we need to call w.addCountFunc(false and then w.addCountFunc(true here?

seems by `metrics.FinishedJobsTotal - metrics.AddedJobsTotal` we can get the jobs in the queue.

And my concern is that not about the number, but about the time cost of flush/conflict jobs. We may record the time and report to metrics like https://github.com/pingcap/ticdc/blob/f0a38a7aaf9f3b11a4d807da275b567642733f58/dm/syncer/syncer.go#L841-L842

We can do it in next PR.

dm/syncer/causality.go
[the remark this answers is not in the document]
glorv: lgtm

dm/syncer/dml_worker.go
db-will: if i understand correctly, we would need to add extra metrics to recording the time spent on flush/conflict job here. And we don't need to refactor/change the current mechanism of recording total number of jobs.

lance6716 (approved):

glorv (commented):
LGTM

dm/syncer/checkpoint_flush_worker.go
```diff
@@ -0,0 +1,116 @@
...
+	addCountFunc func(bool, string, opType, int64, *filter.Table)
+}
+
+// Add add a new flush checkpoint job.
+func (w *checkpointFlushWorker) Add(msg *checkpointFlushTask) {
+	w.input <- msg
+}
+
+// Run read flush tasks from input and execute one by one.
+func (w *checkpointFlushWorker) Run(ctx *tcontext.Context) {
```
GMHDBJD, line 53: ctx only use for log, how about use log.WithField like other components? https://github.com/pingcap/ticdc/blob/32032ca513ddb7c40831b590f9826dbea2d416a8/dm/syncer/dml_worker.go#L75

GMHDBJD (commented):
also https://github.com/pingcap/ticdc/pull/3222#discussion_r763613734
rest LGTM

dm/syncer/checkpoint_flush_worker.go
db-will: actually, ctx is needed here for `w.cp.FlushPointsExcept(ctx, task.snapshotInfo.id, task.exceptTables, task.shardMetaSQLs, task.shardMetaArgs)`

GMHDBJD (approved):
LGTM

PR #3241, opened 2021-11-02
dm/tracker: also ignore some safe error for tracker

lance6716:
close https://github.com/pingcap/dm/issues/2250

check if the error can be ignored, which is "should not exist and already not exist" or "should exist and exist and must be same"

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### Release note 

```release-note
Fix an error when ALTER TABLE DROP COLUMN and its INDEX in one SQL
```


dm/syncer/error.go
```diff
@@ -54,6 +54,19 @@ func ignoreDDLError(err error) bool {
 	}
 }
 
+// ignoreTrackerDDLError is also same with ignoreDDLError, but in order to keep tracker's table structure same as
+// upstream's, we can ignore "already exists" errors because already exists doesn't mean same.
+func ignoreTrackerDDLError(err error) bool {
+	switch {
+	case infoschema.ErrDatabaseExists.Equal(err), infoschema.ErrDatabaseDropExists.Equal(err),
```
Ehco1996, line 61: can you give a example to explain why we can ignore the `ErrDatabaseExists/ErrDatabaseDropExists/ErrTableExists` error?

i mean i can't understand this `we can ignore "already exists" errors because already exists doesn't mean same.`

does that mean when i want to sync 'create/drop database/table a' to downstream, we can countinue to sync regardless of whether the database/table existes or not?

lance6716: sorry i want to say “we can’t ignore“😂

Ehco1996 (approved):
LGTM

lichunzhu (approved):

PR #3245, opened 2021-11-02
p2p: Implement MessageServer

liuzix:
- Implement the server part of `pkg/p2p`.

- Implemented the server part of `pkg/p2p`.

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

### Release note 

```release-note
None
```


pkg/p2p/server.go
```diff
@@ -0,0 +1,845 @@
...
+		return errors.Trace(err)
+	}
+
+	metricsServerStreamCount := serverStreamCount.With(prometheus.Labels{
+		"from": packet.Meta.SenderAdvertisedAddr,
+	})
+	metricsServerStreamCount.Add(1)
+	defer metricsServerStreamCount.Sub(1)
+
+	ctx := context.WithValue(stream.Context(), streamMetaCtxKey, packet.Meta)
```
overvenus, line 556: Where is it used?

pkg/p2p/server.go
```diff
@@ -0,0 +1,845 @@
...
+	close(s.closeCh)
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.isClosed {
+		return
+	}
+	s.isClosed = true
+	close(s.sendCh)
```
overvenus, line 813: Maybe you should close(s.closeCh) after acquiring lock.

pkg/p2p/server.go
```diff
@@ -0,0 +1,845 @@
...
+	defer ticker.Stop()
+
+	for {
+		failpoint.Inject("ServerInjectTaskDelay", func() {
+			log.Info("channel size", zap.Int("len", len(m.taskQueue)))
+		})
+		select {
+		case <-ctx.Done():
+			return errors.Trace(ctx.Err())
+		case <-ticker.C:
```
overvenus, line 192: Could you use timer? Ticker may buffer a tick in its channel, as a result `m.tick()` may be invoked faster twice before next tick.

pkg/p2p/server.go
```diff
@@ -0,0 +1,845 @@
...
+		PeerID: senderID,
+		Epoch:  epoch,
+		sender: sender,
+		metricsAckCount: serverAckCount.With(prometheus.Labels{
+			"to": senderID,
+		}),
+	}
+}
+
+func (p *cdcPeer) abortWithError(ctx context.Context, err error) {
```
nongfushanquan, line 75: when this method will be called

pkg/p2p/server.go
```diff
@@ -0,0 +1,845 @@
...
+	}); err != nil {
+		return nil, nil, errors.Trace(err)
+	}
+
+	return doneCh, poolHandle.ErrCh(), nil
+}
+
+// MustRemoveHandler removes the registered handler for the given topic and wait
+// for the operation to complete.
+func (m *MessageServer) MustRemoveHandler(ctx context.Context, topic string) error {
```
nongfushanquan, line 400: why we need two methods?

pkg/p2p/server.go
liuzix: I think doing so may create a deadlock problem, because a sender might be blocked at writing to `sendCh` with `RLock` taken. Closing `closeCh` before taking the lock can force selects blocking on `sendCh` to break.

pkg/p2p/server.go
liuzix: `MustRemoveHandler` is implemented using `RemoveHandler`. The latter is an asynchronous version.

pkg/p2p/server.go
liuzix: It is called when a peer is forced to deregister by some error.

pkg/p2p/server.go
amyangfei: The drawback to use ticker may cause unbalanced trigger, but the frequency can't exceed the interval. Such as trigger interval is 4s, it won't trigger more than 2 times during 8s, but internal between two ticks is not deterministic

liuzix: I think using a ticker is enough for our purpose here.

pkg/p2p/server.go
```diff
@@ -0,0 +1,843 @@
...
+				zap.String("sender-id", sm.SenderId),
+				zap.String("topic", topic),
+				zap.Int64("skipped-Seq", entry.Sequence),
+				zap.Int64("last-ack", lastAck))
+			m.acksMapLock.Unlock()
+			return nil
+		}
+
+		if lastAck != 0 && entry.Sequence > lastAck+1 {
+			log.Panic("seq is skipped. Bug?", zap.Int64("last-ack", lastAck))
```
amyangfei, line 360: Can also log `entry.Sequence`.

amyangfei: Should we call `m.acksMapLock.Unlock()` before panic, or deadlock may happen.

pkg/p2p/server.go
```diff
@@ -0,0 +1,844 @@
...
+				log.Debug("handler registered", zap.String("topic", task.topic))
+			case taskOnDeregisterHandler:
+				if handler, ok := m.handlers[task.topic]; ok {
+					delete(m.handlers, task.topic)
+					go func() {
+						err := handler.GracefulUnregister(ctx, time.Second*5)
+						if err != nil {
+							log.L().DPanic("failed to gracefully unregister handle",
+								zap.Error(err))
+						}
```
amyangfei, line 217: In which scenario will `GracefulUnregister` return error, is the underlying work pool blocked? What will happen if don't panic here

pkg/p2p/server.go
```diff
@@ -0,0 +1,844 @@
...
+			if cerror.ErrPeerMessageInternalSenderClosed.Equal(err) {
+				peersToDeregister = append(peersToDeregister, peer)
+			}
+		}
+	}
+}
+
+func (m *MessageServer) deregisterPeer(ctx context.Context, peer *cdcPeer, err error) {
+	log.Info("Deregistering peer", zap.String("sender", peer.PeerID),
+		zap.Int64("epoch", peer.Epoch))
```
amyangfei, line 295: Do we need to record `err` here?

pkg/p2p/server.go
liuzix: `GracefulUnregister` will return error if the `ctx` is cancelled or the timeout has passed. `DPanic` here is for debugging and will not have an effect in production.

The underlying workerpool is not blocked in anyway.

amyangfei (approved):

pkg/p2p/server.go
```diff
@@ -0,0 +1,872 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/codes"
+	gRPCPeer "google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	waitUnregisterHandleTimeout = time.Second * 5
```
overvenus, line 40: Could you add some comments?

pkg/p2p/server.go
```diff
@@ -0,0 +1,872 @@
...
+)
+
+// MessageServerConfig stores configurations for the MessageServer
+type MessageServerConfig struct {
```
overvenus, line 44: Should it be in package `pkg/config`?

pkg/p2p/server.go
```diff
@@ -0,0 +1,872 @@
...
+	}
+}
+
+func (p *cdcPeer) abortWithError(ctx context.Context, err error) {
+	if err1 := p.sender.Send(ctx, errorToRPCResponse(err)); err1 != nil {
+		log.Warn("could not send error to peer", zap.Error(err),
+			zap.NamedError("send-err", err1))
+		return
+	}
+	log.Debug("sent error to peer", zap.Error(err))
```
overvenus, line 85: What about `Warn` or `Info`?

pkg/p2p/server.go
```diff
@@ -0,0 +1,872 @@
...
+		peer.abortWithError(ctx, err)
+	}
+}
+
+// MustAddHandler registers a handler for messages in a given topic and waits for the operation
+// to complete.
+func (m *MessageServer) MustAddHandler(
+	ctx context.Context,
+	topic string,
+	tpi interface{},
```
overvenus, line 330: What is tpi? Could you add some comments?

pkg/p2p/server.go
```diff
@@ -0,0 +1,872 @@
...
+	delete(m.peers, peer.PeerID)
+	m.peerLock.Unlock()
+	if err != nil {
+		peer.abortWithError(ctx, err)
+	}
+}
+
+// MustAddHandler registers a handler for messages in a given topic and waits for the operation
+// to complete.
+func (m *MessageServer) MustAddHandler(
```
overvenus, line 327:
```suggestion
func (m *MessageServer) AddHandler(
```

`Must` implies panic if AddHandler fails and should not return error.

pkg/p2p/server.go
```diff
@@ -0,0 +1,872 @@
...
+		return nil
+	})
+	errg.Go(func() error {
+		for {
+			select {
+			case <-ctx.Done():
+				return errors.Trace(ctx.Err())
+			case resp, ok := <-streamHandle.sendCh:
+				if !ok {
+					// cancel the stream when sendCh is closed
```
overvenus, line 612: Could you add an Info log here?

overvenus (commented):
What's unit tests coverage?

pkg/p2p/server.go
```diff
@@ -0,0 +1,872 @@
...
+				return errors.Trace(ctx.Err())
+			case streamHandle.sendCh <- errorToRPCResponse(err):
+			default:
+				log.Warn("sendCh congested, could not send error", zap.Error(err))
+				return errors.Trace(err)
+			}
+		}
+		return nil
+	})
+	errg.Go(func() error {
```
overvenus, line 605: Looks like this goroutine is unnecessary, we can run the for loop in `SendMessage`.

pkg/p2p/server.go
```diff
@@ -0,0 +1,872 @@
...
+	defer s.mu.RUnlock()
+
+	if s.isClosed {
+		return cerror.ErrPeerMessageInternalSenderClosed.GenWithStackByArgs()
+	}
+
+	select {
+	case <-ctx.Done():
+		return errors.Trace(ctx.Err())
+	case s.sendCh <- response:
```
overvenus, line 821: Maybe we should rate limited send frequency.

pkg/p2p/server.go
liuzix: If we did this, we would not be able to select on `m.closeCh`, which makes it impossible to terminate the stream properly if `stream.Send` is blocked by gRPC flow control.

pkg/p2p/server.go
liuzix: This is not a user facing config. It is intended for other pieces of code using this package.

pkg/p2p/server.go
liuzix: `deregisterPeer` already prints the `err`. This debug log here is for debugging.

pkg/p2p/server.go
liuzix: type information

nongfushanquan (approved):

PR #3249, opened 2021-11-02
cli: check sink uri as soon as possible.

3AceShowHand:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```


lance6716:
please complete PR descrption. I can't get why we move a check out of `validate`

pkg/cmd/cli/cli_changefeed_create.go
```diff
@@ -145,6 +145,10 @@ func (o *createChangefeedOptions) addFlags(cmd *cobra.Command) {
 
 // complete adapts from the command line args to the data and client required.
 func (o *createChangefeedOptions) complete(ctx context.Context, f factory.Factory, cmd *cobra.Command) error {
+	if o.commonChangefeedOptions.sinkURI == "" {
```
ben1009, line 148: why only move this one, seems have the same logic
```
func (o *createChangefeedOptions) validateTargetTs() error {
	if o.commonChangefeedOptions.targetTs > 0 && o.commonChangefeedOptions.targetTs <= o.startTs {
		return errors.Errorf("target-ts %d must be larger than start-ts: %d", o.commonChangefeedOptions.targetTs, o.startTs)
	}
	return nil
}
```

lonng: Please add some unit tests for it.

ben1009: ping @3AceShowHand no update for 7 days

PR #3259, opened 2021-11-04
compactor: fix performance issue caused by golang gc

GMHDBJD:
Previously, when we initialized the keymap of the compactor, we allocate buffersize memory space for it. When we set a large worker-count (that is, a large buffer-size) and the downstream pressure is small, the compactor will constantly flush the buffer, and golang gc will cause performance degradation

- do not pre-allocate memory for keymap
- add more metrics

### Check List 

Tests 

 - manually test

### Release note 

```release-note
`None`.
```


dm/syncer/compactor.go
```diff
@@ -61,48 +66,62 @@ func compactorWrap(inCh chan *job, syncer *Syncer) chan *job {
...
+			// if the number of outer jobs is zero or buffer is full, flush the buffer
+			if len(c.outCh) == 0 || len(c.buffer) >= c.bufferSize {
+				c.flushBuffer()
+			}
+			// if no inner jobs and the number of outer jobs is zero, flush the buffer
+		case <-time.After(waitTime):
+			failpoint.Inject("SkipFlushCompactor", func() {
+				failpoint.Continue()
+			})
+			if len(c.outCh) == 0 {
```
glorv, line 122: I think when there are no new job in waitTime, we should always do flush here.

glorv (commented):
rest LGTM

lance6716 (approved):

dm/syncer/compactor.go
GMHDBJD: The reason for changing this behavior is that if the downstream is very slow(the outer channel has jobs), but the job sent by upstream is not fast(maybe slow than 10ms?). At this time, if we still flush jobs, the compactor buffer will always be empty, which does not conform to the logic of compact jobs when the downstream is slow.

GMHDBJD: But if there is no job in the upstream 10ms, seems it doesn't matter if it is not compact?

Ehco1996: agree with @GMHDBJD 
just a question, when downstream is slow, we keep still flush when buffer is full and this will increase the pressure of downsteram, maybe we can incr buffer size here🤔

dm/syncer/syncer.go
```diff
@@ -1248,7 +1250,8 @@ func (s *Syncer) successFunc(queueID int, jobs []*job) {
 		s.addCount(true, queueBucket, sqlJob.tp, 1, sqlJob.targetTable)
 	}
 	s.updateReplicationJobTS(nil, dmlWorkerJobIdx(queueID))
-	metrics.ReplicationTransactionBatch.WithLabelValues(s.cfg.WorkerName, s.cfg.Name, s.cfg.SourceID, queueBucket).Observe(float64(len(jobs)))
+	metrics.ReplicationTransactionBatch.WithLabelValues(s.cfg.WorkerName, s.cfg.Name, s.cfg.SourceID, queueBucket, "statements").Observe(float64(statementsCnt))
```
Ehco1996, line 1253: statements means sql that really  exec to the downstream, rows means sql that come from binlog. am i right?
maybe need update the metrics doc https://docs.pingcap.com/zh/tidb-data-migration/stable/monitor-a-dm-cluster

Ehco1996 (approved):
LGTM

dm/syncer/compactor.go
GMHDBJD: May cause OOM, buffer size too big will also increase the time for stop-task.

dm/syncer/syncer.go
GMHDBJD: statements mean how many statements are in one transaction. rows mean how many rows changed in one transaction. I will update the doc later

dm/syncer/compactor.go
GMHDBJD: 07a89a7

lonng (commented):
Please add some tests.

PR #3260, opened 2021-11-04
dm/loader: support dump tls certs from dm-master

Ehco1996:
as title

dm-master stores the tls certificate data for the user task in its own etcd, so when the worker needs to establish a tls connection to the downstream, it does not need to have a certificate file on the current node

as many of the components lighting relies on (tidb,tikv-client,pd-clinet) do not support the init `tls.Config` directly by raw tls data, dm-worker downloads the  certificate file from dm-master and saves it to the specified path

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`support dump tls certs from dm-master`.
```


dm/dm/config/security.go
```diff
@@ -72,6 +74,26 @@ func (s *Security) LoadTLSContent() error {
 	return nil
 }
 
+// DumpTLSContent dump tls certs data to file.
+func (s *Security) DumpTLSContent() error {
+	if !utils.IsFileExists(s.SSLCA) {
```
lichunzhu, line 79: Can users specify `ssl-ca-bytes` while left `s.ssl-ca` empty?

lichunzhu (commented):
Rest LGTM

dm/dm/config/security.go
Ehco1996: yes, and dmctl use [this way](https://github.com/pingcap/ticdc/blob/b855baca0bea81e785841ccb19e615484a267abc/dm/dm/ctl/master/operate_source.go#L108) to transfer ssl-certs-data to dm-master

lichunzhu: Don't understand what you mention.. What will this function do if path `s.SSLCA` is invalid or empty?

Ehco1996: nice catch！i don't consider this case, i have two plan to slove this issue
A:  simply return err and this would cause task paused with error
B:  generate valid path for user and dump cert data to it

if you agree with plan A, no changes would be required
if you agree with plan B, i will add a test and fix in next commit

lichunzhu: I agree with plan B. It's hard for users to specify a right `SSLCA`, especially when sources transfer from workers to workers

Ehco1996: impl plan B in https://github.com/pingcap/ticdc/pull/3260/commits/ce791107df8cb1aaeea546c7a23747e082458f25, ptal 😬

dm/dm/worker/subtask.go
```diff
@@ -151,6 +151,14 @@ func NewSubTaskWithStage(cfg *config.SubTaskConfig, stage pb.Stage, etcdClient *
 
 // initUnits initializes the sub task processing units.
 func (st *SubTask) initUnits() error {
+	// NOTE: because lighting not support init tls with raw certs bytes, we write the certs data to a file.
```
lance6716, line 154:
```suggestion
	// NOTE: because lightning not support init tls with raw certs bytes, we write the certs data to a file.
```

and rest place

dm/dm/config/security.go
```diff
@@ -72,6 +75,36 @@ func (s *Security) LoadTLSContent() error {
...
+		if err := utils.WriteFileAtomic(s.SSLCA, s.SSLCABytes, 0o600); err != nil {
+			return err
+		}
+	}
+	if !utils.IsFileExists(s.SSLCert) {
+		if err := utils.WriteFileAtomic(s.SSLCert, s.SSLCertBytes, 0o600); err != nil {
+			return err
+		}
+	}
+	if !utils.IsFileExists(s.SSLKey) {
```
lance6716, line 100: overwriting old content is acceptable? maybe user change the cert content

dm/tests/tls/run.sh
```diff
@@ -73,6 +73,78 @@ function setup_mysql_tls() {
...
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT "$cur/conf/ca.pem" "$cur/conf/dm.pem" "$cur/conf/dm.key"
+
+	# operate mysql config to worker
+	run_dm_ctl_with_tls_and_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" $cur/conf/ca.pem $cur/conf/dm.pem $cur/conf/dm.key \
+		"operate-source create $WORK_DIR/source1.yaml" \
+		"\"result\": true" 2 \
+		"\"source\": \"$SOURCE_ID1\"" 1
+
+	echo "start task and check stage"
+	run_dm_ctl_with_tls_and_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" $cur/conf/ca.pem $cur/conf/dm.pem $cur/conf/dm.key \
```
lance6716, line 117: unstable test, because if worker runs quickly it will meet injected error

dm/tests/tls/run.sh
```diff
@@ -73,6 +73,78 @@ function setup_mysql_tls() {
...
+
+	# kill dm-worker 1 and clean the failpoint
+	export GO_FAILPOINTS=''
+	ps aux | grep dm-worker1 | awk '{print $2}' | xargs kill || true
+	check_port_offline $WORKER1_PORT 20
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $WORK_DIR/dm-worker1.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT "$cur/conf/ca.pem" "$cur/conf/dm.pem" "$cur/conf/dm.key"
+
+	# dm-worker will dump task-ca.pem from dm-master and save it to task-ca.pem
+	# let;s try use this file to connect dm-master
```
lance6716, line 138:
```suggestion
	# let's try use this file to connect dm-master
```

dm/tests/tls/run.sh
```diff
@@ -73,6 +73,78 @@ function setup_mysql_tls() {
...
+	mv "$cur/conf/task-ca.pem" "$cur/conf/task-ca.pem.bak"
+
+	# kill dm-worker 1 and clean the failpoint
+	export GO_FAILPOINTS=''
+	ps aux | grep dm-worker1 | awk '{print $2}' | xargs kill || true
+	check_port_offline $WORKER1_PORT 20
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $WORK_DIR/dm-worker1.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT "$cur/conf/ca.pem" "$cur/conf/dm.pem" "$cur/conf/dm.key"
+
+	# dm-worker will dump task-ca.pem from dm-master and save it to task-ca.pem
```
lance6716, line 137:
```suggestion
	# dm-worker will dump task-ca.pem from dm-master and save it to ca.pem
```

lichunzhu (approved):
LGTM.except for the existing comments

dm/tests/tls/run.sh
Ehco1996: it's task-ca.pm actually, because i delete old`task-ca.pem` in [here](https://github.com/Ehco1996/ticdc/blob/857bc5f1d2d33802ff30f03b113f8fdeb2e3b08e/dm/tests/tls/run.sh#L127)

dm/dm/config/security.go
Ehco1996: I don't think it's a good practice to overwrite a user's file, ex: if some cert stores in mysql's dir

dm/dm/worker/subtask.go
Ehco1996: fixed in https://github.com/pingcap/ticdc/pull/3260/commits/e35473de552e3413d1b8d69697d522d1f8bb62b6

dm/tests/tls/run.sh
Ehco1996: remove the check of `result: true` in https://github.com/pingcap/ticdc/pull/3260/commits/dea8ae5649fc9a6e2af263a644e379a81eba8904,
Is there a better way to handle this?

dm/dm/config/security.go
lance6716: MySQL's directory is located in upstream, this is DM-worker. BTW, we'll remove whole folder after finishing load unit.

dm/tests/tls/run.sh
lance6716: you have checked `query-status` below

dm/dm/config/security.go
Ehco1996: we can operate files in dm-worker's directory but i don't think user will put cert files to it ? so i perfer to if the file user configed in yaml is wrong, just return the error

lance6716: Now I think the behavior is that 
- if user specified the path for certs
   - if the cert exists, don't modify it
   - if the cert doesn't exist, we can overwrite certs to dump folder and change the cert path. Because the original user specified path maybe not suitable for this DM-worker
- if user didn't specify the path for certs, overwrite certs to dump folder and change the cert path.

And why can't we change lightning to pass cert content?

Ehco1996: > if the cert doesn't exist, we can overwrite certs to dump folder and change the cert path. Because the original user specified path maybe not suitable for this DM-worker

i will impl this in next commit, and all the other points mentioned above have been implemented in current code


> And why can't we change lightning to pass cert content?

as many of the components lighting relies on (tidb,tikv-client,pd-clinet) do not support the init tls.Config directly by raw tls data, dm-worker downloads the certificate file from dm-master and saves it to the specified path (already mentioned in PR description )

it's too hard  to change so many components 😢

lance6716 (approved):

dm/dm/config/security.go
Ehco1996: implemented in https://github.com/pingcap/ticdc/pull/3260/commits/a71915c9f0eac266d005c77a8a259da3addf14b2
ptal
 i will merge this pr if no more comments

lance6716:
lgtm

lance6716:
@Ehco1996 TLS test for lightnign still can't work, please open an issue as a TODO

https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/774/pipeline

Ehco1996:
> @Ehco1996 TLS test for lightning still can't work, please open an issue as a TODO
> 
> ```
> 2021-11-17T11:56:09.724Z] [2021/11/17 19:56:09.011 +08:00] [ERROR] [lightning.go:248] ["failed to runlightning"] [task=test] [unit=lightning-load] [error="invalid value / unknown config name: cluster"] [errorVerbose="invalid value / unknown config name: cluster\ngithub.com/pingcap/errors.AddStack\n\t/nfs/cache/mod/github.com/pingcap/errors@v0.11.5-0.20211009033009-93128226aaa3/errors.go:174\ngithub.com/pingcap/errors.Trace\n\t/nfs/cache/mod/github.com/pingcap/errors@v0.11.5-0.20211009033009-93128226aaa3/juju_adaptor.go:15\ngithub.com/pingcap/tidb/br/pkg/lightning/common.(*MySQLConnectParam).Connect\n\t/nfs/cache/mod/github.com/pingcap/tidb@v1.1.0-beta.0.20211115203106-b076e193b320/br/pkg/lightning/common/util.go:70\ngithub.com/pingcap/tidb/br/pkg/lightning/restore.DBFromConfig\n\t/nfs/cache/mod/github.com/pingcap/tidb@v1.1.0-beta.0.20211115203106-b076e193b320/br/pkg/lightning/restore/tidb.go:96\ngithub.com/pingcap/tidb/br/pkg/lightning.CleanupMetas\n\t/nfs/cache/mod/github.com/pingcap/tidb@v1.1.0-beta.0.20211115203106-b076e193b320/br/pkg/lightning/lightning.go:752\ngithub.com/pingcap/tidb/br/pkg/lightning.CheckpointRemove\n\t/nfs/cache/mod/github.com/pingcap/tidb@v1.1.0-beta.0.20211115203106-b076e193b320/br/pkg/lightning/lightning.go:739\ngithub.com/pingcap/ticdc/dm/loader.(*LightningLoader).restore\n\t/home/jenkins/agent/workspace/dm_ghpr_integration_test/go/src/github.com/pingcap/ticdc/dm/loader/lightning.go:240\ngithub.com/pingcap/ticdc/dm/loader.(*LightningLoader).Process\n\t/home/jenkins/agent/workspace/dm_ghpr_integration_test/go/src/github.com/pingcap/ticdc/dm/loader/lightning.go:292\ngithub.com/pingcap/ticdc/dm/loader.(*LightningLoader).Resume\n\t/home/jenkins/agent/workspace/dm_ghpr_integration_test/go/src/github.com/pingcap/ticdc/dm/loader/lightning.go:350\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1371"]
> ```
> 
> https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/774/pipeline

@lance6716  
i find it out, this error caused by [clean checkpoint](https://github.com/Little-Wallace/ticdc/blob/7c5fd31718b892bf60ec27d036ecbd7260c090d0/dm/loader/lightning.go#L221) not use the right mysql DSN, but the whole loder process is success so check_sync_diff is passed

i will fix this when https://github.com/pingcap/ticdc/pull/3220 merged
this pr is is too large, i don't want to push any new commit to it 😅

PR #3262, opened 2021-11-04
dm/loader/syncer: support extend column

yufan022:
What problem does this PR solve?
migrate from https://github.com/pingcap/dm/pull/2178
origin issue: https://github.com/pingcap/dm/issues/2111
relation pr: pingcap/tidb-tools#520

What is changed and how it works?
support extract table/schema/source names to specify column.

Load:
reuse reassemble function to rewrite sql.
https://github.com/yufan022/ticdc/blob/da32d489044dcbd1400a35a8c4e162802872b43a/dm/loader/convert_data.go#L185

Sync:
add extract values to rows before handle binlog event.
https://github.com/yufan022/ticdc/blob/da32d489044dcbd1400a35a8c4e162802872b43a/dm/syncer/syncer.go#L2032

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

Related changes

### Release note 

```release-note
Support extracting source/schema/table name to specific column
```


yufan022:
For integration tests, should I create a new folder in `dm/tests` or just add it to some existing folders.

Ehco1996:
> For integration tests, should I create a new folder in `dm/tests` or just add it to some existing folders.

i think this feature would be helpful in `downstream_more_column` [case](https://github.com/pingcap/ticdc/blob/master/dm/tests/downstream_more_column/run.sh) , so could please add a new sub test case here?

yufan022:
I added it to a new folder so that the logic would be clearer

dm/tests/extend_column/run.sh
```diff
@@ -0,0 +1,67 @@
...
+	run_sql_file $cur/data/db1.increment.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
+	run_sql_file $cur/data/db2.increment.sql $MYSQL_HOST2 $MYSQL_PORT2 $MYSQL_PASSWORD2
+
+	run_sql_tidb_with_retry "select count(1) from ${db}.${yb};" "count(1): 6"
+	run_sql_tidb_with_retry "select count(1) from ${db}.${tb};" "count(1): 6"
+	run_sql_tidb_with_retry "select count(1) from ${db}.${yb} where c1=3 and c_table='1';" "count(1): 1"
+	run_sql_tidb_with_retry "select count(1) from ${db}.${yb} where c1=3 and c_table='2';" "count(1): 1"
+	run_sql_tidb_with_retry "select count(1) from ${db}.${tb} where c1=3 and c_table='1' and c_schema='extend_column1' and c_source='replica01';" "count(1): 1"
+	run_sql_tidb_with_retry "select count(1) from ${db}.${tb} where c1=3 and c_table='2' and c_schema='extend_column2' and c_source='replica02';" "count(1): 1"
+
```
yufan022, line 57: `update/delete` also need check, improve later.

dm/tests/extend_column/run.sh
```diff
@@ -0,0 +1,67 @@
...
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
+
+	# operate mysql config to worker
+	cp $cur/conf/source1.yaml $WORK_DIR/source1.yaml
+	dmctl_operate_source create $WORK_DIR/source1.yaml $SOURCE_ID1
+	cp $cur/conf/source2.yaml $WORK_DIR/source2.yaml
+	dmctl_operate_source create $WORK_DIR/source2.yaml $SOURCE_ID2
+
+	# start DM task in all mode
+	# schemaTracker create table from dump data
```
Ehco1996, line 34: `schemaTracker create table from dump data` seems this is not needed

dm/tests/extend_column/conf/dm-task.yaml
```diff
@@ -0,0 +1,73 @@
+---
+name: test
+task-mode: all
+is-sharding: false
+meta-schema: "dm_meta"
+# enable-heartbeat: true
```
Ehco1996, line 6:
```suggestion

```
heartbeat is also deprecated, please also delete heartbeat-update-interval and heartbeat-report-interva

dm/syncer/syncer.go
```diff
@@ -2150,6 +2152,24 @@ func (qec *queryEventContext) String() string {
 		qec.ddlSchema, qec.originSQL, startLocation, currentLocation, lastLocation, shardingReSync, needHandleDDLs, strings.Join(trackInfos, ","))
 }
 
+// genrateExtendColumn generate extended columns by extractor.
+func genrateExtendColumn(data [][]interface{}, r *router.Table, table *filter.Table, source string) [][]interface{} {
```
Ehco1996, line 2156:
```suggestion
func genrateExtendColumn(data [][]interface{}, r *router.Table, table *filter.Table, sourceID string) [][]interface{} {
```

dm/syncer/syncer.go
```diff
@@ -2042,7 +2044,7 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
 	param := &genDMLParam{
 		targetTableID:   utils.GenTableID(targetTable),
 		data:            prunedRows,
-		originalData:    rows,
+		originalData:    extRows,
```
Ehco1996, line 2047: maybe we should keep using rows as `originalData`, and apply `extRows` to `prunedColumns`

dm/loader/loader.go
```diff
@@ -362,7 +362,7 @@ func (w *Worker) dispatchSQL(ctx context.Context, file string, offset int64, tab
 				continue
 			}
 
-			if w.loader.columnMapping != nil {
+			if w.loader.columnMapping != nil || len(table.extendCol) > 0 {
```
Ehco1996, line 365: why we need this check: `len(table.extendCol) > 0` ?

dm/loader/convert_data.go
```diff
@@ -278,7 +282,11 @@ func parseTable(ctx *tcontext.Context, r *router.Table, schema, table, file stri
 			columns = append(columns, col.Name.Name.O)
 		}
 	}
-	if hasGeneragedCols {
+	extendCol, extendVal := r.FetchExtendColumn(schema, table, sourceID)
```
Ehco1996, line 285: seems only need to `FetchExtendColumn` when  `r.TableExtractor !=nil && `rule.SchemaExtractor` !=nil && r.SourceExtractor != nil`, and this can reduce one re match for better performance

dm/syncer/syncer.go
```diff
@@ -2029,7 +2029,9 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
 		return err2
 	}
 
-	prunedColumns, prunedRows, err := pruneGeneratedColumnDML(tableInfo, rows)
+	extRows := genrateExtendColumn(rows, s.tableRouter, sourceTable, s.cfg.SourceID)
```
Ehco1996, line 2032: ditto, always `FetchExtendColumn` is not good for perf

Ehco1996 (commented):
thanks for you work

some tips:  
* you can use `make check` `make fmt` `make tidy` to fmt your code, so that CI would be pass, 
* use `go get -u github.com/pingcap/tidb-tools@master` to update `tidb-tools` version and this can make `r.FetchExtendColumn` really work
* new added `extend_column` it test could be added to `dm/tests/others_integration_2.txt` and this test will running in CI

yufan022:
```

[2021-11-08T04:07:04.141Z] go: downloading github.com/pingcap/tidb-tools v5.2.3-0.20211105044302-2dabb6641a6e+incompatible

[2021-11-08T04:07:19.035Z] # github.com/pingcap/tidb-tools/pkg/utils

[2021-11-08T04:07:19.035Z] open /nfs/cache/mod/github.com/pingcap/tidb-tools@v5.2.3-0.20211105044302-2dabb6641a6e+incompatible/pkg/utils/cpu_posix.go: no such file or directory

[2021-11-08T04:07:19.035Z] # github.com/pingcap/tidb-tools/pkg/etcd

[2021-11-08T04:07:19.035Z] open /nfs/cache/mod/github.com/pingcap/tidb-tools@v5.2.3-0.20211105044302-2dabb6641a6e+incompatible/pkg/etcd/etcd.go: no such file or directory

[2021-11-08T04:07:23.240Z] /usr/local/go/pkg/tool/linux_amd64/vet: chdir /nfs/cache/mod/github.com/pingcap/tidb-tools@v5.2.3-0.20211105044302-2dabb6641a6e+incompatible/pkg/table-filter: no such file or directory

[2021-11-08T04:07:23.240Z] /usr/local/go/pkg/tool/linux_amd64/vet: chdir /nfs/cache/mod/github.com/pingcap/tidb-tools@v5.2.3-0.20211105044302-2dabb6641a6e+incompatible/pkg/table-rule-selector: no such file or directory

[2021-11-08T04:07:23.240Z] /usr/local/go/pkg/tool/linux_amd64/vet: chdir /nfs/cache/mod/github.com/pingcap/tidb-tools@v5.2.3-0.20211105044302-2dabb6641a6e+incompatible/pkg/filter: no such file or directory
```

dm/loader/loader.go
yufan022: If the extend column exists, we need use `reassemble` rewrite SQL.

Ehco1996: ohoho i see, but there is another case we need to handle:
`` `
table.sourceTable != table.targetTable && len(table.extendCol) > 0
```

**Comment by Ehco1996** [comment]
> ```
> 
> [2021-11-08T04:07:04.141Z] go: downloading github.com/pingcap/tidb-tools v5.2.3-0.20211105044302-2dabb6641a6e+incompatible
> 
> [2021-11-08T04:07:19.035Z] # github.com/pingcap/tidb-tools/pkg/utils
> 
> [2021-11-08T04:07:19.035Z] open /nfs/cache/mod/github.com/pingcap/tidb-tools@v5.2.3-0.20211105044302-2dabb6641a6e+incompatible/pkg/utils/cpu_posix.go: no such file or directory
> 
> [2021-11-08T04:07:19.035Z] # github.com/pingcap/tidb-tools/pkg/etcd
> 
> [2021-11-08T04:07:19.035Z] open /nfs/cache/mod/github.com/pingcap/tidb-tools@v5.2.3-0.20211105044302-2dabb6641a6e+incompatible/pkg/etcd/etcd.go: no such file or directory
> 
> [2021-11-08T04:07:23.240Z] /usr/local/go/pkg/tool/linux_amd64/vet: chdir /nfs/cache/mod/github.com/pingcap/tidb-tools@v5.2.3-0.20211105044302-2dabb6641a6e+incompatible/pkg/table-filter: no such file or directory
> 
> [2021-11-08T04:07:23.240Z] /usr/local/go/pkg/tool/linux_amd64/vet: chdir /nfs/cache/mod/github.com/pingcap/tidb-tools@v5.2.3-0.20211105044302-2dabb6641a6e+incompatible/pkg/table-rule-selector: no such file or directory
> 
> [2021-11-08T04:07:23.240Z] /usr/local/go/pkg/tool/linux_amd64/vet: chdir /nfs/cache/mod/github.com/pingcap/tidb-tools@v5.2.3-0.20211105044302-2dabb6641a6e+incompatible/pkg/filter: no such file or directory
> ```

@yufan022  this may caused by nfs cache in nfs, some retry could help, build is successed in [here](https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/337/pipeline/53)

File: dm/syncer/syncer.go (hunk 5 above)
Comment (yufan022) [reply dm/syncer/syncer.go:2051]: The `update/delete` events will check that the lengths of `originalData` and `sourceTableInfo.Columns` are equal.
https://github.com/yufan022/ticdc/blob/dm_2111/dm/syncer/dml.go#L209

Since `sourceTableInfo.Columns` from the downstream table for first time, that will contains `table/source/schema` columns info.

File: dm/loader/convert_data.go (hunk 7 above)
Comment (yufan022) [reply dm/loader/convert_data.go:285]: Since `Extractor` belongs to `TableRule`, it must be matched to get `TableRule`.

File: dm/syncer/syncer.go (hunk 5 above)
Comment (Ehco1996) [reply dm/syncer/syncer.go:2051]: I still perfer to not change the originData because we should keep raw data from binlog not affected by `ExtendColumn`. so the logic of the check [here](https://github.com/yufan022/ticdc/blob/152ae257fd8fabde8caeaba47ef71e250b96a12d/dm/syncer/dml.go#L209) could be updated.

because https://github.com/pingcap/ticdc/pull/3168 is merged , i think we shoud check the error: `ErrSyncerUnitDMLColumnNotMatch` by campare input data `param.data` with `downstreamTableInfo.Columns` ?
ptal @lance6716 @GMHDBJD

File: dm/loader/loader.go (hunk 6 above)
Comment (yufan022) [reply dm/loader/loader.go:365]: I think we don't need handle it.

Since `reassemble` use `table.insertHeadStmt` generate SQL, it seems that `dstTable` is always the target table name. Or is there any other reason?

https://github.com/yufan022/ticdc/blob/dm_2111/dm/loader/convert_data.go#L318
https://github.com/yufan022/ticdc/blob/dm_2111/dm/loader/convert_data.go#L297

File: dm/syncer/syncer.go (hunk 5 above)
Comment (yufan022) [reply dm/syncer/syncer.go:2051]: I agree with you and there has another way: If `extCols > 0` then check `len(originData) != len(extCols)`.

Comment (Ehco1996) [reply dm/syncer/syncer.go:2051]: i can't understand why need to check `len(originData) != len(extCols)`, seems they are always different when there are extCols 😂, my point is just to add extrow to `prunedColumns` so that we can make sure `extrow` + `prunedColumns` have same column count with downstream

File: dm/loader/convert_data_test.go [hunk 9]
```diff
@@ -168,7 +168,7 @@ func (t *testConvertDataSuite) TestParseTable(c *C) {
 	r, err := router.NewTableRouter(false, rules)
 	c.Assert(err, IsNil)
 
-	tableInfo, err := parseTable(tcontext.Background(), r, "test1", "t2", "./dumpfile/test1.t2-schema.sql", "ANSI_QUOTES")
+	tableInfo, err := parseTable(tcontext.Background(), r, "test1", "t2", "./dumpfile/test1.t2-schema.sql", "ANSI_QUOTES", "")
```
Ehco1996, line 171:
```suggestion
	tableInfo, err := parseTable(tcontext.Background(), r, "test1", "t2", "./dumpfile/test1.t2-schema.sql", "ANSI_QUOTES", "mysql-01")
```
this is better for understand, and please fix this in rest test

Ehco1996 (commented):
LGTM when rest comments are resolved

dm/syncer/syncer.go
yufan022: sorry typo fault. check `if extCols > 0` then check `(len(originData) + len(extCols)) != len(ti.Columns) {`

yufan022: For update/delete scene, we need change `delete from a where id = 1` to `delete from a where id = 1 and table = x and source = x and schema = x`.

`where id = 1` comes from `originalData`, So if we can't modify `originData`, the situation will be a little complicated.

Ehco1996: > delete/update from source_table where id = 1 to delete from target_table where id = 1 and table = x and source = x and schema = x

LGTM

> if we can't append extCols to originalData, the situation will be a little complicated.

seems we only need do some changes in func `whereColumnsAndValues` just add extCols and extValues

yufan022: seems just need add extValues, because [sourceTableInfo](https://github.com/yufan022/ticdc/blob/2b46cf74d8dc8e32866170188241ab15555ba605/dm/syncer/syncer.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L2051) is from downStream, it always contains extendColumn info.

I think if the downstream table only has more extended columns than the upstream table, we don't need use `dmctl operate-schema set` to set table schema, just use the downstream [table](https://github.com/yufan022/ticdc/blob/e4da0cfc5d469b447f86abd2e3f94c65cd865919/dm/syncer/syncer.go#L715) directly. Is that Is correct?

Ehco1996: > seems just need add extValues

i don't think so, because the extend col may not  be an index column in downstream, and if it is, the genereated sql will  not  contain the extend cols in current [impl](https://github.com/pingcap/ticdc/blob/8814deda74767c7094294441c9c6517244dd824e/dm/syncer/dml.go#L631)

> I think if the downstream table only has more extended columns than the upstream table, we don't need use dmctl operate-schema set to set table schema, just use the downstream table directly. Is that correct?

yes, we now always use downstream table strucure to `gensql`

yufan022: > i don't think so, because the extend col may not be an index column in downstream, and if it is, the genereated sql will not contain the extend cols in current impl

usually downstream table will change the primary key to union primary key like:

source: `PRIMARY KEY (id)`
target: `PRIMARY KEY(id,source_table)`

In this scene, that [indexColumns.Columns](https://github.com/pingcap/ticdc/blob/8814deda74767c7094294441c9c6517244dd824e/dm/syncer/dml.go#L381) has `id, source_table`, [columns](https://github.com/pingcap/ticdc/blob/8814deda74767c7094294441c9c6517244dd824e/dm/syncer/dml.go#L382) has extend column, but [data](https://github.com/pingcap/ticdc/blob/8814deda74767c7094294441c9c6517244dd824e/dm/syncer/dml.go#L383) does not contains extend columns, just need add extVals to data.

source: `PRIMARY KEY (id)`
target: `does not contains extend column index or primary key`

In this scene, [columns](https://github.com/pingcap/ticdc/blob/8814deda74767c7094294441c9c6517244dd824e/dm/syncer/dml.go#L635) already contains extend column name, just need add extVals to [values](https://github.com/pingcap/ticdc/blob/8814deda74767c7094294441c9c6517244dd824e/dm/syncer/dml.go#L647).

In the above two scenarios, it seems that we only need to add extVals.

lance6716 (commented):
the discussion of https://github.com/pingcap/ticdc/pull/3262/files#r744353579 is a bit long, I'll list the unsettled problems here, please help check if I miss somthing

1. what's the meaning of "originalData"
2. how to try to eliminate the error of "column vs value mismatch"
3. what's the proper way to append `extCols` to WHERE clause

my opinion is 

1 & 2: the problem to me is there's currently one version of upstream table structure / upstream binlog values in DM, but we now have two versions of with and without extended columns. Should we also let DM maintain two versions, or we can discard one of the two versions? I prefer we can discuss it in a separate issue and implement the perfect version in later PRs. Would you like to lead this feature with our help? @yufan022 
3: currently WHERE clause tries to read downstream PK/UK to distinguish a row. If there exists a PK/UK, since it's unqiue we don't need to attach `extCols`. If there's no valid PK/UK which means all columns of PK/UK can be found in binlog, we should attach `extCols` to do a full column matching. Then it's related to the above question 😂

anyway, I think we can **keep this PR lightweight and merge it as long as it's self-consistent**. If this feature has no effects when it's not enabled, it's no harm for other user.

dm/syncer/syncer.go
Ehco1996: i agree with lance6716's comments here, i will resolve this conversation as the comments below have summed it up prety well

Ehco1996:
ping @yufan022

yufan022:
> 1 & 2: the problem to me is there's currently one version of upstream table structure / upstream binlog values in DM, but we now have two versions of with and without extended columns. Should we also let DM maintain two versions, or we can discard one of the two versions? I prefer we can discuss it in a separate issue and implement the perfect version in later PRs. Would you like to lead this feature with our help? @yufan022

I would be happy to work on improving this feature. 

> 3: currently WHERE clause tries to read downstream PK/UK to distinguish a row. If there exists a PK/UK, since it's unqiue we don't need to attach extCols. If there's no valid PK/UK which means all columns of PK/UK can be found in binlog, we should attach extCols to do a full column matching. Then it's related to the above question 😂

Yep.

The current table structure comes from the downstream pre-created, so it always comes with extended columns, my idea is that we just add `value` to ensure that it is equal to the `table.column`.

yufan022:
@Ehco1996 Hi, do we have to make sure that the value [here](https://github.com/yufan022/ticdc/blob/e4da0cfc5d469b447f86abd2e3f94c65cd865919/dm/syncer/dml.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L230) is the `originData`?

Ehco1996:
> @Ehco1996 Hi, do we have to make sure that the value is the `originData`?

you mean `originOldValues` or `originValues`? i don't think we need to make sure this , those fileds is used to distinguish a row and if extend clols can do this job, we need not to keep originData is the data from binlog

dm/loader/convert_data.go
```diff
@@ -278,7 +282,11 @@ func parseTable(ctx *tcontext.Context, r *router.Table, schema, table, file stri
...
+	if len(extendCol) > 0 {
+		columns = append(columns, extendCol...)
+	}
+	if hasGeneragedCols || len(extendCol) > 0 {
```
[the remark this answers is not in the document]
yufan022: Oh, It's not necessary.

Append extend column in history commits at [here](https://github.com/yufan022/dm/blob/f30c17fea76693a3315903ec2aba02fe93f6b930/loader/convert_data.go#L291), It has been moved to the top, I'll remove this condition.

Currently the extended column must be at the end of all columns and there are requirements for the order.

yufan022:
I made some changes to keep `originalData` field unchanged, But it still needs to be replaced in `genAndFilterXXXDMLs`. The key is to add the extended values before or after pushing the DMLjob to the [dmlJobCh](https://github.com/yufan022/ticdc/blob/e4da0cfc5d469b447f86abd2e3f94c65cd865919/dm/syncer/syncer.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L945).

If this is still not the recommended practice, I have to move the logic of appending extended columns into [genSQL](https://github.com/yufan022/ticdc/blob/e4da0cfc5d469b447f86abd2e3f94c65cd865919/dm/syncer/dml.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L724)/[whereColumnsAndValues](https://github.com/yufan022/ticdc/blob/e4da0cfc5d469b447f86abd2e3f94c65cd865919/dm/syncer/dml.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L606).

Ehco1996 (commented):
i think this pr is self-consistent now, please fix the conficts in go mod, let's enable test CI

dm/syncer/dml.go
```diff
@@ -94,6 +98,10 @@ func (s *Syncer) genAndFilterInsertDMLs(tctx *tcontext.Context, param *genDMLPar
 	}
 	downstreamIndexColumns := downstreamTableInfo.AbsoluteUKIndexInfo
 
+	if len(extendValues) > 0 {
+		originalDataSeq = extendData
```
lance6716, line 102: in line 114, `extractValueFromData` has assumed each row in `originalDataSeq` has same length with `ti.Columns`. But now the assumption is broken.

by a brief inspecting, that branch will be triggered when upstream table has a generated column. Also I think it's good to add an integration test to make sure this behaviour.

dm/syncer/dml.go
```diff
@@ -84,6 +86,8 @@ func (s *Syncer) genAndFilterInsertDMLs(tctx *tcontext.Context, param *genDMLPar
 		originalDataSeq = param.originalData
 		columns         = param.columns
 		ti              = param.sourceTableInfo
+		extendData      = param.extendData
+		extendValues    = param.extendValues
```
lance6716, line 90: seems we can remove `extendValues`, the usage can be changed to 

if extendData != nil {originalDataSeq = extendData}

dm/syncer/syncer.go
yufan022: Hi @Ehco1996 @lance6716 , why is this conversation marked as resolved?

About route:
  The logic of [router.FetchExtendColumn](https://github.com/pingcap/tidb-tools/blob/cef6f500d02daca6f6e420e1e6d3466dfb39ddbd/pkg/table-router/router.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L263) and [router.Route](https://github.com/pingcap/tidb-tools/blob/cef6f500d02daca6f6e420e1e6d3466dfb39ddbd/pkg/table-router/router.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L185) is very similar, but more return fields are required. `router.Route` has already been called once in phases [syncer](https://github.com/pingcap/ticdc/blob/e924c71914bf04505a2cd0f002d95d95059eff1f/dm/syncer/syncer.go#L2995) and [loader](https://github.com/pingcap/ticdc/blob/e924c71914bf04505a2cd0f002d95d95059eff1f/dm/loader/loader.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L1234). For performance reasons, I can add the extended column return value of `router.Route` or make a new func return `targetSchema,targetTable,extendCols,extendVals` so that no routing operations are added. If this is necessary.

About cache:
Adding a cache for A is a good idea, but this will result in the cached data not being updated in time after the upstream table migrates the instance, which may result in the downstream extended column data being wrong. If this situation is tolerable, I think adding caching would be very helpful for performance.
One solution is to restart the dm task after the upstream table migration instance?

Ehco1996: i think it's better to optimise performance in next or another PR, i am afraid that this pr will get so large that it will be difficult to merge, so i marked this conversation as resolved

abourt cache or func change, IMO, maybe we need some perf profiling to prove thoses  optimzations are really useful

dm/syncer/dml.go
yufan022: [`extendData != nil`](https://github.com/pingcap/ticdc/blob/43456cc356332066c4a65b0446b9507b381cfe8a/dm/syncer/dml.go#L102) and [`len(columns) != len(ti.Columns)`](https://github.com/pingcap/ticdc/blob/43456cc356332066c4a65b0446b9507b381cfe8a/dm/syncer/dml.go#L114)

You mean that we need to test the behavior of [extractValueFromData](https://github.com/pingcap/ticdc/blob/43456cc356332066c4a65b0446b9507b381cfe8a/dm/syncer/dml.go#L114) when these conditions are met?

Ehco1996:
it-test faild
```

[2021-11-19T00:27:35.433Z] dmctl test cmd: "start-task /home/jenkins/agent/workspace/dm_ghpr_integration_test/go/src/github.com/pingcap/ticdc/dm/tests/extend_column/conf/dm-task.yaml --remove-meta"

[2021-11-19T00:27:37.342Z] run tidb sql failed 1-th time, retry later

[2021-11-19T00:27:39.880Z] run tidb sql failed 1-th time, retry later

[2021-11-19T00:27:41.788Z] run tidb sql failed 2-th time, retry later

[2021-11-19T00:27:43.698Z] run tidb sql failed 3-th time, retry later

[2021-11-19T00:27:45.605Z] run tidb sql failed 4-th time, retry later

[2021-11-19T00:27:47.510Z] run tidb sql failed 5-th time, retry later

[2021-11-19T00:27:49.416Z] run tidb sql failed 6-th time, retry later

[2021-11-19T00:27:51.953Z] run tidb sql failed 7-th time, retry later

[2021-11-19T00:27:53.860Z] run tidb sql failed 8-th time, retry later

[2021-11-19T00:27:55.766Z] run tidb sql failed 9-th time, retry later

[2021-11-19T00:27:57.736Z] run tidb sql failed 10-th time, retry later

[2021-11-19T00:27:59.669Z] TEST FAILED: OUTPUT DOES NOT CONTAIN 'count(1): 6'

[2021-11-19T00:27:59.669Z] ____________________________________

[2021-11-19T00:27:59.669Z] [Fri Nov 19 08:27:57 CST 2021] Executing SQL: select count(1) from extend_column.y;

[2021-11-19T00:27:59.669Z] *************************** 1. row ***************************

[2021-11-19T00:27:59.669Z] count(1): 4

[2021-11-19T00:27:59.669Z] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ptal @yufan022

yufan022:
> extend_column it-test faild 😂
> 
> ```
> 
> [2021-11-19T00:27:35.433Z] dmctl test cmd: "start-task /home/jenkins/agent/workspace/dm_ghpr_integration_test/go/src/github.com/pingcap/ticdc/dm/tests/extend_column/conf/dm-task.yaml --remove-meta"
> 
> [2021-11-19T00:27:37.342Z] run tidb sql failed 1-th time, retry later
> 
> [2021-11-19T00:27:39.880Z] run tidb sql failed 1-th time, retry later
> 
> [2021-11-19T00:27:41.788Z] run tidb sql failed 2-th time, retry later
> 
> [2021-11-19T00:27:43.698Z] run tidb sql failed 3-th time, retry later
> 
> [2021-11-19T00:27:45.605Z] run tidb sql failed 4-th time, retry later
> 
> [2021-11-19T00:27:47.510Z] run tidb sql failed 5-th time, retry later
> 
> [2021-11-19T00:27:49.416Z] run tidb sql failed 6-th time, retry later
> 
> [2021-11-19T00:27:51.953Z] run tidb sql failed 7-th time, retry later
> 
> [2021-11-19T00:27:53.860Z] run tidb sql failed 8-th time, retry later
> 
> [2021-11-19T00:27:55.766Z] run tidb sql failed 9-th time, retry later
> 
> [2021-11-19T00:27:57.736Z] run tidb sql failed 10-th time, retry later
> 
> [2021-11-19T00:27:59.669Z] TEST FAILED: OUTPUT DOES NOT CONTAIN 'count(1): 6'
> 
> [2021-11-19T00:27:59.669Z] ____________________________________
> 
> [2021-11-19T00:27:59.669Z] [Fri Nov 19 08:27:57 CST 2021] Executing SQL: select count(1) from extend_column.y;
> 
> [2021-11-19T00:27:59.669Z] *************************** 1. row ***************************
> 
> [2021-11-19T00:27:59.669Z] count(1): 4
> 
> [2021-11-19T00:27:59.669Z] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> ptal @yufan022

It looks like [s.getTableInfo](https://github.com/pingcap/ticdc/blob/43456cc356332066c4a65b0446b9507b381cfe8a/dm/syncer/syncer.go#L2039) behavior is inconsistent after I merge the master branch.

yufan022:
It's caused by [PR:3295](https://github.com/pingcap/ticdc/pull/3295), starting the `load+increment` task for the first time will read the table structure from the dump file, but the dump file does not contain extension columns.

Before this PR, table structure always load from [downstream](https://github.com/pingcap/ticdc/blob/43456cc356332066c4a65b0446b9507b381cfe8a/dm/syncer/syncer.go#L722)

Ehco1996 (approved):
> The potential solution is to parse the dump file and skip it if the table has extended columns enabled.

this is lgtm and it test [ci](https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/825/pipeline) is passed now i think this pr is ready for merge 🎉

dm/syncer/dml.go
yufan022: @lance6716 @Ehco1996 One more unresolved conversation, please give me some guidance.

Ehco1996: as @lance6716 advice, i think just add a it-test to proof of correctness when met this condition 

* upstream table has a generated column
* downstream has extend colum

yufan022: Got it, let me see it.

GMHDBJD (commented):
I think 
```
create table in downstream(May be automated in the future)
fetch table from upstream
fill the columns and datas for exta columns
```
is better than
```
create table in downstream
fetch downstream table when task start
fill the data for ext rows
```

dm/syncer/dml.go
yufan022: Test generate and extend columns. https://github.com/pingcap/ticdc/pull/3262/commits/8036fc41bd8c85f390b9ecb10de34af6a324e3d4

dm/syncer/syncer.go
```diff
@@ -2040,14 +2040,20 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
...
 		return err2
 	}
 
+	extRows := generateExtendColumn(originRows, s.tableRouter, sourceTable, s.cfg.SourceID)
+	rows := originRows
+	if extRows != nil {
+		rows = extRows
+	}
+
 	prunedColumns, prunedRows, err := pruneGeneratedColumnDML(tableInfo, rows)
```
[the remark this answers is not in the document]
yufan022: Hi, since [ti.Columns](https://github.com/pingcap/ticdc/blob/27424e8bf420ad461268e38cd1edb6a47931319c/dm/syncer/dml.go#L459) contains extend columns, `extRows` and `ti.Columns` are the same length.

yufan022:
> fetch table from upstream

There are two ways to get the table structure from upstream in dm (if my understanding is correct):
1. If task model is `load+increment`, the upstream table structure will be loaded from the dump file at startup.
2. If task model is only `increment`, we need use `dmctl operate-schema set` to set upstream table structure when the upstream and downstream table structures are different.

In the case of the second, if we had thousands of upstream tables, we would need to call `dmctl operate-schema set` for each table, which would require too many operations. so, I used the downstream table structure directly.

Please point out if my understanding is incorrect.

dm/syncer/syncer.go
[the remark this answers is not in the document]
GMHDBJD: Ok, at that time I thought the table structure was init from dump file. But I still prefer the tableinfo should be same as upstream. cc @lance6716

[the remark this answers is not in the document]
yufan022: I left some comments [here](https://github.com/pingcap/ticdc/pull/3262#issuecomment-975290984) about why use the downstream table structure. PTAL

GMHDBJD (approved):
LGTM

Ehco1996:
@lance6716  do you have more suggestions? if not i will merge this PR

lance6716:
> @lance6716 do you have more suggestions? if not i will merge this PR

lgtm

PR #3277, opened 2021-11-04
etcd_worker: batch etcd patch

asddongmen:
This PR is an improvement to the issue's(#3112 ) fix.

1. Batch as many changefeed `patches` as possible into a batch and submit them as an etcd transaction, which can reduce the etcd qps.
~~2. Get the response from etcd `watchCh` as much as possible to reduce the number of etcd worker tick reactor.~~
3. Add metrics for some key points in EtcdWorker.

### Check List 

Tests 

 - Unit test 

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```


pkg/orchestrator/etcd_worker.go
```diff
@@ -283,14 +283,33 @@ func (worker *EtcdWorker) cloneRawState() map[util.EtcdKey][]byte {
 	return ret
 }
 
+var maxBatchPatchSize = 96
+
 func (worker *EtcdWorker) applyPatchGroups(ctx context.Context, patchGroups [][]DataPatch) ([][]DataPatch, error) {
+	getBatchPatches := func(patchGroups [][]DataPatch) ([]DataPatch, int) {
+		batchPatches := patchGroups[0]
```
ben1009, line 290: check len before get patchGroups[0] ?

pkg/orchestrator/etcd_worker.go
```diff
@@ -283,14 +283,33 @@ func (worker *EtcdWorker) cloneRawState() map[util.EtcdKey][]byte {
 	return ret
 }
 
+var maxBatchPatchSize = 96
+
 func (worker *EtcdWorker) applyPatchGroups(ctx context.Context, patchGroups [][]DataPatch) ([][]DataPatch, error) {
+	getBatchPatches := func(patchGroups [][]DataPatch) ([]DataPatch, int) {
```
ben1009, line 289: could the func be a named instead ?

pkg/orchestrator/etcd_worker.go
```diff
@@ -283,14 +283,33 @@ func (worker *EtcdWorker) cloneRawState() map[util.EtcdKey][]byte {
...
+			if (len(batchPatches) + len(patches)) > maxBatchPatchSize {
+				break
+			}
+			batchPatches = append(batchPatches, patches...)
+			patchesNum++
+		}
+
+		return batchPatches, patchesNum
+	}
+
```
ben1009, line 305: any ut cover the change ?

pkg/orchestrator/etcd_worker.go
```diff
@@ -283,14 +283,33 @@ func (worker *EtcdWorker) cloneRawState() map[util.EtcdKey][]byte {
...
 func (worker *EtcdWorker) applyPatchGroups(ctx context.Context, patchGroups [][]DataPatch) ([][]DataPatch, error) {
+	getBatchPatches := func(patchGroups [][]DataPatch) ([]DataPatch, int) {
+		batchPatches := patchGroups[0]
+		patchesNum := 1
+		for _, patches := range patchGroups {
+			if patchesNum == 1 {
+				continue
+			}
+			if (len(batchPatches) + len(patches)) > maxBatchPatchSize {
```
ben1009, line 296: what if patchGroups[0] already > max, need break down or just leave as it is.

pkg/orchestrator/etcd_worker.go
```diff
@@ -283,14 +283,33 @@ func (worker *EtcdWorker) cloneRawState() map[util.EtcdKey][]byte {
 	return ret
 }
 
+var maxBatchPatchSize = 96
+
 func (worker *EtcdWorker) applyPatchGroups(ctx context.Context, patchGroups [][]DataPatch) ([][]DataPatch, error) {
+	getBatchPatches := func(patchGroups [][]DataPatch) ([]DataPatch, int) {
+		batchPatches := patchGroups[0]
+		patchesNum := 1
+		for _, patches := range patchGroups {
```
ben1009, line 292: may be use index, the patchesNum could be index + 1 as result, let for loop do the validation

ben1009 (commented):
is this pr ready for review, if not ignore  the comments

pkg/orchestrator/etcd_worker.go
asddongmen: Leave as it is.

pkg/orchestrator/etcd_worker.go
asddongmen: i will add ut.

pkg/orchestrator/etcd_worker.go
asddongmen: sure

pkg/orchestrator/etcd_worker.go
asddongmen: In any condiction we should make sure a `patches` from `patchGroups` does not be splited.

pkg/orchestrator/etcd_worker.go
asddongmen: good advice~ thanks

pkg/orchestrator/etcd_worker.go
```diff
@@ -128,6 +128,7 @@ func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session,
 	rl := rate.NewLimiter(rate.Limit(tickRate), 1)
 	for {
 		var response clientv3.WatchResponse
```
overvenus, line 130:
```suggestion
```

pkg/orchestrator/etcd_worker.go
```diff
@@ -146,21 +147,46 @@ func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session,
 				return errors.Trace(err)
```
overvenus, line 147:
```suggestion
		case response := <-watchCh:
			// In this select case, we receive new events from Etcd, and call handleEvent if appropriate.
			if err := response.Err(); err != nil {
				return errors.Trace(err)
```

pkg/orchestrator/etcd_worker.go
asddongmen: the caller will do this check.

pkg/orchestrator/batch.go
```diff
@@ -0,0 +1,74 @@
...
+	"go.etcd.io/etcd/clientv3"
+)
+
+const (
+	maxBatchPatchSize    = 64
+	maxBatchResponseSize = 64
+)
+
+// getBatchPatches batch patches
+func getBatchPatches(patchGroups [][]DataPatch) ([]DataPatch, int) {
```
liuzix, line 27: We also need to count bytes here.  https://etcd.io/docs/v3.3/dev-guide/limit/#:~:text=etcd%20is%20designed%20to%20handle,bytes%20flag%20for%20etcd%20server.

pkg/orchestrator/batch.go
```diff
@@ -0,0 +1,73 @@
...
+	num := 0
+	totalSize := 0
+	// store changedState of multiple changefeed
+	batchChangedState := make(map[util.EtcdKey][]byte)
+	for i, patches := range patchGroups {
+		changeState, changedSize, err := getChangedState(state, patches)
+		if err != nil {
+			return nil, 0, 0, err
+		}
+		// if one changefeed
```
asddongmen, line 38: i will complete it lately.

pkg/orchestrator/etcd_worker.go
```diff
@@ -196,8 +213,11 @@ func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session,
 			if !rl.Allow() {
 				continue
 			}
+			startTime := time.Now()
 			// it is safe that a batch of updates has been applied to worker.state before worker.reactor.Tick
 			nextState, err := worker.reactor.Tick(ctx, worker.state)
+			costTime := time.Since(startTime).Seconds()
```
overvenus, line 219: Could you warn it takes more than 1 second?

pkg/orchestrator/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+
+package orchestrator
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	etcdTxnSize = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "ticdc",
+			Subsystem: "etcdWorker",
```
overvenus, line 22:
```suggestion
			Subsystem: "etcd_worker",
```

pkg/orchestrator/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+			Name:      "etcd_txn_size_bytes",
+			Help:      "Bucketed histogram of of etcd txn size.",
+			Buckets:   prometheus.ExponentialBuckets(1, 2, 18),
+		}, []string{"capture"})
+
+	etcdTxnExecDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "ticdc",
+			Subsystem: "etcdWorker",
```
overvenus, line 31:
```suggestion
			Subsystem: "etcd_worker",
```

pkg/orchestrator/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+			Name:      "etcd_txn_exec_duration",
+			Help:      "Bucketed histogram of processing time (s) of a etcd txn.",
+			Buckets:   prometheus.ExponentialBuckets(0.002 /* 2 ms */, 2, 18),
+		}, []string{"capture"})
+
+	etcdWorkerTickDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "ticdc",
+			Subsystem: "etcdWorker",
```
overvenus, line 40:
```suggestion
			Subsystem: "etcd_worker",
```

pkg/orchestrator/etcd_worker.go
```diff
@@ -344,7 +357,11 @@ func (worker *EtcdWorker) applyPatches(ctx context.Context, patches []DataPatch)
 		panic("unreachable")
 	}
 
+	worker.metrics.metricEtcdTxnSize.Observe(float64(size))
+	startTime := time.Now()
 	resp, err := worker.client.Txn(ctx).If(cmps...).Then(ops...).Commit()
+	costTime := time.Since(startTime).Seconds()
```
overvenus, line 363: Could you warn it takes more than 1 second?

pkg/orchestrator/batch.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/pingcap/ticdc/pkg/orchestrator/util"
+)
+
+const (
+	// 1.25 MiB
+	// Ref: https://etcd.io/docs/v3.3/dev-guide/limit/
+	etcdTxnMaxSize = 1024 * (1024 + 256)
+)
+
+func getBatchChangedState(state map[util.EtcdKey][]byte, patchGroups [][]DataPatch) (map[util.EtcdKey][]byte, int, int, error) {
```
liuzix, line 28: Can you add comments on the meaning of the return values?

liuzix (approved):

overvenus (approved):

pkg/orchestrator/batch.go
asddongmen: done

PR #3279, opened 2021-11-04
sorter/leveldb: add cleaner

overvenus:

Add cleaner to clean moved or stopped tables' data.

### Check List 

Tests 

 - Unit test

```
ok      github.com/pingcap/ticdc/cdc/sorter/leveldb     4.437s  coverage: 87.1% of statements
ok      github.com/pingcap/ticdc/cdc/sorter/leveldb/message     0.106s  coverage: 100.0% of statements
```

### Release note 

```release-note
None
```


pkg/actor/actor.go
```diff
@@ -56,7 +56,7 @@ type Mailbox interface {
 
 	// Try to receive a message.
 	// It must be nonblocking and should only be called by System.
-	tryReceive() (message.Message, bool)
+	Receive() (message.Message, bool)
```
sdojjy, line 59: Base on the comment, TryReceive is better?

overvenus: I'll update comment, as we have `Send` for nonblocking send, we should name `Receive` for nonblocking receive.

asddongmen (approved):

cdc/sorter/leveldb/cleaner.go
```diff
@@ -0,0 +1,202 @@
...
+
+	limiter *rate.Limiter
+	router  *actor.Router
+}
+
+var _ actor.Actor = (*CleanerActor)(nil)
+
+// NewCleanerActor returns a cleaner actor.
+func NewCleanerActor(
+	ctx context.Context, id int, db *leveldb.DB, router *actor.Router,
```
hi-rustin, line 49: It looks like we have some parameters that are not used. Are they reserved parameters?

cdc/sorter/leveldb/cleaner.go
```diff
@@ -0,0 +1,202 @@
...
+
+		start := encoding.EncodeTsKey(task.UID, task.TableID, 0)
+		limit := encoding.EncodeTsKey(task.UID, task.TableID+1, 0)
+		iterRange := &lutil.Range{
+			Start: start,
+			Limit: limit,
+		}
+		iter := clean.db.NewIterator(iterRange, nil)
+
+		// Force write the first batch if the task is rescheduled (rate limited).
```
hi-rustin, line 108:
```suggestion
		// Force writes the first batch if the task is rescheduled (rate limited).
```

cdc/sorter/leveldb/cleaner.go
```diff
@@ -0,0 +1,202 @@
...
+					rescheduleDelay = delay
+					break TASKS
+				}
+				batch.Reset()
+				force = false
+			}
+		}
+		// Release iterator in time.
+		iter.Release()
+		// Ignore rate limter and force write remaining kv.
```
hi-rustin, line 134:
```suggestion
		// Ignore rate limiter and force write remaining kv.
```

cdc/sorter/leveldb/cleaner.go
overvenus: No, nice catch!

hi-rustin: Do we need to keep the ctx?

hi-rustin:
@overvenus check failed.

cdc/sorter/leveldb/cleaner.go
overvenus: Addressed, thanks!

cdc/sorter/leveldb/cleaner.go
```diff
@@ -0,0 +1,202 @@
...
+	closedWg *sync.WaitGroup
+
+	limiter *rate.Limiter
+	router  *actor.Router
+}
+
+var _ actor.Actor = (*CleanerActor)(nil)
+
+// NewCleanerActor returns a cleaner actor.
+func NewCleanerActor(
```
amyangfei, line 48: Is leveldb sorter and cleaner shared by all changefeeds in one TiCDC process or each changefeed has a unique one?

hi-rustin (approved):

cdc/sorter/leveldb/cleaner.go
overvenus: They are shared by all changefeeds in one TiCDC process for easy resource management. And they can be shared for each changefeed if we want, no additional modification required.

amyangfei: If shared by multiple changefeeds, what if one changefeed trigger cleaner and cleans data that another changefeed will use?

overvenus: Data is isolated by a UID prefix.

PR #3284, opened 2021-11-04
kafka_consumer: remove useless kafka_consumer Dockerfile

hi-rustin:

part of https://github.com/pingcap/ticdc/issues/2199

remove useless kafka_consumer Dockerfile. We have reduced a root folder.
It is still using 1.13 of golang, which I believe is not used anywhere.
If you still want to keep it, I can move it to the right place.

### Check List 

Tests 

 - No code

Code changes

None
Side effects

None

Related changes

None

### Release note 

```release-note

None
```

### DIFF
--- a/kafka_consumer/Dockerfile
+++ b/kafka_consumer/Dockerfile
@@ -1,9 +1,0 @@
-FROM golang:1.13 as builder
-WORKDIR /go/src/github.com/pingcap/ticdc
-COPY . .
-RUN go mod download
-RUN make kafka_consumer
-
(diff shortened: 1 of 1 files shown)


asddongmen (approved):

amyangfei:
I remember that it was used in an old test (tipocket), while this test seems deprecated and no one is maintaining. I think it is OK to remove it. If anything is really broken by this PR, we can put it back.

amyangfei (approved):

PR #3295, opened 2021-11-05
dm: load table structure from dump files

lance6716:

if the task is full+incremental, we can use table structures from dump file to initiate schema tracker. Compared with using downstream table structure, that's better for the scenario when downstream has more columns.

this is done previously but canceled in https://github.com/pingcap/dm/pull/1915 , because at that time with this feature, generating WHERE clause uses upstream PK/UK which will cause data lost. Now we have merged https://github.com/pingcap/ticdc/pull/3168 so this feature can be turned on without correctness problem.

revert https://github.com/pingcap/dm/pull/1915

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`.
```


GMHDBJD (approved):
I forget that whether we have an integration test like `create table(id int auto increment, uk int unique)` and remove auto increment in downstream. If no, add it in this pr or later downstream PK/UK pr is OK

lance6716:
> I forget that whether we have an integration test like `create table(id int auto increment, uk int unique)` and remove auto increment in downstream. If no, add it in this pr or later downstream PK/UK pr is OK

not understand 🤔 do you mean we should check that we're using downstream PK/UK to generate DML? https://github.com/pingcap/ticdc/pull/3168 should cover the test.

and there's a test to ensure original AskTUG issue is fixed.

https://github.com/pingcap/ticdc/blob/df6eefc0833474559ae46ae367fccf59a7a1bb23/dm/tests/all_mode/run.sh#L47

GMHDBJD:
Just a suggestion about using a common case to test for both features. Can ignore.

lichunzhu (approved):

PR #3299, opened 2021-11-05
cdc/model: a protocol for owner/processor interaction

liuzix:
- We need to implement a protocol for Owner/Processor interaction used for peer-message based scheduling.

- Specified the topic names and message schema for table scheduling.

### Check List 

Tests 

 - Unit test

Code changes

### Release note 

```release-note
None
```


cdc/model/protocol.go
```diff
@@ -0,0 +1,104 @@
...
+	Removing []TableID
+}
+
+// Marshal serializes the message into MsgPack format.
+func (m *SyncMessage) Marshal() ([]byte, error) {
+	raw, err := msgpack.Marshal(m)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	return json.Marshal(raw)
```
amyangfei, line 80: Why is the message marshalled by both msgpack and json

liuzix: Previously I tried to override the JSON marshal methods. It was a mistake here. Fixed.

amyangfei (approved):

cdc/model/protocol.go
```diff
@@ -0,0 +1,98 @@
...
+func DispatchTableTopic(changefeedID ChangeFeedID) p2p.Topic {
+	return fmt.Sprintf("dispatch/%s", changefeedID)
+}
+
+// DispatchTableMessage is the message body for dispatching a table.
+type DispatchTableMessage struct {
+	OwnerRev   int64   `json:"owner-rev"`
+	ID         TableID `json:"id"`
+	IsDelete   bool    `json:"is-delete"`
+}
```
overvenus, line 37: Could you add some tests to lock the field names and types? It can prevent unexpected breaking changes, eg. changing type without rename json tag.

cdc/model/protocol.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"fmt"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/ticdc/pkg/p2p"
+	"github.com/vmihailenco/msgpack/v5"
+)
+
+// This file contains a communication protocol between the Owner and the Processor.
```
overvenus, line 24: As a communication protocol, does it need a versioning mechanism?

liuzix: added

cdc/model/protocol_test.go
```diff
@@ -0,0 +1,93 @@
...
+		Adding:   largeSliceFn(),
+		Removing: largeSliceFn(),
+	}
+}
+
+func TestMarshalDispatchTableMessage(t *testing.T) {
+	msg := &DispatchTableMessage{
+		OwnerRev: 1,
+        ID: TableID(1),
+        IsDelete: true,
```
overvenus, line 69: Please run go fmt.

overvenus (commented):
Rest LGTM

overvenus (approved):

PR #3309, opened 2021-11-05
[DNM]dm/syncer: move syncer's location logic to a separate member

lance6716:

a little refactor of syncer, now locations should be read from a separate member. Also rename and comment the locations to give less misleading information.

see `location_recorder.go`

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
 `None`.
```


lance6716:
will split to smaller PR

PR #2214, opened 2021-07-04
puller/sorter: remove manual GC settings

overvenus:
### What problem does this PR solve? 

Do not change default GC settings, it could reduce the minimum mutator utilization and lead to #2202.

Close #2202 

### Check List 

Tests 

 - Manual test (add detailed scripts or steps below)

Capturing 6k tables.

Related changes

 - Need to cherry-pick to the release branch

### Release note

```release-note
Fix gRPC keepalive error when memory pressure is high.
```

PR #3327, opened 2021-11-08
dm-relay: create binlog reader from relay, refactor relay

D3Hunter:
part 2 of subtask 4 in [#2214 of DM](https://github.com/pingcap/dm/issues/2214)

i ever thought to split this pr, but it's too late after fixing all the test cases.

- relay writer shouldn't be used concurrently and it's not a service, so remove unnecessary lock(except some status) and `stage` state
- remove unnecessary stage in binlog writer; reuse the same binlog writer in relay writer
- reuse relay writer in relay
- move `recover` logic from relay writer to relay
- remove unnecessary connection kill logic(already resolved in this [pr](https://github.com/go-mysql-org/go-mysql/pull/411), and current used version contains it)
- binlog reader created using relay; register listeners implicitly
- remove the transformer entity, make it a method in relay

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


dm/pkg/parser/common.go
```diff
@@ -335,3 +335,27 @@ func SplitDDL(stmt ast.StmtNode, schema string) (sqls []string, err error) {
 func genTableName(schema string, table string) *filter.Table {
 	return &filter.Table{Schema: schema, Name: table}
 }
+
+// CheckIsDDL checks input SQL whether is a valid DDL statement.
+func CheckIsDDL(sql string, p *parser.Parser) bool {
+	sql = utils.TrimCtrlChars(sql)
+
+	if utils.IsBuildInSkipDDL(sql) {
```
lance6716, line 343: maybe address this behaviour "the default skipped DDL will return `false`" in function comment.

dm/relay/binlog_writer.go
```diff
@@ -19,30 +19,27 @@ import (
...
 
-	"github.com/pingcap/ticdc/dm/pkg/binlog/common"
 	"github.com/pingcap/ticdc/dm/pkg/log"
 	"github.com/pingcap/ticdc/dm/pkg/terror"
 )
 
 // BinlogWriter is a binlog event writer which writes binlog events to a file.
 type BinlogWriter struct {
-	cfg *BinlogWriterConfig
+	mu sync.RWMutex
```
lance6716, line 32: seems `FileWriter.out` is only used in one goroutine and driven by upstream binlog event, so we even need no lock.

lance6716 (commented):
will review tests later

lichunzhu (commented):
will submit some comments later

lichunzhu (approved):
lgtm

lance6716 (commented):
rest lgtm, please address existing comments

dm/relay/binlog_writer.go
D3Hunter: "filename/offset" state'll be used in binlog reader later.

dm/pkg/parser/common.go
D3Hunter: maybe another pr

dm/relay/relay.go
```diff
@@ -123,17 +125,21 @@ type Relay struct {
...
 // NewRealRelay creates an instance of Relay.
 func NewRealRelay(cfg *Config) Process {
-	return &Relay{
+	r := &Relay{
 		cfg:       cfg,
 		meta:      NewLocalMeta(cfg.Flavor, cfg.RelayDir),
 		logger:    log.With(zap.String("component", "relay log")),
 		listeners: make(map[Listener]struct{}),
 	}
+	r.writer = NewFileWriter(r.logger)
```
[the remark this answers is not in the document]
D3Hunter: it's using `r.logger`

dm/relay/binlog_writer.go
```diff
@@ -19,30 +19,27 @@ import (
...
+	offset   atomic.Int64
+	file     *os.File
+	filename string
 
 	logger log.Logger
 }
 
 // BinlogWriterStatus represents the status of a BinlogWriter.
 type BinlogWriterStatus struct {
-	Stage    string `json:"stage"`
```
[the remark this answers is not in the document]
D3Hunter: it over complicated for a file write wrapper to have a `stage`

whether it's opened or not can be inferred from `filename`

D3Hunter:
@lance6716  PTAL

lance6716 (approved):

PR #3333, opened 2021-11-08
dm/openapi: use gin as http router instead of echo

Ehco1996:

as cdc and other projects choose to use gin as http router and https://github.com/deepmap/oapi-codegen/pull/475 was merged(oapi can genereate gin-server code from openapi spec), maybe is good to keep it same.

the most change is the code style between two web frameworks,
the biggest different is difficult to implement`error-handling`  in `gin` with middleware way：
`gin. HandlerFunc` not return error 😂 , so i left many `_  = c.Error(err)` in code to make lint pass

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`.
```


dm/dm/master/openapi.go
```diff
@@ -732,36 +802,37 @@ func (s *Server) DMAPIOperateTableStructure(ctx echo.Context, taskName string, s
...
-	var msg string
-	if tErr, ok := err.(*terror.Error); ok {
-		code = int(tErr.Code())
-		msg = tErr.Error()
-	} else {
-		msg = err.Error()
-	}
-	if sendErr := sendHTTPErrorResp(c, code, msg); sendErr != nil {
-		c.Logger().Error(sendErr)
+func terrorHTTPErrorHandler() gin.HandlerFunc {
```
Ehco1996, line 817: I am not confident in the implementation of this middleware... wait for another reviewe with better suggestion

asddongmen:
> /cc @asddongmen

i will review this pr in tomorrow.

dm/dm/master/openapi.go
asddongmen: I think it is good.

asddongmen (approved):

dm/dm/master/openapi.go
```diff
@@ -247,29 +260,39 @@ func (s *Server) DMAPIStartRelay(ctx echo.Context, sourceName string) error {
 	if needUpdate {
 		// update current source relay config before start relay
 		if err := s.scheduler.UpdateSourceCfg(sourceCfg); err != nil {
-			return err
+			_ = c.Error(err)
+			return
 		}
 	}
-	return s.scheduler.StartRelay(sourceName, req.WorkerNameList)
+	if err := s.scheduler.StartRelay(sourceName, req.WorkerNameList); err != nil {
```
lance6716, line 267: bad API :joy:  https://github.com/gin-gonic/gin/issues/2941

lance6716 (approved):

Makefile
```diff
@@ -362,49 +362,49 @@ else
 endif
 
 tools/bin/failpoint-ctl: tools/check/go.mod
-	cd tools/check && $(GO) build -o ../bin/failpoint-ctl github.com/pingcap/failpoint/failpoint-ctl
+	cd tools/check && $(GO) build -mod=mod -o ../bin/failpoint-ctl github.com/pingcap/failpoint/failpoint-ctl
```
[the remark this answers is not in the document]
Ehco1996: `- mod` flag is useful when upgrade pkg version in `tools`

we don't have any go file in tools dir, if i run `go mod tidy` go.mod will be a empty file 😅,do you have any other better way to upgrade those tools ?

PR #3355, opened 2021-11-09
Fix syncer dml worker

db-will:
Currently, the sync mechanism among ddl, dml and flush is broken, and it breaks previous sync mechanism after introducing dml-worker optimization.

for the following code snippet:
https://github.com/pingcap/ticdc/blob/master/dm/syncer/syncer.go#L929
```
	case flush:
		s.addCount(false, adminQueueName, job.tp, 1, job.targetTable)
		s.jobWg.Add(1)
		s.dmlJobCh <- job
		s.jobWg.Wait()
		s.addCount(true, adminQueueName, job.tp, 1, job.targetTable)
		return s.flushCheckPoints()
	case ddl:
		s.addCount(false, adminQueueName, job.tp, 1, job.targetTable)
		s.updateReplicationJobTS(job, ddlJobIdx)
		s.jobWg.Add(1)
		startTime := time.Now()
		s.ddlJobCh <- job
		metrics.AddJobDurationHistogram.WithLabelValues("ddl", s.cfg.Name, adminQueueName, s.cfg.SourceID).Observe(time.Since(startTime).Seconds())
		s.jobWg.Wait()
	case insert, update, del:
		s.dmlJobCh <- job
		s.isTransactionEnd = false
		failpoint.Inject("checkCheckpointInMiddleOfTransaction", func() {
			s.tctx.L().Info("receive dml job", zap.Any("dml job", job))
			time.Sleep(100 * time.Millisecond)
		})
```
and https://github.com/pingcap/ticdc/blob/master/dm/syncer/dml_worker.go#L118
```
	for j := range w.inCh {
		metrics.QueueSizeGauge.WithLabelValues(w.task, "dml_worker_input", w.source).Set(float64(len(w.inCh)))
		if j.tp == flush || j.tp == conflict {
			if j.tp == conflict {
				w.addCountFunc(false, adminQueueName, j.tp, 1, j.targetTable)
			}
			w.wg.Add(w.workerCount)
			// flush for every DML queue
			for i, jobCh := range jobChs {
				startTime := time.Now()
				jobCh <- j
				metrics.AddJobDurationHistogram.WithLabelValues(j.tp.String(), w.task, queueBucketMapping[i], w.source).Observe(time.Since(startTime).Seconds())
			}
			w.wg.Wait()
			if j.tp == conflict {
				w.addCountFunc(true, adminQueueName, j.tp, 1, j.targetTable)
			} else {
				w.flushCh <- j
			}
		} else {
			queueBucket := int(utils.GenHashKey(j.dml.key)) % w.workerCount
			w.addCountFunc(false, queueBucketMapping[queueBucket], j.tp, 1, j.targetTable)
			startTime := time.Now()
			w.logger.Debug("queue for key", zap.Int("queue", queueBucket), zap.String("key", j.dml.key))
			jobChs[queueBucket] <- j
			metrics.AddJobDurationHistogram.WithLabelValues(j.tp.String(), w.task, queueBucketMapping[queueBucket], w.source).Observe(time.Since(startTime).Seconds())
		}
	}
```

we can found out that:
- for ddl and dml, we are not waiting for all dml jobs finished when execute ddl
- for ddl and flush, flush could be still in progress but ddl already finished
- for dml and flush, we ensure it's not broken by using dml worker wg

By comparing with [previous code](https://github.com/pingcap/dm/blob/v2.0.7/syncer/syncer.go#L968), the missing point is: before running ddl, we should guarantee all dml/flush jobs are finished.

As what we do with [previous code](https://github.com/pingcap/dm/blob/v2.0.7/syncer/syncer.go#L968), we use one wait group to sync among ddl, dml, flush together. This will ensure their interactions haven't changed with dml worker optimization.

Besides, flushCh is removed from dml worker as it's not providing the right sync mechanism for us.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/dm/syncer/dml_worker.go
+++ b/dm/syncer/dml_worker.go
@@ -38,7 +38,7 @@ type DMLWorker struct {
 	multipleRows bool
 	toDBConns    []*dbconn.DBConn
 	tctx         *tcontext.Context
-	wg           sync.WaitGroup // counts conflict/flush jobs in all DML job channels.
+	wg           *sync.WaitGroup // counts conflict/flush jobs in all DML job channels.
 	logger       log.Logger
 
 	// for metrics
@@ -54,12 +54,11 @@ type DMLWorker struct {
 	addCountFunc func(bool, string, opType, int64, *filter.Table)
 
 	// channel
-	inCh    chan *job
-	flushCh chan *job
+	inCh chan *job
 }
 
 // dmlWorkerWrap creates and runs a dmlWorker instance and returns flush job channel.
-func dmlWorkerWrap(inCh chan *job, syncer *Syncer) chan *job {
+func dmlWorkerWrap(inCh chan *job, syncer *Syncer) {
 	chanSize := syncer.cfg.QueueSize / 2
 	if syncer.cfg.Compact {
 		chanSize /= 2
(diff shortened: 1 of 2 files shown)


GMHDBJD:
I think https://github.com/pingcap/ticdc/blob/8814deda74767c7094294441c9c6517244dd824e/dm/syncer/syncer.go#L2371 will guarantee that all jobs are flushed before ddl

PR #3366, opened 2021-11-10
sorter,mounter: flow control based on row size

### LINKED ISSUE #3364: Memory consumption of RowChangedEvent is larger than RawKVEntry, it leads to inaccurate flow control
### What did you do?

The goal of the flow control (from sorter to sink) is preventing OOM. In the current implementation, the flow control is based on `RawKVEntry.ApproximateSize()`, and it could cause OOM if there are many tables, even with a very small `per-table-memory-quota`.

In a test of capturing 10k tables, TiCDC OOM on a 64G server with `per-table-memory-quota = 307200` (300KB).

No OOM.

### What did you see instead?

OOM.

### Versions of the cluster

```console
v4.0.x - v5.2.x
```

overvenus:

Close #3364 

Also, it removes mounterNode, as row size only exists after it is mounted.

Flow control based on row size, instead of rawkv size.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

### Release note 

```release-note
None
```


liuzix:
Can you provide test results comparison on sink throughput?

overvenus:
> Can you provide test results comparison on sink throughput?

Updated, 1 cdc, 10 tables, blackhold sink, per-table-memory-quota 20MB, 300k qps.

cdc/entry/mounter.go
```diff
@@ -399,84 +408,108 @@ func (m *mounterImpl) mountRowKVEntry(tableInfo *model.TableInfo, row *rowKVEntr
...
-func formatColVal(datum types.Datum, tp byte) (value interface{}, warn string, err error) {
+const (
+	sizeOfEmptyBytes  = int(unsafe.Sizeof(emptyBytes))
+	sizeOfEmptyString = int(unsafe.Sizeof(""))
+	sizeOfEmptyDatum  = int(unsafe.Sizeof(types.Datum{}))
+)
+
+func sizeOfDatumValue(d types.Datum) int {
+	array := [...]types.Datum{d}
+	return int(types.EstimatedMemUsage(array[:], 1)) - sizeOfEmptyDatum
```
amyangfei, line 428: Is it accurate to put datum into an array, estimate the memory size and minutes `sizeOfEmptyDatum`

cdc/entry/mounter.go
```diff
@@ -399,84 +408,108 @@ func (m *mounterImpl) mountRowKVEntry(tableInfo *model.TableInfo, row *rowKVEntr
...
+		v := datum.GetMysqlJSON().String()
+		return v, len(v) + sizeOfEmptyString, "", nil
 	case mysql.TypeNewDecimal:
-		v := datum.GetMysqlDecimal()
-		if v == nil {
-			return nil, "", nil
+		d := datum.GetMysqlDecimal()
+		if d == nil {
+			return nil, 0, "", nil
 		}
```
amyangfei, line 451: if `d` is a typed nil, the memory size of it is not zero?

cdc/entry/mounter.go
```diff
@@ -399,84 +408,108 @@ func (m *mounterImpl) mountRowKVEntry(tableInfo *model.TableInfo, row *rowKVEntr
...
+) {
 	if datum.IsNull() {
-		return nil, "", nil
+		return nil, 0, "", nil
 	}
 	switch tp {
 	case mysql.TypeDate, mysql.TypeDatetime, mysql.TypeNewDate, mysql.TypeTimestamp:
-		return datum.GetMysqlTime().String(), "", nil
+		v := datum.GetMysqlTime().String()
+		return v, len(v) + sizeOfEmptyString, "", nil
```
amyangfei, line 440: Why adding `sizeOfEmptyString`

overvenus: The `string` struct also takes some bytes.

cdc/entry/mounter.go
overvenus: Yes, nil takes 0 byte.

liuzix (approved):

asddongmen (approved):

PR #3376, opened 2021-11-10
dm-test: fix unstable unit test

D3Hunter:
fix unstable unit test

#### reason: 
if ctx timeout(10ms) before this line, rotate event'll not be written and fail. this is hard to reproduce locally.
```go
_, err = r.handleEvents(ctx, reader2, transformer2, writer2) // returned when ctx timeout
```

**PS**: this comment is misleading, the event is not written as we set `writer2.err = errors.New("writer error for testing")`, so it's removed
```go
// after handle rotate event, we save and flush the meta immediately
```

solved by move creation of ctx before calling handleEvent, **although not solved completely, as timeout is unstable by nature, but since there're many cases depends on timeout or sleep, so will leave it**

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
`None`.
```


lance6716 (approved):

glorv (approved):

lance6716:
```
"msg": "[code=38032:class=dm-master:scope=internal:level=high], Message: some error occurs in dm-worker: ErrCode:11106 ErrClass:\"functional\" ErrScope:\"internal\" ErrLevel:\"high\" Message:\"fail to restart streamer controller: the given relay log pos (dm-integration-test-0s7gq-62dv6-bin.000002, 16782) of meta config is too big, please check it again\" Workaround:\"If the size of the corresponding binlog file has exceeded 4GB, please follow the solution in https://docs.pingcap.com/tidb-data-migration/stable/error-handling#the-relay-unit-throws-error-event-from--in--diff-from-passed-in-event--or-a-replication-task-is-interrupted-with-failing-to-get-or-parse-binlog-errors-like-get-binlog-error-error-1236-hy000-and-binlog-checksum-mismatch-data-may-be-corrupted-returned\" , Workaround: Please execute `query-status` to check status.",
```

not totally fixed 🤔

PR #3395, opened 2021-11-11
test: add test for needSwitchDB

hi-rustin:

See: https://github.com/pingcap/ticdc/pull/3264/files#r745243747

1. move to better place.
2. add test for it

### Check List 

Tests 

 - Unit test

Code changes

None

Side effects

None

Related changes

None

### Release note 

```release-note

None
```


hi-rustin:
/cc @lance6716 @amyangfei 

Could you please take a look? Thanks!

cdc/sink/mysql.go
```diff
@@ -364,6 +354,16 @@ func (s *mysqlSink) execDDL(ctx context.Context, ddl *model.DDLEvent) error {
 	return nil
 }
 
+func needSwitchDB(ddl *model.DDLEvent) bool {
```
lance6716, line 357: to reduce indent level, we can

if len == 0 {
return false
}
if ddl.Type == timodel.ActionCreateSchema || ddl.Type == timodel.ActionDropSchema {
return false
}
return true

also you can use `switch` to replace last `if`

hi-rustin: Done. Thanks!

lance6716 (approved):

Ehco1996 (approved):

hi-rustin:
/hold

Waiting for @amyangfei review.

amyangfei (approved):

PR #3405, opened 2021-11-11
dm: add retry in it full_mode test

Ehco1996:

as title, fix an unstable [test](https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/499/pipeline) 

as title

### Check List 

Tests 

 - Integration test

Code changes

- None

Side effects

- None

Related changes

- None

### Release note 

```release-note
`None`.
```


lance6716:
if sync_diff works well, we should not need this change

https://github.com/pingcap/ticdc/pull/3413 PTAL

PR #3417, opened 2021-11-11
compactor: fix data race

GMHDBJD:
datarace in CI https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/513/pipeline/

save job.tp before add dml job

### Check List 

Tests 

 - original test

### DIFF
--- a/dm/syncer/syncer.go
+++ b/dm/syncer/syncer.go
@@ -919,7 +919,11 @@ func (s *Syncer) addJob(job *job) error {
 		s.tctx.L().Info("All jobs is completed before syncer close, the coming job will be reject", zap.Any("job", job))
 		return nil
 	}
-	switch job.tp {
+
+	// avoid data race with compactor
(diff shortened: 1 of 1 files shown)


dm/syncer/syncer.go
```diff
@@ -919,7 +919,11 @@ func (s *Syncer) addJob(job *job) error {
 		s.tctx.L().Info("All jobs is completed before syncer close, the coming job will be reject", zap.Any("job", job))
 		return nil
 	}
-	switch job.tp {
+
+	// avoid data race with compactor
```
Ehco1996, line 923:
```suggestion
	// avoid job.type data race with compactor.run()
```

Ehco1996 (approved):
LGTM, little suggestion but keep it unchanged is ok

lance6716 (approved):

PR #3422, opened 2021-11-12
ticdc.rules: fix alertmanager rules

### LINKED ISSUE #3332: Alert cdc_checkpoint_high_delay could have false positive or false negative
### What did you do?

If the server time of alertmanager is different from TiCDC/PD nodes, some alert items rely on the server time difference, including
- cdc_checkpoint_high_delay: expr: (time() - ticdc_processor_checkpoint_ts / 1000) > 600
- cdc_resolvedts_high_delay: expr: (time() - ticdc_processor_resolved_ts / 1000) > 300

Conditions:
- If server time of alertmanager is much faster than TiCDC nodes, false positive could be fired.
- If server time of alertmanager is much slower than TiCDC nodes, false negative could be fired.

_No response_

### What did you see instead?

Can use `ticdc_processor_checkpoint_ts_lag` and `ticdc_processor_resolved_ts_lag` instead

### Versions of the cluster

```console
v5.0.1
```

```console
v5.0.1
```

asddongmen:
close #3332 

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes

### Release note 

```release-note
None
```


Ehco1996 (approved):
LGTM, btw do we already have a test for this metric ?

amyangfei (approved):

asddongmen:
> LGTM, btw do we already have a test for this metric ?

no, i will do this

PR #3432, opened 2021-11-12
compactor: fix duplicate entry in safemode

GMHDBJD:
in https://github.com/pingcap/dm/pull/2061, we change `replace` to `insert on duplicate key update`
but `insert on duplicate key update` will error when there is more than one key duplicate.
.e.g.
```
downstream: (1,1),(2,3)
insert (1,3) on duplicate key update will report duplicate key error
replace(1,3) will delete both rows and insert it
```

- use replace instead in safemode
   - insert(safemode) => replace
   - insert => insert
   - update(safemode) => delete + replace
   - update(update pk) => update
   - update(not update pk) => insert on duplicate key update
   - delete => delete

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


lance6716 (approved):

Ehco1996 (approved):
LGTM

lance6716:
/hold

to avoid /merge trigger run-all-tests

PR #3449, opened 2021-11-14
webui(dm): initial implementation

zoubingwu:

Added web ui for dm

- create source
- list all source

image

image

### Release note 

```release-note
Add a web user interface for dm
```


dm/ui/.eslintrc.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  extends: ['@ti-fe/eslint-config'],
```
Ehco1996, line 2: just a question: is this file is the lint rule for JS project ? why all rule is empty?

dm/ui/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules/
```
Ehco1996, line 1: gitignore should move to project root level

dm/ui/.husky/pre-commit
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+export PATH="$PATH:$(pwd)/dm/ui/node_modules/.bin"
+linter=$(pwd)/dm/ui/node_modules/.bin/lint-staged
+if test -f "$linter"; then
```
Ehco1996, line 6: lint-check should move into CI in case other develop not install pre-commit plugin locally

dm/ui/src/App.tsx
```diff
@@ -0,0 +1,66 @@
...
+    path: '/',
+    element: <Layout />,
+    children: [
+      {
+        index: true,
+        element: <div>TODO Dashboard</div>,
+      },
+      {
+        path: '/migration/task',
+        element: <div>TODO 任务列表</div>,
```
Ehco1996, line 21: i would like  all the Chinese to be translated into English (as this is an open-source project)

dm/ui/.eslintrc.js
zoubingwu: Yes, it extends from `@ti-fe/eslint-config`, which is a shared collection of rules used by our front-enders, you can also specify other rules here in this file or override it.

dm/ui/.husky/pre-commit
zoubingwu: I'll add a linter job for github action later, not intend to do too much in one single pull request

Ehco1996: no need to add to github action, just add it to `make-check` step in Makefile

dm/ui/src/models/source.ts
```diff
@@ -0,0 +1,22 @@
+export interface Source {
```
Ehco1996, line 1: maybe we can use openapi spec to generate those models 
https://swagger.io/tools/swagger-codegen/ the spec is [here](https://github.com/pingcap/ticdc/blob/master/dm/openapi/spec/dm.yaml)

dm/README.md
```diff
@@ -59,6 +59,24 @@ See the "Deploy" section in [our doc](https://docs.pingcap.com/tidb-data-migrati
 
 See the "Configuration" section in [our doc](https://docs.pingcap.com/tidb-data-migration/stable/config-overview)
 
+## Web UI
```
Ehco1996, line 62: maybe move those part to dm/ul/readme.md?

zoubingwu: sure

dm/ui/src/models/source.ts
zoubingwu: This is how tidb-dashboard does it if i recall correctely and I honestly don't think this is necessary here. It's slow, and requires java environment . And the generated code is a whole api client, making ui code heavy coupled with it and very difficult to reuse ui component else where.

zoubingwu: also it generate code based on a openapi format json, but the openapi spec itself lacks enough infomation to generate solid, type-safe typescript code

dm/ui/.husky/pre-commit
zoubingwu: @Ehco1996 is node and yarn preinstalled in CI images?

dm/ui/src/models/source.ts
Ehco1996: sgtm

dm/ui/.husky/pre-commit
Ehco1996: currently no,  but we can build/download linter bin to `bin/` dir like [golangci-lint](https://github.com/pingcap/ticdc/blob/1a285a478ddf058ede2d20c59f0850c0b6a1e35a/Makefile#L372)

Ehco1996:
@zoubingwu  ci failed https://github.com/pingcap/ticdc/runs/4428126477?check_suite_focus=true

Ehco1996 (approved):
LGTM

dm/ui/locales/en.json
```diff
@@ -0,0 +1,48 @@
+{
```
baurine, line 1: Using YAML may be more convenient, it supports comments and multiple lines, and easy multiple layers.

dm/ui/src/layouts/SiderMenu.tsx
```diff
@@ -0,0 +1,122 @@
...
+          itemKey: '/cluster/relay-log',
+        },
+      ],
+    },
+  ]
+  return (
+    <div>
+      <div className="flex p-4 justify-center">
+        <img
+          src="https://internals.tidb.io/uploads/default/original/1X/4fde143c268ccde514d0c93e420b8a2304fc2033.png"
```
baurine, line 66: This image can't be accessed inside a private network, it is better to put in local.

baurine:
lgtm

Ehco1996:
could you please take another look for this pr ? thanks 🧡🧡🧡 

/cc @lance6716 @lichunzhu @glorv

Makefile
```diff
@@ -240,6 +240,11 @@ dm: dm-master dm-worker dmctl dm-portal dm-syncer
 dm-master:
 	$(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/dm-master ./dm/cmd/dm-master
 
+dm-master-with-webui:
```
glorv, line 243: Shall we always embed webui into dm-master

glorv (commented):
LGTM

Makefile
Ehco1996: This is also an option, but introduces the problem that every time  build DM code you need to download the node dependencies and then `node build` (this is too slow and need adjust CI config)

so I feel like I should only do this when building a release package?  I see that `pd` does the same thing https://github.com/tikv/pd/blob/master/Makefile#L56

dm/ui/src/layouts/SiderMenu.tsx
zoubingwu: sure

dm/ui/locales/en.json
zoubingwu: A vite plugin is used to automatically scan all files and extract keys to json files during dev, I'll add yaml support later.

zoubingwu:
fixed mentioned issues, need one more approval 

cc @lance6716 @lichunzhu @baurine @glorv

baurine (approved):

lance6716 (commented):
rest lgtm

lance6716 (approved):

PR #3466, opened 2021-11-15
scheduler_v2: moveTableManager

liuzix:
- We need a module for implementing logic related to manual table moves.

- Implemented `moveTableManager"

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/scheduler/move_table_manager.go
```diff
@@ -0,0 +1,185 @@
...
+			job.status = moveTableJobStatusRemoved
+			continue
+		case removeTableResultGiveUp:
+			delete(m.moveTableJobs, tableID)
+			// Giving up means that we can move forward,
+			// so there is no need to return false here.
+			continue
+		case removeTableResultUnavailable:
+		}
+		return false, nil
```
nongfushanquan, line 145: how to handle the return false case

cdc/scheduler/move_table_manager.go
```diff
@@ -0,0 +1,185 @@
...
+func (m *moveTableManagerImpl) Add(tableID model.TableID, target model.CaptureID) bool {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if _, ok := m.moveTableJobs[tableID]; ok {
+		// Returns false if the table is already in a move table job.
+		return false
+	}
+
+	m.moveTableJobs[tableID] = &moveTableJob{
```
nongfushanquan, line 108: why not  use the struct like  map[tableID][{captureID,status}]

3AceShowHand (approved):

amyangfei (approved):

PR #3467, opened 2021-11-15
dm/syncer : fix wrong causality key when schema has prefix index

WizardXiao:
fix wrong causality key when use prefix index. https://github.com/pingcap/ticdc/issues/3286

Causality key will use prefix index to advoid dmls divide into different work group. For example.
`create table t (c1 text, c2 int unique, unique key c1(c1(3)));`
`insert into t values ('1234', 1)` --> key is `123.c1.`
`insert into t values ('1235', 2)` --> key is `123.c1.`

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`support calculate causality key with prefix index`.
```


dm/syncer/dml.go
```diff
@@ -683,14 +685,34 @@ func genKeyList(table string, columns []*model.ColumnInfo, dataSeq []interface{}
 	return buf.String()
 }
 
+// truncateIndexValues truncate prefix index from data.
+func truncateIndexValues(indexColumns *model.IndexInfo, tiColumns []*model.ColumnInfo, data []interface{}) []interface{} {
+	values := make([]interface{}, 0, len(indexColumns.Columns))
+	for i, iColumn := range indexColumns.Columns {
+		tcolumn := tiColumns[i]
+		if data[i] != nil && (tcolumn.Tp == mysql.TypeVarchar || tcolumn.Tp == mysql.TypeVarString ||
```
Ehco1996, line 693:
```suggestion
		if data[i] != nil && tcolumn.EvalType() == types.ETString
```

seems this is enough?

dm/syncer/dml.go
```diff
@@ -683,14 +685,34 @@ func genKeyList(table string, columns []*model.ColumnInfo, dataSeq []interface{}
...
 // genMultipleKeys gens keys with UNIQUE NOT NULL value.
 // if not UNIQUE NOT NULL value, use table name instead.
 func genMultipleKeys(downstreamTableInfo *schema.DownstreamTableInfo, ti *model.TableInfo, value []interface{}, table string) []string {
 	multipleKeys := make([]string, 0, len(downstreamTableInfo.AvailableUKIndexList))
 
 	for _, indexCols := range downstreamTableInfo.AvailableUKIndexList {
 		cols, vals := getColumnData(ti.Columns, indexCols, value)
-		key := genKeyList(table, cols, vals)
+		// handle prefix index
+		truncVals := truncateIndexValues(indexCols, cols, vals)
```
Ehco1996, line 714: do we need to do some extra logic fot other index type? such as  func_index https://dev.mysql.com/doc/refman/8.0/en/create-index.html#create-index-functional-key-parts

dm/syncer/dml.go
WizardXiao: `EvalType()` includes several additional cases, like `TypeUnspecified`,`TypeNull`,`TypeNewDate`,`TypeGeometry`. 
I'm afraid it may has  problem.

dm/syncer/dml.go
WizardXiao: I will test func_index later, but I'm afraid `TruncateIndexValue` does not support.

dm/syncer/dml.go
Ehco1996: i found `TruncateIndexValue` alreadt do precheck in 
https://github.com/pingcap/tidb/blob/fde18397028a3ab30f38ab4ef964b6ef8e0adf7a/tablecodec/tablecodec.go#L1290

so we can just simply remove those check?

WizardXiao: yes, `TruncateIndexValue` will check too. i was intended to do this check to avoid to convert all kinds data to `Datum`, because the prefix index does not  use requently. 
as `Datum` can conveniently convert to data `interface{}`, this check is seems redundant just like your comment, so i will remove this check.

dm/syncer/dml.go
WizardXiao: it seems we do not support function index.
scheam: `CREATE TABLE t1 (col1 INT, col2 INT, unique key func_index ((abs(col1))));`
the function index will be a column `_V$_func_index_0` with offset `2`, and we will have a wrong uk which column offset is wrong.

Ehco1996: we can solve this in another pr, so we need keep https://github.com/pingcap/ticdc/issues/3286 open

Ehco1996 (approved):

dm/syncer/dml.go
WizardXiao: yes, it seems that function index is different from prefix index, and we will support it in another pr.

lichunzhu (approved):

PR #3474, opened 2021-11-16
dm/scheduler: fix inconsistent of relay status

### LINKED ISSUE #3478: scheduler failed to keep consistent relay status if master/worker restart in particular order
### What did you do?

https://github.com/pingcap/ticdc/blob/bc6029e22fbf38612184765ee802431663e4fa10/dm/tests/new_relay/run.sh#L86

prerequisite
1. `start-relay -s source1 worker1`
2.  kill worker1 (don't use kill -9)
3. restart master
4. start worker1

now DM master lost the relay status of that worker in memory, in other words, it will treat that worker as a free worker. but the worker is indeed pulling relay logs.

the inconsistency will become a problem when master tries to bound another source to the worker. At then worker will report same error like https://github.com/pingcap/dm/issues/2204

above integration test will pass, which means after master and worker restarts, two worker status of source2 is bound + relay

### What did you see instead?

two worker status of source2 is bound + free

### Versions of the cluster

```console
master (5.3.0)
```

```console
```

lance6716:

close #3478 

when receive keepalive, get relay source from correct place

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

- need cherry-pick to release

### Release note 

```release-note 
 Fix a bug that when master and worker restart in a particular order, relay status in DM-master is wrong
```


lance6716:
cc @lichunzhu

lichunzhu (approved):

GMHDBJD (approved):
LGTM

PR #3475, opened 2021-11-16
cdc/sorter: add non-blocking function TryAddEntry

sdojjy:
Add an new function named TryAddEntry into sorter interface, this function will not block the caller

For unifiedSorter,  add an default select case,
For entry sorter,  just call AddEntry function, AddEntry append the entry to a memory list, it should be non blocking

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
 `None`.
```


cdc/sorter/sorter.go
```diff
@@ -24,5 +24,7 @@ import (
 type EventSorter interface {
 	Run(ctx context.Context) error
 	AddEntry(ctx context.Context, entry *model.PolymorphicEvent)
+	// TryAddEntry try to add entry to the sorter, returns false if the entry can not be added, otherwise returns true
```
overvenus, line 27: Please keep line width within 80 chars.

cdc/sorter/unified/unified_sorter.go
```diff
@@ -281,6 +281,23 @@ func (s *Sorter) AddEntry(ctx context.Context, entry *model.PolymorphicEvent) {
...
+// TryAddEntry implements the EventSorter interface
+func (s *Sorter) TryAddEntry(ctx context.Context, entry *model.PolymorphicEvent) bool {
+	select {
+	case <-ctx.Done():
+		return false
+	case <-s.closeCh:
+		return false
+	default:
+	}
+	select {
```
overvenus, line 293:
```suggestion
```

Could you merge two select?

overvenus (commented):
Rest LGTM

cdc/sorter/unified/unified_sorter.go
sdojjy: add two select to guarantee the done/close condition is checked first.

cdc/sorter/sorter.go
sdojjy: done

cdc/sorter/sorter.go
```diff
@@ -24,5 +24,7 @@ import (
...
+	TryAddEntry(ctx context.Context, entry *model.PolymorphicEvent) bool
```
liuzix, line 29: Do we need it to return an error? It may be useful to know that the sorter has been cancelled or otherwise irrecoverably unavailable.

cdc/sorter/unified/unified_sorter.go
overvenus: I see, could you add some some comments?

sdojjy: done

overvenus (approved):
LGTM

cdc/sorter/sorter.go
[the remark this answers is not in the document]
sdojjy: done

asddongmen (approved):

cdc/sorter/unified/unified_sorter.go
```diff
@@ -281,6 +281,24 @@ func (s *Sorter) AddEntry(ctx context.Context, entry *model.PolymorphicEvent) {
 	}
 }
 
+// TryAddEntry implements the EventSorter interface
+func (s *Sorter) TryAddEntry(ctx context.Context, entry *model.PolymorphicEvent) (bool, error) {
+	// add two select to guarantee the done/close condition is checked first.
+	select {
+	case <-ctx.Done():
+		return false, cerror.ErrSorterContextIsDone.GenWithStackByArgs()
```
liuzix, line 289: Can we just return `ctx.Err()` here?

sdojjy: done

liuzix (approved):

PR #3480, opened 2021-11-16
ticdc/token-region: don't set metric to 0 since multiple tables share it

amyangfei:

Metrics in token region are shared by multiple tables, so we should not use `set` to a set specific value

Use accurate inc or dec instead

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
Fix kv client cached region metric could be negative.
```


maxshuang:
I test this pr, look good.

asddongmen (approved):

ben1009 (approved):

overvenus:
Cc https://github.com/pingcap/tiflow/issues/4300

PR #3481, opened 2021-11-16
 redolog: add metric

ben1009:
add redo log related metric

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
add redo log related metric
```


ben1009:
> /run-unit-test

+1

ben1009:
image

ben1009:
> > /run-unit-test
> 
> +1

+1

cdc/redo/writer/file.go
```diff
@@ -495,7 +508,11 @@ func (w *Writer) flushAll() error {
 
 	ctx, cancel := context.WithTimeout(context.Background(), defaultS3Timeout)
 	defer cancel()
-	return w.writeToS3(ctx, w.file.Name())
+
+	err = w.writeToS3(ctx, w.file.Name())
+	w.metricFlushAllDuration.Observe(time.Since(start).Seconds())
```
amyangfei, line 513: Is this metric S3 only, if using NFS, this metric is not observed.

amyangfei (commented):
By the way, do we need to separate the metrics of writing files and uploading to S3, including flush duration and write bytes. For two reasons
- Flush duration has different meaning when using NFS or S3 as backend
- The written bytes to local files and to S3 storage are different when using S3 backend

cdc/redo/writer/file.go
ben1009: nfs use the same metric as local file 
```
	start := time.Now()
	err = w.file.Sync()
	w.metricFsyncDuration.Observe(time.Since(start).Seconds())
```

ben1009:
> By the way, do we need to separate the metrics of writing files and uploading to S3, including flush duration and write bytes. For two reasons
> 
> * Flush duration has different meaning when using NFS or S3 as backend
> * The written bytes to local files and to S3 storage are different when using S3 backend
 
could add a `written bytes` for s3, but since network should not be the bottleneck in aws, so not have that added at beginning.

overvenus (approved):
LGTM

amyangfei (approved):

ben1009:
> /run-verify

+1

ben1009:
> /run-integration-test

+1

ben1009:
> > /run-verify
> 
> 
> 
> +1

1

PR #3484, opened 2021-11-16
master: add declare for start/stop relay commands

lichunzhu:
close https://github.com/pingcap/ticdc/issues/3400

Add declare like `start-task`, `operate-source` ... for `start-relay` and `stop-relay`.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


okJiang (commented):
rest lgtm

dm/dm/master/server.go
```diff
@@ -1642,6 +1652,25 @@ func (s *Server) waitOperationOk(
 				} else {
 					return false, "", queryResp, terror.ErrMasterOperRespNotSuccess.Generate("relay is disabled for this source")
 				}
+			case *pb.OperateRelayRequest:
+				// including the situation that source status is nil and relay status is nil
+				relayStatus := queryResp.GetSourceStatus().GetRelayStatus()
+				if expect == pb.Stage_Stopped && relayStatus == nil {
```
Ehco1996, line 1658: seems  we loss the `relayStatus==nil && expect != pb. Stage_Stopped` branch ? 

although ths may not hanppen with the current logic,i think we should still take this branch into thought in case we go into it later when refactor the relay logic

how about change the if-else to

```go
if relayStatus == ni{
  if except == xxx{
  }

}else{
  if except == xxx{
  }
}
```

dm/dm/master/server.go
```diff
@@ -2243,6 +2275,29 @@ func (s *Server) OperateRelay(ctx context.Context, req *pb.OperateRelayRequest)
 		return resp2, nil
 	}
 	resp2.Result = true
+	// TODO: now we make sure req.Source isn't empty and len(req.Worker)>=1 in dmctl,
+	// we need refactor the logic here if this behavior changed in the future
+	var workerToCheck []string
+	for _, worker := range req.Worker {
```
Ehco1996, line 2281: `req.Worker` may empty after https://github.com/pingcap/ticdc/pull/3226 this merged 😅

maybe left a TODO here

Ehco1996 (commented):
restLGTM

Ehco1996 (approved):

GMHDBJD (approved):
LGTM

PR #3494, opened 2021-11-17
ticdc/redo: should cleanup redo logs when removing a paused changefeed

amyangfei:

Owner should cleanup redo logs when removing a paused changefeed

- Extract some ignoralbe errors for cdc processor
- Add a `redoManagerCleanup` function, it is called in when changefeed calls `releaseResources`, no matter whether the changefeed has been initialized.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
Fix a bug that redo logs are not cleaned up when removing a paused changefeed.
```


cdc/processor/processor.go
```diff
@@ -109,6 +109,25 @@ func newProcessor(ctx cdcContext.Context) *processor {
 	return p
 }
 
+// processorIgnoralbeError returns true if the error means the processor exits
+// normally, caused by changefeed pause, remove, etc.
+func processorIgnoralbeError(err error) bool {
```
ben1009, line 114: may be could include err==nil case, since already extract as a func

ben1009 (approved):

cdc/processor/processor.go
amyangfei: addressed

hi-rustin (approved):

PR #3497, opened 2021-11-17
*: make sink method FlushRowChangedEvents non-blocking

asddongmen:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/cdc/processor/pipeline/sink.go
+++ b/cdc/processor/pipeline/sink.go
@@ -114,7 +114,9 @@ func (n *sinkNode) stop(ctx pipeline.NodeContext) (err error) {
 	return
 }
 
-func (n *sinkNode) flushSink(ctx pipeline.NodeContext, resolvedTs model.Ts) (err error) {
+// flushSink will try to flush all rows commitTs before resolvedTs
+// if the action fails, it will return false.
+func (n *sinkNode) flushSink(ctx pipeline.NodeContext, resolvedTs model.Ts) (ok bool, err error) {
 	defer func() {
 		if err != nil {
 			n.status.Store(TableStatusStopped)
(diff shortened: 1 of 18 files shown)


cdc/processor/pipeline/sink.go
```diff
@@ -136,10 +136,17 @@ func (n *sinkNode) flushSink(ctx pipeline.NodeContext, resolvedTs model.Ts) (err
...
 		return errors.Trace(err)
 	}
-	checkpointTs, err := n.sink.FlushRowChangedEvents(ctx, resolvedTs)
-	if err != nil {
-		return errors.Trace(err)
+
+	var checkpointTs uint64
+	var ok bool
+	// retry until ok is true
+	for !ok {
```
sdojjy, line 143: why retry here?

asddongmen: If not retry here, the `resolvedTs` pass in this `Flush` may be lost.

cdc/processor/pipeline/sink.go
```diff
@@ -131,22 +134,24 @@ func (n *sinkNode) flushSink(ctx pipeline.NodeContext, resolvedTs model.Ts) (err
 		resolvedTs = n.targetTs
 	}
 	if resolvedTs <= n.checkpointTs {
-		return nil
+		return ok, nil
```
lonng, line 137: I think we use `true/false` directly if more readable. I don't think it's make sense to define a boolean `ok` and return `ok/!ok`.

liuzix:
require LGT3 because the PR changes public API.

cdc/processor/pipeline/sink.go
asddongmen: Reasonable suggestion, thanks~

asddongmen:
find  better way #3541

PR #3498, opened 2021-11-17
dm/election: fix datarace when campaign slow

Ehco1996:
close https://github.com/pingcap/ticdc/issues/3495

use `campaignMu` to protect `e.campaignWg`

In the for loop, campaignWg.Add(1) and cancelCampaign was assigned.

But in the test, Resign was called in another thread, and cancelCampaign was used in Resgin which race with campaignWg.Add(1) in the for loop.

### Check List 

Tests 

 - Unit test

Code changes

- None

Side effects

Related changes

### Release note 

```release-note
`None`.
```


okJiang:
Can you explain the cause of the problem? The original [link](ci2.pingcap.net/blue/organizations/jenkins/cdc_ghpr_test/detail/cdc_ghpr_test/12109/pipeline) has expired and cannot be traced

Ehco1996:
> Can you explain the cause of the unstable UT? The original [link](ci2.pingcap.net/blue/organizations/jenkins/cdc_ghpr_test/detail/cdc_ghpr_test/12109/pipeline) has expired and cannot be traced.

see here, https://github.com/pingcap/ticdc/issues/3495

okJiang (approved):
In the for loop, `campaignWg.Add(1)` and `cancelCampaign` was assigned.

But in the test, Resign was called in another thread, and cancelCampaign was used in Resgin which race with `campaignWg.Add(1)` in the for loop.

lance6716 (approved):

PR #3499, opened 2021-11-17
cdc/capture: init table actor system if it is enabled

sdojjy:
it's a part of the table pipeline actor refactor.
This PR init table actor system if it is enabled

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/capture/capture.go
+++ b/cdc/capture/capture.go
@@ -28,7 +28,7 @@ import (
 	"github.com/pingcap/ticdc/cdc/model"
 	"github.com/pingcap/ticdc/cdc/owner"
 	"github.com/pingcap/ticdc/cdc/processor"
-	"github.com/pingcap/ticdc/pkg/actor"
+	"github.com/pingcap/ticdc/cdc/processor/pipeline/system"
 	"github.com/pingcap/ticdc/pkg/config"
 	cdcContext "github.com/pingcap/ticdc/pkg/context"
 	cerror "github.com/pingcap/ticdc/pkg/errors"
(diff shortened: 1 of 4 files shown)


cdc/capture/capture.go
```diff
@@ -60,6 +61,9 @@ type Capture struct {
 	etcdClient *etcd.CDCEtcdClient
 	grpcPool   kv.GrpcPool
 
+	tableActorSystem *actor.System
+	tableActorRouter *actor.Router
```
overvenus, line 65: Could you wrap them in a struct? like https://github.com/pingcap/ticdc/pull/3406

overvenus (approved):
rest LGTM

cdc/capture/capture.go
sdojjy: ok

cdc/capture/capture.go
```diff
@@ -354,6 +369,10 @@ func (c *Capture) AsyncClose() {
 	if c.grpcPool != nil {
 		c.grpcPool.Close()
 	}
+	if c.tableActorSystem != nil {
+		err := c.tableActorSystem.Stop()
+		log.Warn("stop table actor system failed", zap.Error(err))
```
liuzix, line 374: Do you think it would be necessary to set `c.tableActorSystem` to `nil`?

liuzix: I think the function `AsyncClose` is assumed to be idempotent.

sdojjy: yes, it's better to set  c.tableActorSystem to nil.

liuzix (approved):

PR #3501, opened 2021-11-17
cdc/sink: only check max-message-bytes for default protocol & craft protocol

3AceShowHand:

When try to create the topic, only check `max-message-bytes` when using the default open protocol or craft protocol.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
None`.
```


overvenus (approved):

amyangfei (approved):

amyangfei (commented):
Is there any unit test to cover the change

3AceShowHand:
> Is there any unit test to cover the change

it's quite hard to cover this case in the unit test, since sarama mock broker does not work well in this case,  it's mocked response does not have valid information about what we want, which always making the case fail.

PR #3516, opened 2021-11-18
processor: let sorter output events less or equal to barrier ts

overvenus:

Do not send resolved ts events that is larger than barrier ts.
When DDL puller stall, it may cause data pile up in memory,
because sorter outputs resolved events that have to wait DDL.

### Check List 

Tests 

 - Unit test
 - Integration test

| master | this PR |
| -- | -- |
| image | image |
| image | image |
| image | image |
| image | image |

### Release note 

```release-note
None
```


sdojjy (approved):
lgtm

cdc/processor/pipeline/sorter.go
```diff
@@ -211,8 +220,28 @@ func (n *sorterNode) Receive(ctx pipeline.NodeContext) error {
 					zap.Uint64("oldResolvedTs", oldResolvedTs))
 			}
 			atomic.StoreUint64(&n.resolvedTs, rawKV.CRTs)
+
+			if resolvedTs > n.barrierTs &&
+				!redo.IsConsistentEnabled(n.replConfig.Consistent.Level) {
+				// Do not send resolved ts events that is larger than
+				// barrier ts.
+				// When DDL puller stall, it may cause data pile up in memory,
```
asddongmen, line 228: I think it is a bit confusing that what this "it" refers to. 
Do you mean `If we send a resolved ts larger than barrier ts` ?

cdc/processor/pipeline/sorter.go
```diff
@@ -211,8 +220,28 @@ func (n *sorterNode) Receive(ctx pipeline.NodeContext) error {
 					zap.Uint64("oldResolvedTs", oldResolvedTs))
 			}
 			atomic.StoreUint64(&n.resolvedTs, rawKV.CRTs)
+
+			if resolvedTs > n.barrierTs &&
+				!redo.IsConsistentEnabled(n.replConfig.Consistent.Level) {
```
amyangfei, line 225: Don't need to check whether redo log is enabled here, since it doesn't break anything when redo log is enabled.

overvenus: I think it may cause deadlock, because when redolog is enable, resolved ts is calculated from sink. If sorter cuts resolved ts, sink can never advance its resolved ts.

amyangfei: sink can forward resolved ts to barrier ts at least, why deadlock happens

amyangfei (approved):

liuzix (approved):

PR #3539, opened 2021-11-19
cdc/sink: adjust kafka producer configuration

3AceShowHand:

This PR tries to make producer configuration, especially `retry` and `timeout` related items be more reasonable.

* For Net-related, Set `DialTimeout` / `WriteTimeout` / `ReadTimeout` to 5s at the moment. `30s -> 5s`
  * **Net-related timeout is the base for almost all operations, and it's the lower bound for all other Timeout**
  * At the moment, `30s` is quite long which will make the owner block.
  * When creating the changefeed, WriteTimeout will affect `NewClient` / `ListTopic` / `DescribeCluster` / `DescribeConfig` operation et al. 
* For Admin-related
  * When creating the changefeed, it will affect the `CreateTopic` operation. 
  * Set single request timeout to `5s`.
* For Metadata-related, it's Timeout is the hard upper bound with retry. set `Metadata.Retry.Max` to infinity, and set hard timeout to `10s`.
  * Current configuration, the maximum block time may be as long as `3690` seconds. take https://github.com/Shopify/sarama/blob/main/config.go#L150 as a reference.
  * When creating the changefeed, the sarama client will try to GetMedadata, just the timeout to `10s`.
  * When the changefeed running, the producer will refresh metadata in the background every 10 minutes. 
* For producer-related, 
  * Set `Producer.Timeout` to `10s`, which is the single request timeout for receiving all required `RequiredAcks`.
  * set `Producer.Retry.Max` to 5, and `Producer.Retry.Backoff` to 100ms, around 50 seconds at all.
  * Original, Set `Producer.Retry.Max` to`600`. In the worst case, would cost 6000s for the message.
 
* Remove `sarama.MaxRequestSize = 1024 * 1024 *1024`, use the default 100MB.
  * For each CDC message, it will be batched into a `batch`, batchByteSize will not be greater than user configured `max-message-bytes` or default `1M`.
  * When a producer sends a request to the broker cluster, it may send multiple batches in a single request.
  * `MaxRequestSize` is identical to [socket.request.max.bytes](https://kafka.apache.org/documentation/?from=from_parent_mindnote#brokerconfigs_socket.request.max.bytes), which is default to 100MB, just use the default.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/sink/producer/kafka/kafka.go
+++ b/cdc/sink/producer/kafka/kafka.go
@@ -610,7 +610,7 @@ func newSaramaConfig(ctx context.Context, c *Config) (*sarama.Config, error) {
 	// replication logs, this process will last from a few seconds to a few minutes.
 	// Kafka cluster will not provide a writing service in this process.
 	config.Metadata.Retry.Max = math.MaxInt
-	config.Metadata.Timeout = 10 * time.Second
+	config.Metadata.Timeout = time.Minute
 
 	config.Producer.Partitioner = sarama.NewManualPartitioner
 	config.Producer.MaxMessageBytes = c.MaxMessageBytes
--- a/cdc/sink/producer/kafka/kafka.go
+++ b/cdc/sink/producer/kafka/kafka.go
@@ -606,7 +606,7 @@ func newSaramaConfig(ctx context.Context, c *Config) (*sarama.Config, error) {
 	// Kafka cluster will not provide a writing service in this process.
 	// Time out in one minute
 	config.Metadata.Retry.Max = math.MaxInt32
-	config.Metadata.Timeout = time.Minute
+	config.Metadata.Timeout = 10 * time.Second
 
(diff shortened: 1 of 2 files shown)


ben1009:
i assume the kafka cluster is provided and maintained by client, so if some of the param may effect the cluster performance, may need reviewed by client and modified base on different load. another reason, the cluster may not always dedicated for cdc only.

3AceShowHand:
> i assume the kafka cluster is provided and maintained by client, so if some of the param may effect the cluster performance, may need reviewed by client and modified base on different load. another reason, the cluster may not always dedicated for cdc only.

The most reasonable way，TiCDC should let user decide how to configure Sink configurations.

Issue #3503, opened 2021-11-17
[ticdc 5.3.0] tikv_scale_plan_tidb case fail: 1. cdc oom, 2. data inconsistency, 3. changefeed stuck


Tammyxia:
### What did you do?

// create 1 changefeed, 3 cdc, 3 tikv
// run tpcc 100 warehouse prepare
// run tpcc run, meanwhile, scale cluster tikv from 3 -> 7, after 15m, scale in to  3
// create table "finishmark"
// wait table "finishmark" sync to downstream 
//check data consistency

### What did you expect to see?

pass

### What did you see instead?

- [x] 1. cdc oom
- [x] 2. The two table has data inconsistency: warehouse，order_line
- [x] 3.  changefeed stuck at almost the time create table "finishmark"

testbed saved for 24 hours.

- upstream grafana:  http://172.16.4.180:30687/

- testbed: cdc-testbed--tps-450050-1-772

- conprof: http://172.16.4.180:30857/?query=%7Bnamespace%3D%22cdc-testbed--tps-450050-1-772%22%2C+job%3D%22ticdc%22%2C+profile_path%3D%22%2Fdebug%2Fpprof%2Fheap%22%7D&from=1637135181536&to=1637135672266&now=false

- test case log: http://172.16.4.180:31714/workflows/testground/plan-exec-450050?tab=workflow&nodeId=plan-exec-450050-3513432500&sidePanel=logs

### Versions of the cluster

Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

```console
TiKV 
Release Version:   5.3.0
Edition:           Community
Git Commit Hash:   d514230a40974393297050645c223bcf1db9aedc
Git Commit Branch: heads/refs/tags/v5.3.0
UTC Build Time:    2021-11-16 12:18:25
Rust Version:      rustc 1.56.0-nightly (2faabf579 2021-07-27)
Enable Features:   jemalloc mem-profiling portable sse protobuf-codec test-engines-rocksdb cloud-aws cloud-gcp
Profile:           dist_release
```

TiCDC version (execute `cdc version`):

```console
/cdc version
Release Version: v5.3.0
Git Commit Hash: f847b331572379527bf37a7f19be20448a74b2c2
Git Branch: heads/refs/tags/v5.3.0
UTC Build Time: 2021-11-16 11:54:34
Go Version: go version go1.16.4 linux/amd64
Failpoint Build: false
```

Tammyxia:
[...]

PR #2617, opened 2021-08-24
sink: reduce lock competition during flushBackendSink

hi-rustin:
### What problem does this PR solve? 
None

### What is changed and how it works?
reduce lock competition during flushBackendSink.

### Check List 

Tests 

 -  Benchmark (TBD)

Code changes

None

Side effects

None

Related changes

 - Need to cherry-pick to the release branch

### Release note 

```release-note
None
```

/cc

PR #3540, opened 2021-11-19
processor,sink(cdc): let sink report resolved ts and do not skip buffer sink flush

overvenus:

Revert #3029 and #2617 temporary for resolving #3503.

Cc https://github.com/pingcap/ticdc/issues/3545

### Check List 

Tests 

 - Integration test

See #3503

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/owner/changefeed.go
+++ b/cdc/owner/changefeed.go
@@ -391,10 +391,6 @@ func (c *changefeed) handleBarrier(ctx cdcContext.Context) (uint64, error) {
 		if !blocked {
 			return barrierTs, nil
 		}
-		log.Info("advance DDL barrier",
-			zap.Uint64("checkpointTs", c.state.Status.CheckpointTs),
-			zap.Uint64("resolvedTs", c.state.Status.ResolvedTs),
-			zap.Uint64("barrierTs", barrierTs))
 		done, err := c.asyncExecDDL(ctx, ddlJob)
 		if err != nil {
 			return 0, errors.Trace(err)
--- a/cdc/owner/changefeed.go
+++ b/cdc/owner/changefeed.go
@@ -391,6 +391,10 @@ func (c *changefeed) handleBarrier(ctx cdcContext.Context) (uint64, error) {
 		if !blocked {
 			return barrierTs, nil
 		}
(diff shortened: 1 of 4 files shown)


cdc/owner/changefeed.go
```diff
@@ -391,6 +391,10 @@ func (c *changefeed) handleBarrier(ctx cdcContext.Context) (uint64, error) {
 		if !blocked {
 			return barrierTs, nil
 		}
+		log.Info("advance DDL barrier",
```
liuzix, line 394: The message here can be confusing to users and support staff. The "DDL barrier" does not seem to be actually changed here. Also, for a single DDL, this piece of code can be executed more than once, causing more confusion.

cdc/sink/table_sink.go
```diff
@@ -55,6 +57,15 @@ func (t *tableSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error
 // is required to be no more than global resolvedTs, table barrierTs and table
 // redo log watermarkTs.
 func (t *tableSink) FlushRowChangedEvents(ctx context.Context, resolvedTs uint64) (uint64, error) {
+	logAbnormalCheckpoint := func(ckpt uint64) {
```
liuzix, line 60: Can you add some comments here?

cdc/sink/table_sink.go
```diff
@@ -55,6 +57,15 @@ func (t *tableSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error
...
+		if ckpt > resolvedTs {
+			log.L().WithOptions(zap.AddCallerSkip(1)).
+				Warn("checkpoint ts > resolved ts, flushed more than emitted",
```
amyangfei, line 63: If this happens, is it better to return an error and pause the changefeed

amyangfei (approved):

cdc/sink/table_sink.go
overvenus: The probability is high after table rescheduling, return error may cause changefeed stuck easily. 
In addition, this situation does not necessarily mean data loss.

3AceShowHand (approved):

PR #3541, opened 2021-11-19
redo, sink(ticdc): make some func non-blocking

asddongmen:
make some method of sink and redo non-blocking to adapt table actor.

1. add 2 types of error `ErrRedoLogBufferBlocking` and `ErrFlushTsBlocking` to indicate the redolog buffer or flushBuffer is blocking.
2. make `logManager` flush resolvedTs in a backend goroutine.
3. add handle logic for the errors mentioned above in related sink model.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```


cdc/processor/pipeline/sink.go
```diff
@@ -315,17 +322,47 @@ func (n *sinkNode) Receive(ctx pipeline.NodeContext) error {
...
-			if err := n.flushSink(ctx, msg.PolymorphicEvent.CRTs); err != nil {
+
+			err := retry.Do(ctx, func() error {
+				err := n.flushSink(ctx, msg.PolymorphicEvent.CRTs)
+				if err != nil {
+					return err
+				}
+				return nil
+			}, retry.WithBackoffBaseDelay(20),
+				retry.WithInfiniteTries(),
```
liuzix, line 333: Please comment on when `flushSink` will be retried. Also can you print a warning if retry keeps failing after a certain duration (say 10 seconds)?

cdc/processor/pipeline/sink.go
```diff
@@ -315,17 +322,47 @@ func (n *sinkNode) Receive(ctx pipeline.NodeContext) error {
...
-		if err := n.emitEvent(ctx, event); err != nil {
+
+		err := retry.Do(ctx, func() error {
+			err := n.emitEvent(ctx, event)
+			if err != nil {
+				return err
+			}
+			return nil
+		}, retry.WithBackoffBaseDelay(20),
+			retry.WithInfiniteTries(),
```
liuzix, line 350: ditto

cdc/processor/pipeline/sink.go
```diff
@@ -315,17 +322,47 @@ func (n *sinkNode) Receive(ctx pipeline.NodeContext) error {
...
+			retry.WithIsRetryableErr(isRetryableError))
+		if err != nil {
 			return errors.Trace(err)
 		}
 	case pipeline.MessageTypeTick:
-		if err := n.flushSink(ctx, n.resolvedTs); err != nil {
+		err := retry.Do(ctx, func() error {
```
liuzix, line 356: Maybe we can wrap the retry logic in a helper function?

cdc/processor/pipeline/sink.go
```diff
@@ -346,3 +392,13 @@ func (n *sinkNode) Destroy(ctx pipeline.NodeContext) error {
 	n.flowController.Abort()
 	return n.sink.Close(ctx)
 }
+
+func isRetryableError(err error) bool {
+	if err == nil {
+		return false
+	}
+	if cerror.ErrRedoLogBufferBlocking.Equal(err) {
```
liuzix, line 400: Just `return cerror.ErrRedoLogBufferBlocking.Equal(err)`

cdc/redo/manager.go
```diff
@@ -51,9 +51,10 @@ const (
 )
 
 const (
-	// supposing to replicate 10k tables, each table has one cached changce averagely.
+	// supposing to replicate 10k tables, each table has one cached change averagely.
 	logBufferChanSize = 10000
-	logBufferTimeout  = time.Minute * 10
+	// supposing to replicate 10k tables, each table has one resolvedTs change averagely.
+	flushBufferChanSize = 10000
```
liuzix, line 57: Can you estimate the memory consumption for each `ManagerImpl`?

cdc/redo/manager.go
```diff
@@ -119,15 +120,22 @@ type cacheRows struct {
 	rows    []*model.RowChangedEvent
 }
 
+// resolvedEv represents a resolvedTs event
+type resolvedEv struct {
```
liuzix, line 124:
```suggestion
type resolvedEvent struct {
```

cdc/redo/manager.go
```diff
@@ -119,15 +120,22 @@ type cacheRows struct {
...
+	tableID    model.TableID
+	resolvedTs uint64
```
liuzix, line 126:
```suggestion
	resolvedTs model.Ts
```

cdc/redo/manager.go
```diff
@@ -375,3 +390,23 @@ func (m *ManagerImpl) bgWriteLog(ctx context.Context, errCh chan<- error) {
...
 	}
 }
+
+func (m *ManagerImpl) bgFlushLog(ctx context.Context, errCh chan<- error) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case event := <-m.flushBuffer:
+			err := m.writer.FlushLog(ctx, event.tableID, event.resolvedTs)
```
liuzix, line 400: Can you check for `event == nil`?

cdc/redo/manager.go
```diff
@@ -375,3 +390,23 @@ func (m *ManagerImpl) bgWriteLog(ctx context.Context, errCh chan<- error) {
...
+			if err != nil {
+				select {
+				case errCh <- err:
+				default:
+					log.Error("err channel is full", zap.Error(err))
```
liuzix, line 405: The error message can be more specific

cdc/sink/buffer_sink.go
```diff
@@ -20,6 +20,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	cerror "github.com/pingcap/ticdc/pkg/errors"
+
```
liuzix, line 24:
```suggestion
```

cdc/sink/buffer_sink.go
```diff
@@ -151,6 +153,8 @@ func (b *bufferSink) FlushRowChangedEvents(ctx context.Context, resolvedTs uint6
 	case <-ctx.Done():
 		return atomic.LoadUint64(&b.checkpointTs), ctx.Err()
 	case b.flushTsChan <- resolvedTs:
+		return atomic.LoadUint64(&b.checkpointTs), nil
+	default:
 	}
-	return atomic.LoadUint64(&b.checkpointTs), nil
+	return atomic.LoadUint64(&b.checkpointTs), cerror.ErrFlushTsBlocking.FastGenByArgs()
```
liuzix, line 159: Is the returned timestamp meaningful when there is a non-nil error?

cdc/sink/manager_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"testing"
 	"time"
 
+	cerror "github.com/pingcap/ticdc/pkg/errors"
+
```
liuzix, line 26:
```suggestion
```

pkg/errors/errors.go
```diff
@@ -120,6 +120,9 @@ var (
 	ErrOldValueNotEnabled        = errors.Normalize("old value is not enabled", errors.RFCCodeText("CDC:ErrOldValueNotEnabled"))
 	ErrSinkInvalidConfig         = errors.Normalize("sink config invalid", errors.RFCCodeText("CDC:ErrSinkInvalidConfig"))
 	ErrCraftCodecInvalidData     = errors.Normalize("craft codec invalid data", errors.RFCCodeText("CDC:ErrCraftCodecInvalidData"))
+	// internal use only
```
liuzix, line 123: Why is it necessary to use two distinct errors?

cdc/processor/pipeline/sink.go
asddongmen: got

cdc/processor/pipeline/sink.go
asddongmen: got

cdc/redo/manager.go
asddongmen: ok

cdc/sink/buffer_sink.go
asddongmen: yes, we can ignore the `ErrFlushTsBlocking` in current implementation, which means that the caller will ignore the error and just accept the `checkpointTs` that not updated.

pkg/errors/errors.go
asddongmen: In current implementation 
1. when the caller meet a `ErrFlushTsBlocking` error, it can just ignore it, because the resolvedTs will continually send by the upstream, which is a implicit retry mechanism.
2. when the caller meet a `ErrRedoLogBufferBlocking` error, it should retry until it success, because the row event will send by the upstream only once.

cdc/sink/buffer_sink.go
liuzix: When is it necessary to differentiate between "blocking" and "checkpoint-ts not updating"?

pkg/errors/errors.go
liuzix: > In current implementation
> 
> 1. when the caller meet a `ErrFlushTsBlocking` error, it can just ignore it, because the resolvedTs will continually send by the upstream, which is a implicit retry mechanism.
> 2. when the caller meet a `ErrRedoLogBufferBlocking` error, it should retry until it success, because the row event will send by the upstream only once.

This seems to be a matter of the different semantics of different methods. But the underlying reasons for returning errors seem to be the same?

cdc/redo/manager.go
asddongmen: event would never be nil

cdc/sink/buffer_sink.go
asddongmen: since n.checkpoints is refreshed asynchronously in a background goroutine,  we return a n.checkpoints every time FlushRowChangedEvents be called, the caller will always check the returned checkpointTs

cdc/processor/pipeline/sink.go
```diff
@@ -151,9 +154,12 @@ func (n *sinkNode) flushSink(ctx context.Context, resolvedTs model.Ts) (err erro
 		return errors.Trace(err)
 	}
 	checkpointTs, err := n.sink.FlushRowChangedEvents(ctx, n.tableID, resolvedTs)
-	if err != nil {
+	// we can ignore ErrFlushTsBlocking error because resolvedTs will continue
+	// to come
```
overvenus, line 158: Do you mean `flushSink` is called periodically when resolvedTs is updated?

cdc/sink/buffer_sink.go
```diff
@@ -191,8 +193,13 @@ func (b *bufferSink) FlushRowChangedEvents(ctx context.Context, tableID model.Ta
...
 		resolvedTs: resolvedTs,
 	}:
+		return b.getTableCheckpointTs(tableID), nil
+	default:
 	}
-	return b.getTableCheckpointTs(tableID), nil
+	// since n.checkpoints is refreshed asynchronously in a background goroutine,
+	// we return a n.checkpoints every time FlushRowChangedEvents be called
+	// the caller will always check the returned checkpointTs
+	return b.getTableCheckpointTs(tableID), cerrors.ErrFlushTsBlocking.FastGenByArgs()
```
overvenus, line 202:
```go
	default:
		// since n.checkpoints is refreshed asynchronously in a background goroutine,
		// we return a n.checkpoints every time FlushRowChangedEvents be called
		// the caller will always check the returned checkpointTs
		return b.getTableCheckpointTs(tableID), cerrors.ErrFlushTsBlocking.FastGenByArgs()
	}
```

I prefer returning blocking error in the default case.

overvenus: By dropping flushMsg, will it cause flowcontrol deadlock? @liuzix

cdc/redo/manager.go
```diff
@@ -371,10 +388,30 @@ func (m *ManagerImpl) bgWriteLog(ctx context.Context, errCh chan<- error) {
...
 	}
 }
+
+func (m *ManagerImpl) bgFlushLog(ctx context.Context, errCh chan<- error) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case event := <-m.flushBuffer:
+			err := m.writer.FlushLog(ctx, event.tableID, event.resolvedTs)
```
overvenus, line 405: By calling `FlushLog`, will it produce a small file? @ben1009

cdc/processor/pipeline/sink.go
```diff
@@ -334,16 +342,35 @@ func (n *sinkNode) HandleMessage(ctx context.Context, msg pipeline.Message) (boo
...
+			atomic.StoreUint64(&n.resolvedTs, event.CRTs)
 			return true, nil
 		}
-		if err := n.emitEvent(ctx, event); err != nil {
+		start := time.Now()
+		err := retry.Do(ctx, func() error {
+			err := n.emitEvent(ctx, event)
+			cost := time.Since(start)
+			if cost >= 10*time.Second {
+				log.Warn("redo log buffer blocking too long", zap.Float64("blocking time(s)", cost.Seconds()))
```
overvenus, line 357:
```suggestion
				log.Warn("redo log buffer blocking too long", zap.Duration("blocking time(s)", cost))
```

overvenus (commented):
Please add tests to cover nonblocking and retry logic.

cdc/sink/buffer_sink.go
asddongmen: > By dropping flushMsg, will it cause flowcontrol deadlock? @liuzix

The deadlock issue will fix in #4084, after it be merged, dropping flushMsg here will not cause flowcontrol deadlock.

cdc/processor/pipeline/sink.go
asddongmen: Yes, it is what I mean, I will polish this comment.

asddongmen:
@overvenus The unit test cases are added, PTAL~ thanks!

cdc/processor/pipeline/sink.go
```diff
@@ -334,16 +343,35 @@ func (n *sinkNode) HandleMessage(ctx context.Context, msg pipeline.Message) (boo
...
+			atomic.StoreUint64(&n.resolvedTs, event.CRTs)
 			return true, nil
 		}
-		if err := n.emitEvent(ctx, event); err != nil {
+		start := time.Now()
+		err := retry.Do(ctx, func() error {
+			err := n.emitEvent(ctx, event)
+			cost := time.Since(start)
+			if cost >= 10*time.Second {
+				log.Warn("redo log buffer blocking too long", zap.Duration("blocking time(s)", cost))
```
overvenus, line 358: What's the frequency of the log? Does it flood when retrying?

cdc/sink/buffer_sink_test.go
```diff
@@ -20,6 +20,8 @@ import (
 	"testing"
 	"time"
 
+	cerrors "github.com/pingcap/tiflow/pkg/errors"
+
```
overvenus, line 24:
```suggestion
	cerrors "github.com/pingcap/tiflow/pkg/errors"
```

cdc/redo/manager.go
```diff
@@ -251,7 +257,18 @@ func (m *ManagerImpl) FlushLog(
...
+	case <-ctx.Done():
+		return ctx.Err()
+	case m.flushBuffer <- resolvedEvent{
+		tableID:    tableID,
+		resolvedTs: resolvedTs,
+	}:
+		return nil
+	default:
+	}
+	return cerror.ErrFlushTsBlocking.FastGenByArgs()
```
overvenus, line 271:
```suggestion
	default:
		return cerror.ErrFlushTsBlocking.FastGenByArgs()
	}
```

cdc/redo/manager.go
```diff
@@ -212,32 +224,26 @@ func (m *ManagerImpl) Enabled() bool {
...
 		// Because the pipeline sink doesn't hold slice memory after calling
 		// EmitRowChangedEvents, we copy to a new slice to manage memory
 		// in redo manager itself, which is the same behavior as sink manager.
 		rows: append(make([]*model.RowChangedEvent, 0, len(rows)), rows...),
 	}:
+		return nil
+	default:
 	}
-	return nil
+	return cerror.ErrRedoLogBufferBlocking.FastGenByArgs()
```
overvenus, line 246:
```go
	default:
		return cerror.ErrRedoLogBufferBlocking.FastGenByArgs()
	}
```

cdc/processor/pipeline/sink.go
```diff
@@ -364,3 +393,28 @@ func (n *sinkNode) Destroy(ctx pipeline.NodeContext) error {
...
+func (n *sinkNode) retryFlushSink(ctx context.Context, resolvedTs model.Ts) error {
+	start := time.Now()
+	err := retry.Do(ctx, func() error {
+		// We must retry until success when flushSink return a ErrRedoLogBufferBlocking
+		// error to ensure the data is safely store by redo log module
+		// Note: the chance that flushSink return a ErrRedoLogBufferBlocking error is low.
+		err := n.flushSink(ctx, resolvedTs)
+		cost := time.Since(start)
+		if cost >= redoLogBlockingWarnLogDuration {
+			log.Warn("redo log buffer blocking too long", zap.Duration("blocking time(s)", cost))
```
overvenus, line 406: What's the frequency of the log? Does it flood when retrying?

cdc/processor/pipeline/sink.go
asddongmen: Yes, it would flood when retrying too many times. I will add some code to limit the warn log output.

cdc/sink/manager_test.go
```diff
@@ -188,7 +190,7 @@ func (s *managerSuite) TestManagerAddRemoveTable(c *check.C) {
 				c.Assert(err, check.IsNil)
 			}
 			_, err := sink.FlushRowChangedEvents(ctx, sink.(*tableSink).tableID, resolvedTs)
```
overvenus, line 192: How about change `FlushRowChangedEvents` to following

```go
type Sink interface {
	FlushRowChangedEvents(ctx context.Context, tableID model.TableID, resolvedTs uint64) (uint64, bool, error)
}
```

So we can use the second return value `bool` to determine whether the sink is block or not.

I think it's more intuitive than comparing `ErrFlushTsBlocking`.

PR #3549, opened 2021-11-20
docs(both): update format of the commit message

hi-rustin:

close https://github.com/pingcap/ticdc/issues/3195

update format of the commit message.

### Check List 

Tests 

 - No code

Code changes

None

Side effects

None

Related changes

None

### Release note 

```release-note
None
```


CONTRIBUTING.md
```diff
@@ -62,30 +62,31 @@ questions: what changed and why. The subject line should feature the what and
...
+capture(ticdc): add comment for variable declaration
 
 Improve documentation.
 ```
 
 The format can be described more formally as follows:
 
 ```
-<subsystem>: <what changed>
+<subsystem>[(ticdc|dm|both)]: <what changed>
```
hi-rustin, line 73: I don't know if this `both` is a good idea, but for now I think it might be better if we write it out explicitly. Plus it will make it easier to automate our checks later.

hi-rustin:
We should need to merge these two documents, but that is not the goal of this PR.

CONTRIBUTING.md
```diff
@@ -62,30 +62,31 @@ questions: what changed and why. The subject line should feature the what and
...
+capture(ticdc): add comment for variable declaration
 
 Improve documentation.
 ```
 
 The format can be described more formally as follows:
 
 ```
-<subsystem>: <what changed>
+<subsystem>(ticdc|dm|both): <what changed>
```
amyangfei, line 73: Recommend to use explicit subsystem name(s), for two reasons
- There may be other subsystem
- Explicit name(s) is good for searching

hi-rustin: Sorry, I don't quite understand what you mean. Do you mean putting the prefix in front?
`ticdc/xxxx,dm/xxxx: xxxx`?

amyangfei: I mean write all subsystems explicitly, such as 
- `ticdc,dm/xxx`, if affects both ticdc and dm
- `ticdc,dm,something-else,another-subsystem/xxx` it affects more than two subsystems.
- `ticdc/xxx` or `dm/xxx` if only affects one of the subsystem

hi-rustin: There seems to be some confusion between subsystem and area, would it be too long if we prefixed both with area? It seems that many git tools have a limit on the length of the first line (including GitHub).

hi-rustin:
/cc @overvenus @maxshuang 

hi-rustin:
/cc @lance6716 @Ehco1996 

dm/CONTRIBUTING.md
```diff
@@ -73,30 +73,31 @@ questions: what changed and why. The subject line should feature the what and
 the body of the commit should describe the why.
 
 ```
-dm-master: add comment for variable declaration
+capture(ticdc): add comment for variable declaration
```
lance6716, line 76: this is dm/CONTRIBUTING.md, so the example could be related to dm

lance6716 (commented):
rest lgtm

dm/CONTRIBUTING.md
Ehco1996: +1

Ehco1996 (approved):
LGTM after suggestion addressed

dm/CONTRIBUTING.md
hi-rustin: Addressed.

CONTRIBUTING.md
hi-rustin: @lance6716 @Ehco1996  Do you have any thoughts on it? Would you rather use`both` or write `dm,ticdc` explicitly?

CONTRIBUTING.md
maxshuang, line 73: LGTM,   Extend can be like:
```
<subsystem0, subsystem1,...>(ticdc|dm|both): <what changed>
```

CONTRIBUTING.md
Ehco1996: i think they are same, if we choose `both` we can add some rule like `both=ticdc,dm...` in bot

maxshuang (approved):

lance6716:
lgtm

overvenus (approved):

PR #3551, opened 2021-11-20
cdc/scheduler_v2: interfaces and mocks for ScheduleDispatcher

liuzix:
- In preparation to merge the ScheduleDispatcher, we need to clearly define the interfaces involved.

- Defined interface `ScheduleDispatcher`.
- Defined interface `ScheduleDispatcherCommunicator`.
- Implemented `mockScheduleDispatcherCommunicator` for testing.

### Check List 

Tests 

 - No code (no unit-testable code)

Code changes

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/scheduler/scheduler_dispatcher.go
+++ b/cdc/scheduler/scheduler_dispatcher.go
@@ -1,0 +1,64 @@
+// Copyright 2021 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package scheduler
+
+import (
+	"github.com/pingcap/ticdc/cdc/model"
+	"github.com/pingcap/ticdc/pkg/context"
+)
+
+// ScheduleDispatcher is an interface for a table scheduler used in Owner.
+type ScheduleDispatcher interface {
+	// Tick is called periodically to update the SchedulerDispatcher on the latest state of replication.
+	// This function should NOT be assumed to be thread-safe. No concurrent calls allowed.
+	Tick(
+		ctx context.Context,
+		checkpointTs model.Ts, // Latest global checkpoint of the changefeed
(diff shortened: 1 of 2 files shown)


cdc/scheduler/scheduler_dispatcher.go
```diff
@@ -0,0 +1,64 @@
...
+	DispatchTable(ctx context.Context,
+		changeFeedID model.ChangeFeedID,
+		tableID model.TableID,
+		captureID model.CaptureID,
+		isDelete bool, // True when we want to remove a table from the capture.
+	) (done bool, err error)
+
+	// Announce announces to the specified capture that the current node has become the Owner.
+	Announce(ctx context.Context,
+		changeFeedID model.ChangeFeedID,
```
amyangfei, line 56: Why does this API need to pass a changefeed ID

liuzix: A changefeed-independent `Announce` seems a little bit hard to implement without changing too many components of our current Owner implementation. Changefeed-wise `Announce` makes the protocol easier to reason about and maintains the structure of the original per-changefeed scheduler.

In our short- to mid- term road map, we do not seem to have an urgent requirement for 10000+ changefeeds, so the performance overhead seems acceptable.

cdc/scheduler/scheduler_dispatcher_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"github.com/pingcap/log"
+	"github.com/pingcap/ticdc/cdc/model"
+	cdcContext "github.com/pingcap/ticdc/pkg/context"
+	"github.com/stretchr/testify/mock"
+	"go.uber.org/zap"
+)
+
+type mockScheduleDispatcherCommunicator struct {
```
ben1009, line 24: seems the mock is modified after generated, any rules should fellow if someone else modify later

cdc/scheduler/scheduler_dispatcher.go
```diff
@@ -0,0 +1,64 @@
...
+		captureID model.CaptureID) (done bool, err error)
+}
+
+const (
```
3AceShowHand, line 60: can you move this const to the top of the file?

cdc/scheduler/scheduler_dispatcher_test.go
liuzix: This mock is not generated. Do you have any specific suggestions about the rules?

amyangfei (approved):

3AceShowHand (approved):

PR #3560, opened 2021-11-22
dm/syncer: fix directly use types.NewDatum

WizardXiao:

fix https://github.com/pingcap/ticdc/issues/3550

use `AdjustBinaryProtocolForDatum` instead of `types.NewDatum`
### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


dm/syncer/dml.go
```diff
@@ -686,14 +687,18 @@ func genKeyList(table string, columns []*model.ColumnInfo, dataSeq []interface{}
 }
 
 // truncateIndexValues truncate prefix index from data.
-func truncateIndexValues(indexColumns *model.IndexInfo, tiColumns []*model.ColumnInfo, data []interface{}) []interface{} {
+func truncateIndexValues(ctx sessionctx.Context, indexColumns *model.IndexInfo, tiColumns []*model.ColumnInfo, data []interface{}) []interface{} {
 	values := make([]interface{}, 0, len(indexColumns.Columns))
+	datums, err := utils.AdjustBinaryProtocolForDatum(ctx, data, tiColumns)
```
Ehco1996, line 692: seems we can use `tablecodec.TruncateIndexValues` to save a for loop below

Ehco1996 (commented):
rest LGTM

dm/syncer/dml.go
WizardXiao: yes, `AdjustBinaryProtocolForDatum` which is diffrent from `types.NewDatum` returns `[]types.Datum` , i will fix.

Ehco1996 (approved):

lance6716 (approved):

PR #3575, opened 2021-11-22
syncer(dm) : fix default collation with upstream in create table statement

WizardXiao:
fix https://github.com/pingcap/ticdc/issues/3420

in this case:
sql1: `create database if not exists test`
sql2: `create database test CHARACTER SET=utf8mb4` 
sql3: `create table test.t1 (id int)` 
sql4: `create table test.t1 (id int) CHARSET=utf8mb4`

if syncer send the same ddl to downstream tidb, database or table will has diffrent `collation` from upstream mysql, because the default collation is different between tidb and mysql. 

it will adjust collation when sync `create database` or `create table` statement.

for database, it will add `collation` by binlog query event and handle ddl like these:
sql1: `create database if not exists test COLLATE = utf8mb4_general_ci`
sql2: `create database test CHARACTER SET=utf8mb4 COLLATE = utf8mb4_general_ci` 

for table has no `charset` and `collation` , as we can not get `collation` from binlog, so there will have issue about timeliness, because dm can set start position at increment mode. so we will check and add warning log in dm. 
in this case, downstream db will use `collation` from database for table, so user can use dm sync dabase ddl sql, but if not, user should make sure dabase's `collation` is right if create dabase manually. so the ddl will be handled like that:
sql3: `create table test.t1 (id int)` and warning log `detect create table risk which use implicit charset and collation`

for table has `charset` but no `collation`, we will add default collation by `SHOW CHARACTER SET`, so the dll will be handled like that:
sql4: `create table test.t1 (id int) CHARSET=utf8mb4 COLLATE = utf8mb4_general_ci`

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
Fix a bug when upstream use implicit collation in create database or create table. 
It will add collation when there is no charset and collation in create database and it will not add collation in create table as table can inherit database. 
By the way, it will not directly alter downstream schema, so that user need make sure consistence when create database or table manually.
```


dm/pkg/utils/db.go
```diff
@@ -632,3 +632,31 @@ func GetTableCreateSQL(ctx context.Context, conn *sql.Conn, tableID string) (sql
 	}
 	return createStr, nil
 }
+
+// GetTableCollation gets table collation info by 'SHOW TABLE STATUS'.
+// See. https://dev.mysql.com/doc/refman/5.7/en/show-table-status.html
```
Ehco1996, line 637:
```suggestion
```
maybe show the resp from db [like other func](https://github.com/pingcap/ticdc/blob/0076d097b2ab252ca81b7a05430c90f34b73754d/dm/pkg/utils/db.go#L193) is better than give a link here

dm/pkg/utils/db.go
```diff
@@ -632,3 +632,31 @@ func GetTableCreateSQL(ctx context.Context, conn *sql.Conn, tableID string) (sql
...
+func GetTableCollation(ctx context.Context, db *sql.DB, schema string, table string) (info string, err error) {
+	querySQL := fmt.Sprintf("SHOW TABLE STATUS from %s like '%s'", schema, table)
```
Ehco1996, line 639: why not using `select table_collation from information_schema.Tables where table_schema="schema" and table_name="table";` ?

this can help reduce scanning logic(L647~L654)

dm/syncer/ddl.go
```diff
@@ -192,6 +192,49 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+func (s *Syncer) adjustTableCollation(ddlInfo *ddlInfo) {
+	if createStmt, ok := ddlInfo.originStmt.(*ast.CreateTableStmt); ok {
+		// create table like old table has no table_options
+		// See https://dev.mysql.com/doc/refman/5.7/en/create-table.html
+		if createStmt.ReferTable != nil {
+			fmt.Printf("RefTable")
+			return
+		}
+		for _, tableOption := range createStmt.Options {
+			fmt.Printf("Collation == %v \n", tableOption)
```
Ehco1996, line 204: please  remove  the dbebug print

dm/syncer/syncer.go
```diff
@@ -2298,6 +2298,8 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
 		if err2 != nil {
 			return err2
 		}
+		// adjust 'Collation' base on upstream
+		s.adjustTableCollation(ddlInfo)
```
Ehco1996, line 2302: how about batch add collation option to those ddl sql after this loop?

dm/syncer/ddl.go
```diff
@@ -192,6 +192,49 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+			// already have 'Collation'
+			if tableOption.Tp == ast.TableOptionCollate {
+				return
+			}
+		}
+		sourceTable := ddlInfo.sourceTables[0]
+		collationInfo, err := utils.GetTableCollation(s.tctx.Ctx, s.fromDB.BaseDB.DB, sourceTable.Schema, sourceTable.Name)
+		if err != nil {
+			fmt.Printf("err == %v \n", err)
+			s.tctx.L().Warn("get source table collation info failed.", zap.Stringer("table", sourceTable), zap.Error(err))
```
Ehco1996, line 214: i perfer not mute this err, seems this err only happen when wo lost connection whit upstream ?

dm/pkg/utils/db.go
WizardXiao: yes,  i will change to `information_schema`. I was worried about access `information_schema` rights, but now it seems `show table status` is the same after search mysql reference.
https://dev.mysql.com/doc/refman/8.0/en/information-schema-introduction.html#information-schema-privileges

dm/pkg/utils/db.go
WizardXiao: ok

dm/syncer/ddl.go
WizardXiao: ok

dm/syncer/ddl.go
WizardXiao: ok, i will return this err.

dm/syncer/syncer.go
WizardXiao: yes, if it can merge collation search into one statement, it may improve performance. i think i will try later.

dm/pkg/utils/db.go
```diff
@@ -632,3 +632,20 @@ func GetTableCreateSQL(ctx context.Context, conn *sql.Conn, tableID string) (sql
 	}
 	return createStr, nil
 }
+
+// GetTableCollation gets table collation info by 'SELECE TABLE_COLLATION FROM INFORMATION_SCHEMA.TABLES'.
+func GetTableCollation(ctx context.Context, db *sql.DB, schema string, table string) (info string, err error) {
```
lance6716, line 637: use wrapped DB to retry on network problem

dm/syncer/ddl.go
```diff
@@ -192,6 +192,45 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
 	return nil
 }
 
+func (s *Syncer) adjustTableCollation(ddlInfo *ddlInfo) error {
+	if createStmt, ok := ddlInfo.originStmt.(*ast.CreateTableStmt); ok {
```
lance6716, line 196: if ... !ok {
return
}

rest logic

dm/syncer/ddl.go
```diff
@@ -192,6 +192,45 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+			return nil
+		}
+		for _, tableOption := range createStmt.Options {
+			// already have 'Collation'
+			if tableOption.Tp == ast.TableOptionCollate {
+				return nil
+			}
+		}
+		sourceTable := ddlInfo.sourceTables[0]
+		collationInfo, err := utils.GetTableCollation(s.tctx.Ctx, s.fromDB.BaseDB.DB, sourceTable.Schema, sourceTable.Name)
```
lance6716, line 209:
```suggestion
		collation, err := utils.GetTableCollation(s.tctx.Ctx, s.fromDB.BaseDB.DB, sourceTable.Schema, sourceTable.Name)
```

dm/syncer/ddl.go
```diff
@@ -192,6 +192,45 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+			s.tctx.L().Error("get source table collation info failed.", zap.Stringer("table", sourceTable), zap.Error(err))
+			return err
+		}
+		if collationInfo == "" {
+			return nil
+		}
+		// add collation
+		if createStmt.Options == nil {
+			createStmt.Options = make([]*ast.TableOption, 1)
+		}
```
lance6716, line 220:
```suggestion
```

dm/syncer/ddl.go
```diff
@@ -192,6 +192,45 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+		createStmt.Options = append(createStmt.Options, &ast.TableOption{Tp: ast.TableOptionCollate, StrValue: collationInfo})
+		routedDDL, err := parserpkg.RenameDDLTable(createStmt, ddlInfo.targetTables)
```
lance6716, line 222: we can move invocation of `adjustTableCollation` inside `genDDLInfo` to avoid this

dm/syncer/ddl.go
WizardXiao: ok

dm/syncer/ddl.go
WizardXiao: ok

dm/pkg/utils/db.go
WizardXiao: ok, i will try.

dm/syncer/ddl.go
WizardXiao: ok

dm/syncer/ddl.go
WizardXiao: ok

dm/pkg/utils/db.go
```diff
@@ -632,3 +632,20 @@ func GetTableCreateSQL(ctx context.Context, conn *sql.Conn, tableID string) (sql
 	}
 	return createStr, nil
 }
+
+// GetTableCollation gets table collation info by 'SELECE TABLE_COLLATION FROM INFORMATION_SCHEMA.TABLES'.
```
dveeden, line 636:
```suggestion
// GetTableCollation gets table collation info by 'SELECT TABLE_COLLATION FROM INFORMATION_SCHEMA.TABLES'.
```

dm/syncer/ddl.go
lance6716: There's nothing to batch. If the query event is CREATE TABLE, after processing there should be at most one DDL to handle 🤔

lance6716 (commented):
also we should check dump/load phase will correctly handle collation.

lance6716:
by `SELECT TABLE_COLLATION FROM INFORMATION_SCHEMA.TABLES` we get currently collation of the table, but what we want is get collation at the replication time. (maybe if we continue replication we'll see alter collations?) please check if we can achieve that.

we have done some work of get SQL mode from the header of query event, maybe this would help
https://github.com/pingcap/ticdc/blob/7a5634bbc0b7b53f7fb70afe4c193fba6ab199b1/dm/pkg/binlog/event/util.go#L359

dm/pkg/utils/db.go
WizardXiao: ok

dm/syncer/ddl.go
WizardXiao: ok

Ehco1996: i think lance suggestion is better

Ehco1996:
> by `SELECT TABLE_COLLATION FROM INFORMATION_SCHEMA.TABLES` we get currently collation of the table, but what we want is get collation at the replication time. (maybe if we continue replication we'll see alter collations?) please check if we can achieve that.
> 
> we have done some work of get SQL mode from the header of query event, maybe this would help
> 
> https://github.com/pingcap/ticdc/blob/7a5634bbc0b7b53f7fb70afe4c193fba6ab199b1/dm/pkg/binlog/event/util.go#L359

seems we can get replication-time-connection from binlog QueryEvent ptal @WizardXiao 
 https://dev.mysql.com/doc/internals/en/query-event.html

Ehco1996:
> by `SELECT TABLE_COLLATION FROM INFORMATION_SCHEMA.TABLES` we get currently collation of the table, but what we want is get collation at the replication time. (maybe if we continue replication we'll see alter collations?) please check if we can achieve that.
> 
> we have done some work of get SQL mode from the header of query event, maybe this would help
> 
> https://github.com/pingcap/ticdc/blob/7a5634bbc0b7b53f7fb70afe4c193fba6ab199b1/dm/pkg/binlog/event/util.go#L359

seems we can get replication-time-connection from binlog QueryEvent ptal @WizardXiao 
 https://dev.mysql.com/doc/internals/en/query-event.html

WizardXiao:
> > by `SELECT TABLE_COLLATION FROM INFORMATION_SCHEMA.TABLES` we get currently collation of the table, but what we want is get collation at the replication time. (maybe if we continue replication we'll see alter collations?) please check if we can achieve that.
> > we have done some work of get SQL mode from the header of query event, maybe this would help
> > https://github.com/pingcap/ticdc/blob/7a5634bbc0b7b53f7fb70afe4c193fba6ab199b1/dm/pkg/binlog/event/util.go#L359
> 
> seems we can get replication-time-connection from binlog QueryEvent ptal @WizardXiao https://dev.mysql.com/doc/internals/en/query-event.html

ok, tanks, i will try.

lichunzhu (commented):
Do we need to do the same thing to columns and databases?

WizardXiao:
> Do we need to do the same thing to columns and databases?

i think this pr will only support tables and if anything go well, we may think about columns and databases later.

WizardXiao:
After some test about mysql binlog, there is a problem when we use `Binlog::QUERY_EVENT:status_vars` to make sure the timeliness of `Collation`, because there is no `Q_CHARSET_DATABASE_CODE` which is characterset and collation of the schema in Mysql 8.0.26. so maybe we need diffrent way to handle `full` or `incremental` phase, and maybe will sacrifice some timeliness.

Ehco1996:
> because there is no Q_CHARSET_DATABASE_CODE which is characterset and collation of the schema in Mysql 8.0.26. 

could you please add the doc link here? i can not find this in mysql doc 😅

>  we need diffrent way to handle handle full or incremental phase

this is also confuses me, in full pahse, loader directy execute create-table-sql  that exported by dumpling with the `collation`  from source 🤔, so we just need to adjust `collation` in  incremental phase

lance6716:
> this is also confuses me, in full pahse, loader directy execute create-table-sql that exported by dumpling with the `collation` from source 🤔, so we just need to adjust `collation` in incremental phase

SHOW CREATE TABLE will not write collation when it's the default collation of the charset

Ehco1996:
> > this is also confuses me, in full pahse, loader directy execute create-table-sql that exported by dumpling with the `collation` from source 🤔, so we just need to adjust `collation` in incremental phase
> 
> SHOW CREATE TABLE will not write collation when ~it's the default collation of the charset~. please see below comments
> 
> https://github.com/mysql/mysql-server/blob/3290a66c89eb1625a7058e0ef732432b6952b435/sql/sql_show.cc#L1983-L1993

thanks!  i think maybe adjust `charset` and `collation` for dumpling and syncer with `INFORMATION_SCHEMA.TABLES` is enough?

* dumpling: there is no need to get  `collation-at-the-replication-time`
* syncer: when a `create-table-sql-without-collation` comes from binlog, use the current collation of the upstream table as the default collation there does not seem to be a problem with the correctness

glorv:
@WizardXiao I think this PR deserve a release note as it's an important bugfix.

WizardXiao:
> @WizardXiao I think this PR deserve a release note as it's an important bugfix.

ok

dm/pkg/binlog/event/util.go
```diff
@@ -354,6 +354,22 @@ func GetParserForStatusVars(statusVars []byte) (*parser.Parser, error) {
 	return parser2, err
 }
 
+// GetServerCollationByStatusVars gets server collation by binlog statusVars.
+func GetServerCollationByStatusVars(statusVars []byte) (string, error) {
+	vars, err := statusVarsToKV(statusVars)
+	b, ok := vars[QCharsetCode]
```
Ehco1996, line 360: maybe check err!=nil first?

dm/syncer/ddl.go
```diff
@@ -192,6 +196,52 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+				hasCharset = true
+			}
+		}
+
+		// just has charset, can not add collation by server collation
+		if hasCharset {
+			tctx.L().Warn("detect create database risk which use implicit collation", zap.String("originSQL", ddlInfo.originDDL))
+			return
+		}
+
```
Ehco1996, line 234: should we add a warning log when no charse and no collation?

dm/syncer/ddl_test.go
```diff
@@ -637,6 +637,57 @@ func (s *testDDLSuite) TestClearOnlineDDL(c *C) {
 	c.Assert(mock.toFinish, HasLen, 0)
 }
 
+func (s *testDDLSuite) TestAdjustTableCollation(c *C) {
```
Ehco1996, line 640:
```suggestion
func (s *testDDLSuite) TestAdjustCollation(c *C) {
```

dm/syncer/syncer.go
```diff
@@ -2285,7 +2285,7 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
 	// TODO: save stmt, tableName to avoid parse the sql to get them again
 	qec.p = parser.New()
 	for _, sql := range qec.splitDDLs {
-		sqls, err2 := s.processOneDDL(qec, sql)
+		sqls, err2 := s.processOneDDL(qec, sql, ev.StatusVars)
```
Ehco1996, line 2288: maybe add `ev.StatusVars` to `qec` is better?

Ehco1996 (commented):
just a question:
dose the adjust of the `createdatabae` in the full phase is placed in the dumpling? 
If so, I can help review it when you raise the PR

WizardXiao:
> just a question: dose the adjust of the `createdatabae` in the full phase is placed in the dumpling? If so, I can help review it when you raise the PR

yes, the full phase is in dumpling and  PR is https://github.com/pingcap/tidb/pull/30292.

dm/pkg/binlog/event/util.go
WizardXiao: it reference `getSQLMode`, maybe its original intention is try to get `QCharsetCode` even has err, because the err may occur in parse other field in `statusVars`.
https://github.com/pingcap/ticdc/blob/master/dm/pkg/binlog/event/util.go#L331

dm/syncer/ddl_test.go
WizardXiao: ok

dm/syncer/syncer.go
WizardXiao: yes, it is more appropriate.

dm/pkg/binlog/event/util.go
```diff
@@ -354,6 +354,22 @@ func GetParserForStatusVars(statusVars []byte) (*parser.Parser, error) {
...
+
+	if !ok {
+		if err == nil {
+			// only happen when this is a dummy event generated by DM
+			err = fmt.Errorf("Q_CHARSET_CODE not found in status_vars %v", statusVars)
+		}
+		return mysql.UTF8MB4DefaultCollation, err
```
lance6716, line 367: when we can't find the collation, I prefer we return the default value of MySQL not TiDB, which is "utf8mb4_general_ci"

https://docs.pingcap.com/tidb/stable/character-set-and-collation

dm/syncer/ddl.go
```diff
@@ -192,6 +196,53 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+			return
+		}
+		for _, tableOption := range createStmt.Options {
+			// already have 'Collation'
+			if tableOption.Tp == ast.TableOptionCollate {
+				return
+			}
+		}
+		tctx.L().Warn("detect create table risk which use implicit collation", zap.String("originSQL", ddlInfo.originDDL))
+	} else if createStmt, ok := ddlInfo.originStmt.(*ast.CreateDatabaseStmt); ok {
```
lance6716, line 215: for current code we could use type switches

dm/syncer/ddl.go
```diff
@@ -192,6 +196,53 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+			// already have 'Collation'
+			if createOption.Tp == ast.DatabaseOptionCollate {
+				return
+			}
+			if createOption.Tp == ast.DatabaseOptionCharset {
+				hasCharset = true
+			}
+		}
+
+		// just has charset, can not add collation by server collation
```
lance6716, line 229: we should also add explicit collation because the default collation of a charset in TiDB is not same with MySQL

dm/pkg/binlog/event/util.go
WizardXiao: ok

lance6716:
please paste your manually test steps in PR description, or link to some doc

rest lgtm

dm/syncer/ddl.go
```diff
@@ -192,6 +196,53 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+		if hasCharset {
+			tctx.L().Warn("detect create database risk which use explicit charset and implicit collation", zap.String("originSQL", ddlInfo.originDDL))
+			return
+		}
+
+		// add collation by server collation from binlog statusVars
+		collation, err := event.GetServerCollationByStatusVars(statusVars)
+		if err != nil {
+			tctx.L().Warn("found error when get collation_server from binlog status_vars", zap.Error(err))
```
Ehco1996, line 238:
```suggestion
			tctx.L().Warn("found error when get collation_server from binlog status_vars, auto use default collation of mysql", zap.Error(err))
```

Ehco1996 (commented):
rest LGTM

dm/pkg/utils/db.go
```diff
@@ -391,6 +391,40 @@ func GetServerUnixTS(ctx context.Context, db *sql.DB) (int64, error) {
...
+		+----------+---------------------------------+---------------------+--------+
+		| Charset  | Description                     | Default collation   | Maxlen |
+		+----------+---------------------------------+---------------------+--------+
+		| armscii8 | ARMSCII-8 Armenian              | armscii8_general_ci |      1 |
+		| ascii    | US ASCII                        | ascii_general_ci    |      1 |
+		| big5     | Big5 Traditional Chinese        | big5_chinese_ci     |      2 |
+		| binary   | Binary pseudo charset           | binary              |      1 |
+		| cp1250   | Windows Central European        | cp1250_general_ci   |      1 |
+		| cp1251   | Windows Cyrillic                | cp1251_general_ci   |      1 |
+	*/if err != nil {
```
okJiang, line 410:
```suggestion
	*/
	if err != nil {
```

WizardXiao:
> please paste your manually test steps in PR description, or link to some doc
> 
> rest lgtm

ok

dm/pkg/binlog/event/util.go
```diff
@@ -354,6 +354,22 @@ func GetParserForStatusVars(statusVars []byte) (*parser.Parser, error) {
...
+func GetServerCollationByStatusVars(statusVars []byte) (string, error) {
+	vars, err := statusVarsToKV(statusVars)
+	b, ok := vars[QCharsetCode]
+
+	if !ok {
+		if err == nil {
+			// only happen when this is a dummy event generated by DM
+			err = fmt.Errorf("Q_CHARSET_CODE not found in status_vars %v", statusVars)
+		}
+		return mysql.Collations[45], err
```
lance6716, line 367: should be "latin1_swedish_ci", which is [8] :thinking: 

https://dev.mysql.com/doc/refman/5.7/en/charset.html

dm/pkg/utils/db.go
```diff
@@ -391,6 +391,40 @@ func GetServerUnixTS(ctx context.Context, db *sql.DB) (int64, error) {
 	return ts, err
 }
 
+// GetCharsetAndDefaultCollation gets charset and default collation map.
+func GetCharsetAndDefaultCollation(ctx context.Context, db *sql.DB) (map[string]string, error) {
```
lance6716, line 395: please use wrapped connection, not sql.DB

dm/syncer/ddl.go
```diff
@@ -192,6 +196,76 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+			tctx.L().Warn("detect create table risk which use implicit charset and collation", zap.String("originSQL", ddlInfo.originDDL))
+			return
+		}
+		// just has charset, can add collation by charset and default collation map
+		collation, ok := charsetAndDefaultCollationMap[strings.ToLower(justCharset)]
+		if !ok {
+			tctx.L().Error("not found charset default collation.", zap.String("originSQL", ddlInfo.originDDL), zap.String("charset", strings.ToLower(justCharset)))
+			return
+		}
+		tctx.L().Warn("detect create table risk which use explicit charset and implicit collation, we will add collation by SHOW CHARACTER SET", zap.String("originSQL", ddlInfo.originDDL), zap.String("collation", collation))
```
lance6716, line 228: no need to print a log at correct path, at least not WARN level :thinking:

dm/syncer/ddl.go
```diff
@@ -192,6 +196,76 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+		createStmt.Options = append(createStmt.Options, &ast.TableOption{Tp: ast.TableOptionCollate, StrValue: collation})
+
+	case *ast.CreateDatabaseStmt:
+		// collation is in create database syntax create_option
```
lance6716, line 232: I think the syntax is too common sense so no need to add a comment for it.

dm/syncer/ddl.go
```diff
@@ -192,6 +196,76 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+
+		// just has charset, can add collation by charset and default collation map
+		if justCharset != "" {
+			tmpCollation, ok := charsetAndDefaultCollationMap[strings.ToLower(justCharset)]
+			if !ok {
+				tctx.L().Error("not found charset default collation.", zap.String("originSQL", ddlInfo.originDDL), zap.String("charset", strings.ToLower(justCharset)))
+				return
+			}
+			collation = tmpCollation
+			tctx.L().Warn("detect create database risk which use explicit charset and implicit collation, we will add collation by SHOW CHARACTER SET", zap.String("originSQL", ddlInfo.originDDL), zap.String("collation", collation))
```
lance6716, line 253: ditto

dm/syncer/ddl.go
```diff
@@ -192,6 +196,76 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+			tctx.L().Error("not found charset default collation.", zap.String("originSQL", ddlInfo.originDDL), zap.String("charset", strings.ToLower(justCharset)))
+			return
+		}
+		tctx.L().Warn("detect create table risk which use explicit charset and implicit collation, we will add collation by SHOW CHARACTER SET", zap.String("originSQL", ddlInfo.originDDL), zap.String("collation", collation))
+		createStmt.Options = append(createStmt.Options, &ast.TableOption{Tp: ast.TableOptionCollate, StrValue: collation})
+
+	case *ast.CreateDatabaseStmt:
+		// collation is in create database syntax create_option
+		// See https://dev.mysql.com/doc/refman/8.0/en/create-database.html
+		var justCharset, collation string
```
lance6716, line 234: var err error
var ok bool

then we don't need `tmpCollation`

lance6716 (commented):
rest lgtm

dm/pkg/utils/db.go
WizardXiao: ok

dm/syncer/syncer.go
```diff
@@ -2318,7 +2320,7 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
 			continue
 		}
 		// We use default parser because sqls are came from above *Syncer.splitAndFilterDDL, which is StringSingleQuotes, KeyWordUppercase and NameBackQuotes
-		ddlInfo, err2 := s.genDDLInfo(qec.p, qec.ddlSchema, sql)
+		ddlInfo, err2 := s.genDDLInfo(qec.p, qec.ddlSchema, sql, qec.eventStatusVars)
```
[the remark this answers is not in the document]
WizardXiao: yes, it is better. i will fix and fix uts.

dm/syncer/ddl.go
WizardXiao: ok, it will be add by `SHOW CHARACTER SET`.

dm/syncer/ddl.go
WizardXiao: ok

dm/pkg/utils/db.go
WizardXiao: ok

dm/pkg/utils/db.go
lance6716:
```suggestion
	*/
	if err != nil {
```

dm/pkg/utils/db.go
```diff
@@ -391,6 +391,46 @@ func GetServerUnixTS(ctx context.Context, db *sql.DB) (int64, error) {
...
+		| ascii    | US ASCII                        | ascii_general_ci    |      1 |
+		| big5     | Big5 Traditional Chinese        | big5_chinese_ci     |      2 |
+		| binary   | Binary pseudo charset           | binary              |      1 |
+		| cp1250   | Windows Central European        | cp1250_general_ci   |      1 |
+		| cp1251   | Windows Cyrillic                | cp1251_general_ci   |      1 |
+		+----------+---------------------------------+---------------------+--------+
+	*/if err != nil {
+		return nil, terror.DBErrorAdapt(err, terror.ErrDBDriverError)
+	}
+	if rows.Err() != nil {
```
lance6716, line 419: check the comment of the function, it says "during iteration", so no need to check it here.

And the error of `rows.Close` is same as `rows.Err()`, if you want to check you can use `rows.Close`.

lance6716 (commented):
rest lgtm

dm/pkg/utils/db.go
WizardXiao: ok

dveeden:
I assume this works fine with the [New framework for collations](https://docs.pingcap.com/tidb/stable/character-set-and-collation#new-framework-for-collations) enabled? Do we need to add tests for that?

WizardXiao:
> I assume this works fine with the [New framework for collations](https://docs.pingcap.com/tidb/stable/character-set-and-collation#new-framework-for-collations) enabled? Do we need to add tests for that?
Current approach will sync collation to tidb and will not check collation conf in tidb, so if `new_collation_enabled` is false, tidb may throw error, and user can set  `new_collation_enabled` be true.

dm/pkg/terror/error_list.go
```diff
@@ -1071,6 +1072,7 @@ var (
 	ErrSyncerParseDDL                       = New(codeSyncerParseDDL, ClassSyncUnit, ScopeInternal, LevelHigh, "parse DDL: %s", "Please confirm your DDL statement is correct and needed. For TiDB compatible DDL, see https://docs.pingcap.com/tidb/stable/mysql-compatibility#ddl. You can use `handle-error` command to skip or replace the DDL or add a binlog filter rule to ignore it if the DDL is not needed.")
 	ErrSyncerUnsupportedStmt                = New(codeSyncerUnsupportedStmt, ClassSyncUnit, ScopeInternal, LevelHigh, "`%s` statement not supported in %s mode", "")
 	ErrSyncerGetEvent                       = New(codeSyncerGetEvent, ClassSyncUnit, ScopeUpstream, LevelHigh, "get binlog event error: %v", "Please check if the binlog file could be parsed by `mysqlbinlog`.")
+	ErrSyncerGetCharsetMap                  = New(codeSyncerGetCharsetMap, ClassSyncUnit, ScopeUpstream, LevelHigh, "get upstream charset and default collation error", "Please check if upstream is normal.")
```
lonng, line 1075:
```suggestion
	ErrSyncerGetCharsetMap                  = New(codeSyncerGetCharsetMap, ClassSyncUnit, ScopeUpstream, LevelHigh, "retrieve upstream charset and default collation error", "Please check if upstream is normal.")
```

Prefer to use `retrieve` instead of `get`, see: https://wikidiff.com/retrieve/get

lance6716 (approved):
> I assume this works fine with the New framework for collations enabled? Do we need to add tests for that?

This PR only handle the increment phase, https://github.com/pingcap/tidb/pull/30292 handles dump/load phase. A testing PR will be submitted after including them two.

dm/pkg/terror/error_list.go
WizardXiao: ok, it is better.

dm/pkg/binlog/event/util.go
```diff
@@ -354,6 +354,23 @@ func GetParserForStatusVars(statusVars []byte) (*parser.Parser, error) {
...
+	vars, err := statusVarsToKV(statusVars)
+	b, ok := vars[QCharsetCode]
+
+	if !ok {
+		if err == nil {
+			// only happen when this is a dummy event generated by DM
+			err = fmt.Errorf("Q_CHARSET_CODE not found in status_vars %v", statusVars)
+		}
+		// mysql default 'latin1_swedish_ci'
+		return mysql.Collations[8], err
```
glorv, line 368:
```suggestion
		return 'latin1_swedish_ci', err
```
`8` is a inner constant in tidb that should not be rely on.  The proper way to do this is to use `mysql.Collations[mysql.CollationNames['latin1_swedish_ci']]`, thus why not `latin1_swedish_ci` directly.

dm/pkg/terror/error_list.go
```diff
@@ -433,6 +433,7 @@ const (
 	codeSyncerParseDDL
 	codeSyncerUnsupportedStmt
 	codeSyncerGetEvent
+	codeSyncerGetCharsetMap
```
glorv, line 436: I think we needn't add this extra error since in most case the cause fetch charset-map error  should be target database internal error or sql connect problem. Return general sql error should be enough.

dm/syncer/ddl.go
```diff
@@ -192,6 +195,72 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+			}
+		}
+		if justCharset == "" {
+			tctx.L().Warn("detect create table risk which use implicit charset and collation", zap.String("originSQL", ddlInfo.originDDL))
+			return
+		}
+		// just has charset, can add collation by charset and default collation map
+		collation, ok := charsetAndDefaultCollationMap[strings.ToLower(justCharset)]
+		if !ok {
+			tctx.L().Error("not found charset default collation.", zap.String("originSQL", ddlInfo.originDDL), zap.String("charset", strings.ToLower(justCharset)))
```
glorv, line 222:
```suggestion
			tctx.L().Warn("not found charset default collation.", zap.String("originSQL", ddlInfo.originDDL), zap.String("charset", strings.ToLower(justCharset)))
```

dm/syncer/ddl.go
```diff
@@ -192,6 +195,72 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+			if createOption.Tp == ast.DatabaseOptionCharset {
+				justCharset = createOption.Value
+			}
+		}
+
+		// just has charset, can add collation by charset and default collation map
+		if justCharset != "" {
+			collation, ok = charsetAndDefaultCollationMap[strings.ToLower(justCharset)]
+			if !ok {
+				tctx.L().Error("not found charset default collation.", zap.String("originSQL", ddlInfo.originDDL), zap.String("charset", strings.ToLower(justCharset)))
```
glorv, line 246:
```suggestion
				tctx.L().Warn("not found charset default collation.", zap.String("originSQL", ddlInfo.originDDL), zap.String("charset", strings.ToLower(justCharset)))
```

dm/syncer/ddl.go
```diff
@@ -192,6 +195,72 @@ func (s *Syncer) clearOnlineDDL(tctx *tcontext.Context, targetTable *filter.Tabl
...
+				return
+			}
+			tctx.L().Info("detect create database risk which use explicit charset and implicit collation, we will add collation by SHOW CHARACTER SET", zap.String("originSQL", ddlInfo.originDDL), zap.String("collation", collation))
+		} else {
+			// has no charset and collation
+			// add collation by server collation from binlog statusVars
+			collation, err = event.GetServerCollationByStatusVars(statusVars)
+			if err != nil {
+				tctx.L().Warn("found error when get collation_server from binlog status_vars, auto use default collation of mysql", zap.Error(err))
```
glorv, line 255: This shouldn't happen in a normal binlog event, so if this actually happened, I think we can just skip here.

dm/pkg/binlog/event/util.go
```diff
@@ -354,6 +354,23 @@ func GetParserForStatusVars(statusVars []byte) (*parser.Parser, error) {
...
+	}
+
+	return mysql.Collations[b[4]], err
```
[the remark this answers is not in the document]
WizardXiao: ok

dm/pkg/binlog/event/util.go
WizardXiao: ok

dm/pkg/terror/error_list.go
WizardXiao: ok, i will remove this error.

dm/syncer/ddl.go
WizardXiao: ok

dm/syncer/ddl.go
WizardXiao: ok

dm/syncer/ddl.go
WizardXiao: ok

dm/pkg/binlog/event/util.go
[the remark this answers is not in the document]
lichunzhu: `@@collation_server` seems not so useful
```sql
mysql> show create database test;
+----------+----------------------------------------------------------------+
| Database | Create Database                                                |
+----------+----------------------------------------------------------------+
| test      | CREATE DATABASE `test` /*!40100 DEFAULT CHARACTER SET utf8 */ |
+----------+----------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> use test;
Database changed
mysql> select @@collation_server;
+--------------------+
| @@collation_server |
+--------------------+
| latin1_swedish_ci  |
+--------------------+
1 row in set (0.00 sec)

mysql> SELECT DEFAULT_COLLATION_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = 'test';
+------------------------+
| DEFAULT_COLLATION_NAME |
+------------------------+
| utf8_general_ci        |
+------------------------+
1 row in set (0.01 sec)

mysql> show create table t;
+-------+----------------------------------------------------------------------------------------+
| Table | Create Table                                                                           |
+-------+----------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `a` varchar(10) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8 |
+-------+----------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> show table status;
+------+--------+---------+------------+------+----------------+-------------+-----------------+--------------+-----------+----------------+---------------------+-------------+------------+-----------------+----------+----------------+---------+
| Name | Engine | Version | Row_format | Rows | Avg_row_length | Data_length | Max_data_length | Index_length | Data_free | Auto_increment | Create_time         | Update_time | Check_time | Collation       | Checksum | Create_options | Comment |
+------+--------+---------+------------+------+----------------+-------------+-----------------+--------------+-----------+----------------+---------------------+-------------+------------+-----------------+----------+----------------+---------+
| t    | InnoDB |      10 | Dynamic    |    0 |              0 |       16384 |               0 |            0 |         0 |           NULL | 2021-12-06 07:02:07 | NULL        | NULL       | utf8_general_ci |     NULL |                |         |
+------+--------+---------+------------+------+----------------+-------------+-----------------+--------------+-----------+----------------+---------------------+-------------+------------+-----------------+----------+----------------+---------+
1 row in set (0.01 sec)
```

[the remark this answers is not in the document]
lichunzhu: I have checked `StatusVars`. It stored `latin1_swedish_ci` instead of `utf8_general_ci`.

glorv (approved):

PR #3576, opened 2021-11-22
pkg/p2p: streamHandle for implementing MessageServer

liuzix:
- `grpc-go` does not provide a thread-safe implementation of `ServerStream`, so we need to forward all messages to a fixed goroutine if we want to send messages from multiple goroutines.
- We want a data structure that supports gracefully closing a channel from the receiver's end.

- Implemented `streamHandle`, which is a specialized data structure for the use of `MessageServer` (in separate PR) in `pkg/p2p`.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


pkg/p2p/server_stream_handle_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	time.Sleep(100 * time.Millisecond)
+	h.Close()
+
+	// Tests double close.
+	// Should not panic.
+	h.Close()
+}
+
+func TestGetStreamMeta(t *testing.T) {
```
overvenus, line 97: Can it be parallel?

overvenus (commented):
rest LGTM

pkg/p2p/server_stream_handle.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/pingcap/ticdc/proto/p2p"
+	"go.uber.org/atomic"
+)
+
+// streamHandle is used to provide the server a non-blocking handle of a stream.
+// It is needed because grpc-go does not provide a stream easily cancellable.
+// Since Go is not friendly with the receiver closing a channel, we need this
+// auxiliary data structure to make the stream cancellation more graceful.
+type streamHandle struct {
+	mu     sync.RWMutex
```
amyangfei, line 31: Better to add comments about which variables are being protected by `mu`, I find the protected variables are not same in `Send` and `Close`.

pkg/p2p/server_stream_handle.go
```diff
@@ -0,0 +1,89 @@
...
+		sendCh:     sendCh,
+		closeCh:    make(chan struct{}),
+		streamMeta: meta,
+	}
+}
+
+// Send sends a message to the stream.
+func (s *streamHandle) Send(ctx context.Context, response p2p.SendMessageResponse) error {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
```
amyangfei, line 53: Should s.mu use `Lock` here, L62 writes to `s.sendCh`

pkg/p2p/server_stream_handle.go
```diff
@@ -0,0 +1,89 @@
...
+	case s.sendCh <- response:
+	case <-s.closeCh:
+		return cerror.ErrPeerMessageInternalSenderClosed.GenWithStackByArgs()
+	}
+
+	return nil
+}
+
+// Close closes the stream handle.
+// We should not call Send after Close.
```
amyangfei, line 71: > We should not call Send after Close.
Is this ensured by code or mechanism in `streamHandle` itself

liuzix: This is an advise for the caller. The `streamHandle` itself returns an error too.

pkg/p2p/server_stream_handle.go
liuzix: We consider `R` to be anything except closing the channel. Read and write are both `R`, and `close` is `W` in this definition. This definition provides an elegant solution for lock protecting go channels.

pkg/p2p/server_stream_handle.go
liuzix: Technically it protects only `sendCh`.

amyangfei (approved):

pkg/p2p/server_stream_handle.go
```diff
@@ -0,0 +1,96 @@
...
+	"go.uber.org/atomic"
+)
+
+// streamHandle is used to provide the server a non-blocking handle of a stream.
+// It is needed because grpc-go does not provide a stream easily cancellable.
+// Since Go is not friendly with the receiver closing a channel, we need this
+// auxiliary data structure to make the stream cancellation more graceful.
+type streamHandle struct {
+	// mu protects sendCh. We should take the shared lock
+	// when sending and the exclusive lock when closing it.
```
3AceShowHand, line 32: please update this comment, just like said `We consider R to be anything except closing the channel. Read and write are both R, and close is W in this definition. This definition provides an elegant solution for lock protecting go channels.`

pkg/p2p/server_stream_handle.go
```diff
@@ -0,0 +1,96 @@
...
+// If called after Close, an error will be returned.
+func (s *streamHandle) Send(ctx context.Context, response p2p.SendMessageResponse) error {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	// We need to read from isClosed while holding the lock.
+	// Otherwise, we risk the sendCh being closed after we have checked
+	// isClosed to be false. If this happens, we will try to send
+	// to a closed channel, which will panic.
+	if s.isClosed.Load() {
```
3AceShowHand, line 62: It looks that only senders which hold the R lock will hit this point, they will only send data, and not close the channel.

So once they found it's not closed, they can safely send the data.

pkg/p2p/server_stream_handle.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Close closes the stream handle.
+// We should not call Send after Close.
+func (s *streamHandle) Close() {
+	if s.isClosed.Swap(true) {
+		// already closed
+		return
+	}
+	// Must close `s.closeCh` while not holding `s.mu`.
```
3AceShowHand, line 84: Will the `Close` and `Send` be called concurrently?

If there are multiple senders hit the line69, and this one closes the channel without holding the exclusive lock, I think this would cause panic if without holding the writer lock.

pkg/p2p/server_stream_handle.go
```diff
@@ -0,0 +1,96 @@
...
+	close(s.closeCh)
```
3AceShowHand, line 85: how about close the `sendCh` first, then close the `closeCh`.

Do we need to set `isClosed` to true here ?

asddongmen (approved):

pkg/p2p/server_stream_handle.go
asddongmen: What if there are multiple goroutine call the Close()?

> how about close the `sendCh` first, then close the `closeCh`.
> 
> Do we need to set `isClosed` to true here ?

3AceShowHand: once they found L80 is true, only the first caller will close the `closeCh`

asddongmen: yes, so we need `to set isClosed to true here`

asddongmen: > how about close the `sendCh` first, then close the `closeCh`.
> 
> Do we need to set `isClosed` to true here ?
`s.isClosed.Swap(true) ` already set it to true, it `swap` and return the pre value.

pkg/p2p/server_stream_handle.go
liuzix: I was talking about what would happen if we did not hold the R lock at this point.

pkg/p2p/server_stream_handle.go
liuzix: > how about close the `sendCh` first, then close the `closeCh`.
> 
> Do we need to set `isClosed` to true here ?

Closing `sendCh` first will very likely cause a data race which leads to undefined behavior, even panics. Unsynchronized sending and closing will race in Go.

3AceShowHand (approved):

PR #3610, opened 2021-11-24
sorter/leveldb(cdc): add leveldb-like interfaces

overvenus:

Add leveldb-like interfaces, so that we can replace goleveldb with other databases easily.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/sorter/leveldb/db/db.go
```diff
@@ -0,0 +1,49 @@
...
+
+// DB is an interface of a leveldb-like database.
+type DB interface {
+	Snapshot() (Snapshot, error)
+	Batch(cap int) Batch
+	Close() error
+	CollectMetrics(captureAddr string, id int)
+}
+
+// A Batch is a sequence of Sets, Merges, Deletes, and/or DeleteRanges that are
```
hi-rustin, line 24: I don't quite understand the comment. Can you explain it?

cdc/sorter/leveldb/db/leveldb.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/syndtr/goleveldb/leveldb/opt"
+	"github.com/syndtr/goleveldb/leveldb/util"
+	"go.uber.org/zap"
+)
+
+// OpenDB opens a leveldb.
+func OpenDB(ctx context.Context, id int, cfg *config.SorterConfig) (DB, error) {
+	lcfg := cfg.LevelDB
+	var option opt.Options
+	option.OpenFilesCacheCapacity = lcfg.MaxOpenFiles / lcfg.Count
```
hi-rustin, line 39: Perhaps it would be better to initialize these parameters by abstracting them into a function?

overvenus: Do you mean builder pattern? `Builder().SetA().SetB().Build()`

cdc/sorter/leveldb/db/db.go
```diff
@@ -0,0 +1,49 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package db
```
amyangfei, line 14: Better to move the interface definition to separated path, instead of the sub-path of leveldb

amyangfei: Besides do we need a concrete package name, `db` is confused.

asddongmen:
rest LGTM

cdc/sorter/leveldb/db/db.go
overvenus: Any suggestion about the name?

overvenus: I'm going to move these files to pkg/db, and name it package db. Feel free to suggest a better name, thanks!

pkg/db/leveldb.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/syndtr/goleveldb/leveldb/iterator"
+	"github.com/syndtr/goleveldb/leveldb/opt"
+	"github.com/syndtr/goleveldb/leveldb/util"
+	"go.uber.org/zap"
+)
+
+// OpenDB opens a leveldb.
+func OpenDB(ctx context.Context, id int, path string, cfg *config.DBConfig) (DB, error) {
+	var option opt.Options
+	option.OpenFilesCacheCapacity = cfg.MaxOpenFiles / cfg.Count
```
hi-rustin, line 38: What I mean is that the process of collecting and completing options can be a separate function.

hi-rustin (approved):
rest LGTM

asddongmen (approved):

PR #3613, opened 2021-11-25
cdc/sink: Refine sink interface

sdojjy:
To fix the bug of #3503 , we should manage table checkpoint per table in the sink, this is PR is a part of the fix. In this PR, only change the interface method, no real logic is changed.

add table ID to Sink interface, below functions will be changed,
```
type Sink interface {
   FlushRowChangedEvents(ctx context.Context, tableID model.TableID, resolvedTs uint64) (uint64, error)
   Barrier(ctx context.Context, tableID model.TableID) error
}
```

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### Release note 

```release-note
 `None`.
```

### DIFF
--- a/cdc/owner/async_sink_test.go
+++ b/cdc/owner/async_sink_test.go
@@ -65,7 +65,7 @@ func (m *mockSink) Close(ctx context.Context) error {
 	return nil
 }
 
-func (m *mockSink) Barrier(ctx context.Context) error {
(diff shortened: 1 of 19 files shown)


overvenus (approved):

asddongmen (approved):

overvenus:
Find a flaky test #3626 .

PR #3621, opened 2021-11-25
cdc/sink: Sink manager manage checkpoint per table

sdojjy:
To fix the bug of #3503 , we should manage table checkpoint per table in the sink, this is PR is a part of the fix. In this PR, sink manager manage checkpointTs per table

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`.
```


cdc/sink/manager.go
```diff
@@ -145,8 +146,13 @@ func (m *Manager) destroyTableSink(ctx context.Context, tableID model.TableID) e
...
 
-func (m *Manager) getCheckpointTs() uint64 {
-	return atomic.LoadUint64(&m.checkpointTs)
+func (m *Manager) getCheckpointTs(tableID model.TableID) uint64 {
+	checkPoints, ok := m.tableCheckpointTsMap.Load(tableID)
+	if ok {
+		return checkPoints.(uint64)
+	}
+	// this function is call when something unexpected, return change level checkpoint is safe
+	return m.changeFeedCheckpointTs
```
overvenus, line 155: Could you return an error? Let's fail fast.

overvenus (approved):
Rest LGTM

cdc/sink/manager.go
sdojjy: Update the comments,
1. if the table flush task can not get the flush lock,  we return changeFeed level checpointTs,  the sinkNode will not update its checkpointTs if returned value is smaller than it.  
2. if error hanppened , sink manager call this func, and also returns the error

cdc/sink/manager.go
```diff
@@ -145,8 +146,14 @@ func (m *Manager) destroyTableSink(ctx context.Context, tableID model.TableID) e
...
-func (m *Manager) getCheckpointTs() uint64 {
-	return atomic.LoadUint64(&m.checkpointTs)
+func (m *Manager) getCheckpointTs(tableID model.TableID) uint64 {
+	checkPoints, ok := m.tableCheckpointTsMap.Load(tableID)
+	if ok {
+		return checkPoints.(uint64)
+	}
+	// cannot find table level checkpointTs because of no table level resolvedTs flush task finished successfully,
+	// for example: first time to flush resolvedTs but cannot get the flush lock, return changefeed level checkpointTs is safe
+	return m.changeFeedCheckpointTs
```
overvenus, line 156: When do we update the `changeFeedCheckepoinTs`? Will it cause checkpoint regression on grafana dashboard? Will it cause owner sending a regressed checkpoint to downstream?

sdojjy: Now we do not update the `changeFeedCheckepoinTs`,  the sinkNode will not update its checkpointTs if the returned value is smaller than it. and we can update it per tick if needed.
grafana and owner get a table's checkpointTs from sinkNode,  if I understand correctly,   grafana and the owner will not be impacted.

cdc/processor/processor.go
```diff
@@ -159,6 +159,7 @@ func (p *processor) tick(ctx cdcContext.Context, state *orchestrator.ChangefeedR
 	if !p.checkChangefeedNormal() {
 		return nil, cerror.ErrAdminStopProcessor.GenWithStackByArgs()
 	}
+	p.sinkManager.UpdateChangeFeedCheckpointTs(state.Status.CheckpointTs)
```
overvenus, line 162: Please add some comments.

overvenus (approved):

amyangfei:
Will these bug fixes be cherry-picked?

asddongmen (approved):

sdojjy:
> Will these bug fixes be cherry-picked?

no

PR #3623, opened 2021-11-25
pkg/p2p(cdc): implement MessageServer

liuzix:
### Reviewer Notes ###
- Code that needs detailed review is about 800 lines. Unit tests takes about 950 lines.

- We need to implement `MessageServer` for receiving peer messages.

- Implemented `MessageServer`, which corresponds to the right half of the picture below.

Screenshot from 2021-11-22 16-57-18

### Check List 

Tests 

 - Unit test

Code changes

### Release note 

```release-note
None
```


pkg/p2p/server.go
```diff
@@ -0,0 +1,831 @@
...
+		})
+
+		if len(acks) == 0 {
+			continue
+		}
+
+		peer.metricsAckCount.Inc()
+		err := peer.sender.Send(ctx, p2p.SendMessageResponse{
+			Ack:        acks,
+			ExitReason: p2p.ExitReason_OK,
```
amyangfei, line 309: Does `ExitReason_OK` mean everything is OK and don't need to exit?

pkg/p2p/server.go
```diff
@@ -0,0 +1,831 @@
...
+
+			log.Debug("skipping peer message",
+				zap.String("sender-id", sm.SenderId),
+				zap.String("topic", topic),
+				zap.Int64("skipped-Seq", entry.Sequence),
+				zap.Int64("last-ack", lastAck))
+			return nil
+		}
+
+		if lastAck != 0 && entry.Sequence > lastAck+1 {
```
amyangfei, line 396: Is `entry.Sequence == lastAck+1` guaranteed by
- client receivers `acks` from server, and sends message with `seq = latest ack + 1`
- server maintains the correct ack corresponding to each message.

pkg/p2p/server.go
liuzix: Yes

pkg/p2p/server.go
liuzix: `lastAck` here is actually the sequence number of the last message received, because we update the `ackManager` for each message. So in other words this is a requirement for the `seq` to be continuous, which can be guaranteed by the client locally.

pkg/p2p/server.go
```diff
@@ -0,0 +1,835 @@
...
+					zap.String("sender", task.sender.GetStreamMeta().SenderId),
+					zap.Int64("epoch", task.sender.GetStreamMeta().Epoch))
+				if err := m.registerPeer(ctx, task.sender, task.clientAddr); err != nil {
+					if cerror.ErrPeerMessageStaleConnection.Equal(err) || cerror.ErrPeerMessageDuplicateConnection.Equal(err) {
+						// These two errors should not affect other peers
+						if err1 := task.sender.Send(ctx, errorToRPCResponse(err)); err1 != nil {
+							return errors.Trace(err)
+						}
+						continue // to handling the next task
+					}
```
overvenus, line 263: Are they covered by tests?

pkg/p2p/server.go
```diff
@@ -0,0 +1,835 @@
...
+
+// handlePendingMessages must be called with `handlerLock` taken exclusively.
+func (m *MessageServer) handlePendingMessages(ctx context.Context, topic string) error {
+	for key, entries := range m.pendingMessages {
+		if key.Topic != topic {
+			continue
+		}
+
+		for _, entry := range entries {
+			if err := m.handleMessage(ctx, entry.StreamMeta, entry.Entry); err != nil {
```
overvenus, line 502: Are they covered by tests?

pkg/p2p/server.go
```diff
@@ -0,0 +1,835 @@
...
+	} else if cerror.ErrPeerMessageStaleConnection.Equal(err) {
+		return p2p.SendMessageResponse{
+			ExitReason:   p2p.ExitReason_STALE_CONNECTION,
+			ErrorMessage: err.Error(),
+		}
+	} else if cerror.ErrPeerMessageReceiverMismatch.Equal(err) {
+		return p2p.SendMessageResponse{
+			ExitReason:   p2p.ExitReason_CAPTURE_ID_MISMATCH,
+			ErrorMessage: err.Error(),
+		}
```
overvenus, line 823: Could you add some tests to cover these errors?

pkg/p2p/server.go
liuzix: Yes. It should have been. I have stabilized the test case `TestServerIncomingConnectionStale`

pkg/p2p/server.go
liuzix: They were to be covered by the `server_client_integration_test.go`. I have modified the unit test case so that these lines are covered.

amyangfei (approved):

pkg/p2p/metrics.go
```diff
@@ -19,6 +19,42 @@ import (
 )
 
 var (
+	serverStreamCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
ben1009, line 22: the grafana template will added later after the code is stable ?

ben1009 (approved):

pkg/p2p/server.go
```diff
@@ -0,0 +1,834 @@
...
+	gRPCPeer "google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+)
+
+// MessageServerConfig stores configurations for the MessageServer
+type MessageServerConfig struct {
+	// The maximum number of entries to be cached for topics with no handler registered
+	MaxTopicPendingCount int
+	// The maximum number of unhandled internal tasks for the main thread.
+	MaxPendingTaskCount int
```
3AceShowHand, line 44: Do we have to unify the naming convention of `MaxTopicPendingCount` and `MaxPendingTaskCount`.

How about rename `MaxPendingTaskCount` to `MaxTaskPendingCount`

pkg/p2p/server.go
```diff
@@ -0,0 +1,834 @@
...
+	// For a correct implementation of the handler, the time it needs
+	// to consume these messages is minimal, as the handler is not
+	// expected to block on channels, etc.
+	WaitUnregisterHandleTimeoutThreshold time.Duration
+}
+
+// cdcPeer is used to store information on one connected client.
+type cdcPeer struct {
+	PeerID string // unique ID of the client
+	Epoch  int64  // epoch is increased when the client retries.
```
3AceShowHand, line 72: can you add a comment about why we need to track the `Epoch` ?

pkg/p2p/server.go
```diff
@@ -0,0 +1,834 @@
...
+	// This can happen when the server is slow.
+	// The upper limit of pending messages is restricted by
+	// MaxTopicPendingCount in MessageServerConfig.
+	pendingMessages map[topicSenderPair][]pendingMessageEntry
+
+	acks *ackManager
+
+	// taskQueue is used to store internal tasks MessageServer
+	// needs to execute serially.
+	taskQueue chan interface{}
```
3AceShowHand, line 121: will you provide an interface for the `internal tasks`?

pkg/p2p/server.go
```diff
@@ -0,0 +1,834 @@
...
+	})
+
+	return errg.Wait()
+}
+
+func (m *MessageServer) run(ctx context.Context) error {
+	ticker := time.NewTicker(m.config.AckInterval)
+	defer ticker.Stop()
+
+	for {
```
3AceShowHand, line 199: does it possible to provide an interface for internal tasks? I think this `switch` is quite clumsy.

pkg/p2p/metrics.go
liuzix: Yes. Before releasing the peer-to-peer scheduling, Grafana panels will be added.

PR #3625, opened 2021-11-26
cdc/sink: buffer sink manage checkpoint ts per table

sdojjy:
To fix the bug of #3503 , we should manage table checkpoint per table in the sink, this is PR is a part of the fix. In this PR, buffer manage checkpointTs per table

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
 `None`.
```


cdc/sink/buffer_sink.go
```diff
@@ -150,8 +152,28 @@ func (b *bufferSink) EmitRowChangedEvents(ctx context.Context, rows ...*model.Ro
...
-	return atomic.LoadUint64(&b.checkpointTs), nil
+	return b.getCheckpointTs(tableID), nil
+}
+
+type flushMsg struct {
+	tableID    model.TableID
+	resolvedTs uint64
+}
+
+func (b *bufferSink) getCheckpointTs(tableID model.TableID) uint64 {
```
asddongmen, line 169: may be getTableCheckpointTs is a better name?
or add a comment here to describe the method return value.

cdc/sink/buffer_sink.go
```diff
@@ -150,8 +152,28 @@ func (b *bufferSink) EmitRowChangedEvents(ctx context.Context, rows ...*model.Ro
...
+	checkPoints, ok := b.tableCheckpointTsMap.Load(tableID)
+	if ok {
+		return checkPoints.(uint64)
+	}
+	return atomic.LoadUint64(&b.changeFeedCheckpointTs)
+}
+
+func (b *bufferSink) UpdateChangeFeedCheckpointTs(checkpointTs uint64) {
+	atomic.StoreUint64(&b.changeFeedCheckpointTs, checkpointTs)
```
asddongmen, line 178: i think it is better to add comments about who will call this method.

overvenus (commented):
Please add some tests

sdojjy:
> Please add some tests

done

overvenus (approved):
LGTM

asddongmen (approved):

Issue #3254, opened 2021-09-14
for `stop-task`, declarative API should wait the task is stoppped


lance6716:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

internally DM should retry and wait the task to stop, when `stop-task`, to implement a declarative API. currently we will return too early when the task has an error 

https://github.com/pingcap/dm/blob/39b5e2098f21260c14373a23069f7e38395d8d7f/dm/master/server.go#L1598-L1625

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

lance6716:
related to unstable CI

https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/241/pipeline/

https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/242/pipeline

fatelei:
let me try

lance6716:
@fatelei do you need our help?

fatelei:
> @fatelei do you need our help?

thanks, currently do need help.

PR #3640, opened 2021-11-26
master(dm): add method check worker status

fatelei:
#3254 

for op is stop-task, add method to check worker status

### Check List 

Tests 

 maybe not need a new tests?

### Related changes

- add method to check worker status when op is stop-task

### DIFF
--- a/dm/dm/master/server.go
+++ b/dm/dm/master/server.go
@@ -1681,6 +1681,10 @@ func (s *Server) waitOperationOk(
 					}
 				}
 			case *pb.StartTaskRequest, *pb.UpdateTaskRequest, *pb.OperateTaskRequest:
+				fmt.Printf("query response: %+v\n", queryResp.Result)
+				if expect == pb.Stage_Stopped && !queryResp.Result {
+					continue
+				}
 				if expect == pb.Stage_Stopped && len(queryResp.SubTaskStatus) == 0 {
 					return true, "", queryResp, nil
 				}


fatelei:
I can do tests in local pc ?

lance6716:
Hi, `stop-task` wants the task to be "not found (stopped)", there's nothing to do with worker status

dm/dm/master/server.go
```diff
@@ -2421,3 +2423,18 @@ func (s *Server) sharedLogic(ctx context.Context, req interface{}, respPointer i
 	*errPointer = terror.ErrMasterRequestIsNotForwardToLeader
 	return true
 }
+
+// checkWorkerStatus when op is stop-task, need check worker status.
```
Ehco1996, line 2427: we also use this func when `check whether source relative worker has been removed by scheduler` so please adjust this comment

dm/dm/master/server.go
```diff
@@ -1629,14 +1630,9 @@ func (s *Server) waitOperationOk(
...
 			if expect == pb.Stage_Stopped {
-				resp := &pb.QueryStatusResponse{
-					Result:       true,
-					SourceStatus: &pb.SourceStatus{Source: sourceID, Worker: cli.BaseInfo().Name},
-				}
-				if w := s.scheduler.GetWorkerByName(cli.BaseInfo().Name); w == nil {
-					return true, "", resp, nil
-				} else if cli.Stage() == scheduler.WorkerOffline {
-					return true, "", resp, nil
+				workerQueryResponse = s.checkWorkerStatus(sourceID, cli)
```
Ehco1996, line 1633:
```suggestion
				if workerQueryResponse = s.checkWorkerStatus(sourceID, cli); workerQueryResponse!=nil{
				    return xxx
			
				}
```

dm/dm/master/server.go
```diff
@@ -1697,8 +1696,11 @@ func (s *Server) waitOperationOk(
 							finished = pb.Stage_Finished
 						}
 						if expect == pb.Stage_Stopped {
+							workerQueryResponse = s.checkWorkerStatus(sourceID, cli)
 							if st, ok2 := subtaskStatus.Status.(*pb.SubTaskStatus_Msg); ok2 && st.Msg == dmcommon.NoSubTaskMsg(taskName) {
-								ok = true
+								if workerQueryResponse != nil {
```
Ehco1996, line 1701: semms we can mv L1699 here

dm/dm/master/server.go
```diff
@@ -1682,7 +1678,10 @@ func (s *Server) waitOperationOk(
 				}
 			case *pb.StartTaskRequest, *pb.UpdateTaskRequest, *pb.OperateTaskRequest:
 				if expect == pb.Stage_Stopped && len(queryResp.SubTaskStatus) == 0 {
-					return true, "", queryResp, nil
+					workerQueryResponse = s.checkWorkerStatus(sourceID, cli)
```
Ehco1996, line 1681: i am not sure if this fix https://github.com/pingcap/ticdc/issues/3254, could you mock `task has an error` in `TestWaitOperationOkRightResult` to verify this patch works fine?

PR #3642, opened 2021-11-26
fix docker file.

3AceShowHand:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/deployments/ticdc/docker-compose/docker-compose-kafka-integration.yml
+++ b/deployments/ticdc/docker-compose/docker-compose-kafka-integration.yml
@@ -2,13 +2,13 @@ version: '2.1'
 
 services:
   zookeeper:
-    image: wurstmeister/zookeeper
+    image: registry-mirror.pingcap.net/wurstmeister/zookeeper
     ports:
       - "2181:2181"
 
   kafka:
     container_name: kafka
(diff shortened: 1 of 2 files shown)


deployments/ticdc/docker-compose/docker-compose-kafka-integration.yml
```diff
@@ -2,13 +2,13 @@ version: '2.1'
 
 services:
   zookeeper:
-    image: wurstmeister/zookeeper
+    image: registry-mirror.pingcap.net/wurstmeister/zookeeper
```
hi-rustin, line 5: This is not consistent with the address used in CI, plus using it may not be friendly to non-Chinese network environments?

hi-rustin (commented):
Again, please complete the PR description when you submit your PR. Or set it to wip or drft for now. Thanks!

deployments/ticdc/docker-compose/docker-compose-kafka-integration.yml
hi-rustin: It looks like the external network cannot access this address.

PR #3647, opened 2021-11-28
deployments(ticdc): refine the network and volumes

hi-rustin:
None

use service network instead of container network.

And do not exit after test finished.

Otherwise it will ask to create the kafka container first.

### Check List 

Tests 

 - No code

Code changes

None

Side effects

None

Related changes

None

### Release note 

```release-note
None
```


hi-rustin:
/cc @asddongmen 

Can you help try it locally?

deployments/ticdc/docker-compose/docker-compose-kafka-integration.yml
```diff
@@ -38,5 +37,7 @@ services:
     command:
       - bash
       - -c
-      - make integration_test_kafka CASE=kafka_messages
-    network_mode: "container:kafka"
+      - make integration_test_kafka CASE=clustered_index & tail -f /dev/null
+    network_mode: "service:kafka"
```
Ehco1996, line 41: just a question: i can't find the differences between container_network and service_network 
it's hard to find the doc about this option, but i find [this](https://github.com/docker/docker.github.io/issues/9725#issuecomment-761832882) 

> Otherwise it will ask to create the kafka container first.

Is this the reason for modifying to service network? btw why change test case to `clustered_index `

hi-rustin: > Is this the reason for modifying to service network? 

Yes. I'm trying in codespaces and using the container it asks to create the network first. Can you help to try it locally?

> btw why change test case to clustered_index

I will think of a better way to solve this problem.

hi-rustin:
Usage:

```console
CASE="clustered_index" docker-compose -f ./deployments/ticdc/docker-compose/docker-compose-kafka-integration.yml up
```

deployments/ticdc/docker/integration-test.Dockerfile
```diff
@@ -55,5 +55,5 @@ COPY . .
 # We always need to clean before we build, please don't adjust its order.
 RUN make clean
 RUN make integration_test_build kafka_consumer cdc
-COPY --from=downloader /root/download/bin/* ./bin
+COPY --from=downloader /root/download/bin/* ./bin/
```
hi-rustin, line 58: Otherwise, it may report an error in some dockers.

deployments/ticdc/docker-compose/docker-compose-kafka-integration.yml
Ehco1996: > CASE="clustered_index" docker-compose -f ./deployments/ticdc/docker-compose/docker-compose-kafka-integration.yml up
```
Step 5/30 : RUN ./download-integration-test-binaries.sh master
...
  0     0    0     0    0     0      0      0 --:--:--  0:01:50 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:02:07 --:--:--     0curl: (7) Failed connect to fileserver.pingcap.net:80; Connection timed out
The command '/bin/sh -c ./download-integration-test-binaries.sh master' returned a non-zero code: 7
```

seems this test only can running in IDC 😅

hi-rustin: Yes, it can only be used in an internal network environment.

hi-rustin: I don't know how to fix this yet, maybe it's to use tiup's mirror link. But that doesn't seem to guarantee that we'll get the latest nightly builds. Also some packages may not be in the mirror.

deployments/ticdc/docker/integration-test.Dockerfile
hi-rustin: I haven't encountered it on my own computer, but @3AceShowHand  seems to have encountered the problem.

asddongmen (approved):

hi-rustin:
/assign @amyangfei 

Could you please help me trigger the merge process? Thanks!

PR #3655, opened 2021-11-29
processor, scheduler_v2(cdc): integrate the Processor with Scheduler(V2) Part I

liuzix:
- We need to integration the processor implementation with the new scheduler in `cdc/scheduler`.

- Minor modification to the `TableExecutor` interface to make it suit the current processor better.
- Implement `TableExecutor` for `processor`.
- Added the `SchedulerV2Enabled` switch in `pkg/config`, which is set to `false` for now.
- Prevent the `processor` from running some logic specific to the Etcd based scheduler, if `SchedulerV2Enabled == true`.

### Check List 

Tests 

 - Unit test

Code changes

### Release note 

```release-note
None
```


cdc/processor/processor.go
```diff
@@ -84,6 +92,127 @@ type processor struct {
...
+
+// IsRemoveTableFinished implements TableExecutor interface.
+func (p *processor) IsRemoveTableFinished(ctx cdcContext.Context, tableID model.TableID) bool {
+	if !p.checkReadyForMessages() {
+		return false
+	}
+
+	table, exist := p.tables[tableID]
+	if !exist {
+		log.Panic("table which was added is not found",
```
amyangfei, line 181:
```suggestion
		log.Panic("table which was removed is not found",
```

cdc/processor/processor.go
```diff
@@ -84,6 +92,127 @@ type processor struct {
...
+	return p.changefeed != nil && p.changefeed.Status != nil
+}
+
+// AddTable implements TableExecutor interface.
+func (p *processor) AddTable(ctx cdcContext.Context, tableID model.TableID) (bool, error) {
+	if !p.checkReadyForMessages() {
+		return false, nil
+	}
+
+	log.Info("adding table", zap.Int64("table-id", tableID))
```
amyangfei, line 106: Can also log changefeed-id

cdc/processor/processor.go
```diff
@@ -84,6 +92,127 @@ type processor struct {
...
+	log.Info("Remove Table finished",
+		cdcContext.ZapFieldChangefeed(ctx),
+		zap.Int64("tableID", tableID))
+
+	return true
+}
+
+// GetAllCurrentTables implements TableExecutor interface.
+func (p *processor) GetAllCurrentTables() []model.TableID {
+	var ret []model.TableID
```
amyangfei, line 204: Better to use  `ret := make([]model.TableID, 0, len(p.tables))`

amyangfei (approved):

asddongmen (approved):

ben1009 (approved):

PR #3693, opened 2021-12-01
sink(ticdc): add tests for enable-tidb-extension and extract the config out

hi-rustin:

None

* extract the config out
* add tests for enable-tidb-extension opts
* Define some magic string as constants

### Check List 

Tests 

 - Unit test

Code changes

Side effects

None

Related changes

None

### Release note 

```release-note
None
```


hi-rustin:
/cc @amyangfei @3AceShowHand 

cdc/sink/producer/kafka/config.go
```diff
@@ -0,0 +1,297 @@
...
+
+// CompleteByOpts the kafka configuration.
+func (c *Config) CompleteByOpts(sinkURI *url.URL, replicaConfig *config.ReplicaConfig, opts map[string]string) error {
+	c.BrokerEndpoints = strings.Split(sinkURI.Host, ",")
+	params := sinkURI.Query()
+	s := params.Get("partition-num")
+	if s != "" {
+		a, err := strconv.Atoi(s)
+		if err != nil {
+			return err
```
amyangfei, line 77: Better to use an RFC error, ditto for other parameter's Atoi error.
Besides has this error branch be tested?

cdc/sink/producer/kafka/config.go
```diff
@@ -0,0 +1,297 @@
...
+		replicaConfig.Sink.Protocol = s
+	}
+
+	s = params.Get("enable-tidb-extension")
+	if s != "" {
+		_, err := strconv.ParseBool(s)
+		if err != nil {
+			return err
+		}
+		if replicaConfig.Sink.Protocol != "canal-json" {
```
amyangfei, line 171: Should we check the boolean value of `enable-tidb-extension` is true here?

cdc/sink/producer/kafka/config.go
```diff
@@ -0,0 +1,297 @@
...
+	case "lz4":
+		config.Producer.Compression = sarama.CompressionLZ4
+	case "zstd":
+		config.Producer.Compression = sarama.CompressionZSTD
+	default:
+		log.Warn("Unsupported compression algorithm", zap.String("compression", c.Compression))
+		config.Producer.Compression = sarama.CompressionNone
+	}
+
+	// Time out in one minute(120 * 500ms).
```
amyangfei, line 270: Curious about, if each retry reaches the timeout (20s defined in L273), will the total retry time be 120 * 20s?

cdc/sink/producer/kafka/config.go
hi-rustin: It shouldn't be necessary, it should only be possible to configure this value in this protocol.

cdc/sink/producer/kafka/config.go
hi-rustin: It should be a total of 20s.

hi-rustin: I'll reset it.

cdc/sink/producer/kafka/config.go
amyangfei: I mean user configure sink-uri with `enable-tidb-extension=false&protocol=canal` or other protocols except for `canal-json`. Then the check will fail here.

amyangfei: althrough `enable-tidb-extension=false` is useless.

cdc/sink/producer/kafka/config.go
hi-rustin: Wraped in https://github.com/pingcap/ticdc/blob/77029891ff124a5a0e07d5b4dbd4c4f40c6c6268/cdc/sink/mq.go#L384

hi-rustin: > Besides has this error branch be tested?

Let me test it.

amyangfei (approved):

3AceShowHand (approved):

PR #3699, opened 2021-12-02
metrics(ticdc): update panels about tikv

hicqu:

Some changes about metrics panel of TiCDC/TiKV, including
* correct the panel "CDC CPU"
* correct the panel "CDC Network Traffic"
* improve the panel "cdc scan speed"
* improve the panel "old value seek operation"

### Check List 

Tests 

 - No code
 - Manual tests
  Reviewers from PingCAP can take a look at [before](http://172.16.7.42:3000/d/YiGL8hBZ1/three-tikv-cdc-ticdc?orgId=1&from=1638423729823&to=1638425294378) and [after](http://172.16.7.42:3000/d/YiGL8hBZ1dfsfsfsfs/three-tikv-cdc-ticdc-dev?orgId=1&from=1638423720268&to=1638425569022).

Code changes

Side effects

Related changes

### Release note 

```release-note
None
```


0xPoe (commented):
~~Could you please post a screenshot after the change?~~
~~Thanks!  💚 💙 💜 💛 ❤️~~
Found it, ignore me.

metrics/grafana/ticdc.json
```diff
@@ -14,7 +14,7 @@
       "type": "grafana",
       "id": "grafana",
       "name": "Grafana",
-      "version": "6.1.6"
+      "version": "7.5.7"
```
liuzix, line 17: Will this cause compatibility problem on older versions of Grafana?

hicqu: Something changes from `table` to `table-old` in the file, so I guess it is *not* compatible.

PR #3721, opened 2021-12-04
pkg/p2p(cdc): fix minor bugs and add basic integration tests

liuzix:
- We need some test cases for the `MessageServer` and `MessageClient` to work together.
- We need to resolve some minor problems.

- Added `server_client_integration_test.go`.
- Fixed a deadlock bug in `MessageClient`.
- Do not GC `ackManager` for now. Better GC mechanism will be added in a subsequent PR.
- Fixed unstable unit test `TestMessageClientBasics`.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


pkg/p2p/server_client_integration_test.go
```diff
@@ -0,0 +1,380 @@
...
+	cancel = func() {
+		grpcServer.Stop()
+		wg.Wait()
+	}
+	return
+}
+
+func runP2PIntegrationTest(ctx context.Context, t *testing.T, size int, numTopics int) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
amyangfei, line 81: L163 already calls `cancel()`

liuzix: Added the deferred cancel here in case there is a panic mid-way.

amyangfei (approved):

asddongmen (approved):

PR #3723, opened 2021-12-04
chore(dm): grpc.WithBackoffMaxDelay has been deprecated

fatelei:

`// Deprecated: use WithConnectParams instead. Will be supported throughout 1.x.` and the client has using `grpc.WithConnectParams`, the `grpc.WithBackoffMaxDelay` is duplicate.

### Check List 

Tests 

 - No need to add tests

### Release note 

```release-note
None
```


lance6716 (approved):
thanks, can you try that we don't need `//nolint:staticcheck` in line 52 now? (remove it and see if CI passes)

Ehco1996 (approved):
LGTM

PR #3724, opened 2021-12-05
capture(cdc): put OwnerRevision into context

liuzix:
- The new scheduler needs to get the `ownerRev` of the owner, which is a number monotonically increasing cluster-wise, marking the order in which owners are elected.

- Put `OwnerRevision` into `context`.
- Added method `GetOwnerRevision` in our wrapped `CDCEtcdClient`.
- Unit tests.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


amyangfei (approved):

asddongmen (approved):

cdc/capture/capture.go
```diff
@@ -283,10 +283,28 @@ func (c *Capture) campaignOwner(ctx cdcContext.Context) error {
...
+		if err != nil {
+			if errors.Cause(err) == context.Canceled {
+				return nil
+			}
+			return errors.Trace(err)
+		}
+
+		// We do a copy of the globalVars here to avoid
+		// accidental modifications and potential race conditions.
+		globalVars := *ctx.GlobalVars()
```
lonng, line 296: The struct `GlobalVars` share some pointer fields. Do we need to implement the deep copy for the `GlobalVars`?

liuzix: I don't think so. Those are globally shared resources such as the pd client, the tikv client, etc, and they do not need to be copied.

PR #3738, opened 2021-12-06
pkg/p2p(cdc): Add MessageRouter

liuzix:
- `MessageRouter` is designed as a registry for all nodes in the cluster, and it manages the `MessageClient`'s used to send messages to the nodes.

- Implemented `MessageRouter`.

### Check List 

Tests 

 - Unit test

Code changes

### Release note 

```release-note
None
```


3AceShowHand (approved):

pkg/p2p/message_router.go
```diff
@@ -0,0 +1,182 @@
...
+
+// RemovePeer implements MessageRouter.
+func (m *messageRouterImpl) RemovePeer(id NodeID) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	delete(m.addressMap, id)
+	if clientWrapper, ok := m.clients[id]; ok {
+		clientWrapper.cancelFn()
+	}
```
amyangfei, line 95: Does it need to remove client(id) from `m.clients`?

pkg/p2p/message_router.go
```diff
@@ -0,0 +1,182 @@
...
+
+	// repeats the logic in fast path after escalating the lock, since
+	// the lock was briefly released.
+	if cliWrapper, ok := m.clients[target]; ok {
+		return cliWrapper.MessageClient
+	}
+
+	addr, ok := m.addressMap[target]
+	if !ok {
+		log.Info("failed to create client, no peer",
```
amyangfei, line 121: Is `log.Warn` better, If `GetClient` returns nil, will the caller retry it later?

liuzix: Yes, the caller will retry if the target is expected to be added.

pkg/p2p/message_router.go
liuzix: The client is removed from `m.clients` only after it is canceled, to prevent two clients to the same target from co-exist (although this is tolerated to same degree).

amyangfei (approved):

PR #3775, opened 2021-12-07
workerpool: limit the rate to output deadlock warning

overvenus:

Close https://github.com/pingcap/ticdc/issues/2706

Ratelimit deadlock log.

### Check List 

Tests 

Related changes

### Release note 

```release-note
Reduce log "synchronize is taking too long, report a bug" in some cases.
```


amyangfei (commented):
Does this PR need a test?

pkg/workerpool/pool_impl.go
```diff
@@ -15,6 +15,7 @@ package workerpool
 
 import (
 	"context"
+	"golang.org/x/time/rate"
```
liuzix, line 18: Should this be moved down to the second group of imports?

liuzix (approved):

amyangfei (approved):

PR #3792, opened 2021-12-08
owner, scheduler, pkg/p2p(cdc): integrate Scheduler(V2) with the Owner (Part II)

liuzix:
**Notes for Reviewers: This PR does not make significant changes to the logic currently runnable from the binary, so the risk is minimal**.

- We need to integrate the Owner with the new scheduler, and we need a data structure that binds the core scheduling logic (in `cdc/scheduler`) to the messaging logic (in `pkg/p2p`).

- Adjust the `scheduler` interface to accommodate the new scheduler.
- Added `MockCluster` in `pkg/p2p` to do a small scale integration test on the new scheduler.
- Implemented `schedulerV2`, which connects the `cdc/scheduler` package (decoupled from the messaging layer) to `pkg/p2p` package.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/owner/scheduler.go
```diff
@@ -37,4 +49,284 @@ type scheduler interface {
...
+	messageServer *p2p.MessageServer,
+	messageRouter p2p.MessageRouter,
+) (*schedulerV2, error) {
+	ret := &schedulerV2{
+		changeFeedID:  changeFeedID,
+		messageServer: messageServer,
+		messageRouter: messageRouter,
+		stats:         &schedulerStats{},
+	}
+	ret.BaseScheduleDispatcher = schedulerv2.NewBaseScheduleDispatcher(changeFeedID, ret, checkpointTs)
```
asddongmen, line 83: I was confused by the packge name...

asddongmen (approved):

nongfushanquan (approved):

amyangfei (approved):

PR #3800, opened 2021-12-09
test(dm): fix testServer.TestServer data race

Ehco1996:
close https://github.com/pingcap/ticdc/issues/3791

`syncMasterEndpoints` use `SetEndpoints` to update master endpoints and this cause data-race with `ha.keepalive` which call `clientv3.(*Client).Endpoints()`

datarace line `clientURLs = append(clientURLs, m.GetClientURLs()...)` is not protected by client's mutex

so use `clientv3.(*Client).autoSync()` to update endpoints

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`.
```


okJiang:
I have a question.

Why the `SerialSuites` can solve the data race? Which `Suite` trigger the data race with `TestServer`?

Ehco1996:
> I have a question.
> 
> Why the `SerialSuites` can solve the data race? Which `Suite` trigger the data race with `TestServer`?

```
Goroutine 238 (running) created at:
  github.com/pingcap/ticdc/dm/dm/worker.(*Server).startKeepAlive()
      /home/jenkins/agent/workspace/atom-ut/ticdc/dm/dm/worker/server.go:249 +0x12c
  github.com/pingcap/ticdc/dm/dm/worker.(*Server).Start()
      /home/jenkins/agent/workspace/atom-ut/ticdc/dm/dm/worker/server.go:133 +0x875
  github.com/pingcap/ticdc/dm/dm/worker.(*testServer).TestServer.func4()
      /home/jenkins/agent/workspace/atom-ut/ticdc/dm/dm/worker/server_test.go:139 +0x3c
Goroutine 177 (running) created at:
  github.com/pingcap/ticdc/dm/dm/worker.(*Server).Start()
      /home/jenkins/agent/workspace/atom-ut/ticdc/dm/dm/worker/server.go:128 +0x849
  github.com/pingcap/ticdc/dm/dm/worker.(*testServer).TestServer.func4()
      /home/jenkins/agent/workspace/atom-ut/ticdc/dm/dm/worker/server_test.go:139 +0x3c
```
see more in https://github.com/pingcap/ticdc/issues/3791

if i am right, two test thread call `server.Start` and this cause etcd-server data-race

okJiang:
I think so. The `s.syncMasterEndpoints(s.ctx)` is an another goroutine which can lead to data race with  `s.startKeepAlive()` .

https://github.com/pingcap/ticdc/blob/5b1bbe86fc7a6b25f15137f55273d8a8dce6cd59/dm/dm/worker/server.go#L128-L133

But `SerialSuites` only changes the `Suites` execution to serial. Seems that doesn't fix the data race.

Ehco1996:
/hold 

wait further investigation

dm/dm/worker/server.go
```diff
@@ -111,6 +111,7 @@ func (s *Server) Start() error {
 		DialKeepAliveTime:    keepaliveTime,
 		DialKeepAliveTimeout: keepaliveTimeout,
 		TLS:                  tls.TLSConfig(),
+		AutoSyncInterval:     syncMasterEndpointsTime,
```
Ehco1996, line 114: this can start a thread to sync endpoint in etcdclient see more here https://pkg.go.dev/go.etcd.io/etcd/clientv3#Config

Ehco1996:
/unhold

@okJiang  i think i found the root cause (updated to pr description) ptal

okJiang:
> /unhold
> 
> @okJiang i think i found the root cause (updated to pr description) ptal

Your work does streamline the code. Great job!

But it doesn't resolved the data race. Because `etcd.Sync()` is same as original, you can see https://github.com/etcd-io/etcd/blob/29292aa7bdafaf65cb5e054591fe0ff07b36f5ee/client/v3/client.go#L181-L193

I search the etcd issue, and find this is a bug that fixed in https://github.com/etcd-io/etcd/issues/13201. So we just fix it by upgrading etcd. 🐶

okJiang (approved):

lance6716 (approved):

PR #3807, opened 2021-12-09
processor, scheduler(cdc): integration scheduler/agent with the Processor

liuzix:
- We need to integrate the scheduler agent with the Processor.

- Implemented `processorAgent`.
- Added integration tests (in-process) of `processorAgent` and `pkg/p2p`.
- Adjusted relevant logic in the rest of the Processor.
- Minor adjustments to the unit test cases of the Processor.
- Minor adjustments to `MessageServer` to make deregistering message handlers idempotent.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/processor/agent.go
```diff
@@ -0,0 +1,369 @@
...
+		// We should wait for the first owner to contact us.
+		// We need to wait for the sync request anyways, and
+		// there would not be any table to replicate for now.
+		log.Debug("waiting for owner to request sync",
+			zap.String("changefeed-id", a.changeFeed))
+		return false
+	}
+
+	client := a.messageRouter.GetClient(a.ownerCaptureID)
+	if client == nil {
```
amyangfei, line 240: Does it need to add log here

cdc/processor/agent.go
```diff
@@ -0,0 +1,369 @@
...
+		actualSeq, ok := client.CurrentAck(topic)
+		if !ok {
+			return false
+		}
+		if actualSeq >= waitSeq {
+			delete(a.barrierSeqs, topic)
+		} else {
+			return false
+		}
+	}
```
amyangfei, line 256: Could it happen this scenario, given two topics, t-a, t-b, and sequence enqueue as follows
```
t-a wait 10, t-b wait 11, t-a wait 12, t-b wait 13...
             t-a ack 10 , t-b-ack 11, t-a ack 12  ...            
```
which leads to `Barrier` always return false.

liuzix: This scenario will not happen, because `Barrier` returning false will prevent sending the checkpoint, which will prevent the Owner from sending any further commands, so the processor will not reply anything.

I will think more about this scenario and try to refactor so that this kind of problem will not be introduced in the future.

asddongmen (approved):

amyangfei (approved):

PR #3811, opened 2021-12-09
sink(ticdc): add ProtocolOpen to replace default protocol

hi-rustin:
Add open-protocol protocol.

part of https://github.com/pingcap/ticdc/issues/3879

* Add the open-protocol protocol and use it as an alias for default protocol.
* No more accepting protocols that are not known

### Check List 

Tests 

 - Unit test
 - Integration test (TBD)

Code changes

Side effects

None

Related changes

None

### Release note 

```release-note
None
```


hi-rustin:
/cc @3AceShowHand @asddongmen 

overvenus (approved):
LGTM

pkg/config/mq_sink_protocol.go
```diff
@@ -31,13 +30,14 @@ const (
...
+	ProtocolOpen
 )
 
 // FromString converts the protocol from string to Protocol enum type.
-func (p *Protocol) FromString(protocol string) {
+func (p *Protocol) FromString(protocol string) error {
 	switch strings.ToLower(protocol) {
 	case "default":
-		*p = ProtocolDefault
+		*p = ProtocolOpen
```
lance6716, line 40: now the input and output by `String()` is different, is this expected? For example, backup the output and apply it to a cluster of lower version

lance6716 (commented):
rest lgtm

pkg/config/mq_sink_protocol.go
hi-rustin: i will fix it in next pr. I will refine the meta dta from  low version cluster.

lance6716 (approved):

PR #3814, opened 2021-12-09
parser(dm): use TiDB special comment for DROP INDEX IF EXISTS

lance6716:
close https://github.com/pingcap/ticdc/issues/3004

as title

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
Better support MySQL -> TiDB -> MySQL scenario by using TiDB special comment
```


Ehco1996 (approved):
LGTM 

some information for other reviewers

* mysql [5.6](https://dev.mysql.com/doc/refman/5.6/en/drop-index.html)/[8.0](https://dev.mysql.com/doc/refman/8.0/en/drop-index.html) does't not support `DROP INDEX IF EXISTS `
* mariadb support this statement but dm not commit to support mariadb as downstream
* tidb support this statement with or without comment

lichunzhu (approved):
LGTM

Issue #3808, opened 2021-12-09
remove unnecessary logs from prometheus metrics request.


3AceShowHand:
### What did you do?

do nothing

### What did you expect to see?

no logs from Prometheus metrics request.

### What did you see instead?

```
[2021/12/06 20:23:39.053 +08:00] [INFO] [http_router.go:144] [/metrics] [status=200] [method=GET] [path=/metrics] [query=] [ip=172.16.6.142] [user-agent=Prometheus/2.27.1] [] [cost=3.763361ms]
[2021/12/06 20:23:54.053 +08:00] [INFO] [http_router.go:144] [/metrics] [status=200] [method=GET] [path=/metrics] [query=] [ip=172.16.6.142] [user-agent=Prometheus/2.27.1] [] [cost=3.207389ms]
[2021/12/06 20:24:09.053 +08:00] [INFO] [http_router.go:144] [/metrics] [status=200] [method=GET] [path=/metrics] [query=] [ip=172.16.6.142] [user-agent=Prometheus/2.27.1] [] [cost=3.145238ms]
[2021/12/06 20:24:24.053 +08:00] [INFO] [http_router.go:144] [/metrics] [status=200] [method=GET] [path=/metrics] [query=] [ip=172.16.6.142] [user-agent=Prometheus/2.27.1] [] [cost=3.026089ms]
[2021/12/06 20:24:39.053 +08:00] [INFO] [http_router.go:144] [/metrics] [status=200] [method=GET] [path=/metrics] [query=] [ip=172.16.6.142] [user-agent=Prometheus/2.27.1] [] [cost=3.480016ms]
[2021/12/06 20:24:54.053 +08:00] [INFO] [http_router.go:144] [/metrics] [status=200] [method=GET] [path=/metrics] [query=] [ip=172.16.6.142] [user-agent=Prometheus/2.27.1] [] [cost=3.304862ms]
[2021/12/06 20:25:09.053 +08:00] [INFO] [http_router.go:144] [/metrics] [status=200] [method=GET] [path=/metrics] [query=] [ip=172.16.6.142] [user-agent=Prometheus/2.27.1] [] [cost=3.227668ms]
[2021/12/06 20:25:24.052 +08:00] [INFO] [http_router.go:144] [/metrics] [status=200] [method=GET] [path=/metrics] [query=] [ip=172.16.6.142] [user-agent=Prometheus/2.27.1] [] [cost=3.060645ms]
[...]

PR #3824, opened 2021-12-10
http_api (ticdc): filter out metrics related log

asddongmen:
#3808 

### Check List 

Tests 

Code changes

Side effects

Related changes

### Release note 

```release-note
None
```


hi-rustin (approved):
It looks fine, but maybe wrapping it as skipPathRegexp(path regex) might work better. But for now it doesn't matter if it's hardcoded because there's only one.

ben1009 (approved):

cdc/http_router.go
```diff
@@ -140,7 +141,10 @@ func logMiddleware() gin.HandlerFunc {
 		if err != nil {
 			stdErr = err.Err
 		}
-
+		// Do not logs metrics related requests when there is no error
```
amyangfei, line 144:
```suggestion
		// Do not log metrics related requests when there is no error
```

Issue #3473, opened 2021-11-16
Upgrade cluster from 4.0.14 to 5.3.0，the stopped changefeed automatically becomes to run


Tammyxia:
### What did you do?

- deploy tidb 4.0.14
- create changefeed: tiup ctl:v4.0.14 cdc changefeed create --pd=http://172.16.4.158:2379 --sink-uri="tidb://root:@172.16.6.47:4000/"  --changefeed-id="mysql-task-1"
- pause this changefeed, make sure it has become stopped status.
- upgrade tidb to 5.3.0
- check changefeed status

### What did you expect to see?

The changefeed is still stopped status.

### What did you see instead?

The changefeed become normal status.

### Versions of the cluster

Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

```console
v4.0.14 -> v5.3.0
```

TiCDC version (execute `cdc version`):

```console
v4.0.14 -> v5.3.0
```

hi-rustin:
I found in my local testing that it's just a task status display issue. It won't really sync data.

PR #3839, opened 2021-12-11
owner(ticdc): correctly handle the state after the old owner is upgraded to the new owner

hi-rustin:
…

part of #3473

correctly handle the state after the old owner is upgraded to the new owner.
Fix it in changefeed manager tick.

### Check List 

Tests 

 - Unit test
 - Manual test (TBD)

Code changes

Side effects

None

Related changes

None

### Release note 

```release-note
Please add a release note.
Fix the problem of changefeed resuming automatically after upgrading cluster
```


hi-rustin:
See: https://github.com/pingcap/ticdc/pull/3861

PR #3851, opened 2021-12-13
etcd/client (ticdc): Prevent revision in WatchWitchChan fallback.

asddongmen:

Prevent revision rollback.

**Note**: Before this PR merge, the lastRevision in WatchWitchChan is initialized to 0. When the watchCh is reset since there has no message been received , revision will be set to 0+1.
This may result in receiving some messages that do not need to be received, lead to a certain extend of performance loss, **but it will not cause any correctness problems.**

Clear and simple in code.

### Check List 

Tests 

 - Unit test

Related changes

### Release note 

```release-note
None
```


pkg/etcd/client.go
```diff
@@ -235,7 +242,7 @@ func (c *Client) WatchWithChan(ctx context.Context, outCh chan<- clientv3.WatchR
 				log.Warn("etcd client watchCh blocking too long, reset the watchCh", zap.Duration("duration", c.clock.Since(lastReceivedResponseTime)), zap.Stack("stack"))
 				cancel()
 				watchCtx, cancel = context.WithCancel(ctx)
-				watchCh = c.cli.Watch(watchCtx, key, clientv3.WithPrefix(), clientv3.WithRev(lastRevision+1))
+				watchCh = c.cli.Watch(watchCtx, key, clientv3.WithPrefix(), clientv3.WithRev(lastRevision))
```
amyangfei, line 245: What's the difference between using `lastRevision+1` and `lastRevision`

pkg/etcd/client.go
```diff
@@ -188,7 +188,14 @@ func (c *Client) WatchWithChan(ctx context.Context, outCh chan<- clientv3.WatchR
 		close(outCh)
 		log.Info("WatchWithChan exited")
 	}()
-	var lastRevision int64
+
+	// get initial revision from opts to avoid revision fall back
+	op := &clientv3.Op{}
```
liuzix, line 193: Can you briefly explain the algorithm? It would probably be even better if you could extract this piece of code into a unit-testable function.

liuzix: Since this piece of code relies on some implementation details of an external package, we would unit-test it.

pkg/etcd/client.go
asddongmen: Before this PR, we could set it `lastRevision+1` is because when:
-  `watchCh` is reset since there has no any msg been received before, revision will be set to 0+1. This may result in receiving some messages that do not need to be received, but no any msg will be lost.
- `watchCh` is reset after `lastReceivedResponseTime` pass since last msg was received, revision will be set to `lastRevision+1` (lastRevision = lastMsg.revision). No any msg will be lost, because the last msg has been received.

After this PR, if we set it `lastRevision + 1` when `watchCh` is reset since there has no any msg been received before. This may result in one msg that `revision = lastRevision` lost, because in the beginning, we set `lastRevision = revision (which we need to watch)`.

In general, we can tolerate receiving redundant msg because the user of the msg will handle this situation, but we cannot lose any msg.

pkg/etcd/client.go
asddongmen: ok, thanks for your advice~

asddongmen: done PTAL~

liuzix (approved):

amyangfei (approved):

PR #3860, opened 2021-12-13
syncer(dm): use an early location to reset binlog and open safemode

lance6716:
close https://github.com/pingcap/ticdc/issues/3487 for release-5.3
https://github.com/pingcap/ticdc/issues/3711 of master branch need a better solution

we use globalCheckpoint which is a correct-but-early location to reset binlog streamer, and open one transaction safemode

add a half of an integration test back

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### Release note 

```release-note
Fix a bug of data loss when DM does finer grained retry
```


dm/syncer/syncer.go
```diff
@@ -1621,11 +1632,13 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 
 			if s.streamerController.CanRetry(err) {
 				// lastLocation is the last finished GTID
-				err = s.streamerController.ResetReplicationSyncer(tctx, lastLocation)
+				err = s.streamerController.ResetReplicationSyncer(tctx, s.checkpoint.GlobalPoint())
 				if err != nil {
 					return err
 				}
 				log.L().Info("reset replication binlog puller", zap.Any("pos", s.checkpoint.GlobalPoint()))
+				_ = s.safeMode.Add(tctx, 1)
```
glorv, line 1640: How about use a local variable like `inFinerRetry` to record this temporary safe-mode
```
safeMode: s.safeMode.Enable() | retrySafeMode,
```

dm/syncer/syncer.go
```diff
@@ -1621,11 +1632,13 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 
 			if s.streamerController.CanRetry(err) {
 				// lastLocation is the last finished GTID
-				err = s.streamerController.ResetReplicationSyncer(tctx, lastLocation)
+				err = s.streamerController.ResetReplicationSyncer(tctx, s.checkpoint.GlobalPoint())
```
glorv, line 1635: If this retry is occurred at `sharding-resyc` phase, enable safe-mode for one transaction is not enough because the start point may be reset to the very beginning point. But since safe-mode is enable in the full sharding-resync phase this is ok. Maybe we can add a comment about this here.

lichunzhu: What does `safe-mode is enable in the full sharding-resync phase` means? To me this is a problem.

lichunzhu (commented):
Rest LGTM

dm/syncer/syncer.go
lance6716: we will open safemode when there's a unsynced sharding table

https://github.com/pingcap/ticdc/blob/d381504a9fc3e5e6615e17627c424ec5bd17b0b4/dm/syncer/syncer.go#L2680

https://github.com/pingcap/ticdc/blob/d381504a9fc3e5e6615e17627c424ec5bd17b0b4/dm/syncer/syncer.go#L2708

please help check it

lance6716:
ping @lichunzhu @glorv

glorv (approved):
LGTM

dm/syncer/syncer.go
lichunzhu: Looks good to me

lichunzhu (approved):

PR #3891, opened 2021-12-15
syncer(dm): don't check flush inside an event and refactor

lance6716:
part of refactor

- `addJob` now becomes a simple function of add job to channel, rest logic goes to `handleJob`
- as title
  - handleRowsEvent will save the table point and check flush after all jobs are added
  - remove a test case because we no longer flush inside an event

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`.
```


lichunzhu (commented):
Don't understand what does `don't check flush inside an event` means.

dm/syncer/syncer.go
```diff
@@ -2252,12 +2275,34 @@ func (s *Syncer) handleRowsEvent(ev *replication.RowsEvent, ec eventContext) err
...
+
+			err2 := s.flushJobs()
+			if err2 != nil {
+				s.tctx.L().DPanic("flush checkpoint failed", zap.Error(err2))
+			}
+			failpoint.Return(nil)
+		}
+	})
+
+	s.checkShouldFlush()
```
lichunzhu, line 2305: It seems that we run `s.checkShouldFlush()` twice for dmls?

dm/syncer/job.go
```diff
@@ -73,6 +73,9 @@ func (t opType) String() string {
 	return ""
 }
 
+// job is the least unit of work for DMLWorker/DDL worker. It's generated by syncer's inner logic or binlog events. Specifically, most
```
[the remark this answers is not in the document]
lance6716: I want to refer to https://github.com/pingcap/tiflow/blob/d84f15b221bf07b29c9949344f38b3248d7f17f6/dm/syncer/dml_worker.go#L33 and we don't have a DDL worker now

dm/syncer/syncer.go
lance6716: fixed in https://github.com/pingcap/tiflow/pull/3891/commits/e549451fae295da5a31ddcbd923d27bd74c19933

lichunzhu (commented):
Should the title be changed from
~~syncer(dm): don't check flush inside an event and refactor~~ to
syncer(dm): don't check flush inside a transaction and refactor

Besides, can you explain what can we benefit from check flush after a transaction and whether it will affect a huge transaction's retry time?

lance6716:
> 

a transaction may has multiple events, and an event may has multiple row changes thus multiple jobs. Before this PR we check if global checkpoint should be flushed in `addJob`, so we may flush in the middle of sending an event.

For old behaviour because the event is not fully sent, we still can't skip it. So we still need replicate this event when retry. Now in position-based replication we have confidence to skip it.

lichunzhu (approved):
LGTM

glorv (approved):

Ehco1996:
need fix lint `github.com/pingcap/tiflow/pkg/errorutil` redeclared in syncer_test.go

PR #3896, opened 2021-12-15
owner, scheduler(cdc): expose information from scheduler

liuzix:
- The PR is part of the solution to the API compatibility problem.
- In the legacy scheduling model, the `cdc cli` as well as the OpenAPI implementation directly reads from Etcd data structures such as `TaskPosition` and `TaskStatus`. But in the new scheduling model, these metadata structures no longer exist.

- Defined `InfoProvider` interface to provide a way to inspect the internal states of the scheduler.
- Implemented `InfoProvider` for the new scheduler.
- Added appropriate switching logic to use `InfoProvider` to supply relevant data when the new scheduler is enabled.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

### Release note 

```release-note
None
```


cdc/scheduler/info_provider.go
```diff
@@ -0,0 +1,83 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package scheduler
```
nongfushanquan, line 14: why not add the file to the schedule_dispatcher.go file

liuzix: I was thinking of improving readability by separating non-core logic into a separate file. Moreover, we might want to refactor the relevant logic into a separate struct implementing `InfoProvider`.

amyangfei (approved):

asddongmen (approved):

Issue #2162, opened 2021-06-25
Make more linters really happy 


hi-rustin:
### Description

Currently, we use "golangci-lint" to execute some linters, but due to some reasons, the entire project code may not really pass some necessary linters.
This issue is for everyone to work together to pass some necessary linters for this project. The specific parameters that need to be supported are shown in the task list below.

### Related Information

* [golangci-lint](https://golangci-lint.run/)

### Task List
- [x] --enable=unconvert(https://github.com/pingcap/ticdc/pull/2172)
- [ ] --enable=[lll](https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/lll.go)
- [x] --enable=[unparam](https://github.com/mvdan/unparam)(https://github.com/pingcap/ticdc/pull/2368)
- [ ] --enable=[gosec](https://github.com/securego/gosec)
- [ ] --enable=[funlen](https://github.com/ultraware/funlen)
- [ ] --enable=[gocyclo](https://github.com/alecthomas/gocyclo)

If there are more necessary linter you would like to enable, please feel free to comment.

hi-rustin:
cc: @amyangfei  @leoppro  @overvenus  

While reading the code, I noticed that we have some unnecessary type conversion code, so I would like to turn on the `unconvert` plugin to improve the code quality. What do you think?

hi-rustin:
/assign 

I am working on this.

hi-rustin:
I found in the process of reading the code that a lot of our code is too long for one screen to read easily. So I suggest we use lint to control the length of each line of code.

hi-rustin:
There are many parameters in our code that are not used. We can use [unparam](https://github.com/mvdan/unparam) to detect them!

Original idea from https://github.com/pingcap/ticdc/pull/2309#issuecomment-881796768. cc: @amyangfei

PR #3898, opened 2021-12-15
misc(ticdc): add more linters and fix issues found by misspell

dveeden:

- Fix misspelled words
- Add more linters to avoid future problems (by copying from the pingcap/tidb `.golangci.yml`)

Differences with pingcap/tidb
- G404 disabled for `gosec` (`G404: Use of weak random number generator (math/rand instead of crypto/rand)`, should be fine as the uses are in non-crypto contexts, like retries)
- `bodyclose` removed
- `rowserrcheck` removed
- `unconvert` removed

Would be great to enable the removed checks once the issues found by them are fixed. Annotations for G404 could be used (`// #nosec G404`) if we want to enable this check.

### Check List 

Tests 

 - No code

### Release note 

```release-note
None
```


.golangci.yml
```diff
@@ -3,6 +3,22 @@ linters:
     - unconvert
     - unparam
     - revive
+    - misspell
```
Ehco1996, line 6: could you please also enable these checker for dm? https://github.com/pingcap/ticdc/blob/master/dm/.golangci.yml

overvenus (approved):
LGTM, thanks!

.golangci.yml
dveeden: It's already included: https://github.com/pingcap/ticdc/blob/71e2a4818e91b8b7fc4ffe90a936f0772c36f2b9/dm/.golangci.yml#L18

hi-rustin:
cc: https://github.com/pingcap/tiflow/issues/2162

.golangci.yml
```diff
@@ -4,6 +4,22 @@ linters:
     - unparam
     - revive
     - depguard
+    - misspell
+    - ineffassign
+    - typecheck
+    - varcheck
+    - unused
```
hi-rustin, line 11: It seems we have listed some lint that is open by default? But it shouldn't matter.

hi-rustin (approved):

.golangci.yml
dveeden: yes indeed

hi-rustin:
Thanks!

Issue #3823, opened 2021-12-10
fatal error: concurrent map read and map write in Optimistic Sharding DDL


GMHDBJD:
### What did you do?
```
fatal error: concurrent map read and map write

goroutine 37885 [running]:
runtime.throw(0x24027ce, 0x21)
	/usr/local/go/src/runtime/panic.go:1117 +0x72 fp=0xc08f804f40 sp=0xc08f804f10 pc=0x438592
runtime.mapaccess2(0x20897e0, 0xc09ce4e2d0, 0xc04d162d40, 0xc04d162d40, 0xc04100c97f)
	/usr/local/go/src/runtime/map.go:469 +0x255 fp=0xc08f804f80 sp=0xc08f804f40 pc=0x40f915
reflect.mapaccess(0x20897e0, 0xc09ce4e2d0, 0xc04d162d40, 0x23e208f)
	/usr/local/go/src/runtime/map.go:1318 +0x3f fp=0xc08f804fb8 sp=0xc08f804f80 pc=0x46bb3f
reflect.Value.MapIndex(0x20897e0, 0xc09ce4e2d0, 0x15, 0x1ffd7a0, 0xc04d162d40, 0x98, 0x20898a0, 0xc09ce4f710, 0x15)
	/usr/local/go/src/reflect/value.go:1189 +0x16e fp=0xc08f805030 sp=0xc08f804fb8 pc=0x4c8d0e
encoding/json.mapEncoder.encode(0xc0722c59c0, 0xc06341c100, 0x20897e0, 0xc09ce4e2d0, 0x15, 0x100)
	/usr/local/go/src/encoding/json/encode.go:813 +0x317 fp=0xc08f8051d8 sp=0xc08f805030 pc=0x52f237
encoding/json.mapEncoder.encode-fm(0xc06341c100, 0x20897e0, 0xc09ce4e2d0, 0x15, 0xc04d160100)
	/usr/local/go/src/encoding/json/encode.go:779 +0x6d fp=0xc08f805218 sp=0xc08f8051d8 pc=0x53b9ed
encoding/json.mapEncoder.encode(0xc0722c59e0, 0xc06341c100, 0x2089780, 0xc09764a560, 0x95, 0x100)
	/usr/local/go/src/encoding/json/encode.go:813 +0x366 fp=0xc08f8053c0 sp=0xc08f805218 pc=0x52f286
[...]

PR #3903, opened 2021-12-16
Optimistic: support start task with inconsistent upstream table schema

### LINKED ISSUE #3629: optimistic: sharding DDL incorrect after stop a source
### What did you do?

start an optimistic task with 3 sources
```
stop-task <task> -s source1
source2 drop column1
source3 drop column1 
```

downstream drop column1

### What did you see instead?

downstream doesn't drop column 1

### Versions of the cluster

```console
master
```

```console
```

### LINKED ISSUE #3708: Optimistic sharding may not get initial table structure
found in CI https://github.com/pingcap/dm/pull/1959/checks?check_run_id=3244290635

in the test case, two worker did INSERT ... VALUES() and ALTER TABLE ... ADD COLUMN in optimistic sharding. one worker has done ADD COLUMN, then another worker starts to get table structure

DM-master's log:

```
[2021/08/04 17:46:31.830 +00:00] [INFO] [optimist.go:660] ["recorded the initial schema"] [component="shard DDL optimist"] [info="{\"task\":\"upgrade_via_tiup_optimistic\",\"source\":\"mysql-replica-01\",\"up-schema\":\"opt_sharding1\",\"up-table\":\"t1\",\"down-schema\":\"opt_db_target\",\"down-table\":\"t_target\",\"ddls\":[\"ALTER TABLE `opt_db_target`.`t_target` ADD COLUMN `c3` INT\"],\"table-before\":\"CREATE TABLE `tbl`(`c1` INT(11) NOT NULL, `c2` TEXT(65535) CHARACTER SET UTF8MB4 COLLATE utf8mb4_bin, PRIMARY KEY (`c1`)) CHARSET UTF8MB4 COLLATE UTF8MB4_BIN\",\"table-after\":\"CREATE TABLE `tbl`(`c1` INT(11) NOT NULL, `c2` TEXT(65535) CHARACTER SET UTF8MB4 COLLATE utf8mb4_bin, `c3` INT(11), PRIMARY KEY (`c1`)) CHARSET UTF8MB4 COLLATE UTF8MB4_BIN\",\"is-deleted\":false,\"version\":1,\"revision\":30,\"ignore-conflict\":false}"]
```

DM-worker's log:

```
[2021/08/04 17:46:31.694 +00:00] [INFO] [optimist.go:175] ["no init schema exists"] [task=upgrade_via_tiup_optimistic] [unit="binlog replication"] [component="shard DDL optimist"] [schema=opt_db_target] [table=t_target] [revision=29]
[2021/08/04 17:46:31.828 +00:00] [INFO] [optimist.go:175] ["no init schema exists"] [task=upgrade_via_tiup_optimistic] [unit="binlog replication"] [component="shard DDL optimist"] [schema=opt_db_target] [table=t_target] [revision=30]
... 
[2021/08/04 17:46:31.849 +00:00] [WARN] [db.go:169] ["execute statements"] [task=upgrade_via_tiup_optimistic] [unit="binlog replication"] [retry=0] [queries="[REPLACE INTO `opt_db_target`.`t_target` (`c1`,`c2`) VALUES (?,?)]"] [arguments="[[111 [49 49 49]]]"] [error="[code=10006:class=database:scope=not-set:level=high], Message: execute statement failed: commit, RawCause: Error 8028: Information schema is changed during the execution of the statement(for example, table definition may be updated by other DDL ran in parallel). If you see this error often, try increasing `tidb_max_delta_schema_count`. [try again later]"]
...
... gen insert sqls failed, schema: opt_db_target, table: t_target: Column count doesn't match value count: 3 (columns) vs 2 (values)
```

the latter worker read table structure from downstream, so it fails with "Column count doesn't match value count" error because downstream has finished ADD COLUMN

### LINKED ISSUE #3786: Support start task with inconsistent upstream table schema for Optimsitic Sharding DDL

When users have some inconsistent sharding tables, they can not start a task either in pessimistic mode or optimistic mode.

start task with inconsistent upstream table schema for Optimsitic Sharding DDL
e.g.

table1(a int primary, b int)
table2(a int primary, c int)

_No response_

To support different upstream table schema when tasks start, master need to know all upstream schema for a optimstic DDL.
The main logic may like
```
start-task
worker put all upstream table names which needed for the task to etcd
master watch table names and save to memory
worker fetch table infos from dump files or upstream databases and then flush to downstream checkpoint when entering sync unit
master receive a DDL and then fetch all sharding table infos from downstream for that DDL
master joined table base on all table infos
master return ddl to worker
```
When stop-task for a source, all the table infos/ddl infos/operations for that source in memory or etcd will be removed.

GMHDBJD:
Issue Number: close #3629 
close #3786 
close #3708 
- worker flush table info when entering sync stage
- master get table info when receiving ddl info
- remove info/operation/locks/drop columns when stop a optimistic task

For DM-master:
- fetch table info for every table when creating a lock rather than use `joined` as init table
- when master restart, no need to rebuild locks base on init schema, just handle info by order
- when the task stops, remove all metadata for optimism, when restarting the task, we can fetch table info from downstream instead.

For DM-worker:
- get and flush table info when entering sync stage

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
support start task with inconsistent upstream table schemas in optimistic shard mode.
```


Makefile
```diff
@@ -275,13 +275,22 @@ dm_generate_openapi: tools/bin/oapi-codegen
...
-	$(GOTEST) -timeout 5m -covermode=atomic -coverprofile="$(DM_TEST_DIR)/cov.unit_test.out" $(DM_PACKAGES) \
+	$(GOTEST) -timeout 5m -covermode=atomic -coverprofile="$(DM_TEST_DIR)/cov.unit_test.out" $(1) \
 	|| { $(FAILPOINT_DISABLE); exit 1; }
 	$(FAILPOINT_DISABLE)
+endef
+
+dm_unit_test: check_failpoint_ctl
+	$(call run_dm_unit_test,$(DM_PACKAGES))
+
+# run unit test for the specified pkg only, like `make dm_unit_test_pkg PKG=github.com/pingcap/dm/dm/master`
```
lance6716, line 291:
```suggestion
# run unit test for the specified pkg only, like `make dm_unit_test_pkg PKG=github.com/pingcap/tiflow/dm/dm/master`
```

lance6716 (commented):
big PR :joy:

dm/syncer/checkpoint.go
```diff
@@ -698,45 +699,68 @@ func (cp *RemoteCheckPoint) FlushPointsExcept(
...
-	tiBytes, err := json.Marshal(ti)
-	if err != nil {
-		return terror.ErrSchemaTrackerCannotSerialize.Delegate(err, sourceSchema, sourceTable)
-	}
-
-	location := point.MySQLLocation()
-	sql2, arg := cp.genUpdateSQL(sourceSchema, sourceTable, location, nil, tiBytes, false)
-	sqls = append(sqls, sql2)
-	args = append(args, arg)
+	batch := 100
```
lance6716, line 711: we can use the batch of DML worker :thinking: 100 is also OK

dm/syncer/checkpoint.go
```diff
@@ -698,45 +699,68 @@ func (cp *RemoteCheckPoint) FlushPointsExcept(
...
+				if p, ok2 := tablePoints[sourceTable]; ok2 {
+					point = p
+				}
+			}
+			// create new point
+			if point == nil {
+				cp.saveTablePoint(table, cp.globalPoint.MySQLLocation(), nil)
+				point = cp.points[sourceSchema][sourceTable]
+			}
+			tiBytes, err := json.Marshal(ti)
```
lance6716, line 738: do we plan to save CREATE TABLE instead of json dump in this sprint?

And a test shows when there're 10000 tables each with 400 columns, the json dump will costs 2.3 GB, so for batch of 100 we'll send 23 MB data per transaction in this scenario :thinking:

lance6716 (commented):
can you describe more details about optimist change in PR description?

dm/syncer/checkpoint.go
GMHDBJD: Since this case needs to change the checkpoint table and is not a feature for optimism, I think we should discuss whether/when to support it, and also need to more careful design.

dm/dm/master/shardddl/optimist.go
```diff
@@ -185,10 +186,45 @@ func (o *Optimist) RemoveMetaData(task string) error {
...
+	// gets all locks for this task
+	locks := o.lk.FindLocksByTask(task)
+	for _, lock := range locks {
+		// remove table by sources for related lock
+		if lock != nil {
+			cols := lock.TryRemoveTableBySources(sources)
+			dropColumns[lock.ID] = cols
+			o.logger.Debug("the tables removed from the lock", zap.String("task", task), zap.Strings("sources", sources))
+		}
+		if !lock.HasTables() {
```
Ehco1996, line 217: do we need put `!lock.HasTables()` in to `if lock != nil`

dm/dm/master/server.go
```diff
@@ -572,6 +572,18 @@ func (s *Server) OperateTask(ctx context.Context, req *pb.OperateTaskRequest) (*
 
 	resp.Result = true
 	resp.Sources = s.getSourceRespsAfterOperation(ctx, req.Name, sources, []string{}, req)
+
+	if expect == pb.Stage_Stopped {
+		// delete meta data for optimist
```
Ehco1996, line 577: can you explian why `when the task stops, remove all metadata for optimism, when restarting the task, we can fetch table info from downstream instead.` what will happen when we not delete  meta here?

dm/pkg/shardddl/optimism/keeper.go
```diff
@@ -20,79 +20,67 @@ import (
...
-					}
-					lk.locks[lockID].tables[source][schema][table] = tableinfo
-				}
-			}
-		}
+	dbConfig, meta := lk.getDownstreamMetaFunc(task)
+	if dbConfig == nil {
+		return nil, terror.ErrMasterOptimisticDownstreamMetaNotFound.Generate(task)
 	}
+	db, err := conn.DefaultDBProvider.Apply(dbConfig)
```
Ehco1996, line 72: defer db.Close()

dm/syncer/syncer.go
```diff
@@ -1483,15 +1475,36 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
+		if s.cfg.Mode == config.ModeAll {
+			delLoadTask = true
+			flushCheckpoint = true
+			err = s.loadTableStructureFromDump(ctx)
+			if err != nil {
+				tctx.L().Warn("error happened when load table structure from dump files", zap.Error(err))
+				cleanDumpFile = false
+			}
+		}
+		tbls := s.optimist.Tables()
```
Ehco1996, line 1488: why always fetch tbls regardless of whether user open the optimist?

dm/tests/shardddl_optimistic/conf/dm-master.toml
```diff
@@ -0,0 +1,10 @@
+# Master Configuration.
+name = "master1"
+master-addr = ":8261"
+advertise-addr = "127.0.0.1:8261"
+peer-urls = "127.0.0.1:8291"
+initial-cluster = "master1=http://127.0.0.1:8291"
+auto-compaction-retention = "3s"
+
+[experimental]
```
Ehco1996, line 9: now we can remove the `experimental` label

dm/dm/master/server.go
GMHDBJD: etcd meta will become larger and larger.

dm/dm/master/shardddl/optimist.go
GMHDBJD: not null now, 
8e2520e

dm/tests/shardddl_optimistic/conf/dm-master.toml
GMHDBJD: keep as compatibility test

dm/syncer/syncer.go
GMHDBJD: update

dm/pkg/shardddl/optimism/keeper.go
GMHDBJD: close when remove

Ehco1996: how about pass dbConfig into DownstreamMeta and lazy open conection ?

dm/dm/master/server.go
Ehco1996: so we also better to do this with pessimist ?

GMHDBJD: Optimistic may have some "Grayscale DDL" e.g. 
tb1: add column new_col1, tb2 add column new_col2. 
then both infos will be kept until they add another finally.
Pessimistic doesn't have this problem.

dm/pkg/shardddl/optimism/keeper.go
GMHDBJD: ece233f

dm/pkg/shardddl/optimism/lock.go
```diff
@@ -75,30 +82,94 @@ type Lock struct {
...
+	for _, schemaTables := range l.tables {
+		for _, tables := range schemaTables {
+			for _, ti := range tables {
+				if firstTable {
+					joined = ti
+					firstTable = false
+				} else {
+					newJoined, err := joined.Join(ti)
+					if err != nil {
+						log.L().Error(fmt.Sprintf("fail to join table info %s with %s", joined, ti), zap.String("lockID", l.ID), log.ShortError(err))
```
Ehco1996, line 162: just a question why can we mute this join err here ?

GMHDBJD: `joinTable` only use when create a lock. User may create a task with conflict upstream. e.g.
```
tb1(a int, b int);
tb2(a int, b varchar(4));
```
In this situation, init joined will error, but we can just ignore it. #4297 have a further implove.

dm/syncer/checkpoint.go
```diff
@@ -698,45 +699,68 @@ func (cp *RemoteCheckPoint) FlushPointsExcept(
...
+			var point *binlogPoint
+			// if point already in memory, use it
+			if tablePoints, ok := cp.points[sourceSchema]; ok {
+				if p, ok2 := tablePoints[sourceTable]; ok2 {
+					point = p
+				}
+			}
+			// create new point
+			if point == nil {
+				cp.saveTablePoint(table, cp.globalPoint.MySQLLocation(), nil)
```
lance6716, line 735: When some checkpoint SQL is failed, we agree that flushed table point should be nil, but can we keep (memory) table point as `ti` 🤔

lance6716 (commented):
will review later

dm/dm/master/server.go
```diff
@@ -573,6 +573,18 @@ func (s *Server) OperateTask(ctx context.Context, req *pb.OperateTaskRequest) (*
...
+		if len(req.Sources) == 0 {
+			err2 = s.optimist.RemoveMetaDataWithTask(req.Name)
```
lance6716, line 580: we should check the resp, make sure worker is really stopped before delete the table

GMHDBJD: Actually we remove the subtask in scheduler at L564, so even worker not really stopped, the meta should be removed.

GMHDBJD: And now master start/stop the task by update etcd, so it only depends on whether put/delete etcd successfully.

dm/pkg/shardddl/optimism/lock.go
```diff
@@ -548,7 +670,15 @@ func (l *Lock) addTables(tts []TargetTable) {
...
-					l.tables[tt.Source][schema][table] = l.joined
+					ti, err := l.FetchTableInfos(tt.Task, tt.Source, schema, table)
+					if err != nil {
+						log.L().Error("source table info not found, use joined table info instead", zap.String("task", tt.Task), zap.String("source", tt.Source), zap.String("schema", schema), zap.String("table", table), log.ShortError(err))
+						l.tables[tt.Source][schema][table] = l.joined
+					} else {
+						t := schemacmp.Encode(ti)
+						log.L().Debug("get source table info", zap.String("task", tt.Task), zap.String("source", tt.Source), zap.String("schema", schema), zap.String("table", table), zap.Stringer("info", t))
+						l.tables[tt.Source][schema][table] = t
+					}
```
GMHDBJD, line 681: main logic for #3786

dm/dm/master/server.go
```diff
@@ -573,6 +573,18 @@ func (s *Server) OperateTask(ctx context.Context, req *pb.OperateTaskRequest) (*
...
+		} else {
+			err2 = s.optimist.RemoveMetaDataWithTaskAndSources(req.Name, sources...)
+		}
+		if err2 != nil {
+			log.L().Error("failed to delete metadata for task", zap.String("task name", req.Name), log.ShortError(err2))
+		}
+	}
 	return resp, nil
 }
```
GMHDBJD, line 589: fix #3629

dm/pkg/shardddl/optimism/schema.go
```diff
@@ -1,151 +0,0 @@
-// Copyright 2020 PingCAP, Inc.
```
GMHDBJD, line 1: no init schema now, so fix #3708

dm/tests/shardddl1/run.sh
```diff
@@ -406,7 +413,7 @@ function DM_UpdateBARule_CASE() {
 
 	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
 		"show-ddl-locks" \
-		"\"ID\": \"test-\`shardddl\`.\`tb\`\"" 1
+		"no DDL lock exists" 1
```
lance6716, line 416: why this is no lock 😵‍💫

dm/tests/shardddl_optimistic/run.sh
```diff
@@ -0,0 +1,477 @@
...
+	run_sql_source2 "alter table ${shardddl1}.${tb1} drop column new_col1"
+	run_sql_source2 "insert into ${shardddl1}.${tb1} values(13);"
+
+	run_sql_tidb_with_retry "select count(1) from ${shardddl}.${tb}" "count(1): 13"
+
+	run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"stop-task test" \
+		"\"result\": true" 3
+
+	cp $cur/conf/double-source-optimistic.yaml $WORK_DIR/task.yaml
```
lance6716, line 394:
```suggestion
```

lance6716 (commented):
rest lgtm

dm/tests/shardddl1/run.sh
GMHDBJD: No DDL infos in DM master because we remove all of them when stop-task. So no lock exists when task starts until master receive first DDL info.

dm/pkg/shardddl/optimism/keeper.go
```diff
@@ -20,79 +20,63 @@ import (
...
 // LockKeeper used to keep and handle DDL lock conveniently.
 // The lock information do not need to be persistent, and can be re-constructed from the shard DDL info.
+// But the drop columns should be persistent.
 type LockKeeper struct {
 	mu    sync.RWMutex
 	locks map[string]*Lock // lockID -> Lock
+
+	downstreamMetaMap     map[string]*DownstreamMeta
+	getDownstreamMetaFunc func(string) (*config.DBConfig, string)
+	// column name -> source -> upSchema -> upTable -> int
```
lance6716, line 44:
```suggestion
	// lockID -> column name -> source -> upSchema -> upTable -> int
```

lance6716 (approved):
/cc @lichunzhu

Ehco1996 (approved):
LGTM

lance6716:
/hold

waiting @lichunzhu review

dm/syncer/checkpoint.go
```diff
@@ -712,45 +713,64 @@ func (cp *RemoteCheckPoint) FlushPointsExcept(
...
+	// should not happened
+	if len(tables) != len(tis) {
+		return errors.Errorf("the length of the tables is not equal to the length of the table infos, left: %d, right: %d", len(tables), len(tis))
 	}
 
-	location := point.MySQLLocation()
-	sql2, arg := cp.genUpdateSQL(sourceSchema, sourceTable, location, nil, tiBytes, false)
-	sqls = append(sqls, sql2)
-	args = append(args, arg)
+	batch := 100
```
lichunzhu, line 725: make 100 as a constant?

lichunzhu (commented):
Will review tests later. Have two questions now:
1. Should we delete obsolete init-schema for dm-clusters upgraded from lower versions?
2. Should we flush table info for newly added tables?

GMHDBJD:
> Should we delete obsolete init-schema for dm-clusters upgraded from lower versions?

I leave a TODO in 
https://github.com/pingcap/tiflow/blob/c9a821c2c23f45e5182d3e230e8fbc22ba345fde/dm/dm/common/common.go#L81  I think we should prune them after upgrading to a major version(e.g. 6.0)

> Should we flush table info for newly added tables

Yes. We still have some problem with newly create/drop table(e.g. #3823), I will create a new pr to solve them.

dm/dm/master/shardddl/optimist.go
```diff
@@ -395,12 +340,18 @@ func (o *Optimist) recoverLocks(
 	}
 
 	for _, info := range infos {
+		if info.IsDeleted {
+			// TODO: handle drop table
+			continue
+		}
+		if !o.tk.SourceTableExist(info.Task, info.Source, info.UpSchema, info.UpTable, info.DownSchema, info.DownTable) {
```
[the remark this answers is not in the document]
GMHDBJD: The logic is same as origin logic https://github.com/pingcap/tiflow/blob/f917d3651530db961d97cf7b704a306d11e46ed5/dm/pkg/shardddl/optimism/keeper.go#L62-L69

For create table while leader is transferred, there may be some problem. I will create a pr for create table with more tests.

dm/syncer/checkpoint.go
GMHDBJD: 5afac75

dm/pkg/shardddl/optimism/lock.go
```diff
@@ -548,7 +670,15 @@ func (l *Lock) addTables(tts []TargetTable) {
 			}
 			for table := range tables {
 				if _, ok := l.tables[tt.Source][schema][table]; !ok {
-					l.tables[tt.Source][schema][table] = l.joined
+					ti, err := l.FetchTableInfos(tt.Task, tt.Source, schema, table)
```
glorv, line 673: As `FetchTableInfos` fetch table schema info from table checkpoint, I think it may cause fetch outdated table info since we do ddl before flush checkpoint. A example:
Worker1:  --ddl1(tbl1) --------  execDDL1   -------- flushTbl1CP ------------->
Worker2:  --------- ddl2(tbl2) ------------- execDDL2 ----- flushTbl2Cp--->
Master:      -----------handleDDL1 --- handleDDL2------------------------->

In this case, when master start handle DDL2 (tbl2), tbl1's checkpoint didn't finished flushing, thus master will fetch outdated table schema for tbl1 and it may impact the correctness of the result.

GMHDBJD: This is a problem, but I don't think it's easy to solve under the existing architecture and it's not a big problem. First, `fetchtableinfo` is only called when the master receives the first DDL (to create a lock), that is, in most cases, DDL2 will not cause the master to call `fetchtableinfo`. In addition, we do not guarantee that DDLs under different sources are executed in order, that is to say, it is uncertain who executes DDL1 and DDL2 first, so the master can only coordinate according to the existing information. For source1, DDL1's Coordination is fine, for source2 there will be a problem if and only if DDL1 and DDL2 are updating the same column (eg DDL1 add column a, DDL2 drop column a). Finally, unless we can ensure that executing DDL and flushing checkpoint are atomic, this problem is difficult to solve. We have had such problems before, such as worker1 crash without flushing checkpoint after executing DDL1, worker2 executing DDL2, and then when master coordinating DDL2, it use the table structure after worker1 executes DDL1.

dm/dm/master/shardddl/optimist.go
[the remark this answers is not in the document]
GMHDBJD: #4497

dm/pkg/shardddl/optimism/lock.go
glorv: LGTM.

dm/tests/shardddl_optimistic/run.sh
```diff
@@ -0,0 +1,476 @@
...
+		restart_master
+	elif [[ "$mod" == "1" ]]; then
+		restart_worker1
+	elif [[ "$mod" == "2" ]]; then
+		restart_worker2
+	else
+		restart_task $cur/conf/double-source-optimistic.yaml
+	fi
+}
+
```
lichunzhu, line 243: Actually, I think it's better to test both four cases for integration tests. It will be easier for us to trace which PR introduces the bug.

dm/tests/shardddl_optimistic/run.sh
```diff
@@ -0,0 +1,476 @@
...
+	run_sql_source1 "insert into ${shardddl1}.${tb1} values(15,'iii');"
+	run_sql_source1 "insert into ${shardddl1}.${tb1} values(16,'jjj');"
+	run_sql_source2 "alter table ${shardddl1}.${tb1} drop column b;"
+	run_sql_source2 "insert into ${shardddl1}.${tb1} values(17,'kkk');"
+	run_sql_source2 "alter table ${shardddl1}.${tb2} add column c text;"
+	run_sql_source2 "insert into ${shardddl1}.${tb2} values(18,'lll');"
+
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+}
```
lichunzhu, line 350: These two functions look exactly the same.

lichunzhu (commented):
Rest LGTM

dm/tests/shardddl_optimistic/run.sh
GMHDBJD: removed

dm/tests/shardddl_optimistic/run.sh
GMHDBJD: I want to test different combinations of them. e.g.(restart master and then restart worker), but too many combinations, so use randome. I add some log for trace in [41cf696](https://github.com/pingcap/tiflow/pull/3903/commits/41cf696a23d1041a75b4cfc70d05f1a4e8cd0c41)

lichunzhu (approved):
LGTM

dm/tests/shardddl_optimistic/run.sh
lichunzhu: I see

PR #3932, opened 2021-12-17
capture(cdc), tcpserver: integrate P2P with server & capture

liuzix:
- Prepares basic dependencies for the peer-messaging based scheduler to work.

- Uses `pkg/tcpserver` to listen to the port.
- Runs the gRPC server.
- Runs the `MessageServer`.
- Maintains the `MessageServer`, `MessageRouter` and `ServerWrapper` in `Capture`.
- Injects `MessageServer` and `MessageRouter` into `context` so that the Owner and Processor can depend on them.
- Calls `SetOnCaptureAdded` and `SetOnCaptureRemoved` on the global state reactor so that the `MessageRouter` can be informed of capture membership changes.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


cdc/capture/capture.go
```diff
@@ -159,11 +175,32 @@ func (c *Capture) reset(ctx context.Context) error {
 	if c.grpcPool != nil {
 		c.grpcPool.Close()
 	}
+
+	if config.SchedulerV2Enabled {
+		c.grpcService.Reset(nil)
+
+		if c.MessageRouter != nil {
+			c.MessageRouter.Close()
+			c.MessageRouter.Wait()
```
overvenus, line 184: Can we wait  in `Close`?

overvenus (commented):
Rest LGTM

cdc/capture/capture.go
liuzix: Separating `Close` and `Wait` in the design of the interface is for two reasons:
1. To provide a finer-grained control over the timing of execution to achieve better unit-test coverage.
2. To avoid introducing deadlock bugs accidentally in the future. The pattern of `close-wait-close-wait` instead of `close-close-wait-wait` has created deadlock bugs before.

amyangfei (approved):

pkg/tcpserver/tcp_server.go
```diff
@@ -93,6 +95,9 @@ func NewTCPServer(address string, credentials *security.Credential) (TCPServer,
 
 // Run runs the mux. The mux has to be running to accept connections.
 func (s *tcpServerImpl) Run(ctx context.Context) error {
+	defer func() {
+		_ = s.rootListener.Close()
```
3AceShowHand, line 99: `s.rootListener.Close()` was called in `Run` and `Close`.

why do we have to call this method here ?

liuzix: `s.rootListener.Close()` provides a reliable way to tell downstream components such as the HTTP API server to exit.

3AceShowHand (approved):

GMHDBJD:
test failed due to https://github.com/pingcap/tiflow/issues/4029

Issue #1147, opened 2020-12-01
Improve usability of Avro


liuzix:
## Feature Request

**Is your feature request related to a problem? Please describe:**
Currently the schema subject name in the Schema Registry is hard coded as `{db}_{table}`, while users could have different expectations.

**Describe the feature you'd like:**
- [ ] Provide a sink-uri option to customize the schema subject name, and default to `{topic}` if nothing is specified.
- [ ] Support multi-topic changefeed for kafka sink.

**Teachability, Documentation, Adoption, Migration Strategy:**
- The first feature request (sink-uri option) is more urgently needed and easier to implement.
- To support scenarios where a large number of tables are being replicated, multi-topic support is needed.

keweishang:
Hi,

We're also facing a similar issue. We can't customize the Schema Registry's subject name and it is fixed as `{db}_{table}` and not configurable when creating the changefeed. Ideally, we'd like for the subject name to be `{topic}`, which is the default TopicNameStrategy and it will be much easier for downstream integration with Kafka. E.g. ksqlDB only supports TopicNameStrategy and we can not use ksqlDB to consume Kafka topics from TiCDC unless we name the Kafka topic as `{db}_{table}`. However, internal we have our own Kafka topic naming convention, and this limitation "breaks" our Kafka topic naming convention and makes integration with Kafka difficult. Thanks!

dveeden:
I think this issue can be closed as more flexible topic naming etc is now supported.

See also:

- https://docs.pingcap.com/tidb/stable/integrate-confluent-using-ticdc
- https://docs.pingcap.com/tidb/stable/release-6.1.0

PR #3936, opened 2021-12-17
avro: Add name strategy option for schema registry

dveeden:

This improves compatibility with KSQL

This is for issue #1147

When creating a changefeed with `--opts nameStrategy=topic` this now
uses the topic name as basis for the schema registry. Before this PR
the schema registry always used `{schema}_{table}` as basis.

This solves the case where KSQL expect the topic name to be used as
basis for the schema registry.

### Example

Setup a test environment and create the changefeed

```
tiup playground --without-monitor --tiflash 0 v5.3.0
confluent local services start
./bin/cdc server
./bin/cdc cli changefeed create --no-confirm \
--changefeed-id="simple-replication-task" \
--sort-engine="unified" \
--sink-uri="kafka://127.0.0.1:9092/cdctest?protocol=avro&kafka-version=2.8.0" \
--opts registry="http://127.0.0.1:8081" \
--opts nameStrategy=topic
```

To verify the `nameStrategy`:
```
$ ./bin/cdc cli changefeed query -c simple-replication-task | jq '.info.opts.nameStrategy'
"topic"
```

Now let's create a table that uses this changefeed:
```sql
CREATE TABLE t1 (
  id INT PRIMARY KEY, 
  n VARCHAR(200) NOT NULL, 
  ts TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6)
);
INSERT INTO t1(id, n) VALUES (1, 'test');
```

This now shows up in KSQL like this:
```
ksql> SHOW TOPICS;

 Kafka Topic                 | Partitions | Partition Replicas
---------------------------------------------------------------
 cdctest                     | 3          | 1
 default_ksql_processing_log | 1          | 1
---------------------------------------------------------------
ksql> PRINT cdctest FROM BEGINNING;
Key format: AVRO or KAFKA_STRING
Value format: AVRO
rowtime: 2021/12/17 07:06:07.227 Z, key: {"id": 1}, value: {"id": 1, "n": "test", "ts": 1639724765242}, partition: 0
^CTopic printing ceased
```

And in the schema registry:
```
$ curl -s http://127.0.0.1:8081/subjects | jq
[
  "cdctest-key",
  "cdctest-value"
]
```

This now allows us to do this:
```
ksql> CREATE STREAM t1str WITH (KAFKA_TOPIC='cdctest', VALUE_FORMAT='avro');

 Message
----------------
 Stream created
----------------

ksql> DESCRIBE t1str;

Name                 : T1STR
 Field | Type
-------------------------
 ID    | INTEGER
 N     | VARCHAR(STRING)
 TS    | TIMESTAMP
-------------------------
For runtime statistics and query details run: DESCRIBE <Stream,Table> EXTENDED;
```

Note that we didnt' have to specify the columns in the `CREATE STREAM`
statement as these were fetched from the schema registry.

See also:
- https://docs.ksqldb.io/en/latest/operate-and-deploy/schema-registry-integration/#schema-inference

#### Considerations

If you have a changefeed it maps to a single topic. With the `nameStrategy=topic` this means that there can only be one set of subjects registered in the registry. As soon as you try to use this topic for more than one table the registry will respond with a "HTTP 409 Conflict".

This doesn't have to be a problem as you could create multiple changefeeds, each with a filter for a single table and an unique topic.

This can even be seen as a feature as this will detect multiple incompatible things being written to the same topic.

Another solution is to implement multi-topic support, where a single changefeed creates one topic per table. However this would run into problems if topic auto creation is disabled, especially with online schema change tools like gh-ost and pt-osc which create temporary tables. However it might be possible to filter out OSC tables as these tables have predictable names.

Without this PR you could reach almost the same functionality by setting the topic in the sink URL to `{schema}_{table}`, this causes the topic name and the name in the schema registry to match for this table. However when a second table replicated via the same changefeed this will register itself under another name and won't result in an error, so this doesn't enforce schema consistency in the same way. Also when having multiple TiDB and CDC instances using the same Kafka infrastructure this makes it hard to keep topics unique as the `{schema}_{table}` name might not be unique across multiple sources. With this PR you could name the topic `{some_prefix}_{schema}_{table}` avoiding this problem.

### Check List 

Tests 

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
A new `nameStrategy` option was added to improve compatibility with Confluent KSQL
```


cdc/sink/codec/avro.go
```diff
@@ -542,13 +542,30 @@ func newAvroEventBatchEncoderBuilder(credential *security.Credential, opts map[s
 		return nil, cerror.ErrPrepareAvroFailed.GenWithStack(`Avro protocol requires parameter "registry"`)
 	}
 
+	topic, ok := opts["topic"]
+	if !ok {
+		topic = ""
+	}
+
+	nameStrategy, ok := opts["nameStrategy"]
+	if !ok {
```
3AceShowHand, line 551: how about just directly returning the error, since the first two cases won't be hit I think.

dveeden: Fixed in 15290e028456b3e0c00c413e1e135cb3d81604c2

keweishang:
Thank you @dveeden for your PR. It would help us a lot because we can now use single-table changefeed and name the Kafka topic as we want without having to match the topic name with the `{schema}_{table}`.

> Another solution is to implement multi-topic support, where a single changefeed creates one topic per table. However this would run into problems if topic auto creation is disabled, especially with online schema change tools like gh-ost and pt-osc which create temporary tables. However it might be possible to filter out OSC tables as these tables have predictable names.

[Debezium MySQL Connector](https://debezium.io/documentation/reference/connectors/mysql.html) uses the above mentioned implementation, it create one Kafka topic per table. Of course, we need to configure the Kafka cluster to enable auto topic creation (or create the topic manually before creating the table), but that doesn't raise any serious issue for us. Debezium also has `database.exclude.list` and `table.exclude.list` [configuration](https://debezium.io/documentation/reference/connectors/mysql.html#mysql-property-table-exclude-list) to filter out temp tables created by gh-ost and pt-osc.

cdc/sink/codec/avro.go
```diff
@@ -542,13 +542,31 @@ func newAvroEventBatchEncoderBuilder(credential *security.Credential, opts map[s
 		return nil, cerror.ErrPrepareAvroFailed.GenWithStack(`Avro protocol requires parameter "registry"`)
 	}
 
+	topic, ok := opts["topic"]
+	if !ok {
+		topic = ""
+	}
+
+	nameStrategy, ok := opts["nameStrategy"]
```
keweishang, line 550: Is this new option `nameStrategy` mandatory or optional. Is the default value of this new option `legacy`?

keweishang (approved):

dveeden:
@keweishang Thanks for the information. That helps a lot.

cdc/sink/codec/avro.go
dveeden: The `nameStrategy` is optional and the default is `legacy`. Only when set explicitly it will use the new behavior.

cdc/sink/mq.go
```diff
@@ -394,6 +394,7 @@ func newKafkaSaramaSink(ctx context.Context, sinkURI *url.URL, filter *filter.Fi
 	if topic == "" {
 		return nil, cerror.ErrKafkaInvalidConfig.GenWithStack("no topic is specified in sink-uri")
 	}
+	opts["topic"] = topic
```
hi-rustin, line 397: Maybe this needs to be put into CompleteConfigsAndOpts?

dveeden: Done in 20e99788d3c3042ad9774102d5274316060c06fe

PTAL

zhangyangyu:
In 6.1, with the implementation of #5338 and #4423 , now it's possible to achieve what this PR does. Generally speaking, you could use one changefeed to replicate multiple tables to kafka, each table corresponds to one topic. This is the solution 2 described by @dveeden in the PR description.

I think we could close this PR now.

Issue #3749, opened 2021-12-06
Rename repo name to pingcap/tiflow


amyangfei:
We are going to rename `pingcap/ticdc` to `pingcap/tiflow`

# Motivation
This repo maintains both TiCDC and DM, the repo name doesn't match the product exactly. We are planning to use a more accurate repo name to make it self-explanatory.

# Procedure

Filing an issue to sort through what needs to be done:

 - [x] 1. Announce to TiDB/TiKV communities via https://internals.tidb.io/. (prepare on ~2021/12/11-2021/12/12~, 2021/12/18-2021/12/19) https://internals.tidb.io/t/topic/512
 - [x] 2. Pause CI, prepare a PR to replace import path.
 - [x] 3. Rename the repo from `pingcap/ticdc` to `pingcap/tiflow`.
 - [x] 4. Reconfigure CI, GitHub bot and automated testing.
 - [x] 5. Prepare and merge PR that changes import path in master branch.
 - [x] 6. Cherry pick import patch PR to release branches.
 - [x] 7. Reconfigure CI, GitHub bot and automated testing for release branches.
 - [x] 8. Check whether we need to update related documents.
 - [ ] 9. Refine `README.md` in root directory.
 - [x] 10. Update dependency in other repos, such as https://github.com/pingcap/tidb-tools

# Some related issues

- Does release branche need cherry-pick
Yes, since we will change the repo name and go version mechanism, we need to cherry the path change to release branches.

- Are released TiCDC urls and go mod versions still available
Yes. GitHub will [redirect all requests to the new URL](https://github.blog/2013-05-16-repository-redirects-are-here/) and go mod download still works. **All old versions can be built as before**.

- Is there any change to feature release?
No, we still release TiCDC and DM as before, there **is no any change to release binary and product usage**, they are independent binaries and follow the same release procedure as TiDB.

amyangfei:
Some remaining problems
- Github action for Arm64 build is failed.

hi-rustin:
> Some remaining problems
> 
>     * [ ]  Github action for Arm64 build is failed.

It may have nothing to do with it, it may simply be that the arm machine went offline. I'll confirm it next Monday. Because it looks like the x86 machine is still online.

lance6716:
I did these steps to my existing PR

```
sed -i "s/github.com\/pingcap\/ticdc/github.com\/pingcap\/tiflow/g" $(find . -type f|grep -vE "bin/|.git")
[...]

PR #3965, opened 2021-12-18
github: replace pingcap/ticdc with pingcap/tiflow in workflow

amyangfei:

Some remaining `pingcap/ticdc`, finished the remaining problem-2 in https://github.com/pingcap/tiflow/issues/3749

**The DM github action tests are verified in #3966.**

These files were introduced from `v5.3.0`, so cherry pick to `release-5.3` is enough.

replace `pingcap/ticdc` with `pingcap/tiflow` in files from `.github` dir

### Check List 

Tests 

 - Unit test
 - Integration test
 
```release-note
None
```


lance6716 (approved):
Do this also has a script to make changes so we can replay the script to our PR?

lonng (approved):

PR #3975, opened 2021-12-20
checker(dm): support concurrent check

okJiang:

Issue Number: close https://github.com/pingcap/tiflow/issues/3974

- support concurrent check table_structure and table_consistency
- the checker thread follows the change of `MydumperConfig.Threads`

### Check List 

Tests 

 - Unit test
 - Manual test

####  10000 tables with default dumpconfig.thread
|added delay|spend time|
|-----------|---------|
|10 ms|31s|
|20 ms|56s|
|50 ms|130s|
|100 ms|175s|

Code changes

Related changes

### Release note 

```release-note
Please add a release note.
```


okJiang:
test passed all. /cc @Ehco1996 @GMHDBJD

dm/tests/checker/run.md
```diff
@@ -0,0 +1,20 @@
+# manual test
```
Ehco1996, line 1: i think this manual test scriptis need not be included in our repo ?

maybe just prepra a doc and link the doc to PR description ?

dm/pkg/checker/utils.go
```diff
@@ -141,3 +144,35 @@ func isMySQLError(err error, code uint16) bool {
...
+	threads := tableNum/tablesWithOneThread + 1
+	for _, num := range cunrrencies {
+		if num >= threads {
+			return num
+		}
+	}
+	return cunrrencies[len(cunrrencies)-1]
+}
+
+func initShardingMock(mock sqlmock.Sqlmock) sqlmock.Sqlmock {
```
Ehco1996, line 166: this func should move to xxx_test.go

Ehco1996 (commented):
will reveiw table_structure.go  later 😂 there one is too large

dm/tests/checker/run.md
okJiang: It is ok for me.

dm/pkg/checker/utils.go
```diff
@@ -141,3 +143,21 @@ func isMySQLError(err error, code uint16) bool {
...
+
+var concurrencies = []int{4, 8, 12, 16, 20, 24, 28, 32}
+
+func getConcurrency(tableNum int) int {
+	// if tableNum < 100, one thread is enough
+	if tableNum < 100 {
+		return 1
+	}
+	threads := tableNum/tablesWithOneThread + 1
+	for _, num := range concurrencies {
```
GMHDBJD, line 157: Can't we calculate the result directly?

dm/pkg/checker/table_structure.go
```diff
@@ -267,114 +311,207 @@ func (c *TablesChecker) checkTableOption(opt *ast.TableOption) *incompatibilityO
...
+				r.Extra = fmt.Sprintf("error on sharding %s", c.targetTableID)
+				r.Instruction = "please set same table structure for sharding tables"
+				return
+			}
+		}
+	}
+
+	for instance, tables := range c.tableMap {
+		concurrency := getConcurrency(len(tables))
+		maxConnections, err := utils.GetMaxConnections(ctx, c.dbs[instance])
```
GMHDBJD, line 477: Still not recommend using max connection. Also if there are some other clients hold the connection?

dm/pkg/checker/table_structure.go
```diff
@@ -267,114 +311,207 @@ func (c *TablesChecker) checkTableOption(opt *ast.TableOption) *incompatibilityO
...
+			var checkTables []*filter.Table
+			if i+1 == concurrency || onethread*(i+1) >= len(tables) {
+				checkTables = tables[onethread*i:]
+			} else {
+				checkTables = tables[onethread*i : onethread*(i+1)]
+			}
+			checkWg.Add(1)
+			go checkFunc(instance, checkTables)
+		}
+	}
```
GMHDBJD, line 504: Why not just determine the value of concurrency and then use something like the producer-consumer model?

dm/pkg/checker/table_structure.go
okJiang: This is indeed a problem, but this is just to prevent the upstream from reporting errors **to a certain extent**. Because we also can't get the real number of upstream connections.

dm/pkg/checker/table_structure.go
okJiang: > use something like the producer-consumer model?

I haven't written many concurrent programs before... So I just wrote it according to my own feelings.

I think there's not much difference between them. Now it is equivalent to passing all the tables that need to be consumed to consumers at one time.

If you think producer-consumer model is better, I can update this to producer-consumer model..

dm/pkg/checker/table_structure.go
Ehco1996: yes, we should not use whole max connection, use `32` as the max concurrencies is enough when `32` < server max connection due to your test report

i say `adjust concurrent by server's max connection` means we should not try open a connection pool that hold connection count larger than server allow (this may get an error from server, not tested)

dm/pkg/checker/table_structure.go
```diff
@@ -60,71 +71,106 @@ func (o *incompatibilityOption) String() string {
...
 		}
 	}
 
-	for name, opts := range options {
-		if len(opts) == 0 {
-			continue
+	for i := 0; i < concurrency; i++ {
+		if onethread == 0 {
+			checkWg.Add(1)
+			go checkFunc(c.tables[0:])
```
Ehco1996, line 144: when start a new backaround thread , we better to pass ctx in, and use selcet to listen this ctx

this is how golang's way to handle timeout or interrupts

dm/pkg/checker/table_structure.go
Ehco1996: agrees with GMHDBJD `producer-consumer model`  is good for maintenance eg:

if i want to change the concurent count , i just incr the worker count
if i want to change the dispatch job logic, i just need to change the producer side
if i want to change the job handle logic, i just need to c change the worker side logic

dm/pkg/utils/db.go
```diff
@@ -632,3 +632,23 @@ func GetTableCreateSQL(ctx context.Context, conn *sql.Conn, tableID string) (sql
...
+	c, err := db.Conn(ctx)
+	if err != nil {
+		return 0, err
+	}
+	defer c.Close()
+	return GetMaxConnectionsForConn(ctx, c)
+}
+
+// GetMaxConnectionsForConn gets max_connections for sql.Conn which is suitable for session variable max_connections.
+func GetMaxConnectionsForConn(ctx context.Context, conn *sql.Conn) (int, error) {
```
Ehco1996, line 647: seems only `GetMaxConnections` used `GetMaxConnectionsForConn` why split this to two func ?

dm/pkg/checker/table_structure.go
```diff
@@ -60,71 +74,128 @@ func (o *incompatibilityOption) String() string {
...
+	concurrency := len(c.tables)/tablesForOneThread + 1
+	if concurrency > maxThreadNum {
+		concurrency = maxThreadNum
+	}
+	maxConnections, err := utils.GetMaxConnections(ctx, c.db)
+	if err != nil {
+		return r, err
+	}
+	if concurrency > maxConnections {
+		concurrency = maxConnections
```
Ehco1996, line 110:
```suggestion
		concurrency = maxConnections / 2
```
as @GMHDBJD  suggestions, we better not use all connections 😬

dm/pkg/checker/table_structure.go
```diff
@@ -60,71 +74,128 @@ func (o *incompatibilityOption) String() string {
...
 
-	for name, opts := range options {
-		if len(opts) == 0 {
-			continue
+	go func() {
+		defer cancel()
+		err, ok := <-errCh
+		if ok {
+			markCheckError(r, err)
+			close(errCh)
```
Ehco1996, line 169: we should only close a channel in a sender goroutine if the sender is the only sender of the channel see more [here](https://go101.org/article/channel-closing.html)

dm/pkg/checker/table_structure.go
```diff
@@ -60,71 +74,128 @@ func (o *incompatibilityOption) String() string {
...
-			if err != nil {
-				markCheckError(r, err)
-				return r
+
+	checkCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	go func() {
+		defer close(inCh)
+		for _, table := range c.tables {
+			select {
```
Ehco1996, line 125: this select should move to out of this for loop? otherwise it will selecet N times 🤔

Ehco1996: ohoho, this thread is producer, so maybe no need to do this job in a gorountine, this is very quick beacuse `inCh` have enough buffer

dm/pkg/checker/table_structure.go
```diff
@@ -60,71 +74,128 @@ func (o *incompatibilityOption) String() string {
...
+			case <-checkCtx.Done():
+				return
+			case table, ok := <-inCh:
+				if !ok {
+					return
+				}
+				statement, err := dbutil.GetCreateTableSQL(ctx, c.db, table.Schema, table.Name)
+				if err != nil {
+					// continue if table was deleted when checking
```
Ehco1996, line 145: just a question: why we can tolatate this case?

dm/pkg/checker/table_structure.go
```diff
@@ -60,71 +74,128 @@ func (o *incompatibilityOption) String() string {
...
+					if isMySQLError(err, mysql.ErrNoSuchTable) {
+						continue
+					}
+					if len(errCh) == 0 {
```
Ehco1996, line 149: how about not use errch , just call cancel to stop checking ?

dm/pkg/checker/table_structure.go
```diff
@@ -267,114 +336,198 @@ func (c *TablesChecker) checkTableOption(opt *ast.TableOption) *incompatibilityO
...
 		parser2, err := dbutil.GetParserForDB(ctx, db)
 		if err != nil {
-			markCheckError(r, err)
-			r.Extra = fmt.Sprintf("fail to get parser for instance %s on sharding %s", instance, c.name)
-			return r
+			r.Extra = fmt.Sprintf("fail to get parser for instance %s on sharding %s", instance, c.targetTableID)
+			return r, err
+		}
+		r.Extra = fmt.Sprintf("instance %s on sharding %s", instance, c.targetTableID)
+		for _, table := range tables {
```
Ehco1996, line 388: how about use fisrstTable = tables[0]

dm/pkg/checker/table_structure.go
```diff
@@ -267,114 +336,198 @@ func (c *TablesChecker) checkTableOption(opt *ast.TableOption) *incompatibilityO
...
 		firstTable    string
 		firstInstance string
+		errCh         = make(chan error, 1)
+		inChs         = make(map[string]chan *filter.Table, len(c.tableMap))
+		checkWg       sync.WaitGroup
 	)
 
-	for instance, schemas := range c.tables {
+	// get first table
+	for instance, tables := range c.tableMap {
```
Ehco1996, line 376: can we also do this in `checkFunc`, not sure if there will be too many shards(such as 200 source) want to check 🤔

dm/pkg/utils/db.go
okJiang: I'm just imitating `GetParserForConn`.

dm/pkg/checker/table_structure.go
okJiang: This comes from the previous code, and I don’t think this case will happen.

dm/pkg/checker/table_structure.go
okJiang: My thought is it can check `checkCtx.Done()` every time. Because if `checkCtx.Done()` and `inCh` is bloken, this thread will dead lock?

dm/pkg/checker/table_structure.go
okJiang: just execute once whatever source numbers.

dm/pkg/checker/table_structure.go
```diff
@@ -60,87 +74,121 @@ func (o *incompatibilityOption) String() string {
...
-				e.Severity = StateWarning
-				e.Instruction = option.instruction
-				r.Errors = append(r.Errors, e)
-			case StateFailure:
-				r.State = StateFailure
-				e := NewError(tableMsg + option.errMessage)
-				e.Instruction = option.instruction
-				r.Errors = append(r.Errors, e)
-			}
+	defer func() {
```
Ehco1996, line 178:
```suggestion
```

Ehco1996: we can wait at the end of func, so no need to use defer

dm/pkg/checker/table_structure.go
```diff
@@ -60,87 +74,121 @@ func (o *incompatibilityOption) String() string {
...
+		checkWg.Wait()
+		log.L().Logger.Info("check table structure over", zap.Time("start time", startTime), zap.String("speed time", time.Since(startTime).String()))
+	}()
+
+	for _, table := range c.tables {
```
Ehco1996, line 183:
```suggestion
	for _, table := range c.tables {
	     inCh <- table
	}
```
no need to  use select here because i think enqueue is very fast because `inCh    = make(chan *filter.Table, len(c.tables))`

dm/pkg/checker/table_structure.go
Ehco1996: this why i said consumer should not close queue

dm/pkg/checker/table_structure.go
glorv: +1

glorv (commented):
rest LGTm

dm/pkg/checker/table_structure.go
```diff
@@ -82,65 +96,92 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
 		Extra: fmt.Sprintf("address of db instance - %s:%d", c.dbinfo.Host, c.dbinfo.Port),
 	}
 
+	startTime := time.Now()
+
+	concurrency := len(c.tables)/tablesForOneThread + 1
```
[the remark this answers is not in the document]
okJiang: If source numbers is very large(> 1000), we should open >32000 goroutines,  will it affect local machine?🤔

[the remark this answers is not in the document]
glorv: The `maxThreadNum ` is `32`, which means we will open at most 32 routine to do this? I don't get how `32000 ` is calculated.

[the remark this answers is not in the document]
okJiang: `maxThreadNum` means goroutine<=32 for every source.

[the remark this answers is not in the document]
glorv: I think all sources should share the single `32` routine pool

dm/pkg/checker/table_structure.go
```diff
@@ -149,8 +203,8 @@ func (c *TablesChecker) Name() string {
 	return "table structure compatibility check"
 }
 
-func (c *TablesChecker) checkCreateSQL(ctx context.Context, statement string) []*incompatibilityOption {
-	parser2, err := dbutil.GetParserForDB(ctx, c.db)
+func (c *TablesChecker) checkCreateSQL(ctx context.Context, db *sql.DB, statement string) []*incompatibilityOption {
+	parser2, err := dbutil.GetParserForDB(ctx, db)
```
glorv, line 207: Maybe be we can prefetch to sql-mode and cache it with db to avoid fetch the sql-mode for every table

dm/pkg/checker/table_structure.go
```diff
@@ -79,68 +95,106 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
...
+	for sourceID, tables := range c.tableMap {
+		for _, table := range tables {
+			isDone := false
+			select {
+			case inCh <- &checkItem{table, sourceID}:
+			case <-checkCtx.Done():
+				isDone = true
+			}
+			if isDone {
+				break
```
glorv, line 190:
```suggestion
outer:
	for sourceID, tables := range c.tableMap {
		for _, table := range tables {
			select {
			case inCh <- &checkItem{table, sourceID}:
			case <-checkCtx.Done():
				break outer
			}
```

dm/pkg/checker/table_structure.go
```diff
@@ -79,68 +95,106 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
...
 			}
 		}
 	}
 
+	close(inCh)
+	checkWg.Wait()
+	log.L().Logger.Info("check table structure over", zap.Time("start time", startTime), zap.String("speed time", time.Since(startTime).String()))
```
glorv, line 197: If there are none tables checked due to ctx.Done, we should check and adjust the check result to Canceled if it is still succeeded.

glorv (commented):
rest LGTM

dm/pkg/checker/table_structure.go
okJiang: where does `outer` use?

glorv: break's label at L220

dm/pkg/checker/table_structure.go
```diff
@@ -58,19 +76,17 @@ func (o *incompatibilityOption) String() string {
 // In generally we need to check definitions of columns, constraints and table options.
 // Because of the early TiDB engineering design, we did not have a complete list of check items, which are all based on experience now.
 type TablesChecker struct {
-	db     *sql.DB
-	dbinfo *dbutil.DBConfig
-	tables map[string][]string // schema => []table; if []table is empty, query tables from db
-
+	dbs      map[string]*sql.DB
+	tableMap map[string][]*filter.Table
```
Ehco1996, line 80: please point out that key of tableMap is `sourceID`

dm/pkg/checker/table_structure.go
```diff
@@ -296,88 +361,183 @@ func (c *ShardingTablesChecker) Check(ctx context.Context) *Result {
...
+		if !ok {
+			markCheckError(r, errors.NotFoundf("client for instance %s", instance))
+			return r
+		}
+		maxConnections, err := utils.GetMaxConnections(ctx, db)
+		if err != nil {
+			markCheckError(r, err)
+			return r
+		}
+		concurrency = int(math.Min(float64(concurrency), float64(maxConnections/2)))
```
Ehco1996, line 518: i am sorry , but there is chance that checker will nerver run when `maxConnections=1` 

maybe use math.Ceil to handel this

```
	// find min max db connection count across all sources
	connectionCnt := math.MaxInt
	for instance := range c.tableMap {
		db, ok := c.dbs[instance]
		if !ok {
			markCheckError(r, errors.NotFoundf("client for instance %s", instance))
			return r
		}
		maxConnections, err := utils.GetMaxConnections(ctx, db)
		if err != nil {
			markCheckError(r, err)
			return r
		}
		if maxConnections < connectionCnt {
			connectionCnt = maxConnections
		}
	}
	concurrency := math.Min(maxThreadNum, math.Ceil(float64(connectionCnt)/2))
```

dm/pkg/checker/table_structure.go
```diff
@@ -296,88 +361,183 @@ func (c *ShardingTablesChecker) Check(ctx context.Context) *Result {
...
-					return r
+					reMu.Unlock()
+					return
 				}
 			}
 		}
 	}
 
+	concurrency := maxThreadNum
+	for instance := range c.tableMap {
```
Ehco1996, line 507:
```suggestion
	for sourceID := range c.tableMap {
```
and adjust rest logic

dm/pkg/checker/table_structure.go
```diff
@@ -79,68 +95,128 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
...
+			case <-checkCtx.Done():
+				return
+			case checkItem, ok := <-inCh:
+				if !ok {
+					return
+				}
+				table := checkItem.table
+				if instance != checkItem.sourceID {
+					instance = checkItem.sourceID
+					p, err = dbutil.GetParserForDB(ctx, c.dbs[instance])
```
Ehco1996, line 137: can we share global parser with checker?,(not sure `ParseOneStmt` is thread safe), if so, we can move the operation of change parser for db out of for loop

dm/pkg/checker/table_structure.go
```diff
@@ -296,88 +361,183 @@ func (c *ShardingTablesChecker) Check(ctx context.Context) *Result {
...
 		Desc:  "check consistency of sharding table structures",
 		State: StateSuccess,
-		Extra: fmt.Sprintf("sharding %s", c.name),
+		Extra: fmt.Sprintf("sharding %s,", c.targetTableID),
 	}
 
+	startTime := time.Now()
+	log.L().Logger.Info("start to check sharding tables")
 	var (
 		stmtNode      *ast.CreateTableStmt
```
Ehco1996, line 370: how about change `stmtNode` to `firstCreateTableStmtNode`

dm/pkg/checker/table_structure.go
```diff
@@ -296,88 +361,183 @@ func (c *ShardingTablesChecker) Check(ctx context.Context) *Result {
...
+		return r
+	}
+
+	stmt, err := parser2.ParseOneStmt(statement, "", "")
+	if err != nil {
+		markCheckError(r, errors.Annotatef(err, "statement %s", statement))
+		return r
+	}
+
+	stmtNode, ok = stmt.(*ast.CreateTableStmt)
```
Ehco1996, line 408: how about wrap get stmtNode code (L396~L406) to a func? this can resued in  `checkFunc`

dm/pkg/checker/table_structure.go
```diff
@@ -296,88 +361,183 @@ func (c *ShardingTablesChecker) Check(ctx context.Context) *Result {
...
+					err = errors.Errorf("Expect CreateTableStmt but got %T", stmt)
+					return
 				}
 
 				if c.checkAutoIncrementPrimaryKey {
-					passed := c.checkAutoIncrementKey(instance, schema, table, ctStmt, info, r)
+					passed := c.checkAutoIncrementKey(instance, table.Schema, table.Name, ctStmt, info, r)
 					if !passed {
-						return r
+						return
```
Ehco1996, line 488: if checkAutoIncrementKey not passed, do we need to append a error  in `r.Errors` ? if we return here how checker know this?

dm/pkg/checker/table_structure.go
```diff
@@ -79,68 +95,128 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
...
+
+	checkCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	checkFunc := func() {
+		defer checkWg.Done()
+		instance := firstInstance
+		p, err := dbutil.GetParserForDB(ctx, c.dbs[instance])
+		if err != nil {
+			reMu.Lock()
```
GMHDBJD, line 121: prefer return error in checkFunc and let outer function call `markCheckError`

dm/pkg/checker/table_structure.go
```diff
@@ -79,68 +95,128 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
...
-					continue
+	for instance := range c.tableMap {
+		firstInstance = instance
+		break
+	}
+
+	checkCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	checkFunc := func() {
```
GMHDBJD, line 116: How about use a named function rather than closure?

dm/pkg/checker/table_structure.go
```diff
@@ -18,23 +18,41 @@ import (
...
+	"go.uber.org/zap"
+
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"github.com/pingcap/tiflow/dm/pkg/utils"
+)
+
+const (
+	// AutoIncrementKeyChecking is an identification for auto increment key checking.
+	AutoIncrementKeyChecking = "auto-increment key checking"
+	maxThreadNum             = 32
```
GMHDBJD, line 44: dumpling use 4 as default value, how about set it to the same value.

okJiang: The only effect is that the time will be quadrupled. At the beginning, @sunzhaoyang  said he hopes the time can be controlled within 1 minute.

dm/pkg/checker/table_structure.go
okJiang: can't, `ParseOneStmt` is not thread safe.

dm/pkg/checker/table_structure_test.go
```diff
@@ -108,10 +83,10 @@ func (t *testCheckSuite) TestShardingTablesChecker(c *tc.C) {
 	mock.ExpectQuery("SHOW CREATE TABLE `test-db`.`test-table-2`").WillReturnRows(createTableRow2)
 
 	result = checker.Check(ctx)
+	printJSON(result)
```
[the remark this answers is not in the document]
okJiang: yep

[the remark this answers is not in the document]
Ehco1996: so we better to delete this code, if you want to keep this log, please use log.Debug instead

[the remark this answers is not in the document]
okJiang: I will update related UT after logic approved.

dm/pkg/checker/table_structure.go
okJiang: I think `(maxConnections +1)/2` is solved this comment

dm/pkg/checker/table_structure.go
GMHDBJD: Did you have a benchmark? For some customer, they would rather take a little longer than affect the upstream. And I think the time needed for checking is less than the time needed for dump data.

dm/pkg/checker/table_structure.go
```diff
@@ -79,68 +103,67 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
...
+		c.firstSourceID = sourceID
+		break
+	}
 
-			opts := c.checkCreateSQL(ctx, statement)
-			if len(opts) > 0 {
-				options[tableName] = opts
-				statements[tableName] = statement
-			}
+	checkCtx, cancel := context.WithCancel(ctx)
```
GMHDBJD, line 115: Seems this WithCancel is no need because you didn't call cancel() in this function.

dm/pkg/checker/table_structure.go
okJiang: test result is 

####  10000 tables
|added delay|spend time|
|-----------|---------|
|10 ms|9.5s|
|20 ms|15.6s|
|50 ms|35s|
|100 ms|68.5s|

I will discuss about upstream affact with PM further.

dm/pkg/checker/table_structure.go
```diff
@@ -79,68 +103,67 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
...
+			}
+			c.reMu.Lock()
+			markCheckError(r, err)
+			c.reMu.Unlock()
+		case <-checkCtx.Done():
+			break outer2
+		}
+	}
+	log.L().Logger.Info("check table structure over", zap.String("spend time", time.Since(startTime).String()))
 	return r
```
GMHDBJD, line 167: how about extract to one or two functions for better review and test? e.g. func dispatchTables ...

dm/pkg/checker/table_structure.go
```diff
@@ -149,41 +172,78 @@ func (c *TablesChecker) Name() string {
...
+	p, err := dbutil.GetParserForDB(ctx, c.dbs[sourceID])
 	if err != nil {
-		return []*incompatibilityOption{
-			{
-				state:      StateFailure,
-				errMessage: err.Error(),
-			},
-		}
+		c.errCh <- err
+		c.cancel()
```
GMHDBJD, line 181: Where is c.cancel be assigned?

dm/pkg/checker/table_structure.go
okJiang: SGTM

okJiang: extract to two functions `dispatchCheckItem()` and `handleErr()` in https://github.com/pingcap/tiflow/pull/3975/commits/90442d12cf17c79db816a5cba7595a0cc08987cc

dm/pkg/checker/mysql_server.go
```diff
@@ -124,13 +124,12 @@ func (pc *MySQLServerIDChecker) Check(ctx context.Context) *Result {
 
 	serverID, err := dbutil.ShowServerID(ctx, pc.db)
 	if err != nil {
-		if utils.OriginError(err) == sql.ErrNoRows {
-			result.Errors = append(result.Errors, NewError("server_id not set"))
-			result.Instruction = "please set server_id in your database"
-		} else {
+		if utils.OriginError(err) != sql.ErrNoRows {
```
GMHDBJD, line 127: Any different? 😂

dm/pkg/checker/table_structure.go
```diff
@@ -546,5 +650,50 @@ func getBriefColumnList(stmt *ast.CreateTableStmt) briefColumnInfos {
...
-	return fmt.Sprintf("sharding table %s consistency checking", c.name)
+	return fmt.Sprintf("sharding table %s consistency checking", c.targetTableID)
+}
+
+func dispatchTableItem(ctx context.Context, tableMap map[string][]*filter.Table, inCh chan *checkItem) {
+	for sourceID, tables := range tableMap {
+		for _, table := range tables {
+			select {
+			case inCh <- &checkItem{table, sourceID}:
+			case <-ctx.Done():
```
GMHDBJD, line 661: I think ctx has a higher priority

dm/pkg/checker/table_structure.go
```diff
@@ -58,19 +75,27 @@ func (o *incompatibilityOption) String() string {
...
 type TablesChecker struct {
-	db     *sql.DB
-	dbinfo *dbutil.DBConfig
-	tables map[string][]string // schema => []table; if []table is empty, query tables from db
-
+	dbs           map[string]*sql.DB
+	tableMap      map[string][]*filter.Table // sourceID => {[table1, table2, ...]}
+	checkWg       sync.WaitGroup
+	reMu          sync.Mutex
+	firstSourceID string
```
GMHDBJD, line 82: Why need firstSourceID? Can't we set it in Check when meet first Table?

dm/pkg/checker/table_structure.go
```diff
@@ -149,41 +144,78 @@ func (c *TablesChecker) Name() string {
...
+					e.Severity = StateWarning
+					e.Instruction = opt.instruction
+					r.Errors = append(r.Errors, e)
+				case StateFailure:
+					r.State = StateFailure
+					e := NewError(tableMsg + opt.errMessage)
+					e.Instruction = opt.instruction
+					r.Errors = append(r.Errors, e)
+				}
+				c.reMu.Unlock()
```
GMHDBJD, line 212: can we call make check error here? or send the result to handleErr/handleResult

dm/pkg/checker/table_structure.go
```diff
@@ -149,41 +144,78 @@ func (c *TablesChecker) Name() string {
...
 	if err != nil {
-		return []*incompatibilityOption{
-			{
-				state:      StateFailure,
-				errMessage: err.Error(),
-			},
-		}
+		c.errCh <- err
+		c.cancel()
+		return
```
GMHDBJD, line 154: So you want to cancel all check progress if any error happended? If so, I think [errGroup](https://pkg.go.dev/golang.org/x/sync/errgroup) meets your requriment. You can just return when error occur.

GMHDBJD (commented):
How about add some integration test.

dm/pkg/checker/mysql_server.go
okJiang: no difference...

dm/pkg/checker/utils.go
```diff
@@ -136,7 +138,9 @@ func markCheckError(result *Result, err error) {
 		} else {
 			state = StateFailure
 		}
-		result.State = state
+		if result.State != StateFailure {
```
Ehco1996, line 141: could please explian why update state only when state not equal

dm/pkg/checker/table_structure.go
```diff
@@ -546,5 +622,35 @@ func getBriefColumnList(stmt *ast.CreateTableStmt) briefColumnInfos {
...
+func dispatchTableItem(ctx context.Context, tableMap map[string][]*filter.Table, inCh chan *checkItem) {
+	for sourceID, tables := range tableMap {
+		for _, table := range tables {
+			select {
+			case <-ctx.Done():
+				log.L().Logger.Warn("ctx canceled before input tables completely")
+				return
+			default:
+			}
+			inCh <- &checkItem{table, sourceID}
```
Ehco1996, line 637: why not put this line in default branch

dm/pkg/checker/table_structure.go
```diff
@@ -296,96 +331,137 @@ func (c *ShardingTablesChecker) Check(ctx context.Context) *Result {
...
+
+			ctStmt, err := getCreateTableStmt(p, statement)
+			if err != nil {
+				return err
+			}
+
+			if c.checkAutoIncrementPrimaryKey {
+				passed := c.checkAutoIncrementKey(sourceID, table.Schema, table.Name, ctStmt, info, r)
+				if !passed {
+					return nil
```
Ehco1996, line 442: the beheavior of `checkAutoIncrementKey` and `checkConsistency` is  not same, checkAutoIncrementKey handle check result inner but checkConsistency not, this confused me when i see 
```
if !passed {
  return nil
}
```
could please adjust this two checker? maybe move 
```
		if !hasMatchedRule {
			r.State = StateFailure
			r.Errors = append(r.Errors, NewError("sourceID %s table `%s`.`%s` of sharding %s have auto-increment key %s and column mapping, but type of %s should be bigint", sourceID, schema, table, c.targetTableID, columnName, columnName))
			r.Instruction = "please handle it by yourself"
			r.Extra = AutoIncrementKeyChecking
			return false
		}
```
to outer like checkConsistency

Ehco1996 (commented):
alse please check why ut blocked , https://ci.pingcap.net/blue/organizations/jenkins/ut-check/detail/ut-check/613/pipeline/
```
panic: test timed out after 5m0s
```

happy new year 🎉

dm/checker/check_test.go
```diff
@@ -247,131 +247,6 @@ func (s *testCheckerSuite) TestBinlogRowImageChecking(c *tc.C) {
 	c.Assert(err, tc.IsNil)
 }
 
-func (s *testCheckerSuite) TestTableSchemaChecking(c *tc.C) {
```
okJiang, line 250: this has been tested in https://github.com/pingcap/tiflow/blob/f710abba6417971370e0d0711d9cd8ca3fa784c4/dm/tests/dmctl_basic/run.sh#L284

the task file is https://github.com/pingcap/tiflow/blob/master/dm/tests/dmctl_basic/conf/dm-task3.yaml

check tables is https://github.com/pingcap/tiflow/blob/f710abba6417971370e0d0711d9cd8ca3fa784c4/dm/tests/dmctl_basic/data/db1.prepare.sql#L26-L27

the failed check items are 
1. sharding table columns are different
2. not exist PK/UK

dm/pkg/checker/table_structure.go
```diff
@@ -296,168 +343,140 @@ func (c *ShardingTablesChecker) Check(ctx context.Context) *Result {
...
 	}
-
-	return true
 }
 
-func (c *ShardingTablesChecker) findAutoIncrementKey(stmt *ast.CreateTableStmt, info *model.TableInfo) map[string]bool {
-	autoIncrementKeys := make(map[string]bool)
-	autoIncrementCols := make(map[string]bool)
-
+func (c *ShardingTablesChecker) hasAutoIncrementKey(stmt *ast.CreateTableStmt) bool {
```
okJiang, line 471: the previous code based on `mapping`(deprecated). If the task doesn't use `mapping`, the check will be failed. Users always disable it now.

So, we change it to report a warning if user set upstream auto-increment-key in sharding mode. And let user resolve it by themselves.

dm/pkg/checker/utils.go
okJiang: `markCheckError` may be called repeatedly. So `StateFailure` can be covered by `StartWarn`

dm/checker/check_test.go
GMHDBJD: But why remove them? Seems they are different test(UT and IT)

dm/pkg/checker/table_structure.go
```diff
@@ -79,111 +106,119 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
...
 	}
+	close(c.optCh)
+	c.wg.Wait()
 
-	for name, opts := range options {
-		if len(opts) == 0 {
-			continue
-		}
-		tableMsg := "table " + name + " "
+	log.L().Logger.Info("check table structure over", zap.String("spend time", time.Since(startTime).String()))
```
GMHDBJD, line 135: how about zap.Duration

dm/pkg/checker/table_structure.go
```diff
@@ -296,168 +343,140 @@ func (c *ShardingTablesChecker) Check(ctx context.Context) *Result {
...
-					r.Errors = append(r.Errors, checkErr)
-					r.Extra = fmt.Sprintf("error on sharding %s", c.name)
-					r.Instruction = "please set same table structure for sharding tables"
-					return r
-				}
-			}
-		}
+	dispatchTableItem(checkCtx, c.tableMap, c.inCh)
+	if err := eg.Wait(); err != nil {
+		c.reMu.Lock()
```
GMHDBJD, line 396: Why sometimes lock but sometimes not? e.g. L384

dm/pkg/checker/table_structure.go
```diff
@@ -296,168 +343,140 @@ func (c *ShardingTablesChecker) Check(ctx context.Context) *Result {
...
+		markCheckError(r, err)
+		c.reMu.Unlock()
 	}
 
+	log.L().Logger.Info("check sharding table structure over", zap.String("spend time", time.Since(startTime).String()))
```
GMHDBJD, line 401: ditto

dm/pkg/checker/table_structure_test.go
```diff
@@ -21,31 +21,21 @@ import (
...
 	ctx := context.Background()
 
-	printJSON := func(r *Result) {
-		rawResult, _ := json.MarshalIndent(r, "", "\t")
-		fmt.Println("\n" + string(rawResult))
-	}
+	// printJSON := func(r *Result) {
+	// 	rawResult, _ := json.MarshalIndent(r, "", "\t")
+	// 	fmt.Println("\n" + string(rawResult))
+	// }
```
GMHDBJD, line 35: remove them?

dm/pkg/checker/table_structure.go
okJiang: only one thread uses `r` before `eg.Go()`

GMHDBJD:
> How about add some integration test.

rest LGTM

dm/pkg/checker/table_structure.go
GMHDBJD: After eg.Wait(), all checkShardingTable exit.

dm/pkg/checker/utils.go
Ehco1996: could you add a comment aroud this code?

dm/pkg/checker/table_structure.go
Ehco1996: how about add lock inner func `markCheckError`? i mean add a mutex into `Result` struct, beacuse `reMu` is just used to lock `result`

dm/pkg/checker/table_structure_test.go
```diff
@@ -125,61 +116,103 @@ func (t *testCheckSuite) TestTablesChecker(c *tc.C) {
...
 
+	checker = NewTablesChecker(
+		map[string]*sql.DB{"test-source": db},
+		map[string][]*filter.Table{"test-source": {
+			{Schema: "test-db", Name: "test-table-1"},
+		}},
+		1)
 	result = checker.Check(ctx)
-
+	printJSON(result)
```
Ehco1996, line 163: there is no printJSON now

dm/checker/check_test.go
okJiang: Indeed. I will revert them later

dm/pkg/checker/table_structure.go
okJiang: GMH means there is also one thread after eg.Wait(), so we needn't lock() them after eg.Wait(). @Ehco1996

Ehco1996: yes, but this relay on current logic to make sure no data-race hanppened, i am worried that something might go wrong here when later refactor.

there are many thread trying to wirte result object before `eg.Wait()`, so we still need a lock . if you move lock/unlock operations into `markCheckError` there is no querstion about `Why sometimes lock but sometimes not? e.g. L384` any more

okJiang:
all comment fixed, ptal again, thx @Ehco1996 @GMHDBJD

Ehco1996 (approved):
LGTM

GMHDBJD (approved):
LGTM

dm/pkg/checker/table_structure.go
```diff
@@ -79,111 +106,119 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
...
-				statements[tableName] = statement
-			}
-		}
+	dispatchTableItem(checkCtx, c.tableMap, c.inCh)
+	c.wg.Add(1)
+	go c.handleOpts(ctx, r)
+	if err := eg.Wait(); err != nil {
+		c.reMu.Lock()
+		markCheckError(r, err)
+		c.reMu.Unlock()
```
okJiang, line 130: here is different from L396. Here can cause data race with `handleOpts()`.

okJiang: If there is no other doubt, please give me a /merge again. @Ehco1996

PR #3985, opened 2021-12-20
worker(dm): worker has higher priority to config relay dir

lance6716:
Support different worker has different path to store relay log

as title

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

Related changes

### Release note 

```release-note
DM-worker can specify a path to store its relay log
```

### DIFF
--- a/dm/dm/worker/config.go
+++ b/dm/dm/worker/config.go
@@ -99,6 +99,8 @@ type Config struct {
 	KeepAliveTTL      int64 `toml:"keepalive-ttl" json:"keepalive-ttl"`
 	RelayKeepAliveTTL int64 `toml:"relay-keepalive-ttl" json:"relay-keepalive-ttl"`
 
+	RelayDir string `toml:"relay-dir" json:"relay-dir"`
+
 	// tls config
 	config.Security
 
--- a/dm/dm/worker/source_worker.go
+++ b/dm/dm/worker/source_worker.go
@@ -16,6 +16,7 @@ package worker
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"sync"
 	"time"
 
@@ -77,6 +78,7 @@ type SourceWorker struct {
 	relayWg      sync.WaitGroup
 	relayHolder  RelayHolder
 	relayPurger  relay.Purger
+	relayDir     string
 
 	startedRelayBySourceCfg bool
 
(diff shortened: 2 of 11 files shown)


dm/dm/worker/config.go
```diff
@@ -99,6 +99,8 @@ type Config struct {
 	KeepAliveTTL      int64 `toml:"keepalive-ttl" json:"keepalive-ttl"`
 	RelayKeepAliveTTL int64 `toml:"relay-keepalive-ttl" json:"relay-keepalive-ttl"`
 
+	RelayDir string `toml:"relay-dir" json:"relay-dir"`
```
Ehco1996, line 102: seems this change need PM's review @sunzhaoyang

Ehco1996: btw why we need this extra config?

dm/tests/print_status/run.sh
```diff
@@ -19,9 +19,12 @@ function run() {
 		"github.com/pingcap/tiflow/dm/syncer/ProcessBinlogSlowDown=sleep(4)")
 	export GO_FAILPOINTS="$(join_string \; ${inject_points[@]})"
 
+	cp $cur/conf/dm-worker1.toml $WORK_DIR/dm-worker1.toml
+	sed -i "s%placeholder%$WORK_DIR/relay_by_worker%g" $WORK_DIR/dm-worker1.toml
```
Ehco1996, line 23: maybe add this test to `relay it test`, add this in `print_status` is little not intuitive

dm/dm/worker/source_worker.go
```diff
@@ -334,6 +342,13 @@ func (w *SourceWorker) EnableRelay(startBySourceCfg bool) (err error) {
 	}
 
 	// 2. initial relay holder, the cfg's password need decrypt
+	// worker's relay-dir has higher priority than source's relay-dir
+	if w.relayDir != "" {
+		workerRelayDir := filepath.Join(w.relayDir, w.name)
```
lichunzhu, line 347: Why do we join a worker name here 🤔

lance6716: to avoid user configurated multiple worker to use same path, in some scenario there will be multiple worker writing relay log

dm/dm/worker/config.go
lance6716: the storage path of relay log is more related to the property of DM-worker, such as disk usage

lichunzhu (approved):
LGTM

lichunzhu:
Remember to update docs after this PR is merged

dm/dm/worker/config.go
sunzhaoyang: LGTM

Issue #3774, opened 2021-10-15
Support migrating GBK charset upstream as-is


lance6716:
### Is your feature request related to a problem? Please describe

related to https://github.com/pingcap/tidb/issues/26812

### Describe the feature you'd like

Now TiDB suport GBK so DM should keep GBK charset and collation when migrating.

PR #3990, opened 2021-12-20
syncer(dm): support GBK replication

lance6716:
close https://github.com/pingcap/tiflow/issues/3774

- use `_binary'x'` to insert GBK data from binlog, and add simple test.
- For DDL, decode the query when it's GBK

Also move some function logic together

### Check List 

Tests 

 - Integration test

Code changes

Side effects

Related changes

### Release note 

```release-note
Support replicate GBK tables as-is
```


lance6716:
waiting nightly TiDB build and https://github.com/pingcap/tidb-tools/issues/572

dm/syncer/syncer.go
```diff
@@ -2333,6 +2333,19 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
 		return nil
 	}
 
+	codec, err := event.GetCharsetCodecByStatusVars(ev.StatusVars)
+	if err != nil {
+		s.tctx.L().Error("get charset codec failed, will treat query as utf8", zap.Error(err))
+	}
+	if codec != nil {
```
lichunzhu, line 2340: weakly suggestion
```suggestion
	} else codec != nil {
```

dm/tests/_utils/test_prepare
```diff
@@ -219,7 +219,9 @@ function run_sql_tidb() {
 function run_sql_tidb_with_retry() {
     rc=0
     for ((k=1; k<11; k++)); do
+        set +e
```
lichunzhu, line 222: Why? Leave a comment here?

dm/tests/gbk/data/db1.increment.sql
```diff
@@ -0,0 +1,10 @@
+use `gbk`;
```
lichunzhu, line 1: Do we need a test for a single gbk charset column?
```sql
create table t20 (id int, name varchar(20) character set gbk, primary key(`id`)) character set utf8;
```

lichunzhu (approved):
Rest LGTM

dm/pkg/binlog/event/util.go
```diff
@@ -376,6 +379,35 @@ func GetServerCollationByStatusVars(statusVars []byte, idAndCollationMap map[int
 	return idAndCollationMap[int(v)], err
 }
 
+// GetCharsetCodecByStatusVars returns an encoding.Encoding to encode and decode original query if needed.
+func GetCharsetCodecByStatusVars(statusVars []byte) (encoding.Encoding, error) {
```
Ehco1996, line 383: this func called `GetCharsetCodecByStatusVars` but only return chartset code when chartset is gbk , this looks litte strange, maybe change a name?

Ehco1996 (commented):
rest LGTM

dm/pkg/binlog/event/util.go
lance6716: the comment says "returns ... if needed", we can return more codec if user wants

Ehco1996: ok i see

Ehco1996 (approved):

PR #4010, opened 2021-12-22
processor(cdc): do not write to workload keys if new scheduler is enabled

liuzix:
- The `workload` Etcd key is not used and has never been read from.
- These keys have sizes `O(n)` (`n` being the number of tables), which risks writing too much data to Etcd.

- Prevent writing to the `workload` keys if the new scheduler is enabled.

### Check List 

Tests 

Code changes

 - Has persistent data change

### Release note 

```release-note
None
```


amyangfei (approved):

cdc/processor/processor.go
```diff
@@ -347,7 +347,10 @@ func (p *processor) tick(ctx cdcContext.Context, state *orchestrator.ChangefeedR
 
 	p.handlePosition(oracle.GetPhysical(pdTime))
 	p.pushResolvedTs2Table()
-	p.handleWorkload()
+
+	if !p.newSchedulerEnabled {
+		p.handleWorkload()
```
overvenus, line 352: Please add some comments.

overvenus (approved):

PR #4034, opened 2021-12-23
dm: adjust lightning loader config 

### LINKED ISSUE #3519: better lightning-loader configuration

currently DM's config for lightning loader is not well-designed

Regarding the new backend configuration item, I hope it could be optimized.

```
tidb:
  backend: "tidb"
```

1. location
Since this configuration is related to target, I think it can be placed in

```
target-database:
  host: "127.0.0.1"
  port: 4400
  user: "root"
  password: ""
  security:
    ssl-ca: "dir-placeholer/task-ca.pem"
    ssl-cert: "dir-placeholer/dm.pem"
    ssl-key: "dir-placeholer/dm.key"
  import-mode: "${value}"  # If multiple downstreams (N->N) are to be added in the future, this approach can also continue to be used
```

2. the value of the configuration item is not easy to understand, for historical reasons Lightning called tidb and local two modes. We can improve it in dm so that new users can understand it better

- `tidb/local` could be replaced with `logical/physical` or `sql/sst`？,looks better to understand

_No response_

_No response_

glorv:

Issue Number: close #3519

Adjust lightning loader by:
- remove the config field `tidb`
- Add new fields `import-mode` and `on-duplicate` in `LoaderConfig`

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### Release note 

```release-note
None
```


dm/dm/worker/subtask.go
```diff
@@ -83,6 +75,14 @@ func createRealUnits(cfg *config.SubTaskConfig, etcdClient *clientv3.Client, wor
 	return us
 }
 
+func newLoadUnit(cfg *config.SubTaskConfig, etcdClient *clientv3.Client, workerName string) unit.Unit {
+	// tidb-lightning doesn't support column mapping currently
+	if cfg.ImportMode == config.LoadModeLoader || cfg.OnDuplicate == config.OnDuplicateError || len(cfg.ColumnMappingRules) > 0 {
```
Ehco1996, line 80: maybe this check should added to `func (m *LoaderConfig) adjust() ` and return an error if user set import mode is `sql` with `ColumnMappingRules`

dm/loader/lightning.go
```diff
@@ -95,12 +97,13 @@ func makeGlobalConfig(cfg *config.SubTaskConfig) *lcfg.GlobalConfig {
 	lightningCfg.TiDB.Psw = cfg.To.Password
 	lightningCfg.TiDB.User = cfg.To.User
 	lightningCfg.TiDB.Port = cfg.To.Port
-	lightningCfg.TiDB.StatusPort = cfg.TiDB.StatusPort
-	lightningCfg.TiDB.PdAddr = cfg.TiDB.PdAddr
-	lightningCfg.TiDB.LogLevel = cfg.LogLevel
-	lightningCfg.TikvImporter.Backend = cfg.TiDB.Backend
+	// tidb backend currently don't need following configs
+	// lightningCfg.TiDB.StatusPort = cfg.TiDB.StatusPort
```
Ehco1996, line 101: how about just remove this dead code

dm/loader/lightning.go
```diff
@@ -197,14 +216,20 @@ func (l *LightningLoader) restore(ctx context.Context) error {
...
+		cfg.Routes = l.cfg.RouteRules
 		if l.cfg.To.Session != nil {
 			for k, v := range l.cfg.To.Session {
 				cfg.TiDB.Vars[k] = v
 			}
 		}
-		cfg.TiDB.StrSQLMode = l.cfg.LoaderConfig.SQLMode
-		cfg.TiDB.Vars = map[string]string{"time_zone": l.timeZone}
+		cfg.TiDB.StrSQLMode = l.sqlMode
+		cfg.TiDB.Vars = map[string]string{
```
Ehco1996, line 228: how about w init `cfg.TiDB.Vars`  with tidb_txn_mode in `makeGlobalConfig` ?

dm/loader/lightning.go
```diff
@@ -159,14 +177,15 @@ func (l *LightningLoader) runLightning(ctx context.Context, cfg *lcfg.Config) er
 	if err := l.checkPointList.UpdateStatus(ctx, lightningStatusRunning); err != nil {
 		return err
 	}
+	log2.SetAppLogger(l.logger.Logger)
```
Ehco1996, line 180: dose this cause datarace when multi lighning running at same time? seems it  both tring to replace global logger

and i found we  alreay call `lightning.log.SetAppLogger` when dm-worker start

dm/tests/print_status/run.sh
```diff
@@ -64,21 +64,14 @@ function check_print_status() {
...
-		if [ $count -eq $status_count ]; then
-			finished_bytes=$(echo "$line" | awk '{print $2}' | tr -d "[finished_bytes=" | tr -d "]")
-			total_bytes=$(echo "$line" | awk '{print $3}' | tr -d "[total_file_count" | tr -d "]")
-			[[ "$finished_bytes" -eq "$total_bytes" ]]
-		fi
-	done
+	echo status_count, $status_count
+	[ $status_count -eq 1 ]
+	finished_bytes=$(cat $status_file | awk '{print $3}' | tr -d "[finished_bytes=" | tr -d "]")
+	total_bytes=$(cat $status_file | awk '{print $4}' | tr -d "[total_file_count" | tr -d "]")
```
lance6716, line 73: maybe this is the constant "59637" as in line 67

lance6716 (commented):
rest lgtm

dm/loader/lightning.go
glorv: GloalConfig doesn't contains `Vars` field.

dm/dm/worker/subtask.go
glorv: In most case, we expect our user just use the "default" value and don't need to pay much attention for whether we user loader or lightning to execute SQL. Thus, lightning can't support some features is our internal restriction and we should auto adjust it instead of report an error to our user especially when user just set the default config.

dm/loader/lightning.go
glorv: done

dm/loader/lightning.go
glorv: 👍

dm/tests/print_status/run.sh
glorv: yes

lance6716 (approved):

Ehco1996 (approved):
LGTM

GMHDBJD:
chaos test failed after this PR merged, please take a look. Seems data lost at load stage.

Issue #3698, opened 2021-12-02
dm: Duplicate entry xx for key 'PRIMARY' on test-infra case tolerance 002


D3Hunter:
[dm-dm-worker-0.log](https://github.com/pingcap/ticdc/files/7639647/dm-dm-worker-0.log)
### What did you do?

test-infra case tolerance 002 (unstable, may not reproduce)
- task `test` runs on worker-1
- after 1 minute, kill worker-1
- `test` scheduled to worker-0
- error after a while

### What did you expect to see?

sync-diff success after sysbench finish

### What did you see instead?

`test` paused due to error in title.

### Versions of the cluster

DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):
```console
Release Version: v5.2.0-master
Git Commit Hash: 77029891ff124a5a0e07d5b4dbd4c4f40c6c6268
Git Branch: master
UTC Build Time: 2021-12-01 14:43:41
Go Version: go version go1.16.4 linux/amd64
(paste DM version here, and you must ensure versions of dmctl, DM-worker and DM-master are same)
```
Upstream MySQL/MariaDB server version:
```console
(paste upstream MySQL/MariaDB server version here)
```
Downstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):
```console
(paste TiDB cluster version here)
```
How did you deploy DM: tiup or manually?
```console
(leave TiUP or manually here)
```
Other interesting information (system version, hardware config, etc):
```console
>
>
```

### current status of DM cluster (execute `query-status <task-name>` in dmctl)

```console
{
    "result": true,
    "msg": "",
    "sources": [
        {
            "result": true,
            "msg": "",
            "sourceStatus": {
                "source": "source-0",
                "worker": "dm-dm-worker-0",
                "result": null,
                "relayStatus": null
            },
            "subTaskStatus": [
                {
                    "name": "test",
                    "stage": "Paused",
                    "unit": "Sync",
[dm-dm-worker-0.log](https://github.com/pingcap/ticdc/files/7639643/dm-dm-worker-0.log)

                    "result": {
                        "isCanceled": false,
                        "errors": [
                            {
                                "ErrCode": 10006,
                                "ErrClass": "database",
                                "ErrScope": "not-set",
                                "ErrLevel": "high",
[...]

PR #4035, opened 2021-12-23
unit(dm): add Kill func for unit

### LINKED ISSUE #3737: multiple worker write for same upstream
### What did you do?

chaos test: pod kill dm

no error log in dm-worker.log

### What did you see instead?

write conflict: insert xxxx

### Versions of the cluster

```console
master
```

```console
```

Ehco1996:

Issue Number: close #3737

### Root case of #3737

* when dm-master node offline, dm-worker will lost the keepalive with master
* dm-worker will use [stopWorker](https://github.com/pingcap/tiflow/blob/abefe57c154efb28f91cf1a15702988e1a7249e7/dm/dm/worker/join.go#L111) to stop running tasks
* if current running unit is `syncer`, syncer will try continue syncing until current trasaction end https://github.com/pingcap/tiflow/blob/abefe57c154efb28f91cf1a15702988e1a7249e7/dm/syncer/syncer.go#L1411
* at the same time master online again or another master node beacme leader will try scheduler task to another running worker
* so there are two workers syncning from same source ,and we may meet duplicate key error

* kill running subtasks when worker lost the keep-alive with dm-master (both for master offline or worker offline )
* add Kill interface func to unit
* impleament kill in syncer unit 
* change `worker.Close` to `worker.Stop(graceful bool)`,  when ungraceful stop, worker will call `killAllSubTasks` to force stop and this will cancel syncer.Run immediately

* **only** when worker lost keepalive unit should call `Kill`

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`fix a bug that multiple worker write for same upstream`.
```


dm/dm/worker/subtask.go
```diff
@@ -379,8 +385,8 @@ func (st *SubTask) closeUnits() {
 
 	for i := cui; i < len(st.units); i++ {
 		u := st.units[i]
-		st.l.Info("closing unit process", zap.Stringer("unit", cu.Type()))
-		u.Close()
+		st.l.Info("closing unit process", zap.Stringer("unit", cu.Type()), zap.Bool("graceful", graceful))
+		u.Close(graceful)
```
Ehco1996, line 389: because close func is idempotent, no harm to call this twice

dm/dm/worker/subtask.go
```diff
@@ -358,7 +358,13 @@ func (st *SubTask) PrevUnit() unit.Unit {
 }
 
 // closeUnits closes all un-closed units (current unit and all the subsequent units).
-func (st *SubTask) closeUnits() {
+func (st *SubTask) closeUnits(graceful bool) {
+	// when not graceful, we want to syncer to exit immediately, so we call u.Close(false) before call cancel
+	// Note that we only implement un graceful close for sync unit
```
Ehco1996, line 363: fix comment

dm/syncer/syncer.go
```diff
@@ -972,7 +973,7 @@ func (s *Syncer) addJob(job *job) error {
 		s.isTransactionEnd = false
 		failpoint.Inject("checkCheckpointInMiddleOfTransaction", func() {
 			s.tctx.L().Info("receive dml job", zap.Any("dml job", job))
-			time.Sleep(100 * time.Millisecond)
+			time.Sleep(500 * time.Millisecond)
```
Ehco1996, line 976: this is because my local computer run this case very fast, sleep 0.1 seconds will cause case failed sometimes

dm/syncer/syncer.go
```diff
@@ -1385,10 +1386,39 @@ func (s *Syncer) syncDML() {
 	}
 }
 
+func (s *Syncer) waitTransactionEndBeforeExit(ctx context.Context) {
```
Ehco1996, line 1389: refactor from `s.Run` L1400~L1426

dm/syncer/syncer.go
```diff
@@ -3160,21 +3164,29 @@ func (s *Syncer) isClosed() bool {
...
+		s.tctx.L().Warn("syncer is closing without graceful")
+	}
 	s.stopSync()
 	s.closeDBs()
 
 	s.checkpoint.Close()
 
-	if err := s.schemaTracker.Close(); err != nil {
-		s.tctx.L().Error("fail to close schema tracker", log.ShortError(err))
+	if graceful {
```
Ehco1996, line 3184: this is may cause confuse , but  i don't have better way. need reviewer suggestion

lance6716 (commented):
weak suggestion: `Close` having a parameter is strange to me, can we use `Kill` or something?

`EtcdServer` has `Stop` and `HardStop`

Ehco1996:
> 

ok i will add new interface func to unit called `HardStop`

D3Hunter:
please paste the root cause of this bug, else it's hard to reason whether this PR fix it.

Ehco1996:
> please paste the root cause of this bug, else it's hard to reason whether this PR fix it.

root cause is updated in PR description, and mean changes to reproduce this bug can found in this [commit](https://github.com/Ehco1996/ticdc/commit/67d06e2c9b7f15b497b08f9ca1d44f813b573a56), ptal

D3Hunter:
we may still met `write conflict` even if connection between worker1 and downstream is slow enough, as `DMLWorker` will wait for `30 * time.Second` before notice of `kill`

also please take a look at #3698

dm/syncer/syncer.go
```diff
@@ -3173,8 +3176,12 @@ func (s *Syncer) Close() {
 
 	s.checkpoint.Close()
 
-	if err := s.schemaTracker.Close(); err != nil {
-		s.tctx.L().Error("fail to close schema tracker", log.ShortError(err))
+	// when close sycner without graceful, Close will be called before s.Process exit.
+	// if we close schematracker worker will pending on `s.checkpoint.Rollback(s.schemaTracker)` forever
+	if closeTracker {
+		if err := s.schemaTracker.Close(); err != nil {
```
D3Hunter, line 3182: this may cause memory leak?

Ehco1996: sadly yes, but i can't find a better way to handle this, maybe don't rollback checkpoint when into `kill` branch?

dm/syncer/dml_worker.go
```diff
@@ -236,9 +237,9 @@ func (w *DMLWorker) executeBatchJobs(queueID int, jobs []*job) {
 		time.Sleep(time.Duration(t) * time.Second)
 	})
 	// use background context to execute sqls as much as possible
-	ctx, cancel := w.tctx.WithTimeout(maxDMLExecutionDuration)
+	ctx, cancel := context.WithTimeout(w.runCtx, maxDMLExecutionDuration)
```
glorv, line 240: Please all adjust the comment

dm/syncer/syncer.go
```diff
@@ -3349,26 +3345,34 @@ func (s *Syncer) Close() {
 	s.stopSync()
 	s.closeDBs()
 
+	// try to rollback checkpoints, if they already flushed, no effect, this operation should call before close schemaTracker
```
glorv, line 3348: pause task will not call `syncer.Close`, but we need to do this rollback in this scenario. /cc @lance6716

dm/tests/checkpoint_transaction/run.sh
```diff
@@ -28,6 +58,29 @@ function run() {
...
 	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
 
+	# test ungraceful stop, worker will not wait transaction finish
+	run_sql_file $cur/data/db1.increment1.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
+	sleep 2
+	# kill dm-master 1 to make worker lost keepalive while a transaction is not finished
+	echo "kill dm-master1"
+	kill_dm_master
+	check_master_port_offline 1
+	sleep 1 # wiat woker lost ha keepalive-ttl is 1 second
```
lance6716, line 68:
```suggestion
	sleep 1 # wait woker lost ha keepalive-ttl is 1 second
```

please check typo in this file

dm/tests/checkpoint_transaction/run.sh
```diff
@@ -28,6 +58,29 @@ function run() {
...
+
+	# check dm-worker2 will exit quickly without waiting for the transaction to finish
```
lance6716, line 70: where is worker2?

lance6716 (changes requested):

Ehco1996:
/hold wait more ctx refine
/run-dm-integration-tests

just trigger a it test

dm/tests/_utils/ha_cases_lib.sh
```diff
@@ -145,6 +145,7 @@ function start_multi_tasks_cluster() {
 }
 
 function cleanup() {
+	cleanup_process $*
```
Ehco1996, line 148: we should cleanup data after cleanup_process otherwise we can't not see the log

dm/dm/worker/server.go
```diff
@@ -441,26 +441,25 @@ func (s *Server) observeSourceBound(ctx context.Context, rev int64) error {
 }
 
 func (s *Server) doClose() {
-	s.cancel()
-	// close server in advance, stop receiving source bound and relay bound
```
Ehco1996, line 445: reason is here: // we should stop current sync first, other wise master may scheduler task on new worker while we are closing

i thinks if receiving source bound and relay bound when we are closing don't make things worse, if this worker is offline, those bound will be transfer to other free worker

dm/dm/worker/subtask.go
```diff
@@ -381,6 +381,14 @@ func (st *SubTask) closeUnits() {
 		u := st.units[i]
 		st.l.Info("closing unit process", zap.Stringer("unit", cu.Type()))
 		u.Close()
+		st.l.Info("closing unit done", zap.Stringer("unit", cu.Type()))
+	}
+}
+
+func (st *SubTask) killCurrentUnit() {
+	if st.CurrUnit() != nil {
+		st.l.Info("kill syncer unit", zap.String("task", st.cfg.Name), zap.Stringer("unit", st.CurrUnit().Type()))
```
Ehco1996, line 390: fix log

dm/tests/new_relay/run.sh
```diff
@@ -26,7 +26,9 @@ function test_cant_dail_upstream() {
 		"start-relay -s $SOURCE_ID1 worker1" \
 		"\"result\": true" 2
 
-	kill_dm_worker
```
Ehco1996, line 29: kill_dm_worker and wait_pattern_exit will waiting process exit forever, this is not good at CI, so use `check_port_offline $WORKER1_PORT 20` instead

dm/syncer/syncer.go
```diff
@@ -431,13 +432,6 @@ func (s *Syncer) Init(ctx context.Context) (err error) {
 			}
 		}
 	}
-	s.checkpointFlushWorker = &checkpointFlushWorker{
```
Ehco1996, line 434: all wokrer's lifecycle  (dml/ddl/checkpointWorker) was moved into `s.Run` and use `s.runCtx` `s.runTctx` `s.runwg` to control

s.Run will make sure all background jobs which created by itself will be closed before it return

dm/syncer/syncer.go
```diff
@@ -280,11 +289,9 @@ func NewSyncer(cfg *config.SubTaskConfig, etcdClient *clientv3.Client, relay rel
 }
 
 func (s *Syncer) newJobChans() {
-	s.closeJobChans()
 	chanSize := calculateChanSize(s.cfg.QueueSize, s.cfg.WorkerCount, s.cfg.Compact)
 	s.dmlJobCh = make(chan *job, chanSize)
 	s.ddlJobCh = make(chan *job, s.cfg.QueueSize)
-	s.checkpointFlushWorker.input = make(chan *checkpointFlushTask, 16)
```
Ehco1996, line 287: the lifecycle of checkpointFlushWorker is moved into `s.Run`

dm/syncer/syncer.go
```diff
@@ -136,7 +149,6 @@ type Syncer struct {
 	binlogType         BinlogType
 	streamerController *StreamerController
 
-	wg    sync.WaitGroup // counts goroutines
```
Ehco1996, line 139: rename to s.runWg

dm/syncer/syncer.go
```diff
@@ -189,8 +201,6 @@ type Syncer struct {
 	filteredUpdate atomic.Int64
 	filteredDelete atomic.Int64
 
-	done chan struct{}
```
Ehco1996, line 192: sync.Run will wait `runWg` before exit now, so no need to use this channel

dm/tests/checkpoint_transaction/run.sh
```diff
@@ -6,6 +6,36 @@ cur=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
...
+		fi
+		num=$(grep "received ungraceful exit ctx, exit now" $WORK_DIR/worker1/log/dm-worker.log | wc -l)
+		if [ $num -lt 1 ]; then
+			continue
+		fi
+		all_matched=true
+		if $all_matched; then
+			echo "check_worker_ungraceful_stop_with_retry after retry: $k"
+			break
+		fi
```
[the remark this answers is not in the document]
Ehco1996: i need this test faild if not `all_matched` so can't use `exit 0` here?

dm/syncer/syncer.go
```diff
@@ -1497,69 +1541,53 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
 	}
 
+	s.runWg.Add(1)
+	// start syncDML worker first because if there is no dml worker checkpointFlushWorker will blocked forever
+	go s.syncDML()
+	s.runWg.Add(1)
+	// start flush checkpoints worker. this worker mast start before s.flushCheckPoints()
+	go func() {
+		defer s.runWg.Done()
+		// also need to use a ctx different s.runCtx, checkpointFlushWorker worker will closed in the first defer
```
lance6716, line 1558:
```suggestion
		// also need to use a ctx different s.runCtx, checkpointFlushWorker worker will be closed in the first defer
```

many passive voice errors

dm/syncer/syncer.go
```diff
@@ -1497,69 +1541,53 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
-			tctx.L().Warn("error happened when load table structure from dump files", zap.Error(err))
+			s.tctx.L().Warn("error happened when load table structure from dump files", zap.Error(err))
 			cleanDumpFile = false
 		}
 	} else {
 		cleanDumpFile = false
 	}
 
+	s.runWg.Add(1)
+	// start syncDML worker first because if there is no dml worker checkpointFlushWorker will blocked forever
```
lance6716, line 1552: why we don't this this order in original code?

Ehco1996: this  case was founded when i hack code ,i want to see all thread lifetime in s.Run so i just commnet the `go s.syncDM` but if `syncDML` not start, `s.Run` will pending on `checkpointFlushWorker.run` as a result dm-woker can not exit any more

this case can not hanppen in real world,  but i perfer to start `syncDML` thread before call `flushCheckPoints` because there is actually an order of priority between

dm/syncer/dml_worker.go
```diff
@@ -36,7 +36,7 @@ type DMLWorker struct {
 	chanSize     int
 	multipleRows bool
 	toDBConns    []*dbconn.DBConn
-	tctx         *tcontext.Context
+	runTCtx      *tcontext.Context
```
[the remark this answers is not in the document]
Ehco1996: agree with you, i rename to `syncCtx` instead do you think this name is better? https://github.com/pingcap/tiflow/pull/4035/commits/a73aa0be1dccbcde85300a3c4a6e3a6a71e216a9

dm/syncer/syncer.go
lance6716: since in real scenario this won't happen, I prefer we delete this comment and line 1555, it only causes confusion

dm/dm/worker/server.go
lance6716: not all functions immediately response to `s.cancel()`, for example, some functions will use context of etcdClient which can cost 30s (so maybe we can close s.etcdClient in advance?). scheduler a new worker will add extra time ro wait doClose() return.

Also, please check if there are other cases that will cause doClose() slower after this change.

dm/syncer/syncer.go
```diff
@@ -123,7 +123,17 @@ const (
...
 
-	tctx *tcontext.Context
+	tctx *tcontext.Context // this ctx only used for logger.
+
+	// this ctx derives from a background ctx and was initialized in s.Run, it is used for some background tasks in s.Run
+	// when this ctx cancelled, syncer will shutdown all background running jobs and not wait transaction end
+	runCtx    *tcontext.Context
+	runCancel context.CancelFunc
+	// this ctx only used for syncDML and syncDDL and only cancelled when ungraceful stop
+	syncTCtx   *tcontext.Context
```
lance6716, line 133: I mean TCtx is ugly 😂 

BTW, by your comment for these two context, one is "when this ctx cancelled, syncer will shutdown all background running jobs and not wait transaction end" other is "only cancelled when ungraceful stop", I can't get the difference of them.

dm/syncer/syncer.go
```diff
@@ -1424,48 +1408,115 @@ func (s *Syncer) syncDML() {
 	}
 }
 
-// Run starts running for sync, we should guarantee it can rerun when paused.
-func (s *Syncer) Run(ctx context.Context) (err error) {
-	runCtx, runCancel := context.WithCancel(context.Background())
-	defer runCancel()
-	tctx := s.tctx.WithContext(runCtx)
+func (s *Syncer) waitBeforeRunExit(ctx context.Context) {
```
lance6716, line 1411: can you add an unit test for this function?

dm/syncer/syncer.go
```diff
@@ -1424,48 +1408,115 @@ func (s *Syncer) syncDML() {
...
+		cp:           s.checkpoint,
+		execError:    &s.execError,
+		afterFlushFn: s.afterFlushCheckpoint,
+		addCountFunc: s.addCount,
+	}
+	s.Unlock()
+	defer func() {
+		s.runCancel()
+		s.closeJobChans()
+		s.checkpointFlushWorker.Close()
```
lance6716, line 1503: will this cause a send-on-closed-channel panic?

dm/syncer/syncer.go
```diff
@@ -3355,45 +3363,46 @@ func (s *Syncer) Close() {
...
-	s.wg.Wait() // wait job workers to return
+// Kill kill syncer without graceful.
+func (s *Syncer) Kill() {
+	s.tctx.L().Warn("kill syncer without graceful")
+	s.runCancel()
+	s.syncCancel()
+	s.Close()
+}
 
+// stopStreamAndRollbackCheckpoint stop stream and rollback checkpoint now it used by Close() and Pause().
```
lance6716, line 3390:
```suggestion
// stopStreamAndRollbackCheckpoint stops stream and rollbacks checkpoint, now it is used by Close() and Pause().
```

I think we do not need to rollback checkpoint when closing, and it's not intuitive that group close streamer and checkpoint into one function.

dm/dm/worker/server.go
Ehco1996: > scheduler a new worker will add extra time ro wait doClose() return.

i thinks this is expected beheaviour, worker should cancel keepalive after sync task really stopped?

dm/syncer/syncer.go
Ehco1996: most logic is tested in current ut `TestRun` , for ungraceful stop and waitTxnend test is added in it test `checkpoint_transaction`

dm/syncer/syncer.go
Ehco1996: actully not, beacuse this defer func can only be called after the job producer(main for loop in s.Run) stopped

dm/syncer/syncer.go
Ehco1996: renamd back to `stopSync` now, i just follow the old logic : rollback checkpoint before syncer.Process return in [here](https://github.com/pingcap/tiflow/blob/9d7ae48d3fee0ea02bfa9dd504e06166bc25b45e/dm/syncer/syncer.go#L703)

dm/syncer/syncer.go
Ehco1996: renamed to syncCtx  and updated the comment about these two ctx in https://github.com/pingcap/tiflow/pull/4035/commits/705c228dd22049a3499c4171ff5ddbbf01f12b10, dose this help?

dm/syncer/syncer.go
lance6716: I think we can only rollback schema checkpoint and tracker when `Pause`, not when `Close`. PTAL @GMHDBJD

dm/dm/worker/server.go
lance6716: I can understand the benefit of this change, originally it's fast and correct, now it's slow and correct.

dm/tests/checkpoint_transaction/run.sh
```diff
@@ -6,6 +6,39 @@ cur=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
...
+		num=$(grep "received ungraceful exit ctx, exit now" $WORK_DIR/worker1/log/dm-worker.log | wc -l)
+		if [ $num -lt 1 ]; then
+			continue
+		fi
+		all_matched=true
+		if $all_matched; then
+			echo "check_worker_ungraceful_stop_with_retry success after retry: $k"
+			break
+		fi
```
lance6716, line 32:
```suggestion
		return 0
```

remove `all_matched` and return 1 when retry is exhausted.

dm/tests/checkpoint_transaction/conf/dm-worker1.toml
```diff
@@ -1,2 +1,5 @@
-name = "worker1"
 join = "127.0.0.1:8261"
+keepalive-ttl = 1
+log-level = "info"
```
lance6716, line 3: in fact `log-level` and `worker-addr` are passed from command line arguments in tests

dm/syncer/syncer.go
lance6716: You forget runCancel previously https://github.com/pingcap/tiflow/pull/4035#discussion_r785619324

currently we have no test to cover this exection path I think.

dm/tests/new_relay/run.sh
```diff
@@ -157,11 +89,87 @@ function test_restart_relay_status() {
...
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+
+	# make sure DM-worker doesn't exit
+	sleep 2
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"query-status -s $SOURCE_ID1" \
+		"injected error" 1
+
+	echo "》》》》》》》》》》》》》》》》》test test_cant_dail_upstream passed"
```
lance6716, line 127: please use English punctuation

dm/tests/new_relay/run.sh
```diff
@@ -157,11 +89,87 @@ function test_restart_relay_status() {
...
+}
+
+function test_cant_dail_downstream() {
+	cleanup_process
 	cleanup_data $TEST_NAME
+	export GO_FAILPOINTS=""
```
lance6716, line 133: `export` should be put in the ending of previous test

dm/dm/worker/server.go
Ehco1996: you are right, i  found i can call `s.cancel` with no wait https://github.com/pingcap/tiflow/pull/4035/commits/e881437b628ee2de07d95be1e6bd6d327b224d75

dm/tests/new_relay/run.sh
Ehco1996: maybe no difference, this is looks like what we do like `cleanup_process`

dm/syncer/syncer.go
Ehco1996: add a ut in https://github.com/pingcap/tiflow/pull/4035/commits/a06c5e82cf6d7fc3e19e4daac5bd09c6685b66d5

dm/syncer/syncer.go
Ehco1996: ping @GMHDBJD  please help me clean the logic about the right time to rollback checkpoint

dm/syncer/syncer.go
```diff
@@ -123,7 +123,17 @@ const (
...
-	tctx *tcontext.Context
+	tctx *tcontext.Context // this ctx only used for logger.
+
+	// this ctx derives from a background ctx and was initialized in s.Run, it is used for some background tasks in s.Run
+	// when this ctx cancelled, syncer will shutdown all background running jobs (except the syncDML and syncDDL) and not wait transaction end.
+	runCtx    *tcontext.Context
+	runCancel context.CancelFunc
+	// this ctx only used for syncDML and syncDDL and only cancelled when ungraceful stop.
+	syncCtx    *tcontext.Context
+	syncCancel context.CancelFunc
```
GMHDBJD, line 134: can we pass different ctx for different function, and only keep the topmost ctx in syncer?

dm/syncer/syncer.go
```diff
@@ -1497,69 +1541,52 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
 		if err != nil {
-			tctx.L().Warn("error happened when load table structure from dump files", zap.Error(err))
+			s.tctx.L().Warn("error happened when load table structure from dump files", zap.Error(err))
 			cleanDumpFile = false
 		}
 	} else {
 		cleanDumpFile = false
 	}
 
+	s.runWg.Add(1)
```
GMHDBJD, line 1551: how about s.runWg.Add(5)?

GMHDBJD:
Should we ungraceful exit the pipeline  in syncer? e.g.
https://github.com/pingcap/tiflow/blob/f710abba6417971370e0d0711d9cd8ca3fa784c4/dm/syncer/compactor.go#L73
https://github.com/pingcap/tiflow/blob/f710abba6417971370e0d0711d9cd8ca3fa784c4/dm/syncer/causality.go#L69
Can you add an ungraceful mock test base on [current isolate test](https://github.com/pingcap/tiflow/blob/f710abba6417971370e0d0711d9cd8ca3fa784c4/dm/tests/ha_cases_1/run.sh#L73)?

dm/syncer/syncer.go
GMHDBJD: Yes, we rollback when task paused. both pause-task and ***pause by error***

dm/syncer/syncer_test.go
```diff
@@ -1703,3 +1687,72 @@ func (s *testSyncerSuite) TestExecuteSQLSWithIgnore(c *C) {
 
 	c.Assert(mock.ExpectationsWereMet(), IsNil)
 }
+
+func genDefaultSubTaskConfig4Test() *config.SubTaskConfig {
```
lance6716, line 1691: this is not an unit test covers https://github.com/pingcap/tiflow/pull/4035#discussion_r793311221

What I want is only to test `waitBeforeRunExit`, not whole Syncer. I think you can construct a simple Syncer, call `waitBeforeRunExit(ctx1)`, cancel ctx1, wait `maxPauseOrStopWaitTime`, check if `runCtx` is canceled

dm/syncer/syncer_test.go
```diff
@@ -1703,3 +1687,72 @@ func (s *testSyncerSuite) TestExecuteSQLSWithIgnore(c *C) {
...
+		LoaderConfig:     loaderCfg,
+		UseRelay:         false,
+	}
+	cfg.Experimental.AsyncCheckpointFlush = true
+	cfg.From.Adjust()
+	cfg.To.Adjust()
+	return cfg
+}
+
+func TestWaitBeforeRunExit(t *testing.T) {
```
lance6716, line 1719: this is not an unit test covers https://github.com/pingcap/tiflow/pull/4035#discussion_r793311221

What I want is only to test `waitBeforeRunExit`, not whole Syncer. I think you can construct a simple Syncer, call `waitBeforeRunExit(ctx1)`, cancel ctx1, wait `maxPauseOrStopWaitTime`, check if `runCtx` is canceled

Ehco1996: add this test  in https://github.com/pingcap/tiflow/pull/4035/commits/1e51bec3be4a8460970648afd7c163ee516a77a4

and i thinks we can keep old test(// test s.Run will not exit unit caller cancel ctx or call s.runCancel)

Ehco1996 (commented):
@GMHDBJD 

> Should we ungraceful exit the pipeline in syncer?

added in https://github.com/pingcap/tiflow/pull/4035/commits/fb5f21fd8072154df2fa15d2302cab7af7e44287

> Can you add an ungraceful mock test base on [current isolate test](https://github.com/pingcap/tiflow/blob/f710abba6417971370e0d0711d9cd8ca3fa784c4/dm/tests/ha_cases_1/run.sh#L73)?

already add a ungraceful test in [checkpoint_transaction](https://github.com/Ehco1996/ticdc/blob/fb5f21fd8072154df2fa15d2302cab7af7e44287/dm/tests/checkpoint_transaction/run.sh#L9)

it's hard to add this test in isolate beacause it's need to exec a long txn to upsteram and make syncer stop at half of this txn. we add many failpoints in `checkpoint_transaction` to achieve this

dm/dm/worker/subtask.go
```diff
@@ -381,6 +381,14 @@ func (st *SubTask) closeUnits() {
...
 		st.l.Info("closing unit process", zap.Stringer("unit", cu.Type()))
 		u.Close()
+		st.l.Info("closing unit done", zap.Stringer("unit", cu.Type()))
+	}
+}
+
+func (st *SubTask) killCurrentUnit() {
+	if st.CurrUnit() != nil {
+		st.l.Info("kill unit", zap.String("task", st.cfg.Name), zap.Stringer("unit", st.CurrUnit().Type()))
+		st.CurrUnit().Kill()
```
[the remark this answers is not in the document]
GMHDBJD: I mean add a log with "kill unit done" like L391

dm/syncer/causality.go
```diff
@@ -66,33 +69,37 @@ func causalityWrap(inCh chan *job, syncer *Syncer) chan *job {
...
-
-			// detectConflict before add
-			if c.detectConflict(keys) {
-				c.logger.Debug("meet causality key, will generate a conflict job to flush all sqls", zap.Strings("keys", keys))
-				c.outCh <- newConflictJob(c.workerCount)
-				c.relation.clear()
+	for {
+		select {
+		case <-c.syncCtx.Ctx.Done():
+			break
```
GMHDBJD, line 75: I think use `return` is more intuitive, like L78

dm/syncer/compactor.go
```diff
@@ -70,6 +73,8 @@ func compactorWrap(inCh chan *job, syncer *Syncer) chan *job {
 func (c *compactor) run() {
 	for {
 		select {
+		case <-c.syncCtx.Ctx.Done():
+			break
```
GMHDBJD, line 77: ditto

dm/syncer/dml_worker.go
```diff
@@ -112,32 +112,39 @@ func (w *DMLWorker) run() {
...
-			queueBucket := int(utils.GenHashKey(j.dml.key)) % w.workerCount
-			w.addCountFunc(false, queueBucketMapping[queueBucket], j.tp, 1, j.targetTable)
-			startTime := time.Now()
-			w.logger.Debug("queue for key", zap.Int("queue", queueBucket), zap.String("key", j.dml.key))
-			jobChs[queueBucket] <- j
-			metrics.AddJobDurationHistogram.WithLabelValues(j.tp.String(), w.task, queueBucketMapping[queueBucket], w.source).Observe(time.Since(startTime).Seconds())
+	for {
+		select {
+		case <-w.syncCtx.Ctx.Done():
+			break
```
GMHDBJD, line 118: ditto

GMHDBJD (commented):
> it's hard to add this test in isolate beacause it's need to exec a long txn to upsteram and make syncer stop at half of this txn. we add many failpoints in checkpoint_transaction to achieve this

I think you can add a failpoint for it to mock a big txn. We have some now, maybe some of them can be reused. https://github.com/pingcap/tiflow/blob/master/dm/syncer/dml_worker.go#L227-L239
Just want to have an automatic test to make sure that we really fix the issue for isolate.

GMHDBJD:
rest LGTM

GMHDBJD (approved):
LGTM

dm/dm/worker/server.go
```diff
@@ -441,30 +441,33 @@ func (s *Server) observeSourceBound(ctx context.Context, rev int64) error {
...
-	s.wg.Wait()
```
lance6716, line 446: this WaitGroup makes sure "stop receiving source bound" happens before "getSourceWorker and Close". Your modification breaks it.

lance6716 (commented):
will review soon

dm/syncer/dml_worker.go
```diff
@@ -112,32 +112,39 @@ func (w *DMLWorker) run() {
...
-		default:
-			queueBucket := int(utils.GenHashKey(j.dml.key)) % w.workerCount
-			w.addCountFunc(false, queueBucketMapping[queueBucket], j.tp, 1, j.targetTable)
-			startTime := time.Now()
-			w.logger.Debug("queue for key", zap.Int("queue", queueBucket), zap.String("key", j.dml.key))
-			jobChs[queueBucket] <- j
-			metrics.AddJobDurationHistogram.WithLabelValues(j.tp.String(), w.task, queueBucketMapping[queueBucket], w.source).Observe(time.Since(startTime).Seconds())
+	for {
+		select {
+		case <-w.syncCtx.Ctx.Done():
```
lance6716, line 117: this will discard jobs in the channel
- dml job will be ignored without setting `s.execError` by fatalFunc which will cause checkpoint not aware about it and forwarded wrongly.
- flush job is not passed into out channel, may cause the generator of flush job waiting forever.

that's critical, please add a test for it.

lance6716 (changes requested):

dm/syncer/syncer.go
```diff
@@ -1423,48 +1407,115 @@ func (s *Syncer) syncDML() {
...
+	rtt := time.Since(t1).Seconds()
+	if tsErr == nil {
+		s.tsOffset.Store(time.Now().Unix() - ts - int64(rtt/2))
+	}
+	return tsErr
+}
+
+// Run starts running for sync, we should guarantee it can rerun when paused.
+func (s *Syncer) Run(ctx context.Context) (err error) {
+	s.Lock()
```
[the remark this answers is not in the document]
Ehco1996: sgtm,remove this lock now

dm/syncer/dml_worker.go
Ehco1996: really thx, current handle job logic not allow us do ungraceful close job pipline now, so i revert this change in 
https://github.com/pingcap/tiflow/pull/4035/commits/05ca26babf53aa0de37cd296a27cf31ffd966e25
patl @GMHDBJD

dm/syncer/syncer.go
```diff
@@ -1423,48 +1407,115 @@ func (s *Syncer) syncDML() {
...
+	runCtx, runCancel := context.WithCancel(context.Background())
+	s.runCtx, s.runCancel = tcontext.NewContext(runCtx, s.tctx.L()), runCancel
+	syncCtx, syncCancel := context.WithCancel(context.Background())
+	s.syncCtx, s.syncCancel = tcontext.NewContext(syncCtx, s.tctx.L()), syncCancel
+	s.checkpointFlushWorker = &checkpointFlushWorker{
```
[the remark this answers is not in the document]
Ehco1996: as checkpointFlushWorker is like dmlworker, i moved this into `newJobChan` and  `closeJobChan`
added in https://github.com/pingcap/tiflow/pull/4035/commits/05ca26babf53aa0de37cd296a27cf31ffd966e25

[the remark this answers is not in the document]
lance6716: I prefer syncer.Init. that's not a JobChan

dm/tests/new_relay/run.sh
```diff
@@ -128,8 +56,12 @@ function test_restart_relay_status() {
 		"list-member -n worker3" \
 		"relay" 1
 
-	kill_dm_worker
-	kill_dm_master
+	echo "kill dm-worker1"
```
lance6716, line 59: why change it? originally we kill all workers

dm/tests/checkpoint_transaction/run.sh
```diff
@@ -41,8 +93,13 @@ function run() {
 		"\"stage\": \"Paused\"" 1
 	# check the point is the middle of checkpoint
 	num=$(grep "not receive xid job yet" $WORK_DIR/worker1/log/dm-worker.log | wc -l)
-	[[ $num -gt 0 ]]
-	sed -e '/not receive xid job yet/d' $WORK_DIR/worker1/log/dm-worker.log >$WORK_DIR/worker1/log/dm-worker.log
```
lance6716, line 45: where is this line?

dm/syncer/dml_worker.go
GMHDBJD: LGTM，but I think if ctx close，flush checkpoint should failed immediately

dm/syncer/syncer.go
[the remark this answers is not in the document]
Ehco1996: move to s.reset now https://github.com/pingcap/tiflow/pull/4035/commits/d6484a68354578ec62968c13a63aa385efe8eae4

dm/syncer/dml_worker.go
Ehco1996: seems `w.execError` will make sure FlushPointsExcept will not triggerd when ctx canceled , i am not sure if this is enough

https://github.com/pingcap/tiflow/blob/master/dm/syncer/checkpoint_flush_worker.go#L74

dm/tests/checkpoint_transaction/run.sh
Ehco1996: this line is deleted because if muti thread is write same log file, the log content can become confusing, so stop the dm-worker and delete whole dm-worker dir in https://github.com/pingcap/tiflow/pull/4035/commits/76180db7ea6f8b62a47929c730a7c02425d3e3c8

dm/syncer/syncer.go
lance6716: dml/ddl worker is not a member of syncer, I think the member initialization of syncer should be put in `Syncer.Init`

dm/syncer/syncer.go
```diff
@@ -1451,47 +1441,107 @@ func (s *Syncer) syncDML() {
...
+		s.tsOffset.Store(time.Now().Unix() - ts - int64(rtt/2))
+	}
+	return tsErr
+}
+
+// Run starts running for sync, we should guarantee it can rerun when paused.
+func (s *Syncer) Run(ctx context.Context) (err error) {
+	runCtx, runCancel := context.WithCancel(context.Background())
+	s.runCtx, s.runCancel = tcontext.NewContext(runCtx, s.tctx.L()), runCancel
+	syncCtx, syncCancel := context.WithCancel(context.Background())
```
lance6716, line 1523: where does `syncCancel` be called for normal close routine?

lance6716 (commented):
rest lgtm

dm/syncer/syncer.go
Ehco1996: > dml/ddl worker is not a member of syncer, I think the member initialization of syncer should be put in Syncer.Init

maybe refine this later, at the moment it can't be put in `init` because every time `s.Run` needs to init a brand new worker, if we `resume` a task, s.Init will not be callded
I don't want to introduce any more changes for now, the checkpointWorker and syncDML/sycnDDL locations can be changed together later

> where does syncCancel be called for normal close routine?

we not call cancel for normal close the reason is [here](https://github.com/pingcap/tiflow/blob/28fe713deaba0d0fc8e53b7e038d99e43d0011dd/dm/syncer/dml_worker.go#L246), but we do have `maxDMLExecutionDuration` to ensure there is a timeout

lance6716: Not calling `cancel` will cause something leak in my memory.

Ehco1996: Learned a lot thanks~ https://github.com/pingcap/tiflow/pull/4035/commits/4809d42ff1a186b02cc08c6bf028c5a0fc310413

lance6716 (approved):

dm/tests/new_relay/run.sh
```diff
@@ -157,11 +85,87 @@ function test_restart_relay_status() {
...
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"query-status -s $SOURCE_ID1" \
+		"\"relayCatchUpMaster\": true" 1 \
+		"dial tcp 127.0.0.1:4000: connect: connection refused" 1
+
+	# restart tidb
+	run_tidb_server 4000 $TIDB_PASSWORD
+	sleep 2
```
GMHDBJD, line 161: What does this do?

GMHDBJD, line 161: What does this do?

dm/tests/new_relay/run.sh
```diff
@@ -157,11 +85,87 @@ function test_restart_relay_status() {
...
+	echo "kill dm-worker1"
+	ps aux | grep dm-worker1 | awk '{print $2}' | xargs kill || true
+	check_port_offline $WORKER1_PORT 20
+
+	export GO_FAILPOINTS="github.com/pingcap/tiflow/dm/pkg/conn/failDBPing=return()"
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+
+	# make sure DM-worker doesn't exit
+	sleep 2
```
GMHDBJD, line 118: Sleep cannot make sure anything.

GMHDBJD (commented):
rest LGTM

dm/tests/new_relay/run.sh
Ehco1996: remove sleep in https://github.com/pingcap/tiflow/pull/4035/commits/f7df28a9d4172661c8f98070c79a4d57d3268801

GMHDBJD (approved):
LGTM

Ehco1996:
FAIL: server_test.go:2144: testMaster.TestGRPCLongResponse

/run-leak-test
/run-verify

PR #4036, opened 2021-12-23
sink(ticdc): set max-message-bytes default to 10m

### LINKED ISSUE #4041: TiCDC sink initialization should check topic or broker's max message bytes configuration for all kinds of protocols.
### What did you do?

In v4.0.16, if use non `open-protocol`, such as `canal-json`, `canal`, `avro` to initialize the Kafka sink, changefeed may meet `message too large` error from the Sarama producer.

Furthermore, the TiCDC cluster was upgraded from v4.0.14 to v4.0.16, we should also consider upgrade compatibility, which means old existing changefeeds should boost and run normal.

changefeed works normally, no matter which kind of protocol.

### What did you see instead?

```
[CDC:ErrKafkaAsyncSendMessage]kafka: Failed to produce message to topic cdc-eps_rt_ods-ods_market_flw_lx_app_log_prod: kafka server: Message was too large, server rejected it to avoid allocation error."]
```

### Versions of the cluster

```console
```

```console
v4.0.16
```

3AceShowHand:
close #4041
set Kafka producer's default `max-message-bytes` to 10M, and use the minimum value among sink-uri's `max-message-bytes`, broker's `message.max.bytes`, topic's `max.message.bytes` to initialize the producer.

* set the default value of `max-message-bytes` to 10M
* use the min value of the 3 mentioned above to initialize the producer.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
Set `max-message-bytes` default to 10M, and use the min value with topic and broker to initialize the producer.
```

### DIFF
--- a/cdc/sink/codec/json.go
+++ b/cdc/sink/codec/json.go
@@ -383,13 +383,13 @@ type JSONEventBatchEncoder struct {
 	messageBuf   []*MQMessage
 	curBatchSize int
 	// configs
-	maxMessageSize int
-	maxBatchSize   int
+	maxMessageBytes int
+	maxBatchSize    int
 }
 
 // GetMaxMessageSize is only for unit testing.
 func (d *JSONEventBatchEncoder) GetMaxMessageSize() int {
-	return d.maxMessageSize
+	return d.maxMessageBytes
 }
 
 // GetMaxBatchSize is only for unit testing.
@@ -468,15 +468,15 @@ func (d *JSONEventBatchEncoder) AppendRowChangedEvent(e *model.RowChangedEvent) 
 		// for single message that longer than max-message-size, do not send it.
 		// 16 is the length of `keyLenByte` and `valueLenByte`, 8 is the length of `versionHead`
 		length := len(key) + len(value) + maximumRecordOverhead + 16 + 8
-		if length > d.maxMessageSize {
+		if length > d.maxMessageBytes {
 			log.Warn("Single message too large",
-				zap.Int("max-message-size", d.maxMessageSize), zap.Int("length", length), zap.Any("table", e.Table))
+				zap.Int("max-message-size", d.maxMessageBytes), zap.Int("length", length), zap.Any("table", e.Table))
 			return EncoderNoOperation, cerror.ErrJSONCodecRowTooLarge.GenWithStackByArgs()
 		}
 
 		if len(d.messageBuf) == 0 ||
 			d.curBatchSize >= d.maxBatchSize ||
-			d.messageBuf[len(d.messageBuf)-1].Length()+len(key)+len(value)+16 > d.maxMessageSize {
+			d.messageBuf[len(d.messageBuf)-1].Length()+len(key)+len(value)+16 > d.maxMessageBytes {
 
 			versionHead := make([]byte, 8)
 			binary.BigEndian.PutUint64(versionHead, BatchVersion1)
@@ -495,10 +495,10 @@ func (d *JSONEventBatchEncoder) AppendRowChangedEvent(e *model.RowChangedEvent) 
 		message.Table = &e.Table.Table
 		message.IncRowsCount()
 
-		if message.Length() > d.maxMessageSize {
+		if message.Length() > d.maxMessageBytes {
 			// `len(d.messageBuf) == 1` is implied
 			log.Debug("Event does not fit into max-message-bytes. Adjust relevant configurations to avoid service interruptions.",
-				zap.Int("event-len", message.Length()), zap.Int("max-message-bytes", d.maxMessageSize))
+				zap.Int("event-len", message.Length()), zap.Int("max-message-bytes", d.maxMessageBytes))
 		}
 		d.curBatchSize++
 	}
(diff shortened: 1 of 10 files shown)


cdc/sink/codec/json.go
```diff
@@ -38,7 +38,7 @@ const (
 	// BatchVersion1 represents the version of batch format
 	BatchVersion1 uint64 = 1
 	// DefaultMaxMessageBytes sets the default value for max-message-bytes
-	DefaultMaxMessageBytes int = 1 * 1024 * 1024 // 1M
+	DefaultMaxMessageBytes int = 10 * 1024 * 1024 // 1M
```
amyangfei, line 41:
```suggestion
	DefaultMaxMessageBytes int = 10 * 1024 * 1024 // 10M
```

cdc/sink/producer/kafka/kafka.go
```diff
@@ -354,17 +354,18 @@ func validateMaxMessageBytesAndCreateTopic(admin kafka.ClusterAdminClient, topic
...
 		}
+
 		if topicMaxMessageBytes < config.MaxMessageBytes {
-			return cerror.ErrKafkaInvalidConfig.GenWithStack(
-				"topic already exist, and topic's max.message.bytes(%d) less than max-message-bytes(%d). "+
-					"Please make sure `max-message-bytes` not greater than topic's `max.message.bytes`",
-				topicMaxMessageBytes, config.MaxMessageBytes)
+			log.Warn("topic's `max.message.bytes` less than user set `max-message-bytes`,"+
+				"use topic's `max.message.bytes` to initialize the Kafka producer",
+				zap.Int("max.message.bytes", topicMaxMessageBytes),
```
amyangfei, line 366:
```suggestion
				zap.Int("message.max.bytes", topicMaxMessageBytes),
```

3AceShowHand: for the topic, it's `max.message.bytes`, for the broker, it's `message.max.bytes`

cdc/sink/codec/json.go
```diff
@@ -38,7 +38,7 @@ const (
 	// BatchVersion1 represents the version of batch format
 	BatchVersion1 uint64 = 1
 	// DefaultMaxMessageBytes sets the default value for max-message-bytes
-	DefaultMaxMessageBytes int = 1 * 1024 * 1024 // 1M
+	DefaultMaxMessageBytes int = 10 * 1024 * 1024 // 10M
```
amyangfei, line 41: Can we move this constant value to pkg/config and use one const only.

cdc/sink/producer/kafka/kafka.go
amyangfei: Got it, I found it in Kafka document, it's strange.

cdc/sink/producer/kafka/kafka_test.go
```diff
@@ -222,46 +231,49 @@ func (s *kafkaSuite) TestValidateMaxMessageBytesAndCreateTopic(c *check.C) {
...
 	detail := &sarama.TopicDetail{
 		NumPartitions: 3,
 		// Does not contain max message bytes information.
 		ConfigEntries: make(map[string]*string),
 	}
 	err = adminClient.CreateTopic("test-topic", detail, false)
 	c.Assert(err, check.IsNil)
-	err = validateMaxMessageBytesAndCreateTopic(adminClient, "test-topic", config)
+	err = validateMaxMessageBytesAndCreateTopic(adminClient, "test-topic", config, cfg)
 	c.Assert(err, check.IsNil)
```
amyangfei, line 266: We can check the value of `cfg.Producer.MaxMessageBytes` here.

amyangfei (commented):
rest lgtm

amyangfei (approved):

pkg/config/sink.go
```diff
@@ -21,6 +21,9 @@ import (
 	cerror "github.com/pingcap/tiflow/pkg/errors"
 )
 
+// DefaultMaxMessageBytes sets the default value for max-message-bytes
+var DefaultMaxMessageBytes int = 10 * 1024 * 1024 // 10M
```
hi-rustin, line 25:
```suggestion
const DefaultMaxMessageBytes  = 10 * 1024 * 1024 // 10M
```

cdc/sink/producer/kafka/kafka.go
```diff
@@ -345,7 +345,7 @@ func kafkaClientID(role, captureAddr, changefeedID, configuredClientID string) (
 	return
 }
 
-func validateMaxMessageBytesAndCreateTopic(admin kafka.ClusterAdminClient, topic string, config *Config) error {
+func validateMaxMessageBytesAndCreateTopic(admin kafka.ClusterAdminClient, topic string, config *Config, saramaConfig *sarama.Config) error {
```
hi-rustin, line 348: This func seems strange at the moment. Can we split it up?
1. getMaxMessageBytes is responsible for getting the right value
2. createTopic is used to create the topic

This way we won't have to pass two configurations and modify them at the same time. Now this function has become very complicated. Originally we didn't modify Sarama's configuration in this method. Now it is not only responsible for creating the topic, but also for setting the sarma configuration correctly, but this configuration is not directly related to creating the topic. It does too many things. I prefer to keep them separate.

hi-rustin: If there is a better way to get it not to set it up at the same time that would work too. I can only think of separating it at the moment.
The sarma configuration has been patched once above via kafka's configuration, but we're modifying it at the same time in this function.

It is currently confusing from either the caller's or the test code's point of view.

3AceShowHand: This worthy another PR to do it.

hi-rustin: I think it might be better to do it in this time. It should just be a simple split method will work.

3AceShowHand: I would leave that to another PR, as a future refinement.

3AceShowHand: I still think it would be better to do this after the release, to prevent some potential new problems.

overvenus (approved):

PR #4039, opened 2021-12-23
owner(cdc): add a warning for owner DDL puller being stuck

liuzix:
- To help diagnosing replication stagnation, we need to be notified of owner's DDL puller being stuck.

- Added a warning if owner's DDL puller lags behind for over 30 seconds.

### Check List 

Tests 

 - Unit test

Side effects

 - Increased code complexity

### Release note 

```release-note
None
```


overvenus (approved):
LGTM

3AceShowHand (approved):

Issue #3470, opened 2021-11-16
Log Should be WARNING instead of ERROR while upstream bin_log OFF


XuJianxu:
### What did you do?

Create a source while upstream bin_log OFF

### What did you expect to see?

The log report WARNING, while upstream create success

### What did you see instead?

ERROR log out put.
[...]

PR #4049, opened 2021-12-24
worker(dm): adjust error to warning when bin_log OFF

okJiang:

Issue Number: close [#3470](https://github.com/pingcap/tiflow/issues/3470)

### Check List 

Tests 

 - Unit test
 - Integration test
 
### Release note 

```release-note
`None`.
```


dm/dm/worker/source_worker.go
```diff
@@ -175,7 +175,11 @@ func (w *SourceWorker) Start() {
 				}
 			}
 			if err2 := w.updateSourceStatus(w.ctx); err2 != nil {
-				w.l.Error("failed to update source status", zap.Error(err2))
+				if terror.ErrNoMasterStatus.Equal(err2) {
+					w.l.Warn("This source's bin_log is OFF, so it only supports full_mode.", zap.String("sourceID", w.cfg.SourceID), zap.Error(err2))
```
Ehco1996, line 179: do we need also add this check to pre-check?

okJiang: This may happen before task start and after create source. So we can't pre-check it.

okJiang: But during the pre-check, we also have `binlogChecker`.

Ehco1996 (approved):

lance6716 (approved):

PR #4071, opened 2021-12-26
pkg/actor(ticdc): stable tests by extending timeout

### LINKED ISSUE #3963: TestBroadcast is unstable
### Which jobs are flaking?

https://github.com/pingcap/tiflow/pull/3935 

TestBroadcast

```log
[2021-12-17T10:31:44.215Z] 

[2021-12-17T10:31:44.215Z] === FAIL: pkg/actor TestBroadcast (5.94s)

[2021-12-17T10:31:44.215Z]     system_test.go:140: Timed out

[2021-12-17T10:31:44.215Z] 
```

https://ci.pingcap.net/blue/organizations/jenkins/atom-ut/detail/atom-ut/5678/pipeline

overvenus:

Close #3963 

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/actor/system_test.go
+++ b/pkg/actor/system_test.go
@@ -220,7 +220,7 @@ func TestActorSendReceive(t *testing.T) {
 		t.Fatal("Timed out")
 	}
 
-	wait(t, time.Second, func() {
+	wait(t, 5*time.Second, func() {
 		err := sys.Stop()
 		require.Nil(t, err)
 	})
@@ -257,7 +257,7 @@ func testBroadcast(t *testing.T, actorNum, workerNum int) {
 	case <-time.After(200 * time.Millisecond):
 	}
 
-	wait(t, time.Second, func() {
+	wait(t, 5*time.Second, func() {
 		err := sys.Stop()
 		require.Nil(t, err)
 	})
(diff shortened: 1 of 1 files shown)


pkg/actor/system_test.go
```diff
@@ -220,7 +220,7 @@ func TestActorSendReceive(t *testing.T) {
 		t.Fatal("Timed out")
 	}
 
-	wait(t, time.Second, func() {
```
amyangfei, line 223: Please add some comments, otherwise the magic number makes it hard to maintain in the feature.

amyangfei (approved):

pkg/actor/system_test.go
```diff
@@ -128,15 +128,16 @@ func TestRouterSendAndSendB(t *testing.T) {
...
+func wait(t *testing.T, f func()) {
 	wait := make(chan int)
 	go func() {
 		f()
 		wait <- 0
 	}()
 	select {
 	case <-wait:
-	case <-time.After(timeout):
+	case <-time.After(5 * time.Second):
```
sdojjy, line 139: why use the hardcoded time here?  is it better to extend the timeout on the function call?

overvenus: The `wait` is specialized in actor testing, 5 seconds are longer enough for most cases, we can pass timeout when it's needed in future.

sdojjy (approved):
LGTM

PR #4074, opened 2021-12-26
codec(cdc): fix encoder `max-message-bytes`

3AceShowHand:

when initializing the encoder, make sure `max-message-bytes` in the `opts`, and set identical to producer's configuration.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/sink/codec/craft.go
+++ b/cdc/sink/codec/craft.go
@@ -103,7 +103,7 @@ func (e *CraftEventBatchEncoder) SetParams(params map[string]string) error {
 	var err error
 	maxMessageBytes, ok := params["max-message-bytes"]
 	if !ok {
-		return cerror.ErrSinkInvalidConfig.Wrap(errors.Errorf("max-message-bytes not found"))
+		return cerror.ErrSinkInvalidConfig.Wrap(errors.New("max-message-bytes not found"))
 	}
 
 	e.maxMessageBytes, err = strconv.Atoi(maxMessageBytes)
--- a/cdc/sink/codec/craft.go
+++ b/cdc/sink/codec/craft.go
@@ -108,7 +108,7 @@ func (e *CraftEventBatchEncoder) SetParams(params map[string]string) error {
 
 	e.maxMessageBytes, err = strconv.Atoi(maxMessageBytes)
 	if err != nil {
-		return cerror.ErrSinkInvalidConfig.Wrap(err)
+		return cerror.WrapError(cerror.ErrSinkInvalidConfig, err)
 	}
 	if e.maxMessageBytes <= 0 || e.maxMessageBytes > math.MaxInt32 {
 		return cerror.ErrSinkInvalidConfig.Wrap(errors.Errorf("invalid max-message-bytes %d", e.maxMessageBytes))
--- a/cdc/sink/codec/craft.go
+++ b/cdc/sink/codec/craft.go
@@ -110,7 +110,7 @@ func (e *CraftEventBatchEncoder) SetParams(params map[string]string) error {
 	if err != nil {
 		return cerror.ErrSinkInvalidConfig.Wrap(err)
 	}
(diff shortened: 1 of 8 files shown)


cdc/sink/codec/craft.go
```diff
@@ -101,13 +101,14 @@ func (e *CraftEventBatchEncoder) Reset() {
 // SetParams reads relevant parameters for craft protocol
 func (e *CraftEventBatchEncoder) SetParams(params map[string]string) error {
 	var err error
+	maxMessageBytes, ok := params["max-message-bytes"]
+	if !ok {
+		return cerror.ErrSinkInvalidConfig.Wrap(errors.New("max-message-bytes not found"))
```
amyangfei, line 106:
```suggestion
		return cerror.ErrSinkInvalidConfig.GenWithStack("max-message-bytes not found")
```
This is more simple, and the error is an RFC error

cdc/sink/codec/craft.go
```diff
@@ -101,13 +101,14 @@ func (e *CraftEventBatchEncoder) Reset() {
...
 
-	e.maxMessageBytes = config.DefaultMaxMessageBytes
-	if maxMessageBytes, ok := params["max-message-bytes"]; ok {
-		e.maxMessageBytes, err = strconv.Atoi(maxMessageBytes)
-		if err != nil {
-			return cerror.ErrSinkInvalidConfig.Wrap(err)
-		}
+	e.maxMessageBytes, err = strconv.Atoi(maxMessageBytes)
+	if err != nil {
+		return cerror.ErrSinkInvalidConfig.Wrap(err)
```
amyangfei, line 111:
```suggestion
		return cerror.WrapError(cerror.ErrSinkInvalidConfig, err)
```
better to use `cerror.WrapError`, it will generate a new rfc error instead of using the original instance.

amyangfei:
Need to confirm:
- Is cherry pick to release branches needed
- whether release note is needed? (a new introduced bug or an old bug)

cdc/sink/codec/json.go
```diff
@@ -615,12 +615,14 @@ func (d *JSONEventBatchEncoder) Reset() {
 func (d *JSONEventBatchEncoder) SetParams(params map[string]string) error {
 	var err error
 
-	d.maxMessageBytes = config.DefaultMaxMessageBytes
-	if maxMessageBytes, ok := params["max-message-bytes"]; ok {
-		d.maxMessageBytes, err = strconv.Atoi(maxMessageBytes)
-		if err != nil {
-			return cerror.ErrSinkInvalidConfig.Wrap(err)
-		}
```
amyangfei, line 623: The original logic is not elegant, but what is the bug in it?
- If `max-message-bytes` is set in sink-uri, the following lines set it: https://github.com/pingcap/tiflow/blob/43a599d08044606dc9af8149a79f92d2cc45d08f/cdc/sink/mq.go#L419-L422
- If `max-message-bytes` is not set in sink-uri, `config.DefaultMaxMessageBytes` will be used here.

3AceShowHand: we have to update the config once more.because we would query Kafka broker to get `topicMaxMessageBytes` or `brokerMessageMaxBytes`, and then choose the minimum of min(`max-message-bytes`, `topicMaxMessageBytes`, `brokerMessageMaxBytes`)

amyangfei: make sense, maybe more scenario tests can be added, either in integration test or test infra.

overvenus (approved):

cdc/sink/codec/json.go
3AceShowHand: we have to refact the configuration in the kafka, `Config`, `sarama.Config`, `opts`, and cover the adjust logic.

PR #4100, opened 2021-12-28
ha, config: add config and etcd KV for TaskCliArgs

lance6716:

Issue Number: ref https://github.com/pingcap/tiflow/issues/4106

as title.

the expected usage in next PR is:
1. master puts the `TaskCliArgs` or removes `TaskCliArgs`(when it is empty)
2. master puts subtask configs
3. worker received subtask configs
4. worker gets `TaskCliArgs`
5. after `TaskCliArgs` is taken effects, worker removes `TaskCliArgs`

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

Related changes

### Release note 

```release-note
`None`.
```


Ehco1996 (approved):
LGTM

just ask a question:

> after TaskCliArgs is taken effects, worker removes TaskCliArgs

when we do shard merge task ,more than one worker will received subtask, so every worker will try to delete this `TaskCliArgs`?  what if wokrer2 try get `TaskCliArgs` after workr1 delete this ? waitting your next PR

GMHDBJD:
I prefer add a config item in task config and only affect in command line than we no need to add new kv in etcd.

Ehco1996:
> 

seems one config item is not enough eg: add `started-time` in subtaskcfg

wokrer1 can deal with this config when first receive this config , but after pause/stop task and the worker receive this subtask config next time, what should worker do?

lance6716:
> LGTM
> 
> just ask a question:
> 
> > after TaskCliArgs is taken effects, worker removes TaskCliArgs
> 
> when we do shard merge task ,more than one worker will received subtask, so every worker will try to delete this `TaskCliArgs`? what if wokrer2 try get `TaskCliArgs` after workr1 delete this ? waitting your next PR

Thanks, I think I should let TaskCliArgs become a subtask level etcd KV, that's more easier.

lance6716:
> I prefer add a config item in task config or subtask config and only affect in command line than we no need to add new kv in etcd.

since `start-time` has higher priority than checkpoint, and when DM worker failover happens it's not expected to start replication from `start-time`, we should remove `start-time` in etcd. Rewriting the subtask KV is more harder than rewriting/deleting a seperate KV.

D3Hunter (approved):
LGTM

Issue #4115, opened 2021-12-28
some changefeeds become [CDC:ErrSnapshotLostByGC] when upgrade from 5.1.0 to 5.1.1-hotfix


Tammyxia:
### What did you do?

- k8s, create tidb cluster v5.1.0, 3 changefeeds: 
    1. "sink-uri": "mysql://root:@downstream1.testbed-tammy-586vr:3306",
    2. "sink-uri": "kafka://downstream2-kafka.testbed-tammy-586vr:9092/cdc-test1?max-message-bytes=1048576\u0026protocol=canal-json",
    3. "sink-uri": "kafka://downstream2-kafka.testbed-tammy-586vr:9092/cdc-test2?max-message-bytes=1048576",
     all changefeed gc-ttl using default value: "gc-ttl\":86400

- run workload: tpcc 50 warehouse, prepare and run

-  upgrade cdc only for cdc hotfix version 5.1.1-xxx, meanwhile, tpcc is running.

### What did you see instead?

- 2 of 3 changefeeds has became failed: "CDC:ErrSnapshotLostByGC" after upgrade successful
- / # /cdc cli changefeed list --pd=http://upstream-pd:2379
[
  {
    "id": "kafka-canal-json-before",
    "summary": {
      "state": "failed",
      "tso": 430093979858763792,
      "checkpoint": "2021-12-28 07:58:48.929",
      "error": {
        "addr": "upstream-ticdc-0.upstream-ticdc-peer.testbed-tammy-586vr.svc:8301",
        "code": "CDC:ErrSnapshotLostByGC",
        "message": "[CDC:ErrSnapshotLostByGC]fail to create or maintain changefeed due to snapshot loss caused by GC. checkpoint-ts 430093979858763792 is earlier than or equal to GC safepoint at 430093979858763792"
      }
    }
  },
  {
    "id": "kafka-default-before",
    "summary": {
      "state": "failed",
      "tso": 430093979858763792,
      "checkpoint": "2021-12-28 07:58:48.929",
      "error": {
        "addr": "upstream-ticdc-0.upstream-ticdc-peer.testbed-tammy-586vr.svc:8301",
        "code": "CDC:ErrSnapshotLostByGC",
        "message": "[CDC:ErrSnapshotLostByGC]fail to create or maintain changefeed due to snapshot loss caused by GC. checkpoint-ts 430093979858763792 is earlier than or equal to GC safepoint at 430093979858763792"
      }
    }
  },
  {
[...]

PR #4116, opened 2021-12-28
tests(ticdc): add upgrade gc safepoint integration tests

overvenus:

Add an upgrade test that covers a case which changefeed checkpoint may be set to gc safepoint by old cdc. 

Issue Number:  ref https://github.com/pingcap/tiflow/issues/4115

### Check List 

Tests 

 - Integration test

Related changes

### Release note 

```release-note
None
```


tests/integration_tests/upgrade_gc_safepoint/run.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/bash
+
+set -e
```
lance6716, line 3:
```suggestion
set -eu
```

I hope we can switch to `set -eu`, see https://github.com/pingcap/tiflow/pull/4160 .

also the problem of wrong position of `pd_cluster_id` in `gc_safepoint` can be found by it https://github.com/pingcap/tiflow/pull/4160#issuecomment-1002933474

overvenus: Please go on with #4160, this PR will not be merged until #4115 is fixed.

PR #4129, opened 2021-12-29
consumer(ticdc): update the logic of consumer.

### LINKED ISSUE #4128: TiCDC add a test to cover the invariant between `MqMessageTypeResolved` Event and other Events.

TiCDC would send `MqMessageTypeResolved` messages to the MQ sink, with protocol such as `open-protocol`, `craft`, and `canal-json`.

We need to find a way to guarantee the semantic of the `MqMessageTypeResolved` message:

If the downstream MQ sink receive a `MqMessageTypeResolved` message with `ts = a`, it should be guaranteed that:
* `all received messages should have ts less than a`
* `no more messages with ts less than a will be received by the downstream`.

We have `kafka-consumer` at the moment, investigating whether it can be used to solve the problem mentioned above.

_No response_

_No response_

3AceShowHand:

Issue Number: close #4128 

Use `kafka-consumer` program to maintain the order relationship between `commitTs` and `watermarkTs`.

`watermarkTs` also known as `ResolvedTs` in TiCDC source code.

Modify the implementation detail of the `kakfa-consumer` program.

* If meet `ResolvedTs` fallback, which means receive `Resolved Event` with `Ts` less than current `ResolvedTs`, just crash the consumer.
* If meet `Row Changed Event` or `DDL Event` with `Ts` not greater than the `globalResolvedTs`, just ignore these events should already be seen.

After this PR, the consumer should still be running stable.

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`.
```


asddongmen:
please remove this TODO:https://github.com/pingcap/tiflow/blob/master/tests/integration_tests/move_table/run.sh#:~:text=%23%20TODO%3A%20remove%20after%20kafka%2Dconsumer/main.go%20is%20fixed 
and try to run kafka integration test

cmd/kafka-consumer/main.go
```diff
@@ -453,11 +453,18 @@ ClaimMessages:
 					log.Fatal("decode message value failed", zap.ByteString("value", message.Value))
 				}
 				resolvedTs := atomic.LoadUint64(&sink.resolvedTs)
-				if resolvedTs < ts {
+				if ts < resolvedTs {
+					log.Fatal("partition resolved ts fallback",
```
overvenus, line 457:
```suggestion
					log.Panic("partition resolved ts fallback",
```

Prefer `log.Panic`.

cmd/kafka-consumer/main.go
```diff
@@ -440,31 +441,37 @@ ClaimMessages:
...
 				}
 			case model.MqMessageTypeResolved:
 				ts, err := batchDecoder.NextResolvedEvent()
 				if err != nil {
-					log.Fatal("decode message value failed", zap.ByteString("value", message.Value))
+					log.Panic("decode message value failed", zap.ByteString("value", message.Value))
 				}
 				resolvedTs := atomic.LoadUint64(&sink.resolvedTs)
-				if resolvedTs < ts {
+				if ts < resolvedTs {
```
overvenus, line 456: Could you add some comments about the ordering of the ts and the resolvedTs?

overvenus (approved):
LGTM

asddongmen (approved):

Issue #4158, opened 2021-12-30
dm-syncer IT is not stable


lance6716:
### Which jobs are flaking?

dm-syncer

### Which test(s) are flaking?

DM-IT

### Jenkins logs or GitHub Actions link

https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/2312/pipeline

### Anything else we need to know

- Does this test exist for other branches as well?

- Has there been a high frequency of failure lately?

Issue #4159, opened 2021-12-30
refine tests


lance6716:
### Is your feature request related to a problem?

to walkaround link-issue-checker in PR

### Describe the feature you'd like

refine tests

PR #4160, opened 2021-12-30
*(both): replace set -e to set -ru and disable some unstable test

lance6716:

Issue Number: ref https://github.com/pingcap/tiflow/issues/4159

close https://github.com/pingcap/tiflow/issues/4158

replace `^set -e\n` to `^set -eu\n` to reduce the chance accidentally format the whole disk. 

Meaning of `set` options can be found in https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html. When we execute some internal shell scripts, both variables in `WORK_DIR=$OUT_DIR/$TEST_NAME` is not set thus empty strings, so a following `rm -rf $WORK_DIR` will become `rm -rf /`. `set -u` can found unbound variables and stop execution.

I added an echo to show what's going to execute

before:

```
$ bash tests/integration_tests/batch_add_table/run.sh 
rm -rf / && mkdir -p /
```

after 

```
$ bash tests/integration_tests/batch_add_table/run.sh
tests/integration_tests/batch_add_table/run.sh: line 7: OUT_DIR: unbound variable
```

### Check List 

Tests 

 - No code

### Release note 

```release-note
 `None`.
```


lonng:
Prefer to add the link https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html to the PR description as the reference of why we add `-u` parameter.

lonng (approved):

lance6716:
/hold

```
/home/jenkins/agent/workspace/cdc_ghpr_integration_test/go/src/github.com/pingcap/tiflow/tests/integration_tests/gc_safepoint/run.sh: line 96: pd_cluster_id: unbound variable
```

https://github.com/pingcap/tiflow/blob/2d5a53478a20402db3928fad90e3e97c6eeb89e2/tests/integration_tests/gc_safepoint/run.sh#L96

I doult if `clear_gc_worker_safepoint` is really worked and this test is ineffective, PTAL at your convenient @hi-rustin

lance6716:
waiting https://github.com/pingcap/tiflow/pull/4116

Ehco1996 (approved):
LGTM thank for this work

lance6716:
seems downstream_diff_index always fails …

lance6716:
temporary removed tests will be enabled in https://github.com/pingcap/tiflow/pull/4172

PR #4176, opened 2021-12-31
ddl(dm): make skipped ddl pass `SplitDDL()`

### LINKED ISSUE #4173: CREATE VIEW will interrupt replication task
### What did you do?

start a task

```
MySQL [mydb]> create view t10 as select * from t1;
```

task is not interrupted

### What did you see instead?

```
                    "name": "test",
                    "stage": "Paused",
                    "unit": "Sync",
                    "result": {
                        "isCanceled": false,
                        "errors": [
                            {
                                "ErrCode": 11007,
                                "ErrClass": "functional",
                                "ErrScope": "internal",
                                "ErrLevel": "high",
                                "Message": "startLocation: [position: (mysql-bin.000001, 1794), gtid-set: 2c43df5b-5eea-11ec-aff5-0242ac110002:1-6], endLocation: [position: (mysql-bin.000001, 2009), gtid-set: 2c43df5b-5eea-11ec-aff5-0242ac110002:1-7], origin SQL: [CREATE ALGORITHM=UNDEFINED DEFINER=`root`@`%` SQL SECURITY DEFINER VIEW `t10` AS select * from t1]: unknown type ddl \u0026{ddlNode:{stmtNode:{node:{text:CREATE ALGORITHM=UNDEFINED DEFINER=`root`@`%!`(MISSING) SQL SECURITY DEFINER VIEW `t10` AS select * from t1 offset:0}}} OrReplace:false ViewName:0xc005fa4a50 Cols:[] Select:0xc0004a2ff0 SchemaCols:[] Algorithm:0 Definer:root@%!S(MISSING)ecurity:0 CheckOption:1}",
                                "RawCause": "",
                                "Workaround": "Please manually handle it using dmctl (skipping the DDL statement or replacing the DDL statement with a specified DDL statement). For details, see https://docs.pingcap.com/tidb-data-migration/stable/handle-failed-sql-statements"
                            }
```

### Versions of the cluster

```console
v5.3.0
```
```console
```
```console
```
```console
(leave TiUP or manually here)
```
```console
>
>
```

```console
```

okJiang:

Issue Number: close #4173 

- make ddl which will be skipped in `skipQueryEvent()` pass `SplitDDL`, and don't report error.

### Check List 

Tests 

 - Integration test

### Release note 

```release-note
`None`.
```


dm/pkg/parser/common.go
```diff
@@ -318,6 +318,7 @@ func SplitDDL(stmt ast.StmtNode, schema string) (sqls []string, err error) {
 		v.Table = table
 
 		return sqls, nil
+	case *ast.CreateViewStmt:
```
okJiang, line 321: How about add all `will be skipped in SkipQueryEvent()` ddl here? @lance6716 

https://github.com/pingcap/tiflow/blob/95ddc8a60b9022320c7a97e0e71d3db5763a59df/dm/pkg/utils/util.go#L45-L100

okJiang: or we execute 

```go
	if utils.IsBuildInSkipDDL(qec.originSQL) {
		return true, nil
	}
```

before `SpiltDDL`

lance6716: there're more cases than CREATE VIEW. I think we should do as https://github.com/pingcap/tiflow/blob/95ddc8a60b9022320c7a97e0e71d3db5763a59df/dm/syncer/syncer.go#L2366-L2380

dm/syncer/syncer.go
```diff
@@ -2357,26 +2357,35 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
...
-			return err2
+			err = err2
+			return
 		}
 		if !needSkip {
-			return err
+			return
 		}
-		// don't return error if parse fail and filter success
+		// don't return error if filter success
```
Ehco1996, line 2374: could you explain why you use defer here ?, old logic is parse -> filter -> track
but now is parse -> handle -> filter 🤔

how about just mute spefic err in https://github.com/pingcap/tiflow/blob/1df27c666ac4266e88a6ca095c4edcb62b12a610/dm/syncer/syncer.go#L2374

okJiang: old logic is

1. parse fail
2. skipSQLByPattern
	1. return err if skip fail
	2. record something if skip success

new logic is

1. for all error (parse fail + DML node + split fail + process fail + ...)
2. skipSQLByPattern
	1. return err if skip fail
	2. record something if skip success

Ehco1996: yes, but this looks like use defer to impl `try/exception`😂 i konw it can solve #4173, but this looks little strange to me

okJiang: In fact, I don't know why we don't `skipSQLByPattern` at beginning directly, but at later operation failed. @lance6716

lance6716: 1. user can configure filter to avoid some ddl being executed downstream
2. user can configure filter to avoid parse error, and we have some built-in filters
3. schema tracker should track every ddl except for the one that will cause error

any implementation is ok

okJiang: > In fact, I don't know why we don't `skipSQLByPattern` at beginning directly, but at later operation failed. @lance6716

> schema tracker should track every ddl except for the one that will cause error

yes, it is order to track every ddl except for the one that will cause error. It's easy to forget. I will add some comment.

okJiang: > yes, but this looks like use defer to impl `try/exception`😂 i konw it can solve #4173, but this looks little strange to me

Indeed. But I haven't thought of a better way temporarily.

dm/tests/all_mode/run.sh
```diff
@@ -377,6 +377,9 @@ function run() {
 	# use sync_diff_inspector to check full dump loader
 	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
 
+	# check create view(should be skipped) success, then skip it
```
Ehco1996, line 380:
```suggestion
	# check create view(should be skipped by func `skipSQLByPattern`) will not stop sync task
```
```suggestion
	# check create view(should be skipped) success, then skip it
```

Ehco1996 (commented):
restLGTM

dm/syncer/syncer.go
Ehco1996: i thought it again,don't have better suggesion either, the problem is we want to mute err that caused by sql which is user want to skipp,but we also want to track these err 😂

so please add more comment in defer func to explain why we need handle err in this way

dm/syncer/syncer.go
```diff
@@ -2357,26 +2357,37 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
...
 
-	stmt, err := parseOneStmt(qec)
-	if err != nil {
-		// return error if parse fail and filter fail
+	defer func() {
+		if err == nil {
+			return
+		}
+		// why not `skipSQLByPattern` at beginning, but at defer?
+		// it is order to track every ddl except for the one that will cause error.
```
lance6716, line 2366:
```suggestion
		// it is in order to track every ddl except for the one that will cause error.
```

dm/tests/all_mode/run.sh
```diff
@@ -377,6 +377,9 @@ function run() {
 	# use sync_diff_inspector to check full dump loader
 	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
 
+	# check create view(should be skipped by func `skipSQLByPattern`) will not stop sync task
+	run_sql_source1 "create view all_mode.t1_v as select * from all_mode.t1 where id=0;"
```
lance6716, line 381: please add a check about task is not interrupted. Currently the test will fail at line 449, that's too far

lance6716 (commented):
rest lgtm

dm/tests/all_mode/run.sh
```diff
@@ -324,11 +324,11 @@ function run() {
...
 	run_sql_source2 "SET @@global.time_zone = '+02:00';"
-	test_expression_filter
-	test_fail_job_between_event
-	test_session_config
-	test_query_timeout
-	test_stop_task_before_checkpoint
+	# test_expression_filter
+	# test_fail_job_between_event
+	# test_session_config
+	# test_query_timeout
```
Ehco1996, line 330: please revert this

Ehco1996 (approved):
LGTM

dm/tests/all_mode/run.sh
```diff
@@ -324,11 +324,11 @@ function run() {
...
+	# test_stop_task_before_checkpoint
```
okJiang, line 331:
```suggestion
	test_expression_filter
	test_fail_job_between_event
	test_session_config
	test_query_timeout
	test_stop_task_before_checkpoint
```

lance6716 (approved):

okJiang:
Is this pr ready to merge? @lance6716

Issue #4177, opened 2021-12-31
syncer: after binlog(handle-error) skip we must reset schema for dm


lichunzhu:
### What did you do?

Use `dmctl binlog skip` to skip a ddl error. For example, to skip errors when we add a not fully dropped column.

### What did you expect to see?

DM can sync following dml correctly.

### What did you see instead?

DM reports error: schema mismatch dmls

### Versions of the cluster

DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):
```console
v5.3.0
```

### current status of DM cluster (execute `query-status <task-name>` in dmctl)

```console
(paste current status of DM cluster here)
```

PR #4178, opened 2021-12-31
syncer(dm): track skipped ddls when using dmctl binlog to skip some ddls

lichunzhu:

Issue Number: close https://github.com/pingcap/tiflow/issues/4177 , close https://github.com/pingcap/tiflow/issues/3239

Track skipped ddls when we use `binlog skip`. However, syncer will only track replaced ddl instead of original ddl for `binlog replace.`

### Check List 

Tests 

 - Unit test
 - Integration test
Use `binlog skip` for *add not fully dropped column` error. DM can replicate data correctly after skip. We don't have to set schema again.

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Support updating table schema when using dmctl binlog to skip some ddls
```


lichunzhu:
/run-all-tests

https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/2426/pipeline

dm/syncer/syncer.go
```diff
@@ -1875,6 +1875,26 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 						// revert currentLocation to startLocation
 						currentLocation = startLocation
 					} else if op == pb.ErrorOp_Skip {
+						ec := eventContext{
```
Ehco1996, line 1878: how about move  https://github.com/pingcap/tiflow/blob/fc70dbde873be0c09fa0f8f603772d2d95d59a8c/dm/syncer/syncer.go#L1923  to  https://github.com/pingcap/tiflow/blob/fc70dbde873be0c09fa0f8f603772d2d95d59a8c/dm/syncer/syncer.go#L1841

so that we can resue struce `ec`, in replaceErr logic, we can just update the filed in `ec`

dm/syncer/syncer.go
```diff
@@ -2967,6 +2987,55 @@ func (s *Syncer) trackDDL(usedSchema string, trackInfo *ddlInfo, ec *eventContex
 	return nil
 }
 
+func (s *Syncer) trackOriginDDL(ev *replication.QueryEvent, ec eventContext, originSQL string) error {
```
Ehco1996, line 2990: seems we can get `originSQL` by use `strings.TrimSpace(string(ev.Query))`

Ehco1996: btw most code here is same as `handleQueryEvent` can we try to merge them?

dm/tests/shardddl2/run.sh
```diff
@@ -485,18 +485,14 @@ function DM_DropAddColumn_CASE() {
 		"\"source 'mysql-replica-02' has no error\"" 1
 
 	# after we skip ADD COLUMN, we should fix the table structure
```
Ehco1996, line 487: `# after we skip ADD COLUMN, we should fix the table structure`

this comment is no need any more

dm/tests/shardddl2/run.sh
```diff
@@ -485,18 +485,14 @@ function DM_DropAddColumn_CASE() {
...
-
-	echo 'CREATE TABLE `tb1` ( `a` int(11) NOT NULL, `b` int(11) DEFAULT NULL, `c` int(11) DEFAULT NULL, PRIMARY KEY (`a`) /*T![clustered_index] NONCLUSTERED */) ENGINE=InnoDB DEFAULT CHARSET=latin1 COLLATE=latin1_bin' >${WORK_DIR}/schema.sql
-	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
-		"binlog-schema update test ${shardddl1} ${tb1} ${WORK_DIR}/schema.sql -s mysql-replica-01" \
-		"\"result\": true" 2
-
-	run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
-		"resume-task test" \
-		"\"result\": true" 3
+	# TODO: support flush checkpoint when ddl is skipped
```
Ehco1996, line 488: can you explian the purpose of this commented TODO test ?

dm/syncer/syncer.go
lichunzhu: Many fields vary when we create this `ec`. I'm afraid that others may lose some fields' changes when they try to deal with this variable. I have removed all the unused fields.

dm/tests/shardddl2/run.sh
lichunzhu: deleted

dm/tests/shardddl2/run.sh
lichunzhu: fix this part in https://github.com/pingcap/tiflow/pull/4178/commits/999b4ecb24e3dab38d67d5fe676da3fa1f9be453

dm/syncer/syncer.go
lichunzhu: I have thought about this before, but finally, I choose to rewrite a new method.
Because many places in `handleQueryEvent` may return an error before skipping. If we got an error before we skip this ddl, we can never skip this ddl through `binlog skip`.

dm/syncer/syncer.go
```diff
@@ -2967,6 +2988,70 @@ func (s *Syncer) trackDDL(usedSchema string, trackInfo *ddlInfo, ec *eventContex
 	return nil
 }
 
+func (s *Syncer) trackOriginDDL(ev *replication.QueryEvent, ec eventContext, originSQL string) (map[string]map[string]struct{}, error) {
+	if originSQL == "BEGIN" || originSQL == "" || utils.IsBuildInSkipDDL(originSQL) {
```
Ehco1996, line 2992: don't need param:`originSQL`, we can use strings.TrimSpace(string(ev.Query)) instead

Ehco1996 (approved):
restLGTM
please also fix the lint

dm/syncer/syncer.go
okJiang: Now it seems that no matter whether this DDL is executed or not, as long as this DDL can be tracked(not build skip ddl, can be parsed, can be split, can genDDLInfo), we need to track it.

So, how about move all `TrackDDL` to a unified place, maybe before `binlog skip`.

Do you think this is feasible?

lance6716 (commented):
/cc @okJiang 

also need a `make check` or `make fmt`

dm/syncer/syncer.go
lichunzhu: Could you please show me which place is `before binlog skip`?

okJiang: This `TrackDDL` seems can't unify with others(at least I didn't get any methods), because it didn't report error as others.

The others `TrackDDL` probably can unify(to be discussed further), but is is another problem. We needn't  talk it in this pr.

okJiang (approved):
LGTM
also close https://github.com/pingcap/tiflow/issues/3239

dm/syncer/syncer.go
```diff
@@ -1875,7 +1875,28 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
...
 					} else if op == pb.ErrorOp_Skip {
+						ec := eventContext{
+							tctx:            tctx,
+							header:          e.Header,
+							startLocation:   &startLocation,
+							currentLocation: &currentLocation,
+							lastLocation:    &lastLocation,
+						}
+						var sourceTbls map[string]map[string]struct{}
+						sourceTbls, err = s.trackOriginDDL(ev, ec)
```
WizardXiao, line 1886: if user use `binlog-schema update` to set schema in tracker, and then use `binlog skip`, we maybe should note that original ddls will be still execute. just a slight worry, i don't find error conditions.

PR #4188, opened 2021-12-31
kafka: support for multi-topic

dveeden:

This adds support for writing to multiple kafka topics

Issue Number: ref #1147

By default the static topic name specified in the URL is used. But tables are registered in the registry by `{schema}_{table}`.

In  #3936 support is added to set the name in the registry to match the topic name. The benefits of this are:
- Allows multiple unrelated CDC instances with similarly named tables to write to the same kafka broker without conflicts by manually specifying unique topic names.
- Having the topic and registry names matching, so KSQL etc can use them.
- With a changefeed per table this works for small-ish setups.

This PR does set the topic name dynamically to `{topic_from_url}_{schema}_{table}`. This makes this easier to setup and maintain.

Unsolved issues / TODO:

- [ ] Rebase this on top of #3936 once it has been merged and add a naming strategy that matches `{topic_from_url}_{schema}_{table}` as the current `{schema}_{table}` naming doesn't match. The prefix is needed in case multiple unrelated schema/table combinations are used with unrelated CDC instances but the same kafka broker.
- [ ] Add a config option for this as we can't enable this by default as that would break backwards compatible behavior
- [ ] Fix the hardcoded `NumPartitions: 3` and `ReplicationFactor: 1`
- [ ] Maybe not create the old static topic name when in multi topic mode
- [ ] Add tests etc.
- [ ] Make sure we set `MaxMessageBytes` etc. for the topics. Maybe re-unify this with `validateMaxMessageBytesAndCreateTopic`

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Support for writing to a topic-per-table in Kafka was added
```

### DIFF
--- a/cdc/sink/producer/kafka/kafka.go
+++ b/cdc/sink/producer/kafka/kafka.go
@@ -68,6 +68,7 @@ type kafkaSaramaProducer struct {
 	closeCh chan struct{}
 	// atomic flag indicating whether the producer is closing
 	closing kafkaProducerClosingFlag
+	admin kafka.ClusterAdminClient
 }
 
 type kafkaProducerClosingFlag = int32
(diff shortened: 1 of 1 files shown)


hi-rustin (commented):
We are currently implementing it according to the RFC. PRs can be found at https://github.com/pingcap/tiflow/issues/4423.
So I will close this PR. thank you for your contribution!

Issue #2183, opened 2021-06-29
Metrics labels should not contain ID, table name or other unbounded sets of values


overvenus:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

Running a changefeed which captures 20k tables produces large metric records.

We should not put ID and table names in metrics labels.

> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.
>
> https://prometheus.io/docs/practices/naming/#labels

Here is a list of problematic metrics

- [x] ticdc_kvclient_region_token
 * not monitored
- [x] ticdc_sorter_consume_count
 *  no need to update
- [x] ticdc_sorter_event_count
 * no need to update
- [x] ticdc_sorter_flush_count_histogram
 * no need to update
- [x] ticdc_sorter_merge_count_histogram
 * no need to update
- [x] ticdc_sorter_merger_start_ts_gauge
 * not monitored
- [x] ticdc_sorter_resolved_ts_gauge
 * not monitored
- [x] ticdc_kvclient_channel_size
 * not monitored
- [x] ticdc_processor_table_memory_consumption
   - [x] histogram
   - [x] update grafana
     - [x] Processor Memory Consumption Per Capture
     - [x] Processor Memory Consumption Per Table -> Per Capture 
- [x] ticdc_processor_table_resolved_ts
  - [x] Table Resolved ts 
- [x] ticdc_puller_event_chan_size
   - [x] histogram
   - [x] update grafana 
     - [x] Puller buffer size
- [x] ticdc_puller_output_chan_size
   - [x] histogram
   - [x] update grafana 
     - [x] Puller buffer size 
- [x] ticdc_puller_resolved_ts
 * not monitored
- [x] ticdc_puller_txn_collect_event_count
 * no need to update

**Any unbounded sets of labels should be avoided, and use other ways to record critical metrics.**

For above metrics, we should

* Remove the `table` and other labels that have high cardinality.
* Use sampling.

Eg. we can rewrite `ticdc_processor_table_memory_consumption` to 

```go
// Sampling, turn gague/counter into histogram.
tableMemoryGauge = prometheus.NewGaugeVec(
	prometheus.GaugeOpts{
		Namespace: "ticdc",
		Subsystem: "processor",
[...]

PR #4198, opened 2022-01-03
processor(ticdc): add min resolved/checkpoint table ID metrics

overvenus:

Replace a high cardinality metric with a low one.

Issue Number: ref #2183

Add min resolved/checkpoint table ID metrics.

### Check List 

image
Side effects

Related changes

### Release note 

```release-note
None
```


3AceShowHand (approved):

3AceShowHand:
It looks good.

We should find a better way to represent the graph.

ben1009:
when/how to use the metric ?

overvenus:
> when/how to use the metric ?

It is useful when there are some locks left in a table which blocks resolved ts, we can get the table ID from the metrics, and resolve lock manually.

ben1009 (approved):

Issue #3654, opened 2021-11-29
table scheduler is changefeed level, which could lead to imbalance in some scenarios


amyangfei:
### Is your feature request related to a problem?

1. Setup a TiDB cluster with 3 TiCDC nodes. Create 10 tables in upstream
2. Create **10 changefeeds**, each changefeed replicates **one table**.

test with cdc master@https://github.com/pingcap/ticdc/commit/39cfb3eda4fee3a340a8b8dfd618195967455359

### Describe the feature you'd like

The tables replication stream can be dispatched to **3 TiCDC nodes evenly**.
But we will observe the tables are not even among 3 TiCDC nodes. This is caused by the table scheduler is in changefeed level, ref: https://github.com/pingcap/ticdc/blob/39cfb3eda4fee3a340a8b8dfd618195967455359/cdc/owner/changefeed.go#L76-L79
and the scheduler algorithm is based on min workload with a random map accessing. https://github.com/pingcap/ticdc/blob/39cfb3eda4fee3a340a8b8dfd618195967455359/cdc/owner/scheduler.go#L206

`workloads` is a map with some undeterminable accessing, but not enough.

**workload ratio is `7:2:1` among three TiCDC nodes.**

```
/tidb/cdc/task/workload/95f9f6e2-7edd-41cf-ad12-c5e91e20c624/test-cf-1
{"57":{"workload":1}}
/tidb/cdc/task/workload/ce10785f-0c81-4882-b0ee-d4b702178549/test-cf-2
{"57":{"workload":1}}
/tidb/cdc/task/workload/ce10785f-0c81-4882-b0ee-d4b702178549/test-cf-3
{"57":{"workload":1}}
/tidb/cdc/task/workload/dbc97296-ecb8-457c-9d62-9f89220635ad/test-cf-10
{"57":{"workload":1}}
/tidb/cdc/task/workload/dbc97296-ecb8-457c-9d62-9f89220635ad/test-cf-4
{"57":{"workload":1}}
/tidb/cdc/task/workload/dbc97296-ecb8-457c-9d62-9f89220635ad/test-cf-5
{"57":{"workload":1}}
/tidb/cdc/task/workload/dbc97296-ecb8-457c-9d62-9f89220635ad/test-cf-6
{"57":{"workload":1}}
/tidb/cdc/task/workload/dbc97296-ecb8-457c-9d62-9f89220635ad/test-cf-7
{"57":{"workload":1}}
[...]

PR #4202, opened 2022-01-04
scheduler(cdc): add randomness to the scheduler(V2)

### LINKED ISSUE #4203: Scheduler v2 may produce imbalance workload in some scenarios
### What did you do?

1. Set up 8 nodes TiCDC cluster with scheduler v2 enabled. 
2. Create a changefeed that captures 60k tables.
3. Kill one TiCDC node.
4. Wait 10 minutes and start the killed node.

We can see evenly distributed tables on each node, but workload is imbalance.

Tables are evenly distributed, and workload is balance on each node.

### What did you see instead?

We can see evenly distributed tables on each node, but workload is imbalance.

| tables | workload |
| -- | -- |
| image | image |

### Versions of the cluster

```console
v5.4.0
```

```console
v5.4.0
```

### LINKED ISSUE #4204: Owner is blocked by scheduler v2 when there are many tables need to be scheduled
### What did you do?

1. Set up 8 nodes TiCDC cluster with scheduler v2 enabled.
2. Create a changefeed that captures 60k tables.

There are about 15 minutes that owner history is blank. 

From flamegraph, we can see owner goroutine filled with `BaseScheduleDispatcher.addTable`. 

| owner history | flamegraph | 
| -- | -- |
| image | image |

Owner should not be blocked by scheduler v2 more than 1 second.

### What did you see instead?

Owner is blocked more than 1 second.

### Versions of the cluster

```console
v5.4.0
```

```console
v5.4.0
```

liuzix:

Issue Number: close #4203 close #4204 ref #3654 
- Since the scheduler is per changefeed, when the changefeeds are many and tables in each changefeed are few, workload imbalance among nodes is likely to happen (#3654).
- If the tables are created in an order that correlates to the workload, the workload can be imbalanced (#4203).
- When the total number of tables is large, finding a target to dispatch a table can be slow and consumes a lot of CPU (#4204).

- Introduced randomness to `scheduler/balancer` to 1) add some randomness in choosing dispatch target, 2) add some randomness in choosing victims when rebalancing.
- Avoided copying table records when trying to find victims when rebalancing.

### Check List 

Tests 

 - Unit test
 - Integration test

Performance
- A benchmark is added to verify the fix for #4204
- Without the changes
```
goos: linux
goarch: amd64
pkg: github.com/pingcap/tiflow/cdc/scheduler
cpu: Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz
BenchmarkAddTable
BenchmarkAddTable-16    	   10000	    519251 ns/op
PASS
```

- With the changes
```
goos: linux
goarch: amd64
pkg: github.com/pingcap/tiflow/cdc/scheduler
cpu: Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz
BenchmarkAddTable
BenchmarkAddTable-16    	   76887	     14000 ns/op
PASS
```

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
None
```


overvenus (commented):
Could you add a benchmark about `addTable`?

cdc/scheduler/balancer.go
```diff
@@ -127,9 +155,22 @@ func (r *tableNumberBalancer) FindVictims(
 		for tableID := range tables {
 			tableList = append(tableList, tableID)
 		}
-		// We sort the tableIDs here so that the result is deterministic,
-		// which would aid testing and debugging.
-		util.SortTableIDs(tableList)
+
+		if r.random != nil {
```
amyangfei, line 159: In which scenario will `r.random` be nil?

cdc/scheduler/balancer.go
```diff
@@ -68,12 +80,7 @@ func (r *tableNumberBalancer) FindTarget(
 
 	captureWorkload := make(map[model.CaptureID]int)
 	for captureID := range captures {
-		captureWorkload[captureID] = 0
-	}
-
-	for captureID, tables := range tables.GetAllTablesGroupedByCaptures() {
```
amyangfei, line 74: Does this performance issue exists in `FindVictims`?

cdc/scheduler/balancer.go
liuzix: When the balancer has been created with `newDeterministicTableNumberRebalancer`.

cdc/scheduler/balancer.go
liuzix: `FindVictims` is called one magnitude less frequently than `FindTarget`, so the complexity is acceptable. In system tests, `FindVictims` did not have any performance problem.

cdc/scheduler/balancer.go
amyangfei: Yep, I found `newDeterministicTableNumberRebalancer` is only used in unit test, will we really use it in the future.

cdc/scheduler/balancer.go
amyangfei: We can add comment about complexity warning to `GetAllTablesGroupedByCaptures`, in case of abuse in the future.

cdc/scheduler/balancer.go
liuzix: I don't have a plan to use it for now. It is mainly for testing purpose.

cdc/scheduler/balancer.go
liuzix: Yes, sure

amyangfei (approved):

overvenus (approved):

overvenus:
```
=== RUN   TestBalancerFindTargetTied
    balancer_test.go:234: 
        	Error Trace:	balancer_test.go:234
        	Error:      	Should be false
        	Test:       	TestBalancerFindTargetTied
--- FAIL: TestBalancerFindTargetTied (0.00s)
```
test fails

PR #4217, opened 2022-01-04
cdc,pkg(ticdc): refactor pdtime.Clock

### LINKED ISSUE #4216: TiKVStorage and TimeAcquirer can be shared
As title.

overvenus:

Issue Number: close #4216

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


pkg/pdtime/clock.go
```diff
@@ -72,47 +73,48 @@ func (c *TimeAcquirerImpl) Run(ctx context.Context) {
...
+				c.mu.err = err
 				c.mu.Unlock()
 			}
 		}
 	}
 }
 
-// CurrentTimeFromCached return current time from pd cache
-func (c *TimeAcquirerImpl) CurrentTimeFromCached() (time.Time, error) {
+// CurrentTime return current time from pd cache
```
asddongmen, line 92:
```suggestion
// CurrentTime returns current time from timeCache
```

pkg/pdtime/clock.go
```diff
@@ -28,41 +28,42 @@ import (
...
-	return &TimeAcquirerImpl{
+// NewClock return a new PDClock
+func NewClock(pdClient pd.Client) *PDClock {
+	return &PDClock{
 		pdClient: pdClient,
+		stopCh:   make(chan struct{}, 1),
 	}
 }
 
 // Run will get time from pd periodically to cache in pdPhysicalTimeCache
```
asddongmen, line 57:
```suggestion
// Run will get time from pd periodically to cache in timeCache
```

pkg/pdtime/clock.go
```diff
@@ -28,41 +28,42 @@ import (
...
+// PDClock cache time get from PD periodically and cache it
+type PDClock struct {
+	pdClient pd.Client
+	mu       struct {
+		sync.RWMutex
+		timeCache time.Time
+		err       error
+	}
+	cancel context.CancelFunc
+	stopCh chan struct{}
```
asddongmen, line 46: why we need this `stopCh`?

overvenus: Nice catch, I want to synchronize `Stop`.

asddongmen: It seems that `Clock.Stop()` is not thread-safety, will it cause any possible problem?

hi-rustin (approved):

pkg/pdtime/clock.go
asddongmen: > It seems that `Clock.Stop()` is not thread-safety, will it cause any possible problem?

I am wrong about this, Never mind.

asddongmen (approved):

Issue #4106, opened 2021-12-28
dmctl can specify `start-time` to start an incremental migration task


lance6716:
### Is your feature request related to a problem?

When the task is related to many sources, we need adjust binlog-file/binlog-pos/binlog-gtid for each source. Too much work !

### Describe the feature you'd like

`start-task task.yaml --start-time "2021-12-28 12:00:00"` can create an incremental migration task which starts roughly at "2021-12-28 12:00:00"

PR #4220, opened 2022-01-05
dm/binlog: find binlog location by timestamp

D3Hunter:
Issue Number: ref #4106 . this pr add interface to find binlog location by timestamp, either remotely or locally.

find target binlog file by binary search, then find the target timestamp inside it linearly.

we find timestamp only at transaction or ddl boundary.

we only parse binlog events as needed. For remote find, the time used to find the location depends on network speed, for a 3.9MB/s network and 1GB binlog file, may takes 4 or more minutes to find the target location in worst case.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`
```


dm/pkg/binlog/event/common.go
```diff
@@ -35,17 +35,20 @@ type DDLDMLResult struct {
 // GenCommonFileHeader generates a common binlog file header.
 // for MySQL:
 //   1. BinLogFileHeader, [ fe `bin` ]
-//   2. FormatDescriptionEvent
-//   3. PreviousGTIDsEvent
+//   2. FormatDescriptionEvent, depends on genGTID
```
lance6716, line 38: I remember this should be irrelevant to `genGTID`

lance6716 (commented):
will review later

dm/pkg/binlog/pos_finder.go
```diff
@@ -0,0 +1,370 @@
...
+
+		if minTs >= targetTs {
+			end = mid - 1
+		} else {
+			// current binlog maybe the target binlog file, we'll backtrace to it later.
+			begin = mid + 1
+		}
+	}
+	if lastTs >= targetTs {
+		if lastMid == 0 {
```
lance6716, line 198: maybe we can check this case (first binlog is still too late) at the beginning of this function

dm/pkg/binlog/pos_finder.go
```diff
@@ -0,0 +1,370 @@
...
+		newSet, err := event.GTIDsFromMariaDBGTIDListEvent(ev)
+		if err != nil {
+			return nil, err
+		}
+		return newSet, nil
+	}
+	return prevSet, nil
+}
+
+func (r *binlogPosFinder) checkTransactionBoundaryEvent(ev *replication.BinlogEvent) (bool, error) {
```
lance6716, line 242:
```suggestion
func (r *binlogPosFinder) checkTransactionBeginEvent(ev *replication.BinlogEvent) (bool, error) {
```

more often used in codes of this PR

dm/pkg/binlog/pos_finder.go
```diff
@@ -0,0 +1,370 @@
...
+		fallthrough
+	case replication.MARIADB_GTID_EVENT:
+		r.everMetGTIDEvent = true
+		transactionBeginEvent = true
+	case replication.QUERY_EVENT:
+		if !r.everMetGTIDEvent {
+			// user may change session level binlog-format=statement, but it's an unusual operation, so we parse it every time
+			// In MySQL 5.6.x without GTID, the timestamp of BEGIN is not the timestamp of the transaction,
+			// To simplify implementation, we use timestamp of BEGIN as the transaction timestamp,
+			// but this may cause some transaction with timestamp >= target timestamp be skipped.
```
lance6716, line 269: > but this may cause some transaction with timestamp >= target timestamp be skipped.

is this caused by GTID_EVENT and QUERY_EVENT has different timestamp? for example GTID_EVENT has ts = 10, QUERY_EVENT has ts = 12. When we input ts = 11, QUERY_EVENT is skipped because of `if !r.everMetGTIDEvent`

lance6716 (commented):
will review later

dm/pkg/binlog/event/generator.go
```diff
@@ -55,6 +73,15 @@ func NewGenerator(flavor string, serverID uint32, latestPos uint32, latestGTID g
 			return nil, terror.ErrBinlogLatestGTIDNotInPrev.Generate(latestGTID, previousGTIDs)
 		}
 
+		ver, err := semver.NewVersion(version)
+		if err != nil {
+			return nil, err
+		}
+		if ver.Compare(*semver.New("5.7.0")) >= 0 && !genGTID {
+			// 5.7+ add anonymous GTID when GTID is disabled
```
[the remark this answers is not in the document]
D3Hunter: anonymous GTID is written into binlog file when GTID is disabled on mysql 5.7+

dm/pkg/binlog/pos_finder.go
```diff
@@ -0,0 +1,370 @@
...
+		fallthrough
+	case replication.ANONYMOUS_GTID_EVENT: // since 5.7, when GTID not enabled. we use this to avoid parsing query event
+		fallthrough
+	case replication.MARIADB_GTID_EVENT:
+		r.everMetGTIDEvent = true
+		transactionBeginEvent = true
+	case replication.QUERY_EVENT:
+		if !r.everMetGTIDEvent {
+			// user may change session level binlog-format=statement, but it's an unusual operation, so we parse it every time
+			// In MySQL 5.6.x without GTID, the timestamp of BEGIN is not the timestamp of the transaction,
```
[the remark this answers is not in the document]
D3Hunter: it's the timestamp of the first statement in a transaction

dm/pkg/binlog/pos_finder.go
D3Hunter: this case may see In MySQL 5.6.x without GTID, timestamp of BEGIN = timestamp of the first statement, such as:
Query BEGIN ts=1
TableMap ts=1
Rows ts=1
Rows ts=5
Xid ts=10

suppose target ts=10, this transaction is skipped in current impl.

dm/pkg/binlog/pos_finder.go
```diff
@@ -0,0 +1,370 @@
...
+	for begin <= end {
+		mid := (begin + end) / 2
+		currBinlog := binaryLogs[mid]
+
+		minTs, err = r.findMinTimestampOfBinlog(currBinlog)
+		if err != nil {
+			return err
+		}
+
+		r.tctx.L().Info("min timestamp in binlog file", zap.Reflect("file", currBinlog), zap.Uint32("ts", minTs))
```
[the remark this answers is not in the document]
D3Hunter: prefer Reflect since we know it's a struct

dm/pkg/binlog/pos_finder.go
D3Hunter: no much gains

dm/pkg/binlog/pos_finder.go
```diff
@@ -0,0 +1,369 @@
...
+		lastMid = mid
+
+		if minTs >= targetTs {
+			end = mid - 1
+		} else {
+			// current binlog maybe the target binlog file, we'll backtrace to it later.
+			begin = mid + 1
+		}
+	}
+	if lastTs >= targetTs {
```
Ehco1996, line 195: is `lastTs >= targetTs` means we can't find user wanted binlog file?, do we still need to fill targetBinlog field?

dm/pkg/binlog/pos_finder.go
```diff
@@ -0,0 +1,369 @@
...
+			// if it's the last binlog file, then this timestamp is out of range,
+			// else the end of this binlog file is the position we want,
+			// since the timestamp of the first event in next binlog >= target timestamp
+			if r.lastBinlogFile {
+				return nil, AboveUpperBoundBinlogPos, nil
+			}
+			break
+		}
+	}
+	if r.tsBeforeFirstBinlog {
```
Ehco1996, line 362: can we handle `r. tsBeforeFirstBinlog is true`  before above for loop ?

dm/pkg/binlog/pos_finder.go
```diff
@@ -0,0 +1,369 @@
+// Copyright 2019 PingCAP, Inc.
```
Ehco1996, line 1: please fix the copyright and other files, but i am wonder why `make check-copyright`don't work 😂

dm/pkg/binlog/pos_finder.go
D3Hunter: 2 cases may causes lastTs >= targetTs, target timestamp either below smallest timestamp in binlog files in that case we need to return the begin of the first binlog and use `posType` to indicate or we need backtrace to previous file.

dm/pkg/binlog/pos_finder.go
D3Hunter: we need to parse `PREVIOUS_GTIDS_EVENT` to get `gtidSet`

dm/pkg/binlog/pos_finder.go
D3Hunter: only checks its existence
https://github.com/pingcap/tiflow/blob/abefe57c154efb28f91cf1a15702988e1a7249e7/scripts/check-copyright.sh#L4

dm/pkg/binlog/pos_finder.go
Ehco1996: got this, we need to return min binlog file and gtid set in  error msg in this case

D3Hunter: in the first case, we need to return the min position without error. When user start a task for the first time, may give a timestamp much smaller than min binlog timestamp.

dm/pkg/binlog/pos_finder_test.go
```diff
@@ -0,0 +1,447 @@
...
+var _ = Suite(&testPosFinderSuite{})
+
+type testPosFinderSuite struct{}
+
+func genBinlogFile(generator *event.Generator, start time.Time, nextFile string) ([]*replication.BinlogEvent, []byte) {
+	insertDMLData := []*event.DMLData{
+		{
+			TableID:    uint64(1),
+			Schema:     fmt.Sprintf("db_%d", 1),
+			Table:      strconv.Itoa(1),
```
lance6716, line 43: the table is db_1.1 :thinking: 

and rest place in this file

lance6716 (approved):

dm/pkg/binlog/pos_finder_test.go
```diff
@@ -0,0 +1,447 @@
...
+
+	{
+		tcctx := tcontext.NewContext(context.Background(), log.L())
+		finder := NewLocalBinlogPosFinder(tcctx, false, flavor, relayDir)
+		location, posType, err := finder.FindByTs(ts)
+		c.Assert(err, IsNil)
+		// start of second transaction
+		c.Assert(location.Position, Equals, mysql.Position{"mysql-bin.000001", xidEvent.Header.LogPos})
+		c.Assert(location.GTIDSetStr(), Equals, "")
+		c.Assert(posType, Equals, InRangeBinlogPos)
```
Ehco1996, line 130: do we need test other posType ?

dm/pkg/binlog/pos_finder_test.go
```diff
@@ -0,0 +1,447 @@
...
+		var targetEventStart = file2Events[len(file2Events)-1].Header.LogPos
+		finder := NewLocalBinlogPosFinder(tcctx, false, flavor, relayDir)
+		location, posType, err := finder.FindByTs(int64(file3Events[0].Header.Timestamp))
+		c.Assert(err, IsNil)
+		c.Assert(location.Position, Equals, mysql.Position{"mysql-bin.000002", targetEventStart})
+		c.Assert(location.GTIDSetStr(), Equals, "")
+		c.Assert(posType, Equals, InRangeBinlogPos)
+	}
+	{
+		var targetEventStart uint32
```
Ehco1996, line 180: can you please add some comments to explian the difference betweed this three `fileEvents` and `targetEvent` and what is the different purepose of this tree test?

dm/pkg/binlog/pos_finder.go
```diff
@@ -0,0 +1,369 @@
...
+	"github.com/pingcap/tiflow/dm/pkg/binlog/event"
+	"github.com/pingcap/tiflow/dm/pkg/binlog/reader"
+	tcontext "github.com/pingcap/tiflow/dm/pkg/context"
+	"github.com/pingcap/tiflow/dm/pkg/gtid"
+	"github.com/pingcap/tiflow/dm/pkg/utils"
+)
+
+type binlogPosFinder struct {
+	remote     bool
+	tctx       *tcontext.Context
```
Ehco1996, line 35: how about use normal ctx instead of tctx as this is a exported pacakge ?

dm/pkg/binlog/pos_finder.go
```diff
@@ -0,0 +1,369 @@
...
+	position := mysql.Position{Name: r.targetBinlog.name, Pos: FileHeaderLen}
+	gtidSet := gtid.MinGTIDSet(r.flavor)
+
+	binlogReader, err := r.startSync(position)
+	if err != nil {
+		return nil, InvalidBinlogPos, err
+	}
+	defer binlogReader.Close()
+	for {
+		ev, err := binlogReader.GetEvent(r.tctx.Ctx)
```
Ehco1996, line 315: how about add a new ctx para to this func? this can help caller to cancel this operation if they want to, as this func may block for a long time

dm/pkg/binlog/pos_finder_test.go
```diff
@@ -0,0 +1,447 @@
...
+			"ffffffff-ffff-ffff-ffff-ffffffffffff:1-3",
+		}
+		var targetEventStart uint32
+		var targetEvent *replication.BinlogEvent
+		cnt := 0
+		for _, ev := range file1Events {
+			if ev.Header.EventType == replication.GTID_EVENT {
+				targetEvent = ev
+
+				finder := NewLocalBinlogPosFinder(tcctx, true, flavor, relayDir)
```
Ehco1996, line 307: this is a light suggestion because most logic for local and remote are seem, we can do this test in next PR's it-test

seems all tests use local finder, do we need test remote finder? can use `MockStream`

dm/pkg/binlog/pos_finder.go
D3Hunter: `tctx` is a wrapper of a golang context and logger, simpler to use and we need both.

dm/pkg/binlog/pos_finder.go
D3Hunter: caller can just cancel this `r.tctx.Ctx` which is passed from caller.

dm/pkg/binlog/pos_finder_test.go
D3Hunter: see `TestMySQL57GTID`

dm/pkg/binlog/pos_finder_test.go
D3Hunter: try read the part where `fileEvents` is generated, i think it's quite verbose from there.

i think the variable name is quite self-explanatory, don't follow what's misleading you, or what you suggest the variable name should be.

dm/pkg/binlog/pos_finder_test.go
D3Hunter: most logic is location neutral, either local or remote, so let `GetBinaryLogs` and `NewTCPReader` test themself.

Ehco1996 (approved):

PR #4224, opened 2022-01-05
test(cdc): Add defaultvalue integration test

### LINKED ISSUE #4223: Add related integration test and test-infra for DefaultValue

maxshuang:
Issue Number: close #4223 

### Check List 

Tests 

 - Integration test

Related changes

### Release note 

```release-note
`None`.
```

### DIFF
--- a/tests/integration_tests/default_value/config.toml
+++ b/tests/integration_tests/default_value/config.toml
@@ -1,0 +1,21 @@
+# Importer Configuration.
+
+log-level = "info"
+
+worker-count = 10
+job-count = 1000
+batch = 10
+
+pd = "127.0.0.1:2379"
+
+[[source-db]]
+host = "127.0.0.1"
+user = "root"
+password = ""
+port = 4000
+
+[[source-db]]
+host = "127.0.0.1"
+user = "root"
+password = ""
+port = 4001
--- a/tests/integration_tests/default_value/config.toml
+++ b/tests/integration_tests/default_value/config.toml
@@ -1,5 +1,4 @@
-# Importer Configuration.
-
+# Source DB config
 log-level = "info"
 
 worker-count = 10
--- a/tests/integration_tests/default_value/main.go
+++ b/tests/integration_tests/default_value/main.go
@@ -1,4 +1,4 @@
-// Copyright 2020 PingCAP, Inc.
+// Copyright 2022 PingCAP, Inc.
 //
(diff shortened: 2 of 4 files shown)


tests/integration_tests/default_value/main.go
```diff
@@ -0,0 +1,738 @@
+// Copyright 2020 PingCAP, Inc.
```
amyangfei, line 1:
```suggestion
// Copyright 2022 PingCAP, Inc.
```

tests/integration_tests/default_value/config.toml
```diff
@@ -0,0 +1,21 @@
+# Importer Configuration.
```
amyangfei, line 1: Seems this file can be removed now.

tests/integration_tests/default_value/main.go
```diff
@@ -0,0 +1,734 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			if enabled {
+				util.MustExec(db, "set global tidb_enable_async_commit = off")
+			} else {
+				util.MustExec(db, "set global tidb_enable_async_commit = on")
+			}
```
amyangfei, line 132: Is it more explicit to set `on` when enabled is true?

amyangfei: Besides we need to check the behavior of global variable, it seems set global variable `only apply to new connection sessions with TiDB`. https://docs.pingcap.com/tidb/stable/system-variables/#system-variables

tests/integration_tests/default_value/config.toml
maxshuang: In run.sh, all these two toml files are used.

tests/integration_tests/default_value/main.go
maxshuang: Get the point. I will remove the this global vasriable, because it seems has no effect on this test.

tests/integration_tests/default_value/config.toml
amyangfei: All right, please refine the comment, it was copied from binlog repo originally, `importer configuration` is misleading

hi-rustin (approved):
It would be a good idea to move the go file to the utils folder. The rest looks good!

amyangfei (approved):

PR #4289, opened 2022-01-12
syncer/checkpoint(dm): update the meaning of IsOlderThanTablePoint

lance6716:

Issue Number: ref #4287

after https://github.com/pingcap/tiflow/pull/3891 , we can be confident to skip event <= table checkpoint when using position based replication

### Check List 

Tests 

 - pass original test

Code changes

### Release note 

```release-note
`None`.
```


Ehco1996 (approved):
LGTM

GMHDBJD (approved):
LGTM

lance6716:
/hold

found a bug in CI

lance6716:
https://github.com/pingcap/tiflow/pull/4289/commits/dd9531495cc9aa9e8701e537438f6b3d7b679652

PTAL @Ehco1996 @GMHDBJD

lance6716:
> can't understand this change, what if added2Queue = false but err2!=nil ?

it means `handleJob` meets error before send the job to DMLQueue or DDLQueue, which will not happen currently.

Ehco1996:
> > can't understand this change, what if added2Queue = false but err2!=nil ?
> 
> it means `handleJob` meets error before send the job to DMLQueue or DDLQueue, which will not happen currently.

yes, but this could not hanppen in handleRowsEvent and how this change fix the `save checkpoint while job is not added to queue` 😵. could you please explain more

lance6716:
I mean "save checkpoint while job is not added to queue" is a bug, the commit has fixed it.

dm/syncer/syncer.go
```diff
@@ -1012,23 +1012,24 @@ func (s *Syncer) handleJob(job *job) (err error) {
 
 	if waitXIDStatus(s.waitXIDJob.Load()) == waitComplete && job.tp != flush {
```
Ehco1996, line 1013: you mean in this line return nil-err and outer logic call s.checkShouldFlush() ? and this cause checkpoint have wrong pos ?

lance6716: yes.

Issue #4222, opened 2022-01-05
ticdc: Refactor the sink module


asddongmen:
### Is your feature request related to a problem?

1. The current definition of [Sink](https://github.com/pingcap/tiflow/blob/master/cdc/sink/sink.go) is blocking, which hinders the implementation of the [table actor](https://github.com/pingcap/tiflow/issues/3881). And it is not conducive to implement flow control of the Sink module in this condiction. 
2. The implemention of the current Sink module is complicated, and there are some redundant components (such as [buffer_sink](https://github.com/pingcap/tiflow/blob/master/cdc/sink/buffer_sink.go)), which are not conducive to understanding and maintenance.

### Describe the feature you'd like

1.  Make this method in Sink non-blocking:
```go
EmitRowChangedEvents(ctx context.Context, rows ...*model.RowChangedEvent)  error
```
2. Make the code related to the Sink module concise.

### Teachability, Documentation, Adoption, Migration Strategy

- [x] Modify the interface and implementation of the Sink interface. In the first step of the whole plan, [SinkNode](https://github.com/pingcap/tiflow/blob/master/cdc/processor/pipeline/sink.go) still maintains blocking semantics for the upper-level caller, and the bottom-level Sink returns false will retry in SinkNode; modify related test cases. #4299
~~- [x] Make redo flush `resolvedTs` async. #4301~~
- [x] Make mq sink  flush `resolvedTs` async. #4356
- [ ] Remove redundant `eventBuffer` in sinkNode #4452 
- [ ] Transform the related calls of SinkNode to non-blocking mode, modify the related code of the table pipeline; add related test cases.
- [x]  Remove the buffer sink and reduce code complexity.
- [x] #5138
- [ ] Add unit test cases to cover all changes.
- [ ] Reduce the number of goroutines in the Sink module.

PR #4299, opened 2022-01-12
sink (ticdc): add a non-blocking API to sink

asddongmen:

Issue Number: ref #4222 

Add a API: `TryEmitRowChangedEvents(ctx context.Context, rows ...*model.RowChangedEvent) (bool, error)` to `Sink`.

### Check List 

Tests 

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
None
```


cdc/sink/sink.go
```diff
@@ -39,6 +39,8 @@ type Sink interface {
 	// FIXME: some sink implementation, they should be.
```
overvenus, line 39:
```suggestion
	// FIXME: some sink implementation is not thread-safe, but they should be.
```

Please correct similar comments in this file, thanks!

cdc/sink/sink.go
```diff
@@ -39,6 +39,8 @@ type Sink interface {
...
 	EmitRowChangedEvents(ctx context.Context, rows ...*model.RowChangedEvent) error
 
+	// TryEmitRowChangedEvents is thread-safe and non-blocking.
```
overvenus, line 42: Please comment thread-safety of the method.

overvenus (commented):
rest LGMT

overvenus (approved):

sdojjy (approved):

PR #4354, opened 2022-01-17
syncer: add a experimental swith for async checkpoint flush

### LINKED ISSUE #4352: [DM] A switch to disable async checkpoint by default
### What did you do?

Async checkpoint now takes effect in release-5.4. We should revert relevant pull requests or add a on-off switch to control it. 

DM should use sync checkpoint in release-5.4.

### What did you see instead?

Async checkpoint now takes effect in release-5.4.

### Versions of the cluster

```console
```
```console
```
```console
```
```console
(leave TiUP or manually here)
```
```console
>
>
```

```console
```

glorv:

Issue Number: close #4352

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
None
```


dm/dm/config/task_test.go
```diff
@@ -708,6 +708,7 @@ func (t *testConfig) TestGenAndFromSubTaskConfigs(c *C) {
 		}
 	)
 
+	stCfg1.Experimental.AsyncCheckpointFlush = true
```
sunzhaoyang, line 711: 这个是默认值吗？

glorv: NO

lichunzhu: should this be false by default？

lichunzhu (approved):
Rest LGTM

dm/syncer/syncer.go
```diff
@@ -972,16 +972,21 @@ func (s *Syncer) addJob(job *job) {
...
-		return
+		return nil
 	}
 
 	jobSeq := s.getFlushSeq()
 	s.tctx.L().Info("Start to async flush current checkpoint to downstream based on flush interval", zap.Int64("job sequence", jobSeq))
-	j := newAsyncFlushJob(s.cfg.WorkerCount, jobSeq)
-	s.addJob(j)
-	s.flushCheckPointsAsync(j)
+	if s.cfg.Experimental.AsyncCheckpointFlush {
```
db-will, line 982: we need to change the log info "Start to async flush current checkpoint..."

db-will:
```
jobSeq := s.getFlushSeq() 
```
Only needed in async mode

glorv: done.

db-will (approved):

dm/dm/config/task.go
```diff
@@ -346,6 +346,11 @@ type TaskConfig struct {
 
 	// extra config when target db is TiDB
 	TiDB *TiDBExtraConfig `yaml:"tidb" toml:"tidb" json:"tidb"`
+
+	// task experimental configs
+	Experimental struct {
+		AsyncCheckpointFlush bool `yaml:"async-checkpoint-flush" toml:"async-checkpoint-flush" json:"async-checkpoint-flush"`
```
GMHDBJD, line 352: Just a disscuss: I prefer not to add an `experimental` config item, because if after GA, we need mark it as deprecated and `config export` may have two `async-checkpoint-flush` entry, which is so confused.

glorv: We can still add deprecated comment for experimental items. After this feature is GA, I think we can just remote this config and always use async flush.

glorv: /cc @lance6716

lance6716: I agree that we can always turn on async flush after GA

lance6716 (approved):

dm/dm/config/task_test.go
glorv: Weakly suggest comment in English please.

glorv:
manually cherry-picked by #4369

PR #4356, opened 2022-01-17
mq (ticdc): Make mq sink flush resolvedTs async

asddongmen:

Issue Number: ref #4222 

### Check List 

Tests 

Code changes

Side effects

Related changes

### Release note 

```release-note
None
```


cdc/sink/mq.go
```diff
@@ -39,11 +40,17 @@ import (
...
+
+type resolvedTsEvent struct {
+	tableID    model.TableID
+	resolvedTs model.Ts
 }
 
 const (
-	defaultPartitionInputChSize = 12800
+	defaultPartitionInputChSize      = 12800
+	defaultResolvedTsEventBufferSize = 12800
```
overvenus, line 53: Can we reduce the channel size?

cdc/sink/mq.go
```diff
@@ -164,19 +173,56 @@ func (k *mqSink) EmitRowChangedEvents(ctx context.Context, rows ...*model.RowCha
...
+	k.statistics.PrintStatus(ctx)
+	return checkpointTs, nil
+}
 
+// bgFlushTs flush resolvedTs to workers and flush the mqProducer
+func (k *mqSink) bgFlushTs(ctx context.Context) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
overvenus, line 203:
```suggestion
			return errors.Trace(ctx.Err())
```

cdc/sink/mq.go
```diff
@@ -185,23 +231,16 @@ flushLoop:
 	for {
 		select {
 		case <-ctx.Done():
-			return 0, ctx.Err()
+			return ctx.Err()
```
overvenus, line 234:
```suggestion
			return errors.Trace(ctx.Err())
```

cdc/sink/mq.go
```diff
@@ -185,23 +231,16 @@ flushLoop:
...
 		case <-k.resolvedReceiver.C:
```
overvenus, line 235: How often the tick `C` is?

cdc/sink/mq.go
```diff
@@ -164,19 +173,56 @@ func (k *mqSink) EmitRowChangedEvents(ctx context.Context, rows ...*model.RowCha
...
+			resolvedTs := msg.resolvedTs
+			err := k.flushTsToWorker(ctx, resolvedTs)
+			if err != nil {
+				return errors.Trace(err)
+			}
+			err = k.mqProducer.Flush(ctx)
+			if err != nil {
+				return errors.Trace(err)
+			}
+			k.tableCheckpointTsMap.Store(msg.tableID, resolvedTs)
```
overvenus, line 214: What if a table is removed between line 213 and 214?

cdc/sink/mq.go
```diff
@@ -164,19 +173,56 @@ func (k *mqSink) EmitRowChangedEvents(ctx context.Context, rows ...*model.RowCha
...
 	}
+	select {
+	case <-ctx.Done():
+		return 0, ctx.Err()
+	case k.resolvedBuffer <- resolvedTsEvent{
+		tableID:    tableID,
+		resolvedTs: resolvedTs,
+	}:
+	}
+	k.statistics.PrintStatus(ctx)
```
overvenus, line 194: How often does it print?

asddongmen: 10 minutes

sdojjy (approved):

cdc/sink/mq.go
asddongmen: Every time a `mqSikn.worker` flushes the producer, we will receive a `C` here.

cdc/sink/mq.go
asddongmen: done

cdc/sink/mq.go
asddongmen: Since CDC does not guarantee exactly onece semantic, it won't cause any poblem here.

Let say `capture1` has got a table name `t1` and `t1.resolvedTs = 10`, `t1.checkpointTs = 5`. If `t1` was moved to `capture2` here, then `t1` will be replicated from `t1.checkPointTs = 5` in `capture2` so that the `data [5, 10]` would be sent twice, but no data will be lost.

cdc/sink/mq.go
```diff
@@ -164,19 +177,59 @@ func (k *mqSink) EmitRowChangedEvents(ctx context.Context, rows ...*model.RowCha
...
+			if err != nil {
+				return errors.Trace(err)
+			}
+			err = k.mqProducer.Flush(ctx)
+			if err != nil {
+				return errors.Trace(err)
+			}
+			// Since CDC does not guarantee exactly once semantic, it won't cause any problem
+			// here even if the table was moved or removed.
+			// ref: https://github.com/pingcap/tiflow/pull/4356#:~:text=What%20if%20a%20table%20is%20removed%20between%20line%20213%20and%20214%3F
```
overvenus, line 220:
```suggestion
			// ref: https://github.com/pingcap/tiflow/pull/4356#discussion_r787405134
```

Let's shorten the link.

cdc/sink/mysql.go
```diff
@@ -224,6 +224,9 @@ func (s *mysqlSink) EmitRowChangedEvents(ctx context.Context, rows ...*model.Row
 // FlushRowChangedEvents will flush all received events, we don't allow mysql
 // sink to receive events before resolving
 func (s *mysqlSink) FlushRowChangedEvents(ctx context.Context, tableID model.TableID, resolvedTs uint64) (uint64, error) {
+	// Since CDC does not guarantee exactly once semantic, it won't cause any problem
+	// here even if the table was moved or removed.
+	// ref: https://github.com/pingcap/tiflow/pull/4356#:~:text=What%20if%20a%20table%20is%20removed%20between%20line%20213%20and%20214%3F
```
overvenus, line 229:
```suggestion
	// ref: https://github.com/pingcap/tiflow/pull/4356#discussion_r787405134
```

overvenus (approved):
LGTM

Issue #2164, opened 2021-06-25
Tracking issue for replace github.com/pingcap/check with github.com/stretchr/testify


ben1009:
## Background

as described in the thread https://internals.tidb.io/t/topic/141/6

```
I did ever use testify in pingcap/errors [[1]](https://github.com/pingcap/errors/pull/36) and it works well with richer matchers and default parallelism support.

I think the modification from gocheck to pingcap/check mainly focuses on parallelism. It is possibly we switch to testify to live with a stronger community instead of a lack of maintenance version of a smaller community.

* [testify ](https://github.com/stretchr/testify) has 188 contributors and latest updated at Apr. 27
* [go-check](https://github.com/go-check/check) has 13 contributors and latest updated at last year.

If you think this direction is good to go, maybe we can draft a pre-RFC to collect wider input.
```

This is somewhat a subtask of https://github.com/pingcap/tidb/issues/26022 that we make a decision that deprecating [pingcap/check](https://github.com/pingcap/check) and adopting [testify](https://github.com/stretchr/testify). Take a look at the tidb issue for more information.

After an offline discussion with @overvenus we think it is fine we make the same change on this repo and here is the tracking issue for the effort.

If you have any further concern, please comment below and reply will be in days.

We first write down a task list and since I have no time to continue the actual work, let's kick off when there is a contributor ask for starting the subtasks.

## Join Forces

If you want to participant the movement, you can comment under this issue to acquire pkg or files to migrate. For example,

```
Hi, I'd like to migrate tests under `errno` pkg. Please create a subtask and assign to me.
```

React will be in time and do the favor for you.

Also you can directly comment under any open issues below without assignee to require assignment.

You can take a look at https://github.com/pingcap/tidb/pull/26375 as an example to do the migration.

## Note
* new unit-test

    - New uts should use testify and goleak directly. Ref to the [Pr 2833](https://github.com/pingcap/ticdc/pull/2833)
    - Pkg github.com/pingcap/tidb/util/testkit should be replaced by github.com/pingcap/tidb/testkit

* parallel

    - Keep test cases as parallel as possible.
    - Don't use t.Parallel() if pkg has failpoint control. Need review Carefully.
    - When some test cases should be run in serial, take use of go testing subtests (t.Run(name string, f func(t *T))) and parallel the parent test only. In this way, test cases in the same subtests set run in serial.
[...]

PR #4358, opened 2022-01-17
dm: new router compatible with regular expression

### LINKED ISSUE #4256: dm table-routing need support regular expressions
demand:
```
db list: test
table list:
t_1
t_2
t_11
t_12
```
now,table routing need write:
```
routes:
  r1:
    s-p: 'test'
    t-p: 't_[0-9]'
    t-s: 'test'
    t-t: 't'
  r2:
    s-p: 'test'
    t-p: 't_[0-9][0-9]'
    t-s: 'test'
    t-t: 't'
  r3:
    t-s: 'test'
    s-p: 'test'
```

i wish:

```
routes:
  r1:
    s-p: 'test'
    t-p: '~t_*'
    t-s: 'test'
    t-t: 't'
```

include version:
v2.x
v1.x
v5.3.0

buchuitoudegou:
implement new router that is compatible with regular expressions

Issue Number: close #4256 

A new router that is compatible with regular expressions is implemented. It basically reused the `filter` module in `tidb-tools` project, which has also been used in the `block-allow-list`. Reusing the existing module ensures modules behaving the same across the whole project.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - New router implemented as a stand-alone module and exporting the same interface with the old one
 - Replace the old router with the new one

Side effects

Related changes

### Release note 

```release-note
Implement a new router that is compatible with regular expression as well as the old router
```


buchuitoudegou:
https://prow.tidb.io/command-help#retest

buchuitoudegou:
/retitle new router implementation compatible with regular expression

dm/regexprrouter/router.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	var (
+		targetSchema string
+		targetTable  string
+	)
+	if len(table) == 0 || len(tblRules) == 0 {
+		// 1. no need to match table or
+		// 2. match no table
+		if len(schmRules) > 1 {
+			return "", "", errors.NotSupportedf("`%s`.`%s` matches %d schema route rules which is more than one.\n", schema, table, len(schmRules))
```
lance6716, line 144: we should use terror package

dm/regexprrouter/router.go
```diff
@@ -0,0 +1,278 @@
...
+		targetTable = table
+	}
+	return targetSchema, targetTable, nil
+}
+
+func (r *RegExprTable) AllRules() (map[string][]interface{}, map[string]map[string][]interface{}) {
+	schemaRouteRules := make(map[string][]interface{})
+	tableRouteRules := make(map[string]map[string][]interface{})
+	for _, filter := range r.filters {
+		curRule := displayRule{
```
lance6716, line 168: seems this is only used for debug? You can use json dump for `filter.rawRule`, no need to add a new structure

buchuitoudegou: Or can I just return two arrays representing the table rules and the schema rules respectively?

dm/regexprrouter/router.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func extractVal(s string, ext interface{}) string {
+	var params []string
+	switch e := ext.(type) {
+	case *oldrouter.TableExtractor:
+		if regExpr, err := regexp.Compile(e.TableRegexp); err == nil {
+			params = regExpr.FindStringSubmatch(s)
+		} else {
+			return ""
```
D3Hunter, line 227:
```suggestion
```

followings too.

dm/regexprrouter/router_test.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	r, err = NewRegExprRouter(false, []*TableRule{})
+	assert.Equal(t, nil, err)
+	for _, rule := range rules {
+		err := r.AddRule(rule)
+		assert.Equal(t, nil, err) // successfully insert
+	}
+}
+
+func TestSchemaRoute(t *testing.T) {
```
D3Hunter, line 58: maybe add some case for ErrWorkerRouteTableDupMatch

dm/regexprrouter/router_test.go
```diff
@@ -0,0 +1,243 @@
...
+	assert.Equal(t, nil, err)
+	for idx := range inputTable {
+		s, n := inputTable[idx].Schema, inputTable[idx].Name
+		expSchm, expName := expectedOutput[idx].Schema, expectedOutput[idx].Name
+		newSchm, newName, err := newRouter.Route(s, n)
+		assert.Equal(t, nil, err)
+		assert.Equal(t, expSchm, newSchm)
+		assert.Equal(t, expName, newName)
+	}
+}
```
D3Hunter, line 243: maybe add some case for `AllRules` and `FetchExtendColumn`

D3Hunter (commented):
maybe we can enhance router in `tidb-tools` directly.

dm/loader/convert_data.go
```diff
@@ -26,10 +26,10 @@ import (
 	parserpkg "github.com/pingcap/tiflow/dm/pkg/parser"
 	"github.com/pingcap/tiflow/dm/pkg/terror"
 	"github.com/pingcap/tiflow/dm/pkg/utils"
+	"github.com/pingcap/tiflow/dm/regexprrouter"
```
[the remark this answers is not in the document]
buchuitoudegou: checked

dm/regexprrouter/router.go
```diff
@@ -0,0 +1,249 @@
...
+package regexprrouter
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/tidb-tools/pkg/filter"
+	oldrouter "github.com/pingcap/tidb-tools/pkg/table-router"
+	"github.com/pingcap/tiflow/dm/pkg/terror"
```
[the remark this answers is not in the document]
buchuitoudegou: checked

dm/regexprrouter/router_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	router "github.com/pingcap/tidb-tools/pkg/table-router"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCreateRouter(t *testing.T) {
```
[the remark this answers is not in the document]
buchuitoudegou: checked

dm/regexprrouter/router.go
```diff
@@ -0,0 +1,249 @@
...
+	TableRule  = oldrouter.TableRule
+	Table      = filter.Table
+	FilterRule = filter.Rules
+)
+
+type FilterType = int32
+
+const (
+	TblFilter  int32 = 1
+	SchmFilter int32 = 2
```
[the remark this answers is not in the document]
buchuitoudegou: checked

dm/regexprrouter/router.go
buchuitoudegou: checked

dm/regexprrouter/router_test.go
buchuitoudegou: checked

dm/regexprrouter/router_test.go
buchuitoudegou: checked

dm/dm/config/subtask.go
```diff
@@ -25,12 +25,12 @@ import (
 	bf "github.com/pingcap/tidb-tools/pkg/binlog-filter"
 	"github.com/pingcap/tidb-tools/pkg/column-mapping"
 	"github.com/pingcap/tidb-tools/pkg/filter"
-	router "github.com/pingcap/tidb-tools/pkg/table-router"
```
Ehco1996, line 28: just a question: why not pull a request to tidb-tools?
maybe i miss some offline talk infomation😂

dm/tests/all_mode/run.sh
```diff
@@ -532,6 +583,10 @@ function run() {
 }
 
 cleanup_data all_mode
+cleanup_data test3gd8917777
```
Ehco1996, line 586: maybe those cleanup can move to test_regexpr_router

dm/tests/all_mode/run.sh
```diff
@@ -320,6 +320,56 @@ function test_expression_filter() {
...
+	sed -i 's/tidb_retry_limit: "fjs"/tidb_retry_limit: "10"/g' $WORK_DIR/regexpr-task.yaml
+	dmctl_start_task "$WORK_DIR/regexpr-task.yaml" "--remove-meta"
+
+	check_sync_diff $WORK_DIR $cur/conf/regexpr_diff_config.toml
+
+	cleanup_data test3gd8917777
+	cleanup_data test4s_2022
+	cleanup_data_upstream test3gd8917777
+	cleanup_data_upstream test4s_2022
+	cleanup_process
```
Ehco1996, line 369: i think call cleanup_data(log) before cleanup_process, because if your process is hung, you can not see logs any more

dm/pkg/router/router_test.go
```diff
@@ -0,0 +1,387 @@
...
+	"fmt"
+	"testing"
+
+	. "github.com/pingcap/check"
+	"github.com/pingcap/tiflow/dm/pkg/terror"
+
+	oldrouter "github.com/pingcap/tidb-tools/pkg/table-router"
+)
+
+var _ = Suite(&testRouteSuite{})
```
Ehco1996, line 26: light suggestion, new added test better use testify instead see https://github.com/pingcap/tiflow/issues/2164

dm/tests/all_mode/run.sh
buchuitoudegou: Would u mind getting me some hint about the `cleanup_data(log)` since I couldn't find such a function. In addition, cleaning the data and then killing the processes is strictly followed the scheme of other tests in that script. Could you describe it in more detail cuz I've never met the situation that I couldn't see the logs. Thx in advance.

buchuitoudegou: p.s. even the some tests fail, the logs remain in the /tmp/dm_test without being broken.

Ehco1996: cleanup_data is [here](https://github.com/pingcap/tiflow/blob/239a21fe146a2b2950dad06e7d4e201791d4d030/dm/tests/_utils/test_prepare#L4)

>  is strictly followed the scheme of other tests 

yes i used to write test like this unitl i found if dm-worker never exit(some bug code with my local dev branch), i can't see the log of dm-worker because whole test dir is deleted. so if there is **no bug** in our code, call cleanup_data before kill process is ok 😂

buchuitoudegou: Yes, deleting the data before killing the processes will make the error logs invisible if the error occurs when exiting the processes😂. But I don't think that is the case we need to test in the integration test😂

Ehco1996: we need not test this case, i just say call cleanup_data after kill process is good for debug 😬

GMHDBJD (approved):
LGTM
Please resolve conflict

dm/pkg/router/router.go
```diff
@@ -0,0 +1,247 @@
...
+			rules = append(rules, filter)
+		}
+	}
+	var (
+		schemaRules = make([]*TableRule, 0, len(rules))
+		tableRules  = make([]*TableRule, 0, len(rules))
+	)
+	for i := range rules {
+		rule := rules[i].rawRule
+		if len(rule.TablePattern) == 0 {
```
lance6716, line 192: ditto

dm/pkg/router/router_test.go
```diff
@@ -0,0 +1,387 @@
...
+		c.Assert(newTbl, Equals, expTable)
+		c.Assert(oldSch, Equals, expSchema)
+		c.Assert(oldTbl, Equals, expTable)
+	}
+}
+
+func (s *testRouteSuite) TestRegExprRoute(c *C) {
+	rules := []*TableRule{
+		{
+			SchemaPattern: "~test.[0-9]+",
```
lance6716, line 177: can you give a meaningful string, it's very hard to check the result :cry:

dm/pkg/router/router_test.go
```diff
@@ -0,0 +1,387 @@
...
+			},
+			SourceExtractor: &SourceExtractor{
+				TargetColumn: "source_name",
+				SourceRegexp: "source_(.*)_(.*)",
+			},
+		},
+		{
+			SchemaPattern: "~s?chema.*",
+			TargetSchema:  "test",
+			TargetTable:   "t2",
```
lance6716, line 278: this usage (no table-pattern but target-table) is strange, should we forbid it in future? @GMHDBJD @lichunzhu

dm/pkg/router/router_test.go
```diff
@@ -0,0 +1,387 @@
...
+	for i := 0; i < 3; i++ {
+		c.Assert(tableRules[i].SchemaPattern, Equals, rules[i+1].SchemaPattern)
+		c.Assert(tableRules[i].TablePattern, Equals, rules[i+1].TablePattern)
+	}
+}
+
+func (s *testRouteSuite) TestDupMatch(c *C) {
+	rules := []*TableRule{
+		{
+			SchemaPattern: "~test[0-9]+.*",
```
lance6716, line 350: also meaningful strings

dm/tests/all_mode/run.sh
```diff
@@ -320,6 +320,61 @@ function test_expression_filter() {
 	echo "[$(date)] <<<<<< finish test_expression_filter >>>>>>"
 }
 
+function test_regexpr_router() {
+	echo "[$(date)] <<<<<< start test_regexpr_router >>>>>>"
+	cleanup_process
+	cleanup_data all_mode
+	cleanup_data test3gd8917777
```
lance6716, line 327: ditto

dm/tests/all_mode/run.sh
```diff
@@ -531,7 +587,6 @@ function run() {
 	run_sql_both_source "SET @@global.time_zone = 'SYSTEM';"
 }
 
-cleanup_data all_mode
```
lance6716, line 534: this is still needed, because your newly added `test_regexpr_router` is not the first case.

lance6716 (commented):
rest lgtm

dm/pkg/router/router_test.go
lichunzhu: IMO we should forbid this kind of behavior

dm/pkg/router/router.go
```diff
@@ -0,0 +1,247 @@
...
+			} else {
+				schmRules = append(schmRules, filterWrapper)
+			}
+		}
+	}
+	var (
+		targetSchema string
+		targetTable  string
+	)
+	if len(table) == 0 || len(tblRules) == 0 {
```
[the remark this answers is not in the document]
buchuitoudegou: `len(table) == 0` is checking if there is no `table` input but merely routing the schema. I forgot if anyone uses the router in this way in DM but it is implemented in `tidb-tools`.
Maybe checking the string by comparing it with `""` is more intuitive.

dm/pkg/router/router.go
```diff
@@ -0,0 +1,247 @@
...
+			return "", "", terror.ErrWorkerRouteTableDupMatch.Generate(schema, table)
+		}
+		if len(schmRules) == 1 {
+			targetSchema, targetTable = schmRules[0].Target.Schema, schmRules[0].Target.Name
+		}
+	} else {
+		if len(tblRules) > 1 {
+			return "", "", terror.ErrWorkerRouteTableDupMatch.Generate(schema, table)
+		}
+		targetSchema, targetTable = tblRules[0].Target.Schema, tblRules[0].Target.Name
```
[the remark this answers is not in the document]
buchuitoudegou: Yes. This is an error case.

dm/pkg/router/router_test.go
buchuitoudegou: What if the users would like to migrate all tables from one or more designated schema(s) to exactly **ONE** table in downstream?

dm/pkg/router/router_test.go
buchuitoudegou: Would you explain more about the hard reading part?

dm/pkg/router/router_test.go
GMHDBJD: They should use TablePattern: "*"

dm/pkg/router/router_test.go
lance6716: I need to look back to `~test3(gd|ct)?8917*` to check `test3gd8917777` is matched as expected (does it have same four digit?) and check other regexp is not matced as expected (like `~test.[0-9]+`), a lot of back and forth jumping.

if they can be `~test1[0-9]*`, `~test2[0-9]*`, `~test[a-z][0-9]*` or other pattern it would be easier to remember.

buchuitoudegou: Yes, it's a bit complicated. Actually I'm trying to build some similar but exclusive patterns and see if they conflicts with the rule that only one routing rule is matched. It's perhaps unnecessarily complex😂. I'll fix it later...

dm/pkg/router/router_test.go
buchuitoudegou: But for now, missing field `TablePattern` is permitted. Would it be troublesome if we forbid it, considering some users may need to modify their old configs?

lance6716:
ping @buchuitoudegou

lance6716 (approved):

PR #4414, opened 2022-01-19
*(dm): recycle resources when initialization of checkpoint in syncer, loader, shard-group-keeper failed

### LINKED ISSUE #4353: *(dm): db not closed when init failed
### description

I found the same problem in more places same as https://github.com/pingcap/tiflow/issues/4142

https://github.com/pingcap/tiflow/blob/4a06fe3104fc5ee37b4328e8f6a8c2cbfdd40199/dm/loader/checkpoint.go#L103-L126

https://github.com/pingcap/tiflow/blob/4a06fe3104fc5ee37b4328e8f6a8c2cbfdd40199/dm/syncer/checkpoint.go#L413-L425

https://github.com/pingcap/tiflow/blob/4a06fe3104fc5ee37b4328e8f6a8c2cbfdd40199/dm/syncer/sharding_group.go#L429-L441

didn't close db once err return

niubell:

Issue Number: close #4353 

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

 - None

Related changes

### Release note 

```release-note
None.
```


okJiang (approved):

dm/loader/checkpoint.go
```diff
@@ -101,7 +102,18 @@ type RemoteCheckPoint struct {
...
+	rollbackHolder := fr.NewRollbackHolder("loader")
+	defer func() {
+		if err != nil {
+			rollbackHolder.RollbackReverseOrder()
+		}
+	}()
+
+	db, dbConns, err = createConns(tctx, cfg, cfg.Name, cfg.SourceID, 1)
 	if err != nil {
 		return nil, err
```
Ehco1996, line 118: seems if err != nil  we have not add add rollbcak func into`rollbackHolder` yes , so there is no rollbak in defer🤔, can we add a ut to test the rowback func really works?

dm/loader/checkpoint.go
```diff
@@ -116,6 +128,7 @@ func newRemoteCheckPoint(tctx *tcontext.Context, cfg *config.SubTaskConfig, id s
 		logger:         tctx.L().WithFields(zap.String("component", "remote checkpoint")),
 	}
 	cp.restoringFiles.pos = make(map[string]map[string]FilePosSet)
+	rollbackHolder.Add(fr.FuncRollback{Name: "CloseRemoteCheckPoint", Fn: cp.Close})
 
 	err = cp.prepare(tctx)
```
Ehco1996, line 133: how about close DB in func `cp.prepre`, so we need't and rollback func anymore

niubell: A uniform rollback can handle all problems like this in the future, i had done it with close in prepare, but i think rollback is better.

dm/loader/checkpoint.go
niubell: > ~seems if err != nil we have not add add rollbcak func into`rollbackHolder` yes , so there is no rollbak in defer🤔, can we add a ut to test the rowback func really works?~
> 
> oh i see it's trying to close db when cp.prepare meet error

bingo

niubell:
I have always tried to sign CLA, but always in vain because of unauthorized email address, i will submit another pull request with a new email address.

Ehco1996 (approved):
LGTM

lance6716 (approved):

PR #3845, opened 2021-12-13
ticdc/owner: Fix ddl special comment syntax error

sdojjy:
### What problem does this PR solve? 
close https://github.com/pingcap/ticdc/issues/3755

### What is changed and how it works?
not use regex to parse and add special TiDB comments,  If a DDL already has special comment like `/*!90000  SHARD_ROW_ID_BITS=4 PRE_SPLIT_REGIONS=3 */`, it wil translate that DDL to `/*!90000 /*T! SHARD_ROW_ID_BITS=4 PRE_SPLIT_REGIONS=3 */ */`, TiDB will throw a syntax error.
this PR use  the TiDB SQL parser to parse the SQL and the Restore it with `format.RestoreStringSingleQuotes|format.RestoreNameBackQuotes|format.RestoreKeyWordUppercase|format.RestoreTiDBSpecialComment` flag

### Check List 

Tests 

 - Unit test

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to update key monitor metrics in both TiCDC document and official document

### Release note 

```release-note
Fix syntax error if DDL has a special comment.
```

PR #4415, opened 2022-01-19
ticdc/owner: Fix ddl special comment syntax error (#3845)

sdojjy:
This is an automated cherry-pick of #3845

close https://github.com/pingcap/ticdc/issues/3755

not use regex to parse and add special TiDB comments,  If a DDL already has special comment like `/*!90000  SHARD_ROW_ID_BITS=4 PRE_SPLIT_REGIONS=3 */`, it wil translate that DDL to `/*!90000 /*T! SHARD_ROW_ID_BITS=4 PRE_SPLIT_REGIONS=3 */ */`, TiDB will throw a syntax error.
this PR use  the TiDB SQL parser to parse the SQL and the Restore it with `format.RestoreStringSingleQuotes|format.RestoreNameBackQuotes|format.RestoreKeyWordUppercase|format.RestoreTiDBSpecialComment` flag

### Check List 

Tests 

 - Unit test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Fix syntax error if DDL has a special comment.
```


amyangfei (approved):

cdc/owner/changefeed.go
```diff
@@ -15,13 +15,15 @@ package owner
...
+	"strings"
 	"sync"
 
 	"github.com/pingcap/errors"
 	"github.com/pingcap/failpoint"
 	"github.com/pingcap/log"
 	timodel "github.com/pingcap/parser/model"
-	"github.com/pingcap/tidb/sessionctx/binloginfo"
+	"github.com/pingcap/tidb/parser"
+	"github.com/pingcap/tidb/parser/format"
```
overvenus, line 26: Can we use pingcap/parser?

overvenus:
/hold

Since pingcap/parser does not support `RestoreTiDBSpecialComment` yet, we should be careful for using tidb/parser, as it may have side effect `init()`.

PR #4479, opened 2022-01-26
validation(dm): manage validator life cycle

### LINKED ISSUE #4459: manage validation life cycle
validator can be start when there's `syncer` unit in the subtask and validation mode is not none, it's terminated when the subtask is terminated.

validator can be in `running` or `stopped` stage
- in `running` when it's started with subtask or started later on the fly
- in `stopped` when `validation stop` is executed

D3Hunter:

Issue Number: close #4459

see issue for validator life cycle description

validator is empty in this pr. `dmctl` command to start/stop validator will be add in later pr

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`
```


dm/dm/config/task.go
```diff
@@ -377,9 +406,10 @@ type TaskConfig struct {
 	BWList map[string]*filter.Rules `yaml:"black-white-list" toml:"black-white-list" json:"black-white-list"`
 	BAList map[string]*filter.Rules `yaml:"block-allow-list" toml:"block-allow-list" json:"block-allow-list"`
 
-	Mydumpers map[string]*MydumperConfig `yaml:"mydumpers" toml:"mydumpers" json:"mydumpers"`
-	Loaders   map[string]*LoaderConfig   `yaml:"loaders" toml:"loaders" json:"loaders"`
-	Syncers   map[string]*SyncerConfig   `yaml:"syncers" toml:"syncers" json:"syncers"`
+	Mydumpers  map[string]*MydumperConfig  `yaml:"mydumpers" toml:"mydumpers" json:"mydumpers"`
+	Loaders    map[string]*LoaderConfig    `yaml:"loaders" toml:"loaders" json:"loaders"`
+	Syncers    map[string]*SyncerConfig    `yaml:"syncers" toml:"syncers" json:"syncers"`
+	Validators map[string]*ValidatorConfig `yaml:"validators" toml:"validators" json:"validators"`
```
Ehco1996, line 412: why we need different `Validators` for different source ? will user want it ? e.g: one souce use `full` but another use `fast` ？

dm/dm/worker/subtask.go
```diff
@@ -733,6 +774,28 @@ func (st *SubTask) HandleError(ctx context.Context, req *pb.HandleWorkerErrorReq
 	return msg, err
 }
 
+func (st *SubTask) getCfg() *config.SubTaskConfig {
+	st.RLock()
+	defer st.RUnlock()
+	return st.cfg
+}
+
+func (st *SubTask) SetCfg(taskConfig config.SubTaskConfig) {
```
Ehco1996, line 783:
```suggestion
func (st *SubTask) SetCfg(subTaskConfig config.SubTaskConfig) {
```

dm/dm/worker/subtask.go
```diff
@@ -257,6 +261,38 @@ func (st *SubTask) run() {
 	go cu.Process(ctx, pr)
 }
 
+func (st *SubTask) StartValidator(expect pb.Stage) {
+	st.Lock()
+	defer st.Unlock()
+	if st.cfg.ValidatorCfg.Mode != config.ValidationFast && st.cfg.ValidatorCfg.Mode != config.ValidationFull {
+		return
```
Ehco1996, line 268: seems this line should not happen, becasue we always adjust the subtaskConfig

dm/dm/worker/source_worker.go
```diff
@@ -787,7 +800,9 @@ func (w *SourceWorker) operateSubTaskStage(stage ha.Stage, subTaskCfg config.Sub
 		if st := w.subTaskHolder.findSubTask(stage.Task); st == nil {
 			// create the subtask for expected running and paused stage.
 			log.L().Info("start to create subtask", zap.String("sourceID", subTaskCfg.SourceID), zap.String("task", subTaskCfg.Name))
-			err := w.StartSubTask(&subTaskCfg, stage.Expect, true)
+			// new subtask, pass Running to validator stage, let internal logic determines it.
+			// as there's no new subtask with a stopped validator, either it has a Running one or none
+			err := w.StartSubTask(&subTaskCfg, stage.Expect, pb.Stage_Running, true)
```
Ehco1996, line 805: always `StartSubTask` with  expected running validator is confused me, if user config validator in task config with all mode, will the validator start automatically when task into incr sync satge ?

but what happened if user start task with full mode?

dm/dm/config/task.go
Ehco1996: btw i perfer to not put this config into subtask config, this two things can be unnecessarily coupled to each other 

seems user can just use ( dmctl validator start/stop --task ---source --mode) to operate validator ?

buchuitoudegou: According to the current design, the user can start a validator when starting the task and thus the validator config is coupled with the task's. 
Referring to the spec, they can specify different validators for different sources. It might possibly be more accessible when they prefer a `full` validator working on small database while a `fast` validator working on a big database.

dm/dm/worker/subtask.go
buchuitoudegou: Adjusting the `subtaskConfig` seems insufficient to guarantee the `mode` is either `fast` or `full` since there are three possible values for it: `none`, `fast`, and `full`, where `none` is the default value (when this field is missing) and indicates the validator should not start.

dm/dm/worker/source_worker.go
buchuitoudegou: It does look confusing. But the validator won't start even passing the `pb.Stage_Running` flag as your previous comment points out...Every time starting the validator, the validator's `mode` will be checked first and forced to return if it's neither `full` and `fast`.
In other words, the validator can still be started in `all` mode migration, where binlog should not be migrated, as long as the mode is set to `full` or `fast`.

buchuitoudegou: But I am also confused by this parameter. It seems always `Running`.

dm/dm/config/task.go
D3Hunter: > why we need different Validators for different source ? will user want it ?

may not, then just use the same validation config. right now there's only one param in `ValidatorConfig`, may add more later, user may want different parameter for different source, just like other units

> btw i perfer to not put this config into subtask config

validation is a `subtask` level unit just like `syncer`, it can only run when there's a `syncer` unit in this subtask actually.

dm/dm/worker/subtask.go
D3Hunter: > seems this line should not happen, becasue we always adjust the subtaskConfig

`st.cfg.ValidatorCfg.Mode` may equals to `none`(i.e. validation is disabled)

dm/dm/worker/source_worker.go
D3Hunter: i explained it in the comment, it's ok to set expect `stage=running` for new subtask, but you're right, it's not explicit and confusing.

dm/dm/worker/source_worker.go
```diff
@@ -1097,3 +1121,146 @@ func (w *SourceWorker) HandleError(ctx context.Context, req *pb.HandleWorkerErro
...
+
+		if etcdutil.IsRetryableError(err) {
+			startRevision = 0
+			retryNum := 1
+			for startRevision == 0 {
+				select {
+				case <-ctx.Done():
+					return nil
+				case <-time.After(500 * time.Millisecond):
+					startRevision, err = w.getCurrentValidatorRevision(w.cfg.SourceID)
```
lance6716, line 1154: we should also reset the validator stage to ones in etcd 🤔 example:

there are three historical stages: (running, rev1), (stopped, rev2), (running, rev3)
At first, the validator starts with subtask, it gets (running, rev1) and watch from rev1. Then it gets (stopped, rev2) and stops validator. At this time etcd compacts and cause CompactError, we enter `if etcdutil.IsRetryableError(err)` and now `getCurrentValidatorRevision` returns rev3, we start watch from rev3 but no KV changes now.

Now the stage in etcd is running but we haven't started validator

D3Hunter: didn't add 1 to revision in the loop so we can get the last one, see:

https://github.com/D3Hunter/tiflow/blob/47f24bbbc7252b9ff772ddab5129393fca0102a6/dm/dm/worker/source_worker.go#L1140

dm/dm/worker/subtask.go
```diff
@@ -515,7 +559,7 @@ func (st *SubTask) Pause() error {
 // TODO: similar to Run, refactor later.
 func (st *SubTask) Resume(relay relay.Process) error {
 	if !st.initialized.Load() {
-		st.Run(pb.Stage_Running, relay)
+		st.Run(pb.Stage_Running, pb.Stage_Running, relay)
```
lance6716, line 562: not sure user want to start validator when resume task @sunzhaoyang

lance6716 (commented):
will review later

D3Hunter:
this case timeout often, but runs ok locally.
```
✖  dm/dm/master (5m0.511s)
```

dm/dm/worker/subtask.go
D3Hunter: changed to get validator stage from etcd.

D3Hunter:
@lance6716 @Ehco1996 ptal

dm/dm/config/task.go
Ehco1996: i found a user sccenario: user want start validation as soon as the sync phase is entered, so this lgtm

dm/syncer/data_validator.go
```diff
@@ -0,0 +1,192 @@
+// Copyright 2021 PingCAP, Inc.
```
Ehco1996, line 1:
```suggestion
// Copyright 2022 PingCAP, Inc.
```

dm/dm/proto/dmworker.proto
```diff
@@ -383,4 +383,10 @@ message GetWorkerCfgRequest {
 
 message GetWorkerCfgResponse {
     string cfg = 1;
-}
\ No newline at end of file
+}
+
+enum ValidatorOp {
+    InvalidValidatorOp = 0;
```
Ehco1996, line 389: beacuse we want to use openapi reimpl dmctl , how about not add new message to this pb? It's too painful to maintain two copies of the same thing at the same time 😷

Ehco1996 (commented):
rest LGTM

dm/dm/proto/dmworker.proto
D3Hunter: when will `openapi reimpl dmctl` GA? 

there're a lot of dmctl command related to validation in later pr, do you mean new commands should be written with `openapi`?

Ehco1996: plan to GA in the next version v6.0, I hope that the new ctl commands will be implemented via openapi if possible, so that there will be less duplication of work later. but this is just weak a suggestion

D3Hunter: cannot find any commands using openapi, maybe we can do it later, together with other commands.

Ehco1996 (approved):
LGTM

dm/dm/worker/source_worker.go
```diff
@@ -455,7 +456,7 @@ func (w *SourceWorker) EnableHandleSubtasks() error {
 
 	// we get the newest subtask stages directly which will omit the subtask stage PUT/DELETE event
 	// because triggering these events is useless now
-	subTaskStages, subTaskCfgM, revSubTask, err := w.fetchSubTasksAndAdjust()
+	subTaskStages, validatorStages, subTaskCfgM, revSubTask, err := w.fetchSubTasksAndAdjust()
```
lance6716, line 459: I found this is the only usage of `validatorStages` from `w.fetchSubTasksAndAdjust()`, meanwhile we have called `getExpectValidatorStage` at two place. Can we only use `getExpectValidatorStage` and not modify `w.fetchSubTasksAndAdjust()` (for example `w.StartSubTask` will call `getExpectValidatorStage` inside)?

dm/pkg/ha/stage.go
```diff
@@ -172,25 +176,34 @@ func GetAllRelayStage(cli *clientv3.Client) (map[string]Stage, int64, error) {
...
 	ctx, cancel := context.WithTimeout(cli.Ctx(), etcdutil.DefaultRequestTimeout)
 	defer cancel()
 
 	var (
 		stm  = make(map[string]Stage)
 		resp *clientv3.GetResponse
 		err  error
+		opts = make([]clientv3.OpOption, 0)
 	)
+	if revision > 0 {
```
lance6716, line 192: can we always use the latest revision? in future we might decouple from etcd, and revision may not be supported in new metadata storage.

dm/syncer/data_validator.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/pingcap/tiflow/dm/dm/unit"
+	"github.com/pingcap/tiflow/dm/pkg/conn"
+	tcontext "github.com/pingcap/tiflow/dm/pkg/context"
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"github.com/pingcap/tiflow/dm/pkg/terror"
+	"github.com/pingcap/tiflow/dm/pkg/utils"
+	"github.com/pingcap/tiflow/dm/syncer/dbconn"
+)
+
+type DataValidator struct {
```
lance6716, line 35: can you left a comment for it? like its lifetime with syncer, main responsibility, interaction with caller

lance6716 (commented):
rest lgtm

dm/pkg/ha/stage.go
D3Hunter: in `operateSubTaskStage`, we need to get validator stage using revision of the subtask.

although is't right to use latest revision(same result), but it's more clear this way.

dm/dm/worker/source_worker.go
D3Hunter: `fetchSubTasksAndAdjust` is fetching in a transaction in batch. for `getExpectValidatorStage`, it's getting for a single task

lance6716 (approved):

D3Hunter:
needs merge @lance6716 @Ehco1996

PR #4493, opened 2022-01-27
master(dm): add cluster id as unique identification of dm-cluster and optimize error message of code:30008

### LINKED ISSUE #3830: Add log for etcd retry

https://github.com/pingcap/ticdc/blob/9721053edae6b146b726bd5daec04a0dc8c57b1a/dm/pkg/retry/strategy.go#L66-L92

we don't the error when retry

log it

_No response_

_No response_

### LINKED ISSUE #4370: Dmctl can't start relay after upgrade from 2.0.1(enable-relay:  true)-> 5.4.0
### What did you do?

config source1 as enable-relay: false
upgrade dm cluster from 2.0.1 -> v5.4.0
try to start-relay for source1

start-relay for source1 success

### What did you see instead?

fail to start-relay,  
use:
 tiup dmctl --master-addr=ip:8261 start-relay  -s mysql-replica-2-002 dm-worker-8101
report :
               "result": {
                    "isCanceled": false,
                    "errors": [
                        {
                            "ErrCode": 30008,
                            "ErrClass": "relay-unit",
                            "ErrScope": "internal",
                            "ErrLevel": "high",
                            "Message": "initial relay unit: dirpath is empty",
                            "RawCause": "",
                            "Workaround": "Please check the `relay-dir` config in source config file."
                        }
                    ],
use _operate-source show_ to check, report:
           "msg": "[code=38032:class=dm-master:scope=internal:level=high], Message: some error occurs in dm-worker: ErrCode:30008 ErrClass:\"relay-unit\" ErrScope:\"internal\" ErrLevel:\"high\" Message:\"initial relay unit: dirpath is empty\" Workaround:\"Please check the `relay-dir` config in source config file.\" , Workaround: Please execute `query-status` to check status.",
            "source": "mysql-replica-2-002",
            "worker": "dm-worker-8101"
        }
    ]

### Versions of the cluster

```console
v5.4.0
```
```console
```
```console
```
```console
(leave TiUP or manually here)
```
```console
>
>
```

```console
```

### LINKED ISSUE #4456: [DM] Add unique identification of dm cluster to integrate with Clinic Diagnotic Toolkits

Clinic Diagnostic Toolkits is a set of TiDB cluster diagnostic tools, it can be also integrated into DM cluster. But without an unique identification of the whole cluster, the logs and metrics cann't be tracked.

We add an unique identification of DM cluster named `cluster id`.
- DM cluster can generate or get `cluster id` during initialization phase, cluster information should be persisted to Etcd.
- We will expose cluster information through OpenAPI

None

- Distinguish different DM clusters.
- Track logs and metrics in the same DM cluster.

niubell:

Issue Number: close #4370, close #4456, close #3830

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
add cluster id as unique identification of dm-cluster and optimize error message of code:30008
```


dm/dm/master/openapi.go
```diff
@@ -901,6 +901,13 @@ func (s *Server) DMAPUpdateTaskTemplate(c *gin.Context, taskName string) {
 	c.IndentedJSON(http.StatusOK, task)
 }
 
+// DMAPIGetClusterID return cluster id of dm cluster
+func (s *Server) DMAPIGetClusterID(c *gin.Context) {
```
D3Hunter, line 905: may add a more common api like `/api/v1/cluster/info` which contains id (only contains id now) so we can add more info to it later.

dm/dm/master/server.go
```diff
@@ -405,6 +411,45 @@ func subtaskCfgPointersToInstances(stCfgPointers ...*config.SubTaskConfig) []con
...
+	}
+
+	s.clusterID = binary.BigEndian.Uint64(resp.Kvs[0].Value)
+	log.L().Info("init cluster id success")
+	return nil
+}
+
+// Only leader is able to return correct cluster id
+func (s *Server) ClusterID() uint64 {
+	return atomic.LoadUint64(&s.clusterID)
```
D3Hunter, line 450: please remove `atomic.load/store` and surround it with a rw lock
1.  `atomic.load/store` only states `atomic`. there's no clear definition in atomic package regards to concurrent access, neither `golang memory model`, this load may not load the latest value.
2. not a frequent call which need this kind of optimization.

dm/dm/master/server.go
```diff
@@ -405,6 +411,45 @@ func subtaskCfgPointersToInstances(stCfgPointers ...*config.SubTaskConfig) []con
 	return stCfgs
 }
 
+func (s *Server) initClusterID(ctx context.Context) error {
+	log.L().Info("init cluster id begin")
+	ctx1, cancel := context.WithTimeout(ctx, time.Second*10)
```
D3Hunter, line 416:
```suggestion
	ctx1, cancel := context.WithTimeout(ctx, etcdutil.DefaultRequestTimeout)
```

dm/dm/master/server.go
```diff
@@ -405,6 +411,45 @@ func subtaskCfgPointersToInstances(stCfgPointers ...*config.SubTaskConfig) []con
...
+	if err != nil {
+		return err
+	}
+
+	// New cluster, generate a cluster id and backfill it to etcd
+	if len(resp.Kvs) == 0 {
+		ts := uint64(time.Now().Unix())
+		clusterID := (ts << 32) + uint64(rand.Uint32())
+		clusterIDBytes := make([]byte, 8)
+		binary.BigEndian.PutUint64(clusterIDBytes, clusterID)
```
D3Hunter, line 429: may just use `uuid.New().String()` in `"github.com/google/uuid"`

dm/dm/master/server.go
```diff
@@ -405,6 +411,45 @@ func subtaskCfgPointersToInstances(stCfgPointers ...*config.SubTaskConfig) []con
...
+		_, err = s.etcdClient.Put(ctx1, common.ClusterIDKey, string(clusterIDBytes))
+		if err != nil {
+			return err
+		}
+		atomic.StoreUint64(&s.clusterID, clusterID)
+		log.L().Info("generate and init cluster id success")
```
D3Hunter, line 435: log the new generated cluster id ?

dm/dm/master/server.go
```diff
@@ -405,6 +411,45 @@ func subtaskCfgPointersToInstances(stCfgPointers ...*config.SubTaskConfig) []con
...
+	if len(resp.Kvs[0].Value) != 8 {
+		return terror.ErrMasterInvalidClusterID.Generate(resp.Kvs[0].Value)
+	}
+
+	s.clusterID = binary.BigEndian.Uint64(resp.Kvs[0].Value)
+	log.L().Info("init cluster id success")
+	return nil
+}
+
+// Only leader is able to return correct cluster id
```
D3Hunter, line 448:
```suggestion
// ClusterID Only leader is able to return correct cluster id
```

dm/dm/master/openapi.go
niubell: > may add a more common api like `/api/v1/cluster/info` which contains id (only contains id now) so we can add more info to it later.

As now we have `/cluster/version`, so i add `/cluster/id`.We can add `/cluster/info` when we need and remove those two in the future.

dm/openapi/spec/dm.yaml
```diff
@@ -1833,3 +1846,12 @@ components:
       required:
         - "success_task_list"
         - "failed_task_list"
+    GetClusterIDResponse:
+      type: object
+      properties:
+        cluster_id:
```
Ehco1996, line 1852: weak suggestion: maybe use string ID is better for human readning?

dm/dm/master/server.go
niubell: > may just use `uuid.New().String()` in `"github.com/google/uuid"`

Random integer type is needed, so i didn't adopt uuid.

dm/dm/master/server.go
niubell: > log the new generated cluster id ?

Users won't use the cluster id now, Clinic system is the only user through http api.

dm/dm/master/openapi.go
Ehco1996: > remove those two in the future.

as this is an OpenAPI Spec and open source , i think maybe we should do less break changes 😂

dm/dm/master/server.go
niubell: Lock is ok, but i just want an atomic integer, and it had nothing to do with `Premature optimization`, right?

dm/openapi/spec/dm.yaml
niubell: > weak suggestion: maybe use string ID is better for human reading?

Random integer type is needed, so i didn't adopt uuid.

dm/dm/master/server.go
```diff
@@ -15,14 +15,17 @@ package master
...
 	"fmt"
+	"math/rand"
 	"net"
 	"net/http"
 	"reflect"
 	"runtime"
 	"sort"
 	"strings"
 	"sync"
+	"sync/atomic"
```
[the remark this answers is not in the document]
niubell: > can you tell me the reason why we choose to introduc `sync/atomic` instead just use `go.uber.org/atomic`

`go.uber.org/atomic` doesn't have `StoreUint64`.

dm/dm/master/openapi.go
niubell: I will change it to `cluster/info`.

dm/dm/master/server.go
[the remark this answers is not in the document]
Ehco1996: there is a similar func https://pkg.go.dev/go.uber.org/atomic?utm_source=gopls#NewUint64

dm/dm/master/server.go
D3Hunter: seems there's a happens-before relation between atomic operations internally(although not documented), see [this](https://github.com/golang/go/discussions/47141), so it's ok to use it.

recommend to use `go.uber.org/atomic`, it's easy to forget to wrap operation with atomic using `sync/atomic`

niubell: > seems there's a happens-before relation between atomic operations internally(although not documented), see [this](https://github.com/golang/go/discussions/47141), so it's ok to use it.
> 
> recommend to use `go.uber.org/atomic`, it's easy to forget to wrap operation with atomic using `sync/atomic`

Yes, memory model of go is different from java, and scope of some locks in our code is not clear.

dm/dm/master/server.go
```diff
@@ -405,6 +410,45 @@ func subtaskCfgPointersToInstances(stCfgPointers ...*config.SubTaskConfig) []con
...
+	if err != nil {
+		return err
+	}
+
+	// New cluster, generate a cluster id and backfill it to etcd
+	if len(resp.Kvs) == 0 {
+		ts := uint64(time.Now().Unix())
+		clusterID := (ts << 32) + uint64(rand.Uint32())
+		clusterIDBytes := make([]byte, 8)
+		binary.BigEndian.PutUint64(clusterIDBytes, clusterID)
```
D3Hunter, line 428: maybe define a `type clusterID uint64`, and abstract newID/encode/decode inside.

D3Hunter (commented):
rest LGTM

dm/dm/master/server.go
niubell: > maybe define a `type clusterID atomic.Uint64`, and abstract newID/encode/decode inside.

If we need to add more function about cluster id，this advice will be better and that's when we will refactor it, i think current situation is ok.

dm/dm/master/server.go
```diff
@@ -37,6 +39,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/pingcap/tiflow/dm/checker"
+	"github.com/pingcap/tiflow/dm/dm/common"
```
Ehco1996, line 42: already have a import call `dmcomman` in L43 😂, could you pelase help to merge this two import?

Ehco1996 (commented):
restLGTM

D3Hunter (approved):

Ehco1996 (approved):
LGTM

Ehco1996:
> [2022-01-28T17:08:42.949Z] coverage: 15.6% of statements
[2022-01-28T17:08:42.949Z] panic: test timed out after 5m0s

seems there is a deadlock in current ut, ptal @niubell

PR #4491, opened 2022-01-27
metrics(dm): fix lag metrics for ddl never updated

Ehco1996:
### What problem does this PR solve?

Issue Number: close #4471

### What is changed and how it works?

if cuurrent max lag is from ddl job, reset this lag after set like skip job

### Check List 

Tests 

 - Integration test

Code changes

Side effects

 - Possible performance regression
 - Increased code complexity

Related changes

### Release note 

```release-note
`None`.
```

PR #4496, opened 2022-01-27
metrics(dm): incr lag metric when syncer blocked

### LINKED ISSUE #3695: calculate secondsBehindMaster when dm sync block

when DM block in the sync stage(e.g. waiting for DDL), the `secondsBehindMaster` will not be updated.

update the `secondsBehindMaster` even DM block. Don't need to be too precise, just make it bigger and bigger

_No response_

_No response_

Ehco1996:

Issue Number: close #3695

* one lag worker calcl lag by jobTS in bucket every  0.1 s
* when a ddl/dml job is sended into worker queue, add this job's TS in to bucket
* when a ddl/dml job executed to downstream, clear the TS in bucket

### Check List 

Tests 

 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`.
```


Ehco1996:
could you please take another look for this pr ? thanks 🧡🧡🧡

@GMHDBJD @lance6716

dm/syncer/syncer.go
```diff
@@ -858,7 +858,13 @@ func (s *Syncer) updateReplicationLagMetric() {
 	}
 	if minTS != int64(0) {
 		lag = s.calcReplicationLag(minTS)
+		// this means the syncer is blocked so no job is being processed and the lag is always equal to the last one.
+		// we just increased the lag time by 1 to make this metric looks a bit more normal.
+		if lag == s.secondsBehindMaster.Load() {
+			lag += 1
```
lichunzhu, line 864: When syncer get blocked, can dm-worker enter this if block?

Ehco1996: yes , this func is called by ticker https://github.com/pingcap/tiflow/blob/801bb607615e42be6be7770a302d692bca037834/dm/syncer/syncer.go#L1598

lichunzhu: What if dml buckets are all empty and syncer gets blocked at a ddl? Will minTS always be `0` and never enter this part?

Ehco1996: yes, but if in this case, it's hard  to calc lag becasue the first ddl job is blocked, i can't known this job TS

one way to handle this is to add a var in syncer struct used to record the first received job's TS,
but I'm worried that it's not worth losing some performance just to deal with this rare situation, which only happens when there is no DML, only a DDL and the first DDL is Blocked

lichunzhu: I think just handle for the second and later on DDLs are enough. But I'm afraid after https://github.com/pingcap/tiflow/pull/4491 we can't handle this well

Ehco1996: cnsidering that this is a relatively rare case (no DML only one pending DDL) maybe we can ignore this case ?
and the lag will be updated after this pengding ddl executed to downstream 😂

lichunzhu: If we use pessimistic ddl lock, will this happen frequently?

Ehco1996: i think it's still rare, unless user only use dm to sync DDL, btw can you provide a better way to solve the problem? i would love to imporve the current situation

lichunzhu: Maybe we should clear these buckets through the heartbeat mechanism: only clear `workerJobTSArray` only when all jobs in this bucket are cleared and handle a heartbeat job.
But after a second thought, it's a bit complicated. The current implementation is also okay.

lichunzhu (approved):
LGTM

Ehco1996:
could you please take another look for this pr ? thanks 🧡🧡🧡 

cc @@lance6716

dm/syncer/syncer.go
```diff
@@ -866,7 +866,13 @@ func (s *Syncer) updateReplicationLagMetric() {
...
```
GMHDBJD, line 872: Will this cause our delay always be 1(even when there is almost no delay)?

Ehco1996: yes, do we have some other ways to check if syncer is blocked? 🤔

GMHDBJD: This may not be acceptable, users may kill us 😂

GMHDBJD: MySQL master will send heartbeat event if no event every ticker，can it make some help？

Ehco1996:
i adjust the logic of calcuate ddl lag, now lag for ddl and dml is same

* one lag worker calcl lag by jobTS in bucket every  0.1 s
* when a ddl/dml job is sended into worker queue, add this job's TS in to bucket
* when a ddl/dml job executed to downstream, clear the TS in bucket

and this ensures that lag will continue to growth when syncer is blocked.

ptal /cc @lichunzhu  @GMHDBJD

dm/tests/metrics/run.sh
```diff
@@ -106,15 +106,19 @@ function run() {
...
 	echo "check dml/skip lag done!"
 
 	# check new metric: dm_syncer_replication_lag_sum,dm_syncer_replication_lag_gauge,
 	# finished_transaction_total,dm_syncer_ideal_qps,dm_syncer_binlog_event_row,replication_transaction_batch exists
 	check_metric $WORKER1_PORT 'dm_syncer_replication_lag_sum{source_id="mysql-replica-01",task="test",worker="worker1"}' 5 -1 999
 	check_metric $WORKER2_PORT 'dm_syncer_replication_lag_sum{source_id="mysql-replica-02",task="test",worker="worker2"}' 5 -1 999
 
-	check_metric $WORKER1_PORT 'dm_syncer_replication_lag_gauge{source_id="mysql-replica-01",task="test",worker="worker1"}' 5 1 999
-	check_metric $WORKER2_PORT 'dm_syncer_replication_lag_gauge{source_id="mysql-replica-02",task="test",worker="worker2"}' 5 1 999
+	check_metric $WORKER1_PORT 'dm_syncer_replication_lag_gauge{source_id="mysql-replica-01",task="test",worker="worker1"}' 5 -1 999
```
Ehco1996, line 120: this is because there is no lag any more , the purpose of this line is to test wheather `dm_syncer_replication_lag_gauge` is exist

okJiang (approved):
LGTM

okJiang (approved):

dm/syncer/syncer.go
Ehco1996: ping @GMHDBJD

lichunzhu (approved):
LGTM

PR #4458, opened 2022-01-24
rfc(dm): add rfc doc of enhancing task manageability

Ehco1996:
### What problem does this PR solve?

Issue Number: ref #4484

### What is changed and how it works?

as title

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`.
```

Issue #4426, opened 2022-01-20
dm: continuous validation


D3Hunter:
### Is your feature request related to a problem?

no

### Describe the feature you'd like

support continuous data validation during sync.

subtasks
- [ ] add pre-check for validation
- [ ] fast validation mode
  - [ ] validation stage observation
  - [ ] checkpoint
  - [ ] save pending/error data into dm_meta
- [ ] full validation mode
- [ ] dmctl commands for validation
- [ ] prometheus monitoring metric for validation, grafana dashboard and alerting

D3Hunter:
it's a new feature, not affects 5.4

buchuitoudegou:
validator will automatically stop itself when encountering some network errors (e.g. failing to persist checkpoint to downstream).

buchuitoudegou:
[2022/06/09 16:15:42.045 +08:00] [INFO] [checkpoint_flush_worker.go:68] ["about to sync flush checkpoint snapshot"] [task=test] [unit="binlog replication"] [snapshot_id=8]
[2022/06/09 16:15:42.112 +08:00] [INFO] [checkpoint_flush_worker.go:109] ["sync flush checkpoint snapshot successfully"] [task=test] [unit="binlog replication"] [snapshot_id=8] [pos="position: (mysql-bin.000007, 856099799), gtid-set: "]
[2022/06/09 16:15:42.113 +08:00] [INFO] [syncer.go:1255] ["after sync flushed checkpoint, gc all causality keys"] [task=test] [unit="binlog replication"]
[2022/06/09 16:15:56.502 +08:00] [INFO] [validator_checkpoint.go:626] ["error count loaded"] [task=test] [unit="continuous validator"] [counts={}]
[2022/06/09 16:15:56.503 +08:00] [INFO] [data_validator.go:412] ["validator status"] [task=test] [unit="continuous validator"] ["processed(i, u, d)"="[188081,309328,154816]"] ["pending(i, u, d)"="[7095,7573,0]"] ["new error rows(not flushed)"=0] ["binlog process speed"="0.58 MB/s"]
[2022/06/09 16:15:56.944 +08:00] [INFO] [status.go:126] ["binlog replication progress"] [task=test] [unit="binlog replication"] ["total binlog size"=768887649] ["last binlog size"=646532312] ["cost time"=30] [bytes/Second=4078511] ["unsynced binlog size"=472461562] ["estimate time to catch up"=115]
[...]

PR #4498, opened 2022-01-28
validation(dm): dmctl commands for validation

### LINKED ISSUE #4490: dmctl commands for validation

buchuitoudegou:

Issue Number: ref #4490

it's the first part of #4490

1. create dmctl commands
2. create grpc service

### Check List 

Tests 

 - Unit test
 - Integration test will be added in the next PR

Code changes

Side effects

Related changes

#4479 

### Release note 

```release-note
None
```


purelind:
/run-integration-test
run-kafka-integration-test

dm/dm/ctl/master/operate_validation.go
```diff
@@ -0,0 +1,160 @@
...
+		Use:   "clear <task-name> <error-id|--all>",
+		Short: "operate validation error",
+		RunE:  clearError,
+	}
+	cmd.Flags().Bool("all", false, "all task")
+	return cmd
+}
+
+func getFlags(cmd *cobra.Command) (taskName string, errID int, isAll bool, err error) {
+	if len(cmd.Flags().Args()) < 1 {
```
D3Hunter, line 59: check for len(cmd.Flags().Args()) > 2

dm/dm/ctl/master/operate_validation.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return err
+	}
+	resp, err = operateError(taskName, errID, isAll, "ignore")
+	if err != nil {
+		return err
+	}
+	common.PrettyPrintResponse(resp)
+	return nil
+}
```
D3Hunter, line 122: those 3 function can be merged into one, only differ in operation type

dm/dm/ctl/master/operate_validation.go
```diff
@@ -0,0 +1,160 @@
...
+		taskName string
+		errID    int
+		isAll    bool
+		resp     *pb.OperateValidationErrorResponse
+	)
+	taskName, errID, isAll, err = getFlags(cmd)
+	if err != nil {
+		return err
+	}
+	resp, err = operateError(taskName, errID, isAll, "resolve")
```
D3Hunter, line 135: maybe define the operation type as a enum

dm/tests/dmctl_command/run.sh
```diff
@@ -6,7 +6,7 @@ cur=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 source $cur/../_utils/test_prepare
 WORK_DIR=$TEST_DIR/$TEST_NAME
 
-help_cnt=45
+help_cnt=46
```
D3Hunter, line 9: there's no integration test for the commands in this pr, maybe add subtask in #4426 and state it in the pr description

dm/dm/master/server.go
```diff
@@ -2580,3 +2580,179 @@ func (s *Server) sharedLogic(ctx context.Context, req interface{}, respPointer i
...
+	// filter the source that we don't want
+	if len(sources) > 0 {
+		filterSource := map[string]interface{}{}
+		for _, source := range sources {
+			filterSource[source] = true // the source we want
+		}
+		for taskName, sourceCfgs := range ret {
+			for source := range sourceCfgs {
+				if _, ok := filterSource[source]; !ok {
+					delete(ret[taskName], source)
```
D3Hunter, line 2610:
```suggestion
					delete(sourceCfgs, source)
```

dm/dm/master/server.go
```diff
@@ -2580,3 +2580,179 @@ func (s *Server) sharedLogic(ctx context.Context, req interface{}, respPointer i
...
+		Result: true,
+	}
+	subTaskCfgs, err = s.getSubTaskCfgs(req.TaskName, []string{})
+	if err != nil {
+		resp.Result = false
+		resp.Msg = err.Error()
+		// nolint:nilerr
+		return resp, nil
+	}
+	// TODO: get validation status from worker
```
D3Hunter, line 2705: todo: parameter check

dm/dm/master/server.go
```diff
@@ -2580,3 +2580,179 @@ func (s *Server) sharedLogic(ctx context.Context, req interface{}, respPointer i
...
+		Result: true,
+	}
+	subTaskCfgs, err = s.getSubTaskCfgs(req.TaskName, []string{})
+	if err != nil {
+		resp.Result = false
+		resp.Msg = err.Error()
+		// nolint:nilerr
+		return resp, nil
+	}
+	// TODO: get validation error from worker
```
D3Hunter, line 2730: todo: parameter check

dm/dm/master/server.go
```diff
@@ -2580,3 +2580,179 @@ func (s *Server) sharedLogic(ctx context.Context, req interface{}, respPointer i
...
+		Result: true,
+	}
+	subTaskCfgs, err = s.getSubTaskCfgs(req.TaskName, []string{})
+	if err != nil {
+		resp.Result = false
+		resp.Msg = err.Error()
+		// nolint:nilerr
+		return resp, nil
+	}
+	// TODO: operate validation error at worker
```
D3Hunter, line 2755: todo: parameter check

dm/dm/proto/dmmaster.proto
```diff
@@ -466,3 +477,85 @@ enum RelayOpV2 {
...
+    string srcTable = 2;
+    string dstTable = 3;
+    string validationStatus = 4;
+    string message = 5;
+}
+
+message GetValidationStatusResponse {
+    bool result = 1;
+    string msg = 2;
+    ValidationStatus status = 3;
```
D3Hunter, line 523:
```suggestion
    repeated ValidationStatus status = 3;
```

dm/dm/proto/dmmaster.proto
```diff
@@ -466,3 +477,85 @@ enum RelayOpV2 {
...
+    string errorType = 7;
+    string status = 8;
+    string time = 9;
+    string message = 10;
+}
+
+message GetValidationErrorResponse {
+    bool result = 1;
+    string msg = 2;
+    ValidationError error = 3;
```
D3Hunter, line 548:
```suggestion
    repeated ValidationError error = 3;
```

dm/dm/ctl/master/operate_validation.go
```diff
@@ -0,0 +1,160 @@
...
+		RunE:  resolveError,
+	}
+	cmd.Flags().Bool("all", false, "all task")
+	return cmd
+}
+
+func NewClearValidationErrorCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "clear <task-name> <error-id|--all>",
+		Short: "operate validation error",
```
Ehco1996, line 51:
```suggestion
		Short: "clear a validation error",
```

dm/dm/ctl/master/operate_validation.go
Ehco1996: maybe also merge to one ctl cmd

dm/dm/master/server.go
```diff
@@ -2580,3 +2580,179 @@ func (s *Server) sharedLogic(ctx context.Context, req interface{}, respPointer i
 	*errPointer = terror.ErrMasterRequestIsNotForwardToLeader
 	return true
 }
+
+func (s *Server) getSubTaskCfgs(taskName string, sources []string) (map[string]map[string]config.SubTaskConfig, error) {
```
Ehco1996, line 2584: perfer add a func call `getStubTaskCfgByTaskAndSource` in`scheduler `

dm/dm/master/server.go
```diff
@@ -2580,3 +2580,179 @@ func (s *Server) sharedLogic(ctx context.Context, req interface{}, respPointer i
...
+					delete(ret[taskName], source)
+				}
+			}
+			if len(ret[taskName]) == 0 {
+				delete(ret, taskName)
+			}
+		}
+	}
+	if len(ret) == 0 {
+		return nil, errors.Errorf("no subtask matched by the current rule")
```
Ehco1996, line 2619: use terror

dm/dm/master/server.go
```diff
@@ -2580,3 +2580,179 @@ func (s *Server) sharedLogic(ctx context.Context, req interface{}, respPointer i
...
+		subTaskCfgs map[string]map[string]config.SubTaskConfig // task-name->sourceID->*config.SubTaskConfig
+	)
+	shouldRet := s.sharedLogic(ctx, req, &resp2, &err2)
+	if shouldRet {
+		return resp2, err2
+	}
+	resp := &pb.StartValidationResponse{}
+	if (req.IsAllTask && len(req.TaskName) > 0) || (!req.IsAllTask && len(req.TaskName) == 0) {
+		resp.Result = false
+		resp.Msg = "either `task-name` or `all-task` should be set"
```
Ehco1996, line 2637: seems this is checked in dmctl ?

dm/dm/ctl/master/query_validation.go
```diff
@@ -0,0 +1,127 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	resp := &pb.GetValidationErrorResponse{}
+	err = common.SendRequest(
+		ctx,
+		"GetValidationError",
+		&pb.GetValidationErrorRequest{
+			IsIgnoredError: isIgnoredError,
+			IsAllError:     isAllError,
```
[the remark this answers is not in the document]
buchuitoudegou: Agree. The two flags are mutually exclusive.

dm/dm/master/server.go
buchuitoudegou: I'll move it later.

dm/dm/ctl/master/operate_validation.go
buchuitoudegou: > maybe also merge to one ctl cmd

What do you mean about merging them into one command? Is it like how `operate-source` does?
```
dmctl validation operate-error <operate-type> ...
```

dm/tests/dmctl_command/run.sh
buchuitoudegou: I've added a subtask and reverted this change.

dm/dm/master/server.go
buchuitoudegou: Is it possible the gprc service is invoked from other clients in addition to `dmctl`, such as openAPI?

Ehco1996: it's possible for user to connect directly to the rpc server but openAPI will not. so it's ok to check twice

dm/dm/ctl/master/query_validation.go
```diff
@@ -0,0 +1,132 @@
...
+	)
+
+	if len(cmd.Flags().Args()) == 0 {
+		cmd.SetOut(os.Stdout)
+		common.PrintCmdUsage(cmd)
+		return errors.New("task name should be specified")
+	}
+	taskName = cmd.Flags().Arg(0)
+	status, err = cmd.Flags().GetString("stage")
+	if status != "" && status != ValidationStageRunning && status != ValidationStageStopped {
```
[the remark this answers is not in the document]
buchuitoudegou: Yes, it's optional. If stage is not set, its value should be `""` and then pass this check.

dm/dm/ctl/master/query_validation.go
```diff
@@ -0,0 +1,132 @@
...
+	ValidationIgnoredErr   = "ignored"
+)
+
+func NewQueryValidationErrorCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "show-errors [--error] <task-name>",
+		Short: "show error of the validation task",
+		RunE:  queryValidationError,
+	}
+	cmd.Flags().String("error", "all", "filtering type of error: all (default) or ignored")
```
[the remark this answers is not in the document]
buchuitoudegou: Are non-ignored errors are a subset of all errors? 
And what kinds of errors should be reported if the `error` flag is not set? I thought `all` is the default value since we should report all kinds of errors (including `ignored` and `non-ignored`, etc.) if users don't set it. 
According to the spec, it seems only these two values are accepted. Is `non-ignored` acceptable here?

dm/dm/ctl/master/query_validation.go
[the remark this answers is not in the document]
D3Hunter: without this flag, default to `running`?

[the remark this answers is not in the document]
buchuitoudegou: I thought the master should return all validator tasks including running and stopped ones. `stage` flag is a filter condition.

dm/dm/ctl/master/query_validation.go
```diff
@@ -0,0 +1,132 @@
...
+	)
+	if len(cmd.Flags().Args()) != 1 {
+		return errors.New("task name should be specified")
+	}
+	taskName = cmd.Flags().Arg(0)
+	errType, err = cmd.Flags().GetString("error")
+	if err != nil {
+		return err
+	}
+	if errType != ValidationAllErr && errType != ValidationIgnoredErr {
```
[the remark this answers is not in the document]
buchuitoudegou: The default value is set when declaring this flag:
```go
cmd.Flags().String("error", "all", "filtering type of error: all (default) or ignored")
```

dm/dm/ctl/master/query_validation.go
```diff
@@ -81,36 +79,36 @@ func queryValidationError(cmd *cobra.Command, _ []string) (err error) {
 
 func NewQueryValidationStatusCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "status <task-name>",
+		Use:   "status [--stage stage] <task-name>",
 		Short: "query validation status of a task",
 		RunE:  queryValidationStatus,
 	}
-	cmd.Flags().String("stage", "running", "filter status")
+	cmd.Flags().String("stage", "", "filter validation tasks status by stages: Running/Stopped")
```
D3Hunter, line 86: why default value removed?

`Running/Stopped` is hard to type, all lower case is better.

dm/dm/ctl/master/start_stop_validation.go
```diff
@@ -0,0 +1,134 @@
...
+	cmd := &cobra.Command{
+		Use:   "stop [-s source ...] [--all-task] [task-name]",
+		Short: "stop validating the completeness of the data",
+		RunE:  startStopValidation("stop"),
+	}
+	cmd.Flags().Bool("all-task", false, "whether to all tasks")
+	return cmd
+}
+
+func startStopValidation(typ string) func(*cobra.Command, []string) error {
```
D3Hunter, line 50:
```suggestion
func startStopValidation(op string) func(*cobra.Command, []string) error {
```

dm/dm/ctl/master/start_stop_validation.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/pingcap/tiflow/dm/dm/config"
+	"github.com/pingcap/tiflow/dm/dm/ctl/common"
+	"github.com/pingcap/tiflow/dm/dm/pb"
+)
+
+func NewStartValidationCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "start [-s source ...] [--mode mode] [--all-task] [task-name]",
+		Short: "start to validate the completeness of the data",
+		RunE:  startStopValidation("start"),
```
D3Hunter, line 33: maybe define `"start"/"stop"` as a constant

dm/dm/ctl/master/query_validation.go
buchuitoudegou: > why default value removed?

The default value `""` means showing both `Running` and `Stopped` validation tasks. If we set the default value as `running` then this command couldn't show all tasks' but only either `Running` or `Stopped` tasks' statuses.

dm/dm/ctl/master/start_stop_validation.go
buchuitoudegou: > ignored + non-ignored = all

As the result I discussed with zhaoyang in the spec, all-errors = ignored errors + resolved-errors + current-error. I think all types of errors and the detailed functionality of this command should be definite before we resolve it😥 Maybe I should discuss it with PM again.

dm/dm/proto/dmmaster.proto
```diff
@@ -108,6 +108,17 @@ service Master {
   rpc TransferSource(TransferSourceRequest) returns(TransferSourceResponse) {}
 
   rpc OperateRelay(OperateRelayRequest) returns(OperateRelayResponse) {}
+
+   // StartValidation starts validation
```
Ehco1996, line 112: don't need this comment, this just repeat the func name😂

dm/dm/proto/dmmaster.proto
```diff
@@ -108,6 +108,17 @@ service Master {
...
+  rpc StartValidation(StartValidationRequest) returns(StartValidationResponse) {}
+  // StopValidation stop validation
```
Ehco1996, line 114: ditto

dm/dm/proto/dmmaster.proto
```diff
@@ -466,3 +477,84 @@ enum RelayOpV2 {
...
+}
+
+message GetValidationErrorResponse {
+    bool result = 1;
+    string msg = 2;
+    repeated ValidationError error = 3;
+}
+
+message OperateValidationErrorRequest {
+    string op = 1; // ignore/resolve/clear
```
Ehco1996, line 551: this op should be an enum like `TaskOp`

dm/dm/master/scheduler/scheduler.go
```diff
@@ -1084,6 +1084,46 @@ func (s *Scheduler) GetSubTaskCfgsByTask(task string) map[string]*config.SubTask
 	return cloneM
 }
 
+func (s *Scheduler) GetSubTaskCfgsByTaskAndSource(taskName string, sources []string) (map[string]map[string]config.SubTaskConfig, error) {
+	var ret map[string]map[string]config.SubTaskConfig // task-name->sourceID->*config.SubTaskConfig
+	if len(taskName) == 0 {
+		ret = s.GetSubTaskCfgs()
```
Ehco1996, line 1090: could you please explain this if's purpose ? why caller will input a empty taskName?

dm/dm/master/scheduler/scheduler.go
```diff
@@ -1084,6 +1084,46 @@ func (s *Scheduler) GetSubTaskCfgsByTask(task string) map[string]*config.SubTask
...
+					delete(sourceCfgs, source)
+				}
+			}
+			if len(ret[taskName]) == 0 {
+				delete(ret, taskName)
+			}
+		}
+	}
+	if len(ret) == 0 {
+		return nil, terror.Annotate(errors.New("no such subtask or source"), "fail to get subtask config by task name and source")
```
Ehco1996, line 1122: perfer to return an empty map and don't return error, caller should check the result and decide whether to report an error

dm/dm/proto/dmmaster.proto
```diff
@@ -108,6 +108,17 @@ service Master {
...
   rpc OperateRelay(OperateRelayRequest) returns(OperateRelayResponse) {}
+
+   // StartValidation starts validation
+  rpc StartValidation(StartValidationRequest) returns(StartValidationResponse) {}
+  // StopValidation stop validation
+  rpc StopValidation(StopValidationRequest) returns(StopValidationResponse) {}
+
+  rpc GetValidationStatus(GetValidationStatusRequest) returns(GetValidationStatusResponse) {}
+
+  rpc GetValidationError(GetValidationErrorRequest) returns(GetValidationErrorResponse) {}
```
Ehco1996, line 119: just a question: if an error hanppend during validation, will the validation process be interrupted? if not, we should take care of the body size of response, it can be  larger than our imagine (maybe add pagination to solve this)

btw maybe we can merge `GetValidationStatus` to `GetValidationStatus` if so user only need one cmd to check the validation work status

buchuitoudegou: > if an error hanppend during validation, will the validation process be interrupted? if not

No, the validator will keep running until the `error-rows` reaches the `max-error-count`. It's probably necessary if the `max-error-count` is too large, but currently the number of errors won't grow infinitely. 

> btw maybe we can merge GetValidationStatus to GetValidationStatus if so user only need one cmd to check the validation work status

What do you mean about "merge `GetValidationStatus` to `GetValidationStatus`"😂?  Do you expect to merge the `query error` command to `query status` command? They are actually doing different things. The `query error` shows which rows are reported erroneous while `query status` shows the validation tasks' statuses (including whether they are still running, some details about the sources, etc.)

dm/dm/master/scheduler/scheduler.go
buchuitoudegou: When users would like to get all tasks of some sources, the `taskName` would be empty.

dm/dm/ctl/master/operate_validation.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/pingcap/tiflow/dm/dm/ctl/common"
+	"github.com/pingcap/tiflow/dm/dm/pb"
+)
+
+func NewIgnoreValidationErrorCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "ignore-error <task-name> <error-id|--all>",
```
Ehco1996, line 29: how about only allow keyword args for new added cli? such as  `dmctl validation ignore-error --task-name=test --error-id=1 --all=true`

we can discuss in this [rfc](https://github.com/pingcap/tiflow/pull/4458#discussion_r805488938)

this will make it much easier to write code for check args

Ehco1996 (commented):
restLGTM

dm/dm/ctl/master/operate_validation.go
buchuitoudegou: It sounds more intuitive and plausible to make user distinguish these two args. But the original one looks more succinct. @D3Hunter PTAL

D3Hunter: those 2 parameters are both mandatory, flags are better for optional ones, i prefer the original one.

D3Hunter (approved):
LGTM

dm/dm/ctl/master/operate_validation.go
Ehco1996: Reasonable

Ehco1996 (approved):
LGTM

PR #4539, opened 2022-02-09
validator(dm): full mode validator, part 1

### LINKED ISSUE #4488: validation(dm): full validation mode

buchuitoudegou:

Issue Number: ref #4488

part 1 of #4488

1. refactor the prototype of the validator
2. migrate the validator to DM

### Check List 

Tests 

 - Unit test

Code changes
- data_validation.go
- validation_cond.go

Side effects

- no

Related changes
- pr #4479 

### Release note 

```release-note
`None`
```


dm/syncer/data_validator.go
```diff
@@ -15,23 +15,73 @@ package syncer
...
+// change of table
+// binlog changes are clustered into table changes
+// the validator validates changes of table-grain at a time.
+type tableChange struct {
+	table *validateTableInfo
+	rows  map[string]*rowChange
+}
+
+// change of a row.
+type rowChange struct {
```
lance6716, line 75: please take a look at https://github.com/pingcap/tiflow/blob/master/pkg/sqlmodel/row_change.go

this library can be shared by DM and CDC, so maybe we can easier port data validator to CDC in future

lance6716 (commented):
BTW, does sync-diff-inspector from tidb-tools have some well-tested functions that you can use?

dm/syncer/data_validator.go
D3Hunter: struct is similar,  with no `prev data` and more extra fields. will stick to it now, there may other changes later.

> BTW, does sync-diff-inspector from tidb-tools have some well-tested functions that you can use?

the data compare part comes from `sync-diff-inspector`, but this logic can be mush simplier in continuous validation, so it's reimplemented, most other code is continuous validation only

lonng:
This PR is too large and please split it into some smalls.

dm/pkg/binlog/position.go
```diff
@@ -428,3 +428,7 @@ func (l *Location) SetGTID(gset gmysql.GTIDSet) error {
 func (l *Location) GetGTID() gtid.Set {
 	return l.gtidSet
 }
+
+func (l *Location) Update(gtidStr string) error {
```
Ehco1996, line 432: we already have `func (l *Location) SetGTID(gset gmysql.GTIDSet) error` , can we just use that func?

dm/syncer/syncer.go
```diff
@@ -172,7 +172,10 @@ type Syncer struct {
 	exprFilterGroup *ExprFilterGroup
 	sessCtx         sessionctx.Context
 
-	closed atomic.Bool
+	// int value of pb.Stage, stage diagram:
+	// new --> running --> stopped --> finished
+	//     |-> finished
+	stage atomic.Int32
```
Ehco1996, line 178: what is stage of `finished` ? syncer will never finished, so seems use one var `isRunning` is enough?

dm/syncer/validator_cond.go
```diff
@@ -0,0 +1,72 @@
...
+package syncer
+
+import (
+	"strings"
+)
+
+type Cond struct {
+	Table     *validateTableInfo
+	ColumnCnt int
+	PkValues  [][]string
```
Ehco1996, line 23: can we reuse RowChange instead ? see https://github.com/pingcap/tiflow/blob/master/pkg/sqlmodel/row_change.go

dm/syncer/data_validator.go
```diff
@@ -138,51 +224,168 @@ func (v *DataValidator) fillResult(err error, needLock bool) {
...
+		case <-v.ctx.Done():
+			return v.ctx.Err()
+		case <-v.waitSyncerTimer.C:
+			fired = true
+			syncLoc = v.syncer.checkpoint.FlushedGlobalPoint()
+			cmp = binlog.CompareLocation(currLoc, syncLoc, v.cfg.EnableGTID)
+			if cmp <= 0 {
+				return nil
+			}
+			v.waitSyncerTimer.Reset(checkInterval)
```
Ehco1996, line 289: do we need to add a log here？ in case of syncer never catch up with master

dm/pkg/binlog/position.go
D3Hunter: syncer calls `SetGTID` when meet `query event`(ddl) or `xid event`, i.e. after the transaction is synced, there is a complete gtid set maintained in `go-mysql`.

but in validator, when meet a `gtid event`, i.e. transaction begin, we need to update the location(simpler than extending `go-mysql` to add a field in this event), so we can check whether syncer synced them

dm/syncer/validator_cond.go
D3Hunter: struct is similar, but nearly no code can be reused, also there're fields special to validation, will leave it before all logic is completed.

dm/syncer/syncer.go
D3Hunter: `isRunning` overlaps with existing `isClosed` fields, i.e. running means not closed, so i merge them into one fields called stage(reusing existing enum), `finished` is set when `Close` is called.

dm/pkg/binlog/position.go
Ehco1996: Reasonably, update has less overhead

dm/syncer/data_validator.go
```diff
@@ -198,3 +402,158 @@ func (v *DataValidator) Stage() pb.Stage {
...
+		return nil
+	}
+
+	var primaryIdx *model.IndexInfo
+	for _, idx := range tableInfo.Indices {
+		if idx.Primary {
+			primaryIdx = idx
+		}
+	}
+	if primaryIdx == nil {
```
lance6716, line 443: there will be no PK in tableInfo.Indices when PKIsHandle in my memory. We might provide a function for this case as well as for `handlePkExCase`

dm/syncer/data_validator.go
```diff
@@ -198,3 +402,158 @@ func (v *DataValidator) Stage() pb.Stage {
...
+		// replication.DELETE_ROWS_EVENTv0, replication.DELETE_ROWS_EVENTv1, replication.DELETE_ROWS_EVENTv2:
+		return rowDeleted
+	}
+}
+
+func genRowKey(pkValues []string) string {
+	return strings.Join(pkValues, "-")
+}
+
+func genColData(v interface{}) []byte {
```
lance6716, line 548: the return value is always cast to string, better change the return type.

dm/dm/config/task.go
```diff
@@ -337,7 +339,8 @@ func (m *SyncerConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 }
 
 type ValidatorConfig struct {
-	Mode string `yaml:"mode" toml:"mode" json:"mode"`
+	Mode        string `yaml:"mode" toml:"mode" json:"mode"`
+	WorkerCount int    `yaml:"worker-count" toml:"worker-count" json:"worker-count"`
```
lonng, line 343: Please fix JSON tag name.

dm/syncer/data_validator.go
D3Hunter: `PKIsHandle = true` when table has primary key like `id int primary key CLUSTERED`, since schema-tracker(we get from it) only use downstream DDL when the task is incremental only, will support this case later.

dm/syncer/data_validator.go
D3Hunter: `dbutil.ColumnData` stores `[]byte`, validation worker uses it. will remove this struct dependency in next pr.

dm/dm/config/task.go
D3Hunter: you mean `-` to `_`?

most of dm code uses `-` actually, we should follow this convention

lonng: Oops, the standard JSON tags should be `_`.

dm/syncer/data_validator.go
```diff
@@ -138,51 +224,169 @@ func (v *DataValidator) fillResult(err error, needLock bool) {
...
+	} else {
+		v.result.Errors = append(v.result.Errors, unit.NewProcessError(err))
+	}
+}
+
+func (v *DataValidator) errorProcessRoutine() {
+	v.errProcessWg.Done()
+	for {
+		select {
+		case err := <-v.errChan:
```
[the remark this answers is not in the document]
D3Hunter: right now channel size is big enough, left a todo in the constructor to enhance error handling in later pr.

dm/dm/config/task.go
D3Hunter: JSON api recommend uses camelCase actually.

for dm, i think we should stick to current convention to make it consistent with current names.

dm/syncer/syncer.go
D3Hunter: will revert and add `isRunning` as other routine(ungraceful kill) may change stage and it's not waiting `s.Process`(no one manages this routine) exist, so stage may leave with a `stopped` value not `finished`. i can move set stage to running/stopped part to a routine managed by `syncer.runWg`, but it looks odd

D3Hunter:
ptal

D3Hunter:
ptal
@lance6716 @Ehco1996

dm/syncer/syncer.go
```diff
@@ -691,6 +693,9 @@ func (s *Syncer) Process(ctx context.Context, pr chan pb.ProcessResult) {
 		<-newCtx.Done() // ctx or newCtx
 	}()
 
+	s.running.Store(true)
```
Ehco1996, line 696: i think this part should put into `syncer.Run`

D3Hunter: the name `syncer.Run` is actually misleading, `Process` is the real `run` which is called in a a separate routine, `syncer.Run` is something like `doRun` which is only called inside `process`. so i put it here.

dm/syncer/data_validator.go
```diff
@@ -15,23 +15,75 @@ package syncer
...
+	checkInterval      = 5 * time.Second
+	validationInterval = 10 * time.Second
+)
+
+type validateTableInfo struct {
+	Source     *filter.Table
+	Info       *model.TableInfo
+	PrimaryKey *model.IndexInfo
+	Target     *filter.Table // target table after route
+	pkIndices  []int
```
lance6716, line 55: what is it used for?

dm/syncer/data_validator.go
```diff
@@ -15,23 +15,75 @@ package syncer
...
+	rows  map[string]*rowChange
+}
+
+// change of a row.
+type rowChange struct {
+	table      *validateTableInfo
+	key        string
+	pkValues   []string
+	data       []interface{}
+	theType    rowChangeType
```
lance6716, line 81: sometimes we use `tp`

dm/syncer/data_validator.go
```diff
@@ -45,45 +97,79 @@ type DataValidator struct {
...
 		cfg:    cfg,
 		syncer: syncerObj,
 		stage:  pb.Stage_Stopped,
 	}
-	c.L = log.With(zap.String("task", cfg.Name), zap.String("unit", "continuous validator"))
-	return c
+	v.L = log.With(zap.String("task", cfg.Name), zap.String("unit", "continuous validator"))
+
+	v.workerCnt = cfg.ValidatorCfg.WorkerCount
+	v.changeEventCount = make([]atomic.Int64, 4)
```
lance6716, line 135: seems there's only 3 types

dm/syncer/data_validator.go
```diff
@@ -138,52 +224,167 @@ func (v *DataValidator) fillResult(err error, needLock bool) {
...
 
-	var errs []*pb.ProcessError
-	if utils.IsContextCanceledError(err) {
-		v.L.Info("filter out context cancelled error", log.ShortError(err))
-	} else {
-		errs = append(errs, unit.NewProcessError(err))
-	}
-
+	// todo: if error, validation is stopped( and cancelled), and we may receive a cancelled error.
+	// todo: do we need this cancel field?
```
lance6716, line 228: For other units like syncer, this field can tell paused-by-user from paused-by-error.

lance6716 (commented):
will review later

dm/syncer/data_validator.go
D3Hunter: when generate select, used to get pk column names and values

dm/syncer/data_validator.go
D3Hunter: used to be 4

dm/syncer/data_validator.go
```diff
@@ -198,3 +399,158 @@ func (v *DataValidator) Stage() pb.Stage {
...
+		Schema: string(ev.Table.Schema),
+		Name:   string(ev.Table.Table),
+	}
+	fullTableName := sourceTable.String()
+	if _, ok := v.unsupportedTable[fullTableName]; ok {
+		return nil
+	}
+
+	targetTable := v.syncer.route(sourceTable)
+	tableInfo, err := v.syncer.schemaTracker.GetTableInfo(sourceTable)
```
lance6716, line 405: here we get TableInfo of upstream, not the DownstreamTableInfo, so we need to handle PK IndexInfo is not in Indices when PkIsHandle

dm/syncer/data_validator.go
```diff
@@ -198,3 +399,158 @@ func (v *DataValidator) Stage() pb.Stage {
...
+				afterPkValue[idx] = string(genColData(afterRow[idx]))
+			}
+			afterKey := genRowKey(afterPkValue)
+			if afterKey != key {
+				// convert to delete and insert
+				v.dispatchRowChange(key, &rowChange{
+					table:      table,
+					key:        key,
+					pkValues:   pkValue,
+					data:       row,
```
lance6716, line 485: remind that in future PR, these values may need further processing like `castUnsigned`

dm/syncer/data_validator.go
```diff
@@ -198,3 +399,158 @@ func (v *DataValidator) Stage() pb.Stage {
...
+		if changeType == rowUpdated {
+			afterRowChangeType := changeType
+			afterRow := ev.Rows[i+1]
+			afterPkValue := make([]string, len(pk.Columns))
+			for _, idx := range pkIndices {
+				afterPkValue[idx] = string(genColData(afterRow[idx]))
+			}
+			afterKey := genRowKey(afterPkValue)
+			if afterKey != key {
+				// convert to delete and insert
```
lance6716, line 480: > // convert to delete and insert

this is similar to `RowChange.SplitUpdate()`, and checking if PK is changed is similar to `RowChange.IsIdentityUpdated()`. I really hope we can use one representation and derive many methods as needed from it. Please left a TODO for it

dm/syncer/data_validator_test.go
```diff
@@ -0,0 +1,460 @@
...
+	baseDB := conn.NewBaseDB(db, func() {})
+	baseConn, err := baseDB.GetBaseConn(context.Background())
+	require.NoError(t, err)
+	return &dbconn.DBConn{
+		BaseConn: baseConn,
+		Cfg:      cfg,
+	}
+}
+
+func Test_validator_StartStop(t *testing.T) {
```
lance6716, line 103: this is different with current test function naming convention

dm/syncer/data_validator.go
```diff
@@ -198,3 +399,158 @@ func (v *DataValidator) Stage() pb.Stage {
...
+	return strings.Join(pkValues, "-")
+}
+
+func genColData(v interface{}) []byte {
+	switch dv := v.(type) {
+	case []byte:
+		return dv
+	case string:
+		return []byte(dv)
+	case decimal.Decimal:
```
lance6716, line 536: we didn't set UseDecimal in BinlogSyncerConfig, so in fact will not meet this type

dm/syncer/validator_cond.go
lance6716: the main inconvenience of RowChange.genWhere is it can't handle INSERT type binlog event. I think we can extract a more generic genWhere and let RowChange and validator use it.

There is also a similar logic of CDC in `prepareDelete`, I really hope we can unify them.

dm/syncer/data_validator.go
```diff
@@ -138,52 +224,167 @@ func (v *DataValidator) fillResult(err error, needLock bool) {
...
-	tctx := tcontext.NewContext(v.ctx, v.L)
-	err := v.streamerController.Start(tctx, lastLocationForTest)
-	if err != nil {
-		v.fillResult(terror.Annotate(err, "fail to restart streamer controller"), true)
+	if err := v.waitSyncerRunning(); err != nil {
+		v.errChan <- terror.Annotate(err, "failed to wait syncer running")
 		return
 	}
 
+	// todo: syncer may change replication location(start from timestamp, sharding resync), how validator react?
```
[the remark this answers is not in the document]
D3Hunter: as we have discussed, syncer may start from a timestamp, but since the task need to be stopped, so this case will not happen.

left a todo, to handle the case the case validator may get a later location when validator starts on task start.

dm/syncer/data_validator.go
```diff
@@ -138,52 +224,167 @@ func (v *DataValidator) fillResult(err error, needLock bool) {
...
+		if err != nil {
+			v.errChan <- terror.Annotate(err, "fail to start streamer controller")
+			return
+		}
+	}
+
 	v.L.Info("start continuous validation")
+	v.startValidateWorkers()
+
+	currLoc := location.CloneWithFlavor(v.cfg.Flavor)
```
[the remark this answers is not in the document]
D3Hunter: no, validator is updating gtidset when met GTID event, not setting gtidset when met xid or query event as syncer

dm/syncer/data_validator.go
D3Hunter: see https://github.com/pingcap/tiflow/pull/4539#discussion_r815644442, will left a todo

dm/syncer/data_validator.go
D3Hunter: not sure about it, will left to todo to handle it later.

dm/syncer/data_validator_test.go
D3Hunter: it's not that easy to filter cases by struct name when using camel case, but you're right

dm/dm/config/task.go
Ehco1996: 😭 In fact the new json tags added are already `_` like  [TaskCliArgs](https://github.com/pingcap/tiflow/blob/21e7eb610c16ec0d123aab5ca7290c3d2b3d6816/dm/dm/config/task_cli_args.go#L31)

I think it needs to be standardized to `_` at least when we need to expose the josn format to the user, which is what all `OpenAPI` currently do.

---
Convention over Configuration
_Who cares what format your database primary keys are described by? Does it really matter whether it’s “id”, “postId”, “posts_id”, or “pid”? Is this a decision that’s worthy of recurrent deliberation? No._

from [ruby on rails](https://rubyonrails.org/doctrine#convention-over-configuration)

dm/syncer/data_validator.go
```diff
@@ -198,3 +386,158 @@ func (v *DataValidator) Stage() pb.Stage {
...
+	}
+	fullTableName := sourceTable.String()
+	if _, ok := v.unsupportedTable[fullTableName]; ok {
+		return nil
+	}
+
+	targetTable := v.syncer.route(sourceTable)
+	tableInfo, err := v.syncer.schemaTracker.GetTableInfo(sourceTable)
+	if err != nil {
+		if schema.IsTableNotExists(err) {
```
Ehco1996, line 409: can we use `sync.skipRowsEvent` like syncer to avoid this error happen ?  

> i guess `skipRowsEvent` have better perfoemance than `GetTableInfo` seems schema need use disk io

Ehco1996 (commented):
rest LGTM

dm/syncer/data_validator.go
```diff
@@ -138,52 +223,155 @@ func (v *DataValidator) fillResult(err error, needLock bool) {
...
-		v.fillResult(terror.Annotate(err, "fail to restart streamer controller"), true)
+	if err := v.waitSyncerRunning(); err != nil {
+		v.errChan <- terror.Annotate(err, "failed to wait syncer running")
 		return
 	}
 
+	// todo: if validator starts on task start, we need to make sure the location we got is the init location of syncer
+	// todo: right now, there's change they have a gap
+	location := v.syncer.checkpoint.FlushedGlobalPoint()
+	if v.streamerController.IsClosed() {
```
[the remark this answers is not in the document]
D3Hunter: it's for test, some fields inside `streamerController` are mocked, will leave a comment

dm/syncer/data_validator.go
```diff
@@ -138,52 +223,155 @@ func (v *DataValidator) fillResult(err error, needLock bool) {
...
-		IsCanceled: isCanceled,
-		Errors:     errs,
+func (v *DataValidator) errorProcessRoutine() {
+	defer v.errProcessWg.Done()
+	for err := range v.errChan {
+		v.fillResult(err, true)
+
+		if errors.Cause(err) != context.Canceled {
+			// todo: need a better way to handle err(auto resuming on some error, etc.)
+			v.stopInner()
```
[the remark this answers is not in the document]
D3Hunter: errChan is closed after all worker routines finished

https://github.com/buchuitoudegou/tiflow/blob/a4b807dd503b584890df904e5a2667c136ddb301/dm/syncer/data_validator.go#L372-L373

dm/syncer/data_validator.go
lance6716: @WizardXiao PTAL

dm/syncer/data_validator.go
lance6716: seems we don't need `castUnsigned`. I forget why the other routine of DM needs this function :cry:

dm/syncer/data_validator.go
D3Hunter: added before `GetTableInfo`

dm/syncer/data_validator.go
lance6716: I'm OK with left a TODO.

Another problem is I'm not sure `GetTableInfo` will race when the TableInfo is changed because syncer is processing a DDL.

lance6716 (commented):
rest lgtm

lance6716 (approved):

Ehco1996 (approved):

dm/syncer/validator_cond.go
```diff
@@ -0,0 +1,72 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package syncer
+
+import (
+	"strings"
+)
+
+type Cond struct {
```
okJiang, line 20: I'm sorry that I can't understand this naming😂.

What does `Cond` mean? And what is its full name? @D3Hunter

lance6716: condition

okJiang: I see that `Cond` is used in `validateDeletedRows` and `validateInsertAndUpdateRows`.

So, why name it condition? I don't know how to establish their connection?

buchuitoudegou: struct `Cond` is used to formulate the `where` clause when querying the downstream database.

okJiang: I see! It means the condition in SQL...

I didn't think of it first. 😅

PR #4553, opened 2022-02-10
worker(dm): add retry for watch when network is weak

### LINKED ISSUE #4548: dm-worker  hang forver when etcd watch ch closed
### What did you do?

one dm-worker node in weak network environment

so thread of `WatchSourceBound` exit by
https://github.com/pingcap/tiflow/blob/1c1015b01d5ad2c738968d763161abfed0b21fc0/dm/pkg/ha/bound.go#L273

How did I determine that exited on this line？ in logs we can see
```
[2022/01/03 15:49:16.622 +08:00] [INFO] [server.go:581] ["receive source bound"] [bound="{\"source\":\"\",\"worker\":\"dm-xxxx\"}"] ["is deleted"=true]

do some unboud source operateions

[2022/01/03 16:22:30.662 +08:00] [INFO] [server.go:602] ["handleSourceBound will quit now"]
```
because there is no error log like `observeSourceBound is failed and will quit now`, se we can make sure `handleSourceBound` exit becasue `sourceBoundCh` or `sourceBoundErrCh` is closed

but the thread of keepalive( keepalive and watchsourceBound may use different grpc stream ) is still working...
```
[2022/01/03 16:22:31.663 +08:00] [INFO] [join.go:93] ["start to keepalive with master"]
[2022/01/03 16:24:49.471 +08:00] [INFO] [keepalive.go:139] ["keep alive channel is closed"]
[2022/01/03 16:24:51.517 +08:00] [WARN] [keepalive.go:122] ["fail to revoke lease"] []
[2022/01/03 16:24:51.517 +08:00] [WARN] [join.go:105] ["keepalive with master goroutine paused"] []
[2022/01/03 16:24:51.517 +08:00] [WARN] [server.go:558] ["worker has not been started, no need to stop"] [source=]
[2022/01/03 16:24:52.517 +08:00] [INFO] [join.go:93] ["start to keepalive with master"]
```

as a result, dm-master think this worker is alive ,but actually the worker is hang forever

dm-worker offline when observeSourceBound quit

### What did you see instead?

_No response_

### Versions of the cluster

master

```console
```

Ehco1996:

Issue Number: close #4548

* add failpoint to reproduce #4548 
* add log when WatchSourceBound channel closed

### Check List 

Tests 

 - Integration test

Code changes

* add retry for observeSourceBound when watch channel closed by weak network
* restart keep alive after observeSourceBound exit
Side effects

Related changes

### Release note 

```release-note
`add retry for watch when network is weak`.
```


dm/pkg/ha/bound.go
```diff
@@ -269,7 +269,11 @@ func WatchSourceBound(ctx context.Context, cli *clientv3.Client, worker string,
 		case <-ctx.Done():
 			return
 		case resp, ok := <-ch:
```
lance6716, line 271: what's the reason of this WatchChan being closed?

Ehco1996: when networkw is poor, all grpc stream is dead, etcd watch will retrn an closed channel

see here https://github.com/etcd-io/etcd/blob/20c89df5e5e2d738efb9c276d954d754eb86918b/client/v3/watch.go#L325

and this [issue](https://github.com/etcd-io/etcd/issues/8980#issuecomment-350070282)

dm/dm/worker/server.go
```diff
@@ -141,9 +141,15 @@ func (s *Server) Start() error {
...
-		// TODO: handle fatal error from observeRelayConfig
-		//nolint:errcheck
-		s.observeRelayConfig(ctx, revRelay)
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				// relay don't affects sync, so no need to restart keepalive
+				log.L().Error("observeRelayConfig meet error will retry now", zap.Error(s.observeRelayConfig(ctx, revRelay)))
```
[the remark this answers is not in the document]
Ehco1996: maybe later, for now just hope it work agin soon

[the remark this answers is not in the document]
GMHDBJD: I think we did I need some retry policies, just like this https://github.com/pingcap/tiflow/blob/5c7d2388fef19f49cdc517a03acd303649fa4e5a/dm/dm/master/scheduler/scheduler.go#L1927-L1941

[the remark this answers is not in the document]
Ehco1996: we already do retry in `observeRelayConfig` see https://github.com/pingcap/tiflow/blob/c3e8be9add2df6ac2f426b8ac35dd650f3dea9d4/dm/dm/worker/server.go#L295

the outer for-loop just keep retry when network is poor

Ehco1996:
ping @GMHDBJD  @niubell  🧡

dm/pkg/ha/bound.go
lance6716: The link you given is not the version and path of etcd we use. 

https://github.com/pingcap/tiflow/blob/b3014061250cb3598c491624135217d4cd76d386/go.mod#L83

lance6716: w.streams will only be nil after `w.Close()`, I don't think that's the root cause

Ehco1996: if you have time can you help locate the problem, I don't have any ideas here at the moment, I'll send you the log in private chat~

Ehco1996:
from etcd
```
Canceled is used to indicate watch failure. If the watch failed and the stream was about to close, before the channel is closed, the channel sends a final response that has Canceled set to true with a non-nil Err()
```

seems grpc streams closed is not the root cause of #4548 , we upadted embed etcd in https://github.com/pingcap/tiflow/pull/4755, let's see if this can solve the problem 😵

lance6716:
> from etcd
> 
> ```
> Canceled is used to indicate watch failure. If the watch failed and the stream was about to close, before the channel is closed, the channel sends a final response that has Canceled set to true with a non-nil Err()
> ```
> 
> seems grpc streams closed is not the root cause of #4548 , we upadted embed etcd in #4755, let's see if this can solve the problem 😵

I tested the behaviour of watch, after I turned off the WiFi of my laptop, watch will retry infinitely rather than terminate.

Ehco1996:
close for no updated now

PR #4569, opened 2022-02-11
sink(ticdc): fix `Unknown system variable 'transaction_isolation'` for some mysql versions

### LINKED ISSUE #4504: sink to mariadb 10.6.5 fails with unknown system variable transaction_isolation
### What did you do?

Create a sink to downstream mariadb 10.6.5.

The downstream can be created without any errors.

### What did you see instead?

```bash
$ ./cdc cli changefeed list
[
  {
    "id": "test",
    "summary": {
      "state": "error",
      "tso": 431019101257728000,
      "checkpoint": "2022-02-07 12:16:27.000",
      "error": {
        "addr": "192.168.101.2:8300",
        "code": "CDC:ErrOwnerUnknown",
        "message": "[CDC:ErrMySQLConnectionError]fail to open MySQL connection: Error 1193: Unknown system variable 'transaction_isolation'"
      }
    }
  }
]

$ mysql -h 127.0.0.1 -P 3306 -u root -p -e "show variables like '%isolation%'";
Enter password:
+---------------+-----------------+
| Variable_name | Value           |
+---------------+-----------------+
| tx_isolation  | REPEATABLE-READ |
+---------------+-----------------+
```

Seems in mariadb 10.6.5 the variable name has been changed from "transaction_isolation" to "tx_isolation", and it causes the failure.

### Versions of the cluster

```console
```

```console
(paste TiCDC version here)
```

maxshuang:

Issue Number: close #4504

Since MySQL 5.7/MySQL 8/Mariadb may use transaction_isolation or tx_isolation, we first select the variable and use the right variable after that. 

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
fix `Unknown system variable 'transaction_isolation'` for some mysql versions
```


zhaoxinyu:
test case `TestApplyDMLs` failed. PTAL @[maxshuang](https://github.com/maxshuang)

maxshuang:
> test case `TestApplyDMLs` failed. PTAL @[maxshuang](https://github.com/maxshuang)

Thanks for reminding. I have fixed it.

zhaoxinyu (approved):

lance6716 (approved):

cdc/sink/mysql_params.go
```diff
@@ -287,6 +285,18 @@ func generateDSNByParams(
...
+	// transaction_isolation is mysql newly introduced variable and will vary from MySQL5.7/MySQL8.0/Mariadb
+	isolation, err := checkTiDBVariable(ctx, testDB, "transaction_isolation", defaultTxnIsolationRC)
+	if err != nil {
+		return "", err
+	}
+	if isolation != "" {
+		dsnCfg.Params["transaction_isolation"] = fmt.Sprintf(`"%s"`, defaultTxnIsolationRC)
+	} else {
+		dsnCfg.Params["tx_isolation"] = fmt.Sprintf(`"%s"`, defaultTxnIsolationRC)
+	}
```
amyangfei, line 298: What about checking and setting, in case of both `transaction_isolation` and `tx_isolation` are unknown system variable.

```go

err = checkAndSet("transaction_isolation")
// check error
err = checkAndSet("tx_isolation)
// check error

func checkAndSet(var string) error {
  isolation, err := checkTiDBVariable(ctx, testDB, var, defaultTxnIsolationRC)
  if err != nil {
    return err
  }
  if isolation != "" {
    dsnCfg.Params[var] = fmt.Sprintf(`"%s"`, defaultTxnIsolationRC)
  }
  return nil
}

```

maxshuang: It may happens for some very old mysql.  
But I think it's ok now.

amyangfei: well, these two variables co-exist in some MySQL versions.

amyangfei (approved):

Issue #4423, opened 2022-01-20
Tracking issue for multiple topics


hi-rustin:
- [x] Support for multiple topics in producer

PR #4598, opened 2022-02-15
sink/kafka(ticdc): remove useless partition flush logic

hi-rustin:

 - remove useless partition flush logic. We just need to count the message number.
 - Put the flush operation into the flush worker.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

None

Side effects

None

Related changes

None

### Release note 

```release-note

None
```


cdc/sink/producer/kafka/kafka.go
```diff
@@ -151,49 +153,28 @@ func (k *kafkaSaramaProducer) SyncBroadcastMessage(ctx context.Context, message
...
-		}
-	}
-	if noEventsToFLush {
-		// no events to flush
-		return nil
+	k.mu.Lock()
+	inflight := k.mu.inflight
+	immediateFlush := inflight == 0
+	if !immediateFlush {
+		k.mu.flushDown = down
```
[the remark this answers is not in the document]
hi-rustin: Waiting for the flush to complete. Select it in flush func.

cdc/sink/producer/kafka/kafka.go
```diff
@@ -117,6 +108,17 @@ func (k *kafkaSaramaProducer) AsyncSendMessage(ctx context.Context, message *cod
...
 
+	msg := &sarama.ProducerMessage{
+		Topic:     k.topic,
+		Key:       sarama.ByteEncoder(message.Key),
+		Value:     sarama.ByteEncoder(message.Value),
+		Partition: partition,
+	}
+	k.mu.Lock()
+	k.mu.inflight++
+	log.Info("emitting inflight messages to kafka", zap.Int64("inflight", k.mu.inflight))
```
amyangfei, line 119: Should we remove this log or use debug level?

cdc/sink/producer/kafka/kafka.go
```diff
@@ -151,49 +153,28 @@ func (k *kafkaSaramaProducer) SyncBroadcastMessage(ctx context.Context, message
...
-				return nil
-			}
-			return cerror.ErrKafkaFlushUnfinished.GenWithStackByArgs()
-		case <-k.flushedReceiver.C:
-			if !checkAllPartitionFlushed() {
-				continue flushLoop
-			}
-			return nil
-		}
+	log.Info("flush waiting for inflight messages", zap.Int64("inflight", inflight))
```
amyangfei, line 170: ditto

cdc/sink/producer/kafka/kafka.go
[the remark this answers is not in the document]
amyangfei: Does the variable mean `flush done`?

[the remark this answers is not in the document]
hi-rustin: yes

amyangfei (approved):

cdc/sink/producer/kafka/kafka.go
```diff
@@ -151,49 +153,28 @@ func (k *kafkaSaramaProducer) SyncBroadcastMessage(ctx context.Context, message
...
 
-	// checkAllPartitionFlushed checks whether data in each partition is flushed
-	checkAllPartitionFlushed := func() bool {
-		for i, target := range targetOffsets {
-			if target > atomic.LoadUint64(&k.partitionOffset[i].flushed) {
-				return false
-			}
-		}
-		return true
+	if immediateFlush {
```
zhaoxinyu, line 166: A race condition may happen in this place. If `k.mu` is unlocked and immediately locked by another goroutine doing `AsyncSendMessage`,  the `k.mu.inflight` will be incremented. As a result, the value hold by `immediateFlush` is no longer correct.

hi-rustin:
/hold

It can be problematic in our current multi-worker model, where we can have endless flush waits. For example, one worker is flushing and the other worker is asyncsend, so the first worker may never finish.

hi-rustin:
I did some local testing and the duration of flush is basically around 10 ms under `sysbench --db-driver=mysql --mysql-host=[127.0.0.1](http://127.0.0.1/) --mysql-port=4000 --mysql-user=root --threads=8 --mysql-db=test --tables=100 --table-size=100000 oltp_write_only prepare` load. So this should be acceptable for now.

In addition, I also tried to do a micro benchmark test, where we tried to add buffer to msgChan when the flush latency is large, but it is not effective either.( This is even an order of magnitude slower.) This optimization has little effect in the case of large writes. The flush latency itself should be acceptable when the writes are small.

```go

// Simulate flush delay
func (m *mockProducer) Flush(ctx context.Context) error {
	if len(m.mqEvent) != 0 {
		time.Sleep(10 * time.Millisecond)
	}
	m.mqEvent = make(map[int32][]*codec.MQMessage)
	m.flushed = true
	return nil
}


func BenchmarkFlush(b *testing.B) {
	worker, _ := newTestWorker()
	ctx, cancel := context.WithCancel(context.Background())
	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		defer wg.Done()
		err := worker.run(ctx)
			b.Error(err)
		}
	}()
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		for i := 0; i < 100000; i++ {
			b.StopTimer()
			if i%100 == 0 {
				time.Sleep(time.Duration(rand.Intn(30)) * time.Millisecond)
			}
			n := rand.Intn(10)
			b.StartTimer()

			if n%2 == 0 {
				worker.msgChan <- mqEvent{
					row: &model.RowChangedEvent{
						CommitTs: 1,
						Table:    &model.TableName{Schema: "a", Table: "b"},
						Columns:  []*model.Column{{Name: "col1", Type: 1, Value: "aa"}},
					},
					partition: int32(i % 10),
				}
			}

			if i%10 == 0 {
				worker.msgChan <- mqEvent{resolvedTs: model.Ts(i)}
			}
		}
	}

	b.StopTimer()
	cancel()
	wg.Wait()
}

```

cdc/sink/mq_flush_worker.go
```diff
@@ -43,21 +43,23 @@ type mqEvent struct {
 
 // flushWorker is responsible for sending messages to the Kafka producer on a batch basis.
 type flushWorker struct {
-	msgChan    chan mqEvent
+	msgChan       chan mqEvent
+	ticker        *time.Ticker
```
overvenus, line 47: When is it closed?

hi-rustin: I assume it shuts down when an error occurs or the flushworker is stopped? Because when an error occurs during the run, it is reported to the processor, which then tries to shut down the sink.

overvenus (approved):

PR #4599, opened 2022-02-16
loader(dm): skip feature for nil etcd client

lance6716:

When use DM as a library, sometimes etcd client is not valid thus we should support nil etcd client

### Check List 

Tests 

 - pass original tests

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`.
```


GMHDBJD (approved):
LGTM

Issue #4582, opened 2022-02-14
row delete event may lose field value when new collation is enabled.


glorv:
### What did you do?

Update tidb dep to newest master version (with new_collation is enabled by default), unit test `TestMounterDisableOldValue` fails consistently.  Seems delete event (with old value disabled and clustered_index enabled) is not compatible with new collation.

Root cause analysis:

When old value is disabled, cdc will parse corresponding pk fields from row handle by function [`tablecodec.DecodeHandleToDatumMap`](https://github.com/pingcap/tiflow/blob/fc3b5ad53eaa51d67360a0efb66b1cd04e6fc01f/cdc/entry/mounter.go#L266). But if new_collation is enabled, this function will skip string fields if the collation is not binary. Thus, the returned event will miss some fields. 
See:  https://github.com/pingcap/tidb/blob/aa24b9a7b40ee51671e43ee1ebea63fdbd0bc57f/tablecodec/tablecodec.go#L482

### What did you see instead?

unit test `TestMounterDisableOldValue` failed. See: https://ci.pingcap.net/blue/organizations/jenkins/ut-check/detail/ut-check/1290/pipeline

### Versions of the cluster

Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

```console
(paste TiDB cluster version here)
```

Upstream TiKV version (execute `tikv-server --version`):

```console
(paste TiKV version here)
```

TiCDC version (execute `cdc version`):

```console
(paste TiCDC version here)
```

PR #4602, opened 2022-02-16
ticdc(mounter): remove unnecessary delete event decode logic

hi-rustin:

Issue Number: close https://github.com/pingcap/tiflow/issues/4582

The original part of the logic was to find the handle column of the delete event when pkishandle. But now the TiCDC puller node pulls the old value by default, so we don't need to do this anymore.

### Check List 

Tests 

 - Integration test

Code changes

None

Side effects

None

Related changes

None

### Release note 

```release-note
None
```


tangenta (approved):

leoppro:
LGTM

liuzix (approved):

overvenus (approved):

PR #4603, opened 2022-02-16
kv(ticdc): replace the connArray with the connHeap

CharlesCheung96:

Issue Number: close #xxx

Replace the connArray with the connHeap in cdc/kv/grpc_pool_impl.go

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

### Release note 

```release-note
 `None`
```

### DIFF
--- a/cdc/kv/grpc_pool_impl.go
+++ b/cdc/kv/grpc_pool_impl.go
@@ -14,6 +14,7 @@
 package kv
 
 import (
+	"container/heap"
 	"context"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


ben1009:
what the problem the pr try to solve, big array?

CharlesCheung96:
> 

1. The changes here are mainly to make the code structure clearer and to optimize the load balancing strategy.
2. In the case of large array capacity, the performance of the heap will indeed be significantly higher than that of arrays. However, we may need to discuss this further, such as confirming the size of the connection pool in a typical case.

PR #4608, opened 2022-02-16
dm-master(dm): use advertise address for ETCD client

### LINKED ISSUE #4511: maybe adverseAddr should be used in some places insted of masterAddr
### What did you do?

I try to use DM with TLS (with --masterAddr=0.0.0.0:8261), but worker cannot connect to master, master log shows 
```
[grpclog.go:42] ["grpc: addrConn.createTransport failed to connect to {0.0.0.0:8261  <nil> 0 <nil>}. Err: connection error: desc = \"transport: authentication handshake failed: x509: certificate is valid for 127.0.0.1, 192.168.122.25, not 0.0.0.0\". Reconnecting..."] [component="embed etcd"]
```

It works after I add 0.0.0.0 to alt_names of cert, bit I think it should not be necessary

Then I search the code, I found somewhere, masterAddr was used as endpoint. maybe adverseAddr should be used instead. For example: https://github.com/pingcap/tiflow/blob/master/dm/dm/master/server.go#L214

related issue: https://github.com/pingcap/dm/issues/966

_No response_

### What did you see instead?

_No response_

### Versions of the cluster

v2.0.7 v5.3.0

```console
```

dsdashun:

Issue Number: close #4511 

Use the advertise address to places where an ETCD client may be needed.  In this PR, when we create an ETCD client, we need to use the advertise address; When we specify the address to join, we actually use the advertise address.   

Also the UTs are adapted to the new logic.  In the current tests, we set the advertise address and master address the same value so that the server start logic can retrieve the advertise address's value.  

What's more, some extra UTs are added to specify the master address as 0.0.0.0:<port>, and set the advertise address to the specific value.  Both TLS / non-TLS scenarios are covered.   

### Check List 

Tests 

 - Unit test

N/A

Side effects

N/A

Related changes

N/A

### Release note 

```release-note
None
```


lance6716 (approved):

lichunzhu (approved):
LGTM

PR #4630, opened 2022-02-18
sink(ticdc): remove useless op return value of AppendRowChangedEvent

hi-rustin:

remove useless op return value of AppendRowChangedEvent.

1. we do not need to return it when we append the msg to the buf.
2. the flushToProducer func never receive EncoderNeedSyncWrite op.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

None

Related changes

None

### Release note 

```release-note
None
```


overvenus (approved):

sdojjy (approved):
LGTM

zhaoxinyu (approved):

PR #4653, opened 2022-02-22
sink/codec(ticdc): remove useless MixedBuild and Reset funcs

hi-rustin:

remove useless MixedBuild and Reset funcs. We have removed the cdc log, so this is completely useless code.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

None

Related changes

None

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/sink/codec/json.go
+++ b/cdc/sink/codec/json.go
@@ -373,11 +373,6 @@ func mqMessageToDDLEvent(key *mqMessageKey, value *mqMessageDDL) *model.DDLEvent
 
 // JSONEventBatchEncoder encodes the events into the byte of a batch into.
 type JSONEventBatchEncoder struct {
-	// TODO remove deprecated fields
-	keyBuf            *bytes.Buffer // Deprecated: only used for MixedBuild for now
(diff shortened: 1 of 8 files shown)


cdc/sink/codec/json.go
```diff
@@ -540,70 +507,14 @@ func (d *JSONEventBatchEncoder) EncodeDDLEvent(e *model.DDLEvent) (*MQMessage, e
...
 // Size implements the EventBatchEncoder interface
 func (d *JSONEventBatchEncoder) Size() int {
-	return d.keyBuf.Len() + d.valueBuf.Len()
-}
-
-// Reset implements the EventBatchEncoder interface
-func (d *JSONEventBatchEncoder) Reset() {
-	d.keyBuf.Reset()
-	d.valueBuf.Reset()
+	return -1
```
amyangfei, line 517: Is there any side effect to `return -1` here (before it is removed)?

hi-rustin: No, because the size of our original json protocol is only for mix encoder. So there is no problem to return -1.

amyangfei (approved):

overvenus (approved):

Issue #4631, opened 2022-02-19
Separate read and write path for leveldb sorter


overvenus:
Separate read and write path to improve read and write throughput.

We plan to add two actors reader and writer, and let table sorter be a thin shim, which delegates AddEntry to writer and Output to reader.

| | |
| - | - |
|![image](https://user-images.githubusercontent.com/2150711/154794792-9c4cc5a3-aee6-4b89-a552-b281b5a78118.png)| ![image](https://user-images.githubusercontent.com/2150711/154796520-a26cd3dd-a91d-4e84-b187-72698d10500a.png) |

PR #4686, opened 2022-02-24
sorter/leveldb(ticdc): separate write from table sorter

overvenus:

Issue Number: ref #4631

It separate write from table sorter.

Note, for now it's dead code, will be enabled in later PRs.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/sorter/leveldb/message/task.go
```diff
@@ -27,11 +28,20 @@ type Task struct {
 	UID     uint32
 	TableID uint64
 
+	// Input unsorted event for writers.
+	// Sorter.AddEntry -> writer.
+	InputEvent *model.PolymorphicEvent
+	// Latest resolved ts / commit ts for readers.
+	// writer -> reader
+	ReadTs ReadTs
```
sdojjy, line 36: It makes the logic complex and unreadable to add so many fields in one struct, some fields are always nil in some cases.
And the `Task` is a part of `Message`,  The size will be larger and larger, maybe we should find a better way to extend the `Task` or `Message`.

cdc/sorter/leveldb/system/system.go
```diff
@@ -68,10 +70,14 @@ func NewSystem(dir string, memPercentage float64, cfg *config.DBConfig) *System
 		WorkerNumber(cfg.Count).Build()
 	compactSystem, compactRouter := actor.NewSystemBuilder("sorter-compactor").
 		WorkerNumber(cfg.Count).Build()
+	writerSystem, writerRouter := actor.NewSystemBuilder("sorter-writer").
```
sdojjy, line 73: we have  `sorter-db` and `sorter-writer`,   can you add some comments about `sorter-db`

cdc/sorter/leveldb/writer.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+
+	// Notify readers that there is something to read.
+	msg := actormsg.SorterMessage(message.Task{
+		UID:     w.uid,
+		TableID: w.tableID,
+		ReadTs: message.ReadTs{
+			MaxCommitTs:   maxCommitTs,
+			MaxResolvedTs: maxResolvedTs,
```
sdojjy, line 97: maybe the write task is not finished,  is it ok to send `maxResolvedTs` to the reader?

cdc/sorter/leveldb/writer.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/pingcap/tiflow/cdc/model"
+	"github.com/pingcap/tiflow/cdc/sorter/encoding"
+	"github.com/pingcap/tiflow/cdc/sorter/leveldb/message"
+	"github.com/pingcap/tiflow/pkg/actor"
+	actormsg "github.com/pingcap/tiflow/pkg/actor/message"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
+)
+
+// writer is a thin shim that batches, translates events into key vaule pairs
```
sdojjy, line 29:
```suggestion
// writer is a thin shim that batches, translates events into key-value pairs
```

cdc/sorter/leveldb/writer.go
overvenus: Yes, as reader/writer send tasks to the same leveldb, so tasks are ordered.
ReadTs will trigger reader to take iterator from leveldb, it happens after writer sends writes to leveldb.
And before leveldb takes iterators, it flushed all pending writes, so iterators see all writes.

cdc/sorter/leveldb/writer.go
```diff
@@ -0,0 +1,103 @@
...
+	// Notify readers that there is something to read.
+	msg := actormsg.SorterMessage(message.Task{
+		UID:     w.uid,
+		TableID: w.tableID,
+		ReadTs: message.ReadTs{
+			MaxCommitTs:   maxCommitTs,
+			MaxResolvedTs: maxResolvedTs,
+		},
+	})
+	// It's ok if send fails, as resolved ts events are received periodically.
```
zhaoxinyu, line 100: Is it possible that resolved ts events keep failing to be sent.

cdc/sorter/leveldb/message/task.go
overvenus: It's good point! Ideally tasks should be an enum type in rust, so tasks can be separated and more readable. Unfortunately, golang does not support it.

> And the Task is a part of Message, The size will be larger and larger, maybe we should find a better way to extend the Task or Message.

Maybe we should make Task be a pointer in Message? Though it makes memory allocation, not sure if it helps performance.

cdc/sorter/leveldb/writer.go
overvenus: Only if reader is blocked for a long time. Eg. reading iterator becomes very slow. I think it rarely happens.

cdc/sorter/leveldb/message/task.go
overvenus: Also see, notes in #4631

cdc/sorter/leveldb/writer.go
```diff
@@ -0,0 +1,110 @@
...
+		err := w.dbRouter.SendB(ctx, w.dbActorID, actormsg.SorterMessage(task))
+		if err != nil {
+			w.reportError("failed to send write request", err)
+			return false
+		}
+	}
+
+	// Notify reader that there is something to read.
+	//
+	// It's ok to noify reader immediately without waiting writes done,
```
zhaoxinyu, line 93:
```suggestion
	// It's ok to notify reader immediately without waiting writes done,
```

zhaoxinyu (approved):

sdojjy (approved):
lgtm

zhaoxinyu:
TestMySQLSinkFlushResovledTs failed.

PR #4703, opened 2022-02-25
owner(cdc): fix two metrics problems

### LINKED ISSUE #4714: TiCDC fails to report checkpoint metrics with pending scheduler operations
TiCDC does not report checkpoint in metrics when there is at least one table that is currently being scheduled. This makes the user unable to see the checkpoint when the cluster is starting up or has just failed over. In this situation, the checkpoint is expected to be blocked from advancing, but the user should have been able to observe it in the metrics.

middle_img_v2_ed2893df-c339-47d3-a1c4-a31b184f83cg

liuzix:
Issue Number: close #4714

- Fixed checkpoint reporting to Prometheus when the checkpoint cannot be advanced.
- Fixed "Table count maintained by owner" when the new scheduler is enabled.
- Optimized the data structures in `util.TableSet` so that reporting the table count metrics will not hinder performance.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test

The image below shows that this PR fix both problems. Before this PR, during scheduling (i.e. there are WIP tables), the checkpoint metrics should lack data.
image

Related changes

### Release note 

```release-note
Fix checkpoint metrics when tables are being scheduled.
```


cdc/scheduler/util/table_set.go
```diff
@@ -145,6 +162,13 @@ func (s *TableSet) RemoveTableRecord(tableID model.TableID) bool {
 	if len(captureIndexEntry) == 0 {
 		delete(s.captureIndex, record.CaptureID)
 	}
+
+	statusCountEntry, ok := s.statusCounts[record.CaptureID]
+	if !ok {
+		log.Panic("unreachable", zap.Int64("tableID", tableID))
+	}
+	statusCountEntry[record.Status]--
```
overvenus, line 170: Will it underflow?

cdc/owner/changefeed.go
```diff
@@ -209,18 +209,25 @@ func (c *changefeed) tick(ctx cdcContext.Context, state *orchestrator.Changefeed
 	if err != nil {
 		return errors.Trace(err)
 	}
+
+	pdTime, _ := ctx.GlobalVars().PDClock.CurrentTime()
```
overvenus, line 213: It might be an empty time.Time, will it cause underflow?

cdc/scheduler/util/table_set.go
liuzix: Other states maintained in `TableSet` is sufficient to guarantee that there is no underflow.

amyangfei (approved):

overvenus (approved):

PR #4741, opened 2022-03-02
cdc/codec: fix nil value in canal codec.

### LINKED ISSUE #4736: canal_json don't support nil value for uint64
### What did you do?

1. Setup kafka inetgration test
2. Create table t1(id int primary key, c1 unsigned int);
3. insert into t1(id) values(1);

kafka sink work normally

### What did you see instead?

Error like:
[2022/03/01 18:27:11.489 +08:00] [PANIC] [canal.go:167] [“unsigned value not in type uint64”] [column="{“name”:“mch_id”,“type”:8,“flag”:193,“value”:null}"]

**Detail**:
trigger senarios:
column define default null at creating + insert without explicit column

https://github.com/pingcap/tiflow/blob/9fdebd6cab82ed32f406485474505f59e56a7e5e/cdc/entry/mounter.go#L526
image

https://github.com/pingcap/tiflow/blob/9fdebd6cab82ed32f406485474505f59e56a7e5e/cdc/sink/codec/canal.go#L166
image

### Versions of the cluster

```console
v5.4.0
```

```console
v5.4.0
```

```console
v5.4.0
```

3AceShowHand:

Issue Number: close #4736 

check nil before use the `Column.Value` in Canal 

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
fix `Canal-JSON` meet `unsigned` SQL Type and nullable, which cause CDC server panic.
```


liuzix (approved):

cdc/sink/codec/canal.go
```diff
@@ -162,6 +162,9 @@ func getJavaSQLType(c *model.Column, mysqlType string) (result JavaSQLType, err
 	// for **unsigned** integral types, type would be `uint64` or `string`. see reference:
 	// https://github.com/pingcap/tiflow/blob/1e3dd155049417e3fd7bf9b0a0c7b08723b33791/cdc/entry/mounter.go#L501
 	// https://github.com/pingcap/tidb/blob/6495a5a116a016a3e077d181b8c8ad81f76ac31b/types/datum.go#L423-L455
+	if c.Value == nil {
```
maxshuang, line 165: I notice v5.4.0 make some uint64 convertion. 
Can you simplely  explain why other branches don't need this convertion?

cdc/sink/codec/canal_flat_test.go
```diff
@@ -118,6 +118,12 @@ func (s *canalFlatSuite) TestNewCanalFlatMessage4DML(c *check.C) {
 			continue
 		}
 
+		// for `Column.Value` is nil, which mean's it is nullable, set the value to `""`
+		if obtainedValue == nil {
+			c.Assert(item.expectedValue, check.Equals, "")
```
maxshuang, line 123: Will this empty value related to the column type？

3AceShowHand: no, I think all column type can be `null`.

PR #4743, opened 2022-03-02
sink(ticdc): add OnTableSinkCreated and OnTableSinkDestroyed hooks for sink

hi-rustin:

add OnTableSinkCreated and OnTableSinkDestroyed hooks for sink.

### Check List 

Tests 

 - [ ] Unit test

Code changes

Side effects

None

Related changes

None

### Release note 

```release-note

None
```


cdc/sink/dispatcher/event_router.go
```diff
@@ -56,27 +57,92 @@ func (r *partitionDispatchRule) fromString(rule string) {
...
+	defaultTopic string
+	rules        []struct {
 		partitionDispatcher partition.Dispatcher
 		topicDispatcher     topic.Dispatcher
 		filter.Filter
 	}
 }
 
+// NewEventRouter creates a new EventRouter
+func NewEventRouter(cfg *config.ReplicaConfig, defaultTopic string) (*EventRouter, error) {
```
hi-rustin, line 69: The check for defaultTopic is insufficient here, but it's not related to this PR, so I'll fix it in the next PR.

cdc/owner/schema.go
```diff
@@ -85,11 +86,31 @@ func (s *schemaWrap4Owner) AllPhysicalTables() []model.TableID {
 	return s.allPhysicalTablesCache
 }
 
+// AllTableNames returns the table names of all tables.
```
amyangfei, line 89: This function is not thread safe, better to add comment to avoid misuse later

amyangfei, line 89:
```suggestion
// AllTableNames returns the table names of all tables that are being replicated.
```

cdc/owner/schema_test.go
```diff
@@ -72,6 +72,23 @@ func TestAllPhysicalTables(t *testing.T) {
...
+	defer helper.Close()
+	ver, err := helper.Storage().CurrentVersion(oracle.GlobalTxnScope)
+	require.Nil(t, err)
+	schema, err := newSchemaWrap4Owner(helper.Storage(), ver.Ver, config.GetDefaultReplicaConfig())
+	require.Nil(t, err)
+	require.Len(t, schema.AllTableNames(), 0)
+	// add normal table
+	job := helper.DDL2Job("create table test.t1(id int primary key)")
+	require.Nil(t, schema.HandleDDL(job))
+	require.Equal(t, schema.AllTableNames(), []model.TableName{{Schema: "test", Table: "t1"}})
```
amyangfei, line 86: Better to change the variable sequence, since
```go
func Equal(t TestingT, expected interface{}, actual interface{}, msgAndArgs ...interface{})
```

amyangfei (commented):
Besides how does passing `all table names` solve the table missing problem

cdc/owner/schema_test.go
hi-rustin: Oops! Thanks!

hi-rustin:
> Besides how does passing `all table names` solve the table missing problem

I think we just need to maintain all the table information that corresponds at that time? It is in the tick before the processing of the ddl, so can we assume that the tables that need to be synchronized at the current moment are those in AllTableNames?

amyangfei (approved):

cdc/owner/ddl_sink.go
```diff
@@ -144,12 +149,16 @@ func (s *ddlSinkImpl) run(ctx cdcContext.Context, id model.ChangeFeedID, info *m
...
+				checkpointTs := s.mu.checkpointTs
 				if checkpointTs == 0 || checkpointTs <= lastCheckpointTs {
+					s.mu.Unlock()
 					continue
 				}
+				tables := s.mu.currentTableNames
+				s.mu.Unlock()
 				lastCheckpointTs = checkpointTs
-				if err := s.sink.EmitCheckpointTs(ctx, checkpointTs); err != nil {
+				if err := s.sink.EmitCheckpointTs(ctx, checkpointTs, tables); err != nil {
```
zhaoxinyu, line 161: Is it possible that the tablesName hold by `tables` is already stale? eg. New tables are added or existing tables are dropped.

hi-rustin: It will be possible, but it is unavoidable? Because we are asynchronous flush.
So we can only update yes tables with checkpoint ts at that time as much as possible?

cdc/owner/schema.go
```diff
@@ -85,11 +86,32 @@ func (s *schemaWrap4Owner) AllPhysicalTables() []model.TableID {
 	return s.allPhysicalTablesCache
 }
 
+// AllTableNames returns the table names of all tables that are being replicated.
+// NOTICE: AllTableNames is not thread-safe.
+func (s *schemaWrap4Owner) AllTableNames() []model.TableName {
+	if s.allTableNamesCache != nil {
+		return s.allTableNamesCache
```
zhaoxinyu, line 93: Why should we get the tableNames from this cache instead of fetching the latest tables from `s.schemaSnapshot.Tables()`?

cdc/owner/ddl_sink.go
hi-rustin: Also we are emitting checkpoint ts first and then executing ddl. so the problem is not too serious. However, there is still a possibility of inconsistency due to asynchronous reasons. For example, a new table is added within 1 second. But I think it takes time for ddl to be distributed in sink, so all we can do is to use the current tables as much as possible?

cdc/owner/schema.go
hi-rustin: Because we may not have ddl for a long time, using cache can improve performance.

cdc/owner/ddl_sink.go
zhaoxinyu: > Also we are emitting checkpoint ts first and then executing ddl. so the problem is not too serious. However, there is still a possibility of inconsistency due to asynchronous reasons. For example, a new table is added within 1 second. But I think it takes time for ddl to be distributed in sink, so all we can do is to use the current tables as much as possible?

If missing some checkpointTs for those new tables is tolerant by downstream, we don't have to maintain the latest tableNames. BTW, I'm wondering how is the checkpointTs handled by downstream?

hi-rustin: It will be used to sort with commit ts.

hi-rustin: > If missing some checkpointTs for those new tables is tolerant by downstream

Theoretically we shouldn't emit checkpoint ts before ddl broadcasts to kafka, so it only sends more, not less?

cdc/sink/dispatcher/event_router.go
```diff
@@ -56,27 +57,92 @@ func (r *partitionDispatchRule) fromString(rule string) {
...
 }
 
+// GetActiveTopics returns a list of the corresponding topics
+// for the tables that are actively synchronized.
+func (s *EventRouter) GetActiveTopics(activeTables []model.TableName) []string {
+	topics := make([]string, 0)
+	topicsMap := make(map[string]bool, len(activeTables))
+	for _, table := range activeTables {
+		topicDispatcher, _ := s.matchDispatcher(table.Schema, table.Table)
+		topicName := topicDispatcher.Substitute(table.Schema, table.Table)
```
zhaoxinyu, line 126: Maybe we can add a warning if the `topicName` generated by topic expression equals to `s.defaultTopic`.

hi-rustin: Should this be expected?

zhaoxinyu: > Should this be expected?

This is not expected. But for the current implementation, we cannot prevent users specifying a default topic which accidentally equals to `topicName`

zhaoxinyu: Or maybe we can add checks somewhere?

hi-rustin: Added. https://github.com/pingcap/tiflow/pull/4743/commits/20a9a413b85ddfde483e370b4ff457f0e223f9f0

zhaoxinyu (approved):

hi-rustin (commented):
Self check

hi-rustin:
/hold

/cc @amyangfei 

Please take a look again. Thanks!

cdc/owner/changefeed.go
```diff
@@ -189,7 +193,12 @@ func (c *changefeed) tick(ctx cdcContext.Context, state *orchestrator.Changefeed
 	default:
 	}
 
-	c.sink.emitCheckpointTs(checkpointTs, c.schema.AllTableNames())
+	// This means that the cached DDL has been executed,
+	// and we need to use the latest table names.
+	if c.currentTableNames == nil {
+		c.currentTableNames = c.schema.AllTableNames()
+	}
+	c.sink.emitCheckpointTs(checkpointTs, c.currentTableNames)
```
amyangfei, line 201: If a DDL is being executed asynchronously, could the `emitCheckpointTs` run with the same `checkpointTs` but different `c.currentTableNames`?

hi-rustin: Before the DDL really executes, we need to use the old currentTableNames, otherwise we will have problems with sending early or stopping early.

amyangfei (approved):

PR #4744, opened 2022-03-02
syncer(dm): fix log flood and performance when frequently call Snapshot

### LINKED ISSUE #4619: replication is slow and many “checkpoint has no change, skip sync flush checkpoint” in DM log
### What did you do?

This issue may be triggered when DM uses GTID replication and with one of below scenarios:
- upstream has a big DML event which contains millions of row changes. **normally this would not happen**, for [default max size of any ROWS_EVENT is 8192](https://dev.mysql.com/doc/refman/5.7/en/replication-options-binary-log.html#option_mysqld_binlog-row-event-max-size)
- an extra large transaction totally has millions of row changes

or DM uses position-based replication and with:
- upstream has a big DML event which contains millions of row changes, **normally this would not happen**

https://asktug.com/t/topic/573236

_No response_

### What did you see instead?

The bug phenomenon is that DM replicates data very slow and there are many “checkpoint has no change, skip sync flush checkpoint” in DM log

### Versions of the cluster

```console
v5.4.0
```

```console
```

lance6716:

Issue Number: close #4619 

always update `lastSnapshotCreationTime` when call `Snapshot`. Now `checkShouldFlush` will return false after seeing `lastSnapshotCreationTime`, so we will not call `Snapshot` too frequently

### Check List 

Tests 

   generate two big transaction and set checkpoint-flush-interval to 5s. Use a complex GTID set which consists of 10 uuid part to start task, meet the first big transaction, create 500 tables and a DML for each table to record 500 table checkpoint, meet the second big transaction
   before this PR, QPS of first big transaction is 16.8k and second one is 3.7k
   after this PR, QPS are both 16.5k

Code changes

Side effects

Related changes

### Release note 

```release-note
Fix that there're lot of log of "checkpoint has no change, skip sync flush checkpoint" and performance may drop
```


dm/syncer/checkpoint.go
```diff
@@ -373,6 +373,15 @@ func (cp *RemoteCheckPoint) Snapshot(isSyncFlush bool) *SnapshotInfo {
 	cp.RLock()
 	defer cp.RUnlock()
 
+	flushGlobalPoint := cp.globalPoint.outOfDate() || cp.globalPointSaveTime.IsZero() || (isSyncFlush && cp.needFlushSafeModeExitPoint.Load())
+
+	// When there's a big transaction in GTID-based replication, after the checkpoint-flush-interval every row change
+	// will trigger a check thus call Snapshot, the Snapshot should be light-weighted. We only check the global point to
+	// return quickly.
+	if !flushGlobalPoint {
```
Ehco1996, line 381: so we dont't need tablepoint level snapshot now ?

lance6716: we save table checkpoint just 8 lines below

db-will: when we received sharded ddl for multiple table on upstream database, using pessimistic mode, will we have a case that global point is not updated, but table checkpoint is updated.

will we have any issues?

lance6716: @GMHDBJD PTAL. Maybe we can add a `IsDirty`/`NeedFlush` in checkpoint interface?

lance6716: in v5.3.0, we always flush checkpoint every 30s no matter the content has been changed.

https://github.com/pingcap/tiflow/blob/20626babf21fc381d4364646c40dd84598533d66/dm/syncer/checkpoint.go#L680-L684

The new async checkpoint flush feature added another check of global checkpoint and table checkpoints. At the moment, this PR wants to only add another check of global checkpoint, but can't handle the pessimistic sharding case. For now I think an atomic boolean IsDirty can represent global or table checkpoint is updated before.

I also want to make sure that, why we add another check compared with v5.3.0? @db-will

GMHDBJD: > will we have a case that global point is not updated, but table checkpoint is updated.

Yes, but we no need to flush table points frequently. The table checkpoint will keep updated in memory, flush when task stopped when existing. If worker was killed, we rely on safemode to handle it. So I think table point can be flushed when global checkpoint updated after 30 second.

lance6716: @GMHDBJD what's the proposed behaviour? 
- flush every 30 seconds no matter checkpoint has changed
- after 30 seconds, check checkpoint is dirty and the dirty variable is turned on when global or table checkpoints are set

GMHDBJD: Keep same as 2.x. Only flushed after 30 seconds and cp.globalPoint.outOfDate(), which is your current implement.

D3Hunter: > IsDirty/NeedFlush in checkpoint interface?

would be more intuitive

lance6716: > Keep same as 2.x. Only flushed after 30 seconds and cp.globalPoint.outOfDate(), which is your current implement.

In v2.0.7, I think the behaviour is
1. in addJob/checkWait, mostly we will flush checkpoints every `checkpoint-flush-interval`
2. in FlushPointsExcept, we generate flush SQL by inspecting outOfDate() for each table point, outOfDate()/SaveTimeIsZero/NeedFlushSafeModeExitPoint for global point.

My current code adds a "step 1.5", which is after every `checkpoint-flush-interval`, we check outOfDate()/SaveTimeIsZero/NeedFlushSafeModeExitPoint of global point to determine if goes to FlushPointsExcept in step 2. That will cause no flushing in pessimistic sharding, if DM-worker also failed to flush when exit, after restart the progress is replayed. Not a big problem but I prefer to avoid it.

Currently I prefer to add IsDirty/NeedFlush to checkpoint interface.

lance6716:
manual test passed and PR description updated, PTAL @GMHDBJD @D3Hunter

GMHDBJD (approved):
LGTM

D3Hunter (approved):
LGTM

db-will (approved):
Looks good!

dm/syncer/checkpoint.go
db-will: > in v5.3.0, we always flush checkpoint every 30s no matter the content has been changed.
> 
> https://github.com/pingcap/tiflow/blob/20626babf21fc381d4364646c40dd84598533d66/dm/syncer/checkpoint.go#L680-L684
> 
> The new async checkpoint flush feature added another check of global checkpoint and table checkpoints. At the moment, this PR wants to only add another check of global checkpoint, but can't handle the pessimistic sharding case. For now I think an atomic boolean IsDirty can represent global or table checkpoint is updated before.
> 
> I also want to make sure that, why we add another check compared with v5.3.0? @db-will

the extra checking is added for preventing calls to async flush continuously, when flushing checkpoint is slow.

PR #4750, opened 2022-03-03
pkg/pipeline(ticdc): optimize CPU usage by withMessage

### LINKED ISSUE #4749: [TiCDC] pipeline.withMessage consumes too much CPU
image

```go
func withMessage(ctx NodeContext, msg Message) NodeContext {
	return messageContext{
		NodeContext: ctx,
		message:     msg,
	}
}
```

This function will cause one allocation per call, because the interface type `NodeContext` needs to hold a pointer to `messageContext`. The frequency of calling this function is equal to the total QPS by row of the TiCDC node, leading to it consuming a disproportionate amount of CPU.

liuzix:

Issue Number: close #4749 

- Avoid allocation for the most common use case.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/server.go
+++ b/cdc/server.go
@@ -257,9 +257,8 @@ func (s *Server) etcdHealthChecker(ctx context.Context) error {
 			for _, pdEndpoint := range s.pdEndpoints {
 				start := time.Now()
 				ctx, cancel := context.WithTimeout(ctx, time.Second*10)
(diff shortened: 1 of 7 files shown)


overvenus (approved):
LGTM

sdojjy (approved):

PR #4767, opened 2022-03-03
validation(dm): full validation mode part 2

### LINKED ISSUE #4488: validation(dm): full validation mode

D3Hunter:

Issue Number: close #4488

- add validation worker which receives row changes from validator, then merge or add into pending row changes.
- at regular interval, validation worker'll validate those changes by querying downstream
- add status routine to print statistics info in validator

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`
```


dm/syncer/data_validator.go
```diff
@@ -77,12 +76,11 @@ type tableChange struct {
 type rowChange struct {
 	table      *validateTableInfo
 	key        string
-	pkValues   []string
+	pkValues   []string // todo: might be removed in later pr
```
buchuitoudegou, line 79: Currently, the `pkValues` is used to construct the query condition whereas the `key` is the concatenation of `pkValues`. Would it be better to deprecate the `key` field but concat it every time we gonna use it? Otherwise how can we construct the `Cond` struct?🤔

dm/syncer/validate_worker.go
```diff
@@ -0,0 +1,348 @@
...
+		select {
+		case change, ok := <-vw.rowChangeCh:
+			if !ok {
+				break outer
+			}
+			// todo: limit number of pending rows
+			vw.updateRowChange(change)
+			vw.receivedRowCount.Inc()
+		case <-time.After(vw.interval):
+			err := vw.validateTableChange()
```
buchuitoudegou, line 86: Can the validation be triggered by the number of pending row? What if the worker gets too much `rowChange` in a short period? Setting a batch threshold can restrict the resource upper bound that the worker consumes at a time.

dm/syncer/validate_worker.go
```diff
@@ -0,0 +1,348 @@
...
+	columns := strings.Join(columnNames, ", ")
+	rowsQuery := fmt.Sprintf("SELECT /*!40001 SQL_NO_CACHE */ %s FROM %s WHERE %s",
+		columns, fullTableName, cond.GetWhere())
+	rows, err := vw.conn.QuerySQL(tctx, rowsQuery, cond.GetArgs()...)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	result := make(map[string][]*sql.NullString)
```
buchuitoudegou, line 283: Similarly, if we scan through and fetch all rows (rather than iterate them one by one) in big data situation, will it be a bit risky.

D3Hunter: yes, add a todo here

https://github.com/buchuitoudegou/tiflow/blob/ebba8bb19d0083a8653b815617ff55cea4a50fe8/dm/syncer/validate_worker.go#L134-L145

dm/syncer/validate_worker.go
D3Hunter: add a todo to limit number of pending rows here

D3Hunter: add a todo to limit number of pending rows here

https://github.com/buchuitoudegou/tiflow/blob/ebba8bb19d0083a8653b815617ff55cea4a50fe8/dm/syncer/validate_worker.go#L82

might be better to trigger validation at that time, will leave a todo.

but what if the validation didn't reduce the number of pending row change(suppose most rows are changed again), and retry might not work then, so we have to keep accumulate row changes until we approach progress of syncer(row changes get updated and most row change can validated successfully)

dm/syncer/data_validator.go
D3Hunter: `pkValues` can also be recreated from `rowChange.data`+`table.pkIndices`, but key is used more often

dm/syncer/validate_worker.go
buchuitoudegou: Making the worker being driven by a fixed number of rows in addition to polling the channel every once in awhile might solve this (i.e. restrict the number of rows) and be more intuitive.

D3Hunter: you mean the way we do in prototype?

in that case we're keeping 2 pending row changes set, one is current accumulating(with a max batch size), another is what's left from previous validation which needs retry and its size may exceed the batch size. it's more complex, and we cannot merge row changes to eliminate duplicates, also for the second one we'll face the same problem as in `todo`.

buchuitoudegou: It is too complicated the way we did in prototype. I think handling the situation that accumulating too many rows is plausible at the time we receive the row change from the validator (according to the newly added todo).

D3Hunter: add in https://github.com/pingcap/tiflow/pull/4767/commits/36f1486309c578f90405fab8242f56250c42ae55

dm/syncer/validate_worker.go
```diff
@@ -0,0 +1,380 @@
...
+	rowDifferent
+)
+
+type validateFailedRow struct {
+	tp      validateFailedType
+	dstData []*sql.NullString
+}
+
+type validateWorker struct {
+	cfg               config.ValidatorConfig
```
Ehco1996, line 60: Is it recommended to use pointer types here?

dm/syncer/validate_worker.go
```diff
@@ -0,0 +1,380 @@
...
+		pendingChangesMap: make(map[string]*tableChange),
+		batchSize:         maxBatchSize,
+	}
+}
+
+func (vw *validateWorker) run() {
+outer:
+	for {
+		select {
+		case change, ok := <-vw.rowChangeCh:
```
Ehco1996, line 91: do we need to select `vw.ctx.Done` here ?

dm/syncer/validate_worker.go
```diff
@@ -0,0 +1,380 @@
...
+		vw.pendingChangesMap[fullTableName] = change
+	}
+	if val, ok := change.rows[row.key]; ok {
+		val.data = row.data
+		val.tp = row.tp
+		val.lastMeetTS = row.lastMeetTS
+		val.failedCnt = 0 // clear failed count
+	} else {
+		change.rows[row.key] = row
+		vw.pendingRowCount.Inc()
```
Ehco1996, line 130: why we need this `pendingRowCount`  i don'e see any usage of this var

dm/syncer/validate_worker.go
```diff
@@ -0,0 +1,380 @@
...
+		var insertUpdateChanges, deleteChanges []*rowChange
+		for _, r := range tblChange.rows {
+			if r.tp == rowDeleted {
+				deleteChanges = append(deleteChanges, r)
+			} else {
+				insertUpdateChanges = append(insertUpdateChanges, r)
+			}
+		}
+		rows := make(map[string]*rowChange)
+		if len(insertUpdateChanges) > 0 {
```
Ehco1996, line 147: seems we can move check `len(deleteChanges) > 0` and `len(insertUpdateChanges) > 0` into func`validateRowChanges `, so logic here can be more simple.

dm/syncer/validate_worker.go
```diff
@@ -0,0 +1,380 @@
...
+			}
+			failedRowCnt += int64(len(rows))
+		}
+	}
+	vw.pendingRowCount.Store(failedRowCnt)
+	vw.pendingChangesMap = failedChanges
+	return nil
+}
+
+func (vw *validateWorker) validateRowChanges(table *validateTableInfo, rows []*rowChange, deleteChange bool) (map[string]*validateFailedRow, error) {
```
Ehco1996, line 181: for `deleteChange bool`, will more row change types be introduced later?

dm/syncer/validate_worker.go
```diff
@@ -0,0 +1,380 @@
...
+			return nil, err
+		}
+		for k, v := range failedRows {
+			res[k] = v
+		}
+	}
+	return res, nil
+}
+
+func (vw *validateWorker) validateRowChangesInner(table *validateTableInfo, rows []*rowChange, deleteChange bool) (map[string]*validateFailedRow, error) {
```
Ehco1996, line 200:
```suggestion
func (vw *validateWorker) batchValidateRowChanges(table *validateTableInfo, rows []*rowChange, deleteChange bool) (map[string]*validateFailedRow, error) {
```

dm/syncer/validate_worker.go
```diff
@@ -0,0 +1,380 @@
...
+				continue
+			}
+		}
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func (vw *validateWorker) getTargetRows(cond *Cond) (map[string][]*sql.NullString, error) {
```
Ehco1996, line 298: for deleted row type check, only selec pk is enough? maybe add another func to do this

Ehco1996 (commented):
will review test later

dm/syncer/validate_worker.go
D3Hunter: in next pr about checkpoint and pereist error rows(validate failed after N tries), need to save it's current data into a meta table, so user can view the detail

dm/syncer/validate_worker.go
D3Hunter: insert/update can be handled the same way, delete need to be handled separately.

no, i think

dm/syncer/validate_worker.go
D3Hunter: need to print internal status regularly or send metrics to prometheus later, not used now.

dm/syncer/validate_worker.go
D3Hunter: worker need to drain this channel, in later pr checkpoint will wait on a wait group just like what syncer checkpoint does.

dm/syncer/validate_worker.go
D3Hunter: both are ok, i think

dm/syncer/validate_worker_test.go
```diff
@@ -0,0 +1,380 @@
...
+	tableInfo1 := genValidateTableInfo(t, tbl1.Schema, tbl1.Name,
+		"create table tbl1(a int primary key, b varchar(100))")
+	tableInfo2 := genValidateTableInfo(t, tbl2.Schema, tbl2.Name,
+		"create table tbl2(a varchar(100) primary key, b varchar(100))")
+	tableInfo3 := genValidateTableInfo(t, tbl3.Schema, tbl3.Name,
+		"create table tbl3(a varchar(100) primary key, b varchar(100))")
+
+	cfg := genSubtaskConfig(t)
+	_, mock, err := conn.InitMockDBFull()
+	mock.MatchExpectationsInOrder(false)
```
Ehco1996, line 49: why need add this?

Ehco1996 (approved):
LGTM

dm/syncer/validate_worker_test.go
D3Hunter: default, sqlmock match sqls in the order we write in code. but in validation it's iterating over a map, the order of sql may different, it's ok in `debug mode`, but unstable in `run mode`

dm/syncer/validate_worker.go
```diff
@@ -0,0 +1,380 @@
...
+
+	err = rows.Scan(colValsI...)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	result := make([]*sql.NullString, len(cols))
+	for i := range colVals {
+		result[i] = &sql.NullString{
+			String: string(colVals[i]),
```
lance6716, line 355: I'm not sure why we need NullString here, because we Scan the row into []byte, "NULL" and NULL from SQL may become same thing. Do you want to Scan into NullString?

Also require an unit test to verify this

D3Hunter: > "NULL" and NULL from SQL may become same thing.

after scan, null value is database corespond to `nil` in golang.

lance6716 (approved):

PR #4768, opened 2022-03-04
scheduler(cdc): add ProcessorEpoch

### LINKED ISSUE #4769: [TiCDC] (unreleased feature) High inter-node latency causes ErrProcessorDuplicateOperations
- When the new scheduler is enabled, high inter-node latency can cause ErrProcessorDuplicateOperations.
- Scenario: 100K tables with 8 TiCDC nodes. Injects a latency of `2000ms` jitter `10ms` between TiCDC nodes. Then restarts all TiCDC nodes all at once.

```
[
  {
    "id": "test-1",
    "summary": {
      "state": "error",
      "tso": 431564666934132811,
      "checkpoint": "2022-03-03 14:22:34.894",
      "error": {
        "addr": "new-sched-test-cdc-3.liuzixiong.tispace:8300",
        "code": "CDC:ErrProcessorDuplicateOperations",
        "message": "[CDC:ErrProcessorDuplicateOperations]table processor duplicate operation, table-id: 26194"
      }
    }
  }
]
```
- Root cause: Processor may send `Sync` to Owner to tell the owner about the processor's internal state. However, the internal state can be updated by a in-flight owner's dispatch command, making the owner's knowledge of the processor' state inconsistent.
- Severity: (1) This problem rarely occurs if the latency is low (sub 100ms) (2) This problem pauses the changefeed, but the changefeed can retry and will eventually succeed if the latency recovers to normal. (3) This problem will not cause data corruption.

liuzix:

Issue Number: close #4769

- Added `epoch` in `Sync` and `DispatchTable` messages, so that outdated dispatches will be ignored by the processor.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test

Restarts all TiCDC nodes at once with a latency of 2000ms injected between each pair of them. The changefeed did not pause and report error, but recovered quickly.
image

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Fix ErrProcessorDuplicateOperations when new scheduler is enabled (disabled by default)
```

### DIFF
--- a/cdc/scheduler/agent.go
+++ b/cdc/scheduler/agent.go
@@ -17,9 +17,8 @@ import (
 	"sync"
 	"time"
 
-	"github.com/google/uuid"
-
 	"github.com/edwingeng/deque"
+	"github.com/google/uuid"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	"github.com/pingcap/tiflow/cdc/model"
--- a/cdc/scheduler/agent.go
+++ b/cdc/scheduler/agent.go
@@ -493,3 +493,9 @@ func (a *BaseAgent) getEpoch() model.ProcessorEpoch {
 
 	return a.epoch
 }
+
+// CurrentEpoch is a public function used in unit tests for
+// checking epoch-related invariants.
+func (a *BaseAgent) CurrentEpoch() model.ProcessorEpoch {
+	return a.getEpoch()
+}
--- a/cdc/scheduler/agent.go
+++ b/cdc/scheduler/agent.go
@@ -17,6 +17,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/uuid"
+
 	"github.com/edwingeng/deque"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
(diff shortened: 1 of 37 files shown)


cdc/scheduler/agent.go
```diff
@@ -272,6 +289,7 @@ func (a *BaseAgent) processOperations(ctx context.Context) error {
 	for tableID, op := range a.tableOperations {
 		switch op.status {
 		case operationReceived:
+			a.logger.Info("Agent start processing operation", zap.Any("op", op))
```
amyangfei, line 292: Is it acceptable that the operation related log is O(#table) scale.

cdc/scheduler/agent.go
```diff
@@ -456,3 +478,23 @@ func (a *BaseAgent) currentOwner() model.CaptureID {
 
 	return a.ownerInfo.OwnerCaptureID
 }
+
+func (a *BaseAgent) resetEpoch() {
+	a.epochMu.Lock()
+	defer a.epochMu.Unlock()
+
+	a.epoch = uuid.New().String()
```
amyangfei, line 486: We only need the unique constraint of this epoch, and don't need the serialization guarantee. Should we add comment about it.

cdc/scheduler/agent.go
liuzix: Without this log, it would be difficult to trace scheduling problems.

cdc/scheduler/agent.go
liuzix: Sure.

amyangfei (approved):

3AceShowHand (approved):

PR #4781, opened 2022-03-04
scripts(ticdc): limit file width

### LINKED ISSUE #4780: Limit code line width for readability
As title.

overvenus:

Issue Number: close #4780

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


hi-rustin (commented):
Can we use [lll](https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/lll.go)?

overvenus:
> Can we use [lll](https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/lll.go)?

lll checks whole file, it may be annoying when we only change one line but it ask us to reformat the whole file.

I'd like to make the reformat processing as smooth as possible by checking added lines only.

scripts/check-diff-line-width.sh
```diff
@@ -0,0 +1,29 @@
...
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+git diff master -U0 cdc pkg |
```
hi-rustin, line 15: What happens on the release branch?

Also, I get check-leaktest-added effects when I test locally and it may not get run?(check-leaktest-added will terminate the check)

overvenus: This check is filtered when it bases on a release branch, see makefile.

docs/swagger/swagger.json
```diff
@@ -711,6 +711,9 @@
                     "items": {
                         "type": "string"
                     }
+                },
```
hi-rustin, line 714: I don't know if we should revert to that change.

scripts/add-leaktest.sh
```diff
@@ -30,5 +30,3 @@ for i in $@; do
 	fi
 	rm -f $i~
 done
-
-git --no-pager diff --exit-code
```
hi-rustin, line 34: After it's removed, does it not know that we need to add `lestleak.AfterTest`?

hi-rustin (approved):

Issue #4360, opened 2022-01-17
Support S3 for DM dump and load


WizardXiao:
### Is your feature request related to a problem?

Upstream data volumes can be large and variable, making them difficult and costly to manage and plan for using EBS. So users want to use DM on s3 storage.

### Describe the feature you'd like

DM supports s3 storage to dump and load data files.

PR #4786, opened 2022-03-07
DM: lightning checkpoint use s3 and change lightning tls dump path

WizardXiao:

Issue Number: ref #4360

1. use lightning version which support s3 when use file checkpoint.
2. change lightning tls config dump path to current path instead of use loader's dir which may use s3 path.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


dm/loader/lightning.go
```diff
@@ -234,15 +235,15 @@ func (l *LightningLoader) restore(ctx context.Context) error {
...
 
-		// TODO lightning checkpoint will support s3 in other pr
-		if storage.IsS3Path(l.cfg.LoaderConfig.Dir) {
-			cfg.Checkpoint.Enable = false
-		} else {
-			cfg.Checkpoint.Driver = lcfg.CheckpointDriverFile
-			cpPath := filepath.Join(l.cfg.LoaderConfig.Dir, lightningCheckpointFileName)
-			cfg.Checkpoint.DSN = cpPath
-			cfg.Checkpoint.KeepAfterSuccess = lcfg.CheckpointOrigin
+		cfg.Checkpoint.Driver = lcfg.CheckpointDriverFile
```
Ehco1996, line 238: why hard code to file checkpoint ?  seems lightning support use `CheckpointDriverMySQL` checkpoint?

lance6716: To reduce the complexibility that checkpoint exists but dump files doesn't exist

dm/loader/lightning.go
```diff
@@ -234,15 +235,15 @@ func (l *LightningLoader) restore(ctx context.Context) error {
...
+		var cpPath string
+		cpPath, err = storage.AdjustPath(l.cfg.LoaderConfig.Dir, lightningCheckpointFileName, string(filepath.Separator), false)
```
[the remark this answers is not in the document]
WizardXiao: if it is a part in url, may be file still at url.path, and this function will handle both url and locl file.
set `needCheckDuplicate` be `false` is to follow the previous which has no check.

dm/dm/config/security.go
```diff
@@ -82,18 +82,33 @@ func (s *Security) LoadTLSContent() error {
...
+				return err
+			}
+		}
 		if err := utils.WriteFileAtomic(s.SSLCA, s.SSLCABytes, 0o600); err != nil {
 			return err
 		}
 	}
 	if s.SSLCert == "" || !utils.IsFileExists(s.SSLCert) {
 		s.SSLCert = path.Join(baseDirPath, "cert.pem")
+		if !utils.IsDirExists(baseDirPath) {
```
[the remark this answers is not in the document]
WizardXiao: ok, i will add a new function.

dm/loader/lightning.go
WizardXiao: our configuration about `loaders` do not support other checkpoint too.

dm/pkg/storage/utils.go
```diff
@@ -26,8 +26,34 @@ import (
...
+// 		u.Path = u.Path + "." + uniqueID
+// 		// u.String will return escaped url and can be used safely in other steps
+// 		return u.String(), err
+// 	}
+// 	return rawURL, nil
+// }
+
 // AdjustPath adjust rawURL, add uniqueId as path suffix.
-func AdjustPath(rawURL string, uniqueID string) (string, error) {
+func AdjustPath(rawURL string, uniqueID string, separator string, needCheckDuplicate bool) (string, error) {
```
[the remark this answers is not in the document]
WizardXiao: ok, after some discussion, i will use original fuction type `func AdjustPath(rawURL string, uniqueID string) (string, error)`, `separator ` will be a part of `uniqueID` which is determined by the caller, `needCheckDuplicate` will be the general logic in this function.

dm/dm/config/security.go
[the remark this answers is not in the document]
WizardXiao: when i fix this function, i notice `utils.IsDirExists` will return false when the path is a file, then it will return error when `os.MkdirAll`, and this error may be as expect.

dm/dm/worker/subtask.go
```diff
@@ -151,7 +150,7 @@ func (st *SubTask) initUnits(relay relay.Process) error {
 	// NOTE: because lightning not support init tls with raw certs bytes, we write the certs data to a file.
 	if st.cfg.NeedUseLightning() && st.cfg.To.Security != nil {
 		// NOTE: LoaderConfig.Dir is always not empty because we only dump certs when we use lightning.
-		if err := st.cfg.To.Security.DumpTLSContent(filepath.Join(st.cfg.LoaderConfig.Dir, "..")); err != nil {
+		if err := st.cfg.To.Security.DumpTLSContent("./" + loader.TmpTLSConfigPath + "_" + st.cfg.Name); err != nil {
```
lance6716, line 153: In the old PR I'm concerned about the certificate files are leaked to other machines by dumping them. Not sure what's our opinions now.

dm/loader/lightning.go
```diff
@@ -234,15 +235,15 @@ func (l *LightningLoader) restore(ctx context.Context) error {
...
-		if storage.IsS3Path(l.cfg.LoaderConfig.Dir) {
-			cfg.Checkpoint.Enable = false
-		} else {
-			cfg.Checkpoint.Driver = lcfg.CheckpointDriverFile
-			cpPath := filepath.Join(l.cfg.LoaderConfig.Dir, lightningCheckpointFileName)
-			cfg.Checkpoint.DSN = cpPath
-			cfg.Checkpoint.KeepAfterSuccess = lcfg.CheckpointOrigin
+		cfg.Checkpoint.Driver = lcfg.CheckpointDriverFile
+		var cpPath string
+		cpPath, err = storage.AdjustPath(l.cfg.LoaderConfig.Dir, string(filepath.Separator)+lightningCheckpointFileName)
```
lance6716, line 240: found that in SubTaskConfig.Adjust, we already make sure `l.cfg.LoaderConfig.Dir` is unique.

dm/pkg/storage/utils.go
```diff
@@ -39,13 +39,13 @@ func AdjustPath(rawURL string, uniqueID string) (string, error) {
 	if u.Scheme == "" {
 		// avoid duplicate add uniqueID
 		if !strings.HasSuffix(rawURL, uniqueID) {
-			return rawURL + "." + uniqueID, nil
```
lance6716, line 42: no need to change this behaviour now

dm/dm/worker/subtask.go
WizardXiao: The reason we need to fix is `LoaderConfig.Dir` may be s3 path, but we don't implement tls at s3 in lightning, so we need to dump it to a local path but not `LoaderConfig.Dir`.

dm/loader/lightning.go
WizardXiao: yes, when add suffix to `l.cfg.LoaderConfig.Dir` to construct a file for `cfg.Checkpoint.DSN`, we actually don't need to check unique, because `l.cfg.LoaderConfig.Dir`'s suffix is 'taskid_sourceid' and `lightningCheckpointFileName` is `tidb_lightning_checkpoint.pb`, but if we check unique may be ok and it will make `storage.AdjustPath`'s  logic be pithier.

dm/pkg/storage/utils.go
WizardXiao: this is because `cfg.Checkpoint.DSN` is a file and it used to be like `l.cfg.LoaderConfig.Dir/fileName`，so if we keep this format，there may be `/xxxx.xx` format in `uniqueID`.

dm/loader/lightning.go
lance6716: `cpPath := filepath.Join(l.cfg.LoaderConfig.Dir, lightningCheckpointFileName)` won't cause any problem and your change introduced https://github.com/pingcap/tiflow/pull/4786#discussion_r825559207

dm/dm/worker/subtask.go
lance6716: I mean if we only keep the certificate in memory, there's less risk to leak it. It's an old problem but may be a security issue @sunzhaoyang

dm/loader/lightning.go
Ehco1996: seems we can keep `cpPath := filepath.Join(l.cfg.LoaderConfig.Dir, lightningCheckpointFileName)`

dm/dm/worker/subtask.go
Ehco1996: btw i forget why not just dump certs to `st.cfg.LoaderConfig.Dir`  😵😵

WizardXiao: this is because `st.cfg.LoaderConfig.Dir` may be a path in s3

dm/loader/lightning.go
WizardXiao: Use `storage.AdjustPath` is because here want to use Lightning checkpoint support S3 in this [PR](https://github.com/pingcap/tidb/pull/32373)

Ehco1996 (approved):
LGTM

dm/dm/worker/subtask.go
WizardXiao: i add some notes to reduce ambiguity.

dm/loader/lightning.go
WizardXiao: i add some notes to reduce ambiguity.

lance6716: @WizardXiao I still can't understand why we need `AdjustPath` again. We have made sure `l.cfg.LoaderConfig.Dir` is unique in SubTaskConfig.Adjust

WizardXiao: this is Lightning checkpoint need a fileName, it used to be `l.cfg.LoaderConfig.Dir/tidb_lightning_checkpoint.pb` and `AdjustPath` for this time use to add  `/tidb_lightning_checkpoint.pb` and make a new path, but this will not change `l.cfg.LoaderConfig.Dir`. Maybe this function name is easy to make ambiguity.

lance6716: > this is because Lightning checkpoint need a fileName

why can't use `cpPath := filepath.Join(l.cfg.LoaderConfig.Dir, lightningCheckpointFileName)`

WizardXiao: this is because `l.cfg.LoaderConfig.Dir` may be a s3 path.

lance6716 (approved):

Issue #4699, opened 2022-02-25
[CC scenario] some captures' sink don't work


Tammyxia:
### What did you do?

- CDC 0 & 7 sink write are 0 rows/s:
![image](https://user-images.githubusercontent.com/78345569/155660314-09de80c7-546b-4031-b4aa-8f27e7b3a8ae.png)
- In the same time, we can see all captures are alive:
![image](https://user-images.githubusercontent.com/78345569/155660532-6f346b2b-d289-4f46-907b-80a0e11aef99.png)

### What did you see instead?

As above.

### Versions of the cluster

Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

```console
v5.4.0
```

Upstream TiKV version (execute `tikv-server --version`):

```console
v5.4.0
```

TiCDC version (execute `cdc version`):

```console
v5.4.0
```

PR #4819, opened 2022-03-09
pkg,sorter,processor(ticdc): support graceful close

overvenus:

Issue Number: ref #4699 

Add Close method to actor, let actor close gracefully.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/processor/pipeline/table_actor.go
```diff
@@ -159,6 +159,11 @@ func NewTableActor(cdcCtx cdcContext.Context,
 	return table, nil
 }
 
+// Close implements Actor interface.
+// TODO: implements table actor stop here.
+func (t *tableActor) Close() {
```
overvenus, line 164: PTAL @sdojjy

3AceShowHand:
can you add more description about the PR, why it can solve the problem that `some captures' sink don't work`

pkg/actor/actor.go
```diff
@@ -41,17 +42,26 @@ type Actor interface {
 	// We choose message to have a concrete type instead of an interface to save
 	// memory allocation.
 	Poll(ctx context.Context, msgs []message.Message) (running bool)
+
+	// Close is called after Poll retrurns false,
```
3AceShowHand, line 46: `retrurns` -> `return`

pkg/actor/system.go
```diff
@@ -68,18 +90,58 @@ func (p *proc) batchReceiveMsgs(batchMsg []message.Message) int {
...
+		&p.state, uint64(procStateMailboxClosed), uint64(procStateClosed)) {
+		p.actor.Close()
+	}
+}
+
+type readyState int
+
+const (
+	// Running state, system needs to poll actors.
+	readyStateRunning readyState = 0
```
3AceShowHand, line 138: how about `readyStateRunning readyState = itoa`, and then ignore the following enum number.

3AceShowHand (approved):
Rest LGTM

overvenus:
> can you add more description about the PR, why it can solve the problem that `some captures' sink don't work`

The issue will be fixed in a later PR. This PR provides graceful close mechanism which is required by the fix.

overvenus:
> > can you add more description about the PR, why it can solve the problem that `some captures' sink don't work`
> 
> The issue will be fixed in a later PR. This PR provides graceful close mechanism which is required by the fix.

See PR https://github.com/pingcap/tiflow/pull/4831

cdc/processor/pipeline/table_actor.go
sdojjy: ok, I will fire a pr after this is merged

pkg/actor/system.go
```diff
@@ -533,6 +605,12 @@ func (s *System) poll(ctx context.Context, id int) {
 					s.handleFatal(
 						"try to remove non-existent actor", p.mb.ID())
 				}
+
+				// Drop all remaining messages.
+				if remainMsgCount != 0 {
+					// TODO: we may need to releaes resources hold by messages.
```
sdojjy, line 611:
```suggestion
					// TODO: we may need to release resources hold by messages.
```

pkg/actor/actor.go
```diff
@@ -41,17 +42,26 @@ type Actor interface {
 	// We choose message to have a concrete type instead of an interface to save
 	// memory allocation.
 	Poll(ctx context.Context, msgs []message.Message) (running bool)
+
+	// Close is called after Poll returns false,
+	// or actor system is stopping and all message has been received.
+	//
+	// Close must be idempotent and nonblocking.
+	Close()
```
sdojjy, line 50: 1.  for normal cases, resources can be released before `Poll` returns `false`.
2. The name `Close` confused me,  it sounds like we can Close the actor directly, but it's just a callback.

pkg/actor/system.go
```diff
@@ -400,13 +459,21 @@ func (s *System) Start(ctx context.Context) {
 }
 
 // Stop the system, cancels all actors. It should be called after Start.
+// Messages sent before this call will be receive by actors.
```
sdojjy, line 462: seems we need `StopNow`, sometimes we need it waste of time to send amessages if the whole system is stopped

sdojjy (approved):

Issue #4126, opened 2021-12-29
[TiCDC] Enable new scheduler (p2p) on master branch


liuzix:
Remaining work to be done for enabling p2p on master branch

P1 items (before branching out release-5.4) :
- [x] Fix move_table test case #4120 
- [x] Add Grafana panels #4108 
- [x] Fix some `cdc cli` commands #3931 
- [x] Add a user-configurable switch to turn on p2p #4151 

P2 items:
- [ ] Refine logging (PR pending).
- [ ] Unit case improvements for existing packages (owner and processor) (PR pending)
- [ ] Enables p2p by default on master branch.

overvenus:
It has been enabled since v6.0.0.

PR #4836, opened 2022-03-10
*(ticdc): enable new scheduler by default

liuzix:

Issue Number: ref #4126

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
Enables a new scheduler.
```


cdc/owner/changefeed_test.go
```diff
@@ -138,6 +138,9 @@ func createChangefeed4Test(ctx cdcContext.Context, t *testing.T) (*changefeed, *
 	}, func() DDLSink {
 		return &mockDDLSink{}
 	})
+	cf.newScheduler = func(ctx cdcContext.Context, startTs uint64) (scheduler, error) {
+		return newSchedulerV1(), nil
+	}
```
amyangfei, line 143: Is this test only available with old scheduler?

amyangfei (approved):

cdc/owner/changefeed_test.go
liuzix: The correctness of the scheduler is not the goal of these tests. The old scheduler is like a "mock scheduler" here.

pkg/cmd/server/server_test.go
```diff
@@ -179,7 +179,8 @@ func TestParseCfg(t *testing.T) {
 			TableActor: &config.TableActorConfig{
 				EventBatchSize: 32,
 			},
-			EnableDBSorter: false,
+			EnableDBSorter:     false,
+			EnableNewScheduler: true,
```
overvenus, line 183: Could you remove `EnableNewScheduler` in debug.go ?

https://github.com/pingcap/tiflow/blob/92fc4c4593cafe3d81936de3f4eb5775683a8bae/pkg/config/debug.go#L35

overvenus (approved):

Issue #4835, opened 2022-03-10
Provide a DM Upstream Simulator for Testing Incremental Replication from Upstream


dsdashun:
### Is your feature request related to a problem?

Currently, when doing some tests for DM on incremental replication, we need to let the upstream continuously generate binlog streams.  So we need a tool to continuously simulate the upstream workloads. The current solutions are not very convenient.  (See the alternatives discussion below ) 

### Describe the feature you'd like

The DM can provide a simulator, which has several features that can solve some problems mentioned above:
* It can continuously apply meaningful modifications on upstream tables with the table schema provided.
* It can define specific workload easily.
* It can simulate batch DDL changes on several sharded tables with one click.  
* After the table schema change, the simulation will use the latest table structures.

### Describe alternatives you've considered

Usually, we simulate upstream workload either by using some benchmark tools like sysbench, or by using some random SQL generating programs like sql-smith.  However, this is not very convenient in some cases.  
* For those benchmark tools, the table schemas are pre-defined.  If we need to provide a bunch of binlog stream from upstream clusters with specified table schemas, we need to modify the code.
* For random sql-generator, the table schema can be defined by our own.  However, when the generated SQL is executed on the upstream, usually no data is actually modified, because the filter clause is purely randomly generated.  So it cannot provide a stable stream of binlogs from the upstream.  
* Some workload is hard to simulate from existing tools.  For example, if we want to simulate a transaction with one insert of table A followed by updating several records on table B, and at last delete that inserted row, the existing tools can hardly do this, we need to write our own code to achieve this kind of simulation. 
* If the upstream clusters have several sharded tables, there is no way to batch apply the DDLs on the set of sharded tables with one command.  
* There is no way to simulate DMLs affecting the online table schema change.  For example, we first let the upstream simulate binlog streams on the current table structure, then we do the DDL on the upstream to change the table structure.  After that we expect the upstream to simulate DMLs using the latest table structures on the fly, without modifying any code.

PR #4838, opened 2022-03-10
simulator(dm): implement some core components

dsdashun:

Issue Number: ref #4835 

Implemented some core components of the simulator:
* modification candidate pool (MCP)
* random SQL generator based on unique keys

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes

N/A

### Release note 

```release-note
None
```


dm/simulator/internal/mcp/mcp.go
```diff
@@ -0,0 +1,132 @@
...
+	"go.uber.org/zap"
+)
+
+var (
+	initRandSeedOnce sync.Once
+	gMCPRand         *rand.Rand
+	gMCPRandLock     sync.Mutex
+)
+
+func init() {
```
lance6716, line 33: this function is only called once by go runtime, no need to use a `sync.Once` I think

And `var gMCPRand = rand.New(rand.NewSource(time.Now().Unix()))` should have same effect, no need to explicitly use an `init()`

dm/simulator/internal/mcp/mcp.go
```diff
@@ -0,0 +1,132 @@
...
+	uk.Lock()
+	uk.RowID = currentLen
+	uk.Unlock()
+	mcp.keyPool = append(mcp.keyPool, uk)
+	return nil
+}
+
+// DeleteUK deletes the unique key from the MCP.
+// It will get the row ID of the UK and delete the UK on that position.
+// If the actual value is different from the input UK, the element will still be deleted.
```
lance6716, line 91: So we can only pass in an index, not an *UniqueKey

dm/simulator/internal/mcp/mcp.go
```diff
@@ -0,0 +1,132 @@
...
+		return errors.Trace(ErrInvalidRowID)
+	}
+	deletedUK = mcp.keyPool[deleteIdx]
+	curLen := len(mcp.keyPool)
+	lastUK := mcp.keyPool[curLen-1]
+	lastUK.Lock()
+	lastUK.RowID = deleteIdx
+	lastUK.Unlock()
+	mcp.keyPool[deleteIdx] = lastUK
+	mcp.keyPool[curLen-1] = deletedUK
```
lance6716, line 121: no need to keep it?

dm/simulator/internal/mcp/mcp_test.go
```diff
@@ -0,0 +1,165 @@
...
+				"id": i,
+			},
+		})
+	}
+	s.mcp = mcp
+}
+
+func (s *testMCPSuite) TestMCPNextUK() {
+	for i := 0; i < 10; i++ {
+		theUK := s.mcp.NextUK()
```
lance6716, line 50: we can assert something, like `theUK` is not nil. In CI there's not a human reading the output.

dm/simulator/internal/mcp/mcp_test.go
```diff
@@ -0,0 +1,165 @@
...
+	)
+	curPoolSize = len(s.mcp.keyPool)
+	for i := 0; i < 5; i++ {
+		err = s.mcp.AddUK(&UniqueKey{
+			RowID: -1,
+			Value: map[string]interface{}{
+				"id": rand.Int(),
+			},
+		})
+		assert.Nil(s.T(), err)
```
lance6716, line 68: we use require package rather than assert package elsewhere. I guess the main reason is that when the first error happens, there is no need to continue a broken test.

And since it's in a suite, we can use `s.Nil`

lance6716 (commented):
will review later

dm/simulator/internal/mcp/mcp.go
dsdashun: Originally, I planned to add the logic that if the RowID is invalid, the UK values are searched in MCP to get the actual index and delete.  So I kept the interface like this, in case in the further this logic is added back.

dm/simulator/internal/mcp/mcp.go
dsdashun: Right.  I'll delete it.

dm/simulator/internal/mcp/mcp_test.go
dsdashun: OK.  I'll refine the tests.

dm/simulator/internal/mcp/mcp_test.go
dsdashun: OK.  I'll fix that.

dm/simulator/internal/sqlgen/impl_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	uk = &mcp.UniqueKey{
+		RowID: -1,
+		Value: map[string]interface{}{
+			"id":      123,
+			"abcdefg": 321,
+		},
+	}
+	sql, err = g.GenUpdateRow(uk)
+	assert.Nil(s.T(), err)
```
D3Hunter, line 131: looks odd, maybe check whether length of uk values equals to length of table uk columns in `generateWhereClause`?

dm/simulator/internal/mcp/mcp_test.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	// test delete at random position
+	for i := 0; i < 5; i++ {
+		theUK := s.mcp.NextUK()
+		deleteRowID := theUK.RowID
+		err = s.mcp.DeleteUK(theUK)
+		assert.Nil(s.T(), err)
+		assert.Equal(s.T(), s.mcp.keyPool[deleteRowID].RowID, deleteRowID, "the new added UK's row ID is abnormal")
+		s.T().Logf("new UK after delete on the index %d: %v\n", deleteRowID, s.mcp.keyPool[deleteRowID])
+	}
```
D3Hunter, line 99: maybe we should check rowid of uk in MCP is still in serial order after all those test

dm/simulator/internal/mcp/mcp_test.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func (s *testMCPSuite) TestMCPAddDeleteInParallel() {
```
D3Hunter, line 102: since operations on MCP is serialized(always locked), i think this test is not needed, but it's ok to keep it.

dm/simulator/internal/mcp/mcp.go
```diff
@@ -0,0 +1,132 @@
...
+
+// ModificationCandidatePool is the core container storing all the current unique keys for a table.
+type ModificationCandidatePool struct {
+	sync.RWMutex
+	keyPool []*UniqueKey
+}
+
+// NewModificationCandidatePool create a new MCP.
+func NewModificationCandidatePool() *ModificationCandidatePool {
+	theKeyPool := make([]*UniqueKey, 0, 8192)
```
GMHDBJD, line 47: Can capacity be configured?

dm/simulator/internal/mcp/mcp.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// NextUK randomly picks a unique key in the MCP.
+func (mcp *ModificationCandidatePool) NextUK() *UniqueKey {
+	mcp.RLock()
+	defer mcp.RUnlock()
+	if len(mcp.keyPool) == 0 {
+		return nil
+	}
+	gMCPRandLock.Lock()
```
D3Hunter, line 60: why not let each MCP has its own rand obj?

dm/simulator/internal/mcp/mcp.go
```diff
@@ -0,0 +1,132 @@
...
+// NewModificationCandidatePool create a new MCP.
+func NewModificationCandidatePool() *ModificationCandidatePool {
+	theKeyPool := make([]*UniqueKey, 0, 8192)
+	return &ModificationCandidatePool{
+		keyPool: theKeyPool,
+	}
+}
+
+// NextUK randomly picks a unique key in the MCP.
+func (mcp *ModificationCandidatePool) NextUK() *UniqueKey {
```
GMHDBJD, line 54: we may need to allocate some uks in batch if the performance is not satisfied

GMHDBJD (commented):
How about add a sqlgen demo?

dm/simulator/internal/mcp/mcp.go
dsdashun: Yes, it is temporarily hard-coded.  I'll add a parameter

dm/simulator/internal/mcp/uk.go
```diff
@@ -0,0 +1,155 @@
...
+	"strings"
+	"sync"
+)
+
+// UniqueKey is the data structure describing a unique key.
+type UniqueKey struct {
+	// It inherits a RWMutex, which is used to modify the metadata inside the UK struct.
+	sync.RWMutex
+	// rowOPLock is the lock for operating on the corresponding row of the unique key
+	rowOPLock sync.Mutex
```
lance6716, line 27: What's RowOperation 🤔

dm/simulator/internal/mcp/uk.go
```diff
@@ -0,0 +1,155 @@
...
+	for k, v := range uk.value {
+		fmt.Fprintf(&b, "%s = %v; ", k, v)
+	}
+	fmt.Fprintf(&b, ") }")
+	return b.String()
+}
+
+// IsValueEqual tests whether two UK's value parts are equal.
+func (uk *UniqueKey) IsValueEqual(otherUK *UniqueKey) bool {
+	if otherUK == nil {
```
lance6716, line 135: maybe check thisUK is nil as well

dm/simulator/internal/mcp/uk_test.go
```diff
@@ -0,0 +1,176 @@
...
+			"col3": 321,
+			"col1": col1Value,
+			"col2": col2Value,
+		},
+	}
+	s.Assert().Equal(false, uk1.IsValueEqual(uk5), "uk1 should not equal uk5 on value")
+	s.Assert().Equal(false, uk5.IsValueEqual(uk1), "uk5 should not equal uk1 on value")
+}
+
+func TestUniqueKeySUite(t *testing.T) {
```
lance6716, line 174:
```suggestion
func TestUniqueKeySuite(t *testing.T) {
```

lance6716 (commented):
will review sqlgen soon

dm/simulator/internal/mcp/mcp.go
```diff
@@ -0,0 +1,117 @@
...
+	defer mcp.RUnlock()
+	return len(mcp.keyPool)
+}
+
+// AddUK adds the unique key into the MCP.
+// It has side effect: the input UK's row ID will be changed.
+func (mcp *ModificationCandidatePool) AddUK(uk *UniqueKey) error {
+	mcp.Lock()
+	defer mcp.Unlock()
+	if len(mcp.keyPool)+1 > cap(mcp.keyPool) {
```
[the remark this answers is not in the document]
dsdashun: This is to limit the overall size of the upstream simulating tables so that the storage is not overwhelmed for some simulating workloads like continuous INSERT operations.  In the design, if the MCP is touching the capacity threshold, an eviction process will be triggered.  Currently the eviction logic is not implemented.  So I just report an error here.

dm/simulator/internal/mcp/uk.go
dsdashun: A row operation is a DML operation on a specific row: UPDATE or DELETE.  In simulation process, If one transaction is updating a row in MCP, I don't want other workers to also modify/delete this row.  So I introduce a lock here.

dm/simulator/internal/mcp/errors.go
```diff
@@ -0,0 +1,28 @@
...
+
+package mcp
+
+import (
+	"github.com/pingcap/errors"
+)
+
+var (
+	// ErrMCPCapacityFull means the capacity of the modification candidate pool (MCP) is full.
+	ErrMCPCapacityFull = errors.New("the capacity of the modification candidate pool is full")
```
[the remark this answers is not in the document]
dsdashun: I put the errors here because the simulator is a relatively independent project now.  So I put all the error definitions in the project scope.  We can move the error definitions into the `pkg/errors` further if needed.

dm/simulator/internal/mcp/mcp_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/pingcap/tiflow/dm/pkg/log"
+)
+
+type testMCPSuite struct {
+	suite.Suite
+	mcp *ModificationCandidatePool
+}
+
+func (s *testMCPSuite) SetupSuite() {
+	s.Require().Nil(log.InitLogger(&log.Config{}))
```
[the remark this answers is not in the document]
dsdashun: If I understand correctly, by default, `s.Nil()` is actually `s.Assert().Nil()`.  I choose to use the `Require()` here because I expect the log initialization should finish before doing rest of the work.  

However, I can remove all those places using `s.Assert().XXX()` with `s.XXX()` so that the code is more concise.

dm/simulator/internal/mcp/uk.go
dsdashun: After further investigation, I think the row operation lock is not needed for the sake of performance.  I'll remove that.

dm/simulator/internal/mcp/uk.go
```diff
@@ -0,0 +1,155 @@
...
+	rowID int
+	// value is the real value of all the UK columns.
+	// The key is the column name, the value is the real value.
+	value map[string]interface{}
+}
+
+// NewUniqueKey creates a new unique key instance.
+// the map values are cloned into the new UK instance,
+// so that the further changes in the value map won't affect the values inside the UK.
+func NewUniqueKey(rowID int, value map[string]interface{}) *UniqueKey {
```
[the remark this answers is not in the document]
dsdashun: Right.  It is possible that there are two items in the MCP having the same UK value.  But that only means this UK value has more chances to be picked, and the simulation logic can still run without problems.

What's more, usually we will insert the row value in the upstream DB first, and then add the corresponding UK in the MCP.  If the insertion failed with a duplicated key error, the logic will be terminated for this operation, and the UK will not be inserted into the MCP.  That means in the normal cases, all the UK values in the MCP are truly unique.

dm/simulator/internal/mcp/uk.go
dsdashun: OK.  I'll fix that.

dm/simulator/internal/sqlgen/sqlgen.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/pingcap/tiflow/dm/simulator/internal/config"
+	"github.com/pingcap/tiflow/dm/simulator/internal/mcp"
+)
+
+// SQLGenerator contains all the operations for generating SQLs.
+type SQLGenerator interface {
+	// GenTruncateTable generates a TRUNCATE TABLE SQL.
+	GenTruncateTable() (string, error)
```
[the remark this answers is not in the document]
dsdashun: I'll take a look at that later.  Currently I still plan to use the string as the output, because it can fulfill my needs for the project, and it will not introduce extra dependencies.

dm/simulator/internal/mcp/mcp_test.go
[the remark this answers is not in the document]
lance6716: Sorry I forget it's Assert not Require.

dm/simulator/internal/sqlgen/impl.go
```diff
@@ -0,0 +1,255 @@
...
+	compareExprs := make([]*ast.BinaryOperationExpr, 0)
+	// iterate the existing UKs, to make sure all the uk columns has values
+	for ukColName := range g.ukMap {
+		val, ok := theUK[ukColName]
+		if !ok {
+			log.L().Error(ErrUKColValueNotProvided.Error(), zap.String("column_name", ukColName))
+			return nil, errors.Trace(ErrUKColValueNotProvided)
+		}
+		compareExprs = append(compareExprs, &ast.BinaryOperationExpr{
+			Op: opcode.EQ,
```
lance6716, line 90: Not sure if we should there's a chance that the column value is NULL. If so, we should fallback to use all columns to identify a row (or support another UK), and we should use `IS NULL` to check the NULL value.

See https://github.com/pingcap/tiflow/blob/fdf32a6f7ed2489ab1c0cdb980c22025d9cf6d75/pkg/sqlmodel/row_change.go#L228

dm/simulator/internal/sqlgen/impl.go
```diff
@@ -0,0 +1,255 @@
...
+			},
+			R: ast.NewValueExpr(val, "", ""),
+		})
+	}
+	return generateCompoundBinaryOpExpr(compareExprs), nil
+}
+
+func generateCompoundBinaryOpExpr(compExprs []*ast.BinaryOperationExpr) ast.ExprNode {
+	switch len(compExprs) {
+	case 0:
```
lance6716, line 104: Do we support using a empty UniqueKeyColumnNames?

For other cases this will cause ill-formed "AND" Op, we should panic or log some error

lance6716 (commented):
rest lgtm

lance6716:
ping @dsdashun

lance6716:
What should we do about this PR @dsdashun

lance6716 (approved):

dm/simulator/internal/sqlgen/sqlgen.go
```diff
@@ -0,0 +1,38 @@
...
+// Package sqlgen is the logic for generating different kinds of SQL statements.
+package sqlgen
+
+import (
+	"github.com/pingcap/tiflow/dm/simulator/internal/config"
+	"github.com/pingcap/tiflow/dm/simulator/internal/mcp"
+)
+
+// SQLGenerator contains all the operations for generating SQLs.
+type SQLGenerator interface {
```
Ehco1996, line 23: need a `GenCreateTableSQL` func to make sqlgen more friendly

maybe do this later

Ehco1996 (approved):
LGTM

lance6716:
/hold

I think we can merge it after the v6.1 LTS release

D3Hunter:
> /hold
> 
> I think we can merge it after the v6.1 LTS release

seems doesn't matter, this module is quite standalone

lance6716:
/unhold
just learned that code freeze for v6.1 is 5/19

Issue #4260, opened 2022-01-10
binlog inject command supports inject at dml 


WizardXiao:
### Is your feature request related to a problem?

Now `binlog inject` command just can inject at ddl position, but user need inject at dml when handling error.

### Describe the feature you'd like

as title.

okJiang:
can we post internal wiki here?

WizardXiao:
i will add more integration test in next pr. please do not close issue.

PR #4848, opened 2022-03-11
DM: support inject at dml when open GTID

WizardXiao:

Issue Number: ref #4260

1. As inject handle will add the original dml to the last event to be inject,  we check weather dml was executed will not check position but gtid and suffix when open gitd, but the last event's suffix will set to be 0, and it will be ignore.
2. Change the location saved table checkpoint or global checkpoint to be a cloned location which suffix is 0, so when check the last event's location will not ignore it, because it will be equal and dml will execute.
3. Change the compare when flush checkpoint to just use compare position, because there is no suffix and the gtid may be equal when handle error.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


dm/syncer/err-operator/operator.go
```diff
@@ -177,6 +177,8 @@ func (h *Holder) MatchAndApply(startLocation, endLocation binlog.Location, curre
 	h.mu.Lock()
 	defer h.mu.Unlock()
 
+	h.logger.Info("try to match and apply a operator", zap.Stringer("startlocation", startLocation), zap.Stringer("endlocation", endLocation), zap.Any("currentEvent", currentEvent))
```
buchuitoudegou, line 180:
```suggestion
	h.logger.Info("try to match and apply an operator", zap.Stringer("startlocation", startLocation), zap.Stringer("endlocation", endLocation), zap.Any("currentEvent", currentEvent))
```

dm/syncer/err-operator/operator.go
```diff
@@ -215,7 +217,7 @@ func (h *Holder) MatchAndApply(startLocation, endLocation binlog.Location, curre
 		}
 	}
 
-	h.logger.Info("match and apply a operator", zap.Stringer("startlocation", startLocation), zap.Stringer("endlocation", endLocation), zap.Stringer("operator", operator))
+	h.logger.Info("match and apply a operator", zap.Stringer("startlocation", startLocation), zap.Stringer("endlocation", endLocation), zap.Any("currentEvent", currentEvent), zap.Stringer("operator", operator))
```
buchuitoudegou, line 220:
```suggestion
	h.logger.Info("match and apply an operator", zap.Stringer("startlocation", startLocation), zap.Stringer("endlocation", endLocation), zap.Any("currentEvent", currentEvent), zap.Stringer("operator", operator))
```

buchuitoudegou (commented):
typos😀

dm/pkg/binlog/position.go
```diff
@@ -274,6 +274,13 @@ func (l Location) Clone() Location {
 	return l.CloneWithFlavor("")
 }
 
+// Clone clones a same location, but reset suffix to 0.
```
Ehco1996, line 277: please left a todo location whit suffix is hadr do understand, let's refine this logic later

dm/syncer/checkpoint.go
```diff
@@ -154,8 +154,8 @@ func (b *binlogPoint) outOfDate() bool {
 func (b *binlogPoint) outOfDateBy(pos binlog.Location) bool {
 	b.RLock()
 	defer b.RUnlock()
-
-	return binlog.CompareLocation(pos, b.flushedPoint.location, b.enableGTID) > 0
+	// locations in savedPoint have no suffix and may have same giid like handle-error, we can use position to check them
+	return binlog.CompareLocation(pos, b.flushedPoint.location, false) > 0
```
Ehco1996, line 158: seems change the logic in `binlog.CompareLocation` is more reasonable ?, hardcode to enable gtid to false is too tricky

dm/pkg/binlog/position.go
WizardXiao: yes, suffix logic is hard to understand and it is used in compare location too much. i think i will refine it in other pr.

dm/syncer/checkpoint.go
WizardXiao: `binlog.CompareLocation` is called in many places and i think we may fix it when we refine suffix. this is `outOfDateBy` is used to check between table checkpoint and global checkpoint for flush, this kind of point should have right position, so  we can use position to check.

Ehco1996: >  add more tests about inject

will you doing this in thie pr or, is this pr ready for review?

WizardXiao:
need fix inject command when it was called in sharding.

dm/syncer/checkpoint.go
WizardXiao: not yet. after some test, it will be more complex when use sharding ddl.

PR #4850, opened 2022-03-11
verification(ticdc): e2e verification, checksum

ben1009:

Issue Number: close https://github.com/pingcap/tiflow/issues/4550

part of https://github.com/pingcap/tiflow/issues/4550, e2e checksum check

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
 `None`.
```


ben1009:
/check-issue-triage-complete

> /run-check-issue-triage-complete

1

cdc/verification/checker.go
```diff
@@ -0,0 +1,218 @@
...
+	log.Debug("do checkSum",
+		zap.String("db", databaseName),
+		zap.String("table", tableName),
+		zap.String("query", query))
+	var checkSum string
+	err := c.db.QueryRowContext(ctx, query).Scan(&checkSum)
+	return checkSum, cerror.WrapError(cerror.ErrMySQLQueryError, err)
+}
+
+// TODO: use ADMIN CHECKSUM TABLE for tidb if needed
```
hicqu, line 179: Result of `admin checksum table` contains row keys, so they can be different even if both upstream and downstream are TiDB. I think we can remove this comment.

cdc/verification/checker.go
```diff
@@ -0,0 +1,218 @@
...
+
+	"github.com/pingcap/log"
+	cerror "github.com/pingcap/tiflow/pkg/errors"
+	"github.com/pingcap/tiflow/pkg/filter"
+	"go.uber.org/zap"
+)
+
+//go:generate mockery --name=checkSumChecker --inpackage
+type checkSumChecker interface {
+	getCheckSum(ctx context.Context, db string, f *filter.Filter) (map[string]string, error)
```
hicqu, line 30: I'm not sure whether the indent is correct or not.

cdc/verification/checker.go
ben1009: is it safe if we just compare Total_kvs, Total_kvs, skip  Checksum_crc64_xor
```
mysql> ADMIN CHECKSUM TABLE t1;
+---------+------------+----------------------+-----------+-------------+
| Db_name | Table_name | Checksum_crc64_xor   | Total_kvs | Total_kvs |
+---------+------------+----------------------+-----------+-------------+
| test    | t1         | 10909174369497628533 |         3 |          75 |
```

cdc/verification/checker.go
ben1009: i will run fmt to make sure

hicqu (approved):

cdc/verification/checker.go
```diff
@@ -0,0 +1,220 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/pingcap/log"
+	cerror "github.com/pingcap/tiflow/pkg/errors"
+	"github.com/pingcap/tiflow/pkg/filter"
+	"go.uber.org/zap"
+)
+
+//go:generate mockery --name=checkSumChecker --inpackage
```
overvenus, line 28: Could you add the commond to Makefile? So that generated files always get updated

overvenus (commented):
Rest LGTM

cdc/verification/checker.go
```diff
@@ -0,0 +1,220 @@
...
+
+	result := make(map[string]string)
+	for _, table := range tables {
+		columns, err := c.getColumns(ctx, table)
+		if err != nil {
+			return result, err
+		}
+		checkSum, err := c.doChecksum(ctx, columns, db, table)
+		if err != nil {
+			return result, err
```
3AceShowHand, line 64: it looks that for one table do check sum failed, all tables after that will not be checked.

How about always do check sum on all tables, and return error if at least one failed.

cdc/verification/checker.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+		result[table] = checkSum
+	}
+
+	return result, nil
+}
+
+func (c *checker) getAllDBs(ctx context.Context) ([]string, error) {
```
3AceShowHand, line 72: it looks that `getAllDBs` and `getAllTables` are almost identical, except for the SQL and some filtering logic.

How about abstracting another function to do the data fetch work, and then do specific logic in this 2 methods.

cdc/verification/checker.go
```diff
@@ -0,0 +1,220 @@
...
+
+	return tables, nil
+}
+
+type columnInfo struct {
+	Field   string
+	Type    string
+	Null    string
+	Key     string
+	Default *string
```
3AceShowHand, line 127: it looks that `Default` can be `null`, so use a pointer to track it?

cdc/verification/checker.go
```diff
@@ -0,0 +1,220 @@
...
+		if len(sourceCheckSum) != len(sinkCheckSum) {
+			log.Warn("source and sink have different checker size",
+				zap.Any("source", sourceCheckSum),
+				zap.Any("sink", sinkCheckSum))
+		}
+
+		for k, v := range sourceCheckSum {
+			target, ok := sinkCheckSum[k]
+			if !ok {
+				log.Warn("cannot find checker at sink, it may eligible to replicate",
```
3AceShowHand, line 208: it may `ineligible` to replicate.

cdc/verification/checker.go
```diff
@@ -0,0 +1,220 @@
...
+					zap.String("sourceChecker", v),
+					zap.String("tableName", k))
+				continue
+			}
+			if v != target {
+				log.Error("checker mismatch", zap.String("source", v), zap.String("sink", target))
+				return false, nil
```
3AceShowHand, line 215: how about always check all tables, this would stop on the first unmatched table.

cdc/verification/checker.go
ben1009: what is the benefit for check all instead, more info for debug?

cdc/verification/checker.go
ben1009: i do not get your point

cdc/verification/checker.go
ben1009: for debug ?

cdc/verification/checker.go
3AceShowHand: if that does not provide more info to help us debug, I think the current is good.

cdc/verification/checker.go
3AceShowHand: yes, if necessary.

3AceShowHand (approved):

cdc/verification/checker.go
ben1009: will added in next pr, have the same case

hi-rustin (approved):

PR #4587, opened 2022-02-14
*(ticdc): switch to unified DML library

lance6716:
### What problem does this PR solve?

Issue Number: ref #3242 

### What is changed and how it works?

as title.

I'll file another PR to better represent omitted values when build RowChange. MySQL binlog also has `binlog_row_image` variable which can be supported as well.

### Check List 

Tests 

 - original unit test https://ci.pingcap.net/blue/organizations/jenkins/tiflow-verify-pipeline/detail/tiflow-verify-pipeline/1071/pipeline
 - original integration test https://ci2.pingcap.net/blue/organizations/jenkins/cdc_ghpr_integration_test/detail/cdc_ghpr_integration_test/9618/pipeline

Code changes

Side effects

Related changes

### Release note 

```release-note 
`None`.
```

PR #4872, opened 2022-03-11
script(cdc): don't check line width for test files

lance6716:

Issue Number: ref #4287 

When I write tests, I think a long string literal (like generated SQL) is good to read. But it can't pass CI.

### Check List 

Tests 

### Release note 

```release-note 
 `None`.
```

### DIFF
--- a/scripts/check-diff-line-width.sh
+++ b/scripts/check-diff-line-width.sh
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Run this scripts via `make limit-line-width`.
+# Run this scripts via `make check-diff-line-width`.
 
 set -e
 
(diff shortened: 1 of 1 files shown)


hi-rustin (commented):
I personally feel that the test needs to be kept at normal width as well. It is no different from the core code. Both need to ensure good readability. But if you have a strong opinion about it, I'm ok with that.

lance6716:
> I personally feel that the test needs to be kept at normal width as well. It is no different from the core code. Both need to ensure good readability. But if you have a strong opinion about it, I'm ok with that.

I made some test file changes like this 
https://github.com/pingcap/tiflow/pull/4587/files#diff-7a385c2f760ae6ee8284c0ec6a649a8953602ae51fca4dd19f7ee28c806e2a4eR87

```diff
			expected: &preparedDMLs{
-				sqls:     []string{"DELETE FROM `common_1`.`uk_without_pk` WHERE `a1` = ? AND `a3` = ? LIMIT 1;"},
+				sqls:     []string{"DELETE FROM `common_1`.`uk_without_pk` WHERE `a1` = ? AND `a3` = ? LIMIT 1"},
				values:   [][]interface{}{{1, 1}},
				rowCount: 1,
			},
```

That's not very easy to find a good layout to pass the CI

overvenus (changes requested):
> I personally feel that the test needs to be kept at normal width as well. It is no different from the core code. Both need to ensure good readability. But if you have a strong opinion about it, I'm ok with that.

Agree, without width limit, we may write tests like crazy long, see

https://github.com/pingcap/tiflow/blob/1e8f99f5e4b80a63266204abcb4bb605ef673d3c/dm/loader/convert_data_test.go#L64-L69

https://github.com/pingcap/tiflow/blob/1e8f99f5e4b80a63266204abcb4bb605ef673d3c/pkg/orchestrator/reactor_state_test.go#L61-L68

lance6716:
ok, I will use str+str to split the long string literal in my original pr.

overvenus:
> ok, I will use str+str to split the long string literal in my original pr.

try SQL formatter https://www.dpriver.com/pp/sqlformat.htm

```go
	longLongLongSQL := `
DELETE FROM `common_1`.`uk_without_pk`
WHERE  `a1` = ?
       AND `a3` = ?
LIMIT 1`
```

PR #4874, opened 2022-03-12
tracker(dm): add UT and fix uncleaned folder

### LINKED ISSUE #4816: Remove schema tracker dir in ut
### Which jobs are flaking?

```
make dm_unit_test
```

```log
(paste logs here)
```

lance6716:

Issue Number: close #4816 

- add tracker.Close
- also add a test case that SHOW CREATE TABLE in downstream returns placement rule comment (I get the output from RFC, not test it in real TiDB)
- move some files in sqlmodel package

### Check List 

Tests 

 - Unit test

Related changes

### Release note 

```release-note
`None`.
```

### DIFF
--- a/dm/pkg/schema/tracker_test.go
+++ b/dm/pkg/schema/tracker_test.go
@@ -90,7 +90,9 @@ func (s *trackerSuite) TestTiDBAndSessionCfg(c *C) {
 	dbConn := &dbconn.DBConn{Cfg: s.cfg, BaseConn: baseConn}
 	// user give correct session config
 
-	_, err = NewTracker(context.Background(), "test-tracker", defaultTestSessionCfg, dbConn)
+	t, err := NewTracker(context.Background(), "test-tracker", defaultTestSessionCfg, dbConn)
+	c.Assert(err, IsNil)
+	err = t.Close()
 	c.Assert(err, IsNil)
 
 	// user give wrong session session, will return error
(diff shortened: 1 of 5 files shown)


lance6716:
ping @GMHDBJD @WizardXiao

GMHDBJD (approved):
LGTM

dm/pkg/schema/tracker_test.go
```diff
@@ -114,6 +116,8 @@ func (s *trackerSuite) TestTiDBAndSessionCfg(c *C) {
 	c.Assert(mock.ExpectationsWereMet(), IsNil)
 	err = tracker.Exec(context.Background(), "", "create database testdb;")
 	c.Assert(err, IsNil)
+	err = tracker.Close()
+	c.Assert(err, IsNil)
```
D3Hunter, line 120: maybe make this and below two into a `defer` after `newTracker` too

D3Hunter (approved):
rest lgtm

dm/pkg/schema/tracker_test.go
lance6716: `tracker` will be lost by overwritten when defer closure call, I need to assign every tracker for every `NewTracker` then

D3Hunter: just in case `tracker.close` is not called for previous assert may failed

Issue #4770, opened 2022-03-04
validator(dm): checkpoint, persist pending/error row change data, and table status


D3Hunter:
- add checkpoint to validator
- persist pending/error row change, and table validate status

PR #4881, opened 2022-03-14
validator(dm): checkpoint, persist pending/error row change data, and table status

D3Hunter:

Issue Number: ref #4770

part 1 of #4770

will add ut for checkpoint in later pr

- flush checkpoint and persist pending/error row change data, and table status together at regular interval
- tableinfo inside syncer checkpoint is loaded before syncer run, validator needs it. It's loaded lazily previously(see [here](https://github.com/pingcap/dm/pull/1271)).
- if row change failed to validate after RowErrorDelay, it's recorded as error
- validation error will be recorded after validator reaches syncer's progress to reduce the chance of false-positive
- handle errors of get event
- trigger validation when we have accumulated too many row changes

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`
```


dm/syncer/validate_worker.go
```diff
@@ -40,6 +40,8 @@ import (
 const (
 	workerChannelSize = 1000
 
+	MaxAccumulatedRowBeforeValidate = 1000
```
buchuitoudegou, line 43: Should it be configurable too?

buchuitoudegou (commented):
This PR is too big😵‍💫...I will review the ut later.

Ehco1996:
will review after codefreeze 😂

dm/syncer/validator_checkpoint.go
```diff
@@ -0,0 +1,436 @@
...
+			message: message,
+		}
+	}
+
+	if err = rows.Err(); err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
```
buchuitoudegou, line 433: What about adding a method to get `validation error` like `loadError`?

dm/syncer/validate_worker.go
```diff
@@ -66,118 +70,179 @@ type validateWorker struct {
...
+		if len(allFailedRows) > 0 {
+			failedChanges[k] = allFailedRows
+		}
+	}
+
+	vw.updatePendingAndErrorRows(failedChanges)
+}
+
+func (vw *validateWorker) updatePendingAndErrorRows(failedChanges map[string]map[string]*validateFailedRow) {
+	vw.Lock()
```
buchuitoudegou, line 205: I'm a bit curious about what this lock used for? It seems all the worker functions are executed synchronously in one go routine? Will they called by other outside code (e.g. DataValidator)?

dm/syncer/syncer.go
```diff
@@ -742,21 +739,10 @@ func (s *Syncer) getTableInfo(tctx *tcontext.Context, sourceTable, targetTable *
 		return nil, terror.ErrSchemaTrackerCannotCreateSchema.Delegate(err, sourceTable.Schema)
 	}
 
-	// if table already exists in checkpoint, create it in schema tracker
-	if ti = s.checkpoint.GetFlushedTableInfo(sourceTable); ti != nil {
```
Ehco1996, line 746: do we need to remove this interface func from checkpoint ?

dm/syncer/syncer.go
```diff
@@ -1560,6 +1546,14 @@ func (s *Syncer) updateTSOffset(ctx context.Context) error {
 
 // Run starts running for sync, we should guarantee it can rerun when paused.
 func (s *Syncer) Run(ctx context.Context) (err error) {
+	if !s.schemaLoaded.Load() {
+		err = s.checkpoint.LoadIntoSchemaTracker(ctx, s.schemaTracker)
```
Ehco1996, line 1550: maybe put this load func into sync.Init

Ehco1996 (commented):
will review later

dm/syncer/data_validator.go
```diff
@@ -366,12 +454,33 @@ func (v *DataValidator) doValidate() {
...
+			err = locationForFlush.SetGTID(ev.GSet)
+			if err != nil {
+				v.errChan <- terror.Annotate(err, "failed to set gtid")
+				return
+			}
+			// TODO: if this routine block on get event, should make sure checkpoint is flushed normally
+			if time.Since(lastFlushCheckpointTime) > metaFlushInterval {
+				lastFlushCheckpointTime = time.Now()
+				if err = v.flushCheckpointAndData(locationForFlush); err != nil {
+					v.L.Warn("failed to flush checkpoint: ", zap.Reflect("error", err))
```
Ehco1996, line 477:
```suggestion
					v.L.Warn("failed to flush checkpoint: ", zap.Error(err))
```

dm/syncer/data_validator.go
```diff
@@ -422,23 +534,77 @@ func (v *DataValidator) startValidateWorkers() {
...
 }
 
 func (v *DataValidator) dispatchRowChange(key string, row *rowChange) {
 	hashVal := int(utils.GenHashKey(key)) % v.workerCnt
 	v.workers[hashVal].rowChangeCh <- row
 }
 
+func (v *DataValidator) genValidateTableInfo(sourceTable *filter.Table) (*validateTableInfo, error) {
+	targetTable := v.syncer.route(sourceTable)
+	// todo: syncer will change schema in schemaTracker, will there be data race?
```
Ehco1996, line 552: tracker is thread-safe  but this may cause different table schema between syncer and validator in extreme cases

dm/syncer/data_validator.go
```diff
@@ -422,23 +534,77 @@ func (v *DataValidator) startValidateWorkers() {
...
 		Schema: string(ev.Table.Schema),
 		Name:   string(ev.Table.Table),
 	}
 	fullTableName := sourceTable.String()
-	if _, ok := v.unsupportedTable[fullTableName]; ok {
+	if state, ok := v.tableStatus[fullTableName]; ok && state.stage == pb.Stage_Stopped {
 		return nil
 	}
 
+	for _, cols := range ev.SkippedColumns {
```
Ehco1996, line 601: how about reuse `checkLogColumns`

dm/syncer/validator_checkpoint.go
```diff
@@ -0,0 +1,436 @@
...
+		}
+
+		err = c.createTable(newCtx)
+
+		c.schemaInitialized.Store(true)
+	}
+	return err
+}
+
+func (c *validatorPersistHelper) createSchema(tctx *tcontext.Context) error {
```
Ehco1996, line 112: this schema and table should also removed in `remove-meta` ?

dm/syncer/validator_checkpoint.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+	queries := make([]string, 0, count+2)
+	args := make([][]interface{}, 0, count+2)
+	nextRevision := c.revision + 1
+
+	c.tctx.L().Info("persist checkpoint and intermediate data")
+
+	// update checkpoint
+	queries = append(queries, `INSERT INTO `+c.checkpointTableName+
+		`(source, binlog_name, binlog_pos, binlog_gtid) VALUES (?, ?, ?, ?) ON DUPLICATE KEY UPDATE
```
Ehco1996, line 198: considering that the amount of data here is much larger than the other checkpoints 
i am a little concerned about performance issues about always use `Insert ON DUPLICATE KEY UPDATE` 

see https://asktug.com/t/topic/33147

dm/syncer/validator_checkpoint.go
D3Hunter: add a load error count in later pr, loadError's not needed in this pr

dm/syncer/validate_worker.go
D3Hunter: yes, validator routine will access it's fields

dm/syncer/syncer.go
D3Hunter: init may timeout and cause task fail, that's why we add this [pr](https://github.com/pingcap/dm/pull/1271)

dm/syncer/data_validator.go
D3Hunter: validator is assuming user hasn't do any destructive ddl such as remove any existing column(in real production database, there should‘t be much delete columns operation in most case), so it's ok to use `current ddl` in syncer to validate previous binlog(it's corresponding  ddl may different) as long as the ddls is validation compatible.

@lance6716 says CDC is storing ddl history, we may reuse it, so validator can use ddl corresponding to binlog, but we may add it in later sprint.

dm/syncer/validator_checkpoint.go
D3Hunter: add a todo

dm/syncer/validator_checkpoint.go
D3Hunter: add a todo

Ehco1996 (approved):
LGTM

D3Hunter:
ping @lance6716

lance6716 (commented):
will review later

dm/syncer/data_validator.go
```diff
@@ -447,119 +610,143 @@ func (v *DataValidator) processRowsEvent(header *replication.EventHeader, ev *re
...
 	}
 
-	targetTable := v.syncer.route(sourceTable)
-	// todo: syncer will change schema in schemaTracker, will there be data race?
-	// todo: what if table is dropped while validator falls behind?
-	tableInfo, err := v.syncer.schemaTracker.GetTableInfo(sourceTable)
+	table, err := v.genValidateTableInfo(sourceTable)
 	if err != nil {
 		if schema.IsTableNotExists(err) {
 			// not a table need to sync
```
lance6716, line 616: For TableNotExists, should we record the unmatched tables and let data validator check if there's a DROP TABLE DDL?

dm/syncer/data_validator.go
```diff
@@ -447,119 +610,143 @@ func (v *DataValidator) processRowsEvent(header *replication.EventHeader, ev *re
...
+	var rev int64
+	v.loadedPendingChanges, rev, err = v.persistHelper.loadPendingChange(tctx)
+	if err != nil {
+		return err
+	}
+	// fix missing data
+	for _, tblChange := range v.loadedPendingChanges {
+		tblChange.table, err = v.genValidateTableInfo(tblChange.table.Source)
+		if err != nil {
+			return err
```
lance6716, line 734: in line 614 we ignored some error. But same error will cause whole routine failed here.

dm/syncer/data_validator.go
```diff
@@ -447,119 +610,143 @@ func (v *DataValidator) processRowsEvent(header *replication.EventHeader, ev *re
...
+	if err != nil {
+		return err
+	}
+
+	var rev int64
+	v.loadedPendingChanges, rev, err = v.persistHelper.loadPendingChange(tctx)
+	if err != nil {
+		return err
+	}
+	// fix missing data
```
lance6716, line 730: could you left a comment to explain that which fields should be recovered from loadPendingChange and what to do for else fields?

dm/syncer/data_validator.go
```diff
@@ -328,20 +379,57 @@ func (v *DataValidator) doValidate() {
...
 	v.startValidateWorkers()
 	defer func() {
 		for _, worker := range v.workers {
 			worker.close()
 		}
 	}()
 
+	metaFlushInterval := v.cfg.ValidatorCfg.MetaFlushInterval.Duration
+	// when gtid is enabled:
+	// gtid of currLoc need to be updated when meet gtid event, so we can compare with syncer
```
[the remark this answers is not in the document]
D3Hunter: add a todo, will check it later.

dm/syncer/data_validator.go
```diff
@@ -422,23 +534,74 @@ func (v *DataValidator) startValidateWorkers() {
...
+		return nil, err
+	}
+	table := &validateTableInfo{
+		Source: sourceTable,
+		Info:   tableInfo,
+		Target: targetTable,
+	}
+
+	var primaryIdx *model.IndexInfo
+	for _, idx := range tableInfo.Indices {
```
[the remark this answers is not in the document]
D3Hunter: i noticed this after reading `pkg/sqlmodel`, planing to use it in later pr.

dm/syncer/data_validator.go
```diff
@@ -328,20 +379,57 @@ func (v *DataValidator) doValidate() {
...
+	// some events maybe processed again, but it doesn't matter for validation
 	currLoc := location.CloneWithFlavor(v.cfg.Flavor)
+	locationForFlush := currLoc.Clone()
+	lastFlushCheckpointTime := time.Now()
 	for {
 		e, err := v.streamerController.GetEvent(v.tctx)
 		if err != nil {
-			v.errChan <- terror.Annotate(err, "fail to get binlog from stream controller")
-			return
+			switch {
```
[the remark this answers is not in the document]
D3Hunter: yes, similar with how syncer.run handles it.

even we merge this switch together with getEvent, we still need a switch kind of thing, since the error returned affect the control path( break/continue/return) of outside code.

maybe we can try to abstract them later when we refactor syncer.

dm/syncer/data_validator.go
```diff
@@ -316,9 +357,19 @@ func (v *DataValidator) doValidate() {
 		return
 	}
 
+	if err := v.loadPersistedData(v.tctx); err != nil {
+		v.errChan <- terror.Annotate(err, "failed to load persisted data")
```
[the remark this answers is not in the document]
D3Hunter: plan to do them together, add some todos.

dm/syncer/data_validator.go
```diff
@@ -111,15 +133,21 @@ type DataValidator struct {
...
 	validateInterval time.Duration
 	workers          []*validateWorker
 	changeEventCount []atomic.Int64
 	workerCnt        int
 
-	// such as table without primary key
-	unsupportedTable map[string]string
+	// whether the validation progress ever reached syncer
+	// if it's false, we don't mark failed row change as error to reduce false-positive
+	reachedSyncer atomic.Bool
```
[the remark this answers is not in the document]
D3Hunter: it's designed to be true after we reached syncer once, so we can set first failed time for row change.

if we init it every time, and validator falls behind and reaches syncer frequently, validate worker will never has any chance to mark a row as error.

we can discuss the strategy later, if it cause too much false-positive.

dm/syncer/data_validator.go
D3Hunter: if the table is dropped in downstream, we cannot validate this case.

will we skip DROP TABLE event but drop it in schma-tracker?

dm/syncer/data_validator.go
D3Hunter: in 614, we just met the row and has to decide whether this row should be validated, so if the err shows this row we cannot validate, it's skipped. here this row should be validated(already persisted), so should not have error when getting table info.

D3Hunter:
ptal @lance6716

dm/syncer/data_validator.go
lance6716: If the table is dropped by operations outside DM, DML on that table will cause error. Generally user will not do that. OK to address this behaviour later.

> will we skip DROP TABLE event but drop it in schma-tracker?

currently schema tracker takes effect before the filter.

dm/syncer/data_validator.go
lance6716: maybe it will happen when

... DML on t1, DROP TABLE t1, DML on other tables, ...

the validator has pending changes of DML on t1, but syncer has dropped t1, so validator can't bootup. Maybe multi-version TableInfo can help it, we can fix it later

D3Hunter: add a todo

dm/syncer/validate_worker.go
```diff
@@ -199,12 +264,28 @@ func (vw *validateWorker) validateRowChanges(table *validateTableInfo, rows []*r
 	return res, nil
 }
 
+func (vw *validateWorker) getPendingChangesMap() map[string]*tableChange {
```
lance6716, line 267: I recall a recently found data race. This function will return same map to caller, the lock can't protect R/W to the map KV. Maybe the lock is not needed?

dm/syncer/data_validator.go
D3Hunter: > If the table is dropped by operations outside DM, DML on that table will cause error.

will add it to `restrictions` in this version.

lance6716 (commented):
will review later

lance6716 (commented):
rest lgtm

dm/syncer/validate_worker.go
D3Hunter: worker will not write it when we're persisting since there's no rowChange in the worker channel, so there'll be no data race.

lock is a must for concurrent r/w accessed and non-atomic variable, even there'll be only one routine may read it as in this case, to make sure we can see the latest value.

dm/syncer/validator_checkpoint.go
```diff
@@ -0,0 +1,438 @@
...
+	c.db = db
+	c.dbConn = conns[0]
+	defer func() {
+		if err == nil {
+			return
+		}
+		dbconn.CloseBaseDB(newCtx, c.db)
+	}()
+
+	if !c.schemaInitialized.Load() {
```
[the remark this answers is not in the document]
D3Hunter: it painful to mock sqls of `create schema/table`, can use this var to skip it.

there's no `init` method for validator, it's inited in `start` every time it's called, it's annoying to see a bunch of logs like `create table if not exists` again.

dm/syncer/data_validator.go
```diff
@@ -447,119 +611,143 @@ func (v *DataValidator) processRowsEvent(header *replication.EventHeader, ev *re
...
-	var primaryIdx *model.IndexInfo
-	for _, idx := range tableInfo.Indices {
-		if idx.Primary {
-			primaryIdx = idx
+	if len(table.Info.Columns) < int(ev.ColumnCount) {
+		v.tableStatus[fullTableName] = &tableValidateStatus{
+			source:  *sourceTable,
+			target:  *table.Target,
+			stage:   pb.Stage_Stopped,
+			message: moreColumnInBinlogMsg,
```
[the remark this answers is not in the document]
D3Hunter: > I guess this is just a checker that has false negative?
since there's no ddl history now(not plan to add it in this version), we restrict that the table validated should not drop/ columns or switch order or ADD COLUMN AFTER.  not a false negative, we just cannot handle it now.

dm/syncer/validate_worker.go
lance6716: OK I'm not sure if golang need a memory barrier 🤔

lance6716 (approved):

PR #4897, opened 2022-03-15
DM: adjust loader's dir to be add subdirectory

### LINKED ISSUE #4360: Support S3 for DM dump and load

WizardXiao:

Issue Number: ref #4360

Adjust path to be subdirectory instead of suffix when we adjust loader's dir in subtask. 

### Check List 

Tests 

 - Integration test

### Release note 

```release-note
None
```


dm/dm/config/subtask.go
```diff
@@ -415,7 +416,7 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 		} else {
 			dirSuffix = c.Name
 		}
-		newDir, err := storage.AdjustPath(c.LoaderConfig.Dir, "."+dirSuffix)
+		newDir, err := storage.AdjustPath(c.LoaderConfig.Dir, string(filepath.Separator)+dirSuffix)
```
WizardXiao, line 419: `storage.AdjustPath` has no change and may be we can use in future, not only add `/xxx`.

lance6716 (approved):
will upgrading from an old version get affected?

lichunzhu (approved):

csuzhangxc:
> will upgrading from an old version get affected?

Can it work if only part of dump data has been imported during the upgrade of DM?

WizardXiao:
> > will upgrading from an old version get affected?
> 
> Can it work if only part of dump data has been imported during the upgrade of DM?

no, it can't, it need full stage has done and syncer is runing now.

WizardXiao:
After discuss with @sunzhaoyang , we decide to just change s3 first, because change local path may have
 risk of compatibility.

dm/dm/config/subtask.go
```diff
@@ -411,11 +411,11 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 		// add suffix
 		var dirSuffix string
 		if isS3 {
-			dirSuffix = c.Name + "." + c.SourceID
+			dirSuffix = "/" + c.Name + "." + c.SourceID
```
Ehco1996, line 414: please add comment for this change

Ehco1996 (approved):
LGTM

dm/dm/config/subtask.go
Ehco1996: i suddenly find that we shoud adapt this change when convert subtask to OpenAPITask
see here https://github.com/pingcap/tiflow/blob/50c98a6ac9eb456ac1cd47d871349b92976eb868/dm/dm/config/task_converters.go#L358

WizardXiao: i add trim path suffix at [this commit](https://github.com/pingcap/tiflow/pull/4897/commits/b481c0f6b746688b12d48798538bc9e49ea34c52), please take a look.

dm/dm/config/task_converters.go
```diff
@@ -355,9 +357,21 @@ func SubTaskConfigsToTaskConfig(stCfgs ...*SubTaskConfig) *TaskConfig {
...
+			dirSuffix = "/" + c.Name + "." + stCfg.SourceID
+		} else {
+			// TODO we will dump local file to dir's subdirectory, but it may have risk of compatibility, we will fix in other pr
+			dirSuffix = "." + c.Name
+		}
 		// if ends with the task name, we remove to get user input dir.
-		loaderCfg.Dir = strings.TrimSuffix(loaderCfg.Dir, dirSuffix)
+		loaderCfg.Dir, err = storage.TrimPath(loaderCfg.Dir, dirSuffix)
+		// because dir comes form subtask, there should not have error.
+		log.L().Warn("parse config comes from subtask error.", zap.Error(err))
```
Ehco1996, line 373: if err!=nil{
}

dm/dm/config/task_converters.go
```diff
@@ -445,9 +459,20 @@ func SubTaskConfigsToOpenAPITask(subTaskConfigList []*SubTaskConfig) *openapi.Ta
...
+		dirSuffix = "/" + oneSubtaskConfig.Name + "." + oneSubtaskConfig.SourceID
+	} else {
+		// TODO we will dump local file to dir's subdirectory, but it may have risk of compatibility, we will fix in other pr
+		dirSuffix = "." + oneSubtaskConfig.Name
+	}
 	// if ends with the task name, we remove to get user input dir.
-	oneSubtaskConfig.LoaderConfig.Dir = strings.TrimSuffix(oneSubtaskConfig.LoaderConfig.Dir, dirSuffix)
+	oneSubtaskConfig.LoaderConfig.Dir, err = storage.TrimPath(oneSubtaskConfig.LoaderConfig.Dir, dirSuffix)
+	// because dir comes form subtask, there should not have error.
+	log.L().Warn("parse config comes from subtask error.", zap.Error(err))
```
Ehco1996, line 474: ditto

dm/dm/config/task_converters.go
WizardXiao: ok

dm/dm/config/task_converters.go
WizardXiao: ok

dm/pkg/storage/utils.go
```diff
@@ -40,21 +41,43 @@ func AdjustPath(rawURL string, uniqueID string) (string, error) {
...
+	if rawURL == "" || uniqueID == "" {
+		return rawURL, nil
+	}
+	u, err := bstorage.ParseRawURL(rawURL)
+	if err != nil {
+		return "", errors.Trace(err)
+	}
+	// not url format, we don't use url library to avoid being escaped or unescaped
+	if u.Scheme == "" {
+		// avoid duplicate add uniqueID, and trim suffix '/' like './dump_data/'
```
lance6716, line 72: Seems this function is only used for the subtask config, and subtask config will not have `/` suffix? we can remove this comment or in later PR

lance6716 (commented):
lgtm

dm/pkg/storage/utils.go
WizardXiao: This is for handle the last `/` of user configuration like `./dump_data/`.  this issue was found out by @sunzhaoyang when discuss how to add subdir.

WizardXiao: ok, i will fix this comment in later pr.

PR #4913, opened 2022-03-16
sink(ticdc): support multi topics

hi-rustin:

- flush worker supports multiple topics

- ddl distribution support for multiple topics

- checkpointts broadcast support for multiple topics (If no table is currently synchronized, the default is sent to the default topic)

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes

None

### Release note 

```release-note
None
```


hi-rustin:
/hold

Please do not merge the release 6.0 branch before it is check out.

hi-rustin:
/hold
Please do not merge this PR until version 6.0 is check out.

cdc/sink/mq.go
```diff
@@ -297,16 +322,15 @@ func (k *mqSink) run(ctx context.Context) error {
...
+func (k *mqSink) asyncWriteToProducer(
+	ctx context.Context, topic string, partition int32, message *codec.MQMessage,
+) error {
+	err := k.mqProducer.AsyncSendMessage(ctx, topic, partition, message)
+	if err != nil {
+		return err
 	}
-
-	return k.mqProducer.SyncBroadcastMessage(ctx, message)
+	return k.mqProducer.Flush(ctx)
```
amyangfei, line 333: Is this a synchronous send, since it calls `Flush` at last.
Besides the function is only used  in L299, the parameter `partition` can be removed.

cdc/sink/dispatcher/event_router.go
```diff
@@ -107,23 +118,58 @@ func NewEventRouter(cfg *config.ReplicaConfig, defaultTopic string) (*EventRoute
...
+// FIXME: Now we can't handle rename tables because
+// we are missing the old and new tables information.
+func (s *EventRouter) GetTopicForDDL(ddl *model.DDLEvent) string {
+	var schema, table string
+	if ddl.PreTableInfo != nil {
+		if ddl.PreTableInfo.Table == "" {
+			return s.defaultTopic
+		}
+		schema = ddl.PreTableInfo.Schema
+		table = ddl.PreTableInfo.Table
```
amyangfei, line 137: If `table` information is nil, should it match the rule `schema.*`

cdc/sink/mq.go
hi-rustin: Fixed.

cdc/sink/dispatcher/event_router.go
hi-rustin: Do you mean a situation where there is neither a new information nor an old information? Does this happen with the ddl supported by cdc?

hi-rustin: When the table does not exist, we consider it to be a schema-level DDL, which should be sent to the default topic according to the spec.

cdc/sink/dispatcher/event_router.go
```diff
@@ -26,6 +26,17 @@ import (
 	"go.uber.org/zap"
 )
 
+// DDLDispatchRule is the dispatch rule for DDL event.
+type DDLDispatchRule int
+
+const (
+	// Broadcast means the DDL event will be broadcast to all the partitions.
+	Broadcast DDLDispatchRule = -1
```
zhaoxinyu, line 34: Maybe `PartitionAll` is a better name to keep consistent with `PartitionZero`.

hi-rustin: Cool!

cdc/sink/mq.go
```diff
@@ -230,9 +232,32 @@ func (k *mqSink) EmitCheckpointTs(ctx context.Context, ts uint64, _ []model.Tabl
...
 	}
-	log.Debug("emit checkpointTs", zap.Uint64("checkpointTs", ts))
-	err = k.writeToProducer(ctx, msg, -1)
-	return errors.Trace(err)
+	// NOTICE: When there is no table sync,
+	// we need to send checkpoint ts to the default topic. T
+	// This will be compatible with the old behavior.
+	if len(tables) == 0 {
+		topic := k.eventRouter.GetDefaultTopic()
+		partitionNum, err := k.topicManager.Partitions(topic)
```
zhaoxinyu, line 240: `Partitions` is expected to return a partition list on the first sight. It's better to replace with name like `GetPartitionCount`.

zhaoxinyu (approved):

cdc/sink/mq.go
hi-rustin: I refer to sarama's command. If you have a strong opinion about it, I can modify it. 

amyangfei (approved):

PR #4928, opened 2022-03-17
validator(dm): impl validation status command

### LINKED ISSUE #4889: Implement  validation status command

buchuitoudegou:

Issue Number: close #4889

this pr looks big for changes of grpc generated code

1. implement `GetValidatorStatus` gRPC service for worker and master
2. Add some uts

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes
do not merge until #4881 is approved and merged.

### Release note 

```release-note
`None`
```


dm/dm/proto/dmworker.proto
```diff
@@ -404,3 +406,22 @@ message CheckSubtasksCanUpdateResponse {
   bool success = 1;
   string msg = 2; // message when error occurred
 }
+
+message GetValidationStatusRequest {
+    string taskName = 1;
+    string FilterStatus = 2;
```
Ehco1996, line 412: this field better to use enum

dm/dm/worker/status.go
```diff
@@ -123,3 +124,25 @@ func (w *SourceWorker) GetUnitAndSourceStatusJSON(stName string, sourceStatus *b
 	}
 	return s
 }
+
+func (w *SourceWorker) GetValidateStatus(stName string, filterStatus string) []*pb.ValidationStatus {
```
Ehco1996, line 128: `stName` is not used?

Ehco1996 (commented):
maybe add a it-test for this cmd

dm/dm/worker/status.go
buchuitoudegou: It seems I forgot to filter subtasks by `stName`😂

buchuitoudegou:
> maybe add a it-test for this cmd

Would it be better to add it-test after finishing all commands of validator? @Ehco1996 /cc @D3Hunter

Ehco1996:
> > maybe add a it-test for this cmd
> 
> Would it be better to add it-test after finishing all commands of validator? @Ehco1996 /cc @D3Hunter

lgtm

dm/dm/proto/dmworker.proto
buchuitoudegou: This field is already enum in DM (i.e., pb.Stage_Running). Did you mean I should use `Stage FilterStatus = 2;` rather than `string FilterStatus = 2;`?

buchuitoudegou: > this field better to use enum

If so, I don't think it works. `FilterStatus` can be empty string which indicates the master should return all the subtasks with either `Running` stage or `Stopped` stage. If using Stage enum here, I can't do this intuitively.

dm/syncer/data_validator.go
```diff
@@ -777,3 +783,35 @@ func genColData(v interface{}) string {
...
+	})
+	ret, err := v.persistHelper.loadTableStatus(v.tctx)
+	if err != nil {
+		return []*pb.ValidationStatus{}
+	}
+	result := make([]*pb.ValidationStatus, 0)
+	for _, tblStat := range ret {
+		result = append(result, &pb.ValidationStatus{
+			SrcTable:         dbutil.TableName(tblStat.source.Schema, tblStat.source.Name),
+			DstTable:         dbutil.TableName(tblStat.target.Schema, tblStat.target.Name),
```
okJiang, line 811:
```suggestion
			SrcTable:         tblStat.source.String(),
			DstTable:         tblStat.target.String(),
```


It is better to unify `DstTable` and `target`.

And a suggestion is that `tblStat.source` and `tblStat.target` also should add a suffix such as `Table` to show their identity in naming.

dm/dm/worker/server_test.go
```diff
@@ -368,6 +368,97 @@ func (t *testServer) TestHandleSourceBoundAfterError(c *C) {
...
+	sourceBound := ha.NewSourceBound(sourceCfg.SourceID, cfg.Name)
+	_, err = ha.PutSourceBound(etcdCli, sourceBound)
+	c.Assert(err, IsNil)
+	w, err := s.getOrStartWorker(sourceCfg, true)
+	c.Assert(err, IsNil)
+	expected := []*pb.ValidationStatus{
+		{
+			Source:           "127.0.0.1",
+			SrcTable:         "`testdb1`.`testtable1`",
+			DstTable:         "`dstdb`.`dsttable`",
```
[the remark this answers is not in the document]
buchuitoudegou: `SrcTable` and `DstTable` are defined in the proto files, where their capitalization indicate themselves as public. Did you mean dstdb.dsttable should be CamelCase? I think it's just a test case string and it's weird using CamelCase as schemas' and tables' names by any means🤔

dm/dm/master/server_test.go
```diff
@@ -2359,3 +2378,132 @@ func (t *testMaster) validatorModeMatch(c *check.C, s *scheduler.Scheduler, task
 	c.Assert(ok, check.IsTrue)
 	c.Assert(cfg.ValidatorCfg.Mode, check.Equals, expectMode)
 }
+
+func (t *testMaster) TestGetValidatorStatus(c *check.C) {
```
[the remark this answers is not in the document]
buchuitoudegou: It seems using `check` is some sort of convention in `dm/dm/master/server_test.go`. I don't know if we should break it in this PR. Perhaps we are supposed to migrate to `testify` in another PR (It's really confusing that both of the two packages are being used in DM, tbh😵‍💫)

dm/syncer/data_validator.go
buchuitoudegou: Will resolve it.

Naming issue: /cc @D3Hunter

dm/dm/worker/server_test.go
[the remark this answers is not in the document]
okJiang: > Did you mean dstdb.dsttable should be CamelCase?

yep. 

> I think it's just a test case string and it's weird using CamelCase as schemas' and tables' names by any means🤔

I don't know whether it is necessary for test case. You can hear what other reviewers say.

dm/dm/worker/server_test.go
```diff
@@ -368,6 +368,97 @@ func (t *testServer) TestHandleSourceBoundAfterError(c *C) {
...
+	// get all
+	resp, err = s.GetWorkerValidateStatus(ctx, &pb.GetValidationStatusRequest{
+		TaskName: "testQueryValidator",
+	})
+	c.Assert(err, IsNil)
+	c.Assert(resp.Result, IsTrue)
+	c.Assert(len(resp.Msg), Equals, 0)
+	c.Assert(resp.Status, DeepEquals, expected)
+	c.Assert(failpoint.Disable("github.com/pingcap/tiflow/dm/syncer/MockValidationQuery"), IsNil)
+	c.Assert(failpoint.Disable("github.com/pingcap/tiflow/dm/dm/worker/WorkerMockValidationQuery"), IsNil)
```
D3Hunter, line 459: move those into defer after `failpoint.Enable` in case assert failed in between.

dm/dm/worker/status.go
```diff
@@ -123,3 +124,25 @@ func (w *SourceWorker) GetUnitAndSourceStatusJSON(stName string, sourceStatus *b
...
+	res := make([]*pb.ValidationStatus, 0)
+	if len(sts) == 0 {
+		return res
+	}
+	sourceIP := w.cfg.DecryptPassword().From.Host
+	for _, st := range sts {
+		if st.cfg.Name != stName || st.validator == nil || !st.validator.Started() {
+			continue
+		}
+		tblStats := st.validator.GetValidationStatus()
```
D3Hunter, line 139: `st.validator` might be accessed concurrently

prefer to access methods of subtask, not access its field directly.

dm/dm/worker/status.go
```diff
@@ -123,3 +124,25 @@ func (w *SourceWorker) GetUnitAndSourceStatusJSON(stName string, sourceStatus *b
...
+		for _, stat := range tblStats {
+			if filterStatus == "" || strings.EqualFold(stat.ValidationStatus, filterStatus) {
```
D3Hunter, line 141: use `==` directly?

dm/syncer/data_validator.go
```diff
@@ -777,3 +783,35 @@ func genColData(v interface{}) string {
...
+			},
+			{
+				SrcTable:         "`testdb2`.`testtable2`",
+				DstTable:         "`dstdb`.`dsttable`",
+				ValidationStatus: pb.Stage_Stopped.String(),
+				Message:          tableWithoutPrimaryKeyMsg,
+			},
+		})
+	})
+	ret, err := v.persistHelper.loadTableStatus(v.tctx)
```
D3Hunter, line 803: can read `v.tableStatus` directly(need to make sure routine safe)

dm/syncer/data_validator.go
D3Hunter: since the struct is named `tableValidateStatus`, it's ok to omit `table` suffix i think

dm/dm/worker/status.go
```diff
@@ -123,3 +124,25 @@ func (w *SourceWorker) GetUnitAndSourceStatusJSON(stName string, sourceStatus *b
...
 	return s
 }
+
+func (w *SourceWorker) GetValidateStatus(stName string, filterStatus string) []*pb.ValidationStatus {
+	sts := w.subTaskHolder.getAllSubTasks()
+	res := make([]*pb.ValidationStatus, 0)
+	if len(sts) == 0 {
+		return res
+	}
+	sourceIP := w.cfg.DecryptPassword().From.Host
```
D3Hunter, line 134: why calls `DecryptPassword`? since we only uses sourceIP

dm/dm/worker/source_worker_test.go
```diff
@@ -759,3 +759,72 @@ func (t *testServer) testSourceWorker(c *C) {
...
+	c.Assert(len(ret), Equals, 1)
+	c.Assert(ret[0], DeepEquals, expected[0])
+	ret = w.GetValidateStatus("testQueryValidator", pb.Stage_Stopped.String())
+	c.Assert(len(ret), Equals, 1)
+	c.Assert(ret[0], DeepEquals, expected[1])
+	ret = w.GetValidateStatus("testQueryValidator", "")
+	c.Assert(len(ret), Equals, 2)
+	c.Assert(ret, DeepEquals, expected)
+	c.Assert(failpoint.Disable("github.com/pingcap/tiflow/dm/syncer/MockValidationQuery"), IsNil)
+	c.Assert(failpoint.Disable("github.com/pingcap/tiflow/dm/dm/worker/WorkerMockValidationQuery"), IsNil)
```
D3Hunter, line 829: doitto

dm/dm/worker/source_worker_test.go
```diff
@@ -759,3 +759,72 @@ func (t *testServer) testSourceWorker(c *C) {
...
+		DialKeepAliveTimeout: keepaliveTimeout,
+	})
+	c.Assert(err, IsNil)
+
+	w, err := NewSourceWorker(cfg, etcdCli, "", "")
+	w.closed.Store(false)
+	c.Assert(err, IsNil)
+	c.Assert(failpoint.Enable("github.com/pingcap/tiflow/dm/syncer/MockValidationQuery", `return(true)`), IsNil)
+	c.Assert(failpoint.Enable("github.com/pingcap/tiflow/dm/dm/worker/WorkerMockValidationQuery", `return(true)`), IsNil)
+	c.Assert(w.StartSubTask(&config.SubTaskConfig{
```
D3Hunter, line 798: maybe we can `recordSubTask` in subtask holder, so we don't need to start the task directly. so we can remove the hacky failpoint in source worker.

dm/dm/worker/subtask.go
```diff
@@ -273,6 +273,9 @@ func (st *SubTask) StartValidator(expect pb.Stage) {
 	}
 	var syncerObj *syncer.Syncer
 	var ok bool
+	failpoint.Inject("WorkerMockValidationQuery", func(_ failpoint.Value) {
```
D3Hunter, line 276: same failpoint name as in `source worker`?

dm/dm/worker/server_test.go
```diff
@@ -368,6 +368,97 @@ func (t *testServer) TestHandleSourceBoundAfterError(c *C) {
 	}), IsTrue)
 }
 
+func (t *testServer) TestServerQueryValidator(c *C) {
```
D3Hunter, line 371: maybe we can just test `w == nil` case, and let integration test to test the other case since there's no other branch and worker.GetValidateStatus is tested already.

dm/dm/worker/status.go
buchuitoudegou: `filterStatus` is possibly of lower case. `EqualFold` will compare the two strings case insensitively.

dm/dm/master/server.go
```diff
@@ -2709,9 +2709,9 @@ func (s *Server) GetValidationStatus(ctx context.Context, req *pb.GetValidationS
 		resp.Msg = "task name should be specified"
 		return resp, nil
 	}
-	if req.FilterStatus != strings.ToLower(pb.Stage_Running.String()) && req.FilterStatus != strings.ToLower(pb.Stage_Stopped.String()) {
+	if req.FilterStatus != "" && req.FilterStatus != strings.ToLower(pb.Stage_Running.String()) && req.FilterStatus != strings.ToLower(pb.Stage_Stopped.String()) {
 		resp.Result = false
-		resp.Msg = fmt.Sprintf("filtering stage should be either `%s` or `%s`", strings.ToLower(pb.Stage_Running.String()), strings.ToLower(pb.Stage_Stopped.String()))
+		resp.Msg = fmt.Sprintf("filtering stage should be either `%s`, `%s`, or empty", strings.ToLower(pb.Stage_Running.String()), strings.ToLower(pb.Stage_Stopped.String()))
 		return resp, err
```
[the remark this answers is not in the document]
buchuitoudegou: So it's simply `nil` here to remove the lint error (maybe I should disable the linting test)🤔

dm/dm/worker/status.go
buchuitoudegou: Did you mean I should wrap it with a method and acquire a lock before accessing it?

D3Hunter: yes, see stopValidator

dm/dm/proto/dmworker.proto
Ehco1996: if ctl not pass FilterStatua as filter, pb will pass `InvalidStage` as default value, It's a long-standing criticism of pb3

> every enum definition must contain a constant that maps to zero as its first element. 
see [here](https://developers.google.com/protocol-buffers/docs/proto3#enum)

Ehco1996: btw, why just `FilterStatus`  is Capital letters 🤣

dm/dm/master/server_test.go
```diff
@@ -2359,3 +2378,132 @@ func (t *testMaster) validatorModeMatch(c *check.C, s *scheduler.Scheduler, task
...
+	)
+	ctrl := gomock.NewController(c)
+	defer ctrl.Finish()
+	server := testDefaultMasterServer(c)
+	server.etcdClient = t.etcdTestCli
+	sources, workers := defaultWorkerSource()
+	sources = sources[:1]
+	workers = workers[:1]
+	startReq := &pb.StartTaskRequest{
+		Task:    taskCfg2,
```
Ehco1996, line 2468: seems this taskCfg2 is not valid?, and you used a mock client, maybe just pass a empty string as task cfg?

Ehco1996 (commented):
restLGTM

dm/dm/proto/dmworker.proto
buchuitoudegou: Will fix it

dm/dm/master/server_test.go
buchuitoudegou: I think it oughts to be valid because the `server.StartTask` will check its validity before starting a task and thus empty string cannot pass the verification.

buchuitoudegou: But this test is removed anyway...

Ehco1996 (approved):
lgtm

dm/dm/ctl/master/query_validation.go
```diff
@@ -115,6 +115,15 @@ func queryValidationStatus(cmd *cobra.Command, _ []string) error {
 			strings.ToLower(pb.Stage_Stopped.String()),
 		)
 	}
+	var pbStage pb.Stage
+	switch stage {
+	case "":
+		pbStage = pb.Stage_InvalidStage
```
D3Hunter, line 121: maybe add a comment here, we use invalid-stage which means to get all

dm/syncer/data_validator.go
```diff
@@ -777,3 +782,33 @@ func genColData(v interface{}) string {
...
+			},
+			{
+				SrcTable:         "`testdb2`.`testtable2`",
+				DstTable:         "`dstdb`.`dsttable`",
+				ValidationStatus: pb.Stage_Stopped.String(),
+				Message:          tableWithoutPrimaryKeyMsg,
+			},
+		})
+	})
+	v.RLock()
```
D3Hunter, line 802: we need to protect all access to `v.tableStatus` using lock. in `processRowsEvent` we're writing to it without lock, may cause data race.

buchuitoudegou: I've tried to fix it by requiring exclusive lock before processing rows event. PTAL again.

D3Hunter: it's ok for now, i'm refactoring `processRowsEvent` to reuse `pkg/sqlmodel`, will try to make the lock range smaller.

D3Hunter (approved):

PR #4942, opened 2022-03-18
scheduler(ticdc): Fix lost finish operation

### LINKED ISSUE #4963: [TiCDC/pre-release] Table dispatch command response is lost when owner is changed during table dispatching
- Scenario: Injects total data loss between TiCDC nodes and the whole PD cluster one at a time, i.e. prevents only one TiCDC node from contacting any PD server at each time. The replication could become stuck.
- Root cause: The dispatch command response could arrive at the owner before the `Sync` message, which has been induced by an owner change. The Owner will refuse to accept the dispatch command response before the Processor is sync'ed.
- Likelyhood: Rare. Never reproduced without using ChaosMesh.

liuzix:

Issue Number: close #4963

- Added a barrier for the `Sync` message.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (Re-run the problematic chaos test case).

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Fix a rare likelihood that replication be stuck if the owner is changed when the new scheduler is enabled (disabled by default).
```


cdc/processor/agent.go
```diff
@@ -322,8 +361,16 @@ func (a *agentImpl) Barrier(_ context.Context) (done bool) {
...
 
-func (a *agentImpl) OnOwnerChanged(ctx context.Context, newOwnerCaptureID model.CaptureID) {
+func (a *agentImpl) OnOwnerChanged(
+	ctx context.Context,
+	newOwnerCaptureID model.CaptureID,
+	newOwnerRev int64,
+) {
+	if a.ownerCaptureID == newOwnerCaptureID && a.ownerRevision == newOwnerRev {
+		return
+	}
```
amyangfei, line 371: In which scenario can `OnOwnerChanged` be called but these two values are not changed

cdc/processor/agent.go
```diff
@@ -166,6 +192,19 @@ func (a *agentImpl) FinishTableOperation(
 	tableID model.TableID,
 	epoch model.ProcessorEpoch,
 ) (done bool, err error) {
+	topic := model.SyncTopic(a.changeFeed)
+	if !a.Barrier(ctx) {
+		if _, exists := a.barrierSeqs[topic]; exists {
+			log.L().Info("Delay sending FinishTableOperation "+
+				"due to pending sync",
```
overvenus, line 199:
```suggestion
			log.L().Info("Delay sending FinishTableOperation due to pending sync",
```

It's okay, as long as it does not exceed 100.

cdc/processor/agent_test.go
```diff
@@ -73,56 +78,61 @@ func newAgentTestSuite(t *testing.T) *agentTestSuite {
 	ownerMessageClient := cluster.Nodes[ownerCaptureID].Router.GetClient(processorCaptureID)
 	require.NotNil(t, ownerMessageClient)
 
-	dispatchResponseCh := make(chan *model.DispatchTableResponseMessage, 1)
+	ret := &agentTestSuite{
+		cluster:      cluster,
+		etcdClient:   etcdCli,
+		etcdKVClient: KVCli,
+
+		dispatchResponseCh: make(chan *model.DispatchTableResponseMessage, 2),
```
overvenus, line 86: Could you add some comments about why the capacity is 2?

amyangfei (approved):

overvenus (approved):

PR #4947, opened 2022-03-18
DM: fix unit test's s3 url format to avoid errors when use `git secrets` check

WizardXiao:

Issue Number: ref #4360

fix urls for unit test which will be error when use `git secrets` check.
### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```


ben1009 (approved):

dm/pkg/storage/utils_test.go
```diff
@@ -76,8 +76,9 @@ func TestIsS3AndAdjustAndTrimPath(t *testing.T) {
 			"s3://bucket3/prefix/path_placeholder?endpoint=https://127.0.0.1:9000&force_path_style=0&SSE=aws:kms&sse-kms-key-id=TestKey&xyz=abc",
 		},
 		{
-			true, "s3://bucket3/prefix/path?access-key=NXN7IPIOSAAKDEEOLMAF&secret-access-key=nREY/7Dt+PaIbYKrKlEEMMF/ExCiJEX=XMLPUANw",
-			"s3://bucket3/prefix/path_placeholder?access-key=NXN7IPIOSAAKDEEOLMAF&secret-access-key=nREY/7Dt%2BPaIbYKrKlEEMMF/ExCiJEX=XMLPUANw",
+			// this will be error
```
lance6716, line 79: what error?

lance6716 (approved):
And I suddently found that we will print access keys to log because they are in config 😂

dm/pkg/storage/utils_test.go
WizardXiao: `git secrets` will be called when `git commit`  and check s3 urls. [git secrets](https://github.com/awslabs/git-secrets).
errors like this.
`git commit -m "test"
dm/pkg/storage/utils_test.go:47:                "s3://bucket4/prefix/path?access-key=NXN7IPIOSAAKDEEOLMAF&secret-access-key=nREY/7Dt+PaIbYKrKlEEMMF/ExCiJEX=XMLPUANw",
dm/pkg/storage/utils_test.go:80:                        true, "s3://bucket3/prefix/path?access-key=NXN7IPIOSAAKDEEOLMAF&secret-access-key=nREY/7Dt+PaIbYKrKlEEMMF/ExCiJEX=XMLPUANw",

[ERROR] Matched one or more prohibited patterns

Possible mitigations:
- Mark false positives as allowed using: git config --add secrets.allowed ...
- Mark false positives as allowed by adding regular expressions to .gitallowed at repository's root directory
- List your configured patterns: git config --get-all secrets.patterns
- List your configured allowed patterns: git config --get-all secrets.allowed
- List your configured allowed patterns in .gitallowed at repository's root directory
- Use --no-verify if this is a one-time false positive`

lance6716: You have fixed an error but left a misleading comment. I think you can remove it or write "git secrets will report error when it's a real AK, so we use a truncated one"

WizardXiao: yes, this will be better. this comment is a test and i forget to delete. i will add this comment.

WizardXiao:
> 

yes, this is risk.

WizardXiao:
> And I suddently found that we will print access keys to log because they are in config 😂

That's a risk.

PR #4953, opened 2022-03-20
test(ticdc): migrate test-infra to testify for pkg/util pkg

### LINKED ISSUE #4879: migrate test-infra to testify for `pkg/util` pkg

mikechengwei:

Issue Number: close #4879

This pr migrate test-infra to testify for pkg/util pkg and and fix files that depend on `testleak.AfterTest` method calls

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
```


cdc/sink/causality_test.go
```diff
@@ -215,6 +213,6 @@ func (s *testCausalitySuite) TestGenKeys(c *check.C) {
 		sort.Slice(keys, func(i, j int) bool {
 			return bytes.Compare(keys[i], keys[j]) > 0
 		})
-		c.Assert(keys, check.DeepEquals, tc.expected)
+		require.Equal(t, keys, tc.expected)
```
hi-rustin, line 216:
```suggestion
		require.Equal(t, c.expected, keys)
```

The old and new assertion functions have different signatures, so please make sure you use the correct order for the rest as well.

hi-rustin (changes requested):

hi-rustin:
Thanks for working on this!

cdc/sink/causality_test.go
mikechengwei: ok

hi-rustin:
Also please refine your PR body. Do not submit the template directly as is. You will need to fill in and modify the contents of it.

mikechengwei:
@hi-rustin hi, I found that the `mq_test.go` changes were relatively small, but some unknown errors occurred. Can you help me see?

hi-rustin:
> @hi-rustin hi, I found that the `mq_test.go` changes were relatively small, but some unknown errors occurred. Can you help me see?

I think it's because these tests just haven't been run! mqSinkSuite does not appear to have hook up gocheck into the "go test" runner.

Also, the reason it fails looks like it's because we're updating the resloved ts asynchronously now. So we can't just get the value. You can try adding a delay to the test or adjusting the flush interval in the test.

hi-rustin:
I would also suggest that you split your PRs, as oversized PRs can easily conflict and are not easily reviewed.
I would suggest starting with a few smaller packages.

zhaoxinyu:
Avoid using testify/suite as much as possible. See https://github.com/pingcap/tiflow/issues/2164.
> Avoid using suite of testify as discussed errno: https://github.com/pingcap/tidb/pull/26082#issuecomment-877230972

ben1009:
could all the change ` Files changed 34` separate into small prs instead ?

overvenus:
Really appreciate your work!
Could you split this PR into smaller PRs? It helps code review and speed up merging (by less confliction and less rebase).

mikechengwei:
ok

mikechengwei:
> I think it's because these tests just haven't been run! mqSinkSuite does not appear to have hook up gocheck into the "go test" runner.
> 
> Also, the reason it fails looks like it's because we're updating the resloved ts asynchronously now. So we can't just get the value. You can try adding a delay to the test or adjusting the flush interval in the test.

No problem.I will split this pr.

mikechengwei:
Close this pr,create new pr #5036 .

Issue #3881, opened 2021-12-14
TiCDC: Implement table pipeline with the actor model 


sdojjy:
### Is your feature request related to a problem?

we introduce the actor model to ticdc, to reduce the goroutine size ,implement table pipeline with the actor model 

### Describe the feature you'd like

mplement table pipeline with the actor model

PR #4999, opened 2022-03-23
pipeline(ticdc): enable table actor by default

sdojjy:

Issue Number: ref https://github.com/pingcap/tiflow/issues/3881

enable table actor by default

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
enable table actor by default
```


ben1009 (commented):
i have a stupid question
https://github.com/pingcap/tiflow/blob/cbc17674f4ab2add8f48a0f9f533aa89cb402343/cdc/processor/pipeline/sorter.go#L88

```
func (n *sorterNode) Init(ctx pipeline.NodeContext) error {
	wg := errgroup.Group{}
	return n.start(ctx, false, &wg, 0, nil)
}

func (n *sorterNode) start(ctx pipeline.NodeContext, isTableActorMode bool, eg *errgroup.Group, tableActorID actor.ID, tableActorRouter *actor.Router) error {
	n.isTableActorMode = isTableActorMode
	n.eg = eg
```

this block is not controlled by the config or dead code

sdojjy:
> tiflow/cdc/processor/pipeline/sorter.go

If table actor is not enable, it will be called by nodeRunner https://github.com/pingcap/tiflow/blob/3a5cbcb0de4cfc5f47194cb03d5903c27403d206/pkg/pipeline/runner.go#L58

```
func (r *nodeRunner) run(ctx context.Context) error {
	nodeCtx := NewNodeContext(ctx, Message{}, r.outputCh)
	defer close(r.outputCh)
	defer func() {
		err := r.node.Destroy(nodeCtx)
		if err != nil && errors.Cause(err) != stdContext.Canceled {
			log.Error("found an error when stopping node",
				zap.String("node", r.name), zap.Error(err))
		}
	}()
	err := r.node.Init(nodeCtx)
	if err != nil {
		return err
	}
```

overvenus (approved):

ben1009 (approved):

PR #5011, opened 2022-03-24
scheduler(cdc), tests: fix TestNoFinishOperationBeforeSyncIsReceived and TestClientSendAnomalies

### LINKED ISSUE #4975: TestClientSendAnomalies is unstable
### Which jobs are flaking?

unit test

TestClientSendAnomalies

```log
[2022-03-22T03:15:01.116Z] --- FAIL: TestClientSendAnomalies (0.10s)
[2022-03-22T03:15:01.116Z]     client_test.go:345: 
[2022-03-22T03:15:01.116Z]         	Error Trace:	client_test.go:345
[2022-03-22T03:15:01.116Z]         	Error:      	An error is expected but got nil.
[2022-03-22T03:15:01.116Z]         	Test:       	TestClientSendAnomalies
```

https://ci2.pingcap.net/blue/organizations/jenkins/cdc_ghpr_leak_test/detail/cdc_ghpr_leak_test/23015/pipeline

### LINKED ISSUE #5012: (TiCDC) TestNoFinishOperationBeforeSyncIsReceived is unstable
```
[2022-03-24T06:36:38.649Z] --- FAIL: TestNoFinishOperationBeforeSyncIsReceived (1.11s)

[2022-03-24T06:36:38.649Z]     agent_test.go:516: 

[2022-03-24T06:36:38.649Z]         	Error Trace:	agent_test.go:516

[2022-03-24T06:36:38.649Z]         	Error:      	Condition never satisfied

[2022-03-24T06:36:38.649Z]         	Test:       	TestNoFinishOperationBeforeSyncIsReceived
```

Cause analysis:
```go
require.Never(t, func() bool {
                // Runs in a separate goroutine, which may not terminate after `Never` returns.
		err := agent.Tick(suite.cdcCtx)
		require.NoError(t, err)

		select {
		case <-suite.ctx.Done():
			return true
		case <-suite.dispatchResponseCh:
			return true
		case <-suite.syncCh:
			return true
		default:
			return false
		}
	}, 100*time.Millisecond, 1*time.Millisecond)

	suite.UnblockSync()
        // The last call to the closure in `Never` may happen after the line above, causing `suite.syncCh` to be popped from.
        // This means that the Eventually will never succeed.

	require.Eventually(t, func() bool {
		select {
		case <-suite.ctx.Done():
			require.Fail(t, "context should not be canceled")
			return false
		case syncMsg := <-suite.syncCh:
			require.Equal(t, &model.SyncMessage{
				ProcessorVersion: version.ReleaseSemver(),
				Epoch:            agent.CurrentEpoch(),
				Running:          nil,
				Adding:           nil,
				Removing:         nil,
			}, syncMsg)
			return true
		default:
		}

		err := agent.Tick(suite.cdcCtx)
		require.NoError(t, err)
		return false
	}, 1*time.Second, 10*time.Millisecond)
```
LT;DR: `require.Never` is not synchronous, and using it may cause unexpected result. So I advise against using `require.Never`.

liuzix:

Issue Number: close #4975 close #5012

- Loosened TestClientSendAnomalies. We accept apparent successes of sending messages after the client is closed.
- Replaced `require.Never` in `TestNoFinishOperationBeforeSyncIsReceived`, as `require.Never` will cause data races.

### Check List 

Tests 

 - Unit test

Related changes

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/p2p/client_test.go
+++ b/pkg/p2p/client_test.go
@@ -341,9 +341,11 @@ func TestClientSendAnomalies(t *testing.T) {
 		time.Sleep(100 * time.Millisecond)
 		closeClient()
 	}()
-	_, err = client.SendMessage(ctx, "test-topic", &testMessage{Value: 1})
-	require.Error(t, err)
-	require.Regexp(t, ".*ErrPeerMessageClientClosed.*", err.Error())
+	_, _ = client.SendMessage(ctx, "test-topic", &testMessage{Value: 1})
+	// There is no need to check for error here, because when a client is closing,
+	// message loss is expected because sending the message is fully asynchronous.
+	// The client implementation is considered correct if `SendMessage` does not
+	// block infinitely.
 
 	wg.Wait()
 
--- a/cdc/processor/agent_test.go
+++ b/cdc/processor/agent_test.go
@@ -66,8 +66,6 @@ type agentTestSuite struct {
 	blockSyncMu   sync.Mutex
 	blockSyncCond *sync.Cond
 	blockSync     bool
-
-	testMu sync.Mutex
 }
 
 func newAgentTestSuite(t *testing.T) *agentTestSuite {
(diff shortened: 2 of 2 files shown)


overvenus (approved):

pkg/p2p/client_test.go
```diff
@@ -341,9 +341,11 @@ func TestClientSendAnomalies(t *testing.T) {
...
 		closeClient()
 	}()
-	_, err = client.SendMessage(ctx, "test-topic", &testMessage{Value: 1})
-	require.Error(t, err)
-	require.Regexp(t, ".*ErrPeerMessageClientClosed.*", err.Error())
+	_, _ = client.SendMessage(ctx, "test-topic", &testMessage{Value: 1})
+	// There is no need to check for error here, because when a client is closing,
+	// message loss is expected because sending the message is fully asynchronous.
+	// The client implementation is considered correct if `SendMessage` does not
+	// block infinitely.
```
amyangfei, line 348: Is the timeline like
timeline|-|-
-- | -- | --
t1 | client is closing |  
t2 |   | client.SendMessage, buffer SendChannelSize messages at most
t3 |   | client.SendMessage will block
t4 | client is closed |  
t5 |   | client.SendMessage returns ErrPeerMessageClientClosed

amyangfei (approved):

pkg/p2p/client_test.go
liuzix: Yes

Issue #5017, opened 2022-03-25
improve the backward compatibility of “SHOW SLAVE HOSTS“ command


duanhuichao:
### What did you do?

With the version upgrade, on the new version of MySQL Cluster, the query results of many system commands will display some new columns. These new columns show new features or information of MySQL.

In general, it is recommended to provide backward compatibility of system commands (e.g., show xxx) rather than hard-coded return value format (i.e., limiting the number of returned columns).

For DM, there are some commands that are difficult to backward-compatible:
(1) SHOW MASTER STATUS
(2) SHOW SLAVE HOSTS
(3) SHOW BINARY LOGS

These commands limit the return format by limiting the number of columns returned:
```
rows, err := db.QueryContext(ctx, `SHOW SLAVE HOSTS`)
rowColumns, err := rows.Columns()
if len(rowColumns) == 5 {
	err = rows.Scan(&serverID, &host, &port, &masterID, &slaveUUID)
} else {
	err = rows.Scan(&serverID, &host, &port, &masterID)
}
```

### What did you expect to see?

Through the reflection mechanism, we only take the value of the required corresponding column and do not limit the number of returned columns.

import (
    "reflect"
)

- [Rows.Columns](http://golang.org/pkg/database/sql/#Rows.Columns) to get column names and positions.
- Passing a slice of pointers []interface{} to [Rows.Scan](https://golang.org/pkg/database/sql/#Rows.Scan) to get the values.
- [reflect.Value](https://golang.org/pkg/reflect/#Value) and [Value.Addr](https://golang.org/pkg/reflect/#Value.Addr) to get a pointer to the destination value.
- [Value.FieldByName](https://golang.org/pkg/reflect/#Value.FieldByName) to get the Value of a struct field if you want to map to struct fields.

see: 
https://stackoverflow.com/questions/21986780/is-it-possible-to-retrieve-a-column-value-by-name-using-golang-database-sql

### What did you see instead?

For the above system commands, when the number of columns returned by MySQL changes, DM cannot work.

### Versions of the cluster

any versions

lance6716:
Hi, which version of MySQL addes new columns? And please address comment in your PR

ryangao0513:
Hello,I encountered the same problem in txsql 8.0.22 patch 20210330

lance6716:
happens again https://asktug.com/t/topic/813202

PR #5018, opened 2022-03-25
(dm) [bugfix] backward compatibility with show command #5017

duanhuichao:
Problem:
With the version upgrade, on the new version of MySQL Cluster,
the query results of many system commands will display some new
columns. These new columns show new features or information of MySQL.

In general, it is recommended to provide backward compatibility of
system commands (e.g., show xxx) rather than hard-coded return value
format (i.e., limiting the number of returned columns).

For DM, there are some commands that are difficult to backward-compatible:
(1) SHOW MASTER STATUS
(2) SHOW SLAVE HOSTS
(3) SHOW BINARY LOGS

Fix:
Through the reflection mechanism, we only take the value of the required
corresponding column and do not limit the number of returned columns.
This patch removes the limit on the number of columns returned by the
'show slave hosts' command.

ref pingcap/tiflow#5017


lance6716:
@duanhuichao Hi, please follow the 2nd & 3rd comment of ti-chi-bot to adjust PR description

lance6716:
I found we have provided a helper function to do similar thing in https://github.com/pingcap/tidb/blob/9b9051f113a6f19857f6a8aa1f28159cb1ec40cc/dumpling/export/sql.go#L694

Could you use that function instead of using reflect? It's more easier to understand.

lonng:
@duanhuichao Thanks for you contribution, and can you add some tests for this PR?

lance6716:
Hi @duanhuichao, another user also meet this problem, so we are going to fix this problem in these days. Can you check the comments to finish your PR?

If you don't have time to response, we will fix it on our own in next week. Thanks for your contribution to find the problem!

lance6716:
replaced by https://github.com/pingcap/tiflow/pull/7372

PR #5025, opened 2022-03-25
validator(dm): checkpoint unit-test & fixing some todos

### LINKED ISSUE #4770: validator(dm): checkpoint, persist pending/error row change data, and table status

D3Hunter:

Issue Number: close #4770

- add unit test for validator checkpoint
- add more statistics infos: pending/error row changes if different kind of row changes
- drop validator meta-tables when `remove-meta`
- make downstream query batch size configurable(default to 100)
- add `startWithSubtask` which means validator starts together with subtask, in this case we use the init sync location as our init validate location. and persist immediately if there's no checkpoint now
- validator should be stopped when `operate-schema`
- add another layer of synchronization to `GetTableInfo/RecreateTables` of tracker, to make sure validator doesn't read an intermediate dropped table.
- depends on methods of syncer not its fields
- if uk of data row is too long, convert it to sha256 digest, to make sure we can put it into dm_meta since it's part of the primary key.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`
```


dm/syncer/syncer.go
```diff
@@ -3962,3 +3970,26 @@ func (s *Syncer) setGlobalPointByTime(tctx *tcontext.Context, timeStr string) er
...
+	return s.checkpoint.FlushedGlobalPoint()
+}
+
+func (s *Syncer) getInitExecutedLoc() binlog.Location {
+	s.RLock()
+	defer s.RUnlock()
+	return s.initExecutedLoc.Clone()
+}
+
+func (s *Syncer) initInitExecutedLoc() {
```
Ehco1996, line 3984: do we need set `currentLocation` and `startLocation` and `lastLocation` with this func?

dm/dm/worker/subtask.go
```diff
@@ -665,6 +674,10 @@ func (st *SubTask) OperateSchema(ctx context.Context, req *pb.OperateWorkerSchem
 		return "", terror.ErrWorkerOperSyncUnitOnly.Generate(st.currUnit.Type())
 	}
 
+	if st.validatorStage() == pb.Stage_Running && req.Op != pb.SchemaOp_ListMigrateTargets {
+		return "", terror.ErrWorkerNotPausedStage.Generate(pb.Stage_Running.String())
```
Ehco1996, line 678: maybe add some hints about validator in this error

dm/syncer/data_validator.go
```diff
@@ -764,7 +836,17 @@ func getRowChangeType(t replication.EventType) rowChangeType {
...
-	return strings.Join(pkValues, "-")
+	// TODO: in scenario below, the generated key may not unique, but it's rare
+	// suppose a table with multiple column primary key: (v1, v2)
+	// for below case, the generated key is the same:
+	// 	 (aaa\t, bbb) and (aaa, \tbbb), the joint values both are "aaa\t\tbbb"
+	join := strings.Join(pkValues, "\t")
+	// if the key is too long, need to make sure it can be stored into database
+	if len(join) > maxRowKeyLength {
+		sum := sha256.Sum256([]byte(join))
+		return hex.EncodeToString(sum[:])
```
Ehco1996, line 847: why we do this opt ? can we just report an error? or we can change the meta schema ?

Ehco1996 (commented):
will reiew test later

dm/syncer/data_validator.go
D3Hunter: pk of user table is part of our pk in meta-schema, so if this pk is long enough(see 3000 bytes), our pk will exceed the limit of mysql(3072 bytes), so make it a hash digest then we can store.

in most case the pk of user table is bigint, it should not go into this branch.

dm/syncer/syncer.go
D3Hunter: ExecutedLoc(executed comes from gtid_executed from mysql, this word tells more information than `start` i think) is like the initial start location.

we can init curr/last location later when we need them.

dm/dm/worker/subtask.go
D3Hunter: https://github.com/pingcap/tiflow/pull/5025/commits/06b8febc7b53814756ea6edac66f173a36d090c1

D3Hunter:
ping @lance6716

D3Hunter:
ping @buchuitoudegou

dm/pkg/schema/tracker.go
```diff
@@ -414,7 +421,27 @@ func (tr *Tracker) CreateTableIfNotExists(table *filter.Table, ti *model.TableIn
...
 }
 
-func (tr *Tracker) BatchCreateTableIfNotExist(tablesToCreate map[string]map[string]*model.TableInfo) error {
+func (tr *Tracker) RecreateTables(logCtx *tcontext.Context, tablesToDrop []*filter.Table, tablesToCreate map[string]map[string]*model.TableInfo) error {
+	tr.Lock()
+	defer tr.Unlock()
+	for _, tbl := range tablesToDrop {
+		// schema changed
+		if err := tr.DropTable(tbl); err != nil {
+			logCtx.L().Warn("failed to drop table from schema tracker",
```
lance6716, line 430: this should be an error level log?

lance6716 (approved):

dm/pkg/schema/tracker.go
```diff
@@ -414,7 +421,27 @@ func (tr *Tracker) CreateTableIfNotExists(table *filter.Table, ti *model.TableIn
...
+		}
+	}
+	for schemaName := range tablesToCreate {
+		// TODO: Figure out how to recover from errors.
+		if err := tr.CreateSchemaIfNotExists(schemaName); err != nil {
+			logCtx.L().Error("failed to rollback schema on schema tracker: cannot create schema",
+				zap.String("schema", schemaName), log.ShortError(err))
+		}
+	}
+	return tr.batchCreateTableIfNotExist(tablesToCreate)
```
buchuitoudegou, line 441: What will happen if the tracker failed to create the corresponding schema before creating the tables? Should we place this to the `err == nil` branch or let the tracker return error immediately after failing to create schemas?

dm/dm/worker/subtask.go
D3Hunter: > I think we should only forbid operations that will change the schema.

yes, if subtask or validator is running we should  permit schema read operation, not write. i copied from previous branch, seems a bug. not sure if it affects v6.0, @Ehco1996 ptal.

dm/pkg/schema/tracker.go
D3Hunter: it's ignored before, copied from

https://github.com/pingcap/tiflow/blob/1a95ac2b5ce472048ecae418aa8b279c7f0d3e59/dm/syncer/checkpoint.go#L913-L920

will leave it this way now.

dm/dm/worker/subtask.go
Ehco1996: I think it is possible to allow read operations even when the task/validator is running,
but, in the [old logic](https://github.com/pingcap/dm/blob/63f574362d0aa1452e43eca3feab4a9ca2d32ee2/dm/worker/subtask.go#L570), OperateSchema was not allowed when the task was not pasued, we can improve this later so it not affects 6.0

Ehco1996 (approved):
lgtm

dm/pkg/schema/tracker.go
buchuitoudegou: Is it a bug or by design? @Ehco1996 @lance6716 🤔

buchuitoudegou (approved):

dm/pkg/schema/tracker.go
D3Hunter: this error happens rarely since it's a embedded tidb, no connection issues, but if tracker's data directory is corrupted or disk full, may cause error. in that case it's hard to recover too, maybe that's why just add a todo.

PR #5031, opened 2022-03-25
cdc/cli: hidden sorter parameters from server command.

### LINKED ISSUE #5029: hidden sorter related parameters from CDC server command.
For the TiUP, the `sorter` related parameter cannot be easily configured, have to use the config file instead. So we hid those parameters.

3AceShowHand:

Issue Number: close #5029 

before:
Screen Shot 2022-03-25 at 3 31 14 PM

after:

image

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`.
```


overvenus (approved):

zhaoxinyu (approved):

hi-rustin (commented):
I wonder if it is possible to put it close to the next line of the added parameter when there is such a large number of hidden parameters.
Otherwise it is easy to be ignored and not easy to find.

Issue #4976, opened 2022-03-22
Scale-in CDC to 1 node, owner exited and failed to start


Tammyxia:
### What did you do?

- 100k table count, cdc 8nodes, 1 changefeed, tikv 6 nodes
- Workload is go-ycsb
- Scale-out TiKV to 7 nodes
- Scale-in TiCDC to 1 node.

### What did you expect to see?

No any error

### What did you see instead?

- CDC owner exited and failed to start from 03/22 00:03, after 12 hours, owner still cannot start.
![image](https://user-images.githubusercontent.com/78345569/159408394-18c7c0df-dcbb-40dc-b26c-ed06e5823492.png)
CDC-0 log: [2022/03/21 16:03:29.891 +00:00] [INFO] [capture.go:436] ["run owner exited"] [error="[CDC:ErrCaptureSuicide]capture suicide"] [errorVerbose="[CDC:ErrCaptureSuicide]capture suicide\ngithub.com/pingcap/errors.AddStack\n\tgithub.com/pingcap/errors@v0.11.5-0.20211224045212-9687c2b0f87c/errors.go:174\ngithub.com/pingcap/errors.(*Error).GenWithStackByArgs\n\tgithub.com/pingcap/errors@v0.11.5-0.20211224045212-9687c2b0f87c/normalize.go:164\ngithub.com/pingcap/tiflow/cdc/capture.(*Capture).runEtcdWorker\n\tgithub.com/pingcap/tiflow/cdc/capture/capture.go:471\ngithub.com/pingcap/tiflow/cdc/capture.(*Capture).campaignOwner\n\tgithub.com/pingcap/tiflow/cdc/capture/capture.go:434\ngithub.com/pingcap/tiflow/cdc/capture.(*Capture).run.func2\n\tgithub.com/pingcap/tiflow/cdc/capture/capture.go:298\nruntime.goexit\n\truntime/asm_amd64.s:1571"]
[...]

PR #5042, opened 2022-03-28
actor(ticdc): add actor log

sdojjy:

Issue Number: ref #4976

add log for actor goroutine lifetime

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/sink/black_hole.go
+++ b/cdc/sink/black_hole.go
@@ -25,7 +25,8 @@ import (
 // newBlackHoleSink creates a black hole sink
 func newBlackHoleSink(ctx context.Context) *blackHoleSink {
 	return &blackHoleSink{
-		statistics: NewStatistics(ctx, "blackhole"),
+		// use `sinkTypeDB` to record metrics
+		statistics: NewStatistics(ctx, sinkTypeDB),
 	}
 }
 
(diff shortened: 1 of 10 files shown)


pkg/actor/system.go
```diff
@@ -447,7 +447,15 @@ func (s *System) Start(ctx context.Context) {
 	for i := 0; i < s.numWorker; i++ {
 		id := i
 		s.wg.Go(func() error {
-			defer pprof.SetGoroutineLabels(ctx)
+			defer func() {
+				log.Info("actor goroutine exited",
+					zap.Int("id", id),
+					zap.String("name", s.name))
+				pprof.SetGoroutineLabels(ctx)
```
overvenus, line 454:
```suggestion
```

We don't need to set labels as it is exiting.

pkg/actor/system.go
```diff
@@ -566,6 +574,9 @@ func (s *System) poll(ctx context.Context, id int) {
 			if !running {
 				// Close the actor and mailbox.
 				p.onActorClosed()
+				log.Info("actor is closed",
+					zap.Uint64("id", uint64(p.mb.ID())),
+					zap.String("name", s.name))
```
overvenus, line 579: We need to add the log after `p.onMailboxEmpty` too.

overvenus: It may be very verbose, when there are many tables.

sdojjy: removed this log

3AceShowHand (approved):

overvenus (approved):

PR #5067, opened 2022-03-30
syncer(dm): add retry for trackTableInfoFromDownstream

### LINKED ISSUE #5015: Add retry for trackTableInfoFromDownstream

```
[2022/03/24 17:41:35.277 +00:00] [ERROR] [subtask.go:331] ["unit process error"] [subtask=test] [unit=Sync] ["error information"="ErrCode:44008 ErrClass:\"schema-tracker\" ErrScope:\"internal\" ErrLevel:\"high\" Message:\"startLocation: [position: (mysql-bin-changelog.000003, 883), gtid-set: ], endLocation: [position: (mysql-bin-changelog.000003, 1003), gtid-set: ], origin SQL: [alter table tbl_1 modify source_id varchar(32)]: cannot parse downstream table schema of `test_1`.`tbl_1` to initialize upstream schema `test`.`tbl_1` in schema tracker\" RawCause:\"driver: bad connection\" "]
```

Add a try mechanism like what we do when execute sql.

_No response_

_No response_

buchuitoudegou:

Issue Number: close #5015 

make `GetParserForConn` retryable. Using `utils.GetParserForConn` directly connects to the downstream database via `sql.Conn` having no retry strategy. Instead, implementing a new `GetParserForConn` with `DBConn` and replacing the old one can implicitly utilize the retry strategy wrapped in `DBConn` and thus, this function can automatically retry itself.

Other functions like `GetTableCreateSQL` are also replaced by the `DBConn`'s version provisioning recoverability.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`
```


buchuitoudegou:
will add a ut later.

dm/syncer/syncer.go
```diff
@@ -756,57 +757,70 @@ func (s *Syncer) getTableInfo(tctx *tcontext.Context, sourceTable, targetTable *
...
+		RetryCount:         10,
+		FirstRetryDuration: retryTimeout,
+		BackoffStrategy:    retry.Stable,
+		IsRetryableFn: func(_ int, _ error) bool {
+			// retry until reaching limit
+			return true
+		},
 	}
+	_, _, err := retryStrategy.Apply(tctx, params, func(tctx *tcontext.Context) (interface{}, error) {
+		parser2, err := utils.GetParserForConn(tctx.Ctx, s.ddlDBConn.BaseConn.DBConn)
```
Ehco1996, line 771: i found tracker has its own conn which set `mysql.DefaultSQLMode` , this may cause the sqlmode to be inconsistent in two places ? see [here](https://github.com/pingcap/tiflow/blob/a9d5d5a6d9120900fdf49ba670cdbaf0b82c04fb/dm/syncer/syncer.go#L3405) and [here](https://github.com/pingcap/tiflow/blob/a9d5d5a6d9120900fdf49ba670cdbaf0b82c04fb/dm/pkg/schema/tracker.go#L543)
cc @WizardXiao

lance6716 (commented):
I prefer we use `DBConn.QuerySQL` rather than we wrap retry again

lance6716:
I think we can make `DBConn.BaseConn` private, so the caller must use provided methods which has considered retry. Maybe a lot of work to do.

buchuitoudegou:
> I prefer we use `DBConn.QuerySQL` rather than we wrap retry again

Did you mean we should make every step retryable rather than roughly retry the whole function?

buchuitoudegou:
> I think we can make `DBConn.BaseConn` private, so the caller must use provided methods which has considered retry. Maybe a lot of work to do.

I will take a look about it👀

dm/syncer/syncer.go
WizardXiao: set `mysql.DefaultSQLMode` is to advoid downstream sql_mode changing and we can not parse the create sql, so we set a high compatibility sql_mode.

Ehco1996: if so can we just use `mysql.DefaultSQLMode` here ?

Ehco1996: if ok to use mysql.DefaultSQLMode(for compatibility), no need retry anymore
 cc @buchuitoudegou @lance6716

lance6716: > no need retry anymore

please read the issue to know what problem retry resolves

buchuitoudegou:
PTAL again @lance6716

dm/syncer/syncer.go
buchuitoudegou: > if ok to use mysql.DefaultSQLMode(for compatibility), no need retry anymore cc @buchuitoudegou @lance6716

This PR attempts to make tracker automatically retry for SQL querying when querying fails due to some network issue.

dm/syncer/dbconn/db.go
```diff
@@ -57,25 +57,32 @@ func CloseBaseDB(logCtx *tcontext.Context, baseDB *conn.BaseDB) {
 // it's not thread-safe.
 type DBConn struct {
 	Cfg      *config.SubTaskConfig
```
lance6716, line 59: maybe Cfg is also not needed to be public?

lance6716 (approved):

dm/syncer/syncer.go
Ehco1996: ohoho mybad, i  for GetTableCreateSQL,I missed this `GetTableCreateSQL ` before, I thought you just want add retry for `GetParserForConn `

dm/syncer/syncer_test.go
```diff
@@ -774,11 +774,13 @@ func (s *testSyncerSuite) TestRun(c *C) {
 	syncer.cfg.CheckpointFlushInterval = 30
 	syncer.fromDB = &dbconn.UpStreamConn{BaseDB: conn.NewBaseDB(db)}
 	syncer.toDBConns = []*dbconn.DBConn{
-		{Cfg: s.cfg, BaseConn: conn.NewBaseConn(dbConn, &retry.FiniteRetryStrategy{})},
-		{Cfg: s.cfg, BaseConn: conn.NewBaseConn(dbConn, &retry.FiniteRetryStrategy{})},
+		// {Cfg: s.cfg, BaseConn: conn.NewBaseConn(dbConn, &retry.FiniteRetryStrategy{})},
```
Ehco1996, line 777: please remove this useless code

Ehco1996 (approved):
restLGTM

Issue #5063, opened 2022-03-30
DM-worker may has high CPU usage and flood log after start a GTID task


lance6716:
### What did you do?

the bug has following prerequisites:
- v5.4.0 or v6.0.0, and
- use relay log
- `enable-gtid: true` in upstream config, and
- start a `all` task when last upstream mysql binlog file has a large size or start a `incremental` task from a middle position of binlog file, or the task is auto resumed at a middle position of binlog file

### What did you expect to see?

works fine

### What did you see instead?

- task almost doesn't get forward after start-task, and
- DM-worker has high CPU (360% in my local PC, which is 12 core), and
- generates lots of log with

```
[2022/03/30 11:30:46.431 +08:00] [INFO] [syncer.go:2020] ["meet heartbeat event and then flush jobs"] [task=test] [unit="binlog replication"]
[2022/03/30 11:30:46.431 +08:00] [INFO] [syncer.go:3247] ["flush all jobs"] [task=test] [unit="binlog replication"] ["global checkpoint"="{{{mysql-bin.000001 113080239} 0xc000010ff8 0} <nil>}(flushed {{{mysql-bin.000001 113080239} 0xc000011208 0} <nil>})"] ["flush job seq"=37]
[2022/03/30 11:30:46.432 +08:00] [INFO] [syncer.go:1114] ["checkpoint has no change, skip sync flush checkpoint"]
```

**For `all` mode task, the problematic duration is related to last upstream mysql binlog file size.**

**For `incremental` task, the problematic duration is related to the specified starting binlog location**

### Versions of the cluster

DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):
```console
v5.4.0, v6.0.0
```
Upstream MySQL/MariaDB server version:
```console
(paste upstream MySQL/MariaDB server version here)
```
Downstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):
```console
(paste TiDB cluster version here)
```
How did you deploy DM: tiup or manually?
```console
(leave TiUP or manually here)
```
Other interesting information (system version, hardware config, etc):
```console
>
>
```

### current status of DM cluster (execute `query-status <task-name>` in dmctl)

```console
(paste current status of DM cluster here)
```

lance6716:
@GMHDBJD we implemented wrong logic about skipped GTID in relay log reader.

[...]

PR #5070, opened 2022-03-30
relay(dm): send one heartbeat for successive skipped GTID

lance6716:

Issue Number: ref #5063 

see https://github.com/pingcap/tiflow/issues/5063#issuecomment-1082678211 

### Check List 

Tests 

 - Integration test
 - Unit tests

Related changes

### Release note 

```release-note
send one heartbeat for successive skipped GTID when enable relay log
```


dm/relay/local_reader.go
```diff
@@ -585,19 +588,24 @@ func (r *BinlogReader) parseFile(
...
-		// for Mariadb, it will bee replaced with MARIADB_GTID_LIST_EVENT
-		// In DM, we replace both of them with HEARTBEAT event
-		if state.replaceWithHeartbeat {
+		// ref https://github.com/pingcap/tiflow/issues/5063#issuecomment-1082678211
+		// heartbeat period is implemented in LocalStreamer.GetEvent
+		if state.skipGTID {
 			switch e.Event.(type) {
 			// Only replace transaction event
 			// Other events such as FormatDescriptionEvent, RotateEvent, etc. should be the same as before
 			case *replication.RowsEvent, *replication.QueryEvent, *replication.GTIDEvent, *replication.XIDEvent, *replication.TableMapEvent:
```
D3Hunter, line 597: seems we missed `*replication.MariadbGTIDEvent`

dm/relay/local_reader.go
```diff
@@ -585,19 +588,24 @@ func (r *BinlogReader) parseFile(
...
-				e = event.GenHeartbeatEvent(e.Header)
+				state.lastSkipGTIDHeader = e.Header
 			default:
 			}
+			return nil
+		} else if lastSkipGTID && state.lastSkipGTIDHeader != nil {
+			// skipGTID is turned off after this event
+			select {
+			case s.ch <- event.GenHeartbeatEvent(state.lastSkipGTIDHeader):
+			case <-ctx.Done():
```
D3Hunter, line 607: prefer remove this branch, since we don't have strong dependency on heartbeat event and `LocalStreamer.GetEvent` has simulate normal heartbeat.

lance6716: mainly to reduce the behaving difference with MySQL, the original purpose is stated in https://github.com/pingcap/dm/pull/1390

D3Hunter (approved):

dm/relay/local_reader.go
```diff
@@ -585,19 +588,25 @@ func (r *BinlogReader) parseFile(
...
+			case *replication.RowsEvent, *replication.QueryEvent, *replication.GTIDEvent,
+				*replication.MariadbGTIDEvent, *replication.XIDEvent, *replication.TableMapEvent:
 				// replace with heartbeat event
-				e = event.GenHeartbeatEvent(e.Header)
+				state.lastSkipGTIDHeader = e.Header
 			default:
 			}
+			return nil
+		} else if lastSkipGTID && state.lastSkipGTIDHeader != nil {
+			// skipGTID is turned off after this event
```
Ehco1996, line 605:
```suggestion
			// lastSkipGTID is turned off after this event
```

Ehco1996 (commented):
lgtm

lance6716:
PTAL @GMHDBJD

dm/relay/local_reader.go
Ehco1996: I read it again and it looks like skipGTID will not be reset because r.advanceCurrentGtidSet(gtidStr) will always return true here 😵

D3Hunter: return false when met un-synced gtids

Ehco1996: thanks, I figured it out, now only send heartbeat when a new event comining and last event is skipped

GMHDBJD (approved):
LGTM

PR #5103, opened 2022-04-02
checker(dm): init checker concurrently

### LINKED ISSUE #5098: dm checker init too slow when there are a lot of sources

The current DM init checker will be started sequentially according to the number of sources, so the time taken will increase linearly with the number of sources， see [here](https://github.com/pingcap/tiflow/blob/master/dm/checker/checker.go#L129)

In addition, there will be external IO in [here](https://github.com/pingcap/tiflow/blob/69485890b08b25405cd50e8085f271d532e674dd/dm/checker/checker.go#L140) and [here](https://github.com/pingcap/tiflow/blob/69485890b08b25405cd50e8085f271d532e674dd/dm/checker/checker.go#L181), so when there are a lot of sources, init will be unacceptably slow

speed up the init

_No response_

_No response_

Ehco1996:

Issue Number: close #5098

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note

`support init checker concurrently`.
```


dm/checker/checker.go
```diff
@@ -65,6 +66,8 @@ type mysqlInstance struct {
 
 	targetDB     *conn.BaseDB
 	targetDBInfo *dbutil.DBConfig
+
+	bw *filter.Filter
```
GMHDBJD, line 70: ba?

dm/checker/checker.go
```diff
@@ -119,75 +122,48 @@ func (c *Checker) Init(ctx context.Context) (err error) {
...
+	eg, ctx2 := errgroup.WithContext(ctx)
+	sourceTargetM := make(map[*mysqlInstance]map[string][]*filter.Table)
+	for idx := range c.instances {
+		i := idx
+		eg.Go(func() error {
+			mapping, fetchErr := c.fetchSourceTargetDB(ctx2, c.instances[i])
+			if fetchErr != nil {
+				return fetchErr
+			}
+			sourceTargetM[c.instances[i]] = mapping
```
lance6716, line 136: data race

dm/checker/checker.go
```diff
@@ -119,75 +122,48 @@ func (c *Checker) Init(ctx context.Context) (err error) {
 	rollbackHolder.Add(fr.FuncRollback{Name: "close-DBs", Fn: c.closeDBs})
 
 	c.tctx = tcontext.NewContext(ctx, log.With(zap.String("unit", "task check")))
+
+	// prepare source target do dbs concurrently
+	eg, ctx2 := errgroup.WithContext(ctx)
+	sourceTargetM := make(map[*mysqlInstance]map[string][]*filter.Table)
```
lance6716, line 128: can you use a more common type as map key?

dm/checker/checker.go
```diff
@@ -119,75 +122,51 @@ func (c *Checker) Init(ctx context.Context) (err error) {
 	rollbackHolder.Add(fr.FuncRollback{Name: "close-DBs", Fn: c.closeDBs})
 
 	c.tctx = tcontext.NewContext(ctx, log.With(zap.String("unit", "task check")))
+
+	// prepare source target do dbs concurrently
+	eg, ctx2 := errgroup.WithContext(ctx)
+	// sourceID => source targetDB mapping, only writing need used mu to avoid concurrent write map
+	var mu sync.Mutex
+	sourceTargetM := make(map[string]map[string][]*filter.Table)
```
lance6716, line 130: nit: if you use preallocated slice to store `map[string][]*filter.Table` and let i-th c.instances access i-th element, there should be no data race

lance6716 (approved):

dm/checker/checker.go
```diff
@@ -65,6 +66,8 @@ type mysqlInstance struct {
 
 	targetDB     *conn.BaseDB
 	targetDBInfo *dbutil.DBConfig
+
+	bAList *filter.Filter
```
okJiang, line 70: `baList` is more common in repo.

dm/checker/checker.go
```diff
@@ -119,75 +122,51 @@ func (c *Checker) Init(ctx context.Context) (err error) {
 	rollbackHolder.Add(fr.FuncRollback{Name: "close-DBs", Fn: c.closeDBs})
 
 	c.tctx = tcontext.NewContext(ctx, log.With(zap.String("unit", "task check")))
+
+	// prepare source target do dbs concurrently
+	eg, ctx2 := errgroup.WithContext(ctx)
+	// sourceID => source targetDB mapping, only writing need used mu to avoid concurrent write map
+	var mu sync.Mutex
+	sourceTargetM := make(map[string]map[string][]*filter.Table)
+	for idx := range c.instances {
```
okJiang, line 131: May the number of instances be too big to effect performance?

okJiang (commented):
rest LGTM

dm/checker/checker.go
```diff
@@ -119,75 +122,51 @@ func (c *Checker) Init(ctx context.Context) (err error) {
...
+		i := idx
```
buchuitoudegou, line 132: Not suggestion. I'm just curious about why idx has to be copied before being used? It seems idx is an integer, which should be copied instead of being moved. (I'm not familiar with golang, would there be any problems if we use the idx directly in the closure by catching it from the outside space? Or idx is caught as ref by default?)

buchuitoudegou (approved):

dm/checker/checker.go
Ehco1996: this is strange in go without passing copy/val, then the value of val will be whatever val is when the goroutines actually runs see more here https://riptutorial.com/go/example/3897/using-closures-with-goroutines-in-a-loop

dm/checker/checker.go
Ehco1996: Not too slow when external IOs are in parallel

okJiang (approved):

PR #5106, opened 2022-04-02
cdc,pkg(ticdc): enable more linters

### LINKED ISSUE #5105: Enable some linters that find bugs
As title

overvenus:

Issue Number: close #5105

Add more linters.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

### Release note 

```release-note
None
```


zhaoxinyu:
leak test TestBatch failed. @hi-rustin PTAL

cdc/sink/mq_flush_worker_test.go
```diff
@@ -176,6 +176,8 @@ func TestBatch(t *testing.T) {
 	batch := make([]mqEvent, 3)
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
```
hi-rustin, line 179: We should not turn on parallel here because batch was reused and the worker itself is not thread-safe.

hi-rustin (changes requested):

cdc/server.go
```diff
@@ -258,17 +258,11 @@ func (s *Server) etcdHealthChecker(ctx context.Context) error {
...
-					log.Warn("etcd health check failed", zap.Error(err))
-					cancel()
-					continue
-				}
-				_, err = httpCli.Do(req)
+				resp, err := httpCli.Get(ctx, fmt.Sprintf("%s/pd/api/v1/health", pdEndpoint))
 				if err != nil {
 					log.Warn("etcd health check error", zap.Error(err))
 				} else {
+					_ = resp.Body.Close()
```
zhaoxinyu, line 265:
```suggestion
		        	_, _ = io.Copy(ioutil.Discard, resp.Body)
					_ = resp.Body.Close()
```

zhaoxinyu, line 265: It's better to discard the response body to reuse the underlying TCP connection.

cdc/sink/codec/schema_registry.go
```diff
@@ -337,6 +333,7 @@ func (m *AvroSchemaManager) ClearRegistry(ctx context.Context, tableName model.T
 	if err != nil {
 		return err
 	}
+	defer resp.Body.Close()
```
zhaoxinyu, line 336: Ditto

cdc/server.go
overvenus: Is there any reference?

zhaoxinyu (approved):

cdc/server.go
zhaoxinyu: https://github.com/golang/go/blob/4d8db00641cc9ff4f44de7df9b8c4f4a4f9416ee/src/net/http/response.go#L64

cdc/sink/mq/mq_flush_worker_test.go
```diff
@@ -173,11 +172,15 @@ func TestBatch(t *testing.T) {
...
-	var wg sync.WaitGroup
 	ctx := context.Background()
-	batch := make([]mqEvent, 3)
-	for _, test := range tests {
+	for i := range tests {
+		test := tests[i]
 		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			batch := make([]mqEvent, 3)
```
hi-rustin, line 181: We shouldn't create a new batch here, it turns out to test its correctness when using the same block of memory.

hi-rustin (commented):
rest LGTM

hi-rustin (approved):
Approve in advance, only the small problem above

hi-rustin:
/hold

Feel free unhold when https://github.com/pingcap/tiflow/pull/5106#discussion_r857451820 addressed.

hi-rustin (approved):
/unhold

Issue #5076, opened 2022-03-30
flaky test-  multi-topic integration test fail


maxshuang:
### Which jobs are flaking?

https://ci2.pingcap.net/blue/organizations/jenkins/cdc_ghpr_kafka_integration_test/detail/cdc_ghpr_kafka_integration_test/10263/pipeline/134

### Which test(s) are flaking?

multi-topic

### Jenkins logs or GitHub Actions link

```log
check diff failed 300-th time, retry later
[2022-03-30T10:42:36.281Z] check diff failed at last
[2022-03-30T10:42:36.281Z] A total of 3 tables need to be compared
[2022-03-30T10:42:36.281Z] 
[2022-03-30T10:42:36.281Z] 
[2022-03-30T10:42:36.281Z] [0A[JComparing the table structure of ``test`.`test2`` ... equivalent
[2022-03-30T10:42:36.281Z] Comparing the table data of ``test`.`test2`` ...
[2022-03-30T10:42:36.281Z] _____________________________________________________________________________
[2022-03-30T10:42:36.281Z] Progress [>------------------------------------------------------------] 0% 0/1
[2022-03-30T10:42:36.281Z] [3A[JComparing the table structure of ``test`.`test3`` ... equivalent
[2022-03-30T10:42:36.281Z] Comparing the table data of ``test`.`test3`` ...
[2022-03-30T10:42:36.281Z] Comparing the table data of ``test`.`test2`` ...
[2022-03-30T10:42:36.281Z] _____________________________________________________________________________
[2022-03-30T10:42:36.281Z] Progress [>------------------------------------------------------------] 0% 0/2
[2022-03-30T10:42:36.281Z] [4A[JComparing the table structure of ``test`.`test1`` ... equivalent
[2022-03-30T10:42:36.281Z] Comparing the table data of ``test`.`test3`` ...
[2022-03-30T10:42:36.281Z] Comparing the table data of ``test`.`test2`` ...
[2022-03-30T10:42:36.281Z] Comparing the table data of ``test`.`test1`` ...
[2022-03-30T10:42:36.281Z] _____________________________________________________________________________
[2022-03-30T10:42:36.281Z] Progress [>------------------------------------------------------------] 0% 0/3
[2022-03-30T10:42:36.281Z] [5A[JComparing the table data of ``test`.`test1`` ... equivalent
[2022-03-30T10:42:36.281Z] Comparing the table data of ``test`.`test3`` ...
[...]

PR #5116, opened 2022-04-05
sink/mq(ticdc): make the topic manager use the correct number of partitions

hi-rustin:

Issue Number: close https://github.com/pingcap/tiflow/issues/5076

make the topic manager use the correct number of partitions.

It uses the wrong number of partitions when kafka turns on auto-create configuration.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

None

Related changes

None

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/sink/manager/kafka/manager.go
+++ b/cdc/sink/manager/kafka/manager.go
@@ -112,7 +112,8 @@ func (m *TopicManager) tryUpdatePartitionsAndLogging(topic string, partitions in
 	}
 }
 
-// CreateTopic creates a topic with the given name.
+// CreateTopic creates a topic with the given name
+// and returns the number of partitions.
 func (m *TopicManager) CreateTopic(topicName string) (int32, error) {
 	start := time.Now()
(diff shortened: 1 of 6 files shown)


zhaoxinyu (approved):

cdc/sink/manager/kafka/manager.go
```diff
@@ -106,27 +106,27 @@ func (m *TopicManager) tryUpdatePartitionsAndLogging(topic string, partitions in
...
 		log.Info(
 			"store topic partition number",
-			zap.String("duration", topic),
+			zap.String("topic", topic),
 			zap.Int32("partitionNumber", partitions),
 		)
 	}
 }
 
 // CreateTopic creates a topic with the given name.
```
asddongmen, line 115: Can you add a comment to explain the meaning of the return `int32` value?

hi-rustin: Good catch! Done!

asddongmen (approved):

PR #5130, opened 2022-04-08
dm/ha: rolling update support to new etcd keys

lichunzhu:

Issue Number: close https://github.com/pingcap/tiflow/issues/5099

Support rolling upgrade in dm. We didn't keep old dm etcd keys to avoid users deploy both old/new versions of dm-workers.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

Related changes

- None

### Release note 

```release-note
None
```

### DIFF
--- a/dm/dm/ctl/master/config.go
+++ b/dm/dm/ctl/master/config.go
@@ -337,9 +337,9 @@ func getRelayCfgs(cli *clientv3.Client) (map[string]map[string]struct{}, error) 
 	if err != nil {
 		return nil, err
 	}
-	// try to get all source cfgs before v6.1.0
+	// try to get all source cfgs before v6.2.0
 	if len(relayCfgsMap) == 0 {
-		relayCfgsMap, _, err = ha.GetAllRelayConfigBeforeV610(cli)
+		relayCfgsMap, _, err = ha.GetAllRelayConfigBeforeV620(cli)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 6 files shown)


lichunzhu:
@WizardXiao PTAL

lance6716 (approved):

WizardXiao (approved):

dm/dm/ctl/master/config.go
```diff
@@ -331,6 +331,22 @@ func getSourceCfgs(cli *clientv3.Client) (map[string]*config.SourceConfig, error
...
+	if len(relayCfgsMap) == 0 {
+		relayCfgsMap, _, err = ha.GetAllRelayConfigBeforeV620(cli)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return relayCfgsMap, nil
+}
+
 func getAllCfgs(cli *clientv3.Client) (map[string]*config.SourceConfig, map[string]map[string]config.SubTaskConfig, map[string]map[string]struct{}, error) {
```
WizardXiao, line 350: just a question. `getAllCfgs` may be just used to export config, here change to support read from old keys is because we support export config when rolling update and there are both new and old version?

lichunzhu: Similar to `getSourceCfgs`. To support dumping relay configs for earlier dm clusters. We only try to dump old keys when there is no relay configs in new dm cluster.

Issue #1386, opened 2021-02-01
Owner OOM when upstream TiDB has too many tables


liuzix:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
- Prepare a TiDB cluster with 40k+ tables, create 8 changefeeds, each of which only replicates a few tables (< 10).

2. What did you expect to see?
- That replication goes smoothly, and that memory consumption is not affected by the number of unreplicated tables.

3. What did you see instead?
- Owner runs out of memory.

4. Versions of the cluster

    - TiCDC version (execute `cdc version`):

        ```
        5.0.0-rc
        ```

amyangfei:
Design document (internal only): [Design: TiCDC Optimization for Large Number of Unreplicated Tables](https://docs.google.com/document/d/1OZdMtDsch7MKccq5sTnHl8_KoLg7E6zeLsB41DFCqE0/edit)

lonng:
I think this issue should not be marked as `severity/critical` because support a large number of tables is not our design goal.

This issue should be marked as a feature request.

@IANTHEREAL  @amyangfei @overvenus How do you think?

IANTHEREAL:
Let's define how many tables are considered to be large number of tables

overvenus:
For now, OOM only occurs when upstream TiDB 1) has many tables and DDL frequently, it's not the case for most users, change to severity/moderate.

overvenus:
In #4068, we found that truncating tables also triggers the issue.

Reproduce steps:

```sh
mysql -h $HOST -P $PORT -u root -E -e 'create database if not exists truncate500;'

for i in $(seq 1 500); do {
    mysql -h $HOST -P $PORT -u root -E -e "
CREATE TABLE if not exists truncate500.orders_$i (
  o_id int(11) NOT NULL,
  o_d_id int(11) NOT NULL,
  o_w_id int(11) NOT NULL,
  o_c_id int(11) DEFAULT NULL,
  o_entry_d datetime DEFAULT NULL,
  o_carrier_id int(11) DEFAULT NULL,
  o_ol_cnt int(11) DEFAULT NULL,
  o_all_local int(11) DEFAULT NULL,
  PRIMARY KEY (o_w_id,o_d_id,o_id),
  KEY idx_order (o_w_id,o_d_id,o_c_id,o_id)
[...]

PR #5144, opened 2022-04-11
schema(cdc): refactor schema snapshot to reduce memory usage

hicqu:

Issue Number: close https://github.com/pingcap/tiflow/issues/1386

Ticdc keeps all versions of schema snapshot for each DDL, and it can only GC them after its checkpoint is advanced. If there are too many DDLs, the checkpoint advancing can be very slow so that all versions of schema snapshot are kept in memory.

Without the PR schema snapshots are deep cloned from the previous one. With the patch all schema snapshots share one MVCC tree to avoid deep clone. This is why memory usage is reduced in the PR.

### Check List 

Tests 

 - Unit test

Create 12K tables WITH THE PATCH:
图片

Create 12K tables WITHOUT THE PATCH:
图片

Code changes

Side effects

Related changes

### Release note 

```release-note
None
```


go.mod
```diff
@@ -95,6 +95,8 @@ require (
 	upper.io/db.v3 v3.7.1+incompatible
 )
 
+require github.com/ngaut/log v0.0.0-20180314031856-b8e36e7ba5ac
```
CharlesCheung96, line 98: Why this dependency needs to be listed separately?

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1155 @@
...
+		return cerror.ErrSnapshotSchemaExists.GenWithStackByArgs(dbInfo.Name.O, id)
+	}
+	s.doCreateSchema(dbInfo, currentTs)
+	s.currentTs = currentTs
+	log.Debug("create schema success", zap.String("name", dbInfo.Name.O), zap.Int64("id", dbInfo.ID))
+	return nil
+}
+
+// Replace a schema. dbInfo will be deep copied.
+// Callers should ensure `dbInfo` information not confict with other schemas.
```
asddongmen, line 499: `confict` seems a typo.

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1155 @@
...
+			}
+			tableDroped = true
+		}
+		return true
+	})
+	for _, vid := range tables {
+		x := s.tables.Delete(vid).(versionedID)
+		info := (*model.TableInfo)(x.target)
+		if info != nil {
+			isEligible := info.IsEligible(s.forceReplicate)
```
asddongmen, line 765: `ineligible := !info.IsEligible(s.forceReplicate)` may be more easy to understand.

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1155 @@
...
+			}
+			partitionDroped = true
+		}
+		return true
+	})
+	for _, vid := range partitions {
+		x := s.partitions.Delete(vid).(versionedID)
+		info := (*model.TableInfo)(x.target)
+		if info != nil {
+			isEligible := info.IsEligible(s.forceReplicate)
```
asddongmen, line 796: ditto

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1155 @@
...
+				target: unsafe.Pointer(tableInfo),
+			})
+			snap.tableNameToID.ReplaceOrInsert(versionedEntityName{
+				prefix: dbinfo.ID,
+				entity: tableInfo.Name.O,
+				tag:    tag,
+				target: tableInfo.ID,
+			})
+
+			isEligible := tableInfo.IsEligible(forceReplicate)
```
asddongmen, line 197: ditto

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1155 @@
...
+	var target int64 = -1
+	return versionedEntityName{prefix, entity, tag, target}
+}
+
+func newVersionedID(id int64, tag uint64) versionedID {
+	var target unsafe.Pointer = unsafe.Pointer(nil)
+	return versionedID{id, tag, target}
+}
+
+// DoHandleDDL is like HandleDDL but doesn't fill schema name into job. It's only for tests.
```
asddongmen, line 1021: `It's only for tests.` is confused, seems it also be used in HandleDDL.

asddongmen (approved):

cdc/owner/schema.go
```diff
@@ -74,36 +74,30 @@ func (s *schemaWrap4Owner) AllPhysicalTables() []model.TableID {
 	if s.allPhysicalTablesCache != nil {
 		return s.allPhysicalTablesCache
 	}
-	tables := s.schemaSnapshot.Tables()
-	s.allPhysicalTablesCache = make([]model.TableID, 0, len(tables))
-	for _, tblInfo := range tables {
+	s.allPhysicalTablesCache = make([]model.TableID, 0)
```
overvenus, line 77: Could you add some comment about why pre-allocate is not needed here.?

cdc/owner/schema.go
```diff
@@ -74,36 +74,30 @@ func (s *schemaWrap4Owner) AllPhysicalTables() []model.TableID {
...
 }
 
 // AllTableNames returns the table names of all tables that are being replicated.
 func (s *schemaWrap4Owner) AllTableNames() []model.TableName {
-	tables := s.schemaSnapshot.Tables()
-	names := make([]model.TableName, 0, len(tables))
-	for _, tblInfo := range tables {
-		if s.shouldIgnoreTable(tblInfo) {
-			continue
+	names := make([]model.TableName, 0)
```
overvenus, line 95:
```suggestion
	names := make([]model.TableName, 0, len(s.allPhysicalTablesCache))
```

cdc/entry/schema_storage_test.go
```diff
@@ -743,58 +768,16 @@ func TestCreateSnapFromMeta(t *testing.T) {
...
+	tableID, ok := snap.TableIDByName("test2", "simple_test5")
 	require.True(t, ok)
 	require.True(t, snap.IsIneligibleTableID(tableID))
 	dbInfo, ok := snap.SchemaByTableID(tableID)
 	require.True(t, ok)
 	require.Equal(t, dbInfo.Name.O, "test2")
-	require.Len(t, snap.tableInSchema, 3)
-}
-
-func TestSnapshotClone(t *testing.T) {
```
overvenus, line 759: Why is it removed?

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1155 @@
...
+		err = s.createTable(tbInfo, currentTs)
+		if err != nil {
+			return errors.Trace(err)
+		}
+	}
+	return nil
+}
+
+// HandleDDL handles the given job.
+func (s *Snapshot) HandleDDL(job *timodel.Job) error {
```
overvenus, line 708: Is is thread-safe?

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1155 @@
...
+	if meta == nil {
+		snap := NewEmptySnapshot(forceReplicate)
+		snap.currentTs = currentTs
+		return snap, nil
+	}
+	return NewSnapshotFromMeta(meta, currentTs, forceReplicate)
+}
+
+// Snapshot stores the source TiDB all schema information and it should be READ-ONLY!
+type Snapshot struct {
```
overvenus, line 83: Is is thread-safe?

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1155 @@
...
+package schema
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+	"unsafe"
+
+	"github.com/google/btree"
+	"github.com/ngaut/log"
```
overvenus, line 23: Please do not use this package.

cdc/entry/schema/snapshot.go
hicqu: Yes all public functions are thread-safe. I will comment it on `Snapshot`.

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1160 @@
...
+	// Partitions and tables share ineligibleTables because their IDs won't conflict.
+	ineligibleTables *btree.BTree
+
+	// if forceReplicate is true, treat ineligible tables as eligible.
+	forceReplicate bool
+
+	currentTs uint64
+
+	rwlock *sync.RWMutex
+	locked int
```
overvenus, line 117: Can we remove `locked`?

overvenus:
Could you describe the difference of schemastaorge between this PR and master?

nongfushanquan (commented):
could you add some comments to explain the method reducing memory usage

cdc/entry/schema/snapshot.go
hicqu: `locked` forbids a thread to try lock multiple times.

hicqu: You are right. The design is wrong.

hicqu:
@overvenus @nongfushanquan I have updated the PR description. Please take a look again, thanks.

cdc/entry/schema/snapshot.go
liuzix: Recursive locking is quite tricky in Go, because there is no "goroutine-local" storage.

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1145 @@
...
+}
+
+// SchemaIDByName gets the schema id from the given schema name.
+func (s *Snapshot) SchemaIDByName(schema string) (int64, bool) {
+	s.rwlock.RLock()
+	defer s.rwlock.RUnlock()
+	return s.doSchemaIDByName(schema)
+}
+
+func (s *Snapshot) doSchemaIDByName(schema string) (id int64, ok bool) {
```
overvenus, line 311:
```suggestion
func (s *Snapshot) schemaIDByNameLocked(schema string) (id int64, ok bool) {
```

I think it's more clear when the method should be invoked.

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1145 @@
...
+}
+
+// NewSnapshotFromMeta creates a schema snapshot from meta.
+func NewSnapshotFromMeta(meta *timeta.Meta, currentTs uint64, forceReplicate bool) (*Snapshot, error) {
+	snap := NewEmptySnapshot(forceReplicate)
+	dbinfos, err := meta.ListDatabases()
+	if err != nil {
+		return nil, cerror.WrapError(cerror.ErrMetaListDatabases, err)
+	}
+	tag := negative(currentTs)
```
overvenus, line 142: Please add some comment about the `tag` and the `negative`.

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1145 @@
...
+
+	for _, dbinfo := range dbinfos {
+		vid := newVersionedID(dbinfo.ID, tag)
+		vid.target = unsafe.Pointer(dbinfo)
+		snap.schemas.ReplaceOrInsert(vid)
+
+		vname := newVersionedEntityName(-1, dbinfo.Name.O, tag)
```
overvenus, line 149:
```suggestion
		vname := newVersionedEntityName(-1, dbinfo.Name.O, tag)
```
Please add some comment about the magic number `-1`.

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1145 @@
...
+	// the associated entity id, negative values are treated as invalid.
+	target int64
+}
+
+// ID with finish timestamp of the associated DDL job.
+type versionedID struct {
+	id  int64
+	tag uint64
+	// the associated entity pointer.
+	target unsafe.Pointer
```
overvenus, line 985: Why does it needs to be an `unsafe.Pointer`?

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1145 @@
...
+		zap.String("schema", tbInfo.TableName.Schema),
+		zap.String("table", tbInfo.TableName.Table),
+		zap.Int64("id", tbInfo.ID))
+	return nil
+}
+
+func (s *Snapshot) doDropTable(tbInfo *model.TableInfo, currentTs uint64) {
+	tag := negative(currentTs)
+	s.tables.ReplaceOrInsert(newVersionedID(tbInfo.ID, tag))
+	s.tableNameToID.ReplaceOrInsert(newVersionedEntityName(tbInfo.SchemaID, tbInfo.TableName.Table, tag))
```
overvenus, line 525: The name `Snapshot` suggests it's read-only, but this method mutates `Snapshot`, should we change the name?

cdc/entry/schema/snapshot.go
hicqu: Because the target can be `*model.DBInfo` or `*model.TableInfo`.

cdc/entry/schema/snapshot.go
hicqu: I'm not sure so I prefer to keep less changes in this PR.

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1151 @@
...
+}
+
+func (s *Snapshot) isIneligibleTableIDLocked(id int64) (ok bool) {
+	tag, ok := s.tableTagByID(id, false)
+	return ok && s.ineligibleTables.Get(newVersionedID(id, tag)) != nil
+}
+
+func (s *Snapshot) tableTagByID(id int64, nilAcceptable bool) (foundTag uint64, ok bool) {
+	s.rwlock.RLock()
+	defer s.rwlock.RUnlock()
```
overvenus, line 398: I'm afraid mixing locked API and unlocked API can easily causing dead lock.

```go
func (s *Snapshot) IsTruncateTableID(id int64) bool {
	s.rwlock.RLock()
	defer s.rwlock.RUnlock()
	tag, ok := s.tableTagByID(id, true)
}
...

func (s *Snapshot) tableTagByID(id int64, nilAcceptable bool) (foundTag uint64, ok bool) {
	s.rwlock.RLock()
	defer s.rwlock.RUnlock()
```

----

 Maybe we can separate thread-safe API and thread-unsafe API into two structs.

```go
// All thread-safe APIs go here
type Snapshot struct {
  mu sync.Mutex
  s snapshot
}

// -----

// All thread-unsafe APIs go here.
type snapshot struct {
}
```

cdc/entry/schema/snapshot.go
overvenus: Can it be interface{}?

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1151 @@
...
+}
+
+// Entity(schema or table) name with finish timestamp of the associated DDL job.
+type versionedEntityName struct {
+	prefix int64 // schema ID if the entity is a table, or -1 if it's a schema.
+	entity string
+	// tag is `uint64.MAX - finish_timestamp`.
+	tag uint64
+	// the associated entity id, negative values are treated as invalid.
+	target int64
```
liuzix, line 978: Can we call this field `id`?

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1151 @@
...
+}
+
+// ID with finish timestamp of the associated DDL job.
+type versionedID struct {
+	id  int64
+	tag uint64
```
liuzix, line 984: Could you add a comment for the definition of `tag`?

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1151 @@
...
+}
+
+func (v1 versionedID) Less(than btree.Item) bool {
+	v2 := than.(versionedID)
+	return v1.id < v2.id || (v1.id == v2.id && v1.tag < v2.tag)
+}
+
+// negative transforms `x` for reverse sorting based on it.
+func negative(x uint64) uint64 {
+	return 0xFFFFFFFFFFFFFFFF - x
```
liuzix, line 1001: Can we use `math.MaxUint64`?

cdc/entry/schema/snapshot.go
```diff
@@ -0,0 +1,1151 @@
...
+			}
+		}
+	}
+
+	snap.currentTs = currentTs
+	return snap, nil
+}
+
+// Copy creates a new schema snapshot based on the given one.
+func (s *Snapshot) Copy() *Snapshot {
```
liuzix, line 194: This method seems to be shallow copying all the composite data structures and this seems to be the central idea that saves memory. Can you add more comments on the behavior of `Copy()`?

liuzix (approved):

hicqu:
@overvenus please take a look again and /merge it if you are free.

PR #5161, opened 2022-04-13
pkg/p2p: make p2p client thread-safe

### LINKED ISSUE #5162: pkg/p2p: MessageClient is not thread-safe when writing to the same topic
**NOTE: This issue is not exposed in the production code of TiCDC, so it is NOT A BUG**

The `MessageClient` in `pkg/p2p` cannot be called by multiple goroutines to write to the same p2p topic. Otherwise, unexpected results such as panicking will occur. We need to address this issue to make the `pkg/p2p` package more usable outside of current usages.

liuzix:

Issue Number: close #5162

- Made `MessageClient` safe to be used by multiple goroutines to write to the **same** p2p topic (writing to different topics has always been thread-safe).

```
BenchmarkSendChanSendAsyncSPSC-24     	  420609	      3634 ns/op
BenchmarkSendChanSendSyncSPSC-24      	 1041715	      1177 ns/op
BenchmarkSendChanSendAsyncMPSC8-24    	   32517	     36439 ns/op
BenchmarkSendChanSendSyncMPSC8-24     	  188133	      6105 ns/op
```
**Note that the `MPSC8` cases actually send `b.N * 8` messages.**
Performance benchmark (indicating the best p2p performance limited by the client's implementation). Note that currently TiCDC's production code only covers the case `BenchmarkSendChanSendAsyncSPSC`. This means a throughput of `275330` messages per second, which is more than enough.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


pkg/p2p/internal/send_chan.go
```diff
@@ -0,0 +1,170 @@
...
+// SendAsync when CPU is saturated, so please refrain from
+// using it unless necessary or for testing.
+func (c *SendChan) SendSync(
+	ctx context.Context,
+	topic string,
+	value []byte,
+	closeCh <-chan struct{},
+	nextSeq func() int64,
+) (int64, error) {
+	ticker := time.NewTicker(10 * time.Millisecond)
```
lance6716, line 70: make 10ms configurable? Benchmark performance is related to it.

pkg/p2p/internal/send_chan_test.go
```diff
@@ -0,0 +1,313 @@
+// Copyright 2021 PingCAP, Inc.
```
lance6716, line 1:
```suggestion
// Copyright 2022 PingCAP, Inc.
```

lance6716 (commented):
left my comments, switch to a hotfix review

pkg/p2p/internal/send_chan.go
```diff
@@ -0,0 +1,170 @@
...
+	default:
+	}
+
+	return true, seq
+}
+
+// Receive receives one message from the channel.
+// If a message is received from `tick`, the function will return
+// (nil, false, nil).
+func (c *SendChan) Receive(ctx context.Context, tick <-chan time.Time) (*proto.MessageEntry, bool, error) {
```
lance6716, line 124: maybe this tick can be replaced by a configurable tick interval, same with `SendSync`?

lance6716 (commented):
lgtm for current code, I'll take a look why old code is not thread-safe. I'll be grateful if you could explain it here.

liuzix:
> lgtm for current code, I'll take a look why old code is not thread-safe. I'll be grateful if you could explain it here.

The old code is not thread-safe because if you use a Go `chan`, there will be no way to increment the `seq` and enqueue the event as an atomic operation.

pkg/p2p/internal/send_chan.go
liuzix: This `tick` is passed from the caller, and is controller by the caller's configuration.

pkg/p2p/internal/send_chan.go
liuzix: `SendSync` is a marginal use case. Not sure if it is worth adding a config?

lance6716: DM in dataflow engine mostly use blocking sending

https://github.com/hanfei1991/microcosm/blob/3ed20d9ee8677fb1d52af5fc3c2c15c00c19c144/pkg/dm/message_pair.go#L52

but we can add the configuration when problem really happens.

lance6716 (approved):

pkg/p2p/internal/send_chan.go
```diff
@@ -0,0 +1,170 @@
...
+	recvIdx int64
+
+	// hasNewMsg is a buffered channel that the sending goroutine(s)
+	// use to notify the receiver end.
+	hasNewMsg chan struct{}
+
+	// notifyChan is an unbuffered channel that
+	// SendSync writes to in order to determine whether the
+	// consumer is ready.
+	notifyChan chan struct{}
```
amyangfei, line 44: Where do we write to this channel?

liuzix: It was a typo. Thank you for pointing this out. Fixing it has actually improved the performance of `SendSync` more than `1000` times.

amyangfei (approved):

PR #5165, opened 2022-04-13
tracker(dm): fix tracker panic when pk of downstream table orders behind

### LINKED ISSUE #5159: dm: tracker panic when pk of downstream table orders behind
### What did you do?

to reproduce
- downstream create table before sync or the table has no pk on create then add multiple pk later.
- tracker will panic when downstream table ddl has, but even if the ddl doesn't satisfy all condition, dm may use the wrong key:
  - multiple normal index before pk
  - multiple uk/pk
  - number of normal index >= number of uk/pk

such as downstream create table below before sync:
```sql
create table t(
    a int not null,
    b int not null,
    c int not null,
    d int not null,
    e int not null,
    index (a),
    index(b),
    index(c),
    unique(d),
    unique(e)
);
```

or add pk after create table in upstream:
```sql
create database test;
use test;
create table t(
    a int not null,
    b int not null,
    c int not null,
    d int not null,
    e int not null,
    index (a),
    index(b),
    index(c)
);

alter table t add unique(d);
alter table t add unique(e);
show create table t;
insert into t values(1,1,1,1,1);
```

code panics since absoluteUKPosition  comes from all indices but index into uk/pk arr:
https://github.com/pingcap/tiflow/blob/2605a84c09587f57f87cfc088e66a769033dcad5/dm/pkg/schema/tracker.go#L569

```log
"Message": "panic error: runtime error: index out of range [3] with length 2",
```

works

### What did you see instead?

panic

### Versions of the cluster

occurs in 5.3, reproduced on 5.4/6.0

```console
```

D3Hunter:

Issue Number: close #5159

### Check List 

Tests 

 - Unit test
 - [x] Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
fix tracker panic when pk of downstream table orders behind
```


lance6716 (approved):

Ehco1996 (approved):

GMHDBJD (approved):
LGTM

PR #5176, opened 2022-04-14
syncer(dm): cache some metrics looking up in hot path

lance6716:
Signed-off-by: lance6716 <lance6716@gmail.com>

### What problem does this PR solve?

Issue Number: ref #5183 

### What is changed and how it works?

- Syncer holds a metrics cache, so some of metrics looking up can be omitted
- Only check checkpoint regression at debug level
- No need to add timeout when GetEvent, it's just reading a channel
- only update location recorder at debug level

### Check List 

Tests 

 - pass original tests
 - **waiting for benchbot test**

Code changes

 - Has exported variable/fields change

Related changes

 - Need to cherry-pick to the release branch

### Release note 

```release-note
Improve syncer's performance
```

PR #5181, opened 2022-04-14
validator(dm): query-status & prometheus monitoring

### LINKED ISSUE #5152: prometheus monitoring metric for validation, grafana dashboard and alerting

### LINKED ISSUE #5164: dmctl query-status command, returns validator status

buchuitoudegou:

Issue Number: close #5152 
close #5164 

- add metric for validator
- return brief info of validator when querying worker's status

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`
```


Ehco1996 (commented):
restLGTM

dm/syncer/data_validator.go
```diff
@@ -489,13 +495,38 @@ func (v *DataValidator) doValidate() {
...
+		v.syncer.currentLocationMu.RLock()
+		syncerLoc := v.syncer.currentLocationMu.currentLocation
+		v.syncer.currentLocationMu.RUnlock()
+		if syncerLoc.Position.Name == currLoc.Position.Name {
+			// same file: record the log pos latency
+			metrics.ValidatorLogPosLatency.WithLabelValues(v.cfg.Name, v.cfg.SourceID, v.cfg.WorkerName).Set(float64(syncerLoc.Position.Pos - currLoc.Position.Pos))
+			metrics.ValidatorLogFileLatency.WithLabelValues(v.cfg.Name, v.cfg.SourceID, v.cfg.WorkerName).Set(float64(0))
+		} else {
+			var syncerLogIdx int64
+			metrics.ValidatorLogPosLatency.WithLabelValues(v.cfg.Name, v.cfg.SourceID, v.cfg.WorkerName).Set(float64(0))
```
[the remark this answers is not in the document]
buchuitoudegou: Did you mean rename `binlogPosLatency`? Current:
1. validator and syncer in the same file: `binlogFileLatency` -> 0, `binlogPosLatency` = `syncer offset` - `validator offset`.
2. validator and syncer not in the same file: `binlogPosLatency` -> 0, `binlogFileLatency` = `sycer file` - `validator file`.

It does looks abnormal that `binlogPosLatency` is 0 at the beginning but increases drastically after validator reaches the same file with syncer.

dm/syncer/data_validator.go
```diff
@@ -489,13 +495,38 @@ func (v *DataValidator) doValidate() {
 		default:
 			currLoc.Position.Pos = e.Header.LogPos
 		}
-
+		// update validator metric
+		metrics.ValidatorBinlogPos.WithLabelValues(v.cfg.Name, v.cfg.SourceID, v.cfg.WorkerName).Set(float64(currLoc.Position.Pos))
```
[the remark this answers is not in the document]
buchuitoudegou: Would you give me some hints about asynchronously updating (e.g. put them into an independent go routine)?

dm/syncer/data_validator.go
```diff
@@ -935,3 +968,31 @@ func (v *DataValidator) getTableStatus() map[string]*tableValidateStatus {
 	}
 	return tblStatus
 }
+
+func (v *DataValidator) GetBriefValidatorStatus() *pb.BriefValidationStatus {
+	v.RLock()
```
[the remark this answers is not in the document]
buchuitoudegou: 1. stop the validator
2. do this in parallel:
    - query-status (config is being read)
    - start validator and change mode (validator config is being written)

dm/syncer/metrics/validator_metrics.go
```diff
@@ -0,0 +1,77 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
```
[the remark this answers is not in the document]
buchuitoudegou: What dashboard did you refer to?

dm/syncer/data_validator.go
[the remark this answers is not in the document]
Ehco1996: i took another look, i found we have the two metrrics `ValidatorBinlogPos` and `ValidatorBinlogFile` 
so user can calcuate the latency by thier own as we also have `BinlogFileGauge ` `BinlogPosGauge ` in [syncer](https://github.com/pingcap/tiflow/blob/536e18274b90934d7611caca25e40f015e1c163a/dm/syncer/metrics/metrics.go#L132)

so maybe just delete `ValidatorLogPosLatency` and `ValidatorLogFileLatency` ?

dm/syncer/metrics/validator_metrics.go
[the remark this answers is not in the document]
Ehco1996: i mean grafana dashboard in [here](https://github.com/pingcap/tiflow/tree/master/dm/metrics/grafana)

dm/syncer/data_validator.go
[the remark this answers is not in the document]
Ehco1996: just a litte suggention ok to not change
maybe add new job type `metrics` and use a new worker thread to update metric like we handle `dmljob`

> What about making them asynchronous as a whole 

lgtm ptal @lance6716

dm/syncer/metrics/validator_metrics.go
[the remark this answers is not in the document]
buchuitoudegou: Idk we have this configuration file😢

Will try to fix it.

dm/syncer/data_validator.go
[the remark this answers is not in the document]
buchuitoudegou: I've discussed it with @D3Hunter. Both latencies are helping users identifying "how far the validator fall behind the syncer". Originally, we use `ValidatorBinlogGauge` (just like the syncer) to imply this message but it seems a bit counterintuitive to let users calculate the difference when using a monitoring system. /cc @sunzhaoyang

dm/syncer/data_validator.go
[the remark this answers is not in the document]
lance6716: https://github.com/pingcap/tiflow/pull/5176 the problem is not related to asynchronous.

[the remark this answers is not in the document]
Ehco1996: >  Metrics caches the metrics variables to avoid looking up by name in metricsproxy each time.

oooo, the hot path is `WithLabelValues` , a cached var is enough and no need do this in async

dm/dm/proto/dmworker.proto
```diff
@@ -478,3 +479,16 @@ enum ValidationErrOp {
...
+message ValidationStats {
+    uint64 processedCount = 1;
+    uint64 succeededCount = 2;
+    uint64 errorCount = 3;
+}
+
+message BriefValidationStatus {
+    string mode = 1;
+    string lastCheckTime = 2;
+    string status = 3;
```
lance6716, line 492: seems this is the `Stage` type.

And I don't know what's the difference of "status", "stage" :joy:

dm/dm/worker/subtask_test.go
```diff
@@ -577,3 +577,50 @@ func (t *testSubTask) TestOperateValidatorError(c *C) {
 	st.StartValidator(pb.Stage_Running, false)
 	c.Assert(st.OperateValidatorError(pb.ValidationErrOp_ClearErrOp, 0, true), IsNil)
 }
+
+func (t *testSubTask) TestGetBriefValidatorStatus(c *C) {
```
lance6716, line 581: For new unit tests you can write it in testify framework

dm/syncer/data_validator.go
```diff
@@ -171,10 +173,16 @@ func NewContinuousDataValidator(cfg *config.SubTaskConfig, syncerObj *Syncer, st
 
 	v.workerCnt = cfg.ValidatorCfg.WorkerCount
 	v.processedRowCounts = make([]atomic.Int64, rowChangeTypeCount)
-	v.workers = make([]*validateWorker, v.workerCnt)
+	v.pendingRowCounts = make([]atomic.Int64, rowChangeTypeCount)
+	v.errorRowCounts = make([]atomic.Int64, errorStateTypeCount)
+	// no pre-allocated space for workers
+	// because checkpoint would try visiting all the workers to persist them, though they are nil
+	// set length to zero to prevent nil pointer panic
+	v.workers = make([]*validateWorker, 0)
```
lance6716, line 181: Where does the checkpoint access happen? `v.workers` is not protected by a lock, maybe there's a R/W race.

And I guess you want to avoid it iterated by `for`, the zero value of slice (nil) is OK, no need to set it to an empty slice.

dm/syncer/data_validator.go
```diff
@@ -935,3 +968,31 @@ func (v *DataValidator) getTableStatus() map[string]*tableValidateStatus {
...
+		ret.Stats.ProcessedCount += uint64(v.processedRowCounts[i].Load())
+		pendingRowCnt += int(v.pendingRowCounts[i].Load())
+	}
+	// currently, errRowCounts has same length with processedRowCounts
+	// to avoid potential careless changes (in the future)
+	// they are iterated separately.
+	for i := 0; i < len(v.errorRowCounts); i++ {
+		ret.Stats.ErrorCount += uint64(v.errorRowCounts[i].Load())
+	}
+	ret.Stats.SucceededCount = ret.Stats.ProcessedCount - ret.Stats.ErrorCount - uint64(pendingRowCnt)
```
lance6716, line 996: seems `loadPersistedData` will load `errorRowCounts`, but other variables will not be loaded. So this calculation may get negative value.

dm/syncer/data_validator.go
buchuitoudegou: > Where does the checkpoint access happen? v.workers is not protected by a lock, maybe there's a R/W race.

https://github.com/pingcap/tiflow/blob/4d6f44dfeadc0c95a29a23b271fe0669d2dbc7a1/dm/syncer/validator_checkpoint.go#L213-L306

There is no data race currently because the method called in `validator_checkpoint` is thread-safe and `workers` won't be accessed concurrently by the checkpoint and the validator. 

The `checkpoint.persist` function will be invoked before the validator starts, in order to fetch persisted state from the database. The purpose of setting the empty slice is to prevent the checkpoint visiting nil worker and calling its method (panic). Otherwise, I have to check if the worker is nil every time before I call worker's methods.

Are there any side-effects if I put off allocating the workers array?

lance6716: you can try delete these four lines

dm/dm/proto/dmworker.proto
buchuitoudegou: They are definitely the same thing. I just followed the naming in the spec...Should I replace it with "Stage" (I bet @sunzhaoyang won't complain of it)?

dm/syncer/data_validator.go
buchuitoudegou: Indeed. `pedingRowCounts` can be recovered in `loadPersistedData`. What about adding a new checkpoint to record `processedCount`:
```sql
CREATE TABLE task_name(
id BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY,
source VARCHAR(32),
row_change_type INT,
processed_count INT,
UNIQUE KEY uk_validator_process(source, row_change_type)
)
```
/cc @D3Hunter

dm/dm/proto/dmworker.proto
sunzhaoyang: As my option.

Stage means different phases of a process, each phase finish different jobs, e.g. ‘dump/load/sync’ 

Status means different situation of the same object, e.g. ‘Running/Pause/Stop’

But  I think we had mixed them in the previous  features, so following the tradition is ok too. The reviewer can decide it.

D3Hunter:
since the next release 6.1 is LTS version, and this pr will add fields to result of `query-status`, so this pr should better be merged after 6.1, will review it later.

buchuitoudegou:
ptal @D3Hunter @Ehco1996

dm/metrics/grafana/DM-Monitor-Professional.json
```diff
@@ -1,45 +1,9 @@
...
-      "id": "timeseries",
-      "name": "Time series",
-      "version": ""
-    }
-  ],
   "annotations": {
     "list": [
       {
         "builtIn": 1,
-        "datasource": "${DS_TEST-CLUSTER}",
```
D3Hunter, line 42: not sure whether we should change this placeholder which may used on cloud or tiup monitoring which has cluster name.

also this file change is so large that it's very hard to review what's changed. @lance6716 how is this file changed in previous pr?

dm/dm/proto/dmworker.proto
```diff
@@ -202,6 +202,19 @@ message SubTaskStatus {
...
+message ValidationBriefStatus {
+    string mode = 1;
+    Stage status = 2;
+    ValidationStats stats = 3;
+}
+
+message ValidationStats {
+    string ProcessedRows = 1;
+    string PendingRows = 2;
+    string ErrorRows = 3;
```
D3Hunter, line 217: remove those and use ValidationStatus instead

dm/CHANGELOG.md
```diff
@@ -224,7 +224,7 @@ As of DM v2.0, DM uses a TiDB-independent version number; after DM v2.0, DM uses
 
 ### Bug fixes
 
-- Fix the issue that restarting a task might cause `fail to initial unit Sync of subtask` error [#1274](https://github.com/pingcap/dm/pull/1274)
+- Fix the issue that restarting a task might cause `fail to initialize unit Sync of subtask` error [#1274](https://github.com/pingcap/dm/pull/1274)
```
D3Hunter, line 227: maybe fix those typo in another pr

dm/metrics/grafana/DM-Monitor-Professional.json
buchuitoudegou: I changed it by mistake and I'm looking for a solution now.

dm/dm/proto/dmworker.proto
buchuitoudegou: It seems ValidationStatus includes some superfluous fields, which will be printed as a result of `query-status` (but it does make code simpler, I will discuss with PM later)

dm/CHANGELOG.md
buchuitoudegou: Will revert it later

dm/dm/proto/dmworker.proto
D3Hunter: better have a sample output so we can compare

dm/metrics/grafana/DM-Monitor-Professional.json
lance6716: ehco's work, this file is related to many CI CD, should talk with them first.

buchuitoudegou: I've asked @Ehco1996 for help. This file can be exported from the grafana server, where, however, the `datasource` should be `${DS_TEST-CLUSTER}`, a placeholder, rather than a specific cluster name.

I'll fix it in the next commit.

D3Hunter: can this file be smaller? hard to review

buchuitoudegou: @Ehco1996

buchuitoudegou:
ptal @D3Hunter @Ehco1996

dm/syncer/data_validator.go
```diff
@@ -628,15 +629,39 @@ func (v *DataValidator) doValidate() {
...
+		}
 		// wait until syncer synced current event
 		err = v.waitSyncerSynced(currLoc)
 		if err != nil {
 			// no need to wrap it in error_list, since err can be context.Canceled only.
 			v.sendError(err)
 			return
 		}
-
+		v.syncer.currentLocationMu.RLock()
```
D3Hunter, line 647: move all those metrics stuff(previous part too) into a separate method to avoid messing the main logic up.

dm/syncer/metrics/validator_metrics.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/pingcap/tiflow/dm/pkg/metricsproxy"
+)
+
+var (
+	ValidatorErrorCount = metricsproxy.NewCounterVec(
```
D3Hunter, line 23: error row change maybe re-find on recover from checkpoint. and using counter is hard to add alerts since we may not find error row change uniformly in time.

maybe use a gauge, and update it on validator status routine.

dm/tests/dmctl_command/run.sh
```diff
@@ -467,6 +467,12 @@ function run_validator_cmd {
 		"new\/ignored\/resolved: 4\/0\/0" 1 \
 		"\"stage\": \"Running\"" 4 \
 		"\"stage\": \"Stopped\"" 1
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
```
D3Hunter, line 470: add a case when validator is not enabled

D3Hunter (commented):
[DM-Monitor-Professional.json](https://github.com/pingcap/tiflow/pull/5181/files#diff-2a3de2d83ed6e4fe3ea8fbf4549eae5d015d26910e7815d5f02c9e82618859be) is not reviewed

dm/metrics/grafana/DM-Monitor-Professional.json
```diff
@@ -51,176 +51,17 @@
...
-        },
-        "tooltipOptions": {
-          "mode": "single"
-        }
-      },
-      "pluginVersion": "7.5.7",
-      "targets": [
-        {
-          "exemplar": true,
-          "expr": "dm_worker_task_state{task=~\"$task\",source_id=~\"$source\"}",
```
Ehco1996, line 203: why this is deleted?

dm/metrics/grafana/DM-Monitor-Professional.json
```diff
@@ -266,7 +107,7 @@
             "alertThreshold": true
           },
           "percentage": false,
-          "pluginVersion": "7.5.7",
+          "pluginVersion": "7.5.11",
```
Ehco1996, line 110: Not sure if this update is working properly?

Ehco1996 (commented):
maybe add a link to

dm/tests/dmctl_command/run.sh
buchuitoudegou: [1565730](https://github.com/pingcap/tiflow/pull/5181/commits/1565730b6e4d7f9a64cbbbea5153bc8ca8ad3078)

buchuitoudegou:
/run-all-tests
 ptal @D3Hunter @Ehco1996

Ehco1996 (approved):
LGTM nicework

buchuitoudegou:
ptal @D3Hunter

buchuitoudegou:
ptal @D3Hunter

dm/syncer/data_validator.go
```diff
@@ -387,7 +391,7 @@ func (v *DataValidator) printStatusRoutine() {
 			speed := float64((currProcessedBinlogSize-prevProcessedBinlogSize)>>20) / interval.Seconds()
 			prevProcessedBinlogSize = currProcessedBinlogSize
 			prevTime = currTime
-
+			v.vmetric.ValidatorErrorCount.Set(float64(v.newErrorRowCount.Load()))
```
D3Hunter, line 394: this fields only stores in-memory new errors, we should load error count using getAllErrorCount which maybe slow that's why we put it in this routine.

dm/syncer/metrics/validator_metrics.go
```diff
@@ -0,0 +1,95 @@
...
+	validatorBinlogPos.DeleteAllAboutLabels(prometheus.Labels{"task": task})
+	validatorBinlogFile.DeleteAllAboutLabels(prometheus.Labels{"task": task})
+}
+
+type ValidatorMetrics struct {
+	ValidatorErrorCount     prometheus.Gauge
+	ValidatorLogPosLatency  prometheus.Gauge
+	ValidatorLogFileLatency prometheus.Gauge
+	ValidatorBinlogFile     prometheus.Gauge
+	ValidatorBinlogPos      prometheus.Gauge
```
D3Hunter, line 84: seems we can remove Validator prefix

dm/syncer/data_validator.go
```diff
@@ -628,15 +665,16 @@ func (v *DataValidator) doValidate() {
 		default:
 			currLoc.Position.Pos = e.Header.LogPos
 		}
-
+		// update validator metric
+		v.updateValidatorBinlogMetric(currLoc)
```
D3Hunter, line 669: maybe we can make it near `v.updateValidatorBinlogLag(currLoc)`, move this part down or reversely

dm/tests/dmctl_command/run.sh
```diff
@@ -552,6 +566,16 @@ function run_validator_cmd {
 	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
 		"validation status test" \
 		"new\/ignored\/resolved: 0\/0\/0" 2
+	run_sql_source1 "insert into dmctl_command.t1 values(-10,'ignore-row')"
+	run_sql_source1 "create table dmctl_command.t_trigger_flush110(id int primary key)" # trigger flush
+	sleep 1                                                                             # wait sync
```
D3Hunter, line 571: we can wait syncer `Synced` using query-status, sleep is not stable, below sleep too, if it's stopped we can check it.

D3Hunter (commented):
rest lgtm

dm/syncer/data_validator.go
```diff
@@ -387,7 +391,12 @@ func (v *DataValidator) printStatusRoutine() {
 			speed := float64((currProcessedBinlogSize-prevProcessedBinlogSize)>>20) / interval.Seconds()
 			prevProcessedBinlogSize = currProcessedBinlogSize
 			prevTime = currTime
-
+			counts, err := v.getAllErrorCount(validatorStatusInterval)
```
D3Hunter, line 394: prefer use `validatorDmctlOpTimeout`, 1m is too long

dm/syncer/data_validator.go
```diff
@@ -387,7 +391,12 @@ func (v *DataValidator) printStatusRoutine() {
...
+			if err == nil {
+				v.vmetric.ErrorCount.Set(float64(counts[pb.ValidateErrorState_NewErr]))
+			} else {
+				v.L.Warn("validator status", zap.Error(err))
```
D3Hunter, line 398:
```suggestion
				v.L.Warn("failed to get error row count", zap.Error(err))
```

dm/tests/dmctl_command/run.sh
```diff
@@ -552,6 +566,18 @@ function run_validator_cmd {
...
 		"validation status test" \
 		"new\/ignored\/resolved: 0\/0\/0" 2
+	run_sql_source1 "insert into dmctl_command.t1 values(-10,'ignore-row')"
+	run_sql_source1 "create table dmctl_command.t_trigger_flush110(id int primary key)" # trigger flush
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"query-status test" \
+		"\"synced\": true" 2
+	run_sql_tidb "select * from dmctl_command.t1"
+	check_not_contains "id: -10" # sync failed due to GO_FAILPOINTS
+	sleep 5                      # wait validator
```
D3Hunter, line 576: can this be removed too?

dm/syncer/data_validator.go
buchuitoudegou: Will change it to 5s

dm/tests/dmctl_command/run.sh
buchuitoudegou: Yes. Because validator is stopped, we don't need to wait for it. It's somehow a placeholder (pretending to wait for the validator) just in case there is something wrong to make stopping the validator failed. Perhaps just leave it here.

D3Hunter: > make stopping the validator failed

then we can just check whether there's `stage: stopped` in `validation status`

D3Hunter (approved):

PR #4249, opened 2022-01-07
owner(cdc): asynchronously close the ddl sink

3AceShowHand:
### What problem does this PR solve?

Issue Number: close #4241, #4384

when closing the `ddlSink` with Kafka as the sink, it would block the `owner` or `processorManager` for a period of time, which had a negative impact on all running changefeeds.

* For Kafka, this would happen when all brokers were shut down.

### What is changed and how it works?

At the owner side:

* close the`ddlSink` asynchronsly
* for changefeed initialization, introduce `changeFeedRunningStatus`, at the moment for `changeFeedClosed`, `changeFeedRunning`, `changeFeedClosing`.
 
`changeFeedClose` -> `changeFeedRunning` if initialize the changefeed.

`changeFeedRunning` -> `changefeedClosing` if close the changefeed

`changeFeedClosing` -> `changeFeedClosed` if the downstream sink fully closed.

without introducing `changeFeedClosing`, a closing changefeed may be reinitializing at the same time, also may be closing multiple times.

At the `processorManager` side:

* initialize and close processor's `sinkManager` in an asynchronous way.
* Like above, introduce `processorRunningStatus`, at the moment for `processorClosed`, `processorInitializing`, `processorRunning`, `processorClosing`.

`processorClosed` -> `processorInitializing`, if try to initialize the sink
`processorInitializing` -> `processorRunning`, if the sinkManager is successfully initialized.
`processorRunning` -> `processorClosing`, if try to close the processor.
`processorClosing` -> `processorClosed` if the sinkManager is fully closed, not matter success or not.

* TablePipeline should not be created before the status is `processorRunning`.

### Check List 

Tests 

 - Manually test

* create 2 changefeed, 1 for MySQL, 1 for Kafka with open protocol, list changefeed show that two works fine.
* kill -9 the Kafka broker, and wait for 1 minute, logs show that owner receives an error and try to close the Kafka changefeed.
* During the Kafka closing period, which would cost another 1 minute, MySQL Changefeed still works normally.

Code changes

Side effects

Related changes

 - Need to cherry-pick to the release branch

### Release note 

```release-note
fix the owner may be blocked when closing the changefeed with Kafka sink which has a bad network connection between Kafka brokers.
```

PR #5186, opened 2022-04-15
sink(ticdc): asynchronously close the mq producers

### LINKED ISSUE #4241: mysql changefeed blocked by abnormal kafka changefeed
### What did you do?

- 1 TiCDC, 1 broker, 1 mysql; 3 kafka changefeeds, 1 mysql changefeed
- At first, all changefeeds work normally
- When kafka service stopped, the 3 kafka changefeeds don't sync data as expected, but the checkpoint of  mysql changefeed also will not move on.
- [
  {
    "id": "kafka-task-1",
    "summary": {
      "state": "normal",
      "tso": 430296185518948355,
      "checkpoint": "2022-01-06 14:14:42.308",
      "error": null
    }
  },
  {
    "id": "kafka-task-3",
    "summary": {
      "state": "normal",
      "tso": 430296185518948355,
      "checkpoint": "2022-01-06 14:14:42.308",
      "error": null
    }
  },
  {
    "id": "kafka-task-4",
    "summary": {
      "state": "normal",
      "tso": 430296185518948355,
      "checkpoint": "2022-01-06 14:14:42.308",
      "error": null
    }
  },
  {
    "id": "mysql-task-1",
    "summary": {
      "state": "normal",
      "tso": 430296201260695559,
      "checkpoint": "2022-01-06 14:15:42.358",
      "error": null
    }
  }
]
[root@CentOS76_VM log]# date
Thu Jan  6 14:42:04 CST 2022

mysql changefeed still can works normally

### What did you see instead?

cdc log has expected ERROR: [ERROR] [changefeed.go:118] ["an error occurred in Owner"] [changefeedID=kafka-task-3] [error="[CDC:ErrKafkaNewSaramaProducer]new sarama producer: kafka: client has run out of available brokers to talk to (Is your cluster reachable?)"]
WARN: [WARN] [client.go:226] ["etcd client outCh blocking too long, the etcdWorker may be stuck"] [duration=20m39.000245874s]

### Versions of the cluster

```console
5.4.0
```

```console
5.4.0
```

3AceShowHand:

Issue Number: close #4241 

as suggested by https://github.com/pingcap/tiflow/pull/4249#issuecomment-1099812788, close the sink asynchronously.

close the producer (kafka / pulsar) in an asynchronous way.

### Check List 

Tests 

Master branch, close the sink Manager would block a period of time. 
Screen Shot 2022-04-18 at 1 58 46 PM

This PR, sink Manager closed immediately, does not block the processor's progress.
Screen Shot 2022-04-18 at 2 11 50 PM

Code changes

Side effects

Related changes

### Release note 

```release-note
None
```


cdc/processor/pipeline/table_actor.go
```diff
@@ -134,12 +134,14 @@ func NewTableActor(cdcCtx cdcContext.Context,
...
 
+		changefeedID: changefeedVars.ID,
+
 		stopCtx: cctx,
 	}
 
 	startTime := time.Now()
 	log.Info("table actor starting",
-		zap.String("changfeed", changefeedVars.ID),
+		zap.String("changefeed", changefeedVars.ID),
```
hi-rustin, line 144: Why not use changefeedID?

hi-rustin (commented):
Can we do some correctness tests? For example, the correctness of shutting down the producer asynchronously after encountering a kafka error and resuming it again?

3AceShowHand:
> Can we do some correctness tests? For example, the correctness of shutting down the producer asynchronously after encountering a kafka error and resuming it again?

It looks already covered by the integration test `kafka_sink_error_resume` ?

hi-rustin:
> > Can we do some correctness tests? For example, the correctness of shutting down the producer asynchronously after encountering a kafka error and resuming it again?
> 
> It looks already covered by the integration test `kafka_sink_error_resume` ?

I don't think it's the same as the scenario where we actually have a problem? Otherwise this integration test should have caught the problem long ago?

3AceShowHand:
> > > Can we do some correctness tests? For example, the correctness of shutting down the producer asynchronously after encountering a kafka error and resuming it again?
> > 
> > 
> > It looks already covered by the integration test `kafka_sink_error_resume` ?
> 
> I don't think it's the same as the scenario where we actually have a problem? Otherwise this integration test should have caught the problem long ago?

3AceShowHand:
> > > Can we do some correctness tests? For example, the correctness of shutting down the producer asynchronously after encountering a kafka error and resuming it again?
> > 
> > 
> > It looks already covered by the integration test `kafka_sink_error_resume` ?
> 
> I don't think it's the same as the scenario where we actually have a problem? Otherwise this integration test should have caught the problem long ago?

I think the test purpose is that `the producer asynchronously closed after encountering a kafka error and resuming it again`, it's should be covered already.

> Otherwise this integration test should have caught the problem long ago?

Do you mean the problem that `blocked a long time`? It's not covered by the case ` kafka_sink_error_resume`, so that does not expose the problem.  Let's have a try to cover it by the integration test.

3AceShowHand:
> > > Can we do some correctness tests? For example, the correctness of shutting down the producer asynchronously after encountering a kafka error and resuming it again?
> > 
> > 
> > It looks already covered by the integration test `kafka_sink_error_resume` ?
> 
> I don't think it's the same as the scenario where we actually have a problem? Otherwise this integration test should have caught the problem long ago?

I think the test purpose is that `the producer asynchronously closed after encountering a kafka error and resuming it again`, it's should be covered already.

> Otherwise this integration test should have caught the problem long ago?

Do you mean the problem that `blocked a long time`? It's not covered by the case ` kafka_sink_error_resume`, so that does not expose the problem.  We have some test in test-infra to inject chaos to make the problems happens periodically. I will try to cover it in the integration test.

cdc/owner/ddl_sink.go
```diff
@@ -264,5 +264,8 @@ func (s *ddlSinkImpl) close(ctx context.Context) (err error) {
 		err = s.syncPointStore.Close()
 	}
 	s.wg.Wait()
-	return err
+	if err != nil && errors.Cause(err) != context.Canceled {
```
hi-rustin, line 267: Do we still use close here, instead of asyncclose?

hi-rustin (commented):
Can we add asynchronous close only to the close method of mq sink so that we don't need to add a new interface for the producer and write duplicate code. This behavior should really just be a special case of an mq sink close.

hi-rustin (approved):

hi-rustin:
Do we only need the two versions of cherry-pick?

hi-rustin:
> Can we add asynchronous close only to the close method of mq sink so that we don't need to add a new interface for the producer and write duplicate code. This behavior should really just be a special case of an mq sink close.

Or maybe we can pass ctx directly into close, I don't know how to do that better. cc: @asddongmen @overvenus @3AceShowHand  What do you think?

3AceShowHand:
> > Can we add asynchronous close only to the close method of mq sink so that we don't need to add a new interface for the producer and write duplicate code. This behavior should really just be a special case of an mq sink close.
> 
> Or maybe we can pass ctx directly into close, I don't know how to do that better. cc: @asddongmen @overvenus @3AceShowHand What do you think?

I would prefer the current implementation, it looks simple enough.

3AceShowHand:
> > Can we add asynchronous close only to the close method of mq sink so that we don't need to add a new interface for the producer and write duplicate code. This behavior should really just be a special case of an mq sink close.
> 
> Or maybe we can pass ctx directly into close, I don't know how to do that better. cc: @asddongmen @overvenus @3AceShowHand What do you think?

> maybe we can pass ctx directly into close

Do you mean `Producer.Close()`?  It looks like we have to start a goroutine just like now.

The current one is simple enough I think.

cdc/sink/mq.go
```diff
@@ -310,8 +310,13 @@ func (k *mqSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
 }
 
 func (k *mqSink) Close(ctx context.Context) error {
-	err := k.mqProducer.Close()
-	return errors.Trace(err)
+	closedCh := k.mqProducer.AsyncClose()
+	select {
+	case <-ctx.Done():
```
asddongmen, line 315: Please add a comment here to tell this `ctx` is a `canceled context` so that this select statement will never block.

cdc/sink/mq.go
```diff
@@ -310,8 +310,13 @@ func (k *mqSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
 }
 
 func (k *mqSink) Close(ctx context.Context) error {
```
asddongmen, line 312: It is better add a comment here to indicate that this func is non-blocking.

sdojjy (approved):

Issue #4464, opened 2022-01-25
changefeed get stuck ...in test case "cdc_scale_sync"


Tammyxia:
### What did you do?

- test infra test case: cdc_scale_sync, main branch
- the changefeed get stuck already for 6 hours: https://tcms.pingcap.net/dashboard/executions/plan/630680
- owner is not stuck because if creating new changefeed, the new one works fine.

### What did you see instead?

![image](https://user-images.githubusercontent.com/78345569/150916447-ec2a9626-72d5-41bc-80c8-ff49ae7f10cd.png)

### Versions of the cluster

Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

```console
main
```

TiCDC version (execute `cdc version`):

```console
[release-version=v5.5.0-nightly] [git-hash=7a227b421dbfcdafee02148e787138798edadf31] [git-branch=heads/refs/tags/v5.5.0-nightly] [utc-build-time="2022-01-24 18:10:05"] [go-version="go version go1.16.4 linux/amd64"] [failpoint-build=false]
```

Tammyxia:
Reproduced again in main version.
After cdc scale finished for about 2 hours, while waiting for sync task completing, changefeed stuck.
It seems tikv is the main reason, because tikv has no grpc message output any more from metrics
![image](https://user-images.githubusercontent.com/78345569/152764736-5f29030e-e13d-48de-86a1-ecea2696ea6b.png)
![image](https://user-images.githubusercontent.com/78345569/152764952-bd6da1e6-0970-457e-a3a1-07c64086f73b.png)

And tikv log ERROR:
[2022/02/06 20:36:48.101 +00:00] [ERROR] [transport.rs:86] ["failed to send significant msg"] [msg="CaptureChange { cmd: ChangeObserver { cdc_id: Some(ObserveHandle { id: ObserveID(1174), observing: true }), rts_id: None, region_id: 1228 }, region_epoch: conf_ver: 5 version: 99, callback: Callback::Read(..) }"]
[2022/02/06 20:36:48.101 +00:00] [ERROR] [endpoint.rs:637] ["cdc initialize fail: Request error message: \"region 1228 not found\" region_not_found { region_id: 1228 }"] [region_id=1228]
[...]

PR #5196, opened 2022-04-18
sink(ticdc): refine sink interface and add init method

### LINKED ISSUE #5197: ticdc: refine sink interface add init method
refine cdc sink interface
add an init method to init and clear up old values 
remove drawback from buffer sink

sdojjy:

We need to clean up the old values of the table checkpoint, otherwise when the table is dispatched back again, it may read the old values.  See: https://github.com/pingcap/tiflow/issues/4464#issuecomment-1085385382. 
To guarantee the value is cleared, we clear old values when the table  pipeline is created.

Issue Number: close #5197 ref https://github.com/pingcap/tiflow/issues/4464

refine cdc sink interface
add an init method to init and clear up old values 
remove drawback from buffer sink

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
`None`.
```


nongfushanquan (commented):
whether the test case in issue 4464  can be reproduced after this pr ?

cdc/sink/manager.go
```diff
@@ -111,30 +110,13 @@ func (m *Manager) destroyTableSink(ctx context.Context, tableID model.TableID) e
 	m.tableSinksMu.Lock()
 	delete(m.tableSinks, tableID)
 	m.tableSinksMu.Unlock()
-	callback := make(chan struct{})
-	select {
-	case <-ctx.Done():
-		return ctx.Err()
-	case m.drawbackChan <- drawbackMsg{tableID: tableID, callback: callback}:
```
hi-rustin, line 118: After we remove this mechanism, how do we ensure that it will stop?

sdojjy:
> whether the test case in issue 4464 can be reproduced after this pr ?

this pr will fix that issue by cleaning the old checkpoint value.

cdc/sink/mysql.go
```diff
@@ -510,6 +510,30 @@ func (s *mysqlSink) dispatchAndExecTxns(ctx context.Context, txnsGroup map[model
 	s.notifyAndWaitExec(ctx)
 }
 
+func (s *mysqlSink) Init(tableID model.TableID) error {
+	s.cleanTableResource(tableID)
+	return nil
+}
+
+func (s *mysqlSink) cleanTableResource(tableID model.TableID) {
```
hi-rustin, line 518: Can we add some tests to cover these cleanups to really take effect?

hi-rustin (approved):
rest LGTM

Approved in advance, but adding more tests is better! Thanks for working on this.

cdc/sink/mysql.go
sdojjy: done

3AceShowHand (approved):

PR #5203, opened 2022-04-19
sink(ticdc): use slice in txnsWithTheSameCommitTs to support splitting transaction

CharlesCheung96:

Issue Number: ref #5231

1. Make memory and unified sorter modules have the same sorting rules as the leveldb sorter
2. Use slices instead of maps in txnsWithTheSameCommitTs to support splitting transaction.
3. Refactor test cases to match real data, which is sorted by commitTs and startTs.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/model/mounter.go
+++ b/cdc/model/mounter.go
@@ -48,3 +48,25 @@ func NewResolvedPolymorphicEvent(regionID uint64, resolvedTs uint64) *Polymorphi
 func (e *PolymorphicEvent) RegionID() uint64 {
 	return e.RawKV.RegionID
 }
+
+// ComparePolymorphicEvents compares two events by CRTs, Resolved, StartTs, Put/Delete order.
+func ComparePolymorphicEvents(i, j *PolymorphicEvent) bool {
+	if i.CRTs == j.CRTs {
+		if j.RawKV.OpType == OpTypeResolved && i.RawKV.OpType != OpTypeResolved {
+			return true
+		} else if i.RawKV.OpType == OpTypeResolved {
+			return false
+		}
+
+		if i.StartTs < j.StartTs {
+			return true
+		} else if i.StartTs > j.StartTs {
+			return false
+		}
+
+		if i.RawKV.OpType == OpTypeDelete && j.RawKV.OpType != OpTypeDelete {
+			return true
+		}
+	}
+	return i.CRTs < j.CRTs
+}
--- a/cdc/model/mounter.go
+++ b/cdc/model/mounter.go
@@ -50,6 +50,7 @@ func (e *PolymorphicEvent) RegionID() uint64 {
 }
 
 // ComparePolymorphicEvents compares two events by CRTs, Resolved, StartTs, Delete/Put order.
+// It returns true if and only if i should precede j.
 func ComparePolymorphicEvents(i, j *PolymorphicEvent) bool {
 	if i.CRTs == j.CRTs {
 		if i.RawKV.OpType == OpTypeResolved {
--- a/cdc/model/mounter.go
+++ b/cdc/model/mounter.go
@@ -49,19 +49,19 @@ func (e *PolymorphicEvent) RegionID() uint64 {
 	return e.RawKV.RegionID
 }
 
-// ComparePolymorphicEvents compares two events by CRTs, Resolved, StartTs, Put/Delete order.
+// ComparePolymorphicEvents compares two events by CRTs, Resolved, StartTs, Delete/Put order.
 func ComparePolymorphicEvents(i, j *PolymorphicEvent) bool {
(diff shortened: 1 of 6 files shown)


3AceShowHand (approved):

cdc/sorter/unified/heap.go
```diff
@@ -25,10 +25,19 @@ type sortHeap []*sortItem
 func (h sortHeap) Len() int { return len(h) }
 func (h sortHeap) Less(i, j int) bool {
 	if h[i].entry.CRTs == h[j].entry.CRTs {
-		if h[j].entry.RawKV.OpType == model.OpTypeResolved && h[i].entry.RawKV.OpType != model.OpTypeResolved {
+		if h[j].entry.RawKV.OpType == model.OpTypeResolved {
```
liuzix, line 28: Consider two `Resolved` events A and B with the same `CommitTs`, before this change,  less(A, B) would be `false`, but after this changed, it would be `true`.

cdc/sorter/memory/entry_sorter.go
```diff
@@ -159,12 +159,27 @@ func (es *EntrySorter) Output() <-chan *model.PolymorphicEvent {
...
 	if i.CRTs == j.CRTs {
-		if i.RawKV.OpType == model.OpTypeDelete {
+		if j.RawKV.OpType == model.OpTypeResolved &&
+			i.RawKV.OpType != model.OpTypeResolved {
+			return true
+		} else if i.RawKV.OpType == model.OpTypeResolved {
+			return false
+		}
+
+		if i.StartTs < j.StartTs && i.StartTs != j.StartTs {
```
liuzix, line 175:
```suggestion
		if i.StartTs < j.StartTs {
```

cdc/sorter/memory/entry_sorter.go
```diff
@@ -159,12 +159,27 @@ func (es *EntrySorter) Output() <-chan *model.PolymorphicEvent {
 }
 
 func eventLess(i *model.PolymorphicEvent, j *model.PolymorphicEvent) bool {
+	return Compare(i, j)
+}
+
+// Compare compares two events by CRTs, Resolved, StartTs, Put/Delete order.
+func Compare(i *model.PolymorphicEvent, j *model.PolymorphicEvent) bool {
```
liuzix, line 166: Maybe this function belongs to `model` together with `PolymorphicEvent`?

CharlesCheung96: Done, PTAL!

cdc/model/mounter.go
```diff
@@ -48,3 +48,27 @@ func NewResolvedPolymorphicEvent(regionID uint64, resolvedTs uint64) *Polymorphi
 func (e *PolymorphicEvent) RegionID() uint64 {
 	return e.RawKV.RegionID
 }
+
+// ComparePolymorphicEvent compares two events by CRTs, Resolved, StartTs, Put/Delete order.
+func ComparePolymorphicEvent(i *PolymorphicEvent, j *PolymorphicEvent) bool {
```
liuzix, line 53:
```suggestion
func ComparePolymorphicEvents(i *PolymorphicEvent, j *PolymorphicEvent) bool {
```

cdc/model/mounter.go
```diff
@@ -48,3 +48,25 @@ func NewResolvedPolymorphicEvent(regionID uint64, resolvedTs uint64) *Polymorphi
 func (e *PolymorphicEvent) RegionID() uint64 {
 	return e.RawKV.RegionID
 }
+
+// ComparePolymorphicEvents compares two events by CRTs, Resolved, StartTs, Delete/Put order.
```
liuzix, line 52:
```suggestion
// ComparePolymorphicEvents compares two events by CRTs, Resolved, StartTs, Delete/Put order.
// It returns true if and only if i should precede j.
```

cdc/sink/mysql/txn_cache.go
```diff
@@ -35,17 +35,23 @@ func (t *txnsWithTheSameCommitTs) Append(row *model.RowChangedEvent) {
 			zap.Any("row", row))
 	}
 	if t.txns == nil {
-		t.txns = make(map[model.Ts]*model.SingleTableTxn)
+		t.txns = make([]*model.SingleTableTxn, 0)
```
liuzix, line 38: `nil` is considered a valid slice in Go. No need to create it by `make`. `append` handles the `nil` case automatically.

CharlesCheung96: Good suggestion, thanks for your review.

liuzix (approved):

PR #5243, opened 2022-04-22
validator(dm): fix dmworker panic due to validator

### LINKED ISSUE #5184: starting validator makes dmworker panic
### What did you do?

1. start task with validator on
2. `dmctl query-status`

print the status of the worker

### What did you see instead?

worker panicked and exited

### Versions of the cluster

commit: bca4b480792536d2548fc3a02c41ebd93f86e25f

```console
```

buchuitoudegou:

Issue Number: close #5184 

put off initializing workers until start validator workers.

### Check List 

Tests 

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`
```

### DIFF
--- a/dm/syncer/data_validator.go
+++ b/dm/syncer/data_validator.go
@@ -171,7 +171,6 @@ func NewContinuousDataValidator(cfg *config.SubTaskConfig, syncerObj *Syncer, st
 
 	v.workerCnt = cfg.ValidatorCfg.WorkerCount
 	v.processedRowCounts = make([]atomic.Int64, rowChangeTypeCount)
-	v.workers = make([]*validateWorker, v.workerCnt)
 	v.validateInterval = validationInterval
 	v.persistHelper = newValidatorCheckpointHelper(v)
 	v.tableStatus = make(map[string]*tableValidateStatus)
@@ -571,6 +570,7 @@ func (v *DataValidator) Stage() pb.Stage {
 
 func (v *DataValidator) startValidateWorkers() {
 	v.wg.Add(v.workerCnt)
+	v.workers = make([]*validateWorker, v.workerCnt)
 	for i := 0; i < v.workerCnt; i++ {
 		worker := newValidateWorker(v, i)
 		v.workers[i] = worker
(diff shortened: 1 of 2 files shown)


dm/syncer/data_validator.go
```diff
@@ -571,6 +570,7 @@ func (v *DataValidator) Stage() pb.Stage {
 
 func (v *DataValidator) startValidateWorkers() {
 	v.wg.Add(v.workerCnt)
+	v.workers = make([]*validateWorker, v.workerCnt)
```
Ehco1996, line 573: which line panic,  can't understand why this fixes the problem, could you please explain more?

buchuitoudegou: ptal https://github.com/pingcap/tiflow/pull/5181#discussion_r852549354

lance6716:
Add a test to avoid regression

dm/syncer/data_validator.go
D3Hunter: v.workers has already inited here
https://github.com/pingcap/tiflow/blob/56ccdf2178468331687ae2f07e83e38111811b27/dm/syncer/data_validator.go#L174

it panics here? not checks worker:
https://github.com/pingcap/tiflow/blob/56ccdf2178468331687ae2f07e83e38111811b27/dm/syncer/validator_checkpoint.go#L191

maybe we should init this field when create workers

buchuitoudegou: > v.workers has already inited here
> 
> https://github.com/pingcap/tiflow/blob/56ccdf2178468331687ae2f07e83e38111811b27/dm/syncer/data_validator.go#L174

Yes, and I removed this to put off initializing the workers array till we start validating. 

https://github.com/pingcap/tiflow/blob/56ccdf2178468331687ae2f07e83e38111811b27/dm/syncer/validator_checkpoint.go#L191
> 
> maybe we should init this field when create workers

It's not because we didn't initialize worker's field but merely the worker itself is nil (because we just initialized the workers' array but the workers inside were nil before starting).

D3Hunter (approved):

D3Hunter:
> Add a test to avoid regression

it's an bug internal to validator, so doesn't affects query-status when validator is off

buchuitoudegou:
I've added a simple ut. PTAL @lance6716

lance6716 (approved):
Another convention is use make([]type, 0, cap) for slices, so we will not access not initialized values

PR #5246, opened 2022-04-22
pkg/gtid: remove gtid.Set interface

lance6716:

Issue Number: close #4287 

compared with mysql.GTIDSet, gtid.Set has two more methods. 
- `Replace` is no longer used
- `Truncate` is used but does not take effects, see https://github.com/pingcap/dm/pull/335#issuecomment-1106069436

### Check List 

Tests 

 - Unit test

Code changes

Related changes

### Release note 

```release-note
 `None`.
```


lance6716:
I'll write a better copy-on-write API for GTID set soon

dm/syncer/checkpoint.go
```diff
@@ -71,7 +72,15 @@ type tablePoint struct {
 }
 
 func (b *tablePoint) String() string {
-	return fmt.Sprintf("location(%v), tableInfo(ID: %d, Name:%s, ColNum: %d, IdxNum: %d, PKIsHandle: %t)", b.location, b.ti.ID, b.ti.Name, len(b.ti.Columns), len(b.ti.Indices), b.ti.PKIsHandle)
+	if b == nil {
+		return ""
+	}
+	var buf strings.Builder
+	buf.WriteString(fmt.Sprintf("location(%s)", b.location))
```
lichunzhu, line 79:
```suggestion
	fmt.Fprintf(&buf, "location(%s)", b.location)
```

dm/syncer/checkpoint.go
```diff
@@ -71,7 +72,15 @@ type tablePoint struct {
...
+	if b.ti != nil {
+		buf.WriteString(fmt.Sprintf(", tableInfo(ID: %d, Name:%s, ColNum: %d, IdxNum: %d, PKIsHandle: %t)", b.ti.ID, b.ti.Name, len(b.ti.Columns), len(b.ti.Indices), b.ti.PKIsHandle))
```
lichunzhu, line 81: ditto

dm/syncer/checkpoint.go
```diff
@@ -183,7 +192,7 @@ func (b *binlogPoint) String() string {
 	b.RLock()
 	defer b.RUnlock()
 
-	return fmt.Sprintf("%v(flushed %v)", b.savedPoint, b.flushedPoint)
+	return fmt.Sprintf("%s(flushed %s)", b.savedPoint.String(), b.flushedPoint.String())
```
lichunzhu, line 195: ditto

lichunzhu (approved):
LGTM

dm/pkg/binlog/position.go
```diff
@@ -398,37 +411,23 @@ func (l *Location) ResetSuffix() {
 	l.Suffix = 0
 }
 
-// SetGTID set new gtid for location
-// Use this func instead of GITSet.Set to avoid change other location.
+// SetGTID set new gtid for location.
+// TODO: don't change old Location and return a new one to copy-on-write.
```
Ehco1996, line 415: >  don't change old Location and return a new one to copy-on-write

I can't remember why we need to do this, can you please give me some tips? is this about the func`location.CloneWithFlavor` ?

Ehco1996 (commented):
lgtm

dm/pkg/binlog/position.go
lance6716: it's a new request. mainly about reducing the possibility that when two variables hold the same reference to a location, changing one of them will accidentally change the other one: SetGTID should not change the input location. I think new API is less error-prone

this problem has another solutions, for example, we use location struct not location pointer, so if a location is assigned to another variables they are different copies. we can choose one solution in following pr.

lance6716:
ptal @Ehco1996 @GMHDBJD

lance6716:
I think this change will reduce efforts on other PRs which are related to GTID, so I change the base branch to master. cc @Ehco1996 @lichunzhu @GMHDBJD 

will resolve conflict later.

Ehco1996 (approved):
LGTM ~ nice work

Issue #5249, opened 2022-04-22
DM-worker reports data race when doing some tests


lichunzhu:
### What did you do?

Run dm-worker normally.

### What did you expect to see?

dm-worker works normally.

### What did you see instead?

DM-worker reports data race for `w.cfg`:

```log
==================
WARNING: DATA RACE
Read at 0x00c00066e5c8 by goroutine 51:
  github.com/pingcap/tiflow/dm/dm/worker.(*SourceWorker).observeValidatorStage()
      /Users/chauncy/code/goPath/src/github.com/pingcap/ticdc/dm/dm/worker/source_worker.go:1163 +0x777
  github.com/pingcap/tiflow/dm/dm/worker.(*SourceWorker).EnableHandleSubtasks.func2()
      /Users/chauncy/code/goPath/src/github.com/pingcap/ticdc/dm/dm/worker/source_worker.go:492 +0xc4

Previous write at 0x00c00066e5c8 by goroutine 192:
  github.com/pingcap/tiflow/dm/dm/worker.(*SourceWorker).refreshSourceCfg()
      /Users/chauncy/code/goPath/src/github.com/pingcap/ticdc/dm/dm/worker/source_worker.go:1286 +0x11b
  github.com/pingcap/tiflow/dm/dm/worker.(*SourceWorker).tryRefreshSubTaskAndSourceConfig()
      /Users/chauncy/code/goPath/src/github.com/pingcap/ticdc/dm/dm/worker/source_worker.go:1292 +0x64
  github.com/pingcap/tiflow/dm/dm/worker.(*SourceWorker).OperateSubTask()
      /Users/chauncy/code/goPath/src/github.com/pingcap/ticdc/dm/dm/worker/source_worker.go:635 +0x46e
  github.com/pingcap/tiflow/dm/dm/worker.(*SourceWorker).operateSubTaskStage()
      /Users/chauncy/code/goPath/src/github.com/pingcap/ticdc/dm/dm/worker/source_worker.go:837 +0x5e4
  github.com/pingcap/tiflow/dm/dm/worker.(*SourceWorker).resetSubtaskStage()
      /Users/chauncy/code/goPath/src/github.com/pingcap/ticdc/dm/dm/worker/source_worker.go:698 +0x3ad
  github.com/pingcap/tiflow/dm/dm/worker.(*SourceWorker).observeSubtaskStage()
[...]

Issue #5235, opened 2022-04-21
syncer reports error reading from server: EOF and paused


lichunzhu:
### Which jobs are flaking?

IT-shardddl3

### Which test(s) are flaking?

DM_097

### Jenkins logs or GitHub Actions link

https://ci2.pingcap.net/job/dm_ghpr_integration_test/5098/display/redirect

### Anything else we need to know

```log
[2022-04-21T09:32:50.215Z] [2022/04/21 17:32:21.444 +08:00] [ERROR] [subtask.go:392] ["unit process error"] [subtask=test] [unit=Sync] ["error information"="ErrCode:50000 ErrClass:\"not-set\" ErrScope:\"not-set\" ErrLevel:\"high\" Message:\"startLocation: [position: (dm-integration-test-go1180-5098-3hvrw-35rq4-bin|000001.000001, 88195), gtid-set: 7260a9b7-c155-11ec-8e12-b6d4cb74e1dd:1-377], endLocation: [position: (dm-integration-test-go1180-5098-3hvrw-35rq4-bin|000001.000001, 88328), gtid-set: 7260a9b7-c155-11ec-8e12-b6d4cb74e1dd:1-378], origin SQL: [alter table shardddl1.tb2 add column col int]: rpc error: code = Unavailable desc = error reading from server: EOF\" "]
```

lichunzhu:
This will happen when dm-master is down and we are trying to put optimistic operation. This problem will make dm-worker's task paused and not easy to find out what's happening.

PR #5250, opened 2022-04-22
worker: add lock protection for w.cfg

lichunzhu:

Issue Number: close https://github.com/pingcap/tiflow/issues/5249

Add lock protection for `w.cfg`.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

- None

Side effects

- None

Related changes

- None

### Release note 

```release-note
None
```

### DIFF
--- a/dm/dm/worker/source_worker.go
+++ b/dm/dm/worker/source_worker.go
@@ -43,9 +43,10 @@ import (
 
 // SourceWorker manages a source(upstream) which is mainly related to subtasks and relay.
 type SourceWorker struct {
-	// ensure no other operation can be done when closing (we can use `WatGroup`/`Context` to archive this)
+	// ensure no other operation can be done when closing (we can use `WaitGroup`/`Context` to archive this)
 	// TODO: check what does it guards. Now it's used to guard relayHolder and relayPurger (maybe subTaskHolder?) since
 	// query-status maybe access them when closing/disable functionalities
+	// This lock is used to guards source worker's source config and subtask holder(subtask configs)
 	sync.RWMutex
 
 	wg     sync.WaitGroup
--- a/dm/dm/worker/source_worker.go
+++ b/dm/dm/worker/source_worker.go
@@ -249,9 +249,12 @@ func (w *SourceWorker) Stop(graceful bool) {
 // updateSourceStatus updates w.sourceStatus.
 func (w *SourceWorker) updateSourceStatus(ctx context.Context) error {
 	w.sourceDBMu.Lock()
+	w.RLock()
+	cfg := w.cfg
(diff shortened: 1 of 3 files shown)


dm/dm/worker/source_worker.go
```diff
@@ -249,9 +249,12 @@ func (w *SourceWorker) Stop(graceful bool) {
 // updateSourceStatus updates w.sourceStatus.
 func (w *SourceWorker) updateSourceStatus(ctx context.Context) error {
 	w.sourceDBMu.Lock()
+	w.RLock()
+	cfg := w.cfg
```
Ehco1996, line 253: seems this lock can not avoid data-race? w.cfg is a pointer

dm/dm/worker/status.go
```diff
@@ -131,7 +131,10 @@ func (w *SourceWorker) GetValidateStatus(stName string, filterStatus pb.Stage) [
 	if st == nil {
 		return res
 	}
-	sourceIP := w.cfg.From.Host + ":" + strconv.Itoa(w.cfg.From.Port)
+	w.RLock()
+	cfg := w.cfg
```
Ehco1996, line 135: ditto

Ehco1996 (commented):
rest LGTM

dm/dm/worker/source_worker.go
lichunzhu: This lock is used to prevent the situation that `w.cfg` is updated in while we are trying to access `w.cfg` at https://github.com/pingcap/tiflow/blob/f29084c/dm/dm/worker/source_worker.go#L1286

Ehco1996: ok, `cfg` here points to `oldCfg` and at first, I wondered if there was a problem with correctness, e.g: getting the `cfg` before the update, but there is no problem because the source cfg only  allowed to be updated when there is no running tasking(running source worker)

Ehco1996 (approved):
LGTM

lance6716 (approved):
Please add comment for SourceWorker.RWMutex that listing known members that it guards

dm/dm/worker/source_worker.go
lichunzhu: To me there is no big problem. `RefreshSourceCfg` won't revise cfg.SourceID. The other race place may let the user get wrong source IP, but that's not a big deal.

lichunzhu:
/run-dm-integration-test
https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/5141/pipeline
same error to https://github.com/pingcap/tiflow/issues/5235

PR #5254, opened 2022-04-24
validator(dm): reuse sqlmodel.RowChange and use defined error in error_list

### LINKED ISSUE #5253: use defined error in error_list, resusing sqlmodel.RowChange

D3Hunter:

Issue Number: close #5253

- reuse sqlmodel.RowChange
- use defined error in error_list

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`.
```


dm/syncer/data_validator.go
```diff
@@ -493,15 +509,16 @@ func (v *DataValidator) doValidate() {
 		// wait until syncer synced current event
 		err = v.waitSyncerSynced(currLoc)
 		if err != nil {
-			v.errChan <- terror.Annotate(err, "failed to wait syncer")
+			// no need to wrapped it in error_list, since err can be context.Canceled only.
```
buchuitoudegou, line 512:
```suggestion
			// no need to wrap it in error_list, since err can be context.Canceled only.
```

dm/syncer/data_validator.go
```diff
@@ -581,55 +598,67 @@ func (v *DataValidator) startValidateWorkers() {
...
 	}
-	// todo: PKIsHandle = true when table has primary key like "id int primary key CLUSTERED", since schema-tracker(we get from it)
-	// todo: only use downstream DDL when the task is incremental only, will support this case later.
-	if primaryIdx == nil {
-		// return partial initialized result
-		return table, errNoPrimaryKey
+	// if current TI has more columns, clone and strip columns
+	if len(tableInfo.Columns) > columnCount {
+		tableInfo = tableInfo.Clone()
+		tableInfo.Columns = tableInfo.Columns[:columnCount]
```
buchuitoudegou, line 656: Considering the following case:
upstream binlog has columns: `a`, `b`, `c`; downstream table has column: `b`, `c`, `d`, `e` (this may happen because we don't forbid DDL event currently)

Simply slicing the `Columns` doesn't work. Should we leave a warning when DDLs come and tries modifying the schemas? Or should we have more decent way to check if the schemas matched?

dm/syncer/data_validator.go
```diff
@@ -698,69 +715,58 @@ func (v *DataValidator) processRowsEvent(header *replication.EventHeader, ev *re
...
 	}
 	return nil
 }
 
-func (v *DataValidator) flushCheckpointAndData(loc binlog.Location) error {
+func (v *DataValidator) persistCheckpointAndData(loc binlog.Location) error {
 	var wg sync.WaitGroup
 	wg.Add(v.workerCnt)
-	flushJob := &rowChange{
+	flushJob := &rowChangeJob{
```
buchuitoudegou, line 769: Little complaint: a bit weird to name a flushJob `rowChangeJob`. It seems these jobs are all assigned to workers so it'd be better called `workerJob` (or something like that). No need to change, I think, because it would be a big work to do. I'm happy to leave it here😂

dm/syncer/validator_checkpoint.go
```diff
@@ -183,13 +183,27 @@ func (c *validatorPersistHelper) createTable(tctx *tcontext.Context) error {
 	return err
 }
 
+type tableChangeDataForPersist struct {
+	sourceTable *filter.Table
+	columnCount int
+	jobs        map[string]*rowChangeDataForPersist
```
buchuitoudegou, line 189:
```suggestion
	rows        map[string]*rowChangeDataForPersist
```
Why not rows?

pkg/sqlmodel/reduce.go
```diff
@@ -58,6 +55,41 @@ func (r *RowChange) IdentityValues() ([]interface{}, []interface{}) {
 	return pre, post
 }
 
+// RowIdentity returns the identity of this row change, caller should
+// call IsIdentityUpdated/SplitUpdate before calling this method to
+// make sure it's not updating the identity itself.
+// we extract identity from preValues for update/delete, postValues for insert.
+// if there's no primary key, return all values.
```
buchuitoudegou, line 62: Does it mean we can identify rows without PK (so that we can remove the `WithoutPKError`)?

D3Hunter: whether row has PK is determined by the tableinfo, it will be skipped if there no PK

where's WithoutPKError?

buchuitoudegou: I mean we can validate the rows without PK (instead of skipping them) because this function helps identifying rows thought they don't have PK.

dm/syncer/data_validator.go
D3Hunter: downstream should have all the columns as in upstream, else syncer will fail too.

we need all columns in the binlog not dropped or switch order in tracker ddl, it's a constraint.

dm/syncer/data_validator.go
D3Hunter: renaming is easy using goland.

pkg/sqlmodel/reduce.go
D3Hunter: this method may used in other place, `if there's no primary key, return all values.` is just a workaround when there's no real identity.

cannot validate pk-less table, see 2 same rows in upstream, only 1 in downstream, validation will success. and validate pk-less table put a lot of load on downstream.

dm/syncer/data_validator.go
Ehco1996: syncer have a hidden feature which may cause different column name between upstream and down stream  see more `s.mappingDML` 😵

Ehco1996 (commented):
will review later

dm/syncer/validate_worker.go
```diff
@@ -281,19 +278,16 @@ func (vw *validateWorker) getErrorRows() []*validateFailedRow {
...
+		pkValues = append(pkValues, r.row.RowStrIdentity())
 	}
-	colCnt := len(rows[0].Data)
+	firstRow := rows[0].row
 	cond := &Cond{
-		Table:     table,
-		ColumnCnt: colCnt,
+		TargetTbl: firstRow.TargetTableID(),
+		Columns:   firstRow.SourceTableInfo().Columns,
+		PK:        firstRow.UniqueNotNullIdx(),
```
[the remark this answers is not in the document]
D3Hunter: it's checked inside `validator.genValidateTableInfo`, those rows are skipped before send into worker

dm/syncer/data_validator.go
```diff
@@ -43,30 +43,32 @@ import (
...
-	Target     *filter.Table // target table after route
+	targetTable         *filter.Table
+	srcTableInfo        *model.TableInfo
+	downstreamTableInfo *schema.DownstreamTableInfo
+
+	message string
 }
 
-type rowChangeType int
+type rowChangeJobType int
```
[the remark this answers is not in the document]
D3Hunter: this type need difference between insert/update/delete, syncer needn't.

dm/syncer/data_validator.go
D3Hunter: it's not changing column name but values(add prefix or suffix). would not support tasks with this hidden feature in this version.

D3Hunter:
ping @lance6716

dm/syncer/data_validator.go
```diff
@@ -656,40 +681,32 @@ func (v *DataValidator) processRowsEvent(header *replication.EventHeader, ev *re
...
+		state = &tableValidateStatus{
 			source: *sourceTable,
-			target: *table.Target,
+			target: *targetTable,
 			stage:  pb.Stage_Running,
 		}
+		v.tableStatus[fullTableName] = state
 	}
+	if validateTbl.message != "" {
+		state.stopped(validateTbl.message)
```
Ehco1996, line 705: this worker will stop when column count mismatch, do we need to add a log here?

Ehco1996 (commented):
rest LGTM

dm/syncer/data_validator.go
D3Hunter: fix in https://github.com/pingcap/tiflow/pull/5254/commits/7efda04e17ac61c3bde2f5d45a3be5f53e10cb11

Ehco1996 (approved):
LGTM

buchuitoudegou (approved):

D3Hunter:
ping @lance6716

PR #5258, opened 2022-04-24
processor(cdc): extract getTableName method

### LINKED ISSUE #5259: cdc simplify the get table name logic.
as title.

3AceShowHand:

Issue Number: close #5259 

* extract `processor.getTableName` method
* `processor.tick` return `nextState`, which is not used, so remove it.

### Check List 

Tests 
 - No code

Code changes

Side effects

Related changes

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/processor/processor.go
+++ b/cdc/processor/processor.go
@@ -927,7 +927,70 @@ func (p *processor) addTable(ctx cdcContext.Context, tableID model.TableID, repl
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


liuzix:
Content LGTM, but maybe we can split the PR into two or three parts? I think two parts are enough, one for processor-related changes, another for the scheduler.

3AceShowHand:
> Content LGTM, but maybe we can split the PR into two or three parts? I think two parts are enough, one for processor-related changes, another for the scheduler.

Thanks for your review, I will revert changes in the `scheduler` part, and do it in other scheduler related PR.

asddongmen (approved):

overvenus (approved):

Issue #5230, opened 2022-04-21
ScanRegions total retry time is too short


overvenus:
### What did you do?

Run tpcc prepare on 4 node TiKV cluster. Create a changefeed to replication upstream write.

### What did you expect to see?

No replication interruption.

### What did you see instead?

Replication is interrupted due to following errors, which is caused by https://github.com/tikv/tikv/issues/12410

```log
[...]

PR #5269, opened 2022-04-25
cdc,retry: fix leader missing by extending region retry duration

overvenus:

Issue Number: close #5230

* Fix replication interruption due to leader missing by extending region retry duration

Other changes
* Remove `oldvalue` in puller and kv client.
* Refine retry error info, includes the original error.

### Check List 

Tests 

 - Unit test

Related changes

### Release note 

```release-note
Fix replication interruption due to leader missing by extending region retry duration
```


cdc/kv/client.go
```diff
@@ -961,7 +959,8 @@ func (s *eventFeedSession) divideAndSendEventFeedToRegions(
 				zap.Reflect("regions", metas),
 				zap.String("changefeed", s.client.changefeed))
 			return nil
-		}, retry.WithBackoffMaxDelay(50), retry.WithMaxTries(100), retry.WithIsRetryableErr(cerror.IsRetryableError))
```
3AceShowHand, line 964: it looks that this is the root cause of the problem.

how about add some comment about retry logic here.

3AceShowHand (approved):

cdc/kv/client.go
```diff
@@ -294,7 +294,6 @@ type CDCKVClient interface {
 		ctx context.Context,
 		span regionspan.ComparableSpan,
 		ts uint64,
-		enableOldValue bool,
```
nongfushanquan, line 297: why remove the enableOldValue flag in this PR?  Is there  an issue related to this item?

pkg/retry/retry_with_opt.go
```diff
@@ -47,6 +47,14 @@ func run(ctx context.Context, op Operation, retryOption *retryOptions) error {
 	default:
 	}
 
+	if retryOption.totalRetryDuration != time.Duration(0) {
+		maxTries := getApproximateMaxTryBasedOnTotalRetryDuration(
```
CharlesCheung96, line 51: Why not just manage the total retry duration here?

cdc/kv/client.go
overvenus: Comments added in kv client config.

cdc/kv/client.go
overvenus: The `enableOldValue ` param has been deprecated monthes ago, it's a cleanup for dead code.

pkg/retry/options.go
```diff
@@ -66,18 +70,19 @@ func WithBackoffMaxDelay(delayInMs int64) Option {
 }
 
 // WithMaxTries configures maximum tries, if tries <= 0 "defaultMaxTries" will be used
```
asddongmen, line 72: `if tries <= 0 "defaultMaxTries" will be used` seems wrong.

pkg/retry/options.go
```diff
@@ -15,14 +15,16 @@ package retry
...
 )
 
 const (
 	// defaultBackoffBaseInMs is the initial duration, in Millisecond
 	defaultBackoffBaseInMs = 10.0
 	// defaultBackoffCapInMs is the max amount of duration, in Millisecond
-	defaultBackoffCapInMs = 100.0
-	defaultMaxTries       = 3
+	defaultBackoffCapInMs   = 100.0
+	defaultMaxTries         = math.MaxInt64
```
asddongmen, line 26: Why do you set this to `math.MaxInt64`?

asddongmen (approved):

CharlesCheung96 (approved):

PR #5281, opened 2022-04-26
sink(ticdc): make mysql sink support split transactions

### LINKED ISSUE #5280: Making mysql sink support split transactions

CharlesCheung96:

Issue Number: close #5280

1. Add a `SplitTxn` field to `RowChangedEvent`, which identifies whether a transaction is split from the current event.
2. Split transactions in flow control module and record the size of transaction using txnSizeEntry.
3. Make sink support split transactions so that a single large transaction can be written concurrently to downstream, which can significantly reduce synchronization latency.

Note: In this PR, FlowControl system still release memory based on [checkPointTs](https://github.com/pingcap/tiflow/blob/master/cdc/sink/flowcontrol/flow_control.go#L96). After sink module supports batch resolve mechanism, FlowControl should release memory based on [`txnSizeEntry`](https://github.com/pingcap/tiflow/pull/5281/files#diff-cfe8ccb173f8b93acc1b97cc5825fdb565d539634714fadffacdf18ea4bc8b26R46).

### Check List 

Tests 

 - Unit test
 - Integration test
> 
> total row of txn: 2 million
> txn size: ~360MB
> downstream: tidb cluster

image
image
image

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`Reduce mysql sink synchronization latency`.
```


cdc/sink/flowcontrol/flow_control.go
```diff
@@ -110,6 +119,62 @@ func (c *TableFlowController) Release(resolvedTs uint64) {
...
+	// Processing txns with the same commitTs.
+	txnEntry := e.(*txnSizeEntry)
+	if txnEntry.commitTs != lastCommitTs {
+		log.Panic("got wrong commitTs from deque, report a bug",
+			zap.Uint64("lastCommitTs", c.lastCommitTs),
+			zap.Uint64("commitTsInDeque", txnEntry.commitTs))
+	}
+
+	// 2. Append row to current transaction entry.
+	if txnEntry.startTs == msg.Row.StartTs &&
```
nongfushanquan, line 155: if txnEntry.startTs  != msg.Row.StartTs ,what should be done here ?

cdc/sink/flowcontrol/flow_control.go
```diff
@@ -82,13 +97,7 @@ func (c *TableFlowController) Consume(commitTs uint64, size uint64, blockCallBac
...
 	}
 
-	c.queueMu.Lock()
-	defer c.queueMu.Unlock()
-	c.queueMu.queue.PushBack(&commitTsSizeEntry{
-		commitTs: commitTs,
-		size:     size,
-	})
-
+	c.enqueueSingleMsg(msg, commitTs, size)
```
nongfushanquan, line 100: this PR doesn't split the big transaction , just make a mark ?

CharlesCheung96: This is divided into two steps：
1. make sink support split transactions, so that a single large transaction can be written concurrently to downstream, which can significantly reduce synchronization latency
2. implement batch resolve mechanism, which can alleviate the oom problem to a certain extent

cdc/sink/flowcontrol/flow_control.go
CharlesCheung96: Then the program will go to the third part and create a new txnEntry, ref: https://github.com/pingcap/tiflow/pull/5281/files#diff-cfe8ccb173f8b93acc1b97cc5825fdb565d539634714fadffacdf18ea4bc8b26R162

cdc/sink/flowcontrol/flow_control.go
```diff
@@ -110,6 +119,59 @@ func (c *TableFlowController) Release(resolvedTs uint64) {
...
+		size:     size,
+		rowCount: 1,
+	})
+	c.txnsWithSameCommitTs++
+	// mark the first data of new txnSizeEntry
+	msg.Row.SplitTxn = true
+	if c.txnsWithSameCommitTs >= batchSize {
+		// TODO(CharlesCheung): add batch resolve mechanism to mitigate oom problem
+		log.Debug("emit batch resolve event throw callback")
+	}
```
amyangfei, line 172: `c.txnsWithSameCommitTs` is a little confusing, it could be 
- either different transactions with same commit-ts but different start-ts
- or the same transactions but exceeds max rows or max size, which is the logic-2 in L151

CharlesCheung96: Therefore, the name is based on the common characteristics of both cases having the same commitTs. Perhaps we should add comments to illustrate this. BTW, do you have any good names to recommend.

amyangfei: What about `batchGroupCount`, and add comment to explain that batch group means a group of batch rows, all row changes have the same commit-ts.

asddongmen (approved):

amyangfei (approved):

Issue #5283, opened 2022-04-27
Tracking issue for Sink Module Refactoring


hi-rustin:
This is a tracking issue for Sink Module Refactoring.
- [x] [remove useless buffers](https://github.com/pingcap/tiflow/pull/5284)
- [x] [remove useless buffer sink](https://github.com/pingcap/tiflow/pull/5302)
- [x] https://github.com/pingcap/tiflow/issues/5425
- [x] https://github.com/pingcap/tiflow/pull/5461

PR #5284, opened 2022-04-27
pipeline/sink(ticdc): remove useless buffer from sinkNode

hi-rustin:

Issue Number: ref https://github.com/pingcap/tiflow/issues/5283

remove useless buffer from sinkNode.
It is a useless buffer, because currently tableSink has buffer itself.

### Check List 

Tests 

 - Unit test

Code changes

None

Side effects

None

Related changes

None

### Release note 

```release-note
None
```


cdc/processor/pipeline/sink.go
```diff
@@ -194,6 +192,7 @@ func (n *sinkNode) addRowToBuffer(ctx context.Context, event *model.PolymorphicE
 		return nil
 	}
 
+	rows := make([]*model.RowChangedEvent, 0, 2)
```
CharlesCheung96, line 195: Whether this slice will be allocated to heap and trigger frequent gc ?

cdc/processor/pipeline/sink.go
```diff
@@ -286,32 +283,17 @@ func splitUpdateEvent(updateEvent *model.PolymorphicEvent) (*model.PolymorphicEv
...
-		for i := range n.rowBuffer {
-			n.rowBuffer[i] = nil
-		}
-		n.rowBuffer = n.rowBuffer[:0]
-	}
-}
-
 // emitRowToSink emits the rows in rowBuffer to backend sink.
-func (n *sinkNode) emitRowToSink(ctx context.Context) error {
+func (n *sinkNode) emitRowToSink(ctx context.Context, rows ...*model.RowChangedEvent) error {
```
CharlesCheung96, line 287: Perhaps we can move this failpoint to tablesink to simplify the logic here. For example, remove emitRowToSink and directly emit RowChangedEvents to table sink.

cdc/processor/pipeline/sink.go
hi-rustin: Why might frequent gc be triggered here?

cdc/processor/pipeline/sink.go
hi-rustin: I'll trt it.

cdc/processor/pipeline/sink.go
overvenus: It is allocated to heap. Could you add a benchmark and show the performance impact? I think 10% regression may be acceptable given that the bottleneck is actually in sink IO not CPU.

cdc/processor/pipeline/sink.go
```diff
@@ -194,6 +192,7 @@ func (n *sinkNode) addRowToBuffer(ctx context.Context, event *model.PolymorphicE
 		return nil
 	}
```
asddongmen, line 194: I think it is better to remove  `addRowToBuffer` , since there is no buffer in `SinkNode` anymore. The `emitRowToSink` is enough.

cdc/processor/pipeline/sink.go
hi-rustin: Each round runs for 30s:
slice: BenchmarkEmitRowToSink-16 1202 74805793 ns/op
direct send: BenchmarkEmitRowToSink-16 914 113538113 ns/op

hi-rustin: I will remove it.

cdc/processor/pipeline/sink.go
hi-rustin: Addressed.

cdc/processor/pipeline/sink.go
hi-rustin: Move it into new emitRowToSink func.

asddongmen (approved):

CharlesCheung96 (approved):

PR #5295, opened 2022-04-27
syncer(dm): fix failed row skipped due to incorrect checkpoint flush

### LINKED ISSUE #5279: failed rows are skipped due to checkpoint flush
### What did you do?

most steps can reuse https://github.com/pingcap/tiflow/issues/5272, but need to set `checkpoint-flush-interval: 1` to make it eaiser to reproduce.**other case may reproduce this bug too, just make sure the failed row should trigger table checkpoint flush**

- create source is in pos mode
- create test.test1 in upstream
- start-task --remove-meta
- `alter table test1 add column c4 int;` in upstream
- `create table test2 (c int primary key);` in upstream or wait 30s, to flush checkpoints
- wait for 1 more seconds, so the next insert will trigger checkpoint flush(this will cause the pos of failed row flushed into table point)
- `insert into test1`. Now task will report error because downstream doesn't have column c4, but before the error, checkpoint is flushed:
https://github.com/pingcap/tiflow/blob/967060a8a8f63388e47540fe175902c99798892c/dm/syncer/syncer.go#L2466
- task is auto resumed, and the failed row is skipped because the pos of the failed row is smaller than current table checkpoint.
https://github.com/pingcap/tiflow/blob/967060a8a8f63388e47540fe175902c99798892c/dm/syncer/syncer.go#L2327

also this case may cause global point larger than table point:
```
+------------+-----------+----------+------------------+------------+
| id         | cp_schema | cp_table | binlog_name      | binlog_pos |
+------------+-----------+----------+------------------+------------+
| mysql-3306 |           |          | mysql-bin.000001 |        766 |
| mysql-3306 | test      | test2    | mysql-bin.000001 |        505 |
| mysql-3306 | test      | test1    | mysql-bin.000001 |        735 |
+------------+-----------+----------+------------------+------------+
```

```yaml
name: foo-task
task-mode: all


target-database:
  host: "127.0.0.1"
  port: 4000
  user: "root"
  password: "" # 如果密码不为空，则推荐使用经过 dmctl 加密的密文

mysql-instances:
  - source-id: "mysql-3306"
    block-allow-list:  "ba-rule1"
    syncer:
      checkpoint-flush-interval: 1
      enable-gtid: true
    filter-rules: ["filter-rule-1"]

block-allow-list:
  ba-rule1:
    do-dbs: ["test"]

filters:
  filter-rule-1:
    schema-pattern: "test"
    table-pattern: "test1"
    events: ["all ddl"]
    action: Ignore
```

for this case it should fail again on the failed row

### What did you see instead?

the row is skipped

### Versions of the cluster

current master

```
Release Version: v6.1.0-alpha-101-ga7bcd922a
Git Commit Hash: a7bcd922ae761e20525d847f46a4377047c46cef
Git Branch: HEAD
UTC Build Time: 2022-04-26 09:26:31
Go Version: go version go1.18 darwin/arm64
```

```console
```

D3Hunter:

Issue Number: close #5279

- checkpoint flush should wait all dml before it finished.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`.
```


GMHDBJD (approved):
LGTM

lance6716 (approved):
maybe my old thought is don't let handleJob calls handleJob itself, but I forget to flush the jobs in queue 😂 

I'm afraid handleJob calling handleJob itself will cause stack overflow in some bad cases, but flushIfOutdated is a common logic so it can also be put inside handleJob. No idea 🤔

PR #5300, opened 2022-04-27
changefeed(ticdc): add namespace info into ChangeFeedID

sdojjy:

Issue Number: ref #5301

make ChangefeedID to an struct
```
type ChangeFeedID struct {
	Namespace string
	ID        string
}
```

after this PR, all changefeed has an default namespace value, "default"
```
func DefaultNamespaceChangeFeedID(id string) ChangeFeedID {
	return ChangeFeedID{
		Namespace: DefaultNamespace,
		ID:        id,
	}
}
```

namesapce and ID  pair is unique in one ticdc cluster

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Please add a release note.
 `None`.
```


cdc/model/changefeed.go
```diff
@@ -31,8 +31,20 @@ import (
...
+const DefaultNamespace = "default"
+
 // ChangeFeedID is the type for change feed ID
-type ChangeFeedID = string
+type ChangeFeedID struct {
+	Namespace string
+	ID        string
+}
+
+func DefaultNamespaceChangeFeedID(id string) ChangeFeedID {
```
asddongmen, line 42: `DefaultChangeFeedID`  is a more concise name.

cdc/model/changefeed.go
```diff
@@ -31,8 +31,20 @@ import (
 	"go.uber.org/zap"
 )
 
+const DefaultNamespace = "default"
+
 // ChangeFeedID is the type for change feed ID
-type ChangeFeedID = string
+type ChangeFeedID struct {
+	Namespace string
```
asddongmen, line 38: Can you add a comment for Namespce?

sdojjy: done

cdc/contextutil/ctx.go
```diff
@@ -109,31 +111,31 @@ func IsOwnerFromCtx(ctx context.Context) bool {
 
 // ChangefeedIDFromCtx returns a changefeedID stored in the specified context.
 // It returns an empty string if there's no valid changefeed ID found.
```
asddongmen, line 113:
```suggestion
// It returns an empty model.changefeedID if there's no changefeedID found.
```

cdc/contextutil/ctx.go
```diff
@@ -109,31 +111,31 @@ func IsOwnerFromCtx(ctx context.Context) bool {
...
+func ChangefeedIDFromCtx(ctx context.Context) model.ChangeFeedID {
+	changefeedID, ok := ctx.Value(ctxKeyChangefeedID).(model.ChangeFeedID)
 	if !ok {
-		return ""
+		return model.ChangeFeedID{}
 	}
 	return changefeedID
 }
 
 // PutChangefeedIDInCtx returns a new child context with the specified changefeed ID stored.
```
asddongmen, line 122:
```suggestion
// PutChangefeedIDInCtx returns a new child context with the specified changefeedID stored.
```

asddongmen (approved):

Issue #5039, opened 2022-03-27
Fix typos


hi-rustin:
Fix typos in the code repository.

PR #5314, opened 2022-04-29
sink/mq/producer(ticdc): fix test typo

hi-rustin:

Issue Number: ref https://github.com/pingcap/tiflow/issues/5039

fix test typo.
It should be sasl-user instead of sasl-username.

### Check List 

Tests 

 - Unit test

Code changes

None

Side effects

None

Related changes

None

### Release note 

```release-note
None
```


overvenus (approved):

hi-rustin:
/hold

I think we can do better.

PR #5331, opened 2022-05-05
validator(dm): better error handling and handle panic

### LINKED ISSUE #5330: better error handling and handle panic

D3Hunter:

Issue Number: close #5330

- write ectd keys on `validation start` even when validator stage is running in etcd already, since validator may stop on error(we don't change the expected stage in this case), so we can resume.
- `validation status` can be used even when validator is stopped
- handle validator/worker-validate panic, so we can see it using `validation status`, and log other panics
- use `basedb` directly, and let `sql.db` handles reconnect stuff.
- add retry on retriable error, add timeout to all db queries.
  - we don't use retry in `dbconn` of syncer since it retry too much and contains a lot of metric stuff.
- use a separate db to do `validation status` since validator may stopped
- flush checkpoint and data on query event to make it easier to test
- fix bug on persist(may delete pending data from other sources)
- fix use the wrong lock on getTableStatusMap
- add more debug log
- revert desc of error `ErrWorkerNotPausedStage` and use another error code for validator to avoid expose validator.

### Check List 

Tests 

 - [x] Unit test
 - [x] Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`.
```


D3Hunter:
> write ectd keys on validation start even when validator stage is running in etcd already, since validator may stop on error(we don't change the expected stage in this case), so we can resume.

ut `TestOperateValidatorTask` fail because of this change, need to discuss with pm about its behavior when starting an auto-stopped validator, i.e. resuming.

dm/syncer/validator_checkpoint_test.go
```diff
@@ -140,21 +138,3 @@ func TestValidatorCheckpointPersist(t *testing.T) {
 	testFunc("")
 	testFunc("failed")
 }
-
-func TestCheckpointNotPanic(t *testing.T) {
```
D3Hunter, line 144: this case can be covered by any i t of validator, so remove it

dm/syncer/validate_worker.go
```diff
@@ -350,45 +367,28 @@ func (vw *validateWorker) validateInsertAndUpdateRows(rows []*rowValidationJob,
...
 	columnNames := make([]string, 0, len(cond.Columns))
 	for _, col := range cond.Columns {
 		columnNames = append(columnNames, dbutil.ColumnName(col.Name.O))
 	}
 	columns := strings.Join(columnNames, ", ")
 	rowsQuery := fmt.Sprintf("SELECT /*!40001 SQL_NO_CACHE */ %s FROM %s WHERE %s",
 		columns, cond.TargetTbl, cond.GetWhere())
-	rows, err := vw.conn.QuerySQL(tctx, rowsQuery, cond.GetArgs()...)
+	// query using sql.DB directly, BaseConn is more than what we need
+	rows, err := vw.db.QueryContext(tctx, rowsQuery, cond.GetArgs()...)
```
buchuitoudegou, line 382: Do we need some extra tricks to handle retryable errors (e.g. connection error) here?

dm/syncer/validator_checkpoint.go
```diff
@@ -352,15 +365,51 @@ func (c *validatorPersistHelper) incrRevision() {
...
+		return data, err
+	}
+
+	return data, err
 }
 
 func (c *validatorPersistHelper) loadCheckpoint(tctx *tcontext.Context) (*binlog.Location, error) {
-	sql := "select binlog_name, binlog_pos, binlog_gtid from " + c.checkpointTableName + " where source = ?"
-	rows, err := c.dbConn.QuerySQL(tctx, sql, c.cfg.SourceID)
+	query := "select binlog_name, binlog_pos, binlog_gtid from " + c.checkpointTableName + " where source = ?"
```
buchuitoudegou, line 411: Why is `select` and `where` not capital? (Not really a big deal though, it's okay to leave it here but other queries capitalize these keywords)

dm/syncer/validator_checkpoint.go
```diff
@@ -519,13 +568,14 @@ func (c *validatorPersistHelper) loadError(filterState pb.ValidateErrorState) ([
...
-	sql := "SELECT (id, source, src_schema_name, src_table_name, dst_schema_name, dst_table_name, data, dst_data, error_type, status, update_time) " +
+	query := "SELECT (id, source, src_schema_name, src_table_name, dst_schema_name, dst_table_name, data, dst_data, error_type, status, update_time) " +
 		"FROM " + c.errorChangeTableName + " WHERE source=?"
 	if filterState != pb.ValidateErrorState_InvalidErr {
-		sql += " AND status=?"
+		query += " AND status=?"
 		args = append(args, int(filterState))
 	}
-	rows, err = c.dbConn.QuerySQL(c.tctx, sql, args...)
+	// we do not retry, let user do it
```
buchuitoudegou, line 577: It's confusing that we have `retryer` in this module but not using it. Automatically retry might make life easier.

buchuitoudegou (commented):
Abandoning `DBConn` in the validator is reasonable because it does do extra metrics for the syncer when querying sql. But one unexpected drawback is we are losing the automatic retry wrapped in `DBConn`, which definitely harms users' experience or we have to do more stuff to make it retryable again.

Besides, this issue also implies that packages or functions that provision ability to query SQL are not well-encapsulated and unified across the whole projects. Do we need to sort it out?

D3Hunter:
> > write ectd keys on validation start even when validator stage is running in etcd already, since validator may stop on error(we don't change the expected stage in this case), so we can resume.
> 
> ut `TestOperateValidatorTask` fail because of this change, need to discuss with pm about its behavior when starting an auto-stopped validator, i.e. resuming.

after discuss with PM, `validation start`(different with `task start`) has two behavior:
- if validator hasn't started ever, it will create and then start the validator
- if validator has start ever, it'll resume the validator.

dm/syncer/validate_worker.go
D3Hunter: no need to retry inside, worker will retry it anyway. see defer in `validateTableChange`

dm/syncer/validator_checkpoint.go
D3Hunter: 个人习惯，no big deal

dm/syncer/validator_checkpoint.go
D3Hunter: those methods only called from dmctl, i.e. user called, which should return fast, and we don't met retryable error often, and if it really happens, user can check it.

for other methods which uses retry, it's because it's run in background, stop-then-resume makes life harder.

dm/pkg/conn/basedb.go
```diff
@@ -175,6 +177,64 @@ func (d *BaseDB) GetBaseConn(ctx context.Context) (*BaseConn, error) {
...
+			}
+			if _, err = tx.ExecContext(tctx.Ctx, q, args[i]...); err != nil {
+				return nil, perrors.Trace(err)
+			}
+		}
+		return nil, perrors.Trace(err)
+	}
+
+	if retryer == nil {
+		retryer = retry.NoRetry
```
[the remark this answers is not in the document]
D3Hunter: thought it maybe used without retry, seems we can add another method then. will remove it.

D3Hunter:
> Abandoning `DBConn` in the validator is reasonable because it does do extra metrics for the syncer when querying sql. But one unexpected drawback is we are losing the automatic retry wrapped in `DBConn`, which definitely harms users' experience or we have to do more stuff to make it retryable again.
> 
> Besides, this issue also implies that packages or functions that provision ability to query SQL are not well-encapsulated and unified across the whole projects. Do we need to sort it out?

`dbconn` should be removed and use `baseconn` directly during refactor i think. and `basedb` should supply api like `execContext`(added some of it in this pr) just like how `sql.DB/sql.Conn` did where they both supply similar api to exec or query, the different is whether you need to use the same connection or not.

dm/pkg/utils/common_test.go
```diff
@@ -257,3 +263,16 @@ func (s *testCommonSuite) TestNonRepeatStringsEqual(c *C) {
...
+	var wg sync.WaitGroup
+	wg.Add(1)
+	// to avoid data race since there's concurrent test case writing log.L()
+	l := log.Logger{Logger: zap.NewNop()}
+	go GoLogWrapper(l, func() {
+		defer wg.Done()
+		panic("should be captured")
+	})
+	wg.Wait()
+	require.True(t, true)
```
lance6716, line 277: I remember if a test panic, go will not treat it as pass. So this line is not needed

dm/syncer/validator_checkpoint.go
buchuitoudegou: lgtm

dm/pkg/conn/basedb.go
```diff
@@ -175,6 +177,61 @@ func (d *BaseDB) GetBaseConn(ctx context.Context) (*BaseConn, error) {
...
+	if tctx.L().Core().Enabled(zap.DebugLevel) {
+		tctx.L().Debug("exec context",
+			zap.String("query", utils.TruncateString(query, -1)),
+			zap.String("argument", utils.TruncateInterface(args, -1)))
+	}
+	return d.DB.ExecContext(tctx.Ctx, query, args...)
+}
+
+func (d *BaseDB) QueryContext(tctx *tcontext.Context, query string, args ...interface{}) (*sql.Rows, error) {
+	if tctx.L().Core().Enabled(zap.DebugLevel) {
```
Ehco1996, line 190: just a question: why this this check `if tctx.L().Core().Enabled(zap.DebugLevel)`, can we just use `tctx.L().Debug(`?

dm/pkg/conn/basedb.go
```diff
@@ -175,6 +177,61 @@ func (d *BaseDB) GetBaseConn(ctx context.Context) (*BaseConn, error) {
...
+		tctx.L().Debug("query context",
+			zap.String("query", utils.TruncateString(query, -1)),
+			zap.String("argument", utils.TruncateInterface(args, -1)))
+	}
+	return d.DB.QueryContext(tctx.Ctx, query, args...)
+}
+
+func (d *BaseDB) DoTxWithRetry(tctx *tcontext.Context, queries []string, args [][]interface{}, retryer retry.Retryer) error {
```
Ehco1996, line 198: maybe we can wrap all db/conns to `SQLWithRetry`, I see lots of same logic in many repos 🤣

https://github.com/pingcap/tidb/blob/f88dab92f9c5ca3f2225a1dab19ff9b4e6aaaaa1/br/pkg/lightning/common/util.go#L141

dm/syncer/data_validator.go
```diff
@@ -353,11 +375,11 @@ func (v *DataValidator) fillResult(err error, needLock bool) {
 
 func (v *DataValidator) errorProcessRoutine() {
 	defer v.errProcessWg.Done()
+	// todo: what if this routine panic? we still need to drain v.errChan.
```
lance6716, line 378: maybe routineWrapper can be changed to

```
func routineWrapper(workFn func(), recoverFn func(error))
```

in recoverFn we can call v.sendError, v.L, or for range v.errChan

dm/pkg/utils/common.go
```diff
@@ -363,3 +363,14 @@ func AdjustBinaryProtocolForDatum(ctx sessionctx.Context, data []interface{}, co
 	}
 	return ret, nil
 }
+
+// GoLogWrapper go routine wrapper, log error on panic.
+func GoLogWrapper(logger log.Logger, fn func()) {
+	defer func() {
+		if err := recover(); err != nil {
+			logger.Error("routine panic", zap.Any("err", err))
```
Ehco1996, line 371: do we need re panic after this log?

dm/syncer/data_validator.go
D3Hunter: > in recoverFn we can call v.sendError, v.L, or for range v.errChan

not easy to recover too. will remove this todo, let it panic, need to check and fix it anyway.

dm/pkg/conn/basedb.go
D3Hunter: it calls `utils.TruncateString(` twice before `tctx.L().Debug`, it's cheaper to check log level first.

dm/pkg/utils/common.go
D3Hunter: both are ok to me.

but without repanic, syncer can continue on validator panic.

D3Hunter:
ptal

dm/syncer/data_validator.go
```diff
@@ -366,14 +387,12 @@ func (v *DataValidator) errorProcessRoutine() {
...
-		return nil
-	case cmp == 0:
+	if cmp >= 0 && !v.reachedSyncer.Load() {
 		v.reachedSyncer.Store(true)
+		v.L.Info("validator progress reached syncer")
+	}
+	if cmp <= 0 {
 		return nil
-	default:
-		v.reachedSyncer.Store(true)
```
lance6716, line 376: again I forget what's the meaning of `reachedSyncer` 😂 and seems if we lucky always have `currLoc < syncLoc`, `reachedSyncer` will always be false

dm/syncer/data_validator.go
```diff
@@ -549,6 +568,22 @@ func (v *DataValidator) doValidate() {
...
+			query := string(ev.Query)
+			if query == "COMMIT" || query == "BEGIN" {
+				break
+			}
+			err = locationForFlush.SetGTID(ev.GSet)
+			if err != nil {
+				v.sendError(terror.Annotate(err, "failed to set gtid"))
+				return
+			}
+			if err = v.persistCheckpointAndData(locationForFlush); err != nil {
```
lance6716, line 582: why this is persistCheckpointAndData but other branches are checkAndPersistCheckpointAndData

lance6716 (commented):
finished a half of files. will review later

dm/syncer/data_validator.go
D3Hunter: only when `reachedSyncer`=true, validator will start mark rows failed to validate as error rows to reduce chance of false-positive.

>  if we lucky always have currLoc < syncLoc, reachedSyncer will always be false

you're right. we need to make sure `reachedSyncer=true` at some time. will leave a todo and fix in next pr.

dm/syncer/data_validator.go
D3Hunter: to make sure checkpoint is persisted on ddl, so i t can check it. but seems we can use checkAndPersistCheckpointAndData and set a very small interval.

dm/syncer/data_validator.go
D3Hunter: `reachedSyncer` is not clear for its usage, renamed in later pr.

D3Hunter:
ping @Ehco1996 @buchuitoudegou

buchuitoudegou (approved):
lgtm

D3Hunter:
ptal

D3Hunter:
ping

Ehco1996 (approved):
LGTM when all comments are resolved

D3Hunter:
> LGTM when all comments are resolved

some outdated comments or comments need further explain are resolved, any other "better to or must change" comments?

Ehco1996:
> > LGTM when all comments are resolved
> 
> some outdated comments or comments need further explain are resolved, any other "better to or must change" comments?

ooo I see, I think it's ready to merge now

D3Hunter:
> finished a half of files. will review later

ping @lance6716

lance6716:
> > finished a half of files. will review later
> 
> ping @lance6716

still doesn't have time 😂 since it's LGT2 we can merge it now.

PR #5340, opened 2022-05-06
syncer: support start from empty gtid config

### LINKED ISSUE #3731: can't start a GTID-based incremental task when upstream GTID is empty
## Bug Report

upstream gtid is empty, and binlog-gtid in meta field in task file is also empty

start successfully

3. What did you see instead?

query-status shows "invalid position (mysql-bin.000001, 4) or GTID not enabled in upstream"

4. Versions of the cluster

    - DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):

        ```
        v2.0.5
        ```

GMHDBJD:

Issue Number: close #3731 

mariadb could use 0-0-0 as empty gtid config
- support user to use uuid:0 as mysql empty gtid config

### Check List 

Tests 
 - Integration test
 - Manual test (mariadb)

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
Support start an incremental task with nil MySQL GTID set.
```

### DIFF
--- a/dm/pkg/gtid/gtid.go
+++ b/dm/pkg/gtid/gtid.go
@@ -100,7 +100,7 @@ func ParserGTID(flavor, gtidStr string) (Set, error) {
 
 // check whether a gtid set is nil(start sync from start)
 // mysql: uuid:0
-// mariadb: x-y-0(no need to handle)
+// mariadb: 0-0-0(no need to handle)
 func IsNilMySQLGTIDSet(gStr string) bool {
 	sp := strings.Split(gStr, ",")
 	if len(sp) != 1 {
--- a/dm/pkg/gtid/gtid.go
+++ b/dm/pkg/gtid/gtid.go
@@ -14,9 +14,13 @@
 package gtid
 
 import (
+	"strings"
+
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/pingcap/errors"
+	"go.uber.org/zap"
 
+	"github.com/pingcap/tiflow/dm/pkg/log"
 	"github.com/pingcap/tiflow/dm/pkg/terror"
 )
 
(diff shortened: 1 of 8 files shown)


dm/pkg/gtid/gtid.go
```diff
@@ -80,14 +84,37 @@ func ParserGTID(flavor, gtidStr string) (Set, error) {
...
 	}
+
 	err = m.Set(gtid)
 	return m, err
 }
 
+// check whether a gtid set is nil(start sync from start)
+// mysql: uuid:0
+// mariadb: x-y-0(no need to handle)
+func IsNilMySQLGTIDSet(gStr string) bool {
```
buchuitoudegou, line 104: Should we add a unit test for this function either?

dm/relay/local_reader_test.go
```diff
@@ -613,7 +613,7 @@ func (t *testReaderSuite) TestStartSyncByGTID(c *C) {
 		{
 			"ba8f633f-1f15-11eb-b1c7-0242ac110002",
 			"ba8f633f-1f15-11eb-b1c7-0242ac110002.000001",
-			"ba8f633f-1f15-11eb-b1c7-0242ac110002:0",
+			"ba8f633f-1f15-11eb-b1c7-0242ac110002:1",
```
buchuitoudegou, line 616: Should we just add a test case rather than replace it? Because it seems `IsNilMySQLGTIDSet` is used to handle such cases (i.e. xxx:0).

dm/pkg/gtid/gtid.go
```diff
@@ -80,14 +84,37 @@ func ParserGTID(flavor, gtidStr string) (Set, error) {
...
 		return nil, terror.ErrNotSupportedFlavor.Generate(flavor)
 	}
+
 	err = m.Set(gtid)
 	return m, err
 }
 
+// check whether a gtid set is nil(start sync from start)
+// mysql: uuid:0
+// mariadb: x-y-0(no need to handle)
```
lance6716, line 103: did you test that mariaDB supports x-y-0?

dm/tests/incremental_mode/run.sh
```diff
@@ -142,16 +120,12 @@ function run() {
...
-	sed -i "s/task-mode-placeholder/incremental/g" $WORK_DIR/dm-task.yaml
-	sed -i "s/binlog-name-placeholder-1//g" $WORK_DIR/dm-task.yaml
-	sed -i "s/binlog-pos-placeholder-1//g" $WORK_DIR/dm-task.yaml
-	sed -i "s/binlog-gtid-placeholder-1/$gtid1/g" $WORK_DIR/dm-task.yaml
-
-	sed -i "s/binlog-name-placeholder-2/$name2/g" $WORK_DIR/dm-task.yaml
-	sed -i "s/binlog-pos-placeholder-2/$pos2/g" $WORK_DIR/dm-task.yaml
-	sed -i "s/binlog-gtid-placeholder-2/$gtid2/g" $WORK_DIR/dm-task.yaml
+	worker1_run_source_1=$(sed "s/$SOURCE_ID1/$SOURCE_ID1\n/g" $WORK_DIR/worker1/log/dm-worker.log | grep -c "$SOURCE_ID1") || true
+	echo "start task in incremental mode with nil gtid/pos"
```
lance6716, line 124:
```suggestion
	echo "start task in incremental mode with zero gtid/pos"
```

dm/tests/incremental_mode/run.sh
```diff
@@ -142,16 +120,12 @@ function run() {
...
+	binlog_name=($(get_binlog_name $MYSQL_HOST1 $MYSQL_PORT1))
```
lance6716, line 125: this should be source2?

dm/relay/local_reader_test.go
GMHDBJD: This gtid is used to generate binlog event, after this pr, xxx:0 will be transformed to nil, but binlog pkg could not generate event with nil gtid.

dm/pkg/gtid/gtid.go
GMHDBJD: add in 044a3e3

dm/pkg/gtid/gtid.go
GMHDBJD: test with 0-0-0, update comment in 044a3e3

lance6716 (approved):

D3Hunter (approved):

Issue #5334, opened 2022-05-05
another implementation of schema tracker


lance6716:
### Is your feature request related to a problem?

Currently schema tracker is an embedded TiDB with unistore as its storage. Before v2.0.7, the storage is mocktikv. It has many problems:

- embedded TiDB may has some bugs: from an issue, user reports that the unistore folder eats up to 18GB space. Here is another issue from a different user https://github.com/pingcap/tiflow/issues/4187
   Although we can change the unstore folder to another place, and clean the folder after the replication task is stopped, some user may still don't have a good place to support a long-running task. In v2.0.1 when the storage is mocktikv, it uses memory rather than disk to serve the embedded TiDB, we also received huge memory usage issue on that version.
- configuration of embedded TiDB: The embedded TiDB should be well configurated about so it can support any table structure that downstream supports. For example, in https://github.com/pingcap/tiflow/issues/5315 user changed the max-index-length but embedded TiDB didn't change it. And new features like clustered index has different behaviours about ALTER PRIMARY KEY.
- Other problems that is caused by TiDB not being designed as a library, like https://github.com/pingcap/tiflow/issues/3402, https://github.com/pingcap/tidb/pull/28027, https://github.com/pingcap/tidb/issues/18714, and many ugly patches on TiDB configurating to turn off some functionalities https://github.com/pingcap/tiflow/blob/9bcebc1f23d29fa739e17c5c65a19b6a7f698f8e/dm/pkg/schema/tracker.go#L111-L119, https://github.com/pingcap/tiflow/blob/9bcebc1f23d29fa739e17c5c65a19b6a7f698f8e/dm/pkg/schema/tracker.go#L170-L172

### Describe the feature you'd like

debezium uses the result of SQL parser to directly mutate the column map and other structures.

https://github.com/debezium/debezium/blob/9f7ede0e0695f012c6c4e715e96aed85eecf6b5f/debezium-core/src/main/java/io/debezium/relational/TableEditorImpl.java

https://github.com/debezium/debezium/blob/b78e3183997b3d4ef044c0d13d9503829124fc92/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/AlterTableParserListener.java#L284

DM can organize the TableInfo on its own, such as keep a map of `{SchemaName -> {SchemaInfo, {TableName -> TableInfo}}`. 
[...]

PR #5341, opened 2022-05-06
tracker(dm): add task name to unistore folder name

lance6716:

Issue Number: ref #5334

sometimes unistore folder is abnormally big, we add task name to the folder name to know which task it belongs.

Also, clean that folder for unit tests by `Close()` 

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### Release note 

```release-note
 `None`.
```

### DIFF
--- a/dm/pkg/schema/tracker.go
+++ b/dm/pkg/schema/tracker.go
@@ -245,7 +245,7 @@ func newTmpFolderForTracker(task string) (string, error) {
 	var buf strings.Builder
 	for _, c := range task {
 		if _, ok := specialChars[c]; ok {
-			buf.WriteByte('_')
+			fmt.Fprintf(&buf, "%%%02X", c)
 		} else {
 			buf.WriteRune(c)
 		}
--- a/dm/pkg/schema/tracker.go
+++ b/dm/pkg/schema/tracker.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"strings"
 	"sync"
@@ -234,23 +235,8 @@ func NewTracker(ctx context.Context, task string, sessionCfg map[string]string, 
 	}, nil
 }
 
-var specialChars = map[rune]struct{}{
-	'`': {}, '~': {}, '!': {}, '@': {}, '#': {}, '$': {}, '%': {}, '^': {}, '&': {},
-	'*': {}, '(': {}, ')': {}, '-': {}, '_': {}, '=': {}, '+': {}, '\\': {}, '|': {},
-	'[': {}, '{': {}, ']': {}, '}': {}, ':': {}, ';': {}, '\'': {}, '"': {}, '<': {},
-	'>': {}, ',': {}, '.': {}, '?': {}, '/': {}, ' ': {},
-}
-
 func newTmpFolderForTracker(task string) (string, error) {
-	var buf strings.Builder
-	for _, c := range task {
-		if _, ok := specialChars[c]; ok {
-			fmt.Fprintf(&buf, "%%%02X", c)
-		} else {
-			buf.WriteRune(c)
-		}
-	}
-	return ioutil.TempDir("./", buf.String()+"-tracker")
+	return ioutil.TempDir("./", url.PathEscape(task)+"-tracker")
 }
 
 // Exec runs an SQL (DDL) statement.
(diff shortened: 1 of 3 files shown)


dm/pkg/schema/tracker.go
```diff
@@ -149,7 +149,7 @@ func NewTracker(ctx context.Context, task string, sessionCfg map[string]string,
 		}
 	}
 
-	storePath, err = ioutil.TempDir("./", "schema-tracker")
+	storePath, err = ioutil.TempDir("./", task+"-schema-tracker")
```
lichunzhu, line 152: What about using a sub-directory?

BTW, I think we should add a data-dir configuration for dm-worker. We have relay-dir,dump-dir, and schema-tracker file directories now, but all the default values of them are the dm-worker binary's dir. For TiUP data-dir and deploy-dir are separated which looks a bit weird. 

Maybe we can do this in the next PR.

lichunzhu (commented):
Rest LGTM

dm/pkg/schema/tracker.go
lance6716: Will do it later, if we can't find the root cause we will change the implementation, so that configuration will be deprecated soon.

D3Hunter (approved):

lichunzhu (approved):
LGTM

lance6716:
/hold

failed to create schema tracker\" RawCause:\"mkdirtemp ab?c/b%c\\303\\213d-schema-tracker: pattern contains path separator\

lance6716:
PTAL for https://github.com/pingcap/tiflow/pull/5341/commits/b3afb8c7d8674b714a0b05664eb8b23fe3f08d1b

@D3Hunter @lichunzhu

lichunzhu:
> PTAL for [b3afb8c](https://github.com/pingcap/tiflow/commit/b3afb8c7d8674b714a0b05664eb8b23fe3f08d1b)
> 
> @D3Hunter @lichunzhu

Will this cause different tasks to have the same schema-tracker file name? What about using dumpling's template https://github.com/pingcap/tidb/blob/master/dumpling/export/prepare.go#L68

lance6716:
> > PTAL for [b3afb8c](https://github.com/pingcap/tiflow/commit/b3afb8c7d8674b714a0b05664eb8b23fe3f08d1b)
> > @D3Hunter @lichunzhu
> 
> Will this cause different tasks to have the same schema-tracker file name? What about using dumpling's template https://github.com/pingcap/tidb/blob/master/dumpling/export/prepare.go#L68

ioutil.TempDir will append different suffix for them.

https://github.com/pingcap/tidb/blob/master/dumpling/export/prepare.go#L68 it's too complicated, I can't understand what it's doing :cry: Can you give an example how to use it for a string input?

lichunzhu:
> ioutil.TempDir will append different suffix for them, so we can only roughly tell them apart.
> 
> https://github.com/pingcap/tidb/blob/master/dumpling/export/prepare.go#L68 it's too complicated, I can't understand what it's doing 😢 Can you give an example how to use it for a string input?

https://github.com/pingcap/dumpling/issues/131#issuecomment-674718065
It can escape the special characters in the object names.
e.g.: `table-partition/TICASE-2024` will be replaced by `table-partition%2FTICASE-2024`.
It seems that we don't need this name for debugging, so I think the current implementation is okay.

lance6716:
https://github.com/pingcap/tiflow/pull/5341/commits/531cbbd99bc25584c9540d31a3575a76f4f2911b

@lichunzhu PTAL

lichunzhu (approved):
LGTM

dm/pkg/schema/tracker.go
```diff
@@ -234,6 +234,25 @@ func NewTracker(ctx context.Context, task string, sessionCfg map[string]string,
...
 }
 
+var specialChars = map[rune]struct{}{
+	'`': {}, '~': {}, '!': {}, '@': {}, '#': {}, '$': {}, '%': {}, '^': {}, '&': {},
+	'*': {}, '(': {}, ')': {}, '-': {}, '_': {}, '=': {}, '+': {}, '\\': {}, '|': {},
+	'[': {}, '{': {}, ']': {}, '}': {}, ':': {}, ';': {}, '\'': {}, '"': {}, '<': {},
+	'>': {}, ',': {}, '.': {}, '?': {}, '/': {}, ' ': {},
+}
+
+func newTmpFolderForTracker(task string) (string, error) {
```
D3Hunter, line 244: why not use `url.PathEscape`

PR #5342, opened 2022-05-06
sink(ticdc): add batch field to resolveTs

### LINKED ISSUE #5398: Add batch field to resolveTs

CharlesCheung96:

Issue Number: close #5398

1. Add batch field to resolveTs
2. Added batch Resolved semantics support to Sink's `FlushRowChangedEvents` interface:

      -  With `resolved.Batch == false`, TiCDC guarantees that all the Events whose commitTs is less than or equal to `resolved.Ts` are sent to Sink through `EmitRowChangedEvents`.
      -  With `resolved.Batch == true`, TiCDC guarantees that all events whose commitTs is less than 'resolved.Ts' are sent to Sink.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/entry/mounter.go
+++ b/cdc/entry/mounter.go
@@ -101,7 +101,7 @@ func NewMounter(schemaStorage SchemaStorage,
 // this method could block indefinitely if the DDL puller is lagging.
 func (m *mounterImpl) DecodeEvent(ctx context.Context, pEvent *model.PolymorphicEvent) error {
 	m.metricTotalRows.Inc()
-	if pEvent.RawKV.OpType == model.OpTypeResolved {
+	if pEvent.IsResolved() {
 		return nil
 	}
 	start := time.Now()
--- a/cdc/model/mounter.go
+++ b/cdc/model/mounter.go
@@ -18,6 +18,8 @@ type PolymorphicEvent struct {
 	StartTs uint64
 	// Commit or resolved TS
 	CRTs uint64
+	// Identify whether the resolved event is in batch mode.
+	Mode ResolvedMode
 
 	RawKV *RawKVEntry
 	Row   *RowChangedEvent
@@ -44,7 +46,7 @@ func NewPolymorphicEvent(rawKV *RawKVEntry) *PolymorphicEvent {
 	}
 }
 
-// NewResolvedPolymorphicEvent creates a new PolymorphicEvent with the resolved ts
+// NewResolvedPolymorphicEvent creates a new PolymorphicEvent with the resolved ts.
 func NewResolvedPolymorphicEvent(regionID uint64, resolvedTs uint64) *PolymorphicEvent {
 	return &PolymorphicEvent{
 		CRTs:  resolvedTs,
@@ -58,13 +60,24 @@ func (e *PolymorphicEvent) RegionID() uint64 {
 	return e.RawKV.RegionID
 }
 
+// IsResolved returns true if the event is resolved. Note that this function can
+// only be called when `RawKV != nil`.
+func (e *PolymorphicEvent) IsResolved() bool {
+	return e.RawKV.OpType == OpTypeResolved
+}
+
+// IsBatchResolved returns true if the event is batch resolved event.
+func (e *PolymorphicEvent) IsBatchResolved() bool {
+	return e.IsResolved() && e.Mode == BatchResolvedMode
+}
+
 // ComparePolymorphicEvents compares two events by CRTs, Resolved, StartTs, Delete/Put order.
 // It returns true if and only if i should precede j.
 func ComparePolymorphicEvents(i, j *PolymorphicEvent) bool {
 	if i.CRTs == j.CRTs {
-		if i.RawKV.OpType == OpTypeResolved {
+		if i.IsResolved() {
 			return false
-		} else if j.RawKV.OpType == OpTypeResolved {
+		} else if j.IsResolved() {
 			return true
 		}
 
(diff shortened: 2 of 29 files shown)


nongfushanquan (approved):

cdc/model/mounter.go
```diff
@@ -13,17 +13,19 @@
...
 package model
 
-// PolymorphicEvent describes an event can be in multiple states
+// PolymorphicEvent describes an event can be in multiple states.
 type PolymorphicEvent struct {
 	StartTs uint64
 	// Commit or resolved TS
 	CRTs uint64
+	// Identify whether the resolved event is in batch mode.
+	Batch bool
```
hi-rustin, line 22: I don't know if mode or batch is better?

cdc/processor/pipeline/sink.go
```diff
@@ -71,7 +71,8 @@ type sinkNode struct {
 	status  TableStatus
 	tableID model.TableID
 
-	resolvedTs   model.Ts
+	// atomic oprations for model.ResolvedTs
+	resolvedTs   atomic.Value
```
hi-rustin, line 75: Prefer it to be the clear type. https://rust.godbolt.org/z/5qKjjPx64 Also it seems to need more instructions. And there is an additional type assertion.

cdc/model/mounter.go
hi-rustin: Also we probably need more documentation on the batch pattern in the code repository, maybe a design document, maybe comments.

cdc/model/mounter.go
```diff
@@ -71,3 +83,19 @@ func ComparePolymorphicEvents(i, j *PolymorphicEvent) bool {
...
+}
+
+// NewResolvedTs creates a new ResolvedTs.
+func NewResolvedTs(t uint64) ResolvedTs {
+	return ResolvedTs{Ts: t, Batch: false}
+}
+
+// NewBatchResolvedTs creates a ResolvedTs with a given batch type.
+func NewBatchResolvedTs(t uint64, b bool) ResolvedTs {
```
3AceShowHand, line 99: It looks that `NewBatchResolvedTs` can directly return a `ResolvedTs` with `Batch` is `True`.

cdc/model/mounter.go
```diff
@@ -49,13 +51,23 @@ func (e *PolymorphicEvent) RegionID() uint64 {
...
+
 // ComparePolymorphicEvents compares two events by CRTs, Resolved, StartTs, Delete/Put order.
 // It returns true if and only if i should precede j.
 func ComparePolymorphicEvents(i, j *PolymorphicEvent) bool {
 	if i.CRTs == j.CRTs {
-		if i.RawKV.OpType == OpTypeResolved {
+		if i.IsResolved() {
 			return false
-		} else if j.RawKV.OpType == OpTypeResolved {
+		} else if j.IsResolved() {
```
3AceShowHand, line 70: It looks that this `else` is unnecessary.

cdc/model/mounter.go
```diff
@@ -71,3 +83,19 @@ func ComparePolymorphicEvents(i, j *PolymorphicEvent) bool {
 	}
 	return i.CRTs < j.CRTs
 }
+
+// ResolvedTs is the resolved timestamp of sink module.
+type ResolvedTs struct {
+	Ts    uint64
+	Batch bool
```
3AceShowHand, line 90: It looks like we can add some comments here, what is the usage of `Batch`, and why do we need it.

cdc/model/mounter.go
CharlesCheung96: This is commented on the [`FlushRowChangedEvents `](https://github.com/pingcap/tiflow/pull/5342/files#diff-d9a3c1b05954cf385789af0a9c06058c18499bebb51676063237d144af0a61daR57) of sink

cdc/model/mounter.go
CharlesCheung96: Else is necessary, otherwise this function will return `false` when `i.IsResolved()==false` and `j.IsResolved()==true`.

cdc/model/mounter.go
CharlesCheung96: > I don't know if mode or batch is better?
This field may be accessed directly from the outside. Ref: https://github.com/pingcap/tiflow/pull/5342/files#diff-176f4cdb8edddd8b6ef31640086f43f3b7d76d969f8a69db5686e87340a03e3dR304

cdc/entry/mounter.go
```diff
@@ -101,7 +101,7 @@ func NewMounter(schemaStorage SchemaStorage,
 // this method could block indefinitely if the DDL puller is lagging.
 func (m *mounterImpl) DecodeEvent(ctx context.Context, pEvent *model.PolymorphicEvent) error {
 	m.metricTotalRows.Inc()
-	if pEvent.RawKV.OpType == model.OpTypeResolved {
+	if pEvent.IsResolved() {
```
amyangfei, line 104: This is a hot path, and `IsResolved` adds one more check for each event.

cdc/processor/pipeline/sink.go
```diff
@@ -291,24 +293,26 @@ func (n *sinkNode) HandleMessage(ctx context.Context, msg pmessage.Message) (boo
 	switch msg.Tp {
 	case pmessage.MessageTypePolymorphicEvent:
 		event := msg.PolymorphicEvent
-		if event.RawKV.OpType == model.OpTypeResolved {
+		if event.IsResolved() {
```
amyangfei, line 296: ditto

cdc/entry/mounter.go
CharlesCheung96: [Nil-value judgment is removed](https://github.com/pingcap/tiflow/pull/5342/files#diff-15a4777aa4f184040a91f7a61bde28f3d0390fc17709257cec96efe5722bd64bR64), PTAL.

cdc/model/mounter.go
CharlesCheung96: This is commented on the [FlushRowChangedEvents ](https://github.com/pingcap/tiflow/pull/5342/files#diff-d9a3c1b05954cf385789af0a9c06058c18499bebb51676063237d144af0a61daR57)of sink module.

amyangfei:
This PR doesn't change any logic, only adds the batch field, is it right?

CharlesCheung96:
> This PR doesn't change any logic, only adds the batch field, is it right?

Yes

amyangfei (approved):

hi-rustin (approved):
LGTM

PR #5362, opened 2022-05-09
*(ticdc): remove scheduler v1

### LINKED ISSUE #5361: Remove TiCDC scheduler v1
Scheduler V1 is no longer used.

overvenus:

Issue Number: close #5361

* Remove scheduler v1
* Remove TaskStatus usage in owner/processor (no longer used)
* Clean up TaskPosition usage in owner/processor (only kept `RunningError`)

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

### Release note 

```release-note
None
```


cdc/model/owner.go
```diff
@@ -208,12 +212,12 @@ func (i *TableReplicaInfo) Clone() *TableReplicaInfo {
 	return &clone
 }
 
-// TaskStatus records the task information of a capture
+// TaskStatus records the task information of a capture.
+//
+// Deprecated: only used in API. TODO: remove API usage.
```
3AceShowHand, line 217: It looks the whole `TaskStatus` is labeled as `deprecated`.

It has 3 different usage:
1. for old scheduler to track scheduling information.
2. for new ticdc to keep compatible with the old version.
3. for OpenAPI, return table's information.

I think the first one usage is deprecated, the last 2 is still useful.

cdc/model/owner.go
```diff
@@ -388,7 +303,6 @@ func (status *ChangeFeedStatus) Unmarshal(data []byte) error {
 
 // ProcInfoSnap holds most important replication information of a processor
 type ProcInfoSnap struct {
```
3AceShowHand, line 305: should `Tables` bne returned or not?

cdc/model/owner.go
```diff
@@ -388,7 +303,6 @@ func (status *ChangeFeedStatus) Unmarshal(data []byte) error {
...
-	CfID      ChangeFeedID                  `json:"changefeed-id"`
-	CaptureID string                        `json:"capture-id"`
-	Tables    map[TableID]*TableReplicaInfo `json:"-"`
```
3AceShowHand, line 393: It looks that `Tables` here is useless before? Have you ever test this?

3AceShowHand:
can we change the signature of `scheduler.Tick`, Just like `scheduler_dispatcher`.
```
Tick(
		ctx context.Context,
		checkpointTs model.Ts
		currentTables []model.TableID,
		captures map[model.CaptureID]*model.CaptureInfo,
	) (newCheckpointTs, newResolvedTs model.Ts, err error)
```

cdc/model/owner.go
overvenus: > 2. for new ticdc to keep compatible with the old version.

I don't think so, we only need to decode keys left by old version, not values.

> 3. for OpenAPI, return table's information.

We will replace them with other structs defined in http_model.go later.

cdc/model/owner.go
overvenus: No usage in ticdc code base.

overvenus:
> can we change the signature of `scheduler.Tick`, Just like `scheduler_dispatcher`.
> 
> ```
> Tick(
> 		ctx context.Context,
> 		checkpointTs model.Ts
> 		currentTables []model.TableID,
> 		captures map[model.CaptureID]*model.CaptureInfo,
> 	) (newCheckpointTs, newResolvedTs model.Ts, err error)
> ```

Good idea, I'd like to change it in a later PR.

sdojjy (approved):

liuzix (approved):

PR #5364, opened 2022-05-09
DM/Openapi: ignore can not connect error when force delete task

### LINKED ISSUE #5363: DM/Openapi: ignore can not connect error when force delete task

Openapi which is force delete task will return error when can not connect downstream.

ignore can not connect error, and return normal.

_No response_

_No response_

WizardXiao:

Issue Number: close #5363

add check when delete task.
### Check List 

Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


dm/dm/master/openapi_controller.go
```diff
@@ -443,14 +444,29 @@ func (s *Server) deleteTask(ctx context.Context, taskName string, force bool) er
 	}
 	defer release()
 
+	ignoreCannotConnectError := func(err error) bool {
+		if err == nil {
+			return true
+		}
+		if force && strings.Contains(err.Error(), "connect: connection refused") {
```
Ehco1996, line 451: how about use 
`errors.Is(err, syscall.ECONNREFUSED)`

Ehco1996 (commented):
rest LGTM

dm/dm/master/openapi_controller.go
WizardXiao: i tried `errors.Is(err, syscall.ECONNREFUSED)` and run inegration test, but it can't ignore `connection refused`. I still don't know the root cause, maybe i will find it out later

Ehco1996: i think this is because this error is  wrapped by `terror` https://github.com/pingcap/tiflow/blob/e91554504e44d3b5faa1b3c8d777b3306dc964c3/dm/pkg/conn/basedb.go#L117

ok not change here

Ehco1996 (approved):
LGTM

WizardXiao:
please help me take a look, tanks. @GMHDBJD @dsdashun

dm/dm/master/openapi_controller.go
```diff
@@ -443,14 +444,29 @@ func (s *Server) deleteTask(ctx context.Context, taskName string, force bool) er
...
-		return adjustErr
+		if !ignoreCannotConnectError(adjustErr) {
+			return adjustErr
+		}
 	}
 	metaSchema := *task.MetaSchema
 	err = s.removeMetaData(ctx, taskName, metaSchema, toDBCfg)
 	if err != nil {
-		return terror.Annotate(err, "while removing metadata")
+		if !ignoreCannotConnectError(err) {
```
buchuitoudegou, line 467: I understand forcing to delete tasks shouldn't have returned errors. But what if the user restarts the same task again but the metadata still exists?

buchuitoudegou (approved):
LGTM but probably result in https://github.com/pingcap/tiflow/issues/5409

dm/dm/master/openapi_controller.go
WizardXiao: i make a discussion note here. 
when user recreates the same task, and the downstream recovers again, it will still have the metadata, then it need remove metadata when start task or manually delete metadata. 
after disscuss with @sunzhaoyang, this is rare, we can just print warning log. but it still has risk，and most of openapi's interface has no warning message when it can run successfully. so i will pose a issue to note this.

Issue #5356, opened 2022-05-07
Goroutine leak in mysql sink


CharlesCheung96:
### Which jobs are flaking?

leak_test

### Which test(s) are flaking?

mysql_test

### Jenkins logs or GitHub Actions link

```log
[2022-05-07T06:20:02.424Z] PASS

[2022-05-07T06:20:02.424Z] goleak: Errors on successful test run: found unexpected goroutines:

[2022-05-07T06:20:02.424Z] [Goroutine 349 in state semacquire, with sync.runtime_Semacquire on top of the stack:

[2022-05-07T06:20:02.424Z] goroutine 349 [semacquire]:

[2022-05-07T06:20:02.424Z] sync.runtime_Semacquire(0xc000c8e3b0?)

[2022-05-07T06:20:02.424Z] 	/usr/local/go/src/runtime/sema.go:56 +0x25

[2022-05-07T06:20:02.424Z] sync.(*WaitGroup).Wait(0xc000c8e3a8)

[2022-05-07T06:20:02.424Z] 	/usr/local/go/src/sync/waitgroup.go:136 +0x9c

[2022-05-07T06:20:02.425Z] github.com/pingcap/tiflow/cdc/sink/mysql.(*mysqlSink).notifyAndWaitExec.func1()

[2022-05-07T06:20:02.425Z] 	/home/jenkins/agent/workspace/cdc_ghpr_leak_test/go/src/github.com/pingcap/tiflow/cdc/sink/mysql/mysql.go:456 +0x45

[2022-05-07T06:20:02.425Z] created by github.com/pingcap/tiflow/cdc/sink/mysql.(*mysqlSink).notifyAndWaitExec

[2022-05-07T06:20:02.425Z] 	/home/jenkins/agent/workspace/cdc_ghpr_leak_test/go/src/github.com/pingcap/tiflow/cdc/sink/mysql/mysql.go:455 +0x31b

[2022-05-07T06:20:02.425Z] ]

[2022-05-07T06:20:02.425Z] FAIL	github.com/pingcap/tiflow/cdc/sink/mysql	8.547s
```

https://ci2.pingcap.net/blue/organizations/jenkins/cdc_ghpr_leak_test/detail/cdc_ghpr_leak_test/25186/pipeline

### Anything else we need to know

- Does this test exist for other branches as well?

- Has there been a high frequency of failure lately?

zhaoxinyu:
[...]

PR #5374, opened 2022-05-10
sink(ticdc): ignore sync.runtime_Semacquire in mysql leak test

CharlesCheung96:

Issue Number: ref #5356

1. Temporarily ignore sync.runtime_Semacquire in mysql leak test.
2. Enable t.Parallel for cdc/sink/mysql, note that some tests which modify global variables cannot be run in parallel.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
`None`.
```


zhaoxinyu (approved):

cdc/sink/mysql/causality_test.go
```diff
@@ -69,6 +71,7 @@ func TestCausality(t *testing.T) {
 }
 
 func TestGenKeys(t *testing.T) {
+	defer leakutil.VerifyNone(t)
```
maxshuang, line 74: can u explain why we can use t.Parallel() and VerifyTestMain() in mysql related package?

CharlesCheung96: VerifyNone is used to identify which tests have coroutine leaks, ref: https://github.com/pingcap/tiflow/issues/5356#issuecomment-1123213475. The root cause is that mysql Sink was not reliably closed. After the PR was fixed, we were able to open `t.Parallel`

asddongmen (approved):

maxshuang (approved):

Issue #5288, opened 2022-04-27
Add upstream to manager all upstream related resources.


asddongmen:
### Is your feature request related to a problem?

as title.

### Describe the feature you'd like

as title.

asddongmen:
done

PR #5264, opened 2022-04-25
sink/flowcontrol(ticdc): make tests run and refactor code

hi-rustin:
### What problem does this PR solve?

Issue Number: ref https://github.com/pingcap/tiflow/issues/5138

### What is changed and how it works?

- make tests run and using testify
- refactor code, split file, rename fileds

### Check List 

Tests 

 - Unit test

Code changes

 - Has exported function/method change
 - Has exported variable/fields change

Side effects
None
Related changes
None

### Release note 

```release-note
None
```

PR #5389, opened 2022-05-11
upstream (ticdc): complete upstream and add unit tests

asddongmen:

Issue Number: ref #5288

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note

None
```


pkg/upstream/upstream.go
```diff
@@ -58,13 +61,20 @@ type Upstream struct {
 	PDClock     pdutil.Clock
 	GCManager   gc.Manager
 
+	// holder count of this upstream
+	hc int32
+	// record the time when Upstream.hc becomes zero.
+	idleTime time.Time
+	// mu should be lock when Upstream.hc or Upstream.idleTime need to be changed.
+	mu sync.Mutex
```
CharlesCheung96, line 69: Perhaps these three fields should be combined into single struct.
ref:https://github.com/pingcap/tiflow/pull/5264/files#diff-cfe8ccb173f8b93acc1b97cc5825fdb565d539634714fadffacdf18ea4bc8b26L145-R33

pkg/upstream/manager.go
```diff
@@ -47,41 +49,44 @@ func NewManager(ctx context.Context) *Manager {
...
 	default:
 	}
-
-	if _, ok := m.ups.Load(clusterID); ok {
+	if _, ok := m.ups.Load(upstreamID); ok {
 		return nil
 	}
 	securityConfig := config.GetGlobalServerConfig().Security
-	up := newUpstream(clusterID, pdEndpoints, securityConfig)
+	up := newUpstream(upstreamID, pdEndpoints, securityConfig)
```
sdojjy, line 70: make securityConfig as the function parameter

sdojjy (approved):

pkg/txnutil/gc/testing.go
```diff
@@ -36,3 +36,5 @@ func (m *MockPDClient) UpdateServiceGCSafePoint(ctx context.Context, serviceID s
 func (m *MockPDClient) GetTS(ctx context.Context) (int64, int64, error) {
 	return oracle.GetPhysical(time.Now()), 0, nil
 }
+
+func (m *MockPDClient) Close() {}
```
CharlesCheung96, line 40: Why do we need an empty close method here?

pkg/upstream/upstream.go
```diff
@@ -198,3 +220,42 @@ func (up *Upstream) IsNormal() bool {
...
+
+func (up *Upstream) unhold() {
+	up.hcMu.mu.Lock()
+	defer up.hcMu.mu.Unlock()
+	up.hcMu.hc--
+	if up.hcMu.hc < 0 {
+		log.Panic("upstream's hc should never less than 0", zap.Uint64("upstreamID", up.ID))
+	}
+	if up.hcMu.hc == 0 && up.hcMu.idleTime.IsZero() {
+		up.hcMu.idleTime = time.Now()
```
CharlesCheung96, line 232: IdleTime should always be updated when `up.hcMu.hc == 0`.

pkg/upstream/upstream.go
```diff
@@ -198,3 +220,42 @@ func (up *Upstream) IsNormal() bool {
...
+
+func (up *Upstream) Release() {
+	up.unhold()
+}
+
+// return true if this upstream idleTime reachs maxIdleDuration.
+func (up *Upstream) shouldClose() bool {
+	up.hcMu.mu.Lock()
+	defer up.hcMu.mu.Unlock()
+	if !up.hcMu.idleTime.IsZero() && time.Since(up.hcMu.idleTime) >= maxIdleDuration {
```
CharlesCheung96, line 257: Should check if `hc==0` here, rather than idleTime.

pkg/upstream/upstream.go
```diff
@@ -198,3 +220,42 @@ func (up *Upstream) IsNormal() bool {
...
+	up.hcMu.hc--
+	if up.hcMu.hc < 0 {
+		log.Panic("upstream's hc should never less than 0", zap.Uint64("upstreamID", up.ID))
+	}
+	if up.hcMu.hc == 0 && up.hcMu.idleTime.IsZero() {
+		up.hcMu.idleTime = time.Now()
+	}
+}
+
+func (up *Upstream) hold() {
```
CharlesCheung96, line 236: The names of `hold` and `hc` are somewhat confusing, perhaps we can find a more appropriate name?

pkg/txnutil/gc/testing.go
asddongmen: To implement PDClient interface.

pkg/upstream/upstream.go
asddongmen: No, `hc` can be `0` when a upstream is created but is not hold by a `changefeed` or `processor`.

CharlesCheung96 (approved):

PR #5397, opened 2022-05-11
validator(dm): stop validate when too many pending rows & handle checkpoint too large

### LINKED ISSUE #5392: stop validation when there too many pending rows to avoid oom. fix checkpoint large transaction problem

D3Hunter:

Issue Number: close #5392

- stop validator when there are too many pending rows or pending row size too large
    - default pending row size = 500M
    - default pending row count = worker count * 2000, i.e. 8000 on default worker count
- persist checkpoint using revision to avoid "transaction too large" problem
- persist pending rows using `insert`(seems faster than upsert)
- add a separate routine to make sure `reachedSyncer=true`(renamed to `markErrorStarted`) at some time, current delay equals to DefaultValidatorRowErrorDelay, i.e. 30m
    - previously we only set it to `true` when validator reached progress of syncer, but on extreme case validator may never reach syncer
- print validate speed(binlog size per sec) in status routine    
- add i t for fast mode

### Check List 

Tests 

 - [x] Unit test
 - [x] Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note 

```release-note
`None`.
```


buchuitoudegou:
I'll take a look at tests later

dm/syncer/data_validator.go
```diff
@@ -320,8 +336,25 @@ func (v *DataValidator) Start(expect pb.Stage) {
 	v.L.Info("started")
 }
 
+func (v *DataValidator) markReachedSyncerRoutine() {
```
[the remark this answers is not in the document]
D3Hunter: field renamed, but forget to rename this routine 😂

D3Hunter:
unstable i t : IT-shardddl4_1, might be caused by CI resource overload, will try run it later.

https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/5619/pipeline/204

D3Hunter:
ptal

dm/syncer/data_validator.go
```diff
@@ -78,6 +78,9 @@ const (
 	validatorDmctlOpTimeout = 5 * time.Second
 )
 
+// to make ut easier, we define it as a var, so we can change it.
+var markErrorRowDelay = config.DefaultValidatorRowErrorDelay
```
buchuitoudegou, line 82: Why is it a global variable rather than a member of the validator (worker)? It cannot be used independently by other components.

dm/syncer/validator_checkpoint.go
```diff
@@ -202,81 +204,49 @@ type tableChangeDataForPersist struct {
...
-	// update checkpoint
-	queries = append(queries, `INSERT INTO `+c.checkpointTableName+
-		`(source, binlog_name, binlog_pos, binlog_gtid) VALUES (?, ?, ?, ?) ON DUPLICATE KEY UPDATE
-			source = VALUES(source),
-			binlog_name = VALUES(binlog_name),
-			binlog_pos = VALUES(binlog_pos),
-			binlog_gtid = VALUES(binlog_gtid)
-		`)
-	args = append(args, []interface{}{c.cfg.SourceID, loc.Position.Name, loc.Position.Pos, loc.GTIDSetStr()})
+var triggeredFailOnPersistForIntegrationTest bool
```
buchuitoudegou, line 213: What about `HasFailForIntegrationTest`...

dm/syncer/validator_checkpoint.go
```diff
@@ -335,34 +305,106 @@ func (c *validatorPersistHelper) persist(tctx *tcontext.Context, loc binlog.Loca
...
+
+	return c.execQueriesWithRetry(tctx, queries, args)
+}
+
+func (c *validatorPersistHelper) persistPendingRows(tctx *tcontext.Context, rev int64) error {
+	count := int(c.validator.getAllPendingRowCount()) + 1
+	queries := make([]string, 0, count)
+	args := make([][]interface{}, 0, count)
+
+	// delete pending row of previous failed persist
```
buchuitoudegou, line 321: Is it previous failed and flushed rows? don't understand failed persist😵‍💫

dm/syncer/validator_checkpoint.go
```diff
@@ -335,34 +305,106 @@ func (c *validatorPersistHelper) persist(tctx *tcontext.Context, loc binlog.Loca
...
 
-	return nil
-}
+	nextRevision := c.revision + 1
 
-func (c *validatorPersistHelper) incrRevision() {
+	// we use "insert" to save pending data to speed up(see https://asktug.com/t/topic/33147) since
+	// the number of pending rows maybe large.
+	// we run sql one by one to avoid potential "transaction too large" error since pending data may quite large.
+	// And use revision field to correspond between checkpoint table and pending row table
```
buchuitoudegou, line 375:
```suggestion
	// And use revision to synchronize versions of checkpoint table and pending row table
```
A bit obscure using `correspond between` (haven't seen such a phrase). I tried paraphrasing it and please let me know if it isn't what you meant.

dm/dm/config/task.go
```diff
@@ -380,6 +386,20 @@ func (v *ValidatorConfig) Adjust() error {
...
+
+	_, err := units.RAMInBytes(v.MaxPendingRowSize)
+	if err != nil {
+		return err
+	}
+	if v.MaxPendingRowCount == 0 {
+		// validator validates every ValidatorMaxAccumulatedRowBeforeValidate rows.
+		// if after 4 validation on each worker, we still cannot reduce the row count,
+		// we take it as a signal that there are too many validation failures.
+		v.MaxPendingRowCount = v.WorkerCount * 4 * ValidatorMaxAccumulatedRowBeforeValidate
```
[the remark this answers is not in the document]
D3Hunter: `ValidatorMaxAccumulatedRowBeforeValidate` is more verbose

dm/syncer/data_validator.go
D3Hunter: it should be a const, but as comments says `to make ut easier, we define it as a var, so we can change it.`

dm/syncer/validator_checkpoint.go
D3Hunter: more verbose

dm/syncer/validator_checkpoint.go
D3Hunter: should be `delete pending rows left by previous failed persist`

dm/dm/config/task.go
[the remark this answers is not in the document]
buchuitoudegou: Yes. I mean replace `ValidatorMaxAccumulatedRowBeforeValidate` with `ValidatorMaxLeftRowBefore`

dm/syncer/validator_checkpoint.go
D3Hunter: should be link or associate

dm/syncer/validator_checkpoint.go
buchuitoudegou: `triggeredFailOnPersistForIntegrationTest` is too long. I suggested to replace it with `HasFailForIntegrationTest`

dm/syncer/validator_checkpoint.go
buchuitoudegou: > should be `delete pending rows left by previous failed persist`

Plz rephrase the comment by this (p.s. `persist`'d better be changed to `persistence`)

dm/syncer/validator_checkpoint.go
```diff
@@ -335,34 +305,106 @@ func (c *validatorPersistHelper) persist(tctx *tcontext.Context, loc binlog.Loca
...
-				srcDataStr, dstDataStr, r.tp, pb.ValidateErrorState_NewErr,
+				string(srcDataStr), string(dstDataStr), r.tp, pb.ValidateErrorState_NewErr,
 			})
 		}
 	}
-	// todo: performance issue when using insert on duplicate? https://asktug.com/t/topic/33147
-	// todo: will this transaction too big? but checkpoint & pending changes should be saved in one tx
-	var err error
+
+	return c.execQueriesWithRetry(tctx, queries, args)
```
buchuitoudegou, line 313: This function simply executes queries one after another without using transactions. What if it succeeded to synchronize the table status but failed to insert error rows? When the validator recovers from the two tables, inconsistency probably occurs that the statuses of tables might posit an error but we cannot find it from error rows. I don't think this inconsistency is tolerable.

dm/syncer/validator_checkpoint.go
D3Hunter: its too long most for the suffix ForIntegrationTest.

`HasFail` is not enough to show its meaning

dm/dm/config/task.go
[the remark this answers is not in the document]
D3Hunter: don't know what ValidatorMaxLeftRowBefore means

but maybe we can remove the suffix `BeforeValidate`

dm/syncer/validator_checkpoint.go
D3Hunter: persist here means the method name, not some verb.

dm/syncer/validator_checkpoint.go
D3Hunter: > What if it succeeded to synchronize the table status but failed to insert error rows?

after resume, validator will start from previous loc, and find those error rows again

> the statuses of tables might posit an error

status of table can be stop or running, it cannot show whether there's error rows.

buchuitoudegou: lgtm

dm/dm/config/task.go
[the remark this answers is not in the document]
buchuitoudegou: lgtm

dm/syncer/validator_checkpoint.go
buchuitoudegou: lgtm

D3Hunter:
ping @lance6716 @Ehco1996

dm/dm/config/task.go
```diff
@@ -380,6 +386,20 @@ func (v *ValidatorConfig) Adjust() error {
 	if v.BatchQuerySize == 0 {
 		v.BatchQuerySize = DefaultValidatorBatchQuerySize
 	}
+	if v.MaxPendingRowSize == "" {
+		v.MaxPendingRowSize = DefaultValidatorMaxPendingRowSize
+	}
+
+	_, err := units.RAMInBytes(v.MaxPendingRowSize)
```
Ehco1996, line 393: how about use this func `ParseFileSize` in dm/pkg/dumpling/utils.go

dm/syncer/data_validator.go
```diff
@@ -333,10 +366,18 @@ func (v *DataValidator) printStatusRoutine() {
 				v.pendingRowCounts[rowUpdated].Load(),
 				v.pendingRowCounts[rowDeleted].Load(),
 			}
+			currProcessedBinlogSize := v.processedBinlogSize.Load()
```
Ehco1996, line 369: seems we can also add a metric for this , can do this later

dm/syncer/validator_checkpoint.go
```diff
@@ -149,7 +151,7 @@ func (c *validatorPersistHelper) createTable(tctx *tcontext.Context) error {
 			revision bigint NOT NULL,
 			create_time timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
 			update_time timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
-			UNIQUE KEY uk_source_schema_table_key(source, schema_name, table_name, row_pk),
+			INDEX idx_source_schema_table_key(source, schema_name, table_name, row_pk),
```
Ehco1996, line 154: it seems that this affects the current checkpoint, is there some extra work that needs to be done?

dm/syncer/validator_checkpoint.go
```diff
@@ -335,34 +305,106 @@ func (c *validatorPersistHelper) persist(tctx *tcontext.Context, loc binlog.Loca
 			args = append(args, []interface{}{
 				c.cfg.SourceID, sourceTable.Schema, sourceTable.Table, r.srcJob.Key,
 				targetTable.Schema, targetTable.Table,
-				srcDataStr, dstDataStr, r.tp, pb.ValidateErrorState_NewErr,
+				string(srcDataStr), string(dstDataStr), r.tp, pb.ValidateErrorState_NewErr,
```
Ehco1996, line 308: so `srcDataStr` is not string, maybe  rename this to `srcData` and `dstData`

dm/syncer/validator_checkpoint.go
```diff
@@ -383,18 +425,18 @@ func (c *validatorPersistHelper) loadPersistedDataRetry(tctx *tcontext.Context)
...
 	var err error
 	data := &persistedData{}
-	data.checkpoint, err = c.loadCheckpoint(tctx)
+	data.checkpoint, data.rev, err = c.loadCheckpoint(tctx)
 	if err != nil {
 		return data, err
 	}
 
-	data.pendingChanges, data.rev, err = c.loadPendingChange(tctx)
+	data.pendingChanges, err = c.loadPendingChange(tctx, data.rev)
```
Ehco1996, line 439: Is it possible that checkpoint's rev is different with loadPendingChange's rev?

Ehco1996 (commented):
rest LGTM

dm/dm/config/task.go
D3Hunter: this method seems used to make `filesize` flag compatible with mydumper

from mydumper:
```
.. option:: --chunk-filesize -F

   Split tables into chunks of this output file size. This value is in MB
```
from dumpling
```
-F or --filesize
The approximate size of the output file. The unit should be explicitly provided (such as 128B, 64KiB, 32MiB, 1.5GiB)
```

but we don't have this burden for compatibility, will not change it now.

dm/syncer/validator_checkpoint.go
D3Hunter: as explained in the pr desc, now we use `revision` field to link checkpoint and pending data, the uk constraint need to be removed, so we can persist without using transaction to avoid "transaction too large" problem.

dm/syncer/validator_checkpoint.go
D3Hunter: as explained in the pr desc, now we use revision field to link checkpoint and pending data

on `persist`, pending change is saved first, then checkpoint, and they save using same `revision` . on success the revision of checkpoint equals to the revision in pending change(there may still have pending data of other revision, but it's trash data that need to clean up). so on loading, we have to use the save `revision` from `checkpoint` table to load data from `pending change` table.

dm/syncer/validator_checkpoint.go
D3Hunter: renamed https://github.com/pingcap/tiflow/pull/5397/commits/2d6ee662e21f19f7a9b33251db86483d7b2d8219

D3Hunter:
@Ehco1996 ptal

dm/syncer/validator_checkpoint.go
Ehco1996: If there is already a task running incremental verification, the user needs to manually delete the checkpoint and start again, right?

However, since this is an undisclosed feature, so lgtm

Ehco1996 (approved):
LGTM

dm/syncer/validator_checkpoint.go
D3Hunter: checkpoint tables of validator cannot be cleaned separately, but can be removed together with task's using `start-task --remove-meta`.

PR #5414, opened 2022-05-13
Optimist: return ConflictNone if a conflict DDL has no conflict with at lease one normal table

GMHDBJD:

Issue Number: ref #4287 

When a conflict DDL has no conflict with at least one normal table, we regard it can be executed in downstream.
e.g.
```
tb1: create table tb(id int, a int)
tb2: create table tb(id int, b int)
tb3: create table tb(id int, b int)
```
when we receive a conflict ddl `alter table tb2 rename b to a`, before this pr, tb2 will be skipped and wait tb3's rename column DDL, after this pr, tb2 will directly executed this DDL because tb1 has already have column `a` and no column `b`. That is to say we use a more optimist algorithm in this pr.

### Check List 

Tests 

 - Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/dm/pkg/shardddl/optimism/lock.go
+++ b/dm/pkg/shardddl/optimism/lock.go
@@ -889,8 +889,8 @@ func (l *Lock) trySyncForOneDDL(source, schema, table string, prevTable, postTab
 
 	log.L().Info("found conflict for DDL", zap.String("source", source), zap.String("schema", schema), zap.String("table", table), zap.Stringer("prevTable", prevTable), zap.Stringer("postTable", postTable), log.ShortError(tableErr))
 
-	if idempotent {
-		log.L().Info("return conflict DDL for idempotent DDL", zap.String("source", source), zap.String("schema", schema), zap.String("table", table), zap.Stringer("prevTable", prevTable), zap.Stringer("postTable", postTable))
+	if idempotent || l.noConflictWithOneNormalTable(source, schema, table, prevTable, postTable) {
+		log.L().Info("directly return conflict DDL", zap.Bool("idempotent", idempotent), zap.String("source", source), zap.String("schema", schema), zap.String("table", table), zap.Stringer("prevTable", prevTable), zap.Stringer("postTable", postTable))
 		l.tables[source][schema][table] = postTable
 		l.finalTables[source][schema][table] = postTable
 		return true, ConflictNone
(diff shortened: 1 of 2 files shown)


dm/pkg/shardddl/optimism/lock.go
```diff
@@ -1085,33 +1076,39 @@ func (l *Lock) allFinalTableLarger() bool {
...
+				}
+
+				// judge joined no error
+				joined, err := postTable.Join(ti)
+				if err != nil {
+					continue
+				}
+
+				// judge this normal table is smaller(same as allTableSmaller)
+				if _, err = joined.Compare(prevTable); err == nil {
```
GMHDBJD, line 1095: copy from https://github.com/pingcap/tiflow/blob/08928bfb12a7ce5a5a808f11488acbb3c36f43c8/dm/pkg/shardddl/optimism/lock.go#L1003

dm/pkg/shardddl/optimism/lock.go
```diff
@@ -1085,33 +1076,39 @@ func (l *Lock) allFinalTableLarger() bool {
...
+					continue
+				}
+
+				// judge this normal table is larger(same as allTableLarger)
+				if joined, err = prevTable.Join(ti); err != nil {
```
GMHDBJD, line 1100: copy from https://github.com/pingcap/tiflow/blob/08928bfb12a7ce5a5a808f11488acbb3c36f43c8/dm/pkg/shardddl/optimism/lock.go#L1043

dm/pkg/shardddl/optimism/lock.go
```diff
@@ -889,8 +889,8 @@ func (l *Lock) trySyncForOneDDL(source, schema, table string, prevTable, postTab
 
 	log.L().Info("found conflict for DDL", zap.String("source", source), zap.String("schema", schema), zap.String("table", table), zap.Stringer("prevTable", prevTable), zap.Stringer("postTable", postTable), log.ShortError(tableErr))
 
-	if idempotent {
-		log.L().Info("return conflict DDL for idempotent DDL", zap.String("source", source), zap.String("schema", schema), zap.String("table", table), zap.Stringer("prevTable", prevTable), zap.Stringer("postTable", postTable))
+	if idempotent || l.noConflictWithOneNormalTable(source, schema, table, prevTable, postTable) {
```
GMHDBJD, line 892: idempotent means the DDL has already been executed in this table before.
noConflictWithOneNormalTable means the DDL has already been executed in at least one other table before.

GMHDBJD:
/run-all-tests
cc @lance6716 @lichunzhu

lichunzhu (approved):
LGTM

Ehco1996 (approved):
LGTM

PR #5440, opened 2022-05-16
Table property filter

hicqu:

Issue Number: ref #5718

Use table property filter to speed up cdc sorter, so that CPU usage and latency can be reduced.

#### Tests 

 - Unit test

Benchmarks
#### vanilla v.s. patched, with modified configurations:
* tikv.min-ts-interval = 100ms
* cdc.owner-flush-interval = 50ms
* cdc.processor-flush-interval = 50ms

### Release note 

```release-note
None
```


overvenus (commented):
Rest LGTM

pkg/db/pebble.go
```diff
@@ -94,14 +96,32 @@ func buildPebbleOption(id int, memInByte int, cfg *config.DBConfig) (pebble.Opti
...
 func OpenPebble(
 	ctx context.Context, id int, path string, memInByte int, cfg *config.DBConfig,
+) (DB, error) {
+	return doOpenPebble(ctx, id, path, memInByte, true, cfg)
+}
+
+func doOpenPebble(
+	ctx context.Context,
+	id int, path string, memInByte int,
+	withTablePropertyCollectors bool,
```
sdojjy, line 113: seems `withTablePropertyCollectors` is always true, why we need this flag?

sdojjy (approved):

pkg/db/pebble.go
hicqu: For some tests in which keys are not valid cdc event key format.

overvenus (approved):

PR #3599, opened 2021-11-23
sink: remove cdclog.

3AceShowHand:
### What problem does this PR solve? 

This PR remove all `cdclog` related code, include `S3Sink` and `fileSink`.

### What is changed and how it works?

* remove all related code and test cases.

### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes

### Release note 

```release-note
None
```

PR #5455, opened 2022-05-17
sink(ticdc): add blob storage sink

gegao-db:

Issue Number: close #xxx

Add a new blob storage sink (initially only supports s3).
The implementation is kind of based on the previously removed s3 sink implementation in https://github.com/pingcap/tiflow/pull/3599.

**Produced file/directory structure**

- `{bucket}/{prefix}/log.meta` (global checkpoint ts, table names and ids)
   On every global checkpoint ts updates
- `{bucket}/{prefix}/t_{table_id}/row.{commit_ts}`
   Flush on every FlushRowChangedEvents
- `{bucket}/{prefix}/ddls/ddl.{commit_ts}`
   Write on every DDL event

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note

```

### DIFF
--- a/cdc/sink/blob/blob.go
+++ b/cdc/sink/blob/blob.go
@@ -1,0 +1,334 @@
+// Copyright 2022 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package blob
+
+import (
+	"context"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/pingcap/errors"
+	backup "github.com/pingcap/kvproto/pkg/brpb"
+	"github.com/pingcap/log"
+	"github.com/pingcap/tidb/br/pkg/storage"
+	"github.com/pingcap/tiflow/cdc/model"
+	"github.com/pingcap/tiflow/cdc/sink/codec"
+	"github.com/pingcap/tiflow/pkg/config"
+	cerror "github.com/pingcap/tiflow/pkg/errors"
+	"github.com/uber-go/atomic"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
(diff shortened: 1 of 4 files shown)


zhaoxinyu:
@gegao-db, Hi, thanks for your work. ❤️ After reviewing this PR, we found some points worth improving:
1. When there are large amount of tables need to be flushed to s3, we need to start the same amount of goroutines in `startFlush()` [function](https://github.com/pingcap/tiflow/pull/5455/files#diff-1a28985c260c9cbec8ab3867c0b008bd3d22bbaac9a53299c8d6be429aedd6efR240), which may not be a good idea. Probably a goroutine pool with a fixed number of goroutines can be utilized to mitigate this overhead.
2. Writing file to s3 may encounter the throttling problem, a rate limiter is needed to control the request frequency.
3. Currently `FlushRowChangedEvents()` is an asynchronous interface, the resolvedTs could be sent to a channel and received from that channel from another goroutine. Please reference [MySQL sink](https://github.com/pingcap/tiflow/blob/master/cdc/sink/mysql/mysql.go#L242) and [MQ sink] (https://github.com/pingcap/tiflow/blob/master/cdc/sink/mq/mq.go#L189).
4. `FlushRowChangedEvents()` in this PR may flush some `RowChangedEvent`s which `CommitTs` is larger than the `ResolvedTs`, is this in line with your expectation?

zhaoxinyu:
@gegao-db, Hi, thanks for your work. ❤️ After reviewing this PR, we found some points worth improving:
1. When there are large amount of tables need to be flushed to s3, we need to start the same amount of goroutines in `startFlush()` [function](https://github.com/pingcap/tiflow/pull/5455/files#diff-1a28985c260c9cbec8ab3867c0b008bd3d22bbaac9a53299c8d6be429aedd6efR240), which may not be a good idea. Probably a goroutine pool with a fixed number of goroutines can be utilized to mitigate this overhead.
2. Writing file to s3 may encounter the throttling problem, a rate limiter is needed to control the request frequency.
3. Currently `FlushRowChangedEvents()` is an asynchronous interface, the resolvedTs could be sent to a channel and received from that channel from another goroutine. Please reference [MySQL sink](https://github.com/pingcap/tiflow/blob/master/cdc/sink/mysql/mysql.go#L242) and [MQ sink] (https://github.com/pingcap/tiflow/blob/master/cdc/sink/mq/mq.go#L189).
4. `FlushRowChangedEvents()` in this PR may flush some `RowChangedEvent`s which `CommitTs` is larger than the `ResolvedTs`, is this in line with your expectation?

PR #5462, opened 2022-05-18
scheduler(cdc): table executor support 2 phase scheduling

3AceShowHand:

Issue Number: close #xxx

This PR makes `TableExecutor` interface can support 2 phase scheduling protocol, by introduce new table status.

`TableStatusPreparing` -> `TableStatusPrepared` -> `TableStatusReplicating` -> `TableStatusStopped`.

#### Tests 

 - Unit test

### Release note 

```release-note
`None`.
```


cdc/scheduler/internal/tp/transport.go
```diff
@@ -18,8 +18,8 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/ngaut/log"
 	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
```
overvenus, line 22: Good catch!

cdc/sorter/leveldb/reader.go
```diff
@@ -505,3 +519,7 @@ func (r *reader) OnClose() {
 	_ = r.state.tryReleaseIterator()
 	r.common.closedWg.Done()
 }
+
+func (r *reader) UpdateStartTs(ts uint64) {
```
overvenus, line 523: Who is its caller?

PR #5472, opened 2022-05-19
dm: periodically print log for lightning and dumpling

### LINKED ISSUE #5448: DM dump and lightning loader didn't periodically print status to log
### What did you do?

https://github.com/pingcap/tiflow/blob/5e858a4766cf0ead2995581e141a43aa0a3c74ab/dm/dumpling/dumpling.go#L210

and

https://github.com/pingcap/tiflow/blob/5e858a4766cf0ead2995581e141a43aa0a3c74ab/dm/loader/lightning.go#L414

don't have `printStatus` like

https://github.com/pingcap/tiflow/blob/5e858a4766cf0ead2995581e141a43aa0a3c74ab/dm/loader/status.go#L37

_No response_

### What did you see instead?

_No response_

### Versions of the cluster

master

```console
```

buchuitoudegou:

Issue Number: close #5448 

### Release note 

```release-note
 `None`.
```

### DIFF
--- a/dm/dumpling/dumpling.go
+++ b/dm/dumpling/dumpling.go
@@ -15,6 +15,7 @@ package dumpling
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"sync"
 	"time"
@@ -222,7 +223,24 @@ func (m *Dumpling) Status(_ *binlog.SourceStatus) interface{} {
 		FinishedRows:      mid.FinishedRows,
 		EstimateTotalRows: mid.EstimateTotalRows,
 	}
+	go m.printStatus(*s)
 	return s
+}
+
+func (m *Dumpling) printStatus(status pb.DumpStatus) {
+	var estimateProgress string
+	if status.FinishedRows >= status.EstimateTotalRows {
+		estimateProgress = "100.00%"
+	} else {
+		estimateProgress = fmt.Sprintf("%.2f %%", status.FinishedRows/status.EstimateTotalRows*100)
+	}
+	m.logger.Info("progress status of dumpling",
+		zap.Int64("total_tables", status.TotalTables),
+		zap.Int64("finished_tables", int64(status.CompletedTables)),
+		zap.Int64("estimated_total_rows", int64(status.EstimateTotalRows)),
+		zap.Int64("finished_rows", int64(status.FinishedRows)),
+		zap.String("estimated_progress", estimateProgress),
+	)
 }
 
 // Type implements Unit.Type.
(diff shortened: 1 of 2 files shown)


dm/dumpling/dumpling.go
```diff
@@ -222,9 +223,26 @@ func (m *Dumpling) Status(_ *binlog.SourceStatus) interface{} {
 		FinishedRows:      mid.FinishedRows,
 		EstimateTotalRows: mid.EstimateTotalRows,
 	}
+	go m.printStatus(*s)
```
Ehco1996, line 226: i forget why here need to use another thread to print, do you have any ideas?

buchuitoudegou: Maybe it could decouple users' querying from probable disk I/O (for log writing)? I don't think it does augment a significant performance though🤣

lance6716: syncer's `printStatus` needs lock thus maybe slow. No need to block RPC. Then I simply use same style for other units.

dm/dumpling/dumpling.go
```diff
@@ -222,9 +223,26 @@ func (m *Dumpling) Status(_ *binlog.SourceStatus) interface{} {
...
 	return s
```
lance6716, line 227: maybe let s := pb.DumpStatus{}, so we can m.printStatus(s) and return &s.

dm/dumpling/dumpling.go
lance6716: I think it's OK we use another goroutine for this light-weighted work or not.

dm/loader/lightning.go
```diff
@@ -412,5 +412,16 @@ func (l *LightningLoader) status() *pb.LoadStatus {
 
 // Status returns the unit's current status.
 func (l *LightningLoader) Status(_ *binlog.SourceStatus) interface{} {
+	go l.printStatus()
 	return l.status()
```
lance6716, line 416: Sorry that we don't need to insist on using another goroutine. s := l.status(), using s to log and returning it is easier.

dm/dumpling/dumpling.go
```diff
@@ -214,6 +215,10 @@ func (m *Dumpling) Status(_ *binlog.SourceStatus) interface{} {
 	if m.core == nil {
 		return &pb.DumpStatus{}
 	}
+	return m.printStatus()
+}
+
+func (m *Dumpling) printStatus() *pb.DumpStatus {
```
lance6716, line 221: I see printStatus and expect its returning is of type error. maybe no need to split a function for Status, or rename it to status like lightning loader.

lance6716 (approved):

buchuitoudegou:
ptal @Ehco1996

Ehco1996 (approved):
LGTM

Issue #5468, opened 2022-05-18
Data loss when upstream transcation conflicts during incremental scan


overvenus:
### What did you do?

 For UPDATE SQL, its prewrite event has both value and old value.
 It is possible that TiDB prewrites multiple times for the same row when
 there are other transcations it conflicts with. For this case,
 if the value is not "short", only the first prewrite contains the value.

 TiKV may output events for the UPDATE SQL as following:

```
 TiDB: [Prwrite1]    [Prewrite2]      [Commit]
       v             v                v                                   Time
 ---------------------------------------------------------------------------->
         ^            ^    ^           ^     ^       ^     ^          ^     ^
 TiKV:   [Scan Start] [Send Prewrite2] [Send Commit] [Send Prewrite1] [Send Init]
 TiCDC:                    [Recv Prewrite2]  [Recv Commit] [Recv Prewrite1] [Recv Init]
```

TiCDC mistakely outputs an event that contains the old value but not contains the value.

The event is translated into DELETE in sink, so the row is lost.

See line L718-L743
https://github.com/pingcap/tiflow/blob/d76503bdd1c04411a3068a1850d3acd4f5d60a3b/cdc/kv/region_worker.go#L718-L743
 

### What did you expect to see?

No data loss

### What did you see instead?

Data is lost.

### Versions of the cluster

TiCDC version (execute `cdc version`):

```console
v5.0.1
```

PR #5477, opened 2022-05-19
kv(ticdc): fix data loss when upstream txn conflicts during scan

overvenus:

Issue Number: close #5468 

Do not match prewrite and commit util it receives init event.
See more in #5468 

#### Tests 

 - Unit test

It caches more events in memory during incremental scan, it cause OOM when upstream workload is write-heavy.

No.

### Release note 

```release-note
Fix data loss when upstream transaction conflicts during cdc reconnection 
```

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -1176,7 +1176,7 @@ func (s *eventFeedSession) receiveFromStream(
 
 	// always create a new region worker, because `receiveFromStream` is ensured
 	// to call exactly once from outter code logic
-	worker := newRegionWorker(s, addr)
+	worker := newRegionWorker(changefeedID, s, addr)
 
 	defer worker.evictAllRegions()
 
--- a/cdc/kv/matcher.go
+++ b/cdc/kv/matcher.go
@@ -14,6 +14,8 @@
 package kv
 
 import (
+	"unsafe"
+
 	"github.com/pingcap/kvproto/pkg/cdcpb"
 	"github.com/pingcap/log"
 	"go.uber.org/zap"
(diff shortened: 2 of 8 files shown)


cdc/kv/matcher.go
```diff
@@ -97,3 +108,19 @@ func (m *matcher) matchCachedRow() []*cdcpb.Event_Row {
 func (m *matcher) rollbackRow(row *cdcpb.Event_Row) {
 	delete(m.unmatchedValue, newMatchKey(row))
 }
+
+func (m *matcher) cacheRollbackRow(row *cdcpb.Event_Row) {
+	m.cachedRollback = append(m.cachedRollback, row)
+}
+
+func (m *matcher) matchCachedRollbackRow(initialized bool) {
+	if !initialized {
```
nongfushanquan, line 117: why need this parameter since the initialized = false would never be called?

liuzix (approved):

cdc/kv/matcher.go
overvenus: No if there is no bug, it's an additional check to make sure we does not misuse `matchCachedRollbackRow` and `state.initialized`.

cdc/kv/client.go
```diff
@@ -1176,7 +1176,7 @@ func (s *eventFeedSession) receiveFromStream(
 
 	// always create a new region worker, because `receiveFromStream` is ensured
 	// to call exactly once from outter code logic
```
3AceShowHand, line 1178:
```suggestion
	// to call exactly once from outer code logic
```

3AceShowHand (approved):

nongfushanquan (approved):

cdc/kv/region_worker.go
```diff
@@ -725,7 +720,7 @@ func (w *regionWorker) handleEventEntry(
 					zap.Uint64("regionID", regionID))
 				return errUnreachable
 			}
-			ok := state.matcher.matchRow(entry)
+			ok := state.matcher.matchRow(entry, state.initialized)
 			if !ok {
 				if !state.initialized {
 					state.matcher.cacheCommitRow(entry)
```
hicqu, line 726: Seems `entry` can be a prewrite without value?

overvenus: It's alway a COMMIT, see line 718.

cdc/model/kv.go
```diff
@@ -84,8 +84,10 @@ type RawKVEntry struct {
 }
 
 func (v *RawKVEntry) String() string {
-	return fmt.Sprintf("OpType: %v, Key: %s, Value: %s, StartTs: %d, CRTs: %d, RegionID: %d",
-		v.OpType, string(v.Key), string(v.Value), v.StartTs, v.CRTs, v.RegionID)
+	// TODO: redact values.
```
zhaoxinyu, line 87: Is any remaining works to do here?

Issue #5352, opened 2022-05-07
Task lists for merging dataflow engine into pingcap/tiflow


amyangfei:
### Is your feature request related to a problem?

This issue lists tasks to do for merging dataflow engine into pingcap/tiflow

### Describe the feature you'd like

1. [x] merge the main code base and make CI happy.
2. [x] Update issue template, add dataflow engine. ref: https://github.com/ti-community-infra/configs/pull/575, https://github.com/pingcap/tiflow/pull/5522
3. [x] ~Update readme of tiflow repo, adding introduction of dataflow engine.~
4. [x] Update CI required fields.
5. [x] Add dataflow engine into verify CI, ref: https://github.com/pingcap/tiflow/pull/3153, https://github.com/pingcap/tiflow/pull/5501, https://github.com/PingCAP-QE/devops-config/pull/128
6. [x] Optimize CI scripts, skip DM/TiCDC integration tests when only updating engine code. https://github.com/PingCAP-QE/ci/pull/940
7. [x] Make dataflow engine lint compatible with tiflow repo. ref: https://github.com/hanfei1991/microcosm/pull/379

amyangfei:
Will update readme of dataflow engine in next month.

PR #5501, opened 2022-05-20
ci(engine): support verify ci for dataflow engine

amyangfei:

Issue Number: ref #5352

- Add verify ci entrypoint for dataflow engine in Makefile

#### Tests 

 - Unit test

No

No

### Release note 

```release-note
None
```


amyangfei:
PTAL, thanks! ❤️ @purelind

Makefile
```diff
@@ -510,3 +510,14 @@ define run_engine_unit_test
 	|| { $(FAILPOINT_DISABLE); exit 1; }
 	$(FAILPOINT_DISABLE)
 endef
+
+engine_unit_test_in_verify_ci: check_failpoint_ctl tools/bin/gotestsum tools/bin/gocov tools/bin/gocov-xml
+	mkdir -p $(ENGINE_TEST_DIR)
+	$(FAILPOINT_ENABLE)
+	@export log_level=error; \
```
lance6716, line 517: not sure if we need some info level log to debug when CI fails

amyangfei: Seems this environment variable doesn't affect the log behavior in unit test.

maxshuang (approved):

PR #5502, opened 2022-05-20
metric[engine]: framework support metric interface

### LINKED ISSUE #5496: BaseMaster/BaseWorker support  multi-projects/tasks metric interface

maxshuang:

Issue Number: close  #5496

1. Offer MetricFactory() method in BaseMaster/BaseWorker/BaseJobMaster
2. pass worker type when creating master/worker

```release-note
None
```


engine/lib/master.go
```diff
@@ -188,6 +191,7 @@ func NewBaseMaster(
 	ctx *dcontext.Context,
 	impl MasterImpl,
 	id libModel.MasterID,
+	tp libModel.WorkerType,
```
amyangfei, line 194: What about passing `WorkerType` in `dcontext.Context`

engine/lib/common.go
```diff
@@ -66,3 +66,37 @@ type WorkerHandle = master.WorkerHandle
...
+// TODO: let user register a unique identifier for the metric prefix
+func WorkerTypeForMetric(t WorkerType) string {
+	switch t {
+	case JobManager:
+		// jobmanager is the framework level job
+		return ""
+	case CvsJobMaster:
+		return "cvs"
+	case CvsTask:
+		return "cvs"
```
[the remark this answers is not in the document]
maxshuang: resolved in latest commit

engine/lib/master.go
maxshuang: What's the principle to pass some params in `dcontext.Context` ?

amyangfei: There are no clear rules, I think passing worker type is reasonable.

liuzix (approved):

PR #5503, opened 2022-05-20
engine: add query status

GMHDBJD:

Issue Number: ref #5504 

- move dmWorker to executor/dm
- use DMWorker instead of dump/load/task
- unify sender interface for master/worker
- automatically call message handler by reflect
- implement query-status api for dm by MessageAgent
- add DebugJob command to send requset/response with Jobmaster by MessageAgent
- register/unregister topic by MessageAgent

QueryStatus result can be found at https://github.com/pingcap/tiflow/blob/e636374d0a1248386b436bff80b523b3f38417d3/engine/jobmaster/dm/api_test.go#L121-L211
Request/Response example can be found at https://github.com/pingcap/tiflow/blob/b41b239b8dd012b9105aa58e326287cd8bcd2b10/engine/servermaster/jobmanager.go#L131-L180

#### Tests 

 - Unit test
 - Integration test
 
### Release note 

```release-note
None
```


lance6716 (commented):
don't forget to rename folder name dmTask to dmtask, like we didn't use jobMaster but jobmaster

engine/executor/dmTask/dmtask.go
```diff
@@ -0,0 +1,209 @@
...
+
+// NewWorkerImpl implements WorkerFactory.NewWorkerImpl
+func (f taskFactory) NewWorkerImpl(ctx *dcontext.Context, workerID libModel.WorkerID, masterID libModel.MasterID, conf lib.WorkerConfig) (lib.WorkerImpl, error) {
+	baseDMTask := NewBaseDMTask(ctx, f.workerType, conf)
+	switch f.workerType {
+	case lib.WorkerDMDump:
+		return newDumpTask(baseDMTask), nil
+	case lib.WorkerDMLoad:
+		return newLoadTask(baseDMTask), nil
+	default:
```
lance6716, line 72: it's safer that we write WorkerDMSync rather than rely on ambiguous default.

engine/executor/dmTask/dmtask.go
```diff
@@ -0,0 +1,209 @@
...
+		return newSyncTask(baseDMTask), nil
+	}
+}
+
+// DMTask defines the interface for dump/load/sync
+type DMTask interface {
```
lance6716, line 78: no need to be public?

engine/executor/dmTask/dmtask.go
```diff
@@ -0,0 +1,209 @@
...
+	onInit(ctx context.Context) error
+	onFinished(ctx context.Context) error
+	createUnitHolder(cfg *config.SubTaskConfig) *unitHolder
+}
+
+// BaseDMTask implements some default methods for dm task
+type BaseDMTask struct {
+	DMTask
+	lib.BaseWorker
```
lance6716, line 87: we should assert struct BaseDMTask implements interface lib.BaseWorker, not embed it. If in future lib.BaseWorker implements more method we will not discover and panic on invocation.

lance6716 (commented):
will review later

engine/executor/dmTask/dmtask.go
GMHDBJD: I just want to be consistent with cvstask. And I think master-worker is code level and job-task is a concept.

GMHDBJD:
PTAL @lance6716 
Since we don't have openapi, I will try to add e2e test in later pr.

engine/executor/dm/task.go
```diff
@@ -0,0 +1,253 @@
...
+
+// task defines the interface for dump/load/sync
+type task interface {
+	onInit(ctx context.Context) error
+	onFinished(ctx context.Context) error
+	createUnitHolder(cfg *config.SubTaskConfig) unit.Holder
+}
+
+// baseTask implements some default methods for dm task
+type baseTask struct {
```
lance6716, line 90: In my view this struct should named as baseDMWorker. Currently worker and task is 1:1, and they are semanticallly similar like they are both "runnable", can "init", but it's more reasonable that only worker can "send/receive message", and names like `MustRegisterWorkerType`, `lib.WorkerImpl` prefer worker not task.

So worker holding task is easier to understand. Maybe worker holding changefeed, worker holding actors are also acceptable for CDC? @liuzix

engine/executor/dm/api.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	taskStatus := runtime.NewTaskStatus(t.workerType, t.taskID, t.stage)
+	if err = json.Unmarshal(statusBytes, taskStatus); err != nil {
+		return dmpkg.QueryStatusResponse{ErrorMsg: err.Error()}, nil
+	}
+	return dmpkg.QueryStatusResponse{TaskStatus: taskStatus}, nil
+}
+
+// StopWorker implements the api of stop worker message which kill itself.
+func (t *baseTask) StopWorker(ctx context.Context, msg dmpkg.StopWorkerMessage) error {
```
lance6716, line 43: or here you want to express the relationship of "task has worker so task can stop its worker"?

engine/executor/dm/task.go
```diff
@@ -0,0 +1,253 @@
...
+	t.unitHolder.Close()
+	return nil
+}
+
+// exit closes the task and exits.
+func (t *baseTask) exit(ctx context.Context, status libModel.WorkerStatus, err error) error {
+	if err := t.CloseImpl(ctx); err != nil {
+		log.L().Warn("fail to close task", log.ShortError(err))
+	}
+	return t.Exit(ctx, status, err)
```
lance6716, line 179: it's weird that a private "exit" calls public "Exit". To solve this problem, maybe rename it closeAndExit or the framework should call CloseImpl in Exit? @liuzix

engine/executor/dm/unit/holder.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/pingcap/tiflow/dm/dm/worker"
+	"github.com/pingcap/tiflow/dm/pkg/backoff"
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"github.com/pingcap/tiflow/engine/jobmaster/dm/metadata"
+	"github.com/pingcap/tiflow/engine/lib"
+	"go.uber.org/zap"
+)
+
+// Holder hold a unit of DM
+type Holder interface {
```
lance6716, line 33: maybe UnitHolder. Holder is too ambiguous to express anything.

engine/jobmaster/dm/api.go
```diff
@@ -0,0 +1,118 @@
...
+		}()
+	}
+	wg.Wait()
+	return jobStatus, nil
+}
+
+// QueryTaskStatus query status for a task
+func (jm *JobMaster) QueryTaskStatus(ctx context.Context, taskID string, workerStatus runtime.WorkerStatus) dmpkg.QueryStatusResponse {
+	if workerStatus.Stage == runtime.WorkerFinished {
+		status, ok := jm.taskManager.GetTaskStatus(taskID)
```
lance6716, line 103: JobMaster has workerManager and taskManager, this makes me also don't want to mix worker and task to one concept.

engine/jobmaster/dm/api.go
```diff
@@ -0,0 +1,118 @@
...
+		if !ok || status.GetStage() != metadata.StageFinished {
+			return dmpkg.QueryStatusResponse{ErrorMsg: fmt.Sprintf("finished task status for task %s not found", taskID)}
+		}
+		return dmpkg.QueryStatusResponse{TaskStatus: status}
+	}
+
+	req := dmpkg.QueryStatusRequest{
+		Task: taskID,
+	}
+	resp, err := jm.messageAgent.SendRequest(ctx, taskID, dmpkg.QueryStatus, req)
```
lance6716, line 113: Do we need to append more unique information to topic? like `HeartbeatPingTopic`

engine/executor/dm/api.go
GMHDBJD: this is align with createWorker 😂

engine/jobmaster/dm/api.go
GMHDBJD: Now, all topic format by DM---JobmasterID---TaskID

engine/executor/dm/unitholder.go
```diff
@@ -0,0 +1,193 @@
...
+	case len(result.Errors) == 0:
+		return metadata.StageFinished
+	default:
+		return metadata.StagePaused
+	}
+}
+
+// Status implement UnitHolder.Status
+func (u *unitHolderImpl) Status(ctx context.Context) interface{} {
+	return u.unit.Status(nil)
```
lance6716, line 192: Do we plan to fetch SourceStatus in future, and make "masterBinlog" works again? If not we can remove "masterBinlog".

engine/executor/dm/unitholder_test.go
```diff
@@ -0,0 +1,151 @@
...
+	u.On("Status").Return(&pb.DumpStatus{})
+	status := unitHolder.Status(context.Background())
+	require.Equal(t, &pb.DumpStatus{}, status)
+
+	unitHolder.resultCh <- pb.ProcessResult{Errors: []*pb.ProcessError{{
+		ErrCode: 0,
+		Message: "context canceled",
+	}}}
+	require.NoError(t, unitHolder.Tick(context.Background()))
+	require.Equal(t, metadata.StageFinished, unitHolder.Stage())
```
lance6716, line 71: If we implements PauseTask in future, PauseTask will cancel the context and seems the "context canceled" will be filtered,  the stage will become StageFinished

engine/executor/dm/worker.go
```diff
@@ -0,0 +1,241 @@
...
+// InitImpl implements lib.WorkerImpl.InitImpl
+func (w *dmWorker) InitImpl(ctx context.Context) error {
+	if err := w.messageAgent.Init(ctx); err != nil {
+		return err
+	}
+	// register jobmaster sender
+	if err := w.messageAgent.UpdateSender(w.masterID, w); err != nil {
+		return err
+	}
+	if w.workerType != lib.WorkerDMSync || w.cfg.Mode != dmconfig.ModeIncrement {
```
lance6716, line 113: this equals to if w.cfg.Mode != dmconfig.ModeIncrement {...}?

engine/executor/dm/worker_test.go
```diff
@@ -0,0 +1,105 @@
...
+	subtaskCfg := &dmconfig.SubTaskConfig{Name: "job-id", SourceID: "task-id", Flavor: mysql.MySQLFlavor}
+	content, err := subtaskCfg.Toml()
+	require.NoError(t, err)
+	RegisterWorker()
+	_, err = registry.GlobalWorkerRegistry().CreateWorker(dctx, lib.WorkerDMDump, "worker-id", "dm-jobmaster-id", []byte(content))
+	require.NoError(t, err)
+	_, err = registry.GlobalWorkerRegistry().CreateWorker(dctx, lib.WorkerDMLoad, "worker-id", "dm-jobmaster-id", []byte(content))
+	require.NoError(t, err)
+	_, err = registry.GlobalWorkerRegistry().CreateWorker(dctx, lib.WorkerDMSync, "worker-id", "dm-jobmaster-id", []byte(content))
+	require.NoError(t, err)
```
lance6716, line 76: seems we can test `messageAgent` is set.

engine/executor/dm/unitholder_test.go
GMHDBJD: Yes, I add the comment at https://github.com/pingcap/tiflow/blob/8c598195388b133aae6d9b10c745ed026f2eae27/engine/executor/dm/unitholder.go#L182

engine/executor/dm/worker_test.go
GMHDBJD: add at L88

engine/executor/dm/unitholder.go
GMHDBJD: I think we shoud implement it in the future.

engine/executor/dm/worker.go
```diff
@@ -0,0 +1,241 @@
...
+	if err := w.messageAgent.UpdateSender(w.masterID, nil); err != nil {
+		return err
+	}
+	w.unitHolder.Close(ctx)
+	w.messageAgent.Close(ctx)
+	return nil
+}
+
+// exit closes the task and exits.
+func (w *dmWorker) exit(ctx context.Context, status libModel.WorkerStatus, err error) error {
```
lance6716, line 162: I think it's a bit weird to let a private function call public function of same name (though Exit is provided by lib.BaseWorker). we can rename exit to closeAndExit.

Or the framework should move CloseImpl inside the Exit? @liuzix

lance6716 (commented):
will review messaage/request/response part later

lance6716 (commented):
reviewing message_agent

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+// sendRequest sends a request message and wait for response.
+func (m *messagePair) sendRequest(ctx context.Context, topic p2p.Topic, command string, req interface{}, sender Sender) (interface{}, error) {
+	msg := message{ID: m.idAllocator.alloc(), Type: requestTp, Command: command, Payload: req}
+	respCh := make(chan interface{}, 1)
+	m.pendings.Store(msg.ID, respCh)
+	defer m.pendings.Delete(msg.ID)
+
+	if err := sender.SendMessage(ctx, topic, msg, true /* block */); err != nil {
```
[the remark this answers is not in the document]
GMHDBJD: update

engine/executor/dm/worker.go
```diff
@@ -0,0 +1,242 @@
...
+
+// OnMasterMessage implements lib.WorkerImpl.OnMasterMessage
+func (w *dmWorker) OnMasterMessage(topic p2p.Topic, message p2p.MessageValue) error {
+	log.L().Info("dmworker.OnMasterMessage", zap.String("topic", topic), zap.Any("message", message))
+	return nil
+}
+
+// CloseImpl implements lib.WorkerImpl.CloseImpl
+func (w *dmWorker) CloseImpl(ctx context.Context) error {
+	// unreigster jobmaster sender
```
lance6716, line 152:
```suggestion
	// unregister jobmaster sender
```

engine/executor/dm/worker.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+// workerStatus gets worker status.
+func (w *dmWorker) workerStatus() libModel.WorkerStatus {
+	stage := w.getStage()
+	code := libModel.WorkerStatusNormal
+	if stage == metadata.StageFinished {
+		code = libModel.WorkerStatusFinished
+	}
+	status := runtime.TaskStatus{Unit: w.workerType, Task: w.taskID, Stage: stage}
```
lance6716, line 222: seems we will get a Normal worker whose stage is Paused. Maybe we can list the combinations of worker statsu and task stage in future.

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+
+var (
+	defaultMessageTimeOut  = time.Second * 2
+	defaultRequestTimeOut  = time.Second * 30
+	defaultResponseTimeOut = time.Second * 2
+	defaultHandlerTimeOut  = time.Second * 30
+)
+
+// generateTopic generate dm message topic.
+func generateTopic(sendID string, receiverID string) string {
```
lance6716, line 41: sendID and receiverID make me think that for a request and response, they two should be swapped. Maybe requestID and taskID?

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+	ctx2, cancel := context.WithTimeout(ctx, defaultResponseTimeOut)
+	defer cancel()
+	log.L().Debug("send response", zap.String("sender-id", senderID), zap.String("command", command), zap.Any("resp", resp))
+	return agent.messagePair.sendResponse(ctx2, generateTopic(agent.id, senderID), msgID, command, resp, sender)
+}
+
+// onMessage receive message/request/response.
+// Forward the response to the corresponding request request.
+// According to the command, the corresponding message processing function of defaultHandler will be called.
```
lance6716, line 281:
```suggestion
// According to the command, the corresponding message processing function of commandHandler will be called.
```

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+// handleResponse receive response.
+func (agent *MessageAgentImpl) handleResponse(id messageID, command string, resp interface{}) error {
+	handler := reflect.ValueOf(agent.commandHandler).MethodByName(command)
+	if !handler.IsValid() {
+		return errors.Errorf("response handler for command %s not found", command)
+	}
+	handlerType := handler.Type()
+	if handlerType.NumOut() != 1 && handlerType.NumOut() != 2 {
```
lance6716, line 312: there's a requirement for commandHandler, we should add the comment above NewMessageAgentImpl

engine/executor/dm/worker.go
```diff
@@ -0,0 +1,242 @@
...
+	return nil
+}
+
+// CloseImpl implements lib.WorkerImpl.CloseImpl
+func (w *dmWorker) CloseImpl(ctx context.Context) error {
+	// unreigster jobmaster sender
+	if err := w.messageAgent.UpdateSender(w.masterID, nil); err != nil {
+		return err
+	}
+	w.unitHolder.Close(ctx)
```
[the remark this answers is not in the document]
GMHDBJD: Here we cannot ignore the error. If we close the loader worker, ignore the unregister error, when we init syncer worker, we will get two receiver for same topic.

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+		return errors.Errorf("wrong response handler type for command %s", command)
+	}
+	ret := reflect.New(handlerType.Out(0).Elem())
+	if bytes, err := json.Marshal(resp); err != nil {
```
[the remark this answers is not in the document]
GMHDBJD: We can't. Resp is marshal and unmarshal by json during send message. It's type is map[string]interface, we should create a new certain type for it.

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+	return errors.Errorf("duplicated response of request %d, and the last response is not consumed", id)
+}
+
+// MessageAgent defines interface for message communication.
+type MessageAgent interface {
+	Init(ctx context.Context) error
+	Tick(ctx context.Context) error
+	Close(ctx context.Context) error
+	UpdateSender(senderID string, sender Sender) error
+	SendMessage(ctx context.Context, senderID string, command string, msg interface{}) error
```
liuzix, line 153: Do you think it is possible to make `SendMessage` and `SendRequest` generic to avoid using `interface{}`?

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+	"github.com/pingcap/errors"
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"github.com/pingcap/tiflow/engine/lib"
+	"github.com/pingcap/tiflow/engine/pkg/p2p"
+	"github.com/pingcap/tiflow/pkg/workerpool"
+	"go.uber.org/zap"
+)
+
+var (
+	defaultMessageTimeOut  = time.Second * 2
```
liuzix, line 34: Can you provide an example when the "MessageTimeOut" will expire? Could 2 seconds be too short?

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+	defaultRequestTimeOut  = time.Second * 30
+	defaultResponseTimeOut = time.Second * 2
```
liuzix, line 36: ditto

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+	defaultHandlerTimeOut  = time.Second * 30
+)
+
+// generateTopic generate dm message topic.
+func generateTopic(senderID string, receiverID string) string {
+	return fmt.Sprintf("DM---%s---%s", senderID, receiverID)
```
liuzix, line 42: how about hex encoding the sender and receiver IDs?

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+	parts := strings.Split(topic, "---")
+	return parts[1], parts[2]
+}
+
+type messageID uint64
+
+type messageType int
+
+const (
+	messageTp messageType = iota
```
liuzix, line 57:
```suggestion
	messageTp messageType = iota+1
```

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+	ID      messageID
+	Type    messageType
+	Command string
+	Payload interface{}
+}
+
+// messageIDAllocator is an id allocator for p2p message system
+type messageIDAllocator struct {
+	mu sync.Mutex
+	id messageID
```
liuzix, line 73: Can you use atomics?

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+func (a *messageIDAllocator) alloc() messageID {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	a.id++
+	return a.id
+}
+
+// Sender defines an interface that supports send message
+type Sender interface {
```
liuzix, line 86: Can you say more on who should implement a `Sender` and where a `Sender` is supposed to be created?

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+	return errors.Errorf("duplicated response of request %d, and the last response is not consumed", id)
+}
+
+// MessageAgent defines interface for message communication.
+type MessageAgent interface {
+	Init(ctx context.Context) error
+	Tick(ctx context.Context) error
+	Close(ctx context.Context) error
+	UpdateSender(senderID string, sender Sender) error
```
liuzix, line 152: Could you comment on when one should call `UpdateSender`?

engine/pkg/dm/message_agent.go
```diff
@@ -0,0 +1,404 @@
...
+	ctx                   context.Context
+	cancel                context.CancelFunc
+	messagePair           *messagePair
+	messageHandlerManager p2p.MessageHandlerManager
+	pool                  workerpool.AsyncPool
+	messageRouter         *lib.MessageRouter
+	wg                    sync.WaitGroup
+	// sender-id -> Sender
+	senders        sync.Map
+	commandHandler interface{}
```
liuzix, line 168: Can you comment on what the `interface{}` should be?

engine/executor/dm/api.go
```diff
@@ -0,0 +1,48 @@
...
+	"encoding/json"
+
+	"github.com/pingcap/errors"
+
+	dmpkg "github.com/pingcap/tiflow/engine/pkg/dm"
+)
+
+// QueryStatus implements the api of query status request.
+func (t *dmWorker) QueryStatus(ctx context.Context, req *dmpkg.QueryStatusRequest) *dmpkg.QueryStatusResponse {
+	// get status from unit
```
lance6716, line 27: in below lines we checked `t.taskID != msg.Task`, but here we don't

engine/executor/dm/api.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"encoding/json"
+
+	"github.com/pingcap/errors"
+
+	dmpkg "github.com/pingcap/tiflow/engine/pkg/dm"
+)
+
+// QueryStatus implements the api of query status request.
+func (t *dmWorker) QueryStatus(ctx context.Context, req *dmpkg.QueryStatusRequest) *dmpkg.QueryStatusResponse {
```
lance6716, line 26:
```suggestion
func (w *dmWorker) QueryStatus(ctx context.Context, req *dmpkg.QueryStatusRequest) *dmpkg.QueryStatusResponse {
```

IDE says receiver name should be consistent

engine/executor/dm/worker.go
```diff
@@ -0,0 +1,242 @@
...
+		stage:      metadata.StageInit,
+		workerType: workerType,
+		taskID:     cfg.SourceID,
+		masterID:   masterID,
+		unitHolder: newUnitHolderImpl(workerType, cfg),
+	}
+
+	// nolint:errcheck
+	ctx.Deps().Construct(func(m p2p.MessageHandlerManager) (p2p.MessageHandlerManager, error) {
+		w.messageAgent = dmpkg.NewMessageAgentImpl(w.taskID, w, m)
```
lance6716, line 98: For all functions called by commandHandler, please add a comment like "FuncXXX is called by refection of commandHandler", otherwise reader will not find how are they called.

lance6716 (commented):
rest lgtm

engine/executor/dm/worker.go
```diff
@@ -0,0 +1,242 @@
...
+		return err
+	}
+	w.unitHolder.Close(ctx)
+	w.messageAgent.Close(ctx)
+	return nil
+}
+
+// closeAndExit closes the task and exits.
+func (w *dmWorker) closeAndExit(ctx context.Context, status libModel.WorkerStatus) error {
+	err := w.CloseImpl(ctx)
```
lance6716, line 163: @liuzix says no need to call CloseImpl. Exit will cause runtime to call it.

GMHDBJD:
> we should continue closing even if UpdateClient meets error
If we close with error, e.g. close loader but unregister with error, when syncer is created, it will register the same topic and then panic

lance6716:
> > we should continue closing even if UpdateClient meets error
> 
> If we close with error, e.g. close loader but unregister with error, when syncer is created, it will register the same topic and then panic

compare

```
if err != nil {return err}
closeOtherMember
```

and

```
err := ...
closeOtherMember
return err
```

I think the latter is more correct because the former will not have the chance to close other member.

And the problem is about when CloseImpl returns error, we should not create next worker unless retry.

engine/executor/dm/worker.go
GMHDBJD: I revert this part now.

lance6716 (approved):

engine/executor/dm/worker.go
```diff
@@ -0,0 +1,243 @@
...
+
+// CloseImpl implements lib.WorkerImpl.CloseImpl
+func (w *dmWorker) CloseImpl(ctx context.Context) error {
+	// unregister jobmaster client
+	err := w.messageAgent.UpdateClient(w.masterID, nil)
+	if err != nil {
+		log.L().Error("failed to update message client", log.ShortError(err))
+	}
+	w.unitHolder.Close(ctx)
+	w.messageAgent.Close(ctx)
```
amyangfei, line 158: These two Close always return nil error?

engine/executor/dm/worker.go
```diff
@@ -0,0 +1,243 @@
...
+
+	status := w.workerStatus()
+	if stage != metadata.StageFinished {
+		return w.UpdateStatus(ctx, status)
+	}
+
+	if w.workerType == lib.WorkerDMDump {
+		if err := w.persistStorage(ctx); err != nil {
+			log.L().Error("failed to persist storage", zap.Error(err))
+			return nil
```
amyangfei, line 210: Is there any side effect to return nil here

engine/servermaster/jobmanager.go
```diff
@@ -107,6 +113,83 @@ func (jm *JobManagerImplV2) PauseJob(ctx context.Context, req *pb.PauseJobReques
 	}}
 }
 
+// DebugJob implements proto/Master.DebugJob
+// Used for request/response with jobmaster
+func (jm *JobManagerImplV2) DebugJob(ctx context.Context, req *pb.DebugJobRequest) (resp *pb.DebugJobResponse) {
```
amyangfei, line 118: Will this be refactored or removed after OpenAPI is finished.

GMHDBJD: Yes

engine/sample/1m1e.yaml
```diff
@@ -9,6 +9,7 @@ services:
                         - "/df-master"
                         - "--config"
                         - "/config.toml"
+                        - "-L=debug"
```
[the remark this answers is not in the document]
GMHDBJD: Easier to troubleshoot problems

engine/executor/dm/worker.go
GMHDBJD: unitholder always return nil, because dm unit alwasy return nil

amyangfei (approved):

Issue #5513, opened 2022-05-23
engine: Tracking issue for merge file resource GC


liuzix:
- [x] Preparation #5514 
- [x] Core GC logic #5557 
- [x] Adjust JobManager #5559 
- [x] Adjust ExecutorManager #5565
- [x] Implement `Discard` and `OnWorkerClosed` #5623 
- [x] Integrate all parts 1) #5596 2) #5646

Issue #5494, opened 2022-05-20
Revert prometheus client_golang from  from v1.12.2 to v1.11.0


maxshuang:
### What did you do?

When merge engine to tiflow repo, client_golang(v1.12.2) will introduce some incompatible change to the go collector and some metric name.

**Detail**
ref : https://github.com/prometheus/client_golang/releases
[CHANGE] ⚠️ Change in collectors.NewGoCollector metrics: Reverting addition of new ~80 runtime metrics by default. You can enable this back with GoRuntimeMetricsCollection option or GoRuntimeMemStatsCollection | GoRuntimeMetricsCollection for smooth transition.
[BUGFIX] Fixed the bug that causes generated histogram metric names to end with _total. ⚠️ This changes 3 metric names in the new Go collector that was reverted from default in this release.
go_gc_heap_allocs_by_size_bytes_total -> go_gc_heap_allocs_by_size_bytes,
go_gc_heap_frees_by_size_bytes_total -> go_gc_heap_allocs_by_size_bytes
go_gc_pauses_seconds_total -> go_gc_pauses_seconds.

### What did you see instead?

some go collector metric lost

### Versions of the cluster

Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

```console
(paste TiDB cluster version here)
```

Upstream TiKV version (execute `tikv-server --version`):

```console
(paste TiKV version here)
```

TiCDC version (execute `cdc version`):

```console
master
```

Ehco1996:
reopen because https://github.com/pingcap/tiflow/pull/5393 upgrade promtheus_client again

PR #5514, opened 2022-05-23
externalresources(engine): Miscellaneous changes required for implementing resource GC.

liuzix:

Issue Number: ref #5513 

- Added required methods in `pkg/orm` (data accessing layer).
- Adjusted interface definitions in `pkg/externalresource/manager`.
- other related changes.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


go.mod
```diff
@@ -55,7 +55,7 @@ require (
 	github.com/pingcap/tidb v1.1.0-beta.0.20220412180037-d07b66ea638c
 	github.com/pingcap/tidb-tools v6.0.0-alpha.0.20220317013353-dfc5146f4746+incompatible
 	github.com/pingcap/tidb/parser v0.0.0-20220412180037-d07b66ea638c
-	github.com/prometheus/client_golang v1.11.0
+	github.com/prometheus/client_golang v1.12.2
```
amyangfei, line 58: Is it necessary to upgrade `client_golang`, seems there is breaking change in 1.12.2, ref: https://github.com/pingcap/tiflow/issues/5494

engine/servermaster/executor_manager.go
```diff
@@ -43,6 +44,7 @@ type ExecutorManager interface {
 	ListExecutors() []string
 	CapacityProvider() scheduler.CapacityProvider
 	GetAddr(executorID model.ExecutorID) (string, bool)
+	WatchExecutors(ctx context.Context) ([]model.ExecutorID, *notifier.Receiver[model.ExecutorID], error)
```
maxshuang, line 47: Could you explain what's the difference of the first param and second param?

engine/pkg/externalresource/manager/test_utils.go
```diff
@@ -0,0 +1,188 @@
...
+	delete(p.executorSet, resModel.ExecutorID(executorID))
+	p.notifier.Notify(model.ExecutorID(executorID))
+}
+
+// HasExecutor returns whether the mock contains the given executor.
+func (p *MockExecutorInfoProvider) HasExecutor(executorID string) bool {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+
+	if _, exists := p.executorSet[resModel.ExecutorID(executorID)]; exists {
```
lance6716, line 67: linter may say return `exists`

engine/pkg/orm/client.go
```diff
@@ -621,6 +636,41 @@ func (c *metaOpsClient) QueryResourcesByExecutorID(ctx context.Context, executor
 	return resources, nil
 }
 
+func (c *metaOpsClient) DeleteResourcesByExecutorID(ctx context.Context, executorID string) error {
+	tx := c.db.WithContext(ctx).Where("executor_id = ?", executorID).
+		Delete(&resourcemeta.ResourceMeta{})
+	if tx.Error != nil {
```
lance6716, line 642: seems we can directly return cerrors.ErrMetaOpFail.Wrap(tx.Error)

lance6716 (commented):
rest lgtm

engine/servermaster/executor_manager.go
lance6716: a snapshot and a continuous event sender on that snapshot. Also please add some comments.

maxshuang: `a continuous event sender on that snapshot`:
I notice we can get the `ExecutorID ` from this receiver, but still can't get what we can do if we can get a  continuous `ExecutorID` event.

engine/pkg/orm/client.go
liuzix: This will cause problems for the caller, because `cerrors.ErrMetaOpFail.Wrap(tx.Error)` returns `(*errors.Error)(nil)` instead of `nil` if `tx.Error == nil`.

engine/servermaster/executor_manager.go
liuzix: > `a continuous event sender on that snapshot`: I notice we can get the `ExecutorID ` from this receiver, but still can't get what we can do if we can get a continuous `ExecutorID` event. @lance6716

The stream of ExecutorIDs is meant to be offline events. Sorry for not making it clearer and I've made some changes in the latest commit.

engine/pkg/externalresource/manager/test_utils.go
liuzix: Done

engine/servermaster/executor_manager.go
maxshuang: Looking forward to your inplementation.
May need to take good care of the dynamic feature of `watching + snapshot` behavior.

lance6716 (approved):

PR #5518, opened 2022-05-23
pkg/log(dm):support reading fields from Ctx

### LINKED ISSUE #5517: engine: We need a better mechanism for structured logging
The problem:

Many of the logs in Data Flow Engine needs to be tagged with metadata of the request/job/tenant/worker, etc. So we need a mechanism to easily attach these fields to the logger.

- [ ] Add necessary features to `dm/pkg/log`.

liuzix:
Issue Number: close #5517

- Put zap fields to contexts with `context.WithValue`.
- Add `WithCtx()` method to `log` package, so that in each function call, the fields can be logged easily.
- Added relevant benchmarks.

#### Tests 

 - Unit test

- No regression for current production code, because no existing execution path is affected.
- Benchmark results are as follows
```
goos: linux
goarch: amd64
pkg: github.com/pingcap/tiflow/dm/pkg/log
cpu: AMD Ryzen 9 5900X 12-Core Processor            
BenchmarkBaseline
BenchmarkBaseline-24    	 9693552	       124.1 ns/op
BenchmarkWithCtx
BenchmarkWithCtx-24     	 6640200	       181.1 ns/op
PASS

Process finished with the exit code 0
```
Using `WithCtx` increases time cost by 50% compared to traditional, but more complex methods. But given the baseline high performance of `zap`, it should be fine as long as no logging is used in CPU intensive operations.

### Release note 

```release-note
None
```

### DIFF
--- a/dm/pkg/log/ctx.go
+++ b/dm/pkg/log/ctx.go
@@ -1,0 +1,47 @@
+// Copyright 2022 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package log
+
+import (
+	"context"
+
+	"go.uber.org/zap"
+)
+
+type logCtxKeyType string
+
+var logCtxKey = logCtxKeyType("zap-fields")
+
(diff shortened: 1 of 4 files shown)


amyangfei (approved):

amyangfei:
```
goos: linux
goarch: amd64
pkg: github.com/pingcap/tiflow/dm/pkg/log
cpu: AMD Ryzen 9 5900X 12-Core Processor            
BenchmarkBaseline
BenchmarkBaseline-24    	 9693552	       124.1 ns/op
BenchmarkWithCtx
BenchmarkWithCtx-24     	 6640200	       181.1 ns/op
PASS

Process finished with the exit code 0
```

Really fast CPU 🚀

lance6716 (approved):

PR #5520, opened 2022-05-23
lib/worker(engine): refactor BaseWorker (Part I)

liuzix:

Issue Number: ref #5521 

- Move `MasterClient` into a separate package.
- Remove `OnMasterFailOver`.

### Reviewer's Guide
- Focus on `engine/lib/worker/master_client.go`.
- The basic structure of `MasterClient` is the same as before, but with some asynchronous enhancement and minor bug fixes.
- Please make sure that removing `OnMasterFailOver` does not harm the correctness of the business logic.

#### Tests 

 - Unit test
 - Integration test

```release-note
None
```


GMHDBJD (approved):
LGTM

engine/lib/worker/master_client.go
```diff
@@ -0,0 +1,247 @@
...
+	sinceLastAcked := clk.Mono().Sub(lastMasterAckedPingTime)
+	if sinceLastAcked <= 2*m.timeoutConfig.WorkerHeartbeatInterval {
+		return true, nil
+	}
+
+	if sinceLastAcked > 2*m.timeoutConfig.WorkerHeartbeatInterval &&
+		sinceLastAcked < m.timeoutConfig.WorkerTimeoutDuration {
+
+		// We ignore the error here
+		_ = m.asyncReloadMasterInfo(context.Background())
```
amyangfei, line 201: Is it better to add a timeout, such as `WorkerTimeoutDuration`, in case of leaving some stale goroutines even worker has exited.

engine/lib/worker/master_client.go
```diff
@@ -0,0 +1,247 @@
...
+	log.L().Debug("sending heartbeat", zap.String("worker", m.workerID))
+	ok, err := m.messageSender.SendToNode(ctx, nodeID, libModel.HeartbeatPingTopic(m.masterID), heartbeatMsg)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	log.L().Info("sending heartbeat success", zap.String("worker", m.workerID),
+		zap.String("master-id", m.masterID))
+	if !ok {
+		// Reloads master info asynchronously.
+		_ = m.asyncReloadMasterInfo(context.Background())
```
amyangfei, line 232: ditto

engine/lib/worker/master_client.go
liuzix: > Besides why removing the context parameter in this function

The semantics of `CheckMasterTimeout` does not seem to require a context. It should always be non-blocking to avoid any surprises.

amyangfei (approved):

PR #5552, opened 2022-05-24
DM(dmctl/openapi): fix incremental task's source meta check

### LINKED ISSUE #5464: DM openapi create task pass the check about incremental task's source meta and cause other interface return error
### What did you do?

1. Use web ui create incremental task and not set source meta 
2. start task success. 
3. check task status.

check task status return normal

### What did you see instead?

check task status return error like `mysql-instance(xxx) must set meta for task-mode incremental, Please check the `meta` config in task configuration file.`

### Versions of the cluster

```console
master
```
```console
```
```console
```
```console
(leave TiUP or manually here)
```
```console
>
>
```

```console
```

WizardXiao:

Issue Number: close #5464 

pass the check of source meta in task's adjust and add check to start task or check task.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


dm/tests/dmctl_basic/conf/dm-task7.yaml
```diff
@@ -0,0 +1,104 @@
+---
```
buchuitoudegou, line 1: Why does this test case need such a complicated task configuration? Is it solely used to emulate the situation without source metadata😵?

buchuitoudegou (approved):
rest LGTM

dm/tests/dmctl_basic/conf/dm-task7.yaml
WizardXiao: i just copy it from other task config like `dm-task1` to `dm-task6`, but it can check not only without source meta and also make sure other configuration item to be normal.

Ehco1996 (approved):
LGTM

PR #5557, opened 2022-05-24
externalresources(engine): implement GCCoordinator & GCRunner

liuzix:

Issue Number: ref #5513

- Implemented `GCCoordinator`.
- Implemented `GCRunner`.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


engine/pkg/externalresource/manager/gc_coordinator.go
```diff
@@ -0,0 +1,239 @@
...
+
+	c.notifier.GCNotify()
+	return nil
+}
+
+func (c *DefaultGCCoordinator) gcByOfflineExecutorID(ctx context.Context, executorID model.ExecutorID) error {
+	log.L().Info("Cleaning up resources meta for offlined executor",
+		zap.String("executor-id", string(executorID)))
+	return c.metaClient.DeleteResourcesByExecutorID(ctx, string(executorID))
+}
```
amyangfei, line 239: If we don't need to perform GC operation for offline executor, does this mean all resources are bounded to one executor, what if we support S3 storage as resource in the future.

engine/pkg/externalresource/manager/gc_runner.go
```diff
@@ -0,0 +1,137 @@
...
+
+	handler, exists := r.gcHandlers[tp]
+	if !exists {
+		log.L().Warn("no gc handler is found for given resource type",
+			zap.Any("resource-id", res.ID))
+	}
+
+	if err := handler(ctx, res); err != nil {
+		return err
+	}
```
amyangfei, line 124: If not exists, `handler` is nil.

liuzix: Fixed

engine/pkg/externalresource/manager/gc_coordinator.go
liuzix: We will adjust this when we implement S3 support. The responsibility of checking whether to perform actual GC should belong to the gcHandler for the resource types.

engine/pkg/externalresource/manager/gc_coordinator.go
```diff
@@ -0,0 +1,239 @@
...
+			continue
+		}
+
+		err = c.runGCEventLoop(ctx, jobReceiver.C, executorReceiver.C)
+		jobReceiver.Close()
+		executorReceiver.Close()
+
+		if gerrors.Is(err, context.Canceled) || gerrors.Is(err, context.DeadlineExceeded) {
+			return errors.Trace(err)
+		}
```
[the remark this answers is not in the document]
liuzix: Added logging.

engine/pkg/externalresource/manager/gc_coordinator.go
```diff
@@ -0,0 +1,249 @@
...
+
+		jobReceiver, executorReceiver, err := c.initializeGC(ctx)
+		if err != nil {
+			log.L().Warn("GC error", zap.Error(err))
+			rl.Take()
+			continue
+		}
+
+		err = c.runGCEventLoop(ctx, jobReceiver.C, executorReceiver.C)
```
maxshuang, line 81: What condition will this loop run again?
I scan the whole pr and find runGCEventLoop will not return if no error happen.

engine/pkg/externalresource/manager/gc_coordinator.go
```diff
@@ -0,0 +1,249 @@
...
+	pkgOrm "github.com/pingcap/tiflow/engine/pkg/orm"
+)
+
+// DefaultGCCoordinator implements interface GCCoordinator.
+// It is responsible for triggering file resource garbage collection.
+type DefaultGCCoordinator struct {
+	executorInfos ExecutorInfoProvider
+	jobInfos      JobStatusProvider
+	metaClient    pkgOrm.ResourceClient
+	notifier      GCNotifier
```
maxshuang, line 39: Does it designed to notify the executor to do some real gc through rpc?

engine/pkg/externalresource/manager/gc_coordinator.go
liuzix: You are right. This loop is for retrying on errors.

maxshuang: So `Run` will be called multi-times?

liuzix: No, it will not.

maxshuang (approved):

PR #5561, opened 2022-05-24
worker[engine]: fix bug that worker impl call 'Exit', but may not return

### LINKED ISSUE #5523: [engine]: BaseWorker Exit() will not make worker exit if user don't return
### What did you do?

Implement WorkerImpl, call Exit() but forget to return  

worker exit as expected

### What did you see instead?

worker doesn't exit if user don't return

91J9RxCYhq

### Versions of the cluster

engine version:

master

maxshuang:

Issue Number: close #5523 

Set the error on errCenter when call 'Exit'

#### Tests 

 - Unit test

### Release note 

```release-note
`None`.
```

### DIFF
--- a/engine/lib/worker.go
+++ b/engine/lib/worker.go
@@ -429,11 +429,11 @@ func (w *DefaultBaseWorker) Exit(ctx context.Context, status libModel.WorkerStat
 	w.workerStatus.ErrorMessage = status.ErrorMessage
 	w.workerStatus.ExtBytes = status.ExtBytes
 	if errRet = w.statusSender.UpdateStatus(ctx, w.workerStatus); errRet != nil {
-		return errRet
+		return
 	}
 
 	errRet = derror.ErrWorkerFinish.FastGenByArgs()
-	return errRet
+	return
 }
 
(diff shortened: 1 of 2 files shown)


engine/lib/worker.go
```diff
@@ -415,19 +415,25 @@ func (w *DefaultBaseWorker) OpenStorage(ctx context.Context, resourcePath resour
...
 	w.workerStatus.ExtBytes = status.ExtBytes
-	if err1 := w.statusSender.UpdateStatus(ctx, w.workerStatus); err1 != nil {
-		return err1
+	if errRet = w.statusSender.UpdateStatus(ctx, w.workerStatus); errRet != nil {
+		return errRet
 	}
 
-	return derror.ErrWorkerFinish.FastGenByArgs()
+	errRet = derror.ErrWorkerFinish.FastGenByArgs()
+	return errRet
```
liuzix, line 436:
```suggestion
	return derror.ErrWorkerFinish.FastGenByArgs()
```

maxshuang: From `return errRet` to `return` to avoid useless copy

gozssky (approved):

Issue #4757, opened 2022-03-03
Tracking issue for reducing high latency


overvenus:
The status quo and the goal of the tracking issue.

Scenario | Status quo (2022/03/02) | Goal
-- | -- | --
**Planned outage** | |
TiCDC rolling upgrade | 2-8 mins | <= 10s
TiKV/TiDB rolling upgrade | <= 1min | <= 30s
PD rolling upgrade | <= 1min | <= 1min
TiKV scale out/off | <= 1min | <= 30s
TiCDC scale out/off | 2-8 mins | <= 5s
PD scale out/off | <= 1min | <= 1min
TiDB scale out/off | <= 1min | <= 30s
**Unplaned outage**<br>(less than ⅓ total node number, outage last 5mins) | |
TiKV/TiDB outage (power down, disk failure etc) | 2-4 mins | <= 2min
TiCDC  outage (power down, disk failure etc) | 2-8 mins | 2-8 mins
PD  outage (power down, disk failure etc) | 2-8 mins | 2-8 mins
**No outage** | |
Delay | <= 10s | <= 2s
Delay spikes (99%) | <= 1 min | <= 5s

## Extreme high latency (>30 min)

- [x] https://github.com/pingcap/tiflow/issues/4516

## Latency spike

- [x] https://github.com/pingcap/tiflow/issues/4756
- [x] https://github.com/pingcap/tiflow/issues/4761
- [x] https://github.com/pingcap/tiflow/issues/4762
- [x] https://github.com/pingcap/tiflow/issues/3529
- [x] https://github.com/tikv/tikv/issues/12166
- [x] #7277
 

## Two-phase scheduling

Two-phase scheduling aims to solve high latency spike (up to minutes) caused by moving table (move a table from a TiCDC to another). 

- [x] Cleanup scheduler v1 code https://github.com/pingcap/tiflow/pull/5362
- [x] Cleanup schedulerv2 interfaces in owner and processor https://github.com/pingcap/tiflow/pull/5396
- [x] Sketch two-phase scheduler (task based scheduling, move table task, add table task, remove table task) https://github.com/pingcap/tiflow/pull/5450 
- [x] Implement two-phase replication_set state transition https://github.com/pingcap/tiflow/pull/5450
- [x] Implement replicationManager https://github.com/pingcap/tiflow/pull/5562
- [x] Implement two-phase scheduler coordinator https://github.com/pingcap/tiflow/pull/5710
[...]

PR #5562, opened 2022-05-24
tp: implement capture manager

overvenus:

Issue Number: ref #4757 

Implement capture manager to mange capture membership.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/scheduler/internal/tp/capture_manager.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func newCaptureStatus(rev schedulepb.OwnerRevision) *CaptureStatus {
+	return &CaptureStatus{OwnerRev: rev, State: CaptureStateUninitialize}
+}
+
+func (c *CaptureStatus) handleHeartbeatResponse(
+	resp *schedulepb.HeartbeatResponse, epoch schedulepb.ProcessorEpoch,
+) {
+	epochMismatch := c.State != CaptureStateUninitialize &&
```
3AceShowHand, line 58: `epochMismatch` is only used once, and the evalution is not easy to read.

How about directly check the condition in the Line 60.

cdc/scheduler/internal/tp/capture_manager.go
```diff
@@ -0,0 +1,160 @@
...
+		Captures:      make(map[string]*CaptureStatus),
+		heartbeatTick: heartbeatTick,
+	}
+}
+
+func (c *captureManager) captureTableSets() map[model.CaptureID]*CaptureStatus {
+	return c.Captures
+}
+
+func (c *captureManager) checkCaptureInitialized() bool {
```
3AceShowHand, line 98: how about rename it to `allCapturesInitialized`

cdc/scheduler/internal/tp/capture_manager.go
```diff
@@ -0,0 +1,160 @@
...
+	allInitialized := true
+	for _, captrueStatus := range c.Captures {
+		if captrueStatus.State == CaptureStateUninitialize {
+			allInitialized = false
+		}
+	}
```
3AceShowHand, line 104: this can be further simplified.

no need to check all captures before return `false`, and `allInitialized` is unnecessary.

3AceShowHand (approved):

PR #5563, opened 2022-05-24
[DNM] openapi(engine): implement openapi

gozssky:

Issue Number: close #xxx

This PR is a prototype of openapi implementation for dataflow engine. 
I will finish the whole implementation in multiple PRs, since there are a lot of changes.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/executor/server.go
```diff
@@ -336,6 +350,20 @@ func (s *Server) Run(ctx context.Context) error {
 		return s.taskRunner.Run(ctx)
 	})
 
+	wg.Go(func() error {
```
liuzix, line 353: Can this be moved into the implementation of `jobAPIServer`? This way it would be easier for unit testing.

engine/servermaster/openapi.go
```diff
@@ -0,0 +1,201 @@
...
+	apiOpVarTenantID = "tenant_id"
+	// apiOpVarProjectID is the key of project id in HTTP API.
+	apiOpVarProjectID = "project_id"
+	// apiOpVarJobID is the key of job id in HTTP API.
+	apiOpVarJobID = "job_id"
+)
+
+// OpenAPI provides API for servermaster.
+type OpenAPI struct {
+	server *Server
```
liuzix, line 40: We can abstract out an interface for `Server`.

engine/servermaster/openapi.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		return
+	}
+	if resp.JobMasterInfo == nil {
+		c.AbortWithError(http.StatusInternalServerError, errors.New("couldn't find job master info"))
+		return
+	}
+	// TODO: check tenant id and project id.
+	executorID := model.ExecutorID(resp.JobMasterInfo.ExecutorId)
+	addr, ok := o.server.executorManager.GetAddr(executorID)
```
liuzix, line 184: You can move the `ExecutorInfoProvider` interface to a common package and modify it to let it provide the information you need, such as the address of executors.

engine/executor/server.go
```diff
@@ -193,6 +201,12 @@ func (s *Server) makeTask(
 		log.L().Error("Failed to create worker", zap.Error(err))
 		return nil, err
 	}
+	if jm, ok := newWorker.(lib.BaseJobMaster); ok {
+		engine := gin.New()
+		apiGroup := engine.Group(jobAPIBasePath(jm.JobMasterID()))
+		jm.OnOpenAPIInitialized(apiGroup)
```
liuzix, line 207: `OnOpenAPIInitialized` is called before `Init`. Could you add some comments?

gozssky: I will add it later.

engine/executor/server.go
gozssky: Yes, it is more reasonable.

engine/executor/openapi.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/http"
+	"strings"
+	"sync"
+
+	libModel "github.com/pingcap/tiflow/engine/lib/model"
+)
+
+const jobAPIPrefix = "/api/v1/jobs/"
+
+func jobAPIBasePath(masterID libModel.MasterID) string {
```
maxshuang, line 26: Use a `jobID` may be better. `MasterID` is the inner implement concept.

maxshuang: Pls make an alias of `JobID = MasterID` in engine/model.

engine/servermaster/http.go
```diff
@@ -0,0 +1,58 @@
...
+	pprofGroup.GET("/trace", gin.WrapF(pprof.Trace))
+	pprofGroup.GET("/threadcreate", gin.WrapF(pprof.Handler("threadcreate").ServeHTTP))
+
+	// Failpoint API
+	if util.FailpointBuild {
+		// `http.StripPrefix` is needed because `failpoint.HttpHandler` assumes that it handles the prefix `/`.
+		router.Any("/debug/fail/*any", gin.WrapH(http.StripPrefix("/debug/fail", &failpoint.HttpHandler{})))
+	}
+
+	// Promtheus metrics API
```
maxshuang, line 56: Why we need to regiter `metrics` twice?

engine/executor/openapi.go
gozssky: OK.

engine/lib/base_jobmaster.go
```diff
@@ -79,6 +81,7 @@ type JobMasterImpl interface {
 
 	Workload() model.RescUnit
 	OnJobManagerMessage(topic p2p.Topic, message interface{}) error
+	OnOpenAPIInitialized(apiGroup *gin.RouterGroup)
```
maxshuang, line 84: Why `JobMasterImpl`  has the `OnOpenAPIInitialized ` as the `BaseJobMaster `?

engine/lib/base_jobmaster.go
```diff
@@ -237,6 +240,11 @@ func (d *DefaultBaseJobMaster) CurrentEpoch() libModel.Epoch {
 	return d.master.currentEpoch.Load()
 }
 
+// OnOpenAPIInitialized implements BaseJobMaster.BaseJobMaster
```
maxshuang, line 243: This logic looks strange.
`OnOpenAPIInitialized ` is the thing framework require user to implement.

Just call this `impl.OnOpenAPIInitialized()`. in  BaseJobMaster.Init() logic

engine/lib/base_jobmaster.go
amyangfei: `BaseJobMaster.OnOpenAPIInitialized` is triggered from the perspective of executor, and `JobMasterImpl.OnOpenAPIInitialized` does the real work that passes router group, they are two functions actually. Maybe we can use two names to make it easier to understand.

engine/lib/base_jobmaster.go
amyangfei, line 243:
```suggestion
// OnOpenAPIInitialized implements BaseJobMaster.OnOpenAPIInitialized
```

engine/lib/base_jobmaster.go
amyangfei: `BaseJobMaster.OnOpenAPIInitialized` is triggered from the perspective of executor, and `JobMasterImpl.OnOpenAPIInitialized` does the real work that passes router group, they are two functions actually. Maybe we can use two names to make it easier to understand.

amyangfei: Can add a comment to this function that user logic must not access `apiGroup` after this function returns, otherwise data race could happen.

gozssky:
The whole implementation will be tracked in https://github.com/pingcap/tiflow/issues/5689.

Issue #5504, opened 2022-05-20
Tracking issue for implementing DM into engine


GMHDBJD:
# This issue track the progress of DM
- [ ] precheck
  - [x] task cfg adjust
  - [ ] source cfg adjust
  - [ ] database adjust
  - [x] checker
- [ ] shard ddl
- [ ] command
  - [ ] binlog
  - [ ] binlog-schema
  - [ ] config
  - [ ] relay
  - [ ] operate-task
  - [ ] query-status
    - [x] task-status
    - [ ] source-status
  - [ ] shard-ddl-lock
  - [ ] validation
- [ ] metrics
- [ ] openapi
- [ ] upgrade
- [ ] error code
- [ ] log
- [ ] test 
- [ ] Unified metadata/config format（yaml/toml/json)

lance6716:
"error code" development will be postponed until Web UI has decided the error format.

PR #5581, opened 2022-05-25
e2e(engine): test two running DM jobs

lance6716:

Issue Number: ref #5504 

add test of two running jobs, so we can test about metrics later

#### Tests 

 - No code

### Release note 

```release-note
 `None`.
```


GMHDBJD (approved):
LGTM

PR #5593, opened 2022-05-25
scheduler(2pc): agent for 2 phase scheduling

3AceShowHand:

Issue Number: ref #4757 

2 phase scheduling agent implementation.

#### Tests 

 - Unit test

### Release note 

```release-note
`None`.
```


dm/pkg/schema/tracker_test.go
```diff
@@ -662,7 +662,7 @@ func (s *trackerSuite) TestAllSchemas(c *C) {
 			table := schema.Tables[0]
 			c.Assert(table.Name.O, Equals, "a")
 			c.Assert(table.Columns, HasLen, 1)
-			// the table should be equivalent to the result of GetTable.
+			// the table should be equivalent to the result of GetTableMeta.
```
overvenus, line 665: No need to modify dm files.

cdc/scheduler/internal/tp/schedulepb/table_schedule_manual_test.go
```diff
@@ -14,77 +14,87 @@
...
 		}
+
 		for i := 0; i < total; i++ {
-			msg.Checkpoints[int64(10000+i)] = Checkpoint{
-				CheckpointTs: 433331421532337260,
-				ResolvedTs:   433331421532337261,
-			}
+			msg.HeartbeatResponse.Tables = append(msg.HeartbeatResponse.Tables,
+				TableStatus{
+					TableID: model.TableID(i),
```
overvenus, line 38:
```suggestion
					TableID: model.TableID(10000+i),
```

Let's close to the real case. Table ID starts from a large number for the this case.

cdc/scheduler/internal/tp/replication_set.go
```diff
@@ -573,6 +576,7 @@ func (r *ReplicationSet) handleRemoveTable() ([]*schedulepb.Message, error) {
 			zap.Any("replicationSet", r), zap.Int64("tableID", r.TableID))
 		return nil, nil
 	}
+	// todo: OldState must be `replicating` here
```
overvenus, line 579: How do we deal with these todo?

cdc/scheduler/internal/tp/agent.go
```diff
@@ -27,31 +40,589 @@ type agent struct {
...
 	tableExec internal.TableExecutor
 
-	tables       map[model.TableID]*schedulepb.TableStatus
-	runningTasks map[model.TableID]*schedulepb.Message
+	// runningTasks track all in progress dispatch table task
+	runningTasks map[model.TableID]*dispatchTableTask
+
+	// pendingTasks is a queue of dispatch table task yet to be processed.
+	// the Deque stores *dispatchTableTask.
+	pendingTasks deque.Deque
```
overvenus, line 48: Why do we need a deque? can we handle tasks immediately? Owner control the total number of task that are running concurrently.

cdc/scheduler/internal/tp/replication_set.go
3AceShowHand: It looks `OldState` is only used for logging, we can remove it from the log, since we know that it's must be in desired state.

cdc/scheduler/internal/tp/agent.go
```diff
@@ -27,31 +40,589 @@ type agent struct {
...
+// id: the incoming owner's capture ID
+// revision: the incoming owner's revision as generated by Etcd election.
+// version: the incoming owner's semantic version string
+func (a *agent) handleOwnerInfo(id model.CaptureID, revision int64, version string) bool {
+	if a.ownerInfo.revision.Revision == revision {
+		if a.ownerInfo.captureID != id {
+			// This panic will happen only if two messages have been received
+			// with the same ownerRev but with different ownerIDs.
+			// This should never happen unless the election via Etcd is buggy.
+			log.Panic("owner IDs do not match",
```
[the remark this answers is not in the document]
3AceShowHand: This does not panic, because message from different owner should have different revision.

This is for two owner with `the same revision, but different owner ID`, which is unrecoveriable etcd error.

overvenus (approved):

PR #5596, opened 2022-05-25
externalresource(engine): Integrate all parts related to GC (Part I)

liuzix:

Issue Number: ref #5513 

- Implemented `LocalFileResourceController`
- Implemented `mockCluster` and some integration tests for GC.
- Minor adjustments of method names.
- Fixed minor bugs.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None.
```


engine/pkg/externalresource/resourcetypes/local_file.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/pingcap/tiflow/engine/client"
+	"github.com/pingcap/tiflow/engine/pb"
+	derrors "github.com/pingcap/tiflow/engine/pkg/errors"
+	resModel "github.com/pingcap/tiflow/engine/pkg/externalresource/resourcemeta/model"
+)
+
+// LocalFileResourceType defines operations specific to
+// the local file type.
+type LocalFileResourceType struct {
```
maxshuang, line 27: This name sounds like a Type.

liuzix: You are right. I will change it.

maxshuang:
rest LGTM.

engine/pkg/externalresource/resourcetypes/interfaces.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"context"
+
+	resModel "github.com/pingcap/tiflow/engine/pkg/externalresource/resourcemeta/model"
+)
+
+// ResourceTypeController is an interface providing relevant operations related
+// to one resource type.
+type ResourceTypeController interface {
```
maxshuang, line 24: IMHO, why not the `ResourceTypeHandler` since it only has `GetHandler` interface?

maxshuang (approved):

engine/client/executor_client.go
```diff
@@ -15,19 +15,31 @@ package client
...
+
+	"github.com/pingcap/tiflow/engine/pb"
+	"github.com/pingcap/tiflow/pkg/retry"
 )
 
 // ExecutorClient defines an interface that supports sending gRPC from server
 // master to executor.
 type ExecutorClient interface {
-	baseExecutorClient
+	// baseExecutorClient
```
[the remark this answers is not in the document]
liuzix: Yes. It seems `baseExecutorClient` should not be part of the public interface.

engine/pkg/externalresource/resourcetypes/interfaces.go
liuzix: More methods are planned in future PRs.

amyangfei (approved):

PR #5601, opened 2022-05-26
codec(ticdc): use handlekey instead of primary key for avro

### LINKED ISSUE #5597: avro codec should use unique key as message key when there is no primary key
### What did you do?

N/A

as title

### What did you see instead?

currently avro would only use primary key as message key

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

zhangyangyu:

Issue Number: close #5597 

Avro shall use Handle Key as message key in case when there is no primary key, we could still try to use a not null unique key , makes downstream easy to use these events.

When there is no handle key, such as force replica situations, Avro message keys would be null. Downstream consumers need to accomodate for this. (This is already the current behavriour)

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/model/sink.go
+++ b/cdc/model/sink.go
@@ -325,15 +325,12 @@ func (r *RowChangedEvent) HandleKeyColumns() []*Column {
 		}
 	}
 
-	if len(pkeyCols) == 0 {
-		log.Panic("Cannot find handle key columns.", zap.Any("event", r))
-	}
-
+	// It is okay not to have handle keys, so the empty array is an acceptable result
 	return pkeyCols
 }
 
-// PrimaryKeyColInfos returns the column(s) and colInfo(s) corresponding to the primary key(s)
-func (r *RowChangedEvent) PrimaryKeyColInfos() ([]*Column, []rowcodec.ColInfo) {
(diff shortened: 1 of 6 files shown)


3AceShowHand (approved):

lance6716:
> we could still try to use a not null unique key

I remember handle key is not a "not null unique key", please check it @3AceShowHand

zhangyangyu:
> > avro codec should use unique key as message key when there is no primary key
> 
> > we could still try to use a not null unique key
> 
> I remember handle key is not a "not null unique key", please check it @3AceShowHand

how do you think @leoppro

zhaoxinyu (approved):

liuzix:
There should be an end-to-end test for this issue. Holding for now.

lance6716:
> > > avro codec should use unique key as message key when there is no primary key
> > 
> > 
> > > we could still try to use a not null unique key
> > 
> > 
> > I remember handle key is not a "not null unique key", please check it @3AceShowHand
> 
> how do you think @leoppro

I have checked the code, the HandleKey flag is fine.

maxshuang (approved):

PR #5623, opened 2022-05-28
externalresource(engine): implement Discard local resource & OnWorkerClosed

liuzix:
Issue Number: ref #5513

- Implemented removing temporary files when `OnWorkerClosed` is called.
- Implemented `Discard()` on the resource handle.
- Refactored the implementation of the resource handle.
- Use hex-encoding to generate paths on the local file system.

#### Tests 

 - Unit test
 - Integration test

#### Questions 

```release-note
None.
```


amyangfei (approved):

engine/pkg/externalresource/broker/file_manager.go
```diff
@@ -163,7 +174,7 @@ func (m *LocalFileManager) RemoveResource(creator libModel.WorkerID, resName res
 	}
 
 	// Note that the resourcePath is actually a directory.
-	if err := os.RemoveAll(resourcePath); err != nil {
+	if err := os.RemoveAll(filePath); err != nil {
 		return derrors.ErrRemovingLocalResource.Wrap(err)
```
maxshuang, line 178: Our local file resource is a file or a directory?
I think `resourcePath` is a more broad concept which we can have diferent implement. 
Why we change the name?

liuzix: - Our resource is a directory. This design choice is made so that the `brStorage` can be easily reused.
- `filePath` is for `the path of the resource directory in the file system`.

maxshuang: Sounds a little ambiguity for `filePath` here. 
Pls modify the comment for it. `Note that the resourcePath is actually a directory.`

rest LGTM.

maxshuang (approved):

PR #5646, opened 2022-05-30
servermaster,executor(engine): Enable local file resource GC

liuzix:

Issue Number: ref #5513 

- Setting up required components in both Servermaster and Executor.
- Made Executor's local data path configurable.
- e2e test will be added in another PR.

#### Tests 

 - Unit test
 - Integration test (only no regression is guaranteed)

No.

No.

### Release note 

```release-note
None.
```


engine/servermaster/server.go
```diff
@@ -748,6 +754,24 @@ func (s *Server) runLeaderService(ctx context.Context) (err error) {
...
 
+	s.gcRunner = externRescManager.NewGCRunner(s.frameMetaClient, map[resModel.ResourceType]externRescManager.GCHandlerFunc{
+		"local": resourcetypes.NewLocalFileResourceType(clients).GCHandler(),
+	})
+	s.gcCoordinator = externRescManager.NewGCCoordinator(s.executorManager, s.jobManager, s.frameMetaClient, s.gcRunner)
+
+	errg, errgCtx := errgroup.WithContext(ctx)
+	defer func() {
+		err := errg.Wait()
+		log.L().Error("errgroup exited with error", zap.Error(err))
```
amyangfei, line 765: Is it better to check `err` is not nil and then log error.

amyangfei:
And please resolve conflicts, rest lgtm

engine/executor/config.go
```diff
@@ -216,3 +216,10 @@ func (c *Config) configFromFile(path string) error {
 	}
 	return nil
 }
+
+func getDefaultLocalStorageDir(executorName string) string {
+	// Use hex encoding in case there are special characters in the
+	// executor name.
+	encodedExecutorName := hex.EncodeToString([]byte(executorName))
+	return filepath.Join(defaultLocalStorageDirPrefix, encodedExecutorName)
```
lance6716, line 224: Do we need to check existence of the path, or use ioutil.TempDir to make sure always use a new directory?

lance6716 (approved):

amyangfei (approved):

Issue #5493, opened 2022-05-20
[engine] Tracking issue for multi-projects/tasks metric/log/meta


maxshuang:
### What did you do?

Tracking todolist for multi-projects/tasks metric/log/meta:
1. multi-projects support
- [x] #5590

2. metric 
- [x] support multi-projects/tasks metric
- [x] #5496
- [x] #5525

3. log
- [x] #5608
- [x] replace user log with multi-projects/tasks log
   - [x] replace the framework dm/log initialization with pingcap/log
   - [x] replace framework component to support multi-projects log 

4. meta
- [x] framework and user meta support multi-projects
- [x] user meta support sql backend
- [x] adjust resourcemeta to support multi-projects

PR #5652, opened 2022-05-30
syncer(dm): prepare promutil.Factory injection and cache metrics

lance6716:

Issue Number: ref #5493

syncer will check the injected Factory to create new metrics. And cache some metrics to avoid calling some heavy functions in hot path.

#### Tests 

 - pass original test. Test about injected Factory will come in future PR.
 - benchbot (86s vs 115s) origin_img_v2_f6898c67-0411-442c-9c1b-2bd350f3494g

### Release note 

```release-note
 `None`.
```


lance6716:
PTAL @gozssky , similar to lightning's change

dm/syncer/dbconn/db.go
```diff
@@ -84,7 +85,12 @@ func (conn *DBConn) ResetConn(tctx *tcontext.Context) error {
 }
 
 // QuerySQL does one query.
-func (conn *DBConn) QuerySQL(tctx *tcontext.Context, query string, args ...interface{}) (*sql.Rows, error) {
+func (conn *DBConn) QuerySQL(
+	tctx *tcontext.Context,
+	metricProxies *metrics.Proxies,
```
gozssky, line 90: Can we put `metricsProxies` in `tctx`?

lance6716: tcontext is a common package, not only for syncer's metrics. And some functions don't have a `tctx` argument, so I put metrics outside `tctx`

gozssky (approved):

dm/dm/config/subtask.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/pingcap/tidb-tools/pkg/column-mapping"
 	extstorage "github.com/pingcap/tidb/br/pkg/storage"
 	"github.com/pingcap/tidb/util/filter"
+	"github.com/pingcap/tiflow/engine/pkg/promutil"
```
Ehco1996, line 30: this import order is wrong 🤔

dm/syncer/metrics/metrics.go
```diff
@@ -30,302 +31,350 @@ const (
 	BinlogEventCostStageGenQuery      = "gen-query"
 )
 
-// below variables are exported to syncer package.
-var (
-	BinlogReadDurationHistogram = metricsproxy.NewHistogramVec(
+// Metrics groups syncer's metric variables.
+type Metrics struct {
```
Ehco1996, line 35: how about renaming to `TaskMetrics`

dm/syncer/metrics/metrics.go
```diff
@@ -30,302 +31,350 @@ const (
...
+	RemainingTimeGauge              prometheus.Gauge
+	ShardLockResolving              prometheus.Gauge
+	FinishedTransactionTotal        prometheus.Counter
+	FlushCheckPointsTimeInterval    prometheus.Observer
+}
+
+// Proxies provides the ability to clean Metrics values when syncer is closed.
+type Proxies struct {
+	Metrics                         *Metrics
+	BinlogReadDurationHistogram     *metricsproxy.HistogramVecProxy
```
Ehco1996, line 59: it seems that only `Metrics` is accessed by callers, maybe make other members like  `BinlogReadDurationHistogram` to private members?

Ehco1996 (commented):
rest LGTM

dm/syncer/metrics/metrics.go
lance6716: some metrics are not a property of a single task, for example BinlogMasterPosGauge. Welcome to give a better name 😂

dm/syncer/metrics/metrics.go
lance6716: some variables like UnsyncedTableGauge are accessed in syncer package, I'll make rest of them private.

dm/syncer/metrics/metrics.go
Ehco1996: In this case I do not have a good name too, the name is too difficult 😭

Ehco1996 (approved):
LGTM

PR #5661, opened 2022-05-30
log(engine): support multi-projects/tasks  log

### LINKED ISSUE #5608: support zap log wrapper for multi-projects/tasks log

maxshuang:

Issue Number: close #5608 

Return logger with const fields

#### Tests 

 - Unit test

### Release note 

```release-note
 `None`.
```


engine/pkg/logutil/log.go
```diff
@@ -0,0 +1,49 @@
+package logutil
```
lonng, line 1: The license header is missing.

engine/lib/model/typedefs.go
```diff
@@ -24,6 +24,8 @@ type (
 	WorkerID = string
 	// Epoch is an increasing only value
 	Epoch = int64
+	// JobID is the alias of MasterID
+	JobID = MasterID
```
lonng, line 28: This code makes me confused because the `MasterID` is an alias of `string` and we make `JobID` as an alias of `MasterID`. Do you want both `JobID` and `MasterID` to be the same type as `string`?

maxshuang: Yes. They need to be the same type as `string`.
`JobID` is some kind of outer concept for user.
`MasterID` is the inner concept for framework schedule。

engine/pkg/logutil/log.go
maxshuang: Solved in latest commit.

engine/pkg/logutil/log.go
```diff
@@ -0,0 +1,62 @@
...
+	// constFieldJobKey is used to recognize jobs of the same job type
+	constFieldJobKey = "job_id"
+	// FieldWorkerKey is used to recognize workers of the same job
+	constFieldWorkerKey = "worker_id"
+)
+
+// NewLogger4Framework return a new logger for framework
+func NewLogger4Framework() log.Logger {
+	return log.L().WithFields(
+		zap.String(constFieldFrameworkKey, constFieldFrameworkValue),
```
GMHDBJD, line 41: Can we use zap.Bool?

GMHDBJD (approved):
rest LGTM

engine/lib/model/typedefs.go
```diff
@@ -24,7 +24,9 @@ type (
 	WorkerID = string
 	// Epoch is an increasing only value.
 	Epoch = int64
-	// JobType is the type for job.
+	// JobID is the alias of MasterID
+	JobID = MasterID
```
gozssky, line 28: JobID is redeclared below.

engine/pkg/logutil/log.go
```diff
@@ -0,0 +1,62 @@
...
+	constFieldFrameworkValue = true
+
+	/// app const label
+	// constFieldTenantKey and constFieldProjectKey is used to recognize metric for tenant/project
+	constFieldTenantKey  = "tenant"
+	constFieldProjectKey = "project_id"
+	// constFieldJobKey is used to recognize jobs of the same job type
+	constFieldJobKey = "job_id"
+	// FieldWorkerKey is used to recognize workers of the same job
+	constFieldWorkerKey = "worker_id"
```
amyangfei, line 35: Have a question here, what if business logs the same key, such as
`["worker test"] [tenant=tenant1] [project_id=proj1] [job_id=job1] [worker_id=worker1] [type=worker][worker_id=dm-worker-1]`

engine/pkg/logutil/log_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/hpcloud/tail"
+	"github.com/pingcap/log"
+	"github.com/pingcap/tiflow/engine/pkg/tenant"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+var testFile = "test.log"
+
+func init() {
```
amyangfei, line 30: In server logic, we still need to initialize global logger via the DM API, such as https://github.com/pingcap/tiflow/blob/84145a17b4ea207125c1b9410458a2b4efd428d5/engine/cmd/master/main.go#L50-L55

Is there any side effect?

maxshuang: I will replace the initialize global logger part after this pr is merged.

engine/lib/model/typedefs.go
maxshuang: Solved in lastest commit.

engine/pkg/logutil/log.go
maxshuang: Our log may be used  by Es, so even if we have the same key here, we still can scan the oriented log.

amyangfei (approved):

Issue #5089, opened 2022-04-01
DM worker can't exit in CI


lance6716:
### Which jobs are flaking?

IT-ha_cases3_1

### Which test(s) are flaking?

DM IT

### Jenkins logs or GitHub Actions link

```
[2022-03-31T09:47:44.177Z] kill dm-worker2
[2022-03-31T09:47:44.177Z] kill: sending signal to 1344 failed: No such process
[2022-03-31T09:47:44.177Z] [Thu Mar 31 17:47:44 CST 2022] <<<<<< START DM-WORKER on port 8263, config: /home/jenkins/agent/workspace/dm_ghpr_integration_test/go/src/github.com/pingcap/tiflow/dm/tests/ha_cases3_1/conf/dm-worker2.toml >>>>>>
[2022-03-31T09:47:44.438Z] dmctl test cmd: "pause-task test"
[2022-03-31T09:47:47.719Z] got=3 expected=3
[2022-03-31T09:47:47.974Z] dmctl test cmd: "resume-task test"
[2022-03-31T09:47:47.974Z] got=3 expected=3
[2022-03-31T09:49:39.369Z] use sync_diff_inspector to check increment data
[2022-03-31T09:49:39.369Z] check diff successfully
[2022-03-31T09:49:39.369Z] [Thu Mar 31 17:49:29 CST 2022] <<<<<< finish test_isolate_master_and_worker >>>>>>
[2022-03-31T09:49:39.369Z] 5 dm-master alive
[2022-03-31T09:49:39.369Z] 6 dm-worker alive
[2022-03-31T09:49:39.369Z] 0 dm-syncer alive
...
[2022-03-31T09:50:11.859Z] wait process dm-worker.test exit...
[2022-03-31T09:50:12.794Z] wait process dm-worker.test exit...
[2022-03-31T09:50:13.728Z] wait process dm-worker.test exit...
[2022-03-31T09:50:14.661Z] wait process dm-worker.test exit...
[2022-03-31T09:50:14.661Z] process dm-worker.test didn't exit after 30 seconds
```

DM worker log : 
[worker.log](https://github.com/pingcap/tiflow/files/8394408/worker.log)

### Anything else we need to know

- Does this test exist for other branches as well?

- Has there been a high frequency of failure lately?

lance6716:
catch it in my local PC!

[dlv.txt](https://github.com/pingcap/tiflow/files/8644391/dlv.txt)

lance6716:
[...]

PR #5667, opened 2022-05-31
test(dm): show full result of ps on kill fail

D3Hunter:

Issue Number: ref #5089

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
`None`.
```


dm/tests/_utils/wait_process_exit
```diff
@@ -16,5 +16,5 @@ while [ $WAIT_COUNT -lt 120 ]; do
 	((WAIT_COUNT++))
 done
 
-echo "process $process didn't exit after 120 seconds, current processlist: $(pgrep $process)"
+echo "process $process didn't exit after 120 seconds, current processlist: $(ps -ef | grep $process | grep -v 'grep')"
```
lance6716, line 19: maybe `ps $process` is enough

lance6716 (approved):

dm/tests/_utils/wait_process_exit
D3Hunter: > maybe `ps $process` is enough.

`ps` can't filter by name, has to grep

buchuitoudegou (approved):
lgtm. It seems this PR isn't about to resolve the issue but rather probe the root cause when it happens again🤣

PR #5672, opened 2022-05-31
promutil(engine): avoid embedding Mutex and prometheus.Registry in Registry

gozssky:

Issue Number: ref https://github.com/pingcap/tiflow/issues/5673

Problem Summary:

Avoid embedding `Mutex` and `prometheus.Registry` in `Registry`. 

Refer to uber go style:
* https://github.com/uber-go/guide/blob/master/style.md#avoid-embedding-types-in-public-structs
* https://github.com/uber-go/guide/blob/master/style.md#zero-value-mutexes-are-valid

Make `Mutex` and `prometheus.Registry` become private normal members.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/engine/pkg/promutil/registry.go
+++ b/engine/pkg/promutil/registry.go
@@ -16,10 +16,11 @@ package promutil
 import (
 	"sync"
 
-	libModel "github.com/pingcap/tiflow/engine/lib/model"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


lance6716 (approved):
/cc @maxshuang

maxshuang (approved):

Issue #5655, opened 2022-05-30
changefeed may stuck since table actor poll too slow. 


asddongmen:
### What did you do?

0. deploy a cdc cluster with 3 nodes
1.  create a changefeed sink to kafka
```
./cdc cli changefeed create --pd=*** --sink-uri="kafka://[***/default_topic?protocol=canal-json](http://***/default_topic?protocol=canal-json)" --config=/tmp/cf-config --changefeed-id=canal-json-test
```
2. write data 
```
sysbench oltp_write_only --db-driver=mysql --mysql-host=[***](http://***/) --mysql-port=4000 --mysql-user=root --mysql-password='***' --mysql-db=sbtest2 --threads=20 --table_size=100000 --tables=20 --report-interval=10 --time=86400 run
```

### What did you expect to see?

cdc work normanly 

### What did you see instead?

changefeed stuck
![origin_img_v2_1207e011-0f6b-4bc4-af0f-fefa9925670g](https://user-images.githubusercontent.com/20351731/170987020-67646bbd-78fa-46cf-a57a-30625dc39fa8.png)

### Versions of the cluster

Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

```console
(paste TiDB cluster version here)
```

Upstream TiKV version (execute `tikv-server --version`):

```console
(paste TiKV version here)
```

TiCDC version (execute `cdc version`):

```console
master branch
```

PR #5675, opened 2022-05-31
sink/mq(ticdc): make EmitCheckpointTs and FlushRowChangedEvents non-blocking

hi-rustin:

Issue Number: ref https://github.com/pingcap/tiflow/issues/5655

We use an unbounded buffer.
- make EmitCheckpointTs non-blocking
- make FlushRowChangedEvents non-blocking

#### Tests 

 - Unit test

Maybe, I will test it.
No
### Release note 

```release-note
Fix the problem that sending data to Kafka is very slow and stuck
```


cdc/sink/mq/mq_flush_worker.go
```diff
@@ -201,9 +192,6 @@ func (w *flushWorker) asyncSend(
 // until it encounters an error or is interrupted.
 func (w *flushWorker) run(ctx context.Context) (retErr error) {
 	defer func() {
-		// Note that errCh is sent to exactly once.
-		w.errCh <- retErr
```
nongfushanquan, line 205: why remove the retErr?

hi-rustin: > * remove flush worker errch, because we do not need it. We now send messages to msgchan without blocking anymore.

cc: @nongfushanquan

hi-rustin: Also, it was supposed to solve the problem of buffer sink getting stuck at that time. It doesn't exist now either.

cdc/sink/mq/mq.go
```diff
@@ -342,6 +338,7 @@ func (k *mqSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
 
 // Close the producer asynchronously, does not care closed successfully or not.
 func (k *mqSink) Close(ctx context.Context) error {
+	k.flushWorker.closeMsgChan()
 	go k.mqProducer.Close()
```
lonng, line 342: Do we need to close the `mqProducer` asynchronous?

hi-rustin: Required because sarama gets stuck when kafka is not healthy. So we need to process it asynchronously, otherwise it will get stuck in ticdc.

cdc/sink/mq/mq_flush_worker.go
```diff
@@ -270,3 +252,7 @@ func (w *flushWorker) flushAndNotify(ctx context.Context) error {
 
 	return nil
 }
+
+func (w *flushWorker) closeMsgChan() {
```
lonng, line 256: Why not name the method `close` so we can put more finalize code into it?

hi-rustin: For now we only need to close this one resource, so name it more precisely. If we have a need to extend it, we can rename it in the future.

cdc/sink/mq/mq.go
lonng: It's better to add some comments here to make the code clear.

cdc/sink/mq/mq_flush_worker.go
lonng: LGTM, even though the name is a little bit ugly.

cdc/sink/mq/mq.go
```diff
@@ -342,6 +338,7 @@ func (k *mqSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
 
 // Close the producer asynchronously, does not care closed successfully or not.
 func (k *mqSink) Close(ctx context.Context) error {
+	k.flushWorker.closeMsgChan()
```
CharlesCheung96, line 341: Why doesn't `k.resolvedBuffer` need to be closed?

hi-rustin: see: func (k *mqSink) bgFlushTs(ctx context.Context) error {

cdc/sink/mq/mq_flush_worker.go
```diff
@@ -235,17 +223,11 @@ func (w *flushWorker) run(ctx context.Context) (retErr error) {
...
 	select {
 	case <-ctx.Done():
 		return errors.Trace(ctx.Err())
-	case err, ok := <-w.errCh:
-		if !ok {
-			return cerror.ErrMQWorkerClosed.GenWithStackByArgs()
-		}
-		return err
-	case w.msgChan <- event:
+	case w.msgChan.In() <- event:
```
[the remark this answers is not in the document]
hi-rustin: It's possible, but it doesn't matter, the worker's run exits and no more messages are processed. And sink will be close soon.

pkg/chann/chann.go
```diff
@@ -0,0 +1,255 @@
...
+// list, however, the function panics if there is more than one option is
+// provided.
+func New[T any](opts ...Opt) *Chann[T] {
+	cfg := &config{
+		cap: -1, len: 0,
+		typ: unbounded,
+	}
+
+	if len(opts) > 1 {
+		panic("chann: too many arguments")
```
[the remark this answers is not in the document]
hi-rustin: It is necessary, because it is an optional parameter, but it can be at most one parameter.

cdc/sink/mq/mq.go
hi-rustin: I will add it.

hi-rustin: Thanks!

hi-rustin: Done

cdc/sink/mq/mq.go
```diff
@@ -342,6 +338,9 @@ func (k *mqSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
 
 // Close the producer asynchronously, does not care closed successfully or not.
 func (k *mqSink) Close(ctx context.Context) error {
+	k.flushWorker.close()
```
asddongmen, line 341: will it happens that mqSink.Close() has been called while mq.bgFlushTs() being called in a separate goroutine?

cdc/sink/mq/mq_flush_worker.go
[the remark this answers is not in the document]
CharlesCheung96: However, writing data to a closed chan will lead to panic.
Ref: https://github.com/pingcap/tiflow/pull/5675/files/87b4774faf66b3fcdc126b90f325e33b934a3b47#diff-28a85a85eb9cf5f4d9dd0de5f7e9a0d5ee84e1b2244c914d6ee48bd814705603R182

pkg/chann/main_test.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2021 PingCAP, Inc.
```
CharlesCheung96, line 1:
```suggestion
// Copyright 2022 PingCAP, Inc.
```

cdc/sink/mq/mq.go
hi-rustin: I don't think there will be such a problem because the global sink close is the processor's responsibility, and the processor stops all table pipelines before shutting down the sink.

cdc/sink/mq/mq_flush_worker.go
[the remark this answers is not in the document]
hi-rustin: I don't think there will be such a problem because the global sink close is the processor's responsibility, and the processor stops all table pipelines before shutting down the sink.

pkg/chann/chann.go
```diff
@@ -0,0 +1,255 @@
...
+// 	// or
+//  ch := chann.New[float64](chann.Cap(-1))
+//
+// If the chann.Cap specified a non-negative integer, the returned channel
+// is either unbuffered (0) or buffered (positive).
+//
+// Note that although the input arguments are  specified as variadic parameter
+// list, however, the function panics if there is more than one option is
+// provided.
+func New[T any](opts ...Opt) *Chann[T] {
```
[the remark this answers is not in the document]
hi-rustin: That would have to pass parameters, but we're taking advantage of the fact that there are no parameters. It represents an unbounded chan.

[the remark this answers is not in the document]
hicqu: Got it.

hicqu:
LGTM.

cdc/sink/mq/mq.go
CharlesCheung96: In my opinion, `bgFlushTs` will not be affected by the table pipeline. Normally, it exits only after ctx.done().
Maybe Sink needs to maintain a ctx and cancel.

sdojjy (approved):

asddongmen (approved):

cdc/sink/mq/mq_test.go
```diff
@@ -241,6 +242,11 @@ func TestPulsarSinkEncoderConfig(t *testing.T) {
 	errCh := make(chan error, 1)
 
 	sink, err := NewPulsarSink(ctx, sinkURI, fr, replicaConfig, opts, errCh)
+	// FIXME: mock pulsar client doesn't support close,
+	// so we can't call sink.Close() to close it.
+	// We will leak goroutine if we don't close it.
+	defer sink.flushWorker.close()
```
3AceShowHand, line 248: how about group these 2 defer calls toghter, by `defer func()`

zhaoxinyu (approved):

cdc/sink/mq/mq_test.go
hi-rustin: Done. Thanks!

cdc/sink/sink.go
```diff
@@ -173,12 +173,17 @@ func Validate(ctx context.Context, sinkURI string, cfg *config.ReplicaConfig, op
 	errCh := make(chan error)
 	ctx = contextutil.PutRoleInCtx(ctx, util.RoleClient)
 	// TODO: find a better way to verify a sinkURI is valid
-	s, err := New(ctx, model.DefaultChangeFeedID("sink-verify"),
+	tempCtx, cancel := context.WithCancel(ctx)
```
CharlesCheung96, line 176:
```suggestion
	ctx, cancel := context.WithCancel(contextutil.PutRoleInCtx(ctx, util.RoleClient))
```

cdc/sink/sink.go
```diff
@@ -173,12 +173,17 @@ func Validate(ctx context.Context, sinkURI string, cfg *config.ReplicaConfig, op
...
+	s, err := New(tempCtx, model.DefaultChangeFeedID("sink-verify"),
 		sinkURI, sinkFilter, cfg, opts, errCh)
 	if err != nil {
+		cancel()
 		return err
 	}
-	err = s.Close(ctx)
+	// NOTICE: We have to cancel the context before we close it,
+	// otherwise we will be writing the closed chan after closing the sink.
```
CharlesCheung96, line 184:
```suggestion
	// otherwise we will write data to closed chan after sink closed.
```

cdc/sink/sink.go
```diff
@@ -173,11 +173,16 @@ func Validate(ctx context.Context, sinkURI string, cfg *config.ReplicaConfig, op
 	errCh := make(chan error)
 	ctx = contextutil.PutRoleInCtx(ctx, util.RoleClient)
```
CharlesCheung96, line 174:
```suggestion
```

PR #5685, opened 2022-05-31
api(ticdc): add /api/v2/tso: get tso from pd

### LINKED ISSUE #5586: TiCDC Open API V2: return tso get from pd
provide an api which returns the pd tso
response like:
```go
type Tso struct {
   Timestamp int64 `json:"timestamp"`
   LogicTime int64 `json:"logic-time"`
}
```

crelax:

Issue Number: close #5586

Add API Get tso from PD to OpenAPI v2

#### Tests 

 - Unit test
 - Integration test
 - No code

No
No
### Release note 

```release-note
TiCDC Open API V2: return tso get from pd 
```


cdc/api/v2/api.go
```diff
@@ -14,25 +14,71 @@
...
+	"github.com/pingcap/tiflow/cdc/model"
+	"github.com/pingcap/tiflow/cdc/owner"
+	"github.com/pingcap/tiflow/pkg/upstream"
 )
 
 // OpenAPIV2 provides CDC v2 APIs
 type OpenAPIV2 struct {
 	capture *capture.Capture
+	// use for unit test only
+	testStatusProvider owner.StatusProvider
```
lonng, line 31: I think it's not the best way to mock the behavior of `OpenAPIV2`.

We can use failpoint into the method `statusProvider`, eg:
```go
func (h *OpenAPIV2) statusProvider() owner.StatusProvider {
	failpoint.Inject("mockStatusProvider", func(){
		failpoint.Return(newMockStatusProvider())
	})
	return h.capture.StatusProvider()
}
```

To avoid pollute the definition of type `OpenAPIV2`.

crelax: Thx & Fixed. For now, I erased the dependence on StatusProvider by removal since tso_test didn't use it.

asddongmen (approved):

PR #5524, opened 2022-05-23
server(cdc): avoid printing help messages when cdc server exits

crelax:
### What problem does this PR solve?

Issue Number: close #5519

### What is changed and how it works?
In `func RunE()` of the `cdc server` command, if no error occurs in the validation, return nil directly in spite of errors thrown by the execution. Handle the execution errors using `cobra.CheckErr()`

#### Tests 

 - Integration test 
   - `tests/integration_tests/cdc_server_tips`: The tests cover two scenarios: one `cdc server` command with invalid arguments should print usage tips, while the other is validated and should not.
 - Manual test
   - Start a tidb cluster and add a cluster using `cdc server`
   - Create a changefeed task using `cli changefeed`
   - Press `Ctrl + C` and in the terminal of `cdc server` and check whether there are prompt messages printed
 
##### Will it cause performance regression or break compatibility?
No
##### Do you need to update user documentation, design documentation or monitoring documentation?
No
### Release note 

```release-note
Clear redundant cdc server prompt in a normal exit
```

PR #5686, opened 2022-05-31
pkg/cmd: tiny refactor of run CDC server command

lonng:

It's not a good idea to implement the `run` method on the `options` type. This PR moves `run` into an individual function `runServer` and keep the `options` type clean.

**Note**: I deleted the `cobra.CheckErr` from `command.RunE` and return the error directly because the root command will check the error and print it out to standard error.

#### Tests 

 - No code

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -109,56 +109,6 @@ func (o *options) addFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.allowedCertCN, "cert-allowed-cn", "", "Verify caller's identity (cert Common Name). Use ',' to separate multiple CN")
 }
 
-// run runs the server cmd.
-func (o *options) run(cmd *cobra.Command) error {
-	cancel := util.InitCmd(cmd, &logutil.Config{
-		File:                 o.serverConfig.LogFile,
-		Level:                o.serverConfig.LogLevel,
-		FileMaxSize:          o.serverConfig.Log.File.MaxSize,
-		FileMaxDays:          o.serverConfig.Log.File.MaxDays,
-		FileMaxBackups:       o.serverConfig.Log.File.MaxBackups,
-		ZapInternalErrOutput: o.serverConfig.Log.InternalErrOutput,
-	})
-	defer cancel()
-
-	tz, err := ticdcutil.GetTimezone(o.serverConfig.TZ)
-	if err != nil {
-		return errors.Annotate(err, "can not load timezone, Please specify the time zone through environment variable `TZ` or command line parameters `--tz`")
-	}
-
-	config.StoreGlobalServerConfig(o.serverConfig)
-	ctx := contextutil.PutTimezoneInCtx(cmdcontext.GetDefaultContext(), tz)
-	ctx = contextutil.PutCaptureAddrInCtx(ctx, o.serverConfig.AdvertiseAddr)
-
-	version.LogVersionInfo()
-	if ticdcutil.FailpointBuild {
-		for _, path := range failpoint.List() {
-			status, err := failpoint.Status(path)
-			if err != nil {
-				log.Error("fail to get failpoint status", zap.Error(err))
-			}
-			log.Info("failpoint enabled", zap.String("path", path), zap.String("status", status))
-		}
-	}
-
-	util.LogHTTPProxies()
-	cdc.RecordGoRuntimeSettings()
-	server, err := cdc.NewServer(strings.Split(o.serverPdAddr, ","))
-	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/cmd/server/server.go
```diff
@@ -295,9 +245,18 @@ func NewCmdServer() *cobra.Command {
...
+				File:                 o.serverConfig.LogFile,
+				Level:                o.serverConfig.LogLevel,
+				FileMaxSize:          o.serverConfig.Log.File.MaxSize,
+				FileMaxDays:          o.serverConfig.Log.File.MaxDays,
+				FileMaxBackups:       o.serverConfig.Log.File.MaxBackups,
+				ZapInternalErrOutput: o.serverConfig.Log.InternalErrOutput,
+			})
+			defer cancel()
+
+			return runServer(o)
```
lonng, line 259: @crelax Can we use `return runServer(o)` directly and mark the `SilenceUsage` of the command as false to avoid printing the help message (https://github.com/pingcap/tiflow/pull/5524).

lonng: Ref: https://github.com/pingcap/tiflow/pull/5686/commits/0ab05f46be5cf475abd62764afdbc28fe55fb6cf

crelax: So sorry that I just noticed this pr and replied late. 
Yes, it can avoid printing redundant help messages by marking `SilenceUsage`. Nevertheless, when the user's inputs are lacking or invalid, silence mode also eliminates the usage tips and that's the time we consider the tips necessary. :)

crelax: We may want to throw an error and print usage tips when the command failed for invalid inputs, and remain silent and just report the error when the error occurs within `run` function since irrelevant to users' inputs.

lonng: @crelax Got it, I will revert the `cobra.CheckErr` part.

lonng: Address in commit https://github.com/pingcap/tiflow/pull/5686/commits/cd66f542d4507c71597f8b52c4435bb3a3cbd340

asddongmen (approved):

hi-rustin:
Thanks for working on this.


> It's not a good idea to implement the `run` method on the `options` type. This PR moves `run` into an individual function `runServer` and keep the `options` type clean.

I don't really see anything wrong with this, there doesn't seem to be anything wrong with making the structure with the information responsible for implementing the method. And it's natural. Also I was referring to kubectl https://github.com/kubernetes/kubectl/blob/182c187ab6e547583c5e51219a2f375de2eaf6f6/pkg/cmd/create/create.go#L234

It seems natural that a method that knows the context and has all the information is responsible for executing the business logic. Passing it to a normal function and then running it doesn't seem to have any additional advantages.

I'm okay with you moving it out, but it will result in a inconsistent style with the other commands.

lonng:
@hi-rustin I think kinds of `options` and `config` types were used to store some context information to pass between different call sites.

What if we pass the `options` to some function and the function called the `run` method but we didn't aware of that?

hi-rustin:
> What if we pass the `options` to some function and the function called the `run` method but we didn't aware of that?

I think this is an abuse of the options, which should only be cli and flag related and should not be called in reverse in other business logic. It is the collector and maintainer of cli information, and it is also the constructor of cli commands. It should not be used for anything other than constructing cli commands. Much less should it be passed directly to other business logic for use (which is hard to do since each command is a separate process), if the business logic really needs this data, it should be translated into real ticdc meta information or passed through other data structures. For example, for server, the real configuration is serverConfig instead of using options directly. So I don't think there's any problem with binding cli-related logic to options, and it won't cause the abuse you're talking about.

Our current approach will make the cli writing style more consistent:

1. complete options
2. verify options
3. options run

This is a uniform organizational pattern.

lonng:
> > What if we pass the `options` to some function and the function called the `run` method but we didn't aware of that?
> 
> I think this is an abuse of the options, which should only be cli and flag related and should not be called in reverse in other business logic. It is the collector and maintainer of cli information, and it is also the constructor of cli commands. It should not be used for anything other than constructing cli commands. Much less should it be passed directly to other business logic for use (which is hard to do since each command is a separate process), if the business logic really needs this data, it should be translated into real ticdc meta information or passed through other data structures. For example, for server, the real configuration is serverConfig instead of using options directly. So I don't think there's any problem with binding cli-related logic to options, and it won't cause the abuse you're talking about.
> 
> Our current approach will make the cli writing style more consistent:
> 
> 1. complete options
> 2. verify options
> 3. options run
> 
> This is a uniform organizational pattern.

Make sense

Issue #5689, opened 2022-06-01
Tracking issue for implementing OpenAPI for engine


gozssky:
## Development Task

There is a prototype implementation in https://github.com/pingcap/tiflow/pull/5563.

- [x] HTTP handler  registration in executor https://github.com/pingcap/tiflow/pull/5696
- [x] OpenAPI framework in servermaster https://github.com/pingcap/tiflow/pull/5713
- [x] OpenAPI forwarding from servermaster to executor https://github.com/pingcap/tiflow/pull/5713
- [x] OpenAPI implementation for servermaster https://github.com/pingcap/tiflow/pull/6669

PR #5696, opened 2022-06-01
executor(engine): support openapi registration for job master

gozssky:

Issue Number: ref https://github.com/pingcap/tiflow/issues/5689

This PR adds support for job master to register its openapi to executor.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


amyangfei (approved):

engine/lib/base_jobmaster.go
```diff
@@ -63,6 +64,23 @@ type BaseJobMaster interface {
 	IsBaseJobMaster()
 }
 
+// BaseJobMasterExt extends BaseJobMaster with some extra methods.
+// These methods are used by framework and are not visible to JobMasterImpl.
+type BaseJobMasterExt interface {
```
maxshuang, line 69: May be we can refine the Master/Worker logic later.

engine/lib/base_jobmaster.go
```diff
@@ -63,6 +64,23 @@ type BaseJobMaster interface {
...
+	// OnOpenAPIInitialized is used to initialize the openapi handler.
```
maxshuang, line 70: How about using a name for action, like 'TriggerOpenAPIInitialize'.

engine/pkg/promutil/util.go
```diff
@@ -77,12 +77,12 @@ func HTTPHandlerForMetric() http.Handler {
 //		$JobType_$Namespace_$Subsystem_$Name(actual)
 
 // NewFactory4Master return a Factory for jobmaster
-func NewFactory4Master(info tenant.ProjectInfo, jobType libModel.JobType, jobID libModel.MasterID) Factory {
```
maxshuang, line 80: I think `MasterID` may be more proper here because this is the function for inner usage(master/worker abstract).
But the `jobType` make me struggle.

gozssky: > But the jobType make me struggle.

It's the same to me.

engine/lib/base_jobmaster.go
gozssky: Done.

maxshuang (approved):

PR #5704, opened 2022-06-01
syncer/binlogstream (dm): split this package from syncer

lance6716:

Issue Number: ref #4287 

as title. Split the package so the logic is more clear and easier to test

```
// Package binlogstream is used by syncer to read binlog.
// All information related to upstream binlog stream should be kept in this package,
// such as reset binlog to a location, maintain properties of the binlog event
// and stream, inject or delete binlog events with binlog stream, etc.
```

#### Tests 

 - pass original tests

### Release note 

```release-note
`None`.
```


GMHDBJD (approved):
LGTM

lichunzhu (approved):
LGTM

PR #5715, opened 2022-06-02
sink/mq(ticdc): gracefully close chann

hi-rustin:

Issue Number: ref https://github.com/pingcap/tiflow/issues/5655

gracefully close chann. We must finish consuming the data here, otherwise it will cause the channel to not close properly.

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```


CharlesCheung96 (approved):

cdc/sink/mq/mq.go
```diff
@@ -197,6 +197,15 @@ func (k *mqSink) FlushRowChangedEvents(
...
 // bgFlushTs flush resolvedTs to workers and flush the mqProducer
 func (k *mqSink) bgFlushTs(ctx context.Context) error {
+	defer func() {
+		go func() {
+			// We must finish consuming the data here,
+			// otherwise it will cause the channel to not close properly.
+			for range k.resolvedBuffer.Out() {
+				// Do nothing. We do not care about the data.
+			}
+		}()
```
overvenus, line 207: Shall we wait the for loop done in the defer func?

hi-rustin: No, closing is also done asynchronously, as long as we can return the error and the upper level calls close, we will always finish consuming the data and close the goruntine in chann normally.

overvenus: What about moving the cleanup for loop to `mqSink.Close()`?

hi-rustin: Good idea. Makes sense! Thanks

hi-rustin: Done

cdc/sink/mq/mq.go
```diff
@@ -347,6 +347,11 @@ func (k *mqSink) EmitDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
 // table sinks before closing it. So there is no writing after closing.
 func (k *mqSink) Close(ctx context.Context) error {
 	k.resolvedBuffer.Close()
+	// We must finish consuming the data here,
+	// otherwise it will cause the channel to not close properly.
+	for range k.resolvedBuffer.Out() {
```
asddongmen, line 352: Is there a possible this for loop will block the `processorManager`.

overvenus (approved):
LGTM

cdc/sink/mq/mq.go
hi-rustin: We stop sending data before we stop sink. so I don't think so, and it doesn't make sense for the consumer to get stuck after closing?

hi-rustin:
/hold

I will test it.

Issue #5748, opened 2022-06-06
chann: unbounded channel may result in a spin loop if there is no receiver after the close


hi-rustin:
https://github.com/pingcap/tiflow/blob/12319d58deb82d732a9f4ea2c03bb5c9dfdbda2e/pkg/chann/chann.go#L191

This can lead to leaking goroutine on close and cause cpu spin.

PR #5749, opened 2022-06-06
pkg/chann(ticdc): prevent unbounded channel may not be close

hi-rustin:

Issue Number: close https://github.com/pingcap/tiflow/issues/5748

prevent unbounded channel may not be close. When there are no consumers, drop that data.

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```

### DIFF
--- a/pkg/chann/chann.go
+++ b/pkg/chann/chann.go
@@ -96,6 +96,7 @@ type Chann[T any] struct {
 	in, out chan T
 	close   chan struct{}
 	cfg     *config
+	q       []T
 }
 
 // New returns a Chann that may represent a buffered, an unbuffered or
@@ -127,7 +128,7 @@ func New[T any](opts ...Opt) *Chann[T] {
 	for _, o := range opts {
 		o(cfg)
 	}
-	ch := &Chann[T]{cfg: cfg}
+	ch := &Chann[T]{cfg: cfg, close: make(chan struct{})}
 	switch ch.cfg.typ {
 	case unbuffered:
 		ch.in = make(chan T)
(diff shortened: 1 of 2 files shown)


hi-rustin (commented):
Self check

pkg/chann/chann.go
```diff
@@ -213,11 +158,89 @@ func (ch *Chann[T]) Close() {
...
+	for e := range ch.in {
+		ch.q = append(ch.q, e)
+	}
+	for len(ch.q) > 0 {
+		select {
+		case ch.out <- ch.q[0]:
+		// The default branch exists because we need guarantee
+		// the loop can terminate. If there is a receiver, the
+		// first case will ways be selected. See https://github.com/golang-design/chann/issues/3.
+		default:
```
CharlesCheung96, line 225: Is it possible to cause data loss here?  Or we don't care if the data is consumed properly after chan is closed?

hi-rustin: We don't care about data loss, it's more about deleting that data and gc as soon as possible here.

CharlesCheung96: After chan closed, slow consumers can actually only consume part of their data, which is pretty confusing.
So why not just close `out`, which provides a more explicit guarantee of the semantics.

pkg/chann/chann.go
```diff
@@ -96,6 +96,7 @@ type Chann[T any] struct {
 	in, out chan T
 	close   chan struct{}
 	cfg     *config
+	q       []T
```
CharlesCheung96, line 99:
```suggestion
    q       []T
	in, out chan T
	close   chan struct{}
	cfg     *config
```
`q` is similar to `in` and `out`.

pkg/chann/chann.go
```diff
@@ -213,11 +158,89 @@ func (ch *Chann[T]) Close() {
...
+		close(ch.close)
 	default:
 		ch.close <- struct{}{}
 	}
 }
 
+// unboundedProcessing is a processing loop that implements unbounded
+// channel semantics.
+func (ch *Chann[T]) unboundedProcessing() {
+	var nilT T
```
CharlesCheung96, line 170: Maybe `emptyT` is more appropriate?

pkg/chann/chann.go
hi-rustin: Would this be less friendly to consumers who can handle it quickly?

CharlesCheung96: No matter how fast consumers consume, there is always the risk of losing data, which can be quite insidious and dangerous. In this sense, it is always safer to lose a continuous piece of data than to lose a random piece of data.

pkg/chann/chann.go
hi-rustin: Done

pkg/chann/chann.go
hi-rustin: Renamed it to zeroT.

pkg/chann/chann.go
hi-rustin: Removed the default. 

CharlesCheung96 (approved):

overvenus:
Could you feedback your changes to upstream?

hi-rustin:
> Could you feedback your changes to upstream?

In fact, I'm merging upstream changes.

pkg/chann/chann.go
```diff
@@ -213,11 +164,87 @@ func (ch *Chann[T]) Close() {
...
+			}
+		}
+		if cap(ch.q) < 1<<5 {
+			ch.q = make([]T, 0, 1<<10)
+		}
+	}
+}
+
+// unboundedTerminate terminates the unbounde channel's processing loop
+// and make sure all unprocessed elements either be consumed if there is
```
hi-rustin, line 216:
```suggestion
// and make sure all unprocessed elements be consumed if there is
```

asddongmen (approved):

PR #5631, opened 2022-05-29
fix: https://github.com/pingcap/tiflow/issues/5625

fatelei:
### What problem does this PR solve?

Issue Number: close #5625 

#### Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

### Release note 

```release-note
Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

If you don't think this PR needs a release note then fill it with `None`.
```

PR #5751, opened 2022-06-06
tests(ticdc): change cdc cron mq integrations to pull triggered

### LINKED ISSUE #5750: cdc cron mq tests not guarding code properly
### Which jobs are flaking?

N/A

N/A

https://github.com/pingcap/tiflow/actions/workflows/ticdc_integration_cron.yaml

The action contains two test and runs them cronly. But from the result I think there is almost nobody careing about them. The canal job fails consistently for 4 months. And with the GA of avro feature, we should take the corresponding test seriously.

zhangyangyu:

Issue Number: close #5750 

### Release note 

```release-note
None
```


zhangyangyu:
This PR depends on https://github.com/pingcap/tiflow/pull/5631.

zhaoxinyu (approved):

hi-rustin:
I'm worried that it will still cause a lot of queuing for pr tasks in tidb. @zhangyangyu Are they using a self-built runner? But it seems that the concurrency of each organization does not improve because of the self-built runner? (I'm not sure)

zhangyangyu:
> I'm worried that it will still cause a lot of queuing for pr tasks in tidb. @zhangyangyu Are they using a self-built runner? But it seems that the concurrency of each organization does not improve because of the self-built runner? (I'm not sure)

Not sure either. Check the configurations of `tiflow` and `tidb`, no self-hosted runner in use yet. But I think it's another problem. We'd better first make these tests actually in use to ensure a bottom line for canal and avro. Currently almost no verify CI is enabled for canal and avro even they are in GA status. If the limited resource is a problem, let's push the EE team to improve it :-(. 

And from the check results of this PR, avro-integration takes about 14m to succeed, canal-integration takes less in my local test if it could succeed. Maybe it could be worse than the unstable dm tests. ;-)

hi-rustin:
> Currently almost no verify CI is enabled for canal and avro even they are in GA status.

`canal-json` has its own tests on jenkins. `canal` does not have GA.

> If the limited resource is a problem, let's push the EE team to improve it :-(.

I've tried, but our ci doesn't support dind very well, and migrating all those tests to Jenkins is difficult.

hi-rustin:
> Check the configurations of `tiflow` and `tidb`, no self-hosted runner in use yet.

tiflow's `Arm Build (ARM64) (pull_request) ` uses a self-built runner.

hi-rustin:
The main reason why other tests cannot use the self-built runner is that dind is not supported, or dind is basically unavailable.

PR #5755, opened 2022-06-06
dm: fix the problem that ssl-*-bytes are printed in log

lichunzhu:

Issue Number: close https://github.com/pingcap/tiflow/issues/5754

Use regex to replace ssl-bytes info with '*'.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


dm/pkg/utils/util.go
```diff
@@ -101,14 +101,18 @@ var (
 	builtInSkipDDLPatterns *regexp.Regexp
 
 	passwordPatterns = `(password: (\\")?)(.*?)((\\")?\\n)`
-	passwordRegexp   *regexp.Regexp
+	sslPatterns      = `(s*sl-(.*?)-bytes:)([\s\\\nn]+)(-\W[0-9]+[\s\\\nn]+)+?([a-zA-Z])`
```
lance6716, line 104: can we use 

`ssl-ca-bytes:((\\{4}n\s{4}-\s\d+)+)` , ssl-key-bytes... 

three variables? that's easier to understand what it matches.

lance6716: three regexp variables are

ssl-ca-bytes:((\\{4}n\s{4}-\s\d+)+)
ssl-key-bytes:((\\{4}n\s{4}-\s\d+)+)
ssl-cert-bytes:((\\{4}n\s{4}-\s\d+)+)

And we can find how `SubTaskConfig` become the log string in #5754, then imitate the logic in unit test. So in future if `SubTaskConfig` changed the test will fail

lichunzhu: emmm, this regex can't match in my test.
Besides, I use three variables at first, but they are almost the same except for `ca`,`key` part. Introduce three variables will also make code more complicated..

lance6716: maybe we can use `ssl-(ca|key|cert)-bytes:((\\n\s{4}-\s\d+)+\\n\s*)`

I'm just not aware of why your expression is very complicated 😂

lance6716:
```diff
diff --git a/dm/pkg/utils/util.go b/dm/pkg/utils/util.go
index cd743bd8b..22a08287f 100644
--- a/dm/pkg/utils/util.go
+++ b/dm/pkg/utils/util.go
@@ -101,7 +101,7 @@ var (
 	builtInSkipDDLPatterns *regexp.Regexp
 
 	passwordPatterns = `(password: (\\")?)(.*?)((\\")?\\n)`
-	sslPatterns      = `(s*sl-(ca|key|cert)-bytes:)([\s\\\nn]+)(-\s\d+[\s\\\nn]+)+?([a-zA-Z])`
+	sslPatterns      = `(ssl-(ca|key|cert)-bytes:)((\\n\s{4}-\s\d+)+)`
 
 	passwordRegexp *regexp.Regexp
 	sslRegexp      *regexp.Regexp
@@ -179,7 +179,7 @@ func IsBuildInSkipDDL(sql string) bool {
 // HideSensitive replace password with ******.
 func HideSensitive(input string) string {
 	output := passwordRegexp.ReplaceAllString(input, "$1******$4")
-	output = sslRegexp.ReplaceAllString(output, "$1 \"******\"\\\\\\\\n    $5")
+	output = sslRegexp.ReplaceAllString(output, "$1 \"******\"")
 	return output
 }
 
diff --git a/dm/pkg/utils/util_test.go b/dm/pkg/utils/util_test.go
index 7abd4598f..e16dc834d 100644
--- a/dm/pkg/utils/util_test.go
+++ b/dm/pkg/utils/util_test.go
@@ -17,11 +17,13 @@ import (
 	"context"
 	"errors"
 	"os"
+	"testing"
 	"time"
 
 	"github.com/go-mysql-org/go-mysql/mysql"
 	. "github.com/pingcap/check"
 	"github.com/pingcap/tidb/errno"
+	"github.com/stretchr/testify/require"
 )
 
 func (t *testUtilsSuite) TestDecodeBinlogPosition(c *C) {
@@ -75,7 +77,7 @@ func (t *testUtilsSuite) TestWaitSomething(c *C) {
 	c.Assert(count, Equals, 5)
 }
 
-func (t *testUtilsSuite) TestHideSensitive(c *C) {
+func TestName(t *testing.T) {
 	strs := []struct {
 		old string
 		new string
@@ -94,14 +96,14 @@ func (t *testUtilsSuite) TestHideSensitive(c *C) {
 			`\n\ntarget-database:\n  host: \"127.0.0.1\"\n  port: 4000\n  user: \"test\"\n  password: \"******\"\n\nmysql-instances:\n  - source-id: \"mysql-replica-01\"\n`,
 		}, { // operate source
 			`user: root\n  password: /Q7B9DizNLLTTfiZHv9WoEAKamfpIUs=\n  port: 3306 security:\n ssl-ca-bytes:\n    - 45\n    ssl-key-bytes:\n    - 45\n    ssl-cert-bytes:\n    - 45\npurge:`,
-			`user: root\n  password: ******\n  port: 3306 security:\n ssl-ca-bytes: "******"\\\\n    ssl-key-bytes: "******"\\\\n    ssl-cert-bytes: "******"\\\\n    purge:`,
+			`user: root\n  password: ******\n  port: 3306 security:\n ssl-ca-bytes: "******"\n    ssl-key-bytes: "******"\n    ssl-cert-bytes: "******"\npurge:`,
 		}, { // start task with ssl
 			`\n\ntarget-database:\n  host: \"127.0.0.1\"\n  port: 4000\n  user: \"test\"\n  password: \"\"\n security:\n ssl-ca-bytes:\n    - 45\n    ssl-key-bytes:\n    - 45\n    ssl-cert-bytes:\n    - 45\nmysql-instances:\n  - source-id: \"mysql-replica-01\"\n`,
-			`\n\ntarget-database:\n  host: \"127.0.0.1\"\n  port: 4000\n  user: \"test\"\n  password: \"******\"\n security:\n ssl-ca-bytes: "******"\\\\n    ssl-key-bytes: "******"\\\\n    ssl-cert-bytes: "******"\\\\n    mysql-instances:\n  - source-id: \"mysql-replica-01\"\n`,
+			`\n\ntarget-database:\n  host: \"127.0.0.1\"\n  port: 4000\n  user: \"test\"\n  password: \"******\"\n security:\n ssl-ca-bytes: "******"\n    ssl-key-bytes: "******"\n    ssl-cert-bytes: "******"\nmysql-instances:\n  - source-id: \"mysql-replica-01\"\n`,
 		},
 	}
 	for _, str := range strs {
-		c.Assert(HideSensitive(str.old), Equals, str.new)
+		require.Equal(t, str.new, HideSensitive(str.old))
 	}
 }

```

the output is `ssl-ca-bytes: \"******\"\\n` but some tests require `ssl-ca-bytes: \"******\"\\\\\\\\n`. I think that's not right. And the indent for next line should not be changed.

lichunzhu: agree. already addressed

dm/pkg/utils/util_test.go
```diff
@@ -75,7 +77,7 @@ func (t *testUtilsSuite) TestWaitSomething(c *C) {
 	c.Assert(count, Equals, 5)
 }
 
-func (t *testUtilsSuite) TestHidePassword(c *C) {
+func TestName(t *testing.T) {
```
lance6716, line 80:
```suggestion
func TestHideSensitive(t *testing.T) {
```

lance6716 (approved):

dm/dm/pb/hide_password.go
```diff
@@ -17,19 +17,19 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
-// HidePwdFunc should be overwrite by utils.HidePassword, this variable is for avoiding cycle import.
-var HidePwdFunc = func(s string) string {
+// HideSensitiveFunc should be overwrite by utils.HidePassword, this variable is for avoiding cycle import.
```
WizardXiao, line 20:
```suggestion
// HideSensitiveFunc should be overwrite by utils.HideSensitive, this variable is for avoiding cycle import.
```

WizardXiao (approved):

Issue #5521, opened 2022-05-23
lib(engine): Refactor BaseWorker


liuzix:
The `BaseWorker` needs to be refactored to make the code more structured and easy to maintain.

PR #5768, opened 2022-06-07
framework/worker(engine): fix worker exit sequence

### LINKED ISSUE #5764: Exit should make sure CloseImpl happens before InitImpl of next worker

For some workers run one after another, developer may want to call Exit to close previous worker and create next worker. And the developer expects CloseImpl of previous worker happens before InitImpl of next worker, for example, he/she write some exclusive resource releasing logic. However the engine can't ensure it.

as title

_No response_

_No response_

liuzix:

Issue Number: ref #5521 close #5764

- Guarantees that `CloseImpl` be called before initiating the worker exit sequence.
- Added some options for creating `ExitController`.

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
None.
```


engine/lib/worker/exit_controller.go
```diff
@@ -71,6 +127,14 @@ func (c *ExitController) PollExit() error {
 		return nil
 	}
 
+	c.callPrepareOnce.Do(func() {
```
lance6716, line 130: not familiar with framework, will the first poll block following call to poll? otherwise the first PollExit will enter callPrepareOnce but it may stuck at prepareExitFunc, and seconds PollExit will not wait prepareExitFunc so Exit finishes before CloseImpl

liuzix: Plus, `sync.Once` prevents the scenario you describe.

maxshuang, line 130: Why we need another sync.Once here? w.closeImplOnce can't achieve the `once` precisely?

maxshuang, line 130: Another problem concerns me a lot.
We design the `Tick` interface to implement the period behavior of business, but DM will run its logic in a independent goroutine, so we call `CloseImpl` before stop calling `Tick` seems ok.

But, to some extend,  other business will observe that `CloseImpl` will be called before stop calling `Tick`, which
means `a man is dead but its heart is beating`.

liuzix: The one in `ExitController` is for redundancy. I can change it to using an atomic, because there is no risk of racing to calling the `prepareExitFunc`.

liuzix: It does not seem that `Tick` can be called after `CloseImpl`, based on the implementation of `BaseWorker`.

maxshuang: My mistake.

maxshuang (approved):

GMHDBJD:
Can you revert this part and check the dm e2e test pass? https://github.com/pingcap/tiflow/pull/5503/commits/4bec2b1d4f24753503f92d9ca724100d6a2646d2

GMHDBJD (approved):
LGTM

PR #5778, opened 2022-06-07
engine(dm): add operate task

GMHDBJD:

Issue Number: ref #5504 

- implement operate-task
- implement OnWorkerStatusUpdated
- refactor unitholder
- move auto-resume to worker level 

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/executor/dm/unitholder.go
```diff
@@ -16,66 +16,45 @@ package dm
...
 // unitHolder hold a unit of DM
 type unitHolder interface {
 	Init(ctx context.Context) error
-	Tick(ctx context.Context) error
-	Stage() metadata.TaskStage
-	Status(ctx context.Context) interface{}
 	Close(ctx context.Context) error
+	Pause(ctx context.Context) error
+	Resume(ctx context.Context) error
+	Stage() (metadata.TaskStage, *pb.ProcessResult)
```
lance6716, line 37: maybe another function to return ProcessResult. Stage and Status is separated, so Result can also be separated.

engine/executor/dm/unitholder.go
```diff
@@ -89,105 +68,142 @@ func newUnitHolderImpl(workerType lib.WorkerType, cfg *dmconfig.SubTaskConfig) *
 
 // Init implement UnitHolder.Init
 func (u *unitHolderImpl) Init(ctx context.Context) error {
+	u.processMu.Lock()
```
lance6716, line 71: maybe caller can make sure Init is not concurrent with Pause/Resume, so no need to lock here. Otherwise event if we use a lock, Pause/Resume may happen before Init and go wrong.

Also for Close

engine/executor/dm/unitholder.go
```diff
@@ -89,105 +68,142 @@ func newUnitHolderImpl(workerType lib.WorkerType, cfg *dmconfig.SubTaskConfig) *
...
+	u.fieldMu.Lock()
+	u.runCtx, u.runCancel = runCtx, runCancel
+	u.fieldMu.Unlock()
+
+	resultCh := make(chan pb.ProcessResult, 1)
+	u.processWg.Add(1)
+	go func() {
+		defer u.processWg.Done()
+		u.unit.Process(runCtx, resultCh)
+		u.fetchAndHandleResult(resultCh)
```
lance6716, line 88: Do you have any experience to reuse fetchResultAndUpdateStage?

engine/executor/dm/unitholder.go
```diff
@@ -89,105 +68,142 @@ func newUnitHolderImpl(workerType lib.WorkerType, cfg *dmconfig.SubTaskConfig) *
...
-	u.mu.Lock()
-	defer u.mu.Unlock()
-	u.lastResult = r
+	// cancel process
+	u.fieldMu.Lock()
+	u.runCancel()
+	u.fieldMu.Unlock()
+	u.processWg.Wait()
+	// TODO: refactor unit.Syncer
+	// unit needs to manage its own life cycle
```
lance6716, line 108: agree

lance6716 (commented):
will review soon

engine/executor/dm/unitholder.go
```diff
@@ -89,105 +68,142 @@ func newUnitHolderImpl(workerType lib.WorkerType, cfg *dmconfig.SubTaskConfig) *
...
 
 // Close implement UnitHolder.Close
 func (u *unitHolderImpl) Close(ctx context.Context) error {
-	u.cancel()
+	u.processMu.Lock()
+	defer u.processMu.Unlock()
+
+	u.fieldMu.Lock()
+	// cancel process
+	u.runCancel()
```
lance6716, line 146: will framework call CloseImpl before or concurrent to InitImpl?

engine/executor/dm/worker.go
```diff
@@ -103,6 +113,7 @@ func newDMWorker(ctx *dcontext.Context, masterID libModel.MasterID, workerType l
 
 // InitImpl implements lib.WorkerImpl.InitImpl
 func (w *dmWorker) InitImpl(ctx context.Context) error {
```
lance6716, line 115: If InitImpl meets error, InitImpl itself should be responsible to clean the started components?

engine/executor/dm/worker.go
```diff
@@ -187,15 +199,17 @@ func (w *dmWorker) persistStorage(ctx context.Context) error {
 
 // tryUpdateStatus updates status when task stage changed.
```
lance6716, line 200: line 179: setupstorage -> setupStorage

engine/executor/dm/worker.go
```diff
@@ -187,15 +199,17 @@ func (w *dmWorker) persistStorage(ctx context.Context) error {
...
 func (w *dmWorker) tryUpdateStatus(ctx context.Context) error {
-	stage := w.unitHolder.Stage()
-	if stage == w.getStage() {
+	stage, _ := w.unitHolder.Stage()
+	currentStage := w.getStage()
```
lance6716, line 203: I forget why we have to keep 2 stage. Please left some comments

engine/executor/dm/unitholder.go
GMHDBJD: I think they should be together, and for finished and error, it should return the result

engine/executor/dm/unitholder.go
GMHDBJD: For init it is right, for close the framework needs to be guaranteed, here is to enhance some robustness

lance6716: Maybe we can add unit test to make sure Pause after Close is OK, for example, no panic will happen. (I haven't review unit tests)

engine/executor/dm/unitholder.go
GMHDBJD: Not very realistic, they are owned by different objects（subtask/unitholder）. Maybe we can merge fetchResult into process later

engine/executor/dm/worker.go
GMHDBJD: rename variable

engine/executor/dm/worker.go
GMHDBJD: cc @liuzix @amyangfei

engine/executor/dm/unitholder.go
GMHDBJD: No

engine/jobmaster/dm/dm_jobmaster.go
```diff
@@ -188,8 +188,12 @@ func (jm *JobMaster) onWorkerFinished(taskStatus runtime.TaskStatus, worker lib.
...
 // OnWorkerStatusUpdated implements JobMasterImpl.OnWorkerStatusUpdated
 func (jm *JobMaster) OnWorkerStatusUpdated(worker lib.WorkerHandle, newStatus *libModel.WorkerStatus) error {
-	// No need to do anything here, because we update it in OnWorkerOnline
-	return nil
+	// we alreay update finished status in OnWorkerOffline
+	if newStatus.Code == libModel.WorkerStatusFinished || len(newStatus.ExtBytes) == 0 {
+		return nil
+	}
+	log.L().Debug("on worker status updated", zap.String("extra bytes", string(newStatus.ExtBytes)), zap.String("id", jm.workerID), zap.String("worker_id", worker.ID()))
+	return jm.OnWorkerOnline(worker)
```
lance6716, line 196: why we treat it (such as task paused by error) as WorkerOnline? maybe we can extract updateStatus, then let OnWorkerStatusUpdated and OnWorkerOnline to call it.

engine/jobmaster/dm/metadata/job.go
```diff
@@ -110,6 +112,11 @@ func (jobStore *JobStore) UpdateStages(ctx context.Context, taskIDs []string, st
 	}
 
 	job := state.(*Job)
+	if len(taskIDs) == 0 {
```
lance6716, line 115: in which path this is empty 🤔

engine/jobmaster/dm/task_manager.go
```diff
@@ -200,22 +187,22 @@ func (tm *TaskManager) GetTaskStatus(taskID string) (runtime.TaskStatus, bool) {
...
-}
-
-func (tm *TaskManager) operateTaskMessage(ctx context.Context, taskID string, stage metadata.TaskStage) error {
-	if stage != metadata.StageRunning && stage != metadata.StagePaused {
-		return errors.Errorf("invalid expected stage %d for task %s", stage, taskID)
+func genOp(runtimeStage, expectedStage metadata.TaskStage) dmpkg.OperateType {
+	switch {
+	case runtimeStage == metadata.StageRunning && expectedStage == metadata.StagePaused:
+		return dmpkg.Pause
+	case runtimeStage == metadata.StagePaused && expectedStage == metadata.StageRunning:
```
lance6716, line 194: this is duplicated with auto resume in worker?

lance6716 (commented):
will review test later

engine/jobmaster/dm/dm_jobmaster.go
GMHDBJD: for jobmaster, it only need to know whether the worker is online or offline. TaskManager operate task when worker online, WorkerManager scheduler worker when worker offline. For a task status updated, the worker still online and we can operate for it, so I think we should rely on OnWorkerOnline.

engine/jobmaster/dm/metadata/job.go
GMHDBJD: If not specify tasks. e.g. `operate-job job-id pause`

engine/jobmaster/dm/task_manager.go
GMHDBJD: No, auto resume will only resume for metadata.StageError

engine/jobmaster/dm/dm_jobmaster.go
lance6716: Directly let OnWorkerStatusUpdated call OnWorkerOnline is strange. The two functions are callbacks of framework, and have very clear meaning. I still prefer we extract our private function and let them two use it.

engine/jobmaster/dm/task_manager.go
lance6716: auto resume may need to check if the expected stage is paused, then no need to resume an error task

engine/jobmaster/dm/dm_jobmaster.go
lance6716: OK, we can left a comment to explain it

engine/jobmaster/dm/task_manager.go
GMHDBJD: jobmaster: resume task if the task is paused but expected to running.
worker: resume task if the task is error.

> auto resume may need to check if the expected stage is paused

if expected stage is paused, jobmaster will send a pause request, worker will paused it, and set stage to pausing/paused, and then will not auto resume

lance6716: a small improvement is if worker is in error stage, and user now set expected stage to paused, master will not send pause  request immediately. Worker may resume 5 miinutes later and change the stage to running, then master send pause request. We can add this case.

engine/jobmaster/dm/dm_jobmaster.go
lance6716: I am now thinking a debug scenario. When I see the logic is inside OnWorkerOnline, I may skip checking the caller because the meaning of OnWorkerOnline is very clear, then I jump to the logic of dmWorker online. But the bug may lie in dmWorker update stage.

The point is we want to share some code in the two callback, but the callback has very clear definition that indicates its usage, so in my mind even though the code is same the two functions are still separate functions, they should not be related to each other.

GMHDBJD: done

engine/executor/dm/unitholder.go
```diff
@@ -89,105 +68,146 @@ func newUnitHolderImpl(workerType lib.WorkerType, cfg *dmconfig.SubTaskConfig) *
...
+	u.processMu.Lock()
+	defer u.processMu.Unlock()
 
-func (u *unitHolderImpl) checkAndAutoResume() {
-	result := u.getResult()
-	if result == nil || len(result.Errors) == 0 {
-		return
+	stage, _ := u.Stage()
+	if stage != metadata.StagePaused && stage != metadata.StageError {
+		return errors.Errorf("failed to resume unit with stage %d", stage)
```
lance6716, line 119: maybe give a string name to stage later

lance6716 (approved):

amyangfei (approved):

PR #5794, opened 2022-06-08
pb(engine): Remove pb job type and use strict sql mode for framework metastore 

### LINKED ISSUE #5801: engine: remove JobType  from proto to avoid regenerate pb

No

Remove JobType  from proto to avoid regenerate pb

_No response_

_No response_

maxshuang:

Issue Number: close #5801 

1. Remove `JobType` in `master.proto` to avoid regenerate pb when adding a new job type
2. Change `WorkerType` type to int16,  `WorkerStatusCode` to int8
3. Add `strict sql mode` to framework metastore connection to avoid some unexpected overflow and truncation of meta data

#### Tests 

 - Unit test

### Release note 

```release-note
`None`.
```


amyangfei:
Please fix e2e test

engine/model/job.go
```diff
@@ -0,0 +1,91 @@
...
+	JobTypeJobManager
+	JobTypeCVSDemo
+	JobTypeDM
+	JobTypeCDC
+	JobTypeFakeJob
+)
+
+var jobTypeName = map[int32]string{
+	2: "CVSDemo",
+	3: "DM",
```
amyangfei, line 41:
```suggestion
var jobTypeName = map[JobType]string{
	JobTypeCVSDemo: "CVSDemo",
	JobTypeDM: "DM",
	...
```

engine/model/job.go
```diff
@@ -0,0 +1,91 @@
...
+	4: "CDC",
+	5: "FakeJob",
+}
+
+var jobTypeValue = map[string]int32{
```
amyangfei, line 46: ditto

engine/model/job.go
```diff
@@ -0,0 +1,91 @@
...
+	tp, ok := JobTypeByName(name)
+	if !ok {
+		log.L().Panic("unexpected fail when get job type by name", zap.String("job_name", name))
+	}
+
+	return tp
+}
+
+// JobTypeByName get JobType by readable job name
+func JobTypeByName(name string) (JobType, bool) {
```
gozssky, line 64: How about `ParseJobType` and `MustParseJobType`?

engine/model/job.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"go.uber.org/zap"
+)
+
+type (
+	// JobID is the unique identifier of a job
+	JobID = string
+	// JobType is the type of a type, like `CVSDEMO`, `DM`, `CDC`.
+	JobType = int32
```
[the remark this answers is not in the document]
maxshuang: I adjust `WorkerType` to `int16` to avoid potential overflow in DB.
Seems that as a memory variable, `int32` is an acceptable type.

I am a little struggle.

engine/model/job.go
```diff
@@ -0,0 +1,91 @@
...
+	name, ok := JobNameByType(tp)
+	if !ok {
+		log.L().Panic("unexpected fail when get job name by type", zap.Int32("job_type", tp))
+	}
+
+	return name
+}
+
+// JobNameByType get job name by job type
+func JobNameByType(tp JobType) (string, bool) {
```
gozssky, line 84: How about implementing `fmt.Stringer()`. I think it would be idiomatic.

engine/model/job.go
maxshuang: Solve in latest commit.

engine/model/job.go
maxshuang: Solve in latest commit.

engine/model/job.go
```diff
@@ -0,0 +1,91 @@
...
+	JobTypeInvalid = JobType(iota)
+	// job type for job manager, only for inner use
+	JobTypeJobManager
+	JobTypeCVSDemo
+	JobTypeDM
+	JobTypeCDC
+	JobTypeFakeJob
+)
+
+var jobTypeName = map[int32]string{
```
gozssky, line 39: Usually, using a switch statement is faster than the map. See https://stackoverflow.com/questions/46789259/map-vs-switch-performance-in-go.

engine/chaos/cases/case_fake_job.go
```diff
@@ -23,8 +23,8 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/pingcap/tiflow/dm/pkg/log"
-	pb "github.com/pingcap/tiflow/engine/enginepb"
 	"github.com/pingcap/tiflow/engine/lib/fake"
+	dfModel "github.com/pingcap/tiflow/engine/model"
```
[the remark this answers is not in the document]
maxshuang: Done

engine/model/job.go
maxshuang: How about `GetJobTypeByName` ?

engine/model/job.go
maxshuang: This function has a meaning to check if the job type is registered.
`fmt.Stringer()` seems not so precise。

gozssky: This purpose would make me think the string is actually the job type identifier. To check whether a type is valid, it's unnecessary to get its string name. In most cases, I think the integer value is the only identifier and the string is just for human-readable.

BTW, here we should use `JobTypeName` instead.

gozssky: Besides, `fmt.Stringer()` is more convenient for log and debug.

maxshuang: I will think about it. Seems it's ambiguous for the name.

engine/model/job.go
```diff
@@ -0,0 +1,77 @@
...
+	JobTypeJobManager
+	JobTypeCVSDemo
+	JobTypeDM
+	JobTypeCDC
+	JobTypeFakeJob
+)
+
+// Define job type name
+const (
+	JobNameInvalid    = "Invalid"
```
gozssky, line 36: `JobName` -> `JobTypeName`?

gozssky (commented):
rest LGTM

engine/model/job.go
maxshuang: Done

engine/model/job.go
```diff
@@ -0,0 +1,77 @@
...
+	JobTypeNameInvalid    = "Invalid"
+	JobTypeNameJobManager = "JobManager"
+	JobTypeNameCVSDemo    = "CVSDemo"
+	JobTypeNameDM         = "DM"
+	JobTypeNameCDC        = "CDC"
+	JobTypeNameFakeJob    = "FakeJob"
+)
+
+// jobNameToType maintains the valid job name and job type
+var jobNameToType = map[string]JobType{
```
gozssky, line 45:
```suggestion
var jobTypeNameToType = map[string]JobType{
```

engine/model/job.go
```diff
@@ -0,0 +1,77 @@
...
+	JobTypeNameCVSDemo: JobTypeCVSDemo,
+	JobTypeNameDM:      JobTypeDM,
+	JobTypeNameCDC:     JobTypeCDC,
+	JobTypeNameFakeJob: JobTypeFakeJob,
+}
+
+// GetJobTypeByName get JobType by readable job name
```
gozssky, line 52:
```suggestion
// GetJobTypeByName get JobType by readable job type name.
```

gozssky (approved):

amyangfei (approved):

engine/framework/model/master.go
```diff
@@ -25,32 +25,23 @@ import (
...
-	"id",
-	"type",
-	"status",
-	"node_id",
-	"address",
-	"epoch",
-	"config",
-}
+// Job master statuses
+const (
```
liuzix, line 32: Can we hard code the statuses here instead of using iota? Adding or removing might cause confusion when using iota.

engine/framework/model/status.go
```diff
@@ -58,12 +59,12 @@ var WorkerUpdateColumns = []string{
...
-	JobID        string           `json:"job-id" gorm:"column:job_id;type:varchar(64) not null;uniqueIndex:uidx_wid,priority:1;index:idx_wst,priority:1"`
-	ID           string           `json:"id" gorm:"column:id;type:varchar(64) not null;uniqueIndex:uidx_wid,priority:2"`
-	Type         int              `json:"type" gorm:"column:type;type:tinyint not null"`
+	ProjectID    tenant.ProjectID `json:"project-id" gorm:"column:project_id;type:varchar(64) not null"`
+	JobID        MasterID         `json:"job-id" gorm:"column:job_id;type:varchar(64) not null;uniqueIndex:uidx_wid,priority:1;index:idx_wst,priority:1"`
+	ID           WorkerID         `json:"id" gorm:"column:id;type:varchar(64) not null;uniqueIndex:uidx_wid,priority:2"`
+	Type         WorkerType       `json:"type" gorm:"column:type;type:smallint not null"`
 	Code         WorkerStatusCode `json:"code" gorm:"column:status;type:tinyint not null;index:idx_wst,priority:2"`
-	ErrorMessage string           `json:"error-message" gorm:"column:errmsg;type:varchar(128)"`
+	ErrorMessage string           `json:"error-message" gorm:"column:errmsg;type:varchar(1024)"`
```
liuzix, line 67: Can we use `blob` or `text` here?

engine/framework/model/master.go
maxshuang: I think it's a convention for the use of `iota`.
What about `Adding comment for the MasterStatusCode in DB`?

maxshuang: Hard code may be better.

Issue #5663, opened 2022-05-31
CDC cli use open api to communicate with cdc server


sdojjy:
cli command list

- [x] cdc cli capture list 
- [x]  dc cli processor list  
- [x] cdc cli processor query
- [x] cdc cli changefeed pause    
- [x]  cdc cli changefeed resume    
- [x] cdc cli changefeed remove
- [x]  cdc cli changefeed list  
- [x] cdc cli changefeed query  
- [x]  cdc cli changefeed update   
- [x] cdc cli changefeed create  
- [ ] cdc cli changefeed cyclic   create-marktables 
- [x] cdc cli changefeed statistics        
- [x] cdc cli tso query        
- [x] cdc cli unsafe delete-service-gc-safepoint
- [x] cdc cli unsafe show-metadata
- [x] cdc cli unsafe resolve-lock

PR #5802, opened 2022-06-08
http&migration(ticdc): Return http error if server is not ready

sdojjy:

Issue Number: ref #5663 

1. remove /tidb/cdc gc service safe point after the migration is done
2. add upstream info
3. if server is not ready, return error

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```

### DIFF
--- a/pkg/migrate/migrate.go
+++ b/pkg/migrate/migrate.go
@@ -23,9 +23,12 @@ import (
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	"github.com/pingcap/tiflow/cdc/model"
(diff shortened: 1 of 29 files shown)


pkg/migrate/migrate.go
```diff
@@ -124,46 +216,50 @@ func (m migrator) migrate(ctx context.Context) error {
 			newKey := newPrefix + oldKey[len(oldPrefix):]
 			beforeKV[newKey] = v.Value
```
asddongmen, line 217: beforeKV should be delete if there is no need to check data before and after migrate.

PR #5806, opened 2022-06-09
registry(engine): Simplify worker registration with generics

liuzix:

Issue Number: ref #5805

- Made `NewSimpleWorkerFactory` generic.

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
None.
```


engine/lib/registry/factory.go
```diff
@@ -40,35 +41,45 @@ type WorkerFactory interface {
...
 	}
 }
 
 // NewWorkerImpl implements WorkerFactory.NewWorkerImpl
-func (f *SimpleWorkerFactory) NewWorkerImpl(
+func (f *SimpleWorkerFactory[T, C]) NewWorkerImpl(
 	ctx *dcontext.Context,
 	workerID libModel.WorkerID,
 	masterID libModel.MasterID,
 	config WorkerConfig,
```
lance6716, line 74: can we require config is of type C?

lance6716 (commented):
rest lgtm

engine/lib/registry/factory.go
liuzix: It seems you cannot implement the `WorkerFactory` interface this way. There has to be a place where you do a `type erasure`.

lance6716 (approved):

amyangfei (approved):

PR #5820, opened 2022-06-09
fb/latency(cdc): agent to be table state awared, to handle different p2p messages.

3AceShowHand:

Issue Number: ref #4757

let the `agent` handle different table state, this makes the message report table's state more accurate.

#### Tests 

 - Unit test

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/scheduler/internal/tp/table.go
+++ b/cdc/scheduler/internal/tp/table.go
@@ -1,0 +1,338 @@
+// Copyright 2022 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tp
+
+import (
+	"context"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
+	"github.com/pingcap/tiflow/cdc/model"
+	"github.com/pingcap/tiflow/cdc/processor/pipeline"
+	"github.com/pingcap/tiflow/cdc/scheduler/internal"
+	"github.com/pingcap/tiflow/cdc/scheduler/internal/tp/schedulepb"
+	"go.uber.org/zap"
+)
+
+// table is a state machine that manage the table's state,
+// also tracking its progress by utilize the `TableExecutor`
+type table struct {
+	id model.TableID
+
+	status   schedulepb.TableStatus
+	executor internal.TableExecutor
+
+	task *dispatchTableTask
+}
+
+func newTable(tableID model.TableID, executor internal.TableExecutor) *table {
+	return &table{
+		id:       tableID,
+		executor: executor,
+		status:   schedulepb.TableStatus{TableID: tableID},
+		task:     nil,
+	}
+}
+
+// refresh the table' status, return true if the table state changed,
+// should be called before any table operations.
+func (t *table) refresh() bool {
+	oldState := t.status.State
+
+	meta := t.executor.GetTableMeta(t.id)
+	newState := tableStatus2PB(meta.State)
+
+	t.status.State = newState
+	t.status.Checkpoint.CheckpointTs = meta.CheckpointTs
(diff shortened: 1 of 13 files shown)


cdc/scheduler/internal/tp/table.go
```diff
@@ -0,0 +1,360 @@
...
+	oldState := t.status.State
+
+	meta := t.executor.GetTableMeta(t.id)
+	newState := tableStatus2PB(meta.State)
+
+	t.status.State = newState
+	t.status.Checkpoint.CheckpointTs = meta.CheckpointTs
+	t.status.Checkpoint.ResolvedTs = meta.ResolvedTs
+
+	return oldState != newState
```
overvenus, line 60: Could you add a log if a table changes its state?

pkg/config/server_config.go
```diff
@@ -128,7 +128,7 @@ var defaultServerConfig = &ServerConfig{
 		},
 		Messages: defaultMessageConfig.Clone(),
 
-		EnableTwoPhaseScheduler: false,
+		EnableTwoPhaseScheduler: true,
```
overvenus, line 131: It may cause config tests fail

cdc/scheduler/internal/tp/table.go
```diff
@@ -0,0 +1,375 @@
...
+				zap.Int64("tableID", t.id), zap.Any("status", status))
+		}
+	}
+	return nil
+}
+
+func (t *table) handleAddTableTask(ctx context.Context) (result *schedulepb.Message, err error) {
+	status := t.status
+	stateChanged := true
+	for stateChanged {
```
overvenus, line 145: Could you unify the for loop at line 102?

cdc/scheduler/internal/tp/table.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+}
+
+func newRemoveTableResponseMessage(status schedulepb.TableStatus) *schedulepb.Message {
+	message := &schedulepb.Message{
+		MsgType: schedulepb.MsgDispatchTableResponse,
+		DispatchTableResponse: &schedulepb.DispatchTableResponse{
+			Response: &schedulepb.DispatchTableResponse_RemoveTable{
+				RemoveTable: &schedulepb.RemoveTableResponse{
+					Status:     &status,
```
overvenus, line 89: Will the `Status` be a pointer to the `table.Status`? Will there be a data race?

cdc/scheduler/internal/tp/table.go
```diff
@@ -0,0 +1,355 @@
...
+	"github.com/pingcap/tiflow/cdc/scheduler/internal/tp/schedulepb"
+	"go.uber.org/zap"
+)
+
+// table is a state machine that manage the table's state,
+// also tracking its progress by utilize the `TableExecutor`
+type table struct {
+	id model.TableID
+
+	status   schedulepb.TableStatus
```
overvenus, line 33: Why not just `schedulepb.TableState`?

3AceShowHand: since this is also used to build `heartbeatResponse`, which needs checkpoint.

`status` is more useful in such use case.

overvenus: What about query `TableMeta()` when building `HeartbeatResponse`? In this way, we don't need `refresh` to reconcile checkpoint and resolved ts.

3AceShowHand: but we still need reconcile state. I don't think there is any difference.

cdc/scheduler/internal/tp/table.go
3AceShowHand: no, this is no data race.

cdc/scheduler/internal/tp/table.go
```diff
@@ -0,0 +1,364 @@
...
+	return &table{
+		id:       tableID,
+		state:    schedulepb.TableStateAbsent, // use `absent` as the default state.
+		executor: executor,
+		task:     nil,
+	}
+}
+
+// getTableState get the table' state, return true if the table state changed
+func (t *table) getTableState() (schedulepb.TableState, bool) {
```
overvenus, line 49:
```suggestion
func (t *table) getAndUpdateTableState() (schedulepb.TableState, bool) {
```

overvenus (approved):

3AceShowHand:
failure by `dm_unit_test` https://ci2.pingcap.net/blue/organizations/jenkins/cdc_ghpr_test/detail/cdc_ghpr_test/16801/pipeline

cdc related test passed.

PR #5831, opened 2022-06-12
api(ticdc): extract forwardToOwnerMiddleware

### LINKED ISSUE #5830: Simplify forwardToOwner in TiCDC API/v1

`forwardToOwner()` logic occurs in several handlers in TiCDC API v1, and it can be extracted as a Gin Middleware.

Extracting forwardToOwner() as a Gin Middleware.

_No response_

_No response_

eastfisher:

Issue Number: close #5830

Extract a function called forwardToOwnerMiddleware() which is used as a Gin Middleware.

#### Tests 

 - No code

No.

No.

### Release note 

```release-note
`None`
```


sdojjy:
why make this change? some ready-only requests can be handled by the non-owner nodes.

eastfisher:
> why make this change? some ready-only requests can be handled by the non-owner nodes.

Hi! I found that these handlers all check current server is owner or not. If it is not owner, then forward request to owner, or handle it locally.

Other handlers which does not need to check owner role are not changed in this PR.

asddongmen:
Thank you for your contribution!

asddongmen (approved):

sdojjy (approved):

PR #5712, opened 2022-06-01
sink(ticdc): remove force consume when redo is disabled

CharlesCheung96:
### What problem does this PR solve?

Issue Number: close #5453, close #1683

### What is changed and how it works?
In order to solve the OOM problem of single table, we refactor flowControl and Sink modules on the basis of splitting transaction, so that they can obtain data in batches and send it to the downstream, which reduces the memory consumption of individual table to constant level: 
1. In flowcontrol module, allocate and release based on resolvedTs and batchID.
2. In sink module, replace model.Ts with model.ResolvedTs to support batch-based memory release.
3. In sinkNode, release memory based on checkpointTs and batchID.

#### Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 > down stream: TiDB
total row of txn: 30 million
txn size: ~5.4 GB
downstream: tidb cluster
memory of cdc node: 32 GB

The results show that memory consumption of single table is constant. However, a lower quota value may affect throughput, so it is recommended to increase the quota of single tables to 30-40 MB.

image
image
image
image
image
image

> compare with #5454 
quota: 100MB
down stream: TiDB
total row of txn: 30 million
txn size: ~5.4 GB
downstream: tidb cluster
memory of cdc node: 32 GB

[up-TiCDC_2022-06-06T10_33_19.155Z.json.zip](https://github.com/pingcap/tiflow/files/8843760/up-TiCDC_2022-06-06T10_33_19.155Z.json.zip)

[heap_profile.zip](https://github.com/pingcap/tiflow/files/8843828/heap_profile.zip)

##### Will it cause performance regression or break compatibility?
No

##### Do you need to update user documentation, design documentation or monitoring documentation?
No

### Release note 

```release-note
`Optimize the memory consumption required for TiCDC to process large transactions`.
```

Issue #5231, opened 2022-04-21
Tracking issue for splitting big transaction


CharlesCheung96:
### Before asking a question, make sure you have

- [X] Searched existing Stack Overflow questions
- [X] Googled your question
- [X] Searched open and closed [GitHub issues](https://github.com/pingcap/tiflow/issues?q=is%3Aissue)
- [X] Read the [documentation](https://docs.pingcap.com/tidb/stable)

### What is your question?

Tracking issue for splitting big transaction

PR #5848, opened 2022-06-13
tests(ticdc): add big txn integration test

CharlesCheung96:

Issue Number: ref #5231 

1.  <s>According to the analysis in #5712, increase per table quota to 35MB to improve sink throughput.</s>
2. Add an integration test for big txn (INSERT scenario, about 100MB per txn) to verify that cdc can properly sync transactions whose size exceeds the single-table quota.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


tests/integration_tests/big_txn/run.sh
```diff
@@ -0,0 +1,56 @@
...
+	run_sql "CREATE DATABASE big_txn;"
+	go-ycsb load mysql -P $CUR/conf/workload -p mysql.host=${UP_TIDB_HOST} -p mysql.port=${UP_TIDB_PORT} -p mysql.user=root -p mysql.db=big_txn
+	run_cdc_server --workdir $WORK_DIR --binary $CDC_BINARY
+
+	if [ "$SINK_TYPE" == "kafka" ]; then
+		run_kafka_consumer $WORK_DIR "kafka://127.0.0.1:9092/$TOPIC_NAME?protocol=open-protocol&partition-num=4&version=${KAFKA_VERSION}&max-message-bytes=10485760"
+	fi
+
+	check_table_exists "big_txn.USERTABLE" ${DOWN_TIDB_HOST} ${DOWN_TIDB_PORT}
+	# run_sql_file $CUR/data/test.sql ${UP_TIDB_HOST} ${UP_TIDB_PORT}
```
CharlesCheung96, line 38:
```suggestion
```

CharlesCheung96, line 38:
```suggestion
```

pkg/config/server_config.go
```diff
@@ -37,6 +37,9 @@ const (
 
 	// DebugConfigurationItem is the name of debug configurations
 	DebugConfigurationItem = "debug"
+
+	// DefaultTableMemoryQuota is the default memory quota for each table.
+	DefaultTableMemoryQuota = 35 * 1024 * 1024 // 35 MB
```
asddongmen, line 42: Is 35 MB a reasonable value? Would it cause OOM more easily when the table amount is huge?

asddongmen: @overvenus Do you have any advice on that?

3AceShowHand:
LGTM

Please also consider the impact of set the default memory quota to 35MB.

It looks that this would cause OOM more easily, which may be due to a low memory mount or the huge amount of tables.

pkg/config/server_config.go
overvenus: Please do not change the default setting, it may break many table (60k) scenario.

What about setting 35MB in the big_txn test?

CharlesCheung96: It's a matter of tradeoffs: after this [pr](https://github.com/pingcap/tiflow/pull/5712), the memory consumption of single table is constant. However, the test [results](https://github.com/pingcap/tiflow/pull/5712#issue-1256591663) show a lower quota value may affect throughput, so it is recommended to increase the quota of single tables to 30-40 MB.

CharlesCheung96: Maybe we could change this parameter as needed in the future.

asddongmen (approved):

overvenus (approved):

PR #5879, opened 2022-06-15
sink/mq,sink/mysql(ticdc): remove useless opts for MQ and MySQL sink

hi-rustin:

Issue Number: ref https://github.com/pingcap/tiflow/issues/5849

remove useless opts for MQ sink.

This PR simply removes the dead code.
Before.
You pass opts it won't do anything, but it won't report an error.

After the change.
You pass opts again cli will report an error.

Upgrade
The original parameters won't work after the upgrade, and it won't panic.

#### Tests 

 - Unit test

No
No, I didn't find any docs about this arg.
### Release note 

```release-note
None
```

### DIFF
--- a/cdc/api/validator/validator.go
+++ b/cdc/api/validator/validator.go
@@ -130,7 +130,6 @@ func VerifyCreateChangefeedConfig(
 	// init ChangefeedInfo
 	info := &model.ChangeFeedInfo{
 		SinkURI:           changefeedConfig.SinkURI,
-		Opts:              make(map[string]string),
 		CreateTime:        time.Now(),
 		StartTs:           changefeedConfig.StartTS,
 		TargetTs:          changefeedConfig.TargetTS,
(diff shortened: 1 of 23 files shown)


hi-rustin:
I tested:
v6.1.0 -> this pr
v5.0.6 -> this pr
v4.0.14 -> this pr

```shell
./cdc cli changefeed create --sink-uri="blackhole://" --changefeed-id="test1" --opts "registry=http://schema-registry:8081"
```

hi-rustin:
This PR simply removes the dead code.
Before.
You pass opts it won't do anything, but it won't report an error.

After the change.
You pass opts again cli will report an error.

Upgrade
The original parameters won't work after the upgrade, and it won't panic.

leoppro (approved):

nongfushanquan (approved):
LGTM

PR #5882, opened 2022-06-15
executor/runtime,framework(engine): Move event loop execution down to framework

liuzix:

Issue Number: close #xxx

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note

```

### DIFF
--- a/engine/executor/worker/internal/runnables.go
+++ b/engine/executor/worker/internal/runnables.go
@@ -17,9 +17,10 @@ import (
 	"context"
 	"time"
 
-	"github.com/pingcap/tiflow/dm/pkg/log"
 	"go.uber.org/atomic"
 	"go.uber.org/zap"
+
+	"github.com/pingcap/tiflow/dm/pkg/log"
 
 	"github.com/pingcap/tiflow/engine/model"
 )
@@ -82,8 +83,8 @@ func (c *RunnableContainer) Info() RuntimeInfo {
 	return c.info
 }
 
-// OnInitialized is the callback when the runnable instance is initialized
-func (c *RunnableContainer) OnInitialized() {
+// OnLaunched is the callback when the runnable instance is launched.
+func (c *RunnableContainer) OnLaunched() {
 	oldStatus := c.status.Swap(TaskRunning)
 	if oldStatus != TaskSubmitted {
 		log.L().Panic("unexpected status", zap.Int32("status", oldStatus))
--- a/engine/executor/worker/task_committer.go
+++ b/engine/executor/worker/task_committer.go
@@ -17,10 +17,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/pingcap/tiflow/dm/pkg/log"
 	"go.uber.org/atomic"
 	"go.uber.org/zap"
 
+	"github.com/pingcap/tiflow/dm/pkg/log"
 	"github.com/pingcap/tiflow/engine/executor/worker/internal"
 	"github.com/pingcap/tiflow/engine/pkg/clock"
 )
--- a/engine/executor/worker/task_runner.go
+++ b/engine/executor/worker/task_runner.go
@@ -21,7 +21,6 @@ import (
 	"github.com/pingcap/errors"
 	"go.uber.org/atomic"
 	"go.uber.org/zap"
-	"golang.org/x/sync/semaphore"
 
 	"github.com/pingcap/tiflow/dm/pkg/log"
 	"github.com/pingcap/tiflow/engine/executor/worker/internal"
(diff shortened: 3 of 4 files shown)


engine/framework/eventloop/runner.go
```diff
@@ -0,0 +1,115 @@
...
+const (
+	defaultEventLoopTickInterval = 50 * time.Millisecond
+)
+
+// Runner runs a task.
+type Runner[T task] struct {
+	task T
+
+	alreadyRun atomic.Bool
+	label      string
```
amyangfei, line 40: What is the meaning of `label` here. Does it only alias to `worker id` or will be a vector in the future?

engine/framework/eventloop/runner.go
```diff
@@ -0,0 +1,115 @@
...
+		label: label,
+		clk:   clock.New(),
+	}
+}
+
+// Run is called by the runtime. Cancelling the
+// ctx cancels the task immediately and forcefully.
+func (r *Runner[R]) Run(ctx context.Context) error {
+	if r.alreadyRun.Swap(true) {
+		panic(fmt.Sprintf("duplicate calls to Run: %s", r.label))
```
amyangfei, line 58: Better to use log.L().Panic

engine/framework/worker.go
```diff
@@ -94,6 +94,7 @@ type BaseWorker interface {
 	// When `err` is not nil, the status code is assigned WorkerStatusError.
 	// Otherwise worker should set its status code to a meaningful value.
 	Exit(ctx context.Context, status frameModel.WorkerStatus, err error) error
```
maxshuang, line 96: Why we need two Exit interface method?

engine/framework/eventloop/runner.go
maxshuang: Same confusion to me. Please add some comments for the lable.

engine/framework/eventloop/runner.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/pingcap/tiflow/engine/pkg/clock"
+	derrors "github.com/pingcap/tiflow/engine/pkg/errors"
+)
+
+const (
+	defaultEventLoopTickInterval = 50 * time.Millisecond
+)
+
+// Runner runs a task.
+type Runner[T task] struct {
```
maxshuang, line 36: Why we need a generic here?
Seems  eventloop of the framwork only cares about the common abilities of task, which means the base object can alread meet our need.

engine/framework/worker.go
maxshuang: As a user, should I really care about this stuff?
I think the framwork has some promise about the correct time sequence of callback, so the user only need to implement the callback.

engine/framework/eventloop/runner.go
maxshuang: I can't figure out what scenarios we will require a generic but not an interface `task` here.

engine/framework/eventloop/runner.go
liuzix: log.L().Panic will not put the zapped fields into the panic info, making it harder to debug

engine/framework/eventloop/runner.go
maxshuang: Because the DefaultBaseMaster and DefaultBaseWorker is a part of the framwork, which means we can control the implement of it.

engine/framework/worker.go
liuzix: > As a user, should I really care about this stuff?

No, you shouldn't. This is not a call back, and the user does not need to implement it. The framework provides an implementation.

engine/framework/eventloop/runner.go
liuzix: It does not seem necessary, but it can be good for readability, so that the reader of the code knows what the Runner runs.

engine/framework/worker.go
maxshuang: Oh, it's my fault for not making me clear.
I mean: 
Should user use the `NotifyExit`? Or this interface method will be hidden after we refine the BaseMaster/MasterImpl interface?

engine/framework/eventloop/runner.go
maxshuang: Ok. I hold my opinion.
@amyangfei @gozssky  PTAL~

engine/framework/worker.go
liuzix: The method is hidden. `Runner` is not exposed to the user.

engine/framework/eventloop/runner.go
amyangfei: Maybe we can remove generic to make it simple, if generic is needed in the feature, we can add it back.

amyangfei (approved):

maxshuang (approved):

PR #5892, opened 2022-06-15
sink/mysql(ticdc): add comments

hi-rustin:

Issue Number: close https://github.com/pingcap/tiflow/issues/5891

Add a comment to mysql explaining how to detect conflicts

#### Tests 

 - No code

No 
No 
### Release note 

```release-note
None
```

### DIFF
--- a/cdc/sink/mysql/causality.go
+++ b/cdc/sink/mysql/causality.go
@@ -22,6 +22,13 @@ import (
 	"github.com/pingcap/tiflow/cdc/model"
 )
 
+const (
+	// noConflicting means that the transaction has no conflicts with anything else on the worker.
+	noConflicting = -1
+	// multipleConflicting means that the transaction has multiple conflicts with other workers.
+	multipleConflicting = -2
+)
+
 // causality provides a simple mechanism to improve the concurrency of SQLs execution under the premise of ensuring correctness.
 // causality groups sqls that maybe contain causal relationships, and syncer executes them linearly.
 // if some conflicts exist in more than one groups, then syncer waits all SQLs that are grouped be executed and reset causality.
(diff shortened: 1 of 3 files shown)


cdc/sink/mysql/causality.go
```diff
@@ -53,21 +60,34 @@ func (c *causality) reset() {
 // detectConflict detects whether there is a conflict
 func (c *causality) detectConflict(keys [][]byte) (bool, int) {
```
nongfushanquan, line 61: please add some comments about how to use the return value ( bool , int ).

asddongmen (approved):

cdc/sink/mysql/causality.go
```diff
@@ -53,21 +60,34 @@ func (c *causality) reset() {
...
+			if conflictingWorkerIndex != workerIndex {
+				return true, multipleConflicting
 			}
 		}
 	}
 
-	return firstIdx != -1, firstIdx
+	// 1) no conflict
+	// 2) conflict with the same worker
+	return conflictingWorkerIndex != noConflicting, conflictingWorkerIndex
```
CharlesCheung96, line 90: Maybe better to separate conflictingWorkerIndex to conflictFlag and workerIndex?

lance6716:
maybe link https://pingcap.com/zh/blog/tidb-binlog-source-code-reading-8

cdc/sink/mysql/causality.go
hi-rustin: Done

cdc/sink/mysql/causality.go
hi-rustin: I think it's really just coupled together, and the fact that we distinguish it internally doesn't change the semantics of that function. So I don't think it's too much of a problem to just leave it that way. Unless we're aiming to refactor it. Then we should really redesign the API.

CharlesCheung96 (approved):

PR #5223, opened 2022-04-20
*(dm): add some warning and adjust some config

lance6716:
Signed-off-by: lance6716 <lance6716@gmail.com>

### What problem does this PR solve?

Issue Number: ref #5220

### What is changed and how it works?

No. 1, 2, 5, 6 subtask of #5220 

### Check List 

Tests 

 - Integration test

Code changes

 - Has exported variable/fields change

Side effects

Related changes

### Release note 

```release-note
Some commands and configuration will be deprecated in future:
- `start-relay` / `stop-relay` with worker name
- `is-sharding` in task configuration
```

PR #5901, opened 2022-06-16
DM(dmctl): Enhance dmctl task commands

WizardXiao:

Issue Number: ref #5900 

change task commands format to `task xxx` and implement commands by openapi interfaces.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


dm/dm/ctl/master/task_check.go
```diff
@@ -0,0 +1,33 @@
...
+
+// NewTaskCheckCmd creates a task check command.
+func NewTaskCheckCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "check <config-file> [--error count] [--warn count] [--start-time]",
+		Short: "Check the configuration file of the task",
+		//TODO check task by openapi
+		RunE: checkTaskFunc,
+	}
+	cmd.Flags().Int64P("error", "e", common.DefaultErrorCnt, "max count of errors to display")
```
buchuitoudegou, line 29:
```suggestion
	cmd.Flags().Int64P("error", "e", common.DefaultErrorCnt, "maximum number of errors to display")
```

dm/dm/ctl/master/task_create.go
```diff
@@ -0,0 +1,138 @@
...
+	// we check if `shard-mode` is explicitly set, to distinguish between "" from default value
+	shardModeSet := false
+	for i := range lines {
+		if bytes.HasPrefix(lines[i], []byte("shard-mode")) {
+			shardModeSet = true
+			break
+		}
+	}
+
+	if isShardingSet && !task.IsSharding && task.ShardMode != "" {
```
buchuitoudegou, line 117: Did you mean:
```
is-sharding: false
shard-mode = "xxx"
```
If `is-sharding` is deprecated, what about setting the `is-sharding` to `true`? Is it legal?

dm/dm/ctl/master/task_create.go
```diff
@@ -0,0 +1,138 @@
...
+		common.PrintLinesf("The behaviour of `is-sharding` and `shard-mode` is conflicting. `is-sharding` is deprecated, please use `shard-mode` only.")
+		return nil, errors.New("please check output to see error")
+	}
+	if shardModeSet && task.ShardMode == "" && task.IsSharding {
```
buchuitoudegou, line 121: Why is `shard-mode` set but `task.ShardMode` remains empty?

dm/dm/ctl/master/task_list.go
```diff
@@ -0,0 +1,109 @@
...
+	listReq := &openapi.DMAPIGetTaskListParams{WithStatus: &more, Stage: stage, SourceNameList: sourceList}
+	params := []interface{}{listReq}
+	listResp := &openapi.GetTaskListResponse{}
+	sendErr := common.SendOpenapiRequest(ctx, "DMAPIGetTaskList", params, http.StatusOK, listResp)
+	if sendErr != nil {
+		return errors.Trace(sendErr)
+	}
+	common.PrettyPrintOpenapiResp(true, listResp)
+
+	// if resp.Result && taskName == "" && len(sources) == 0 && !more {
```
buchuitoudegou, line 100: Should this be removed?

buchuitoudegou (commented):
Should we add some integration tests to ensure this pr works as expected?

WizardXiao:
> Should we add some integration tests to ensure this pr works as expected?

yes, i'm adding the integration test.

dm/dm/ctl/master/task_list.go
WizardXiao: yes, i removed.

dm/dm/ctl/master/task_create.go
WizardXiao: this is the same as `start-task` at [here](https://github.com/pingcap/tiflow/blob/master/dm/dm/ctl/master/start_task.go#L90) to make `task create/start` compatible with `start-task` behavior. 
bwt, taking this out of `start-task` is because we may just maintain commands which call openapi interface and we may extend the commands behavior but grpc interface does not support in the future, but at present, it is better to be compatible.

dm/dm/ctl/master/task_create.go
WizardXiao: the same as above.

WizardXiao:
There has some change about openapi interface:
1. make openapi open as default [here](https://github.com/pingcap/tiflow/pull/5901/files#diff-4b54a4428156c470ed89777702318bd2cd8f8c4fdccb19d32627b82a35bc3d53R63)
2. make task status's sourelist filter effect [here](https://github.com/pingcap/tiflow/pull/5901/files#diff-7d9499db50ffd3895e6f45ab0aee935dac04929a8e4a200d584c37331746d095R517)
3. change task list soure filter and status filter to be true if there has one of them matches [here](https://github.com/pingcap/tiflow/pull/5901/files#diff-7123b0f55cb365015d32584f0a70280f555b8abb799739f530d5540436fbeb8cR626), and it is to support `stop task -s source1`
4. add `Paused` status to task list, because task will be `Paused` after `stop task` now, [here](https://github.com/pingcap/tiflow/pull/5901/files#diff-66c38807b124dc58be9646df20d35503b1812893d3cf81d95c65d9c9b89b9db5R1755)
/cc @GMHDBJD @Ehco1996 @buchuitoudegou

dm/dm/ctl/master/task_create.go
buchuitoudegou: Should we pull out the same logic with `start-task`? Or it would be a bit difficult to make them aligned.

dm/tests/dmctl_advance/check_list/task.sh
```diff
@@ -0,0 +1,518 @@
...
+	task_stop_op
+	task_delete_op
+	task_update_op
+	task_status_op
+	task_list_op
+	other_op
+	https_availability
+}
+
+function task_create_op() {
```
buchuitoudegou, line 22: What about adding failed cases with redundant parameters (i.e. too many arguments error)

dm/openapi/spec/dm.yaml
```diff
@@ -1752,6 +1752,7 @@ components:
       enum:
         - Stopped
         - Running
+        - Paused
```
Ehco1996, line 1755: why we need Paused stage?   [This](https://github.com/pingcap/tiflow/blob/master/dm/docs/RFCS/20220124_enhance_task_manageability.md) was originally designed to remove the overlap between paused and stopped states

WizardXiao: ok, i will remove `Paused`, but i think i will change the status to be `Stopped` after `task stop` in other pr.

dm/dm/ctl/master/task_create.go
WizardXiao: I thought about it, too，but i'm afraid we may change the logic about check the input for extending commands and we may have to think about grpc interface, because previous commands still call this function.

dm/tests/dmctl_advance/check_list/task.sh
```diff
@@ -0,0 +1,518 @@
...
+		"task delete test -f -y" \
+		"Delete task test success" 1
+	run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"task delete test1 -f -y" \
+		"Delete task test1 success" 1
+}
+
+# check,pause,resume still use grpc interface, just Simply test availability
+function other_op() {
+	# commond tips
```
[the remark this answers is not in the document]
WizardXiao: ok, it should be command.

dm/dm/ctl/master/task_create.go
WizardXiao: i find out this [pr](https://github.com/pingcap/tiflow/pull/5223), maybe it is the aim which will return waring or error.

dm/dm/ctl/master/task_operate.go
```diff
@@ -0,0 +1,179 @@
...
+	if resp.Total == 0 {
+		common.PrettyPrintOpenapiResp(false, fmt.Sprintf("there is no tasks of source: %v", sources))
+		return nil
+	}
+	result := batchTaskOperate(cmd, batchSize, sources, resp.Data, operateFunc)
+	common.PrettyPrintInterface(result)
+	return nil
+}
+
+func batchTaskOperate(cmd *cobra.Command, batchSize int, sources []string, tasks []openapi.Task,
```
Ehco1996, line 134: it seem that don't need batch request here? sotp task api support suourceNameList param

dm/dm/ctl/master/task_create.go
WizardXiao: this logic has been pulled out in [convertToOpenapiTaskFromFile](https://github.com/pingcap/tiflow/pull/5901/files#diff-277a02d1cc0098a357a31a6eefddaa67508f1997d193cc928c21485a30a9767fR73) and it will be called by `task create`, `task start` or `task update` which will really use file content, not only task name,  `task check` will call `convertToOpenapiTaskFromFile` too, but it wasn't called by previous commands like `start-task` or `check-task`. i can pull out some logic and let it called by both previous commands and new commands, but if we need extending new commands, we may need think about previous too. considering that the previous instructions will be hidden, i am not sure if this is necessary.

dm/dm/ctl/master/task.go
```diff
@@ -0,0 +1,39 @@
...
+func NewTaskCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "task <command>",
+		Short: "manage or show task operations",
+	}
+	cmd.AddCommand(
+		NewTaskCreateCmd(),
+		NewTaskStartCmd(),
+		NewTaskStopCmd(),
+		NewTaskPauseCmd(),
```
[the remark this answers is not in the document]
WizardXiao: should we use `task pause` and `task resume` instead `pause-task` and `resume-task` , or we just hide `pause-task` and `resume-task`. @sunzhaoyang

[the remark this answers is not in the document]
Ehco1996: in new task cmd, we  only have start/stop task no pause/resume any more, see https://github.com/pingcap/tiflow/blob/master/dm/docs/RFCS/20220124_enhance_task_manageability.md#dmctl-commands-for-task

dm/dm/ctl/master/task_create.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, errors.New("please check output to see error")
+	}
+
+	taskStr = string(fileContent)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	// convert to openapi task
+	params = []interface{}{openapi.ConverterTaskRequest{TaskConfigFile: &taskStr}}
+	convertRes := &openapi.ConverterTaskResponse{}
+	sendError = common.SendOpenapiRequest(ctx, "DMAPIConvertTask", params, http.StatusOK, convertRes)
```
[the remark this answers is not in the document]
WizardXiao: ok, it seems that `convertTaskConfig` need get some info from `scheduler`, may be it  can be pulled out. let me have a try.

dm/dm/ctl/master/task_operate.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	taskName := common.GetTaskNameFromArgOrFile(cmd.Flags().Arg(0))
+	sources, err := common.GetSourceArgs(cmd)
+	if err != nil {
+		return err
+	}
+
+	// check task is exist
+	notExist, err := isTaskNotExist(taskName)
```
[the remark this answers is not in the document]
WizardXiao: i'm afraid not, if we stop task, there is no check and will return success, It may be considered that there are no subtask need to handle. [here](https://github.com/pingcap/tiflow/blob/0294e283f28bdccce80d3ed2c71b3e735e0923fb/dm/dm/master/scheduler/scheduler.go#L1681). I found it out when I was writing the test too.

dm/dm/ctl/master/task_operate.go
WizardXiao: `batchTaskOperate` is called by `task stop -s xxx` which will stop all task about the source, and `task stop xx -s xx,xx` will just call api support suourceNameList param.

dm/dm/ctl/master/task_update.go
```diff
@@ -0,0 +1,61 @@
...
+	if len(cmd.Flags().Args()) != 1 {
+		cmd.SetOut(os.Stdout)
+		common.PrintCmdUsage(cmd)
+		return errors.New("please check output to see error")
+	}
+	task, err := convertToOpenapiTaskFromFile(cmd.Flags().Arg(0))
+	if err != nil {
+		return errors.Trace(err)
+	}
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
WizardXiao: ok

dm/dm/master/openapi_view.go
```diff
@@ -514,7 +514,29 @@ func (s *Server) DMAPIGetTaskStatus(c *gin.Context, taskName string, params open
 		_ = c.Error(err)
 		return
 	}
-	resp := openapi.GetTaskStatusResponse{Total: len(*task.StatusList), Data: *task.StatusList}
+	resp := openapi.GetTaskStatusResponse{}
+
+	// filter by sources
+	if params.SourceNameList != nil {
```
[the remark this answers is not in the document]
WizardXiao: ok, but need `sourcelist` as a param to `s.getTask`.

dm/dm/ctl/master/task_create.go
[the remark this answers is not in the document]
Ehco1996: > convertTaskConfig need get some info from scheduler

sorry, i forget this now i think it's ok use api to do this convert

dm/dm/ctl/master/task_operate.go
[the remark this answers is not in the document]
Ehco1996: if so, i think better to change [here](https://github.com/pingcap/tiflow/blob/0294e283f28bdccce80d3ed2c71b3e735e0923fb/dm/dm/master/scheduler/scheduler.go#L1681)  because the coment said

> no subtask need to update, this should not happen

But this actually happened 🤣, my point is if server found user want to operate an object whis is not exist, server should retun an error instead of asking the client to do this verification

dm/dm/ctl/master/task_operate.go
Ehco1996: Screen Shot 2022-06-24 at 09 25 16


i think all task cmds must have a `task-name` arg, if user want stop all subtasks about a source, he can use `api/v1/source/source-name/disable` instead

PR #5945, opened 2022-06-20
schema(cdc): refactor (*TableInfo)findHandleIndex by reusing public method exposed by tidb

### LINKED ISSUE #5902: (*TableInfo).findHandleIndex() has similar logic with getPrimaryKey

https://github.com/pingcap/tiflow/blob/b9d096912f3ba392a314c19ef8fee03910f6bd86/cdc/model/schema_storage.go#L144

https://github.com/pingcap/tidb/blob/4ae78cdead54097679407314ce33b895e3c01d16/ddl/ddl_api.go#L1638-L1644

very similar

we can make getPrimaryKey in TiDB a public function and reuse it.

_No response_

_No response_

crelax:

Issue Number: close #5902 

resue the method  in tidb

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```


cdc/model/schema_storage.go
```diff
@@ -146,27 +146,10 @@ func (ti *TableInfo) findHandleIndex() {
...
-				(len(ti.Indices[handleIndexOffset].Columns) == len(ti.Indices[i].Columns) &&
-					ti.Indices[handleIndexOffset].ID > ti.Indices[i].ID) {
-				handleIndexOffset = i
-			}
-		}
-	}
-	if handleIndexOffset >= 0 {
-		ti.HandleIndexID = ti.Indices[handleIndexOffset].ID
+
+	pk := ti.GetPrimaryKey()
```
amyangfei, line 150: `ti.GetPrimaryKey` does not completely equal to original logic, given the scenario of no primary key
- If there is no UNIQUE INDEX on NOT NULL columns, `ti.GetPrimaryKey` returns nil, but original logic could choose one unique index
- If there exist multiple NOT NULL unique indexes, `ti.GetPrimaryKey` always return the first one, but original logic choose one with fewer columns (it has no correctness problem, but should check out whether exists compatibility issue)

Issue #5949, opened 2022-06-21
[engine] gRPC usage overhaul


liuzix:
There exists a couple of systemic problems in our current usage of gRPC, including but not limited to the following:
- Lack of a unified retry framework
- Lack of an exception handling mechanism that can propagate structured error across gRPC call boundaries
- Lack of a unified logging mechanism

There are also problems related to individual components regarding gRPC:
- Executor client wrappers that are hard to mock and maintain
- Lack of a good mechanism to manage the membership changes of executors
- Coupling of service clients running on the same node, such as `ExecutorService` and `BrokerService`.

PR #5950, opened 2022-06-21
pkg/rpcerror(engine): implement utils for creating rpc-friendly errors.

liuzix:
Issue Number: ref #5949 

- Implemented `engine/pkg/rpcerror`.

#### Tests 

 - Unit test

No.

No.

### Release note 

```release-note
None.
```


engine/pkg/rpcerror/attributes.go
```diff
@@ -0,0 +1,173 @@
...
+func (c DeadlineExceeded) grpcStatusCode() codes.Code {
+	return codes.DeadlineExceeded
+}
+
+// AlreadyExists can be passed as a parameter to Error to
+// mark the grpc error code of this error as AlreadyExists.
+type AlreadyExists struct{}
+
+func (c AlreadyExists) grpcStatusCode() codes.Code {
+	return codes.NotFound
```
maxshuang, line 71: codes.AlreadyExists?

engine/pkg/rpcerror/grpc.go
```diff
@@ -0,0 +1,112 @@
...
+	detail := st.Details()[0]
+	errPB, ok := detail.(*pb.ErrorV2)
+	if !ok {
+		return errors.Trace(errIn)
+	}
+
+	// Tries to find the prototype from the registry
+	val, exists := prototypeRegistry.Load(errPB.Name)
+	if !exists {
+		return errors.Trace(
```
lance6716, line 76: maybe we should distinguish the error converted from gRPC and the error of logic

lance6716 (commented):
there are many types of error, maybe give a short explanation of relationship among them

engine/pkg/rpcerror/attributes.go
```diff
@@ -0,0 +1,173 @@
...
+	return false
+}
+
+type grpcStatusCoder interface {
+	grpcStatusCode() codes.Code
+}
+
+// Canceled can be passed as a parameter to Error to
+// mark the grpc error code of this error as Canceled.
+type Canceled struct{}
```
lonng, line 44: Why do we define so many types? Can we define the errors as an enum type? If so, it's easy to implement the `grpcStatusCode` method.

engine/pkg/rpcerror/normalize.go
```diff
@@ -0,0 +1,212 @@
...
+
+	"github.com/pingcap/errors"
+	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	pb "github.com/pingcap/tiflow/engine/enginepb"
+)
+
+type normalizeOpts struct {
```
lonng, line 31: I don't think we need to define a new type `normalizeOpts` because the fields of it are the same as `Prototype[E errorInfo]`.
Why not reuse them directly?

engine/pkg/rpcerror/normalize.go
```diff
@@ -0,0 +1,212 @@
...
+	_, exists := prototypeRegistry.LoadOrStore(opts.name, ret)
+	if exists {
+		log.L().Warn("duplicate error type", zap.String("name", opts.name))
+	}
+
+	return ret
+}
+
+// Gen generates a stackless instance of the given error.
+func (p *Prototype[E]) Gen(e *E) error {
```
lonng, line 90: It's better to use `Normalize()` as the method name.

engine/pkg/rpcerror/grpc.go
liuzix: I think this is fine, because it throws an error indicating that the input cannot be recognized.

engine/pkg/rpcerror/attributes.go
liuzix: Because types can be used as parameters to `Error[E, R]`.

engine/pkg/rpcerror/normalize.go
liuzix: `Normalize()` is used to define error prototypes in `pingcap/error`, not to generate an instance, I think.

engine/pkg/rpcerror/normalize.go
liuzix: `Prototype` has type paramaters, which can make defining `NormalizeOpt` tricky.

engine/pkg/rpcerror/normalize.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/pingcap/errors"
+	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	pb "github.com/pingcap/tiflow/engine/enginepb"
+)
+
+type normalizeOpts struct {
+	name string
```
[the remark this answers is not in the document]
liuzix: I feel like using this package only internally. Any error message intended for outside the system will be generated through `pingcap/errors`. In existing usage cases, some RFC codes already seem unnecessary and unwarranted, such as `ErrPipelineTryAgain` used in TiCDC. I do not think confusing RFC codes with internal error identifiers is a great idea.

engine/pkg/rpcerror/grpc.go
gozssky: How can we guarantee the error name is registered in `prototypeRegistry`? Especially when the server and client are in different processes.

engine/pkg/rpcerror/attributes.go
gozssky: Why not just put `statusCode` and `retryable` as `Error`'s member variables directly?

engine/pkg/rpcerror/grpc.go
```diff
@@ -0,0 +1,112 @@
...
+	"reflect"
+
+	"github.com/gogo/status"
+	"github.com/pingcap/errors"
+	"google.golang.org/grpc/codes"
+
+	pb "github.com/pingcap/tiflow/engine/enginepb"
+)
+
+func tryUnwrapNormalizedError(errIn error) (typeErasedNormalizedError, bool) {
```
maxshuang, line 28: What does `typeErasedNormalizedError ` means literally?

engine/pkg/rpcerror/normalize.go
```diff
@@ -0,0 +1,215 @@
...
+
+// Gen generates a stackless instance of the given error.
+func (p *Prototype[E]) Gen(e *E) error {
+	return &normalizedError[E]{
+		prototype: p,
+		inner:     *e,
+	}
+}
+
+// GenWithStack is similar to Gen but attaches the call stack to it.
```
maxshuang, line 100: I am some kind of confuse about the responsibility of `normalizeError` and `Prototype`.
Can we remain the `Gen` and `GenWithStack` for the `Prototype` which means it will generate some forms of `normalizeError`?
Move some methods like  `Is`, `Convert`, `fromJsonBytes`  to the `normalizeError`?

engine/pkg/rpcerror/normalize.go
```diff
@@ -0,0 +1,215 @@
...
+// is the type name of the type argument to Normalize.
+func WithName(name string) NormalizeOpt {
+	return func(opts *normalizeOpts) {
+		opts.name = name
+	}
+}
+
+var prototypeRegistry = &sync.Map{}
+
+// Prototype is a generator for all instances
```
maxshuang, line 59: IMHO,  a structual error called `normalizeError` has four members: `msg`, `name`, `isRetryable` and `grpcStatusCode`. 
Do we really need a `prototype` which has a generic param and `msg`, `name`?
Why not some methods for `normalizeError` for easy understanding.

engine/pkg/rpcerror/normalize.go
liuzix: The names seem sub-optimal. I will change.

engine/pkg/rpcerror/normalize.go
liuzix: How do you distinguish an `error factory` and an instance of an error?

engine/pkg/rpcerror/normalize.go
```diff
@@ -0,0 +1,215 @@
...
+	return &pb.ErrorV2{
+		Name:    e.name(),
+		Details: e.mustMarshalJSON(),
+	}
+}
+
+func (e *normalizedError[E]) statusCode() codes.Code {
+	if statusCoder, ok := any(e.inner).(grpcStatusCoder); ok {
+		return statusCoder.grpcStatusCode()
+	}
```
maxshuang, line 200: `e.inner` is the generic with `IsErrorInfo` limit. 
Why we need another `grpcStatusCoder ` method here? 
Seems strange.

engine/pkg/rpcerror/grpc.go
maxshuang: I get the scene and the point of `type erasure`.
Are we face the same situation? 
It's not so easy to get the meaning of `typeErasedNormalizedError` and what it does.

engine/pkg/rpcerror/normalize.go
maxshuang: Same question as https://github.com/pingcap/tiflow/pull/5950/files#r905973563

engine/pkg/rpcerror/normalize.go
liuzix: We might not limit the usage of this package to processing gRPC-related error. Multiple interfaces seem good for decoupling.

engine/pkg/rpcerror/grpc.go
liuzix: How do you implement `Is` and `Convert`?

engine/pkg/rpcerror/attributes.go
liuzix: > Why not just put `statusCode` and `retryable` as `Error`'s member variables directly?

It's not convenient for defining an error type with a type definition only. For example

```go
type ErrMyCustomError struct {
    rpcerror.Error[rpcerror.Retryable, rpcerror.Unavailable]

    field1 int
    field2 string
}
```

engine/pkg/rpcerror/grpc.go
liuzix: They would be registered as part of the initialization process of global variables.

gozssky: Does the initialization still work if the client process does not import the package where the error is defined?

maxshuang (approved):

lance6716 (approved):

PR #5955, opened 2022-06-21
cdc: merge two-phase scheduler

overvenus:

Issue Number: ref #4757 

Add two-phase scheduler to reduce latency spike during rolling restart.

Commits are cherry-picked from [fb/latency](https://github.com/pingcap/tiflow/tree/fb/latency) branch.

#### Tests 

 - Unit test

### Release note 

```release-note
Add two-phase scheduler to reduce latency spike during rolling restart.
```


amyangfei:
Should it be `tide/merge-method-merge` instead of `tide/merge-method-rebase`?

overvenus:
> Should it be `tide/merge-method-merge` instead of `tide/merge-method-rebase`?

Both are fine, they all keep commit history, while rebase keep a linear history.

3AceShowHand (approved):

cdc/api/v1/api.go
```diff
@@ -695,6 +696,68 @@ func (h *OpenAPI) ListCapture(c *gin.Context) {
...
+// @Summary Drain captures
+// @Description Drain all tables at the target captures in cdc cluster
+// @Tags capture
+// @Accept json
+// @Produce json
+// @Success 200,202
+// @Failure 500,400 {object} model.HTTPError
+// @Router	/api/v1/captures/drain [post]
+func (h *OpenAPI) DrainCapture(c *gin.Context) {
+	if !h.capture.IsOwner() {
```
lonng, line 709: We use middleware to forward the traffic to the owner and will the path reachable?

scripts/check-diff-line-width.sh
```diff
@@ -23,7 +23,10 @@ BASE_HASH=$(git --no-pager log -E --grep='\(#[0-9]+\)$' -n 1 --format=format:%H)
 WARN_THRESHOLD=80
 ERROR_THRESHOLD=100
 
-git --no-pager diff $BASE_HASH -U0 -- cdc pkg cmd -- ':(exclude)*_gen.go' -- ':(exclude)*_gen_test.go' |
+git --no-pager diff $BASE_HASH -U0 -- cdc pkg cmd \
+	-- ':(exclude)*_gen.go' \
+	-- ':(exclude)*_gen_test.go' \
+	-- ':(exclude)*.pb.go' |
```
sdojjy, line 29: exclude *_mock.go  ?

overvenus: I prefer exclude them in another PR.

pkg/config/scheduler_config.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"time"
+
+	cerrors "github.com/pingcap/tiflow/pkg/errors"
+)
+
+// SchedulerConfig configs TiCDC scheduler.
+type SchedulerConfig struct {
+	HeartbeatTick        int          `toml:"heartbeat-tick" json:"heartbeat-tick"`
```
sdojjy, line 24: from the field name`HeartbeatTick`, I don't know what it means.

sdojjy (approved):

pkg/config/scheduler_config.go
overvenus: Comments added, thanks!

PR #6016, opened 2022-06-23
logutil(cdc, engine): add some util function to logutil for cdc and engine

### LINKED ISSUE #6015: Merge some log util function of engine to tiflow/pkg/logutil

No

Need merge some log util function of engine to tiflow/pkg/logutil

_No response_

_No response_

maxshuang:

Issue Number: close #6015 

Merge some util function of engine to tiflow/pkg/logutil

#### Tests 

 - Unit test

### Release note 

```release-note
 `None`.
```

### DIFF
--- a/cmd/kafka-consumer/main.go
+++ b/cmd/kafka-consumer/main.go
@@ -91,10 +91,13 @@ func init() {
 	flag.StringVar(&key, "key", "", "Private key path for Kafka SSL connection")
 	flag.Parse()
 
-	err := logutil.InitLoggerForCDC(&logutil.Config{
+	err := logutil.InitLogger(&logutil.Config{
 		Level: logLevel,
 		File:  logPath,
-	})
+	},
+		logutil.WithInitGRPCLogger(),
+		logutil.WithInitSaramaLogger(),
+	)
 	if err != nil {
 		log.Panic("init logger failed", zap.Error(err))
 	}
--- a/cmd/kafka-consumer/main.go
+++ b/cmd/kafka-consumer/main.go
@@ -91,7 +91,7 @@ func init() {
 	flag.StringVar(&key, "key", "", "Private key path for Kafka SSL connection")
 	flag.Parse()
 
-	err := logutil.InitLogger(&logutil.Config{
+	err := logutil.InitLoggerForCDC(&logutil.Config{
 		Level: logLevel,
 		File:  logPath,
 	})
--- a/pkg/logutil/context.go
+++ b/pkg/logutil/context.go
@@ -1,0 +1,42 @@
+// Copyright 2022 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logutil
+
+import (
+	"context"
+
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
+)
+
+type ctxLogKeyType struct{}
+
+// CtxLogKey indicates the context key for logger
+// public for test usage.
+var CtxLogKey = ctxLogKeyType{}
+
+// FromContext return the logger in context, or return global logger
+// if logger not found in context
+func FromContext(ctx context.Context) *zap.Logger {
+	if ctxlogger, ok := ctx.Value(CtxLogKey).(*zap.Logger); ok {
+		return ctxlogger
+	}
+
+	return log.L()
+}
+
+// NewContextWithLogger attaches a new logger to context and return the context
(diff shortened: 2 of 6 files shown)


pkg/logutil/context.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
+)
+
+type ctxLogKeyType struct{}
+
+// CtxLogKey indicates the context key for logger
+// public for test usage.
+var CtxLogKey = ctxLogKeyType{}
```
lonng, line 27: Why we make the `CtxLogKey` public?

pkg/logutil/context.go
```diff
@@ -0,0 +1,42 @@
...
+package logutil
+
+import (
+	"context"
+
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
+)
+
+type ctxLogKeyType struct{}
```
lonng, line 23: Can we use `var ctxLogKey = struct{}{}` to avoid defining a new type because we don't want others to initialize an instance for this type?

pkg/logutil/log.go
```diff
@@ -83,6 +84,30 @@ func SetLogLevel(level string) error {
 	return nil
 }
 
+// InitLoggerForCDC initializes logger for CDC, including grpc, sarma etc.
+func InitLoggerForCDC(cfg *Config) error {
```
lonng, line 88: Can we make the `grpc/sarama` related logical as options of `InitLogger` instead of defining a new `InitLoggerForCDC` interface because it is not extensible?
In this way, we will add more and more functions like this if we want to initialize more environments.

Prefer:
```go
func InitLogger(cfg *Config, opts ...LoggerOpt) error {}
```

pkg/logutil/context_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/pingcap/log"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+func TestContextWithLogger(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.TODO()
+	require.Equal(t, FromContext(ctx), log.L())
```
amyangfei, line 29:
```suggestion
	require.Equal(t, log.L(), FromContext(ctx))
```
ditto for L33

cmd/kafka-consumer/main.go
```diff
@@ -91,7 +91,7 @@ func init() {
 	flag.StringVar(&key, "key", "", "Private key path for Kafka SSL connection")
 	flag.Parse()
 
-	err := logutil.InitLogger(&logutil.Config{
+	err := logutil.InitLoggerForCDC(&logutil.Config{
```
amyangfei, line 94: Should `pkg/cmd/util/helper.go` also needs to initialize logger this way?

pkg/logutil/context.go
maxshuang: Yes, it's better.

pkg/logutil/log.go
```diff
@@ -107,24 +132,36 @@ func InitLogger(cfg *Config) error {
...
+	return nil
+}
 
-	var level zapcore.Level
-	err = level.UnmarshalText([]byte(cfg.Level))
-	if err != nil {
-		return errors.Trace(err)
+// InitLoggerWithWriteSyncer initializes a zap logger with specified write syncer.
+// For easy unit test when we use zaptest.Buffer for the zapcore.WriteSyncer
+func InitLoggerWithWriteSyncer(cfg *Config, output, errOutput zapcore.WriteSyncer) error {
```
amyangfei, line 141: Is this function used in test only?

pkg/logutil/context.go
maxshuang: Let user use the `context.Value(CtxLogKey)` freely.
But it can be private.

pkg/logutil/context.go
gozssky: No, if we have another ctxKey, there will be a collision. See https://go.dev/play/p/K-OneBfEsUT.

gozssky: Another example with pointers: https://go.dev/play/p/lT3_zN69m8E.

pkg/logutil/log.go
maxshuang: Currently Yes. But user can customize their `WriteSyncer`.

cmd/kafka-consumer/main.go
maxshuang: No. It doesn't  rely on `grpc` and `sarama` actually.

pkg/logutil/log.go
gozssky: Can we put it in `Config` or an option?

maxshuang: Move to option in latest commit.

pkg/logutil/log.go
maxshuang: Fixed.

pkg/logutil/context.go
maxshuang: Fixed.

cmd/kafka-consumer/main.go
maxshuang: Fix.

amyangfei (approved):

lonng (approved):

PR #6049, opened 2022-06-27
pkg/causality(cdc): Implement a new conflict detector

liuzix:

Issue Number: ref #6050

- Implemented a DAG-based conflict detecting algorithm
- Mostly wait-free implementation.

#### Preliminary benchmarks
```
goos: linux
goarch: amd64
pkg: github.com/pingcap/tiflow/pkg/causality/tests
cpu: AMD Ryzen 9 5900X 12-Core Processor            

BenchmarkLowConflicts
BenchmarkLowConflicts-24       	 2072618	       592.0 ns/op
BenchmarkMediumConflicts
BenchmarkMediumConflicts-24    	 2691535	       485.9 ns/op
BenchmarkHighConflicts
BenchmarkHighConflicts-24      	 1000000	      1484 ns/op

PASS

```
These benchmarks are designed to determine the CPU overhead of conflict detection, so I did not simulate latency of the Workers.

- BenchmarkLowConflicts: workerNum = 8, batchSize = 8, conflictRate = 00006. **1689189 QPS**.
- BenchmarkMediumConflicts: workerNum = 8, batchSize = 8, conflictRate = 0.0155. **2061855 QPS**.
- BenchmarkHighConflicts: workerNum = 8, batchSize = 8, conflictRate = 0.41. **673854 QPS**.

In these benchmarks, I used a conflict model where the conflicts between any pair of batches are **independent**.

#### Tests 

 - Unit test

- No.

- No.

### Release note 

```release-note
None
```


lance6716 (commented):
please also consider an interface for DM's usage, where DM does not need to keep a transaction atomically replicated. Hope we can use same pkg in future.

pkg/causality/internal/slots.go
```diff
@@ -0,0 +1,94 @@
...
+			panic("elem list is not found")
+		}
+		found := false
+		for e := elemList.Front(); e != nil; e = e.Next() {
+			if elem.Equals(e.Value.(E)) {
+				if found {
+					panic("found")
+				}
+				found = true
+				elemList.Remove(e)
```
hicqu, line 90: Here we can break the inner loop?

liuzix: Sure. I added the `panic` to aid debugging.

pkg/causality/conflict_detector.go
```diff
@@ -0,0 +1,154 @@
...
+// limitations under the License.
+
+package causality
+
+import (
+	"sync"
+	"time"
+
+	"go.uber.org/atomic"
+
```
hi-rustin, line 21: Please delete this blank line.

hi-rustin:
It would be nice if we could add a design document. It's a little difficult for me to read it directly now.

tools/check/go.sum
```diff
@@ -27,6 +27,10 @@ cloud.google.com/go v0.84.0/go.mod h1:RazrYuxIK6Kb7YrzzhPoLmCVzl7Sup4NrbKPg8KHSU
 cloud.google.com/go v0.87.0/go.mod h1:TpDYlFy7vuLzZMMZ+B6iRiELaY7z/gJPaqbMx6mlWcY=
 cloud.google.com/go v0.90.0/go.mod h1:kRX0mNRHe0e2rC6oNakvwQqzyDmg57xJ+SZU1eT2aDQ=
 cloud.google.com/go v0.93.3/go.mod h1:8utlLll2EF5XMAV15woO4lSbWQlk8rer9aLOfLh7+YI=
+cloud.google.com/go v0.94.1/go.mod h1:qAlAugsXlC+JWO+Bke5vCtc9ONxjQT3drlTTnAplMW4=
```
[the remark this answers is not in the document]
liuzix: Automatically generated. Did not review by myself. Will fix.

pkg/causality/conflict_detector.go
liuzix: My IDE is separating internal and external imports. Do you think it would make the imports more readable?

tools/check/go.sum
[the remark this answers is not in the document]
liuzix: This seems related to the btree package. I will try to switch to google's own btree, and see if it's a better choice.

pkg/causality/interfaces.go
```diff
@@ -0,0 +1,38 @@
...
+// limitations under the License.
+
+package causality
+
+type (
+	conflictKey = int64
+)
+
+type txnEvent interface {
+	ConflictKeys() []conflictKey
```
hicqu, line 21: It should be an ordered slice? Or there could be dead locks I guess.

pkg/causality/internal/slots.go
```diff
@@ -0,0 +1,90 @@
...
+func (s *Slots[E]) Add(elem E, keys []int64, onConflict func(dependee E)) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	for _, key := range keys {
+		needInsert := true
+		if elemList, ok := s.slots[key%s.numSlots]; ok {
+			for e := elemList.Front(); e != nil; e = e.Next() {
+				e := e.Value.(E)
+				if e.Equals(elem) {
```
hicqu, line 57: Is it possible? I suppose it's not. So it's only necessary to call `onConflict` on the last element.

hicqu: Only call `onConflict` on the last element will be a obvious improvement because it reduces the relationship size from O(n^2) to O(n), if there are n pending txns in the detecter.
If the improvement is possible, I also suggest to remove the `Sleep` logic from `ConflictDetecter.Add`.

pkg/causality/interfaces.go
liuzix: We are not locking individual slots. So this should not be a problem.

pkg/causality/internal/slots.go
liuzix: Very good observation!

liuzix: Benchmark results have been updated.

pkg/causality/conflict_detector.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/pingcap/tiflow/pkg/causality/internal"
+)
+
+// ConflictDetector implements a logic that dispatches transaction
+// to different workers in a way that transactions modifying the same
+// keys are never executed concurrently and have their original orders
+// preserved.
+type ConflictDetector[Worker worker[Txn], Txn txnEvent] struct {
+	workers []Worker
+	slots   *internal.Slots[*internal.Node]
```
nongfushanquan, line 31: could you add some comments about the slots,finishedTxnQuene and resolvedTxnQueue.

pkg/causality/interfaces.go
```diff
@@ -0,0 +1,38 @@
...
+// This is temporary solution before the actual data structure
+// of and transaction is decided on.
+// TODO remove this.
+type OutTxnEvent[T txnEvent] struct {
+	Txn      T
+	Callback func(errIn error)
+}
+
+type worker[Txn txnEvent] interface {
+	Add(txn *OutTxnEvent[Txn])
```
hicqu, line 37: How about `Add(txn Txn, unlock func())`? So `OutTxnEvent` can be removed.

hicqu:
rest LGTM.

pkg/causality/conflict_detector.go
```diff
@@ -0,0 +1,169 @@
...
+	resolvedTxnQueue *containers.SliceQueue[resolvedTxn[Txn]]
+
+	// nextWorkerID is used to dispatch transactions round-robin.
+	nextWorkerID atomic.Int64
+
+	wg      sync.WaitGroup
+	closeCh chan struct{}
+}
+
+type finishedTxn[Txn txnEvent] struct {
```
CharlesCheung96, line 54: Is it possible to replace `finishedTxn and resolvedTxn ` with a generic version of `Node`,  which holds a `txn` field internally.

pkg/causality/internal/node.go
```diff
@@ -0,0 +1,241 @@
...
+
+// Node is a node in the dependency graph used
+// in conflict detection.
+type Node struct {
+	id int64 // immutable
+
+	mu             sync.Mutex
+	conflictCounts map[workerID]int
+	// TODO maybe google's btree is better?
+	dependers  btree.Map[nodeID, *Node]
```
maxshuang, line 46: Why we need a btree.Map here? Or we just need an order map?

liuzix: Yes. An ordered map seems enough. But I have not looked into if there's any production-ready ordered map implementation in Go.

hicqu: Seems one node can only have 0 or 1 dependers?

hicqu (approved):

pkg/causality/internal/slots.go
```diff
@@ -0,0 +1,90 @@
...
+				} else {
+					onConflict(lastElem)
+				}
+			}
+		} else {
+			s.slots[key%s.numSlots] = list.New()
+		}
+
+		if needInsert {
+			s.slots[key%s.numSlots].PushBack(elem)
```
maxshuang, line 68: I have come up with an idea.
Since we use `slot` to compress the key space and construct dependency DAG therefore, we may suffer from some `false positive` dependency([a]->[a',b]->[a,c]) and long path depenpency link([a]->[a,b]->[a,c]).

May be we can `increasing path compression` in `Union-find` algorithm to find find more accurate the dependency DAG and shorten the path, so we can resolve the node faster.

maxshuang (approved):

pkg/causality/internal/slots.go
liuzix: I'm confused by your notation for paths. Can you clarify what `DAG([a]->[[a,b],[a,c]]) and shorten the path([a]->[[a,b],[a,c]])` means?

pkg/causality/conflict_detector.go
hi-rustin: It looks OK. But it's different from what we were doing. And it may be difficult for people to follow this rule in practice.

pkg/causality/conflict_detector.go
```diff
@@ -0,0 +1,171 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package causality
+
+import (
+	"sync"
+
+	"go.uber.org/atomic"
+
```
hi-rustin, line 20:
```suggestion
```

pkg/causality/conflict_detector.go
```diff
@@ -0,0 +1,171 @@
...
+		// resolved too, thus incurring potential risk of
+		// deadlocking.
+		d.resolvedTxnQueue.Push(txnResolvedEvent[Txn]{
+			txn:      txn,
+			node:     node,
+			workerID: workerID,
+		})
+	})
+
+	return nil
```
hi-rustin, line 111: Can we remove this useless return value?

pkg/causality/conflict_detector.go
```diff
@@ -0,0 +1,171 @@
...
+			}
+		}
+	}
+}
+
+// sendToWorker should not call txn.Callback if it returns an error.
+func (d *ConflictDetector[Worker, Txn]) sendToWorker(
+	txn Txn, node *internal.Node, unlock func(), workerID int64,
+) {
+	if workerID == -1 {
```
hi-rustin, line 164: Better to make `-1` be a const.

pkg/causality/interfaces.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2022 PingCAP, Inc.
```
hi-rustin, line 1: I prefer to rename this file to `worker.go`, because `interfaces` means nothing.

pkg/causality/internal/node.go
```diff
@@ -0,0 +1,276 @@
...
+
+// ID returns the node's ID. For debug only.
+func (n *Node) ID() int64 {
+	return n.id
+}
+
+// Free must be called if a node is no longer used.
+// We are using sync.Pool to lessen the burden of GC.
+func (n *Node) Free() {
+	if n.id == -1 {
```
hi-rustin, line 92: Better to make `-1` be a const.

pkg/causality/tests/driver.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
+	"go.uber.org/atomic"
+	"go.uber.org/zap"
+
```
hi-rustin, line 24:
```suggestion
```

pkg/causality/tests/worker.go
```diff
@@ -0,0 +1,102 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"sync"
+
+	"github.com/pingcap/errors"
+
```
hi-rustin, line 20:
```suggestion
```

hi-rustin (commented):
some nits

pkg/causality/internal/slots.go
```diff
@@ -0,0 +1,90 @@
...
+type Eq[T any] interface {
+	Equals(other T) bool
+}
+
+// Slots implements slot-based conflict detection.
+// It holds references to E, which can be used to build
+// a DAG of dependency.
+type Slots[E Eq[E]] struct {
+	mu    sync.Mutex
+	slots map[int64]*list.List
```
hi-rustin, line 31: Can we use `sync.Map` here?

pkg/causality/internal/node.go
```diff
@@ -0,0 +1,276 @@
...
+	btreeFreeList = btree.NewFreeListG[*Node](1024)
+)
+
+// Node is a node in the dependency graph used
+// in conflict detection.
+type Node struct {
+	id int64 // immutable
+
+	mu             sync.Mutex
+	conflictCounts map[workerID]int
```
hi-rustin, line 50: Maybe it might be better to add some comments for this field.

pkg/causality/internal/slots.go
liuzix: We still need a mutex to make `Add` race-free.

hi-rustin: Got it! Thanks!

hi-rustin (approved):
Thanks a lot for you working on this!  💚 💙 💜 💛 ❤️

pkg/causality/internal/slots.go
maxshuang: `DAG([a]->[[a,b],[a,c]])` need to be corrected as `DAG([a]->[[a',b],[a,c]])`.
It means, for a single slot, key 1(%3), 4(%3), 7(%3) may be assigned to the this slot, so we may have some `false positive`  dependency like [7, x] -> [4,x] -> [1, x] -> [7',x]. 
Every time we resolve the dependency, it need to follow the dependency chain, like [7',x] first,  [1, x] second, and then [4,x], [7, x] for last.

But the [7,x] is depent on [7',x] actually, not [4,x]. So if we can correct the dependency chain increasingly, it may be helpful. For example:
1. [7, x] -> [4,x] -> [1, x] -> [7',x]
2. [4,x] -> [1, x] -> [7',x]
               |
     [7,x]--
3. [4,x] -> [1, x] -> [7',x]
                               |
                     [7,x]--
 As we find the right dependency item, after we resolve [7',x], we can resolve [7,x] next.
 Just a shot in my mind.

Issue #5344, opened 2022-05-06
after print "rollback checkpoint" in log, unistore folder size grows


lance6716:
### What did you do?

https://github.com/pingcap/tiflow/blob/630896740566eb0f7a81062506e92177e755f404/dm/syncer/checkpoint.go#L715-L727

mock the logic of rollback checkpoint

```
	for i := 0; i < 10000; i++ {
		err = tracker.DropTable(tbl)
		c.Assert(err, IsNil)

		err = tracker.CreateSchemaIfNotExists("test")
		c.Assert(err, IsNil)
		err = tracker.CreateTableIfNotExists(tbl, oriTi)
		c.Assert(err, IsNil)
	}
```

during the test

```
# lance @ LAPTOP-AK9BOS6J in ~/go/src/github.com/pingcap/tiflow on git:630896740
 x [19:38:27]
$ du -h /tmp/tidb-unistore-temp944770732
16M     /tmp/tidb-unistore-temp944770732/kv
16M     /tmp/tidb-unistore-temp944770732

# lance @ LAPTOP-AK9BOS6J in ~/go/src/github.com/pingcap/tiflow on git:630896740
 x [19:38:42]
$ du -h /tmp/tidb-unistore-temp944770732
104M    /tmp/tidb-unistore-temp944770732/kv
104M    /tmp/tidb-unistore-temp944770732

# lance @ LAPTOP-AK9BOS6J in ~/go/src/github.com/pingcap/tiflow on git:630896740
 x [19:40:22]
$ du -h /tmp/tidb-unistore-temp944770732
195M    /tmp/tidb-unistore-temp944770732/kv
195M    /tmp/tidb-unistore-temp944770732
```

So if the task is resumed for many times (which means the task has a resumable error), the size of unistore folder will grow.

### What did you expect to see?

the size of `/tmp/tidb-unistore-temp944770732` has an upper-bound

### Versions of the cluster

v5.3.1

### current status of DM cluster (execute `query-status <task-name>` in dmctl)

```console
(paste current status of DM cluster here)
```

lance6716:
after some discussion, this space may be occupied by DDL history job and we have no way to clean it. Maybe the fix that we close schema tracker and remove the unistore folder after the task is paused.

Or we think about #5334

GMHDBJD:
> Maybe the fix is that we close schema tracker

looks good to me

lance6716:
[...]

PR #6052, opened 2022-06-27
syncer(dm): init schemaTracker when syncer run

### LINKED ISSUE #6014: separate schemaTracker's construction and initialization

buchuitoudegou:

Issue Number: close #6014 

The validator will get the old schemaTracker because the syncer re-builds the schemaTracker every time it starts. To avoid that, I'm about to decouple the creating and initialization of the schemaTracker so that the syncer will only initialize the schemaTracker when it starts to run and the schemaTracker remains the same instance in the whole lifecycle of the task.

1. create dump schemaTracker when initializing the syncer
2. initialize the schemaTracker when running

#### Tests 

 - Unit test

### Release note 

```release-note
`None`
```


lance6716:
> schemaTracker remains the same instance in the whole lifecycle of the task

this will cause https://github.com/pingcap/tiflow/issues/5344 and https://github.com/pingcap/tiflow/issues/4187

buchuitoudegou:
> > schemaTracker remains the same instance in the whole lifecycle of the task
> 
> this will cause #5344 and #4187

The schemaTracker can still be `Close`'d every time the task is paused, where the persistent storage is removed. Just avoid creating a new instance.
https://github.com/pingcap/tiflow/blob/9035b2fd8a686edb10e8fab325ac1ccfe05c089d/dm/pkg/schema/tracker.go#L400

lance6716 (commented):
And also can we add a test of "two tracker"?
rest lgtm

dm/pkg/schema/tracker.go
```diff
@@ -99,16 +99,25 @@ type DownstreamTableInfo struct {
...
-func NewTracker(
+func (tr *Tracker) Init(
 	ctx context.Context,
 	task string,
 	sessionCfg map[string]string,
 	downstreamConn *dbconn.DBConn,
 	logger log.Logger,
-) (*Tracker, error) {
+) error {
+	if tr == nil {
```
D3Hunter, line 118: when will tr be nil?

if it's for test, we can change test instead

dm/pkg/schema/tracker.go
```diff
@@ -233,22 +242,39 @@ func NewTracker(
...
 
-	return &Tracker{
-		storePath: storePath,
-		store:     store,
-		dom:       dom,
-		se:        se,
-		dsTracker: dsTracker,
-	}, nil
+// NewTracker creates a new tracker. It's preserved for test.
+func NewTracker(
```
D3Hunter, line 265: prefer to name `NewDumpTracker` as `NewTracker`, and remove this one.

for test, we change them to `Init`

dm/pkg/schema/tracker.go
buchuitoudegou: This line is copied from the old version. I remember it will let the `load_task` IT fail if deleted (probably this test case tries to `pause-task` before initializing the syncer). will take a look tomorrow.

dm/pkg/schema/tracker.go
buchuitoudegou: Maybe leaving a function invoking `NewTracker` and subsequently `Init` is better, so as to make the test more succinct.

D3Hunter: then name it explicitly as something like NewTrackerForTest. `NewTracker` is how we name normal ctor

lance6716: IDE should be able to easily rename them all

buchuitoudegou: 8dceb81

dm/syncer/syncer.go
```diff
@@ -1696,7 +1698,13 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 		}
 	}
 
-	s.schemaTracker, err = schema.NewTracker(ctx, s.cfg.Name, s.cfg.To.Session, s.downstreamTrackConn, s.tctx.L())
+	if s.schemaTracker == nil {
+		s.schemaTracker, err = schema.NewTracker(ctx, s.cfg.Name, s.cfg.To.Session, s.downstreamTrackConn, s.tctx.L())
```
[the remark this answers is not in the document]
buchuitoudegou: 8dceb81

dm/pkg/schema/tracker.go
```diff
@@ -99,16 +99,25 @@ type DownstreamTableInfo struct {
 	WhereHandle *sqlmodel.WhereHandle
 }
 
-// NewTracker creates a new tracker. `sessionCfg` will be set as tracker's session variables if specified, or retrieve
+// NewDumpTracker simply returns an empty Tracker,
```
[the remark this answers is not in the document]
buchuitoudegou: 8dceb81

dm/pkg/schema/tracker.go
```diff
@@ -233,22 +242,40 @@ func NewTracker(
...
+		return err
 	}
 
 	// init downstreamTracker
 	dsTracker := &downstreamTracker{
 		downstreamConn: downstreamConn,
 		tableInfos:     make(map[string]*DownstreamTableInfo),
 	}
+	tr.Lock()
+	defer tr.Unlock()
```
D3Hunter, line 254: now this lock is used in 3 places, please refine comments in `Tracker` definition

buchuitoudegou: [77cfbca](https://github.com/pingcap/tiflow/pull/6052/commits/77cfbcab957523dfd1092a03463b27183f28436f)

D3Hunter (approved):

Issue #6013, opened 2022-06-23
executor manager stop hangs the whole server master


amyangfei:
### What did you do?

chaos test failure case: https://github.com/pingcap/tiflow/runs/7013417892?check_suite_focus=true

detail logs:
[chaos-base-logs.network-emulation-dataflow.zip](https://github.com/pingcap/tiflow/files/8966604/chaos-base-logs.network-emulation-dataflow.zip)
 

### What did you see instead?

- Job is submitted, but job master is not created.
- executor manager hangs

It seems `errctx.Err()` doesn't return a nil error

```go
goroutine 1168 [semacquire, 5 minutes]:
sync.runtime_Semacquire(0xc00045fc00?)
        runtime/sema.go:56 +0x25
sync.(*WaitGroup).Wait(0x4436a0?)
        sync/waitgroup.go:136 +0x52
github.com/pingcap/tiflow/engine/servermaster.(*ExecutorManagerImpl).Stop(0x281e5ea?)
        github.com/pingcap/tiflow/engine/servermaster/executor_manager.go:278 +0x1f
github.com/pingcap/tiflow/engine/servermaster.(*Server).runLeaderService.func12.1()
        github.com/pingcap/tiflow/engine/servermaster/server.go:763 +0x2c
github.com/pingcap/tiflow/engine/servermaster.(*Server).runLeaderService.func12()
        github.com/pingcap/tiflow/engine/servermaster/server.go:767 +0x8a
golang.org/x/sync/errgroup.(*Group).Go.func1()
        golang.org/x/sync@v0.0.0-20220513210516-0976fa681c29/errgroup/errgroup.go:74 +0x64
created by golang.org/x/sync/errgroup.(*Group).Go
        golang.org/x/sync@v0.0.0-20220513210516-0976fa681c29/errgroup/errgroup.go:71 +0xa5
```

Besides we can find a lot of `github.com/pingcap/tiflow/engine/pkg/errctx.(*errCtx).Done.func1.1()` goroutines

```bash
$ grep "errctx/ctx.go.41" server-master-0_goroutines.log |wc -l
856
```

### Versions of the cluster

Dataflow Engine version (run `df-master -V` or `df-executor -V`):

master@49960d68d8b795758c0a1322219eea76b81a32d0

gozssky:
It is unnecessary to create a goroutine for each `errctx`. I will optimize it first.

gozssky:
[...]

PR #6061, opened 2022-06-27
pkg/errctx(engine): avoid creating goroutinue in errctx

gozssky:

Issue Number: ref https://github.com/pingcap/tiflow/issues/6013.

 Avoid creating goroutinue in errctx.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/pkg/errctx/ctx.go
```diff
@@ -16,45 +16,57 @@ package errctx
 import (
 	"context"
 	"sync"
+	"time"
 )
 
 type errCtx struct {
-	context.Context
-	center *ErrCenter
+	ctx context.Context
```
liuzix, line 23: Why are we not embedding it directly?

gozssky: The [Uber Go Style Guide](https://github.com/uber-go/guide/blob/master/style.md#avoid-embedding-types-in-public-structs) doesn't suggest embedding types in public structs. But `errCtx` is not public struct, so it's okay to embed it directly.

gozssky: Done.

engine/pkg/errctx/ctx.go
```diff
@@ -20,41 +20,45 @@ import (
...
+	c.mu.Lock()
+	if c.err != nil {
+		c.mu.Unlock()
+		return // already canceled
+	}
+	if c.Context.Err() != nil {
+		c.err = c.Context.Err()
+		c.mu.Unlock()
+		return // parent is canceled
+	}
```
amyangfei, line 51: Better to add a unit test to cover this logic.
Besides is it necessary to record parent context error instead of the passing in error?

gozssky: > // If Done is not yet closed, Err returns nil.
	// If Done is closed, Err returns a non-nil error explaining why:
	// Canceled if the context was canceled
	// or DeadlineExceeded if the context's deadline passed.
	// After Err returns a non-nil error, successive calls to Err return the same error.
	Err() error

`Err()` shouldn't return a different error if it has returned the parent context error.

gozssky:
e2e tests failed. Because some structures may store the `context.Context` passed by the caller, which is not a good practice.

gozssky:
@liuzix @amyangfei @maxshuang e2e test passed. PTAL thanks.

amyangfei (approved):

maxshuang (approved):

Issue #6094, opened 2022-06-28
Some test suites have not been migrated to testify


amyangfei:
### Is your feature request related to a problem?

ref: https://github.com/pingcap/tiflow/issues/2164

```go
cdc/model/owner_test.go
19:     "github.com/pingcap/check"

pkg/util/testleak/leaktest.go
29:     "github.com/pingcap/check"

pkg/util/testleak/fake.go
21:     "github.com/pingcap/check"

pkg/txnutil/gc/gc_manager_test.go
21:     "github.com/pingcap/check"

pkg/txnutil/gc/gc_service_test.go
20:     "github.com/pingcap/check"
```

### Describe the feature you'd like

- [x] All tests in cdc packages use `github.com/stretchr/testify` instead of `github.com/pingcap/check`
- [ ] Add a script to verify that `goleak` is enabled for each package.
- [ ] Remove `leak_test` entry from Makefile after it is removed from CI. ref: https://github.com/ti-community-infra/configs/pull/644

hi-rustin:
@amyangfei Can we disable the leak test now?

amyangfei:
Leak tests have been removed after `release-5.4`

hi-rustin:
Thanks for working on this!

PR #6096, opened 2022-06-28
tests(cdc): migrate pingcap/check to testify, remove unused leaktest

amyangfei:

Issue Number: ref #6094

- Use testify in pkg/txnutil/gc package
- Remove unused leaktest

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


CharlesCheung96:
Perhaps we should add a new script to verify that `goleak` is enabled for each `pkg`.

amyangfei:
> Perhaps we should add a new script to verify that `goleak` is enabled for each `pkg`.

Good point @CharlesCheung96

CharlesCheung96:
image
Whether `leak-test` should be removed?

amyangfei:
> image
> 
> Whether `leak-test` should be removed?

Yes, I have communicated with CI colleague

CharlesCheung96 (approved):

amyangfei:
ptal @maxshuang

maxshuang (approved):

PR #6110, opened 2022-06-28
cdc: drain waits for owner resign and table move out

overvenus:

Issue Number: ref #4757 

Drain TiCDC ownership and tables before shutdown.

Tests show there is no (or little) latency when a TiCDC shutdown.

#### Tests 

 - Unit test

Run sysbench

```shell
# Upstream workload.
sysbench --db-driver=mysql --mysql-host=172.16.5.33 --mysql-port=48104 --mysql-user=root --create_secondary=off \
    --mysql-db=testtable1k --tables=1000 --table-size=1000000 --time=100000000 --threads=5 oltp_update_non_index run

# stop TiCDC manually
$ date && systemctl stop cdc-47915 && date
Tue Jun 28 20:48:32 CST 2022
Tue Jun 28 20:49:31 CST 2022
```

image

### Release note 

```release-note
None
```


cdc/capture/capture.go
```diff
@@ -542,27 +569,70 @@ func (c *Capture) AsyncClose() {
...
+		ticker := time.NewTicker(drainInterval)
+		defer ticker.Stop()
+		for {
+			complete := c.drainImpl(ctx)
+			if complete {
+				return
+			}
+			ticker.Reset(drainInterval)
+			select {
+			case <-ctx.Done():
```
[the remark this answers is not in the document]
overvenus: Comments added.

hicqu (approved):

cdc/capture/capture.go
```diff
@@ -542,27 +569,73 @@ func (c *Capture) AsyncClose() {
...
-	// Disable drain for now.
-	//
-	// TODO close done when
-	// 1. there is no table in the capture, and
-	// 2. the capture is not owner.
+	const drainInterval = 100 * time.Millisecond
 	done := make(chan struct{})
-	close(done)
+	go func() {
+		defer close(done)
```
3AceShowHand, line 578: how about groups 2 `defer` together

cdc/capture/capture.go
```diff
@@ -542,27 +569,73 @@ func (c *Capture) AsyncClose() {
...
+		ticker := time.NewTicker(drainInterval)
+		defer ticker.Stop()
+		for {
```
3AceShowHand, line 581: how about rewrite the for loop like

```
		for {
			select {
			case <- ticker.C:
				if c.drainImpl(ctx) {
					return
				}
			case <- ctx.Done():
				return
			}
		}
```

cdc/capture/capture.go
```diff
@@ -542,27 +569,73 @@ func (c *Capture) AsyncClose() {
...
-	log.Info("draining the capture, removing all tables in the capture")
+func (c *captureImpl) Drain(ctx context.Context) <-chan struct{} {
+	log.Info("draining capture, removing all tables in the capture")
 
-	// Disable drain for now.
-	//
-	// TODO close done when
-	// 1. there is no table in the capture, and
-	// 2. the capture is not owner.
+	const drainInterval = 100 * time.Millisecond
```
3AceShowHand, line 575: the `drainInterval` looks too small, how about set as 1 seconds?

cdc/processor/manager.go
```diff
@@ -197,13 +218,22 @@ func (m *Manager) handleCommand() error {
 	case commandTpWriteDebugInfo:
 		w := cmd.payload.(io.Writer)
 		m.writeDebugInfo(w)
+	case commandTpQueryTableCount:
+		count := 0
+		for _, p := range m.processors {
+			count += len(p.GetAllCurrentTables())
```
3AceShowHand, line 224: `GetAllCurrentTables` is a method of the `TableExecutor` interface, we aims to make the `TableExecutor` one per table, so the method could be moved to the `Processor` interface if it will be add in the future.

cdc/scheduler/internal/tp/scheduler_drain_capture.go
```diff
@@ -76,7 +89,14 @@ func (d *drainCaptureScheduler) Schedule(
 	defer d.mu.Unlock()
 
 	if d.target == captureIDNotDraining {
-		return nil
+		stopping := checkStoppingCapture(captures)
```
3AceShowHand, line 92: how about add a comment to make others know that `target` can be set by signal.

3AceShowHand:
rest LGTM

sdojjy (approved):

cdc/capture/capture.go
overvenus: > how about rewrite the for loop like
> 
> ```
> 		for {
> 			select {
> 			case <- ticker.C:
> 				if c.drainImpl(ctx) {
> 					return
> 				}
> 			case <- ctx.Done():
> 				return
> 			}
> 		}
> ```

It waits a tick before the first drain, relevantly slow.

cdc/capture/capture.go
overvenus: I think 100ms is okay, as it does not have heavy CPU workload nor IO workload.

cdc/scheduler/internal/tp/scheduler_drain_capture.go
```diff
@@ -76,7 +89,14 @@ func (d *drainCaptureScheduler) Schedule(
 	defer d.mu.Unlock()
 
 	if d.target == captureIDNotDraining {
```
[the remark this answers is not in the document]
overvenus: There is a `getTarget` already, and it requires a lock. If we call `getTarget` here, deadlock occurs.

cdc/processor/manager.go
overvenus: Good point, we can do it in a later PR.

3AceShowHand (approved):

Issue #5928, opened 2022-06-20
Tracking issue for refactor sink


hi-rustin:
Implement MQ sink and MySQL sink in asynchronous mode to improve the throughput of the sink and reduce the coupling of modules.

- https://github.com/pingcap/tiflow/pull/7214
- [Add MQ worker batch flush duration](https://github.com/pingcap/tiflow/pull/7263)
- Add flushed rows dashboards in the dataflow row
- Add more metrics for progress tracking
- Add table sink wrapper and table progress heap https://github.com/pingcap/tiflow/pull/7453

hi-rustin:
Enable pull-based sink preparations:
1. [ ] Fix all panic issue
2. [ ] Solve the no-data table advance issue
3. [ ] Add sorter stats info
4. [ ] Handle err from `sinkMnager.GetTableStats`
5. [ ] Use SourceManger in SinkMnager instead of SortEngine and move mounted iter into source manager
6. [ ] Add test for pullerWrapper and SourceManger
7. [ ] Add test for redo log worker
8. [ ] Add test for generating redo task function

PR #6118, opened 2022-06-29
sinkv2(ticdc): fix update resolved ts not taking effect issue

hi-rustin:

- delete buffer after flush
- set correct eventID
- add more tests

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```


cdc/sinkv2/tablesink/event_table_sink.go
```diff
@@ -58,18 +74,22 @@ func (e *eventTableSink[E]) UpdateResolvedTs(resolvedTs model.ResolvedTs) {
 		return
 	}
 	resolvedEvents := e.eventBuffer[:i]
+	e.eventBuffer = append(make([]E, 0, len(e.eventBuffer[i:])), e.eventBuffer[i:]...)
```
asddongmen, line 77: How about add a comment to explain why do you to reset the `eventBuffer` in this way.

asddongmen (approved):

cdc/sinkv2/eventsink/event.go
```diff
@@ -32,7 +32,7 @@ type CallbackFunc func()
 type CallbackableEvent[E TableEvent] struct {
 	Event       E
 	Callback    CallbackFunc
-	TableStatus *pipeline.TableState
+	TableStatus pipeline.TableState
```
overvenus, line 35: Why need `TableStatus` and why change to value?

hi-rustin: It's a mistake. Fixed and added test for it.

hicqu (approved):

Issue #5066, opened 2022-03-30
Nondeterministic DDLs may cause more data inconsistency


lichunzhu:
### What did you do?

Use DM-master to replicate DDL `alter table add column x datetime default current_timestamp` to downstream. The downstream/upstream table already has some data in tables.

### What did you expect to see?

The upstream/downstream data can be consistent.

### What did you see instead?

DM upstream/downstream tables x column diff.

### Versions of the cluster

v6.0.0

### current status of DM cluster (execute `query-status <task-name>` in dmctl)

```console
(paste current status of DM cluster here)
```

lichunzhu:
# Root Cause
Some kind of ddls will rewrite data in tables with uncertainty:
`add column default current_timestamp`
# Fix
We can use either way to fix this problem:
1. alert an error immediately and notify users to fix data after this kind of ddl. Like we did for add not fully dropped columns now.
2. fix the data automatically for users.

lance6716:
For MySQL 5.7, AFAIK the undetermistic DDL is DEFAULT CURRENT_TIMESTAMP. MySQL 8.0.13 will have more cases.

https://stackoverflow.com/questions/46134550/mysql-set-default-id-uuid

So we can check if the CURRENT_TIMESTAMP in binlog header is the same as the time of DDL

lance6716:
https://dev.mysql.com/doc/refman/5.6/en/server-system-variables.html#sysvar_timestamp

> Set the time for this client. This is used to get the original timestamp if you use the binary log to restore rows.

D3Hunter:
> For MySQL 5.7, AFAIK the only undetermistic DDL is DEFAULT CURRENT_TIMESTAMP. MySQL 8.0.13 will have more cases.
> 
> https://stackoverflow.com/questions/46134550/mysql-set-default-id-uuid
> 
> So we can check if the CURRENT_TIMESTAMP in binlog header is the same as the time of DDL

seems header timestamp is the ddl execute time https://dev.mysql.com/doc/internals/en/unsafe-statements.html

https://github.com/mysql/mysql-server/blob/8d8c986e5716e38cb776b627a8eee9e92241b4ce/sql/item_timefunc.cc#L1997

[...]

PR #6119, opened 2022-06-29
syncer(dm): add the function GetTimezoneByStatusVars()

lyzx2001:

Issue Number: ref #5066 

The function GetTimezoneByStatusVars() is added to retrieve the upstream time zone for "ALTER TABLE ADD COLUMN x DATETIME", but not for "ALTER TABLE ADD COLUMN x TIMESTAMP", since there is no time zone information for the latter one. This function will be used in further fixing issue #5066.

It uses the function statusVarsToKV() to retrieve the time zone and convert it to string. If the result is "SYSTEM", replace it with Syncer.upstreamTZ, in order to guarantee that the absolute time zone of upstream will be synchronized to downstream.

Also, I have added the unit test for this function. The input data for this unit test were retrieved using the function Dump() in go-mysql.

#### Tests 

 - Unit test
 

No.

No.

### Release note 

```release-note
None
```


dm/pkg/binlog/event/util.go
```diff
@@ -408,6 +408,27 @@ func GetCharsetCodecByStatusVars(statusVars []byte) (encoding.Encoding, error) {
...
 
+// GetTimezoneByStatusVars returns the timezone of upstream for "datetime" type alter table commands
+// For "ALTER TABLE ADD x datetime", not for "ALTER TABLE ADD x timestamp".
+func GetTimezoneByStatusVars(statusVars []byte, upstreamTZStr string) (string, error) {
+	vars, err := statusVarsToKV(statusVars)
+	b, ok := vars[QTimeZoneCode]
+
+	if !ok {
+		if err == nil {
+			// only happen when this is a dummy event generated by DM
```
lance6716, line 419: this comment should be updated, as in the comment of function some DDL will also not write timezone to status vars

dm/pkg/binlog/event/util_test.go
```diff
@@ -133,3 +139,38 @@ func (t *testUtilSuite) TestStatusVarsToKV(c *C) {
...
+			nil,
+		},
+	}
+
+	for _, t := range testCases {
+		var upstreamTZStr string // to stimulate Syncer.upstreamTZ
+		upstreamTZStr = "+0:00"
+		vars, err := GetTimezoneByStatusVars(t.input, upstreamTZStr)
+		if t.err != nil {
+			c.Assert(err.Error(), Equals, t.err.Error())
```
lance6716, line 170: please also use testify package in this file. Also change the old tests.

dm/tests/many_tables/run.sh
```diff
@@ -16,6 +16,7 @@ function prepare_data() {
 		for j in $(seq 2); do
 			run_sql "INSERT INTO many_tables_db.t$i VALUES ($j,${j}000$j),($j,${j}001$j);" $MYSQL_PORT1 $MYSQL_PASSWORD1
 		done
+		run_sql "INSERT INTO many_tables_db.t$i VALUES (9, 90009);" $MYSQL_PORT1 $MYSQL_PASSWORD1
```
lance6716, line 19: changes of this file should not be included

dm/pkg/binlog/event/util.go
```diff
@@ -408,6 +408,27 @@ func GetCharsetCodecByStatusVars(statusVars []byte) (encoding.Encoding, error) {
...
+			err = fmt.Errorf("Timezone not found in status_vars %v", statusVars)
+		}
+		return "", err
+	}
+
+	s := string(b[1:])
```
D3Hunter, line 425: plz leave a comment about the structure of this value

dm/pkg/binlog/event/util.go
```diff
@@ -408,6 +408,27 @@ func GetCharsetCodecByStatusVars(statusVars []byte) (encoding.Encoding, error) {
...
 }
 
+// GetTimezoneByStatusVars returns the timezone of upstream for "datetime" type alter table commands
+// For "ALTER TABLE ADD x datetime", not for "ALTER TABLE ADD x timestamp".
+func GetTimezoneByStatusVars(statusVars []byte, upstreamTZStr string) (string, error) {
+	vars, err := statusVarsToKV(statusVars)
+	b, ok := vars[QTimeZoneCode]
+
+	if !ok {
+		if err == nil {
```
D3Hunter, line 418: this part of code is quite anti-pattern, seems imitate logic in getSQLMode(GetServerCollationByStatusVars too). it's introduced in https://github.com/pingcap/dm/pull/1176/files, to fix some undocumented var which may causes syncer stop. 

we should leave a comment to explain why we didn't check error first. or better to refactor this part of code, we don't care about undocumented vars anyway, it should not return error.

D3Hunter: fired an issue https://github.com/pingcap/tiflow/issues/6121

lyzx2001: I have modified this part of the code. Please take a look at whether it corresponds to what you meant.

dm/tests/many_tables/run.sh
lyzx2001: Have removed this part of the changes.

dm/pkg/binlog/event/util_test.go
lyzx2001: Done.

dm/pkg/binlog/event/util.go
lyzx2001: Done.

dm/pkg/binlog/event/util.go
```diff
@@ -408,6 +408,26 @@ func GetCharsetCodecByStatusVars(statusVars []byte) (encoding.Encoding, error) {
 	}
 }
 
+// GetTimezoneByStatusVars returns the timezone of upstream for "datetime" type alter table commands.
+// For "ALTER TABLE ADD COLUMN x DATETIME", not for "ALTER TABLE ADD COLUMN x TIMESTAMP".
+func GetTimezoneByStatusVars(statusVars []byte, upstreamTZStr string) (string, error) {
+	vars, err := statusVarsToKV(statusVars)
+	b, _ := vars[QTimeZoneCode]
```
D3Hunter, line 415: didn't check whether `QTimeZoneCode` exist. previous impl is ok, but maybe need add some comments about why we didn't check error first as usual.

as of `undocumented vars`, it's related to `statusVarsToKV`, we can refactor it later.

dm/pkg/binlog/event/util_test.go
```diff
@@ -17,24 +17,27 @@
 package event
 
 import (
+	"github.com/stretchr/testify/require"
```
lance6716, line 20: configurate the IDE to group and sort the import, like

https://www.jetbrains.com/help/go/creating-and-optimizing-imports.html#use-the-goimports-local-grouping

dm/pkg/binlog/event/util_test.go
```diff
@@ -123,13 +126,50 @@ func (t *testUtilSuite) TestStatusVarsToKV(c *C) {
...
-	for _, t := range testCases {
-		vars, err := statusVarsToKV(t.input)
-		if t.err != nil {
-			c.Assert(err.Error(), Equals, t.err.Error())
+	for _, test := range testCases {
+		vars, err := statusVarsToKV(test.input)
+		if test.err != nil {
+			require.Equal(t, test.err.Error(), err.Error())
+		} else {
+			require.Nil(t, err)
```
lance6716, line 134:
```suggestion
			require.NoError(t, err)
```

dm/pkg/binlog/event/util_test.go
```diff
@@ -123,13 +126,50 @@ func (t *testUtilSuite) TestStatusVarsToKV(c *C) {
...
+		}
+		require.Equal(t, vars, test.output)
+
```
lance6716, line 137:
```suggestion
```

dm/pkg/binlog/event/util_test.go
```diff
@@ -123,13 +126,50 @@ func (t *testUtilSuite) TestStatusVarsToKV(c *C) {
...
+	for _, test := range testCases {
+		var upstreamTZStr string // to simulate Syncer.upstreamTZ
+		upstreamTZStr = "+0:00"
+		vars, err := GetTimezoneByStatusVars(test.input, upstreamTZStr)
+		if test.err != nil {
+			require.Equal(t, test.err.Error(), err.Error())
 		} else {
-			c.Assert(err, IsNil)
+			require.Nil(t, err)
```
lance6716, line 171:
```suggestion
			require.NoError(t, err)
```

dm/pkg/binlog/event/util_test.go
```diff
@@ -123,13 +126,50 @@ func (t *testUtilSuite) TestStatusVarsToKV(c *C) {
...
 		}
-		c.Assert(vars, DeepEquals, t.output)
+		require.Equal(t, vars, test.output)
```
lance6716, line 173:
```suggestion
		require.Equal(t, test.output, vars)
```

The first is expected and second is got. Please check parameter names.

dm/pkg/binlog/event/util.go
lyzx2001: Understand. Added the comment.

lance6716 (approved):

lance6716:
ptal @D3Hunter

D3Hunter (approved):

PR #6169, opened 2022-07-04
master(dm): DM didn't show error reason when meet SQL error

### LINKED ISSUE #6032: DM didn't show error reason when meet SQL error
### What did you do?

https://asktug.com/t/topic/694378/12?u=lance6716

an error message, maybe the table name is too long.

### What did you see instead?

_No response_

### Versions of the cluster

v6.0.0

```console
```

ForwardStar:

Issue Number: close #6032 

Getting the task status with OpenAPI may result in only an error message without error code, raw cause, etc. In this pull request I wrap the information together into the field "ErrorMsg".

#### Tests 

 - Unit test

### Release Note
```release-note
None
```


lance6716 (approved):

dm/openapi/fixtures/task.go
```diff
@@ -53,8 +53,45 @@ var (
...
-	  }
-`
+	}
+	`
+
+	noShardErrNameJSONStr = `
+	{
+		"enhance_online_schema_change": true,
+		"meta_schema": "dm_meta",
+		"name": "orders_netpay_pay_inf_nuccinterconn_zh_ext_2022_[6-12]",
```
dsdashun, line 63: Do we need to change the task name into a more generic name, and not hard-coded ?  For example, put this variable inside `GenNoShardErrNameOpenAPITaskForTest` as a string template, and generate the task name in a random sequence of letters with specific length, and generate the final task.

dm/dm/master/openapi_controller_test.go
```diff
@@ -445,6 +447,105 @@ func (s *OpenAPIControllerSuite) TestTaskController() {
...
+				Worker: worker1Name,
+			},
+			SubTaskStatus: [](*pb.SubTaskStatus){&pb.SubTaskStatus{
+				Result: &pb.ProcessResult{
+					Errors: [](*pb.ProcessError){&pb.ProcessError{
+						ErrCode:  10006,
+						ErrClass: "database",
+						ErrScope: "downstream",
+						ErrLevel: "high",
+						Message:  "fail to initialize unit Load of subtask orders_netpay_pay_inf_nuccinterconn_zh_ext_2022_[6-12] : execute statement failed: CREATE TABLE IF NOT EXISTS `dm_meta`.`orders_netpay_pay_inf_nuccinterconn_zh_ext_2022_[6-12]_lightning_checkpoint_list` (\n\t\ttask_name varchar(255) NOT NULL,\n\t\tsource_name varchar(255) NOT NULL,\n\t\tstatus varchar(10) NOT NULL DEFAULT 'init' COMMENT 'init,running,finished',\n\t\tPRIMARY KEY (task_name, source_name)\n\t);\n",
```
dsdashun, line 489: I suggest to use the task name as place holder, and fill in with `task.Name` instead.

dm/dm/master/openapi_controller_test.go
```diff
@@ -445,6 +447,105 @@ func (s *OpenAPIControllerSuite) TestTaskController() {
...
+						RawCause: "Error 1059: Identifier name 'orders_netpay_pay_inf_nuccinterconn_zh_ext_2022_[6-12]_lightning_checkpoint_list' is too long.",
```
dsdashun, line 490: ditto

dm/openapi/fixtures/task.go
ForwardStar: I think it is difficult to put the name in this multi-lined string. But I think we can use the default NoShard task and manually assign it into task.Name

dsdashun (approved):
LGTM

PR #6183, opened 2022-07-05
validator(dm): add new it & fix some tips in dmctl

buchuitoudegou:

Issue Number: ref #4426 

#### Tests 

 - Integration test

### Release note 

```release-note
`None`.
```


buchuitoudegou:
docs/swagger is automatically updated when I run `make check`. PTAL @Ehco1996 if they should be changed.

dm/dm/ctl/master/start_stop_validation.go
```diff
@@ -39,7 +39,7 @@ func NewStartValidationCmd() *cobra.Command {
 		RunE:  startValidation,
 	}
 	cmd.Flags().Bool("all-task", false, "whether applied to all tasks")
-	cmd.Flags().String("mode", "", "specify the mode of validation: full, fast")
+	cmd.Flags().String("mode", "", "specify the mode of validation: full (default), fast")
```
lance6716, line 42:
```suggestion
	cmd.Flags().String("mode", "full", "specify the mode of validation: full (default), fast")
```

why the default value is not set here 🤔

lance6716 (approved):

dm/dm/ctl/master/start_stop_validation.go
buchuitoudegou: It's set in the server (I didn't notice this either until I found the implicit logic in the server code😂)

https://github.com/pingcap/tiflow/blob/83a7cd9a1879e73e4077511f209b8c247da3c76e/dm/dm/master/server.go#L2748-L2751

But maybe the tips should be revised too, because the `full` mode is by default only when the user has never enabled the validator while, on the contrary, the `mode` cannot be reset by `validation start --mode=xxx`, in the case of which `mode` is ignored.

D3Hunter: seems we can set default as @lance6716 suggested, since we already checked whether there's explicit mode or not.

dm/tests/validator_basic/conf/task.yaml
```diff
@@ -0,0 +1,45 @@
+---
```
D3Hunter, line 1: maybe name this file as `task-sharding`

docs/swagger/docs.go
```diff
@@ -796,6 +796,9 @@ var doc = `{
                 },
                 "schema-registry": {
                     "type": "string"
+                },
```
D3Hunter, line 799: cames from `make check`?

seems we can ignore it or file another pr

dm/tests/validator_basic/run.sh
```diff
@@ -685,12 +685,42 @@ function test_validation_syncer_stopped() {
 		"new\/ignored\/resolved: 0\/0\/0" 1
 }
 
+function test_dup_autopk() {
+	# successfully validate sharding tables
+	# though the downstream table deletes the primary key
+	# because unique key can identify a row accurately.
```
D3Hunter, line 691:
```suggestion
	# because not-null-unique key can identify a row accurately.
```

D3Hunter (commented):
rest lgtm

dm/dm/ctl/master/start_stop_validation.go
buchuitoudegou: Will fix it.

dm/tests/validator_basic/run.sh
buchuitoudegou: d28f43f

docs/swagger/docs.go
buchuitoudegou: d28f43f

dm/tests/validator_basic/conf/task.yaml
buchuitoudegou: d28f43f

dm/dm/ctl/master/start_stop_validation.go
buchuitoudegou: d28f43f

D3Hunter (approved):

PR #6188, opened 2022-07-05
syncer(dm): solve the issue of nondeterministic DDLs may cause data inconsistency

### LINKED ISSUE #5066: Nondeterministic DDLs may cause more data inconsistency

lyzx2001:

Issue Number: close #5066 

To solve the issue that nondeterministic DDLs may cause data inconsistency, I retrieve the timestamp information from the header of the binlog event, and retrieve the timezone information using the function GetTimezoneByStatusVars(). Then I pass these information to function syncDDL() by adding timestamp and timezone attributes in the job struct and queryEventContext struct. I put "SET TIME_ZONE/TIMESTAMP = retrieved_timezone/retrieved_timestamp" at the start of all the SQL commands, and put "SET TIME_ZONE/TIMESTAMP = DEFAULT" at the end of all the SQL commands. Now when downstream is synchronizing commands like "ALTER TABLE ADD COLUMN DATETIME/TIMESTAMP", it will generate the same data as upstream's.

For the integration test, I tested "ALTER TABLE ADD COLUMN DATETIME" and "ALTER TABLE ADD COLUMN TIMESTAMP" commands on two different tables. After these two commands, I let the program sleep for 1 second to manually cause inconsistency in the actual timestamp of upstream and downstream. Then I use check_sync_diff to check whether the data are consistent. After this, use "INSERT INTO TABLE" commands to check whether timezone and timestamp have been set back to default. 

#### Tests 

 - Unit test
 - Integration test
 

I have added timestamp and timezone attributes in the job struct and queryEventContext struct, maybe it will affect other related functions. The returned parameters of the function str2TimezoneOrFromDB() in dm/syncer/util.go are modified, a parameter called tzStr with type string is added, to store the original string of upstream's time zone. I have modified the related functions that call str2TimezoneOrFromDB().
Need to update design documentation that two attributes timestamp and timezone have been added to the job struct in dm/syncer/job.go and the queryEventContext struct in dm/syncer/syncer.go. Also, the returned parameters of the function str2TimezoneOrFromDB() in dm/syncer/util.go are modified, a parameter called tzStr with type string is added, to store the original string of upstream's time zone.

### Release note 

```release-note
Fix the issue of encountering data inconsistency when using nondeterministic DDLs like "ALTER TABLE ADD COLUMN x DATETIME/TIMESTAMP DEFAULT CURRENT_TIMESTAMP".
Notice: This fix only works for the non-sharding cases.
```


dm/syncer/syncer.go
```diff
@@ -1321,6 +1321,20 @@ func (s *Syncer) syncDDL(queueBucket string, db *dbconn.DBConn, ddlJobChan chan
 		if !ok {
 			return
 		}
+
+		// set timezone
+		if ddlJob.timezone != "" {
+			setTimezoneSQL := fmt.Sprintf("SET SESSION TIME_ZONE = '%s'", ddlJob.timezone)
+			ddlJob.ddls = append([]string{setTimezoneSQL}, ddlJob.ddls...)
+			setTimezoneSQLDefault := fmt.Sprintf("SET SESSION TIME_ZONE = DEFAULT")
```
lance6716, line 1329: maybe we should not revert it to default. For two sequential DDL, if the latter one uses the same timezone as the former one, binlog will not write timezone again. So for latter one we should keep the timezone

Please check mysqlbinlog's output to see if it's right.

dm/syncer/syncer.go
```diff
@@ -2795,6 +2811,13 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
 		s.tctx.L().Warn("found error when get sql_mode from binlog status_vars", zap.Error(err))
 	}
 
+	qec.timezone, err = event.GetTimezoneByStatusVars(ev.StatusVars, s.upstreamTZ.String())
+	if err != nil {
+		s.tctx.L().Warn("found error when get timezone from binlog status_vars", zap.Error(err))
+	}
+
+	qec.timestamp = e.Header.Timestamp
```
lance6716, line 2819: e.Header should be contained in ec.header, no need to add a parameter

dm/syncer/syncer.go
```diff
@@ -2795,6 +2811,13 @@ func (s *Syncer) handleQueryEvent(ev *replication.QueryEvent, ec eventContext, o
 		s.tctx.L().Warn("found error when get sql_mode from binlog status_vars", zap.Error(err))
 	}
 
+	qec.timezone, err = event.GetTimezoneByStatusVars(ev.StatusVars, s.upstreamTZ.String())
```
lance6716, line 2814: I'm a bit afriad that the string converted from s.upstreamTZ is not the same as the original string which builds s.upstreamTZ. So maybe we need to add s.upstreamTZStr to store the original string.

Can you find some material to check it?

dm/syncer/syncer.go
lyzx2001: Yes. If the time zone is the same as that of last time, the time zone information will not be found in binlog.

dm/syncer/syncer.go
lyzx2001: Added s.upstreamTZStr.

dm/syncer/syncer.go
```diff
@@ -1321,6 +1323,27 @@ func (s *Syncer) syncDDL(queueBucket string, db *dbconn.DBConn, ddlJobChan chan
...
 		}
+
+		// set timezone
+		if ddlJob.timezone != "" {
+			s.timezone_last_time = ddlJob.timezone
+			setTimezoneSQL := fmt.Sprintf("SET SESSION TIME_ZONE = '%s'", ddlJob.timezone)
+			ddlJob.ddls = append([]string{setTimezoneSQL}, ddlJob.ddls...)
+			setTimezoneSQLDefault := fmt.Sprintf("SET SESSION TIME_ZONE = DEFAULT")
+			ddlJob.ddls = append(ddlJob.ddls, setTimezoneSQLDefault)
+		} else {
```
D3Hunter, line 1334: > For two sequential DDL, if the latter one uses the same timezone as the former one, binlog will not write timezone again.

should check `s.timezone_last_time` in case we start task in the middle of 2 ddl, one has timezone, one not

dm/tests/many_tables/run.sh
```diff
@@ -25,12 +27,24 @@ function incremental_data() {
 			run_sql "INSERT INTO many_tables_db.t$i VALUES ($j,${j}000$j),($j,${j}001$j);" $MYSQL_PORT1 $MYSQL_PASSWORD1
 		done
 	done
+
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"pause-task test" \
+		"\"result\": true" 2
+
+	run_sql "ALTER TABLE many_tables_db.t1 ADD x datetime DEFAULT current_timestamp;" $MYSQL_PORT1 $MYSQL_PASSWORD1
```
D3Hunter, line 35: not sure whether up/down stream uses different tz, if not, can refer this https://github.com/pingcap/tiflow/blob/9f92cb6ca9298b5b6c212e89754edb79b180847d/dm/tests/start_task/run.sh#L71-L77

dm/syncer/syncer.go
lyzx2001: Yes. I have modified it as it will use s.timezone_last_time if current timezone is "".

D3Hunter: `s.timezone_last_time` might be empty in this case, should check it

D3Hunter (approved):
this fix only works for non-sharding case, maybe add it to `release note` part in desc

lyzx2001:
> this fix only works for non-sharding case, maybe add it to `release note` part in desc

Done.

dm/tests/many_tables/run.sh
```diff
@@ -120,6 +153,8 @@ cleanup_process
 
 run $*
 
+trap - EXIT
```
lance6716, line 156: before we reset the trap, we should manually run `restore_timezone` so it will be run once at any case

Issue #6152, opened 2022-07-01
TiCDC should give some hints when user pass `::1` to SinkURI


hi-rustin:
### Is your feature request related to a problem?

cc: https://github.com/pingcap/tiflow/issues/6135

### Describe the feature you'd like

If the user forgets to add `[]` when passing the ipv6 address, we should report an error and prompt the user to use `[::1]`.
See https://www.ietf.org/rfc/rfc2732.txt, we have to use brackets to wrap IPv6 address.

PR #6190, opened 2022-07-05
sink(ticdc): precheck the sink URI and give some hints

hi-rustin:

Issue Number: close https://github.com/pingcap/tiflow/issues/6152

- add precheck for sinkURI
- when we use wrong IPv6 format, give some hints

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```


hi-rustin:
Manual testing:
```log
1. ./cdc cli changefeed create --pd="http://127.0.0.1:2379" --sink-uri="mysql://root@::1]:3306/" --changefeed-id="simple-replication-task1" --sort-engine="unifaied"
2. [CDC:ErrSinkURIInvalid]sink uri host is not valid IPv6 address, when using IPv6 address in URI, please use [ipv6-address]:port
```

cdc/sink/validator.go
```diff
@@ -0,0 +1,84 @@
...
+// Maybe we should support the dry-run mode to validate sink.
+func Validate(ctx context.Context, sinkURI string, cfg *config.ReplicaConfig) error {
+	if err := preCheckSinkURI(sinkURI); err != nil {
+		return err
+	}
+
+	errCh := make(chan error)
+	ctx, cancel := context.WithCancel(contextutil.PutRoleInCtx(ctx, util.RoleClient))
+	s, err := New(ctx, model.DefaultChangeFeedID("sink-verify"),
+		sinkURI, cfg, errCh)
```
overvenus, line 38:
```suggestion
	s, err := New(ctx, model.DefaultChangeFeedID("sink-verify"), sinkURI, cfg, errCh)
```

pkg/util/uri.go
```diff
@@ -0,0 +1,62 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import "strings"
+
+// IsValidIPv6AddressInURI reports whether hostPort is a valid IPv6 address in URI.
+// See: https://www.ietf.org/rfc/rfc2732.txt.
+func IsValidIPv6AddressInURI(hostPort string) bool {
```
overvenus, line 20: What about [`net.ParseIP`](https://pkg.go.dev/net#ParseIP)?

overvenus (commented):
Rest LGTM

pkg/util/uri.go
hi-rustin: The main purpose of this function is to determine the format in the URI, not whether it is an IP.

cdc/sink/validator.go
hi-rustin: Done. Thanks!

pkg/util/uri.go
hi-rustin: Added in IsIPv6Address.

overvenus (approved):

nongfushanquan (approved):

Issue #6022, opened 2022-06-24
engine package related enhancement


amyangfei:
### Is your feature request related to a problem?

Some package names and layouts can be optimized.

### Describe the feature you'd like

- [ ] Move engine binary to root path of cmd, rename to `tiflow` command, and makes `df-master` and `df-executor` as subcommand of `tiflow`. ~The command will be either `tiflow master` or `tiflow engine master`~. The command will be `tiflow master`, `tiflow executor` and `tiflow cli`.
  - [x] 1. Move df-executor, df-master to tiflow executor and tiflow master #6111
  - [ ] 2. refactor server master config and executor config
  - [ ] 3. add more unit tests
  - [x] 4. migrate df-master-client to tiflow cli #6181
  - [x] 5. rename other binaries with `df` prefix #6200
- [x] Rename utilities library to singular form, such as renaming `github.com/pingcap/tiflow/engine/etcdutils` to `github.com/pingcap/tiflow/engine/etcdutil` #6030

PR #6200, opened 2022-07-06
*(engine): rename remaining df prefix to tiflow

amyangfei:

Issue Number: ref #6022

- rename remaining df prefix to tiflow

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


Makefile
```diff
@@ -2,7 +2,7 @@
 .PHONY: build test check clean fmt cdc kafka_consumer coverage \
 	integration_test_build integration_test integration_test_mysql integration_test_kafka bank \
 	dm dm-master dm-worker dmctl dm-syncer dm_coverage \
-	engine tiflow df-demo df-chaos-case
+	engine tiflow tiflow-demo tiflow-chaos-case
```
CharlesCheung96, line 5: Is sub-command more appropriate here? Such as `tiflow demo` and `tiflow chaos`.

amyangfei: I prefer to use standalone binary, for two reasons
- demo and chaos test will not be used in product environment
- It will make binary too large to put them together.

amyangfei:
ptal @CharlesCheung96

CharlesCheung96 (approved):

Issue #6208, opened 2022-07-07
[Engine] Improvements on communication mechanisms


liuzix:
- [x] Refactor Executor clients https://github.com/pingcap/tiflow/pull/6209
- [x] Refactor master clients https://github.com/pingcap/tiflow/pull/6436
- [x] Unify error handling style of internal gRPC services https://github.com/pingcap/tiflow/pull/6788
- [ ] Improve the quality of `pkg/p2p`.
- [ ] Tracing for Open API and gRPC methods

PR #6209, opened 2022-07-07
client(engine): Refactor the executor client

liuzix:
Issue Number: ref #6208

- Rewrote ExecutorClient-related logic to (1) Provide a better decoupled interface (2) Improve the overall code quality and test coverage.

#### Tests 

 - Unit test

No.

No.

### Release note 

```release-note
None.
```


engine/pkg/client/executor_client.go
```diff
@@ -0,0 +1,106 @@
...
+func NewExecutorClient(conn *grpc.ClientConn) ExecutorClient {
+	return &executorClientImpl{
+		ExecutorServiceClient: NewExecutorServiceClient(enginepb.NewExecutorClient(conn)),
+		BrokerServiceClient:   NewBrokerServiceClient(enginepb.NewBrokerServiceClient(conn)),
+		conn:                  conn,
+	}
+}
+
+// Close closes the gRPC connection maintained by the client.
+func (c *executorClientImpl) Close() {
```
lonng, line 58: Why not propagate the error to its caller?

engine/pkg/client/executor_group.go
```diff
@@ -0,0 +1,209 @@
...
+		logger:        logger,
+	}
+}
+
+// GetExecutorClient tries to get the ExecutorClient for the given executor.
+func (g *DefaultExecutorGroup) GetExecutorClient(id model.ExecutorID) (ExecutorClient, bool) {
+	g.mu.RLock()
+	defer g.mu.RUnlock()
+
+	if client, ok := g.clients[id]; ok {
```
lonng, line 80: Why not use `return g.clients[id]`?

liuzix: Go does not support this syntax. Refer to https://stackoverflow.com/questions/49921496/passing-on-return-from-a-map-get

engine/pkg/client/executor_group.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return ret, nil
+}
+
+// UpdateExecutorList updates the stored clients using a map from executor IDs
+// to their addresses.
+// Note: This method will not wait for the clients to be fully connected.
+// In the rare case where grpc.Dial does fail, an error will be returned to the caller
+// of this method, and the caller should retry appropriately.
+func (g *DefaultExecutorGroup) UpdateExecutorList(executors map[model.ExecutorID]string) error {
```
amyangfei, line 115: Where will this API be used?

liuzix: There are no existing use cases for now, but I'm imagining that this can be useful for some optimizations such as periodically re-fetching the list of executors, etc.

engine/pkg/client/executor_client.go
liuzix: - There is probably no easy remedy for any error that could be returned here.
- If we have an alert system, we can report the error so that we can discover a problem if errors occur frequently.

gozssky (approved):

Issue #6160, opened 2022-07-04
TiCDC supprots filterling DDL/DML event by event type and SQL expression.


asddongmen:
### Is your feature request related to a problem?

TiCDC already supports filtering DDL/DML event by schema name and table name, but this capability is no longer sufficient to meet the needs of users.

### Describe the feature you'd like

TiCDC should support:
1.  Filter out DDL event by its event type.
2. Filter out DML event by its event type.
3. Filter out DML event by SQL expression.

### Teachability, Documentation, Adoption, Migration Strategy
- [x] move filter from sink to mounter.  https://github.com/pingcap/tiflow/pull/5936
- [x] Add dml expression filter. https://github.com/pingcap/tiflow/pull/6215
- [x] Add dml/ddl event filter type. https://github.com/pingcap/tidb-tools/pull/655
- [x] Add changefeed not retry error. #6310 
- [ ] Integrate row filter function into current filter logic.  https://github.com/pingcap/tiflow/pull/6095

_No response_

PR #6215, opened 2022-07-07
filter (ticdc): add dml expression filter

asddongmen:

Issue Number: ref #6160

Add strcut `dmlExprFilter` to `pkg/filter`, it is a part of cdc row filter.
Note: This PR will not change any behavior of cdc, it only:
1. Add strcut `dmlExprFilter`.
2. Add a filter config struct `EventFilterRule` to `pkg/config/filter.go`.
3. Add a field `RowChangedDatums` to `model.RowChangedEvent` to facilitate `dmlExprFilter`'s function.

#### Tests 

 - Unit test

### Release note 

```release-note

None
```


cdc/model/sink.go
```diff
@@ -269,6 +270,15 @@ type RowChangedEvent struct {
 
 	// SplitTxn marks this RowChangedEvent as the first line of a new txn.
 	SplitTxn bool `json:"-" msg:"-"`
+	// This filed is only used by expression filter.
+	// It will be set to nil after used.
+	RowChangedDatums *RowChangedDatums `json:"-" msg:"-"`
```
overvenus, line 275:
```suggestion
	RowChangedDatums RowChangedDatums `json:"-" msg:"-"`
```

Prefer value to ease GC pressure.

pkg/filter/expr_filter.go
```diff
@@ -0,0 +1,374 @@
...
+		if err != nil {
+			return nil, err
+		}
+		r.deleteExprs[tableName] = expr
+	}
+	return r.deleteExprs[tableName], nil
+}
+
+func (r *dmlExprFilterRule) getSimpleExprOfTable(expr string,
+	ti *model.TableInfo,
```
overvenus, line 199:
```suggestion
func (r *dmlExprFilterRule) getSimpleExprOfTable(
	expr string, ti *model.TableInfo,
```

pkg/filter/expr_filter.go
```diff
@@ -0,0 +1,374 @@
...
+			return nil, cerror.ErrExpressionColumnNotFound.
+				FastGenByArgs(getColumnFromError(err), ti.TableName.Table, expr)
+		}
+		log.Error("failed to parse expression", zap.Error(err))
+		return nil, cerror.ErrExpressionParseFailed.FastGenByArgs(err, expr)
+	}
+	return e, nil
+}
+
+func (r *dmlExprFilterRule) shouldSkipDML(row *model.RowChangedEvent,
```
overvenus, line 217: Do it too.

pkg/filter/expr_filter.go
```diff
@@ -0,0 +1,374 @@
...
+	for _, rule := range f.rules {
+		if rule.tableMatcher.MatchTable(schema, table) {
+			res = append(res, rule)
+		}
+	}
+	return res
+}
+
+// shouldSkipDML skips dml event by sql expression.
+func (f *dmlExprFilter) shouldSkipDML(row *model.RowChangedEvent,
```
overvenus, line 356: Could you add a benchmark? Since it applies to all changes, it may become a bottleneck.

cdc/model/sink.go
```diff
@@ -269,6 +270,15 @@ type RowChangedEvent struct {
...
+}
+
+// RowChangedDatums is used to store the changed datums of a row.
+type RowChangedDatums struct {
+	RowDatums    []types.Datum `json:"-" msg:"-"`
+	PreRowDatums []types.Datum `json:"-" msg:"-"`
```
overvenus, line 281: Please add them to ApproximateBytes().

pkg/config/filter.go
```diff
@@ -24,4 +25,18 @@ type FilterConfig struct {
 	*filter.MySQLReplicationRules
 	IgnoreTxnStartTs []uint64           `toml:"ignore-txn-start-ts" json:"ignore-txn-start-ts"`
 	DDLAllowlist     []model.ActionType `toml:"ddl-allow-list" json:"ddl-allow-list,omitempty"`
+	EventFilters     []*EventFilterRule `toml:"event-filters" json:"event-filters"`
+}
+
+// EventFilterRule is used by sql event filter and expression filter
+type EventFilterRule struct {
```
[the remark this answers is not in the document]
asddongmen: ok

cdc/model/sink.go
asddongmen: This struct's life cycle limits in mounter and will be set to nil after here :https://github.com/pingcap/tiflow/blob/master/cdc/entry/mounter.go#L122, so it is no need to add them to `ApproximateBytes()`.

cdc/model/sink.go
asddongmen: We will assign a value to this filed and set it to nil in a same function scope, so I think it is acceptable use a pointer here.

pkg/filter/expr_filter.go
asddongmen: yes.

sdojjy (approved):

pkg/filter/expr_filter.go
asddongmen: // cmd: go test -benchmem -run=^$ -bench ^BenchmarkSkipDML$ github.com/pingcap/tiflow/pkg/filter
// goos: maxOS 12.3.1
// goarch: arm64
// cpu: Apple M1 Pro
// BenchmarkSkipDML/insert-1
// BenchmarkSkipDML/insert-1-10              990166              1151 ns/op             768 B/op         24 allocs/op
// BenchmarkSkipDML/insert-2
// BenchmarkSkipDML/insert-2-10             1000000              1187 ns/op             768 B/op         24 allocs/op
// BenchmarkSkipDML/update
// BenchmarkSkipDML/update-10                698208              1637 ns/op            1480 B/op         43 allocs/op
// BenchmarkSkipDML/delete
// BenchmarkSkipDML/delete-10               1000000              1112 ns/op             768 B/op         24 allocs/op

cdc/model/mounter.go
```diff
@@ -17,9 +17,21 @@ import (
 	"math"
 
 	"github.com/pingcap/log"
+	"github.com/pingcap/tidb/types"
 	"go.uber.org/zap"
 )
 
+// RowChangedDatums is used to store the changed datums of a row.
+type RowChangedDatums struct {
```
overvenus, line 25: How about define it in cdc/entry/mounter.go?

asddongmen: It will cause cycle import.

pkg/filter/expr_filter.go
```diff
@@ -0,0 +1,398 @@
...
+	"github.com/pingcap/tiflow/dm/pkg/utils"
+	"github.com/pingcap/tiflow/pkg/config"
+	cerror "github.com/pingcap/tiflow/pkg/errors"
+	"go.uber.org/zap"
+)
+
+// dmlExprFilterRule only be used by dmlExprFilter.
+// This struct is mostly a duplicate of `ExprFilterGroup` in dm/pkg/syncer,
+// but have slightly changed to fit the usage of cdc.
+type dmlExprFilterRule struct {
```
overvenus, line 38: Does it support GBK encoding?

overvenus (approved):

pkg/filter/expr_filter.go
asddongmen: Yes, I have added a unit test case to cover GBK charset.

overvenus (approved):

PR #6216, opened 2022-07-07
engine(DM): implement dm openapi

GMHDBJD:

Issue Number: ref #5504 
- implement openapi for dm
- remove debug job command

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// DMAPIGetJobConfig implements the api of get job config.
+func (jm *JobMaster) DMAPIGetJobConfig(c *gin.Context) {
+	cfg, err := jm.GetJobCfg(c.Request.Context())
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+	bs, err := cfg.Yaml()
```
lance6716, line 71: If the output is directly passed to Web frontend, I guess JSON is more common to be used. Will discuss later

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,251 @@
...
+// DMAPISetBinlogOperator implements the api of set binlog operator.
+func (jm *JobMaster) DMAPISetBinlogOperator(c *gin.Context, taskName string) {
+	var req openapi.SetBinlogOperatorRequest
+	if err := c.Bind(&req); err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	r := &dmpkg.BinlogRequest{
+		Sources: []string{taskName},
```
lance6716, line 141: seems the field should be `Task`, but in e2e_dm_test.go we use `[]string{source1, source2}` as tasks. Very confusing

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	switch req.Op {
+	case openapi.SetBinlogOperatorRequestOpInject:
+		r.Op = pb.ErrorOp_Inject
+	case openapi.SetBinlogOperatorRequestOpSkip:
+		r.Op = pb.ErrorOp_Skip
+	case openapi.SetBinlogOperatorRequestOpReplace:
+		r.Op = pb.ErrorOp_Replace
+	default:
+		_ = c.Error(errors.New("unsupport op type '' for set binlog operator"))
```
lance6716, line 157:
```suggestion
		_ = c.Error(errors.New("unsupported op type '' for set binlog operator"))
```

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	if params.BinlogPos != nil {
+		req.BinlogPos = *params.BinlogPos
+	}
+	resp, err := jm.Binlog(c.Request.Context(), req)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+	// TODO: use no content response, now return content because delete may failed.
```
lance6716, line 182: Not sure if we should return a 500 errors when resp.ErrorMsg is not empty

engine/test/e2e/e2e_dm_test.go
```diff
@@ -177,154 +185,269 @@ func testSimpleAllModeTask(
...
+
+func queryStatus(ctx context.Context, client *http.Client, jobID string, tasks []string, t *testing.T) (*dm.JobStatus, error) {
+	url := fmt.Sprintf(baseURL+"/status", jobID)
+	for i, task := range tasks {
+		if i == 0 {
+			url += "?"
+		} else {
+			url += "&"
+		}
+		url += "tasks=" + task
```
lance6716, line 317: maybe https://pkg.go.dev/net/url#Values.Encode because it also handle special characters encoding

lance6716 (commented):
rest lgtm

amyangfei:
PTAL @gozssky @maxshuang

engine/jobmaster/dm/openapi.go
GMHDBJD: In engine we use tasks, in DM we use source...

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,251 @@
...
+
+	router.PUT("/config", wrapper.DMAPIUpdateJobConfig)
+
+	router.GET("/schema/tasks/:task-name", wrapper.DMAPIGetSchema)
+
+	router.PUT("/schema/tasks/:task-name", wrapper.DMAPISetSchema)
+
+	router.GET("/status", wrapper.DMAPIGetJobStatus)
+
+	router.PUT("/status", wrapper.DMAPIOperateJob)
```
lance6716, line 47: should use POST according to out OpenAPI doc

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,251 @@
...
+	var tasks []string
+	if params.Tasks != nil {
+		tasks = *params.Tasks
+	}
+	resp, err := jm.QueryJobStatus(c.Request.Context(), tasks)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+	c.IndentedJSON(http.StatusOK, resp)
```
lance6716, line 61: should return 500 when status has error

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,251 @@
...
+	default:
+		_ = c.Error(errors.New("unsupported op type '' for set binlog operator"))
+		return
+	}
+	resp, err := jm.Binlog(c.Request.Context(), r)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+	c.IndentedJSON(http.StatusOK, resp)
```
lance6716, line 165:
```suggestion
	c.IndentedJSON(http.StatusCreated, resp)
```

for POST

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,251 @@
...
+	if params.BinlogPos != nil {
+		req.BinlogPos = *params.BinlogPos
+	}
+	resp, err := jm.Binlog(c.Request.Context(), req)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+	// TODO: use no content response, now return content because delete may failed.
+	c.IndentedJSON(http.StatusOK, resp)
```
lance6716, line 183: for successful deletion, status code is 204

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,251 @@
...
+		return
+	}
+
+	r := &dmpkg.BinlogSchemaRequest{
+		Op:       op,
+		Sources:  []string{taskname},
+		Database: database,
+		Table:    table,
+	}
+	resp := jm.BinlogSchema(c.Request.Context(), r)
```
lance6716, line 219: don't forget to check error message

engine/jobmaster/dm/openapi.go
lance6716: JSON is almost the only accepted format in our doc

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,251 @@
...
+	err := jm.OperateTask(c.Request.Context(), op, nil, tasks)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+	c.Status(http.StatusOK)
+}
+
+// DMAPIGetBinlogOperator implements the api of get binlog operator.
+func (jm *JobMaster) DMAPIGetBinlogOperator(c *gin.Context, taskName string, params openapi.DMAPIGetBinlogOperatorParams) {
```
lance6716, line 116: hope there's not many operators so we don't need to add a `total` field and paging

lance6716 (commented):
there's an unified error structure for "API error structure" we can do it in next PR 😂

engine/jobmaster/dm/openapi.go
GMHDBJD: It's not create status, it's change status, so I think PUT is more accurate according to the doc.

engine/jobmaster/dm/openapi.go
GMHDBJD: add a todo

engine/jobmaster/dm/openapi.go
GMHDBJD: add a todo in terrorHTTPErrorHandler

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,300 @@
...
+	switch req.Op {
+	case openapi.SetBinlogOperatorRequestOpInject:
+		r.Op = pb.ErrorOp_Inject
+	case openapi.SetBinlogOperatorRequestOpSkip:
+		r.Op = pb.ErrorOp_Skip
+	case openapi.SetBinlogOperatorRequestOpReplace:
+		r.Op = pb.ErrorOp_Replace
+	default:
+		// nolint:errcheck
+		_ = c.Error(errors.New("unsupported op type '' for set binlog operator"))
```
amyangfei, line 192: Can add `req.Op` in error.

tests/utils/many_sorters_test/many_sorters.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/pingcap/tiflow/cdc/sorter"
 	"github.com/pingcap/tiflow/cdc/sorter/unified"
 	"github.com/pingcap/tiflow/pkg/config"
+	cerrors "github.com/pingcap/tiflow/pkg/errors"
```
amyangfei, line 33: Is this file(test caes) dead code, it is not reasonable to exist missing import. 🤔

amyangfei (approved):

engine/jobmaster/dm/openapi.go
```diff
@@ -0,0 +1,300 @@
...
+// TODO: extract terror from worker response
+func terrorHTTPErrorHandler() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		c.Next()
+		gErr := c.Errors.Last()
+		if gErr == nil {
+			return
+		}
+		var (
+			code int
```
lance6716, line 77: there should be a default error code, we can do it in next PR. And this function should be put before DMAPIGetJobStatus

engine/jobmaster/dm/openapi/spec/dm.yaml
```diff
@@ -0,0 +1,330 @@
...
+          schema:
+            type: string
+      requestBody:
+        description: "request body"
+        content:
+          "application/json":
+            schema:
+              $ref: "#/components/schemas/SetBinlogOperatorRequest"
+      responses:
+        "200":
```
lance6716, line 216: not sure what's the effect of HTTP status code in this file. post should be 201

lance6716: rest lgtm

engine/jobmaster/dm/openapi.go
GMHDBJD: add another TODO... Now I think 0 is ok.

lance6716 (approved):

PR #6217, opened 2022-07-07
syncer(dm): batch processing error when too many tables

### LINKED ISSUE #6161: the 'txn-entry-size-limit' config is not effective in DM
### What did you do?

https://asktug.com/t/topic/694876

TiDB config:
```
+------+---------------------+--------------------------------------+-----------+
| Type | Instance            | Name                                 | Value     |
+------+---------------------+--------------------------------------+-----------+
| tidb | xxxxx:4000 | performance.txn-entry-size-limit     | 125829120 |
| tikv | xxxxx:20161 | raftstore.raft-entry-max-size        | 128MiB    |
+------+---------------------+--------------------------------------+-----------+
```
DM error:
```
"errors": [
      {
          "ErrCode": 50000,
          "ErrClass": "not-set",
          "ErrScope": "not-set",
          "ErrLevel": "high",
          "Message": "[kv:8025]entry too large, the max entry size is 6291456, the size of data is 7890542",
          "RawCause": "",
          "Workaround": ""
      }
],
```

DM task work normally.

### What did you see instead?

DM task get an error.

### Versions of the cluster

```
tiup is checking updates for component dmctl ...
Starting component `dmctl`: /home/admin/.tiup/components/dmctl/v6.1.0/dmctl/dmctl -V
Release Version: v6.1.0
Git Commit Hash: 9f5e3cea2fae0a3dce6cc5ca79e3bcd654b8aa99
Git Branch: heads/refs/tags/v6.1.0
UTC Build Time: 2022-06-07 11:55:24
Go Version: go version go1.18.2 linux/amd64
```
TiDB Version
```
MySQL [(none)]> SELECT tidb_version()\G
tidb_version(): Release Version: v6.1.0
Edition: Community
Git Commit Hash: https://github.com/pingcap/tidb/commit/1a89decdb192cbdce6a7b0020d71128bc964d30f
Git Branch: heads/refs/tags/v6.1.0
UTC Build Time: 2022-06-05 05:15:11
GoVersion: go1.18.2
Race Enabled: false
TiKV Min Version: v3.0.0-60965b006877ca7234adaced7890d7b029ed1306
1 row in set (0.00 sec)
```

```console
```

ForwardStar:

Issue Number: close #6161 

When processing batch creating tables, it may exceed tin-entry-size-limit and hence return an error. We solve this problem with divide and conquer techniques.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


dm/pkg/schema/tracker.go
```diff
@@ -493,6 +493,22 @@ func (tr *Tracker) CreateTableIfNotExists(table *filter.Table, ti *model.TableIn
 	return tr.dom.DDL().CreateTableWithInfo(tr.se, schemaName, ti, ddl.OnExistIgnore)
 }
 
+// SplitBatchCreateTableAndHandle will split the batch if it exceeds the kv entry size limit.
+func (tr *Tracker) SplitBatchCreateTableAndHandle(schema model.CIStr, info *[]*model.TableInfo, l int, r int) error {
+	if err := tr.dom.DDL().BatchCreateTableWithInfo(tr.se, schema, (*info)[l:r], ddl.OnExistIgnore); err != nil {
```
D3Hunter, line 498: what happens if len((*info)[l:r]) == 0?

prefer we create them in batch, such as 100 at a time

ForwardStar: I think this will only happen if txn-entry-size-limit is too small that even one create table job could not be handle. Maybe we can throw a new defined error "txn-entry-size-limit too small"? However, I think divide-and-conquer may behave faster than fixed-size batches.

D3Hunter: other error may happen, such as `data dir` of the embedded tidb is corrupted, might keep failing and dividing

D3Hunter: > However, I think divide-and-conquer may behave faster than fixed-size batches.

need bench test, if its improvement is trival compared to some fixed batch-size. perfer to use batch, it's simpler

dm/pkg/schema/tracker.go
```diff
@@ -509,7 +529,7 @@ func (tr *Tracker) BatchCreateTableIfNotExist(tablesToCreate map[string]map[stri
 			cloneTis = append(cloneTis, cloneTi)
 		}
 		schemaName := model.NewCIStr(schema)
-		if err := tr.dom.DDL().BatchCreateTableWithInfo(tr.se, schemaName, cloneTis, ddl.OnExistIgnore); err != nil {
+		if err := tr.SplitBatchCreateTableAndHandle(schemaName, cloneTis, 0, len(cloneTis)); err != nil {
```
lance6716, line 532: I'm a bit nervous about zero length cloneTis. Do we have a test for it?

lance6716 (approved):

D3Hunter (approved):

dm/pkg/schema/tracker.go
ForwardStar: Let me try. If error occurs I would just skip this operation.

ForwardStar: I check the codes and think there is no case with zero length cloneTis. For safety, I add a test in ``schema_test.go``.

PR #6225, opened 2022-07-08
syncer(dm): adjust safe mode duration when start/resume

### LINKED ISSUE #6224: DM: adjust safe mode duration when DM resume

We want to adjust safe mode duration when DM resume 

add a configuration adjust safe mode duration when DM resume

_No response_

_No response_

okJiang:

Issue Number: close #6224

add configuration to adjust safe mode duration

- set "0s", not step in safe mode. But if dm exists checkpoint, dm must enter safe-mode. So dm will report an **error**.
- set "1s" or "1h", step in safe mode one second or one hour
- do not set it, step in safe mode as same as internal default( now is 60s )

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
add configuration to adjust safe mode duration
```


GMHDBJD (changes requested):
/hold

okJiang:
/hold 

need understand safe mode more

dm/cmd/dm-syncer/config.go
```diff
@@ -168,6 +168,7 @@ func (c *commonConfig) newSubTaskConfig(args []string) (*config.SubTaskConfig, e
 	fs.StringVar(&cfg.LogRotate, "log-rotate", "day", "log file rotate type, hour/day")
 	fs.BoolVar(&cfg.EnableGTID, "enable-gtid", false, "enable gtid mode")
 	fs.BoolVar(&cfg.SafeMode, "safe-mode", false, "enable safe mode to make syncer reentrant")
+	fs.BoolVar(&cfg.DisableSafeModeResume, "disable-safe-mode-resume", false, "disable safe mode to resume ")
```
lonng, line 171: Why must this command-line argument be added?

okJiang: Sry, it was a mistake. What we need do is that just add it to task config file.

I will remove it.

lance6716:
we already have `safe-mode-duration`, can `safe-mode-duration = 0` be used for this scenario? @okJiang

okJiang:
> we already have `safe-mode-duration`, can `safe-mode-duration = 0` be used for this scenario? @okJiang

Yes，we can use it. @lance6716 
- set 0, not step in safe mode when resume
- set x, step in safe mode x second when resume
- set -1(default), step in safe mode as same as internal ( now is 5s )

dm/cmd/dm-syncer/config.go
```diff
@@ -262,8 +262,9 @@ type syncerConfig struct {
 	EnableGTID  bool `toml:"enable-gtid" json:"enable-gtid"`
 	AutoFixGTID bool `toml:"auto-fix-gtid" json:"auto-fix-gtid"`
 
-	SafeMode   bool   `toml:"safe-mode" json:"safe-mode"`
-	ConfigFile string `json:"config-file"`
+	SafeMode               bool   `toml:"safe-mode" json:"safe-mode"`
+	SafeModeResumeDuration string `toml:"safe-mode-resume-duration" json:"safe-mode-resume-duration"`
```
lance6716, line 266: we don't maintain dm-syncer now. But changing it is also no harm

dm/dm/config/task.go
```diff
@@ -324,19 +324,19 @@ type SyncerConfig struct {
 	AutoFixGTID bool `yaml:"auto-fix-gtid" toml:"auto-fix-gtid" json:"auto-fix-gtid"`
 	EnableGTID  bool `yaml:"enable-gtid" toml:"enable-gtid" json:"enable-gtid"`
 	// deprecated
-	DisableCausality bool `yaml:"disable-detect" toml:"disable-detect" json:"disable-detect"`
-	SafeMode         bool `yaml:"safe-mode" toml:"safe-mode" json:"safe-mode"`
+	DisableCausality       bool   `yaml:"disable-detect" toml:"disable-detect" json:"disable-detect"`
+	SafeMode               bool   `yaml:"safe-mode" toml:"safe-mode" json:"safe-mode"`
+	SafeModeResumeDuration string `yaml:"safe-mode-resume-duration" toml:"safe-mode-resume-duration" json:"safe-mode-resume-duration"`
```
lance6716, line 329: Can it be named `safe-mode-duration`? Or the `resume` means it only takes effect when resuming not starting?

dm/dm/config/task.go
```diff
@@ -324,19 +324,19 @@ type SyncerConfig struct {
...
 	EnableANSIQuotes bool `yaml:"enable-ansi-quotes" toml:"enable-ansi-quotes" json:"enable-ansi-quotes"`
 }
 
 // DefaultSyncerConfig return default syncer config for task.
 func DefaultSyncerConfig() SyncerConfig {
 	return SyncerConfig{
-		WorkerCount:             defaultWorkerCount,
-		Batch:                   defaultBatch,
-		QueueSize:               defaultQueueSize,
-		CheckpointFlushInterval: defaultCheckpointFlushInterval,
```
lance6716, line 339: I see it's adjusted again in SubTaskConfig.Adjust. Maybe we should also add an adjust in TaskConfig.adjust

dm/dm/config/task.go
```diff
@@ -1103,6 +1110,7 @@ func NewSyncerConfigsForDowngrade(syncerConfigs map[string]*SyncerConfig) map[st
 			EnableGTID:              syncerConfig.EnableGTID,
 			DisableCausality:        syncerConfig.DisableCausality,
 			SafeMode:                syncerConfig.SafeMode,
+			SafeModeResumeDuration:  syncerConfig.SafeModeResumeDuration,
```
lance6716, line 1113: This function is `NewSyncerConfigsForDowngrade` ! lower versions don't have `safe-mode-resume-duration`, so we should not write it.

dm/syncer/syncer.go
```diff
@@ -3740,6 +3740,18 @@ func (s *Syncer) Pause() {
...
+	var err error
+	defer func() {
+		if err != nil {
+			pr <- pb.ProcessResult{
+				IsCanceled: false,
+				Errors: []*pb.ProcessError{
+					unit.NewProcessError(err),
+				},
+			}
+			return
```
lance6716, line 3752: lint should report this return is redundant

dm/dm/config/task.go
okJiang: It seems `safe-mode-duration` is better to meet demand. LGTM.

Confirm later.

okJiang:
all test passed. ptal again~ @lance6716 @GMHDBJD

dm/dm/config/task.go
```diff
@@ -788,6 +792,19 @@ func (c *TaskConfig) adjust() error {
...
 			inst.Syncer = &defaultCfg
 		}
+		if inst.Syncer.QueueSize == 0 {
+			inst.Syncer.QueueSize = defaultQueueSize
+		}
+		if inst.Syncer.CheckpointFlushInterval == 0 {
+			inst.Syncer.CheckpointFlushInterval = defaultCheckpointFlushInterval
+		}
+		if inst.Syncer.SafeModeDuration == "" {
+			inst.Syncer.SafeModeDuration = strconv.Itoa(2 * inst.Syncer.CheckpointFlushInterval)
```
lance6716, line 802: I agree that when CheckpointFlushInterval is changed, we should align SafeModeDuration to 2 * inst.Syncer.CheckpointFlushInterval. But seems you forget to add `s` (seconds) to the value. And please check why test didn't find the problem.

And defaultSafeModeDuration can be removed

dm/syncer/syncer.go
```diff
@@ -1618,6 +1618,9 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 	if err != nil {
 		return err
 	}
+	if !fresh && s.cfg.SafeModeDuration == "0s" {
```
lance6716, line 1621: maybe we can tolerate 0s, we already have the feature that syncer will try to exit gracefully, without left the outdated checkpoint.

And s.cliArgs.SafeModeDuration should have higher priority

dm/dm/config/task.go
```diff
@@ -788,6 +792,19 @@ func (c *TaskConfig) adjust() error {
 			defaultCfg := DefaultSyncerConfig()
 			inst.Syncer = &defaultCfg
 		}
+		if inst.Syncer.QueueSize == 0 {
```
GMHDBJD, line 795: These lines already cover at https://github.com/pingcap/tiflow/blob/master/dm/dm/config/subtask.go#L447-L452

dm/syncer/safe_mode.go
```diff
@@ -48,11 +48,39 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+		initPhaseSeconds := s.cfg.SafeModeDuration
+
+		failpoint.Inject("SafeModeInitPhaseSeconds", func(val failpoint.Value) {
+			seconds, _ := val.(string)
+			initPhaseSeconds = seconds
+			s.tctx.L().Info("set initPhaseSeconds", zap.String("failpoint", "SafeModeInitPhaseSeconds"), zap.String("value", seconds))
+		})
+		var duration time.Duration
+		var err error
+		if initPhaseSeconds == "" {
```
GMHDBJD, line 60: Seems already adjust and it could not be empty

dm/syncer/syncer.go
GMHDBJD, line 1621: I think safe-mode-duration = '0' should also be valid

GMHDBJD, line 1621: Why not fresh and no safemode means error? If we have safemode exit point, we don't need safemode

dm/syncer/safe_mode.go
okJiang: for some unit test

dm/dm/config/task.go
okJiang: https://github.com/pingcap/tiflow/pull/6225#discussion_r918647405.

dm/pkg/terror/error_list.go
```diff
@@ -1131,6 +1134,7 @@ var (
 	ErrSyncerUnsupportedStmt                = New(codeSyncerUnsupportedStmt, ClassSyncUnit, ScopeInternal, LevelHigh, "`%s` statement not supported in %s mode", "")
 	ErrSyncerGetEvent                       = New(codeSyncerGetEvent, ClassSyncUnit, ScopeUpstream, LevelHigh, "get binlog event error: %v", "Please check if the binlog file could be parsed by `mysqlbinlog`.")
 	ErrSyncerDownstreamTableNotFound        = New(codeSyncerDownstreamTableNotFound, ClassSyncUnit, ScopeInternal, LevelHigh, "downstream table %s not found", "")
+	ErrSyncerReprocessWithSafeModeFail      = New(codeSyncerReprocessWithSafeModeFail, ClassSyncUnit, ScopeInternal, LevelMedium, "Your `safe-mode-duration` in task.yaml is set 0s, so you can't re-process with safe mode.", "Please stop and re-start this task. If you want to start task successfully, you need set `safe-mode-duration` > `0s`")
```
lance6716, line 1137:
```suggestion
	ErrSyncerReprocessWithSafeModeFail      = New(codeSyncerReprocessWithSafeModeFail, ClassSyncUnit, ScopeInternal, LevelMedium, "your `safe-mode-duration` in task.yaml is set to 0s, the task can't be re-processed without safe mode currently", "Please stop and re-start this task. If you want to start task successfully, you need set `safe-mode-duration` > `0s`.")
```

dm/syncer/safe_mode.go
```diff
@@ -43,35 +56,68 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
-	if s.checkpoint.SafeModeExitPoint() != nil {
-		//nolint:errcheck
-		s.safeMode.Add(tctx, 1) // enable and will revert after pass SafeModeExitLoc
-		s.tctx.L().Info("enable safe-mode for safe mode exit point, will exit at", zap.Stringer("location", *s.checkpoint.SafeModeExitPoint()))
+	exitPoint := s.checkpoint.SafeModeExitPoint()
+	failpoint.Inject("SafeModeDurationIsZero", func() {
+		exitPoint = nil
+	})
+	if exitPoint != nil {
+		if binlog.CompareLocation(*exitPoint, s.getInitExecutedLoc(), s.cfg.EnableGTID) > 0 {
```
lance6716, line 64: seems initExecutedLoc is only recorded once when Syncer is started, it will not be updated when subtask pause/resume.

dm/syncer/safe_mode.go
```diff
@@ -43,35 +56,68 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
-				if err != nil {
-					// send error to the fatal chan to interrupt the process
-					s.runFatalChan <- unit.NewProcessError(err)
-				}
-				if !s.safeMode.Enable() {
-					s.tctx.L().Info("disable safe-mode after task initialization finished")
+		initPhaseSeconds := s.cfg.SafeModeDuration
+
+		failpoint.Inject("SafeModeInitPhaseSeconds", func(val failpoint.Value) {
+			seconds, _ := val.(string)
```
lance6716, line 75:
```suggestion
			seconds := val.(string)
```

dm/syncer/safe_mode.go
```diff
@@ -43,35 +56,68 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+		})
+		var duration time.Duration
+		if initPhaseSeconds == "" {
+			duration = time.Second * time.Duration(2*s.cfg.CheckpointFlushInterval)
+		} else {
+			duration, err = time.ParseDuration(initPhaseSeconds)
+			if err != nil {
+				// send error to the fatal chan to interrupt the process
+				s.runFatalChan <- unit.NewProcessError(err)
+				s.tctx.L().Error("enable safe-mode failed due to duration parse failed", zap.Duration("duration", duration))
```
lance6716, line 87:
```suggestion
				s.tctx.L().Error("enable safe-mode failed due to duration parse failed", zap.Duration("duration", duration))
```

when duration parsing failed, `duration` is the zero value. You should return `initPhaseSeconds`?

dm/syncer/safe_mode.go
```diff
@@ -43,35 +56,68 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+			}
+		}
+		s.tctx.L().Info("enable safe-mode because of task initialization", zap.Duration("duration", duration))
+
+		if int64(duration) > 0 {
+			//nolint:errcheck
+			s.safeMode.Add(tctx, 1) // enable and will revert after 2 * CheckpointFlushInterval
+			go func() {
+				var err error
+				defer func() {
```
lance6716, line 98: this defer is duplicated with line 36?

dm/syncer/safe_mode.go
okJiang: I don't think so. `Syncer.Process()` is also called by `Syncer.Resume()`

lance6716: `initExecutedLoc` will be set only when it's `nil`, so it will not be written again when pause/resume

okJiang: You are right! I added `beginLocation` in `Syncer`, and use it to compare with `exitPoint`. `beginLocation` will init whatever start/resume.

engine/jobmaster/dm/config/dm_subtask_3307.toml
```diff
@@ -58,6 +59,7 @@ ansi-quotes = false
   port = 3307
   user = "root"
   password = "******"
+  syncer-config-name = "sync-01"
```
okJiang, line 62:
```suggestion
```

dm/syncer/syncer.go
```diff
@@ -3172,10 +3171,7 @@ func (s *Syncer) handleQueryEventPessimistic(qec *queryEventContext) error {
 			zap.String("sourceTableID", sourceTableID),
 			zap.Stringer("start location", startLocation),
 			log.WrapStringerField("end location", currentLocation))
-		err = s.safeMode.DescForTable(qec.tctx, ddlInfo.targetTables[0]) // try disable safe-mode after sharding group synced
```
okJiang, line 3175: remove re-sync safe mode

dm/syncer/safe_mode_test.go
```diff
@@ -80,21 +87,45 @@ func TestEnableSafeModeInitializationPhase(t *testing.T) {
...
 	s.cfg.SafeMode = false
+	s.cfg.SafeModeDuration = ""
 	mockCheckpoint.safeModeExitPoint = &binlog.Location{Position: mysql.Position{Name: "mysql-bin.000123", Pos: 123}}
+	mockCheckpoint.globalPoint = binlog.Location{Position: mysql.Position{Name: "mysql-bin.000123", Pos: 120}}
+	s.initInitExecutedLoc()
 	s.enableSafeModeInitializationPhase(s.tctx)
 	require.True(t, s.safeMode.Enable())
 
 	// test enable by initPhaseSeconds
+	s.checkpoint = &mockCheckpointForSafeMode{}
```
okJiang, line 106: fix previous unstable test

dm/tests/safe_mode/run.sh
```diff
@@ -239,7 +298,7 @@ function run() {
 		# DM-worker1 is sharding lock owner and exits
 		if [ "$(check_port_return $WORKER1_PORT)" == "0" ]; then
 			echo "DM-worker1 is sharding lock owner and detects it offline"
-			export GO_FAILPOINTS='github.com/pingcap/tiflow/dm/syncer/SafeModeInitPhaseSeconds=return(0)'
```
okJiang, line 242: this failpoint is for closing safe mode in re-sync.

If dm needn't enable safe mode in re-sync, do we consider removing it?

okJiang:
the it is passed, ptal again. @lance6716 @GMHDBJD 

It should be noted that if disable safe-mode derectly when beginLocation equal safeModeExitPoint, it cause lots of bugs. I can't do it this pr and revert it. https://github.com/pingcap/tiflow/pull/6225#discussion_r923607530.

dm/pkg/terror/error_list.go
```diff
@@ -1132,6 +1135,7 @@ var (
 	ErrSyncerUnsupportedStmt                = New(codeSyncerUnsupportedStmt, ClassSyncUnit, ScopeInternal, LevelHigh, "`%s` statement not supported in %s mode", "")
 	ErrSyncerGetEvent                       = New(codeSyncerGetEvent, ClassSyncUnit, ScopeUpstream, LevelHigh, "get binlog event error: %v", "Please check if the binlog file could be parsed by `mysqlbinlog`.")
 	ErrSyncerDownstreamTableNotFound        = New(codeSyncerDownstreamTableNotFound, ClassSyncUnit, ScopeInternal, LevelHigh, "downstream table %s not found", "")
+	ErrSyncerReprocessWithSafeModeFail      = New(codeSyncerReprocessWithSafeModeFail, ClassSyncUnit, ScopeInternal, LevelMedium, "your `safe-mode-duration` in task.yaml is set to 0s, the task can't be re-processed without safe mode currently", "Please stop and re-start this task. If you want to start task successfully, you need set `safe-mode-duration` > `0s`.")
```
lance6716, line 1138: > you need set `safe-mode-duration` > `0s`

could you explain how should user use `safe-mode-duration = 0s`? In the begining the user can set `safe-mode-duration = 0s`, but when error happens, it must be changed to `safe-mode-duration > 0s`. And the user now have no chance to set it to 0s again, because we don't support "disable safe-mode because beginLocation equal safeModeExitPoint", the task will meet safemode and can't start.

okJiang: As we know, if DM worker arise an abnormal exit and self-start, `Syncer` will enter safe mode.

The user maybe use DM as a transfer tool, and do some operation(sum/count) on the downstream. If DM synced the binlog A(status = 0) to the downstream, and user did some operation(sum/count) and changed its status = 1, then DM worker crashed and re-started, DM will enter safe mode at this moment. This is cause binlog A(status =0) re-sync.

So user can set `safe-mode-duration = 0s` to report an error when DM worker crash and re-start. At this time, they can make some manual adjustments to the business.

lance6716: I mean "safe-mode-duration = 0s" can only be used for once.

okJiang: Oh, I find I forget to add related integration test. I will add it at once!

> So user can set safe-mode-duration = 0s to report an error when DM worker crash and re-start. At this time, they can make some manual adjustments to the business.

Next, user can stop-task, adjust `safe-mode-duration>0s` in `task.yaml`, and start-task successfully.

After that, user can stop-task, adjust `safe-mode-duration=0s` in `task.yaml`, and start-task again. At this time, DM worker exist the `safeModeExitPoint`, user can use it again!

okJiang: The operation is a little complex indeed. But it works and PM suggests.

lance6716: > After that, user can stop-task, adjust safe-mode-duration=0s in task.yaml, and start-task again. At this time, DM worker exist the safeModeExitPoint, user can use it again!

Why DM worker will exit safeModeExitPoint? safeModeExitPoint is always written to checkpoint at its best, so when DM start task, it will enter safe mode. And since user set `safe-mode-duration=0s`, it will report error immediately. (because we don't support "disable safe-mode because beginLocation equal safeModeExitPoint")

okJiang: > DM worker **exist** the safeModeExitPoint, user can use it again!

not

> DM worker will **exit** safeModeExitPoint

I mean DM worker **exist** the safeModeExitPoint, so DM will enter safe mode even if safe-mode-duration=0s(it doesn't report error).

lance6716: This does not behave same as PR description "- set "0s", not step in safe mode. But if dm exists checkpoint, dm must enter safe-mode. **So dm will report an error.**"

okJiang: I am sorry that I didn't illustrate it in the PR description😂: priority `safeModeExitPoint` > `safe-mode-duration` always.

okJiang: > But if dm exists checkpoint, dm must enter safe-mode. So dm will report an error."

I think `exists checkpoint` does not mean `exit safeModeExitPoint`

lance6716: if priority safeModeExitPoint > safe-mode-duration always, why will DM report error the first time it pauses? DM will always have safeModeExitPoint.

okJiang: The latest code reserved the safe-mode logic, but used `compare(exitPoint, beginPosition)` to determine whether need consider `duration=0`, you can see details in https://github.com/pingcap/tiflow/pull/6225/commits/78fcacaf4218f2c083fb445b7ddf53612dcde60f

okJiang: And I withdraw the conclusion that priority `safeModeExitPoint` > `safe-mode-duration`

dm/syncer/safe_mode.go
```diff
@@ -42,36 +56,68 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+	}
+
+	// when exitPoint equal beginLocation, it means safe mode doesn't enable
+	// because safe mode will disable quickly in the syncer main logic
+	if exitPoint != nil && binlog.CompareLocation(*exitPoint, *s.beginLocation, s.cfg.EnableGTID) == 0 {
+		s.tctx.L().Info("safe-mode: exitPoint equal to beginLocation")
+		return
+	}
+	if s.safeMode.Enable() && int64(duration) == 0 {
+		fresh, err2 := s.IsFreshTask(tctx.Ctx)
```
GMHDBJD, line 116: Why fresh should be treated differently

dm/syncer/safe_mode.go
```diff
@@ -42,36 +56,68 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
-			select {
-			case <-tctx.Context().Done():
-			case <-time.After(time.Duration(initPhaseSeconds) * time.Second):
-			}
-		}()
+		}
+	}
+
+	// when exitPoint equal beginLocation, it means safe mode doesn't enable
+	// because safe mode will disable quickly in the syncer main logic
```
GMHDBJD, line 110: Why enable and close later instead of not enable it in the first place?

dm/syncer/syncer.go
```diff
@@ -236,6 +236,8 @@ type Syncer struct {
 	locations *locationRecorder
 	// initial executed binlog location, set once for each instance of syncer.
 	initExecutedLoc *binlog.Location
+	// beginLocation is the beginning location when syncer start/resume.
+	beginLocation *binlog.Location
```
GMHDBJD, line 240: Can we just use s.checkpoint.GlobalPoint()?

GMHDBJD:
Will this part https://github.com/pingcap/tiflow/blob/b4bbb862318b6a516e591678a2fe098cbba2b6d4/dm/syncer/syncer.go#L3147 affect user? Can we remove this logic?

dm/syncer/safe_mode.go
```diff
@@ -42,36 +56,68 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
 		//nolint:errcheck
 		s.safeMode.Add(tctx, 1) // add 1 but has no corresponding -1, so keeps enabled
 		s.tctx.L().Info("enable safe-mode by config")
+		return
```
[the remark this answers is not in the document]
okJiang: I will remove it

GMHDBJD:
Another requirement is to report an error when safemode-duration is too short, did you implement it?

dm/syncer/safe_mode.go
GMHDBJD: > Determine if it is the first time to start the task this machine.

GMHDBJD: > Determine if it is the first time to start the task this machine.
> 
> If it is, `duration = 0` doesn't matter.

It's the first time to start sync, why it doesn't matter? If user use --consistency none for dumpling, I think we should deal with it aslo

okJiang:
> Will this part
> 
> https://github.com/pingcap/tiflow/blob/b4bbb862318b6a516e591678a2fe098cbba2b6d4/dm/syncer/syncer.go#L3147
> 
> affect user? Can we remove this logic?

I don't confirm. Have you any suggestion? @lance6716

lance6716:
> > Will this part
> > https://github.com/pingcap/tiflow/blob/b4bbb862318b6a516e591678a2fe098cbba2b6d4/dm/syncer/syncer.go#L3147
> > 
> > affect user? Can we remove this logic?
> 
> I don't confirm. Have you any suggestion? @lance6716

I also don't know why we need to open safe mode. Maybe it's because for the old versions when shard resync we didn't maintain checkpoint correctly, so we need safemode as walkaround. I think if CI can pass it's OK to remove it

okJiang:
> Will this part
> 
> https://github.com/pingcap/tiflow/blob/b4bbb862318b6a516e591678a2fe098cbba2b6d4/dm/syncer/syncer.go#L3147
> 
> affect user? Can we remove this logic?

remove this will bring much workload. Don't move it now

okJiang:
> Another requirement is to report an error when safemode-duration is too short, did you implement it?

we can add some note in document.

okJiang:
it passed https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/7890/pipeline

dm/tests/safe_mode/run.sh
```diff
@@ -183,9 +183,83 @@ function safe_mode_recover() {
...
+	dmctl_start_task "$WORK_DIR/dm-task-safe-mode-duration.yaml"
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+
+	sed -i "s/safe-mode-duration: \"30s\"/safe-mode-duration: \"0s\"/" $WORK_DIR/dm-task-safe-mode-duration.yaml
+	# stop and start task success
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"stop-task test" \
+		"\"result\": true" 3
+	dmctl_start_task "$WORK_DIR/dm-task-safe-mode-duration.yaml"
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
```
lance6716, line 251: we should insert some data before check sync diff, otherwise there's no new data change to be checked

dm/syncer/safe_mode.go
```diff
@@ -43,35 +57,77 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+	initPhaseSeconds := s.cfg.SafeModeDuration
+
+	failpoint.Inject("SafeModeInitPhaseSeconds", func(val failpoint.Value) {
+		initPhaseSeconds = val.(string)
+		s.tctx.L().Info("set initPhaseSeconds", zap.String("failpoint", "SafeModeInitPhaseSeconds"), zap.String("value", initPhaseSeconds))
+	})
+	failpoint.Inject("SafeModeDurationSet", func(val failpoint.Value) {
+		initPhaseSeconds = val.(string)
+		s.tctx.L().Info("set initPhaseSeconds", zap.String("failpoint", "SafeModeDurationSet"), zap.String("value", initPhaseSeconds))
+	})
```
GMHDBJD, line 71: This two failpoint do the same thing

okJiang: `SafeModeInitPhaseSeconds` was used in so many places. So `SafeModeInitPhaseSeconds` keep the previous logic.

`SafeModeDurationSet` only change the safe-mode-duration

dm/syncer/safe_mode.go
```diff
@@ -43,35 +57,77 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+	}
+
+	if int64(duration) == 0 {
+		failpoint.Inject("SafeModeInitPhaseSeconds", func(val failpoint.Value) {
+			seconds := val.(string)
+			if seconds == "0s" {
+				failpoint.Return()
 			}
-		}()
+		})
```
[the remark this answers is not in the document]
okJiang: make sure safe mode disable when failpoint `SafeModeInitPhaseSeconds = 0s`

Because `SafeModeInitPhaseSeconds = 0s` was used in so many places

dm/syncer/safe_mode.go
```diff
@@ -43,35 +60,90 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+		if !fresh {
```
lance6716, line 145: start-task without --remove-meta is not a fresh task. So your walkaround "Please stop and re-start this task. If you want to start task successfully, you need set `safe-mode-duration` > `0s`." should not work. Please check why test didn't find this problem

dm/dm/config/subtask.go
```diff
@@ -450,6 +452,12 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 	if c.SyncerConfig.CheckpointFlushInterval == 0 {
 		c.SyncerConfig.CheckpointFlushInterval = defaultCheckpointFlushInterval
 	}
+	if c.SyncerConfig.SafeModeDuration == "" {
+		c.SyncerConfig.SafeModeDuration = strconv.Itoa(2*c.SyncerConfig.CheckpointFlushInterval) + "s"
+	}
+	if _, err := time.ParseDuration(c.SyncerConfig.SafeModeDuration); err != nil {
+		return terror.ErrConfigInvalidSafeModeDuration.Generate(c.SyncerConfig.SafeModeDuration, err)
+	}
```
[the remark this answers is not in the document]
okJiang: It's ok

dm/syncer/safe_mode.go
```diff
@@ -43,35 +60,90 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+	failpoint.Inject("SafeModeInitPhaseSeconds", func(val failpoint.Value) {
+		initPhaseSeconds = val.(string)
+		s.tctx.L().Info("set initPhaseSeconds", zap.String("failpoint", "SafeModeInitPhaseSeconds"), zap.String("value", initPhaseSeconds))
+	})
+	failpoint.Inject("SafeModeDurationSet", func(val failpoint.Value) {
+		initPhaseSeconds = val.(string)
+		s.tctx.L().Info("set initPhaseSeconds", zap.String("failpoint", "SafeModeDurationSet"), zap.String("value", initPhaseSeconds))
+	})
+	if initPhaseSeconds == "" {
+		duration = time.Second * time.Duration(2*s.cfg.CheckpointFlushInterval)
```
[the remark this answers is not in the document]
okJiang: `s.cfg.SafeModeDuration` is set to `initPhaseSeconds` in L80

dm/syncer/safe_mode.go
```diff
@@ -43,35 +60,90 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+			return
+		}
+	}
+	exitPoint := s.checkpoint.SafeModeExitPoint()
+	beginLocation := s.checkpoint.GlobalPoint()
+	failpoint.Inject("SafeModeDurationSetBeginLoc", func() {
+		beginLocation = binlog.MustZeroLocation(mysql.MySQLFlavor)
+	})
+	if exitPoint != nil {
+		s.tctx.L().Info("compare exitPoint and beginLocation", zap.Stringer("exit point", exitPoint), zap.Stringer("begin location", beginLocation))
```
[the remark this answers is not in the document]
okJiang: checked rest places

dm/syncer/safe_mode.go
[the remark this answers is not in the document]
okJiang: "make sure safe mode disable"  is inaccurate.

maybe disable InitPhaseSeconds affect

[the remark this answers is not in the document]
okJiang: > > make sure safe mode disable when failpoint `SafeModeInitPhaseSeconds = 0s`
> > Because `SafeModeInitPhaseSeconds = 0s` was used in so many places
> 
> I can't understand why failpoint.Return() is related to "make sure safe mode disable"

Because now, `safe-mode-duration=0s` will report an error when `task not fresh`

dm/syncer/safe_mode.go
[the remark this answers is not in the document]
lance6716: I mean why duration is not set to `s.cfg.SafeModeDuration` in line 92? maybe it's to handle 0s for ConsistencyTypeNone?

When ConsistencyTypeNone, safeModeExitPoint is set in https://github.com/pingcap/tiflow/blob/f388d67b343cfb6b4f502253cb8c562695ac95b8/dm/syncer/checkpoint.go#L1194 And for the first time it's "fresh", no error should happen.

dm/syncer/safe_mode.go
```diff
@@ -43,35 +60,90 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
-				if !s.safeMode.Enable() {
-					s.tctx.L().Info("disable safe-mode after task initialization finished")
+		s.tctx.L().Info("enable safe-mode because of task initialization", zap.Duration("duration", duration))
+
+		if int64(duration) > 0 {
+			//nolint:errcheck
+			s.safeMode.Add(tctx, 1) // enable and will revert after 2 * CheckpointFlushInterval
+			go func() {
+				defer func() {
+					err = s.safeMode.Add(tctx, -1)
```
lance6716, line 124: this `err` may race with the accessing from outer goroutine.

dm/syncer/safe_mode.go
```diff
@@ -30,6 +36,17 @@ func (s *Syncer) enableSafeModeByTaskCliArgs(tctx *tcontext.Context) {
...
 
 func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
+	var err error
+	defer func() {
+		if err != nil {
+			// send error to the fatal chan to interrupt the process
+			s.runFatalChan <- unit.NewProcessError(err)
+		}
+		if !s.safeMode.Enable() {
+			s.tctx.L().Warn("enable safe-mode failed")
```
lance6716, line 46: maybe we should not write log as "failed", because when checkpoint equals safeModeExitPoint it's expected

dm/syncer/syncer_test.go
```diff
@@ -1001,10 +1002,6 @@ func (s *testSyncerSuite) TestRun(c *C) {
 		updateJobMetricsFn: func(bool, string, *job) {},
 	}
 
-	// When crossing safeModeExitPoint, will generate a flush sql
```
lance6716, line 1004: these deletion is unexpected, see my next comment in syncer.go

dm/syncer/syncer.go
```diff
@@ -2240,7 +2240,7 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 
 		// check pass SafeModeExitLoc and try disable safe mode, but not in sharding or replacing error
 		safeModeExitLoc := s.checkpoint.SafeModeExitPoint()
-		if safeModeExitLoc != nil && shardingReSync == nil {
+		if s.safeMode.Enable() && safeModeExitLoc != nil && shardingReSync == nil {
```
lance6716, line 2243: your change will cause the problem stated in line 2248 ~ 2252. Please keep old logic or flush FlushSafeModeExitPoint when it's set to nil in safe_mode.go

dm/syncer/safe_mode.go
okJiang: If so, maybe remove it is better

dm/syncer/safe_mode.go
okJiang: Good catch! I will update it.

dm/syncer/syncer.go
okJiang: I think it is the safest that exec `FlushSafeModeExitPoint` in the safe_mode.go.

But, a discussion, L2248-L2252 does also not occur without `FlushSafeModeExitPoint when it's set to nil in safe_mode.go`. Because `enableSafeModeInitializationPhase` is only called in the `Syncer.Run()` beginning. At this time, `SafeModeExitPoint` just Load from Syncer.go:456 and Syncer.go:1660. So, it is same as the disk's `SafemodeExitPoint`.

GMHDBJD:
https://github.com/pingcap/tiflow/blob/8f3fd7a7f30fa6eefc9bc95879eee2653ab6c937/dm/dm/config/task.go#L1075 Copy the new config item for downgrade and omit it if it has default value

dm/dm/config/task.go
```diff
@@ -1112,6 +1131,15 @@ func NewSyncerConfigsForDowngrade(syncerConfigs map[string]*SyncerConfig) map[st
 	return syncerConfigsForDowngrade
 }
 
+// omitDefaultVals change default value to empty value for new config item.
+// If any default value for new config item is not empty(0 or false or nil),
+// we should change it to empty.
+func (c *SyncerConfigForDowngrade) omitDefaultVals() {
+	if c.SafeModeDuration == DefaultSyncerConfig().SafeModeDuration {
```
GMHDBJD, line 1138: prefer
```
const safeModeDuration
```
rather than define it in DefaultSyncerConfig

dm/syncer/safe_mode.go
```diff
@@ -43,35 +57,101 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
 		s.safeMode.Add(tctx, 1) // add 1 but has no corresponding -1, so keeps enabled
 		s.tctx.L().Info("enable safe-mode by config")
 	}
-	if s.checkpoint.SafeModeExitPoint() != nil {
+
+	var duration time.Duration
+
+	exportCfg := export.DefaultConfig()
```
GMHDBJD, line 63: No need for this part of the logic

dm/syncer/safe_mode.go
```diff
@@ -43,35 +57,101 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+		s.tctx.L().Info("compare exitPoint and beginLocation", zap.Stringer("exitPoint", exitPoint), zap.Stringer("beginLocation", beginLocation))
+		if binlog.CompareLocation(*exitPoint, beginLocation, s.cfg.EnableGTID) == 0 {
+			s.tctx.L().Info("exitPoint equal to beginLocation, so disable the safe mode")
+			s.checkpoint.SaveSafeModeExitPoint(nil)
+			// must flush here to avoid the following situation:
+			// 1. quit safe mode
+			// 2. push forward and replicate some sqls after safeModeExitPoint to downstream
+			// 3. quit because of network error, fail to flush global checkpoint and new safeModeExitPoint to downstream
+			// 4. restart again, quit safe mode at safeModeExitPoint, but some sqls after this location have already been replicated to the downstream
+			if err = s.checkpoint.FlushSafeModeExitPoint(s.runCtx); err != nil {
```
GMHDBJD, line 112:
```
return s.checkpoint.FlushSafeModeExitPoint
```

dm/syncer/safe_mode.go
```diff
@@ -43,35 +57,101 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
+				return
+			}
+			return
+		}
+
 		//nolint:errcheck
 		s.safeMode.Add(tctx, 1) // enable and will revert after pass SafeModeExitLoc
-		s.tctx.L().Info("enable safe-mode for safe mode exit point, will exit at", zap.Stringer("location", *s.checkpoint.SafeModeExitPoint()))
+		s.tctx.L().Info("enable safe-mode for safe mode exit point, will exit at", zap.Stringer("location", *exitPoint))
```
GMHDBJD, line 120: Will `*exitPoint` panic?

GMHDBJD (commented):
rest LGTM

dm/dm/config/task.go
okJiang: The consideration is define it in DefaultSyncerConfig can follow `CheckpointFlushInterval` directly.

dm/syncer/safe_mode.go
okJiang: `exitPoint` is not nil

dm/dm/config/task.go
GMHDBJD: You can directly define it as
```
var (
    SafeModeDuration = strconv.Itoa(2*defaultCheckpointFlushInterval) + "s"
)
```

dm/config/subtask.go
```diff
@@ -450,6 +451,14 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 	if c.SyncerConfig.CheckpointFlushInterval == 0 {
 		c.SyncerConfig.CheckpointFlushInterval = defaultCheckpointFlushInterval
 	}
+	if c.SyncerConfig.SafeModeDuration == "" {
+		c.SyncerConfig.SafeModeDuration = defaultSafeModeDuration
```
lance6716, line 455: when user has set non-empty CheckpointFlushInterval and empty SafeModeDuration, we should set SafeModeDuration to 2*CheckpointFlushInterval rather than defaultSafeModeDuration

dm/config/task.go
```diff
@@ -788,6 +792,20 @@ func (c *TaskConfig) adjust() error {
...
 			inst.Syncer = &defaultCfg
 		}
+		if inst.Syncer.QueueSize == 0 {
+			inst.Syncer.QueueSize = defaultQueueSize
+		}
+		if inst.Syncer.CheckpointFlushInterval == 0 {
+			inst.Syncer.CheckpointFlushInterval = defaultCheckpointFlushInterval
+		}
+		if inst.Syncer.SafeModeDuration == "" {
+			inst.Syncer.SafeModeDuration = defaultSafeModeDuration
```
lance6716, line 802: also here

dm/config/task.go
```diff
@@ -1112,6 +1132,15 @@ func NewSyncerConfigsForDowngrade(syncerConfigs map[string]*SyncerConfig) map[st
 	return syncerConfigsForDowngrade
 }
 
+// omitDefaultVals change default value to empty value for new config item.
+// If any default value for new config item is not empty(0 or false or nil),
+// we should change it to empty.
+func (c *SyncerConfigForDowngrade) omitDefaultVals() {
+	if c.SafeModeDuration == DefaultSyncerConfig().SafeModeDuration {
```
lance6716, line 1139: the check should also be updated to 2*CheckpointFlushInterval

dm/config/task.go
```diff
@@ -1136,16 +1165,16 @@ type TaskConfigForDowngrade struct {
...
 	CleanDumpFile           bool                                 `yaml:"clean-dump-file"`
 	EnableANSIQuotes        bool                                 `yaml:"ansi-quotes"`
 	RemoveMeta              bool                                 `yaml:"remove-meta"`
 	// new config item
-	MySQLInstances   []*MySQLInstanceForDowngrade `yaml:"mysql-instances"`
-	ExprFilter       map[string]*ExpressionFilter `yaml:"expression-filter,omitempty"`
-	OnlineDDL        bool                         `yaml:"online-ddl,omitempty"`
-	ShadowTableRules []string                     `yaml:"shadow-table-rules,omitempty"`
-	TrashTableRules  []string                     `yaml:"trash-table-rules,omitempty"`
+	Syncers          map[string]*SyncerConfigForDowngrade `yaml:"syncers,omitempty"`
```
lance6716, line 1172: no need to move Syncer here, Mydumpers/Loaders/Syncer are more relevant

dm/tests/safe_mode/run.sh
```diff
@@ -183,9 +183,93 @@ function safe_mode_recover() {
...
+	sed -i "/relay-binlog-name/i\relay-dir: $WORK_DIR/worker1/relay_log" $WORK_DIR/source1.yaml
+	dmctl_operate_source create $WORK_DIR/source1.yaml $SOURCE_ID1
+	# worker2 -> source2
+	run_dm_worker $WORK_DIR/worker2 $WORKER2_PORT $cur/conf/dm-worker2.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
+	cp $cur/conf/source2.yaml $WORK_DIR/source2.yaml
+	sed -i "/relay-binlog-name/i\relay-dir: $WORK_DIR/worker2/relay_log" $WORK_DIR/source2.yaml
+	dmctl_operate_source create $WORK_DIR/source2.yaml $SOURCE_ID2
+
+	dmctl_start_task "$cur/conf/dm-task-safe-mode-duration.yaml" "--remove-meta"
```
lance6716, line 208: failpoint `SafeModeDurationSetBeginLoc` makes tests can't reflect real scenario, the execution path is changed. You can inject some errors in `DMLWorker.executeBatchJobs`

dm/tests/safe_mode/run.sh
```diff
@@ -183,9 +183,93 @@ function safe_mode_recover() {
...
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+	run_dm_worker $WORK_DIR/worker2 $WORKER2_PORT $cur/conf/dm-worker2.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
+
+	cp $cur/conf/dm-task-safe-mode-duration.yaml $WORK_DIR/dm-task-safe-mode-duration.yaml
+	sed -i "s/safe-mode-duration: \"0s\"/safe-mode-duration: \"30s\"/" $WORK_DIR/dm-task-safe-mode-duration.yaml
+
+	dmctl_start_task "$WORK_DIR/dm-task-safe-mode-duration.yaml"
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
```
lance6716, line 242: also insert some data to verify the task is normal

dm/tests/safe_mode/run.sh
okJiang: After https://github.com/pingcap/tiflow/pull/6225/commits/def5228201236d7ee3c226abbc6c1f9ae080d02a, the prepare sql can be executed to downstream. So needn't inject errors in `DMLWorker.executeBatchJobs`

dm/tests/safe_mode/run.sh
okJiang: I think the task has been verified normal already.

1. L212/L213 inserted data
2. L222-224 showed the failed task
3. After modifying the `safe-mode-duration` from `0s` to `30s`, we can see task is normal here(L242), which shows the rest inserted data in L212/L213 can be synced to downstream.

dm/syncer/safe_mode.go
```diff
@@ -43,35 +53,79 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
-				s.tctx.L().Info("set initPhaseSeconds", zap.String("failpoint", "SafeModeInitPhaseSeconds"), zap.Int("value", seconds))
-			})
-			s.tctx.L().Info("enable safe-mode because of task initialization", zap.Int("duration in seconds", initPhaseSeconds))
-			select {
-			case <-tctx.Context().Done():
-			case <-time.After(time.Duration(initPhaseSeconds) * time.Second):
+		}
+	}
+
+	if int64(duration) == 0 && !s.cfg.SafeMode {
```
okJiang, line 122: omit one condition, will add it later

dm/syncer/safe_mode.go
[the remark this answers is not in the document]
lance6716: What's the expected test scenario for it?

lance6716 (commented):
will review after lunch

dm/tests/safe_mode/run.sh
```diff
@@ -183,9 +183,92 @@ function safe_mode_recover() {
...
+	# worker2 -> source2
+	run_dm_worker $WORK_DIR/worker2 $WORKER2_PORT $cur/conf/dm-worker2.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
+	cp $cur/conf/source2.yaml $WORK_DIR/source2.yaml
+	sed -i "/relay-binlog-name/i\relay-dir: $WORK_DIR/worker2/relay_log" $WORK_DIR/source2.yaml
+	dmctl_operate_source create $WORK_DIR/source2.yaml $SOURCE_ID2
+
+	dmctl_start_task "$cur/conf/dm-task-safe-mode-duration.yaml" "--remove-meta"
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+	sleep 1
```
lance6716, line 209:
```suggestion
```

dm/tests/safe_mode/run.sh
```diff
@@ -183,9 +183,92 @@ function safe_mode_recover() {
...
+	# restart workers
+	kill_dm_worker
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+	run_dm_worker $WORK_DIR/worker2 $WORKER2_PORT $cur/conf/dm-worker2.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
+
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"query-status test" \
+		"your \`safe-mode-duration\` in task.yaml is set to 0s" 1
```
lance6716, line 223: why we meet error here? I think DM should enter "safe mode exit point" path

okJiang: Because after unexpectedly exiting in L216, the beginning location is not equal to the `exitPoint`.

lance6716: L216 will not make sure dm-worker exit in the middle of executing DML, maybe we need some failpoint

dm/syncer/syncer.go
```diff
@@ -2189,7 +2189,7 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 
 		// check pass SafeModeExitLoc and try to disable safe mode, but not in sharding or replacing error
 		safeModeExitLoc := s.checkpoint.SafeModeExitPoint()
-		if safeModeExitLoc != nil && shardingReSync == nil {
+		if s.safeMode.Enable() && safeModeExitLoc != nil && shardingReSync == nil {
```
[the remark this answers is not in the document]
okJiang: It's because `Syncer` dosen't enter safe-mode when `exitPoint` equal to `beginPosition` due to the safe_mode.go(L82-L92). So, when dm restart in this situation, dm will `SaveSafeModeExitPoint` and disable safe mode(cause error) in L2204 again without `s.safeMode.Enable()` of this line.

Therefore, I added it.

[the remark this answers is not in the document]
lance6716: in safe_mode.go line 84 we should save nil to `SafeModeExitPoint`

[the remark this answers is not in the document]
okJiang: However, I don't know what's any problems that checking safe-mode enable before disable safe-mode.🤔

dm/config/task.go
```diff
@@ -1140,7 +1169,7 @@ type TaskConfigForDowngrade struct {
 	BAList                  map[string]*filter.Rules             `yaml:"block-allow-list"`
 	Mydumpers               map[string]*MydumperConfig           `yaml:"mydumpers"`
 	Loaders                 map[string]*LoaderConfigForDowngrade `yaml:"loaders"`
-	Syncers                 map[string]*SyncerConfigForDowngrade `yaml:"syncers"`
+	Syncers                 map[string]*SyncerConfigForDowngrade `yaml:"syncers,omitempty"`
```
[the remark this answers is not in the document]
okJiang: fixed in https://github.com/pingcap/tiflow/pull/6225/commits/40eab235f9a76b7eb78792a4c42a8d72a1183f8c

dm/syncer/safe_mode.go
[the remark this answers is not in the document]
okJiang: removed it https://github.com/pingcap/tiflow/pull/6225/commits/f70bfb5e4b7576e6df6f2e211f569510225778fe @lance6716

dm/tests/safe_mode/run.sh
okJiang: added in https://github.com/pingcap/tiflow/pull/6225/commits/82d42e10eac8d54744071a00ceb582148d222da8

lance6716:
please check the log, don't blindly rerun CI. @okJiang

okJiang:
shardddl2 failed
/run-dm-integration-test

okJiang:
ptal~ @lance6716 @GMHDBJD

dm/syncer/safe_mode.go
```diff
@@ -43,35 +53,68 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
...
-				err := s.safeMode.Add(tctx, -1)
-				if err != nil {
-					// send error to the fatal chan to interrupt the process
-					s.runFatalChan <- unit.NewProcessError(err)
-				}
-				if !s.safeMode.Enable() {
-					s.tctx.L().Info("disable safe-mode after task initialization finished")
+		s.tctx.L().Info("enable safe-mode because of task initialization", zap.Duration("duration", duration))
+
+		if int64(duration) > 0 {
```
lance6716, line 95:
```suggestion
		if duration > 0 {
```

and line 117

dm/syncer/safe_mode.go
```diff
@@ -43,35 +53,68 @@ func (s *Syncer) enableSafeModeInitializationPhase(tctx *tcontext.Context) {
 		s.safeMode.Add(tctx, 1) // add 1 but has no corresponding -1, so keeps enabled
 		s.tctx.L().Info("enable safe-mode by config")
 	}
-	if s.checkpoint.SafeModeExitPoint() != nil {
+
+	var duration time.Duration
```
lance6716, line 57: the duration has complicated read and write, can we change the logic to

if exitPoint != nil {
  check checkpoint is equal to exitPoint, return nil
  check has "0s" safe-mode-duration, return error
} else {
  turn on safe mode for safe-mode-duration
}

dm/tests/safe_mode/run.sh
```diff
@@ -150,6 +149,7 @@ function safe_mode_recover() {
 		check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
 		check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
 
+		# sleep 3
```
lance6716, line 152:
```suggestion
```

please check other `sleep` in changed content, they should not be needed?

dm/tests/safe_mode/run.sh
```diff
@@ -183,9 +183,96 @@ function safe_mode_recover() {
...
+		"query-status test" \
+		"your \`safe-mode-duration\` in task.yaml is set to 0s" 1
+
+	# stop and start task success
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"stop-task test" \
+		"\"result\": true" 3
+
+	# restart workers
+	kill_dm_worker
```
lance6716, line 235: this is not needed?

dm/tests/safe_mode/run.sh
```diff
@@ -183,9 +183,96 @@ function safe_mode_recover() {
...
+
+	sed -i "s/safe-mode-duration: \"30s\"/safe-mode-duration: \"0s\"/" $WORK_DIR/dm-task-safe-mode-duration.yaml
+	# stop and start task success
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"stop-task test" \
+		"\"result\": true" 3
+
+	export GO_FAILPOINTS=""
+	# restart workers
+	kill_dm_worker
```
lance6716, line 255: ditto

dm/config/task.go
```diff
@@ -325,8 +327,9 @@ type SyncerConfig struct {
 	AutoFixGTID bool `yaml:"auto-fix-gtid" toml:"auto-fix-gtid" json:"auto-fix-gtid"`
 	EnableGTID  bool `yaml:"enable-gtid" toml:"enable-gtid" json:"enable-gtid"`
 	// deprecated
-	DisableCausality bool `yaml:"disable-detect" toml:"disable-detect" json:"disable-detect"`
-	SafeMode         bool `yaml:"safe-mode" toml:"safe-mode" json:"safe-mode"`
+	DisableCausality bool   `yaml:"disable-detect" toml:"disable-detect" json:"disable-detect"`
+	SafeMode         bool   `yaml:"safe-mode" toml:"safe-mode" json:"safe-mode"`
+	SafeModeDuration string `yaml:"safe-mode-duration" toml:"safe-mode-duration" json:"safe-mode-duration"`
```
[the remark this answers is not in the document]
okJiang: What do you mean??

dm/syncer/safe_mode.go
okJiang: yes, you are the right. I'm sry I forgot it😂

dm/syncer/safe_mode_test.go
```diff
@@ -80,21 +94,48 @@ func TestEnableSafeModeInitializationPhase(t *testing.T) {
 	// test enable by config
 	s.cliArgs = nil
 	s.cfg.SafeMode = true
+	s.cfg.SafeModeDuration = "0" // test safeMode's priority higher than SafeModeDuration's
 	mockCheckpoint := &mockCheckpointForSafeMode{}
+	mockCheckpoint.globalPoint = binlog.Location{}
+	mockCheckpoint.tablePoint = make(map[string]map[string]binlog.Location)
 	s.checkpoint = mockCheckpoint
 	s.enableSafeModeInitializationPhase(s.tctx)
 	require.True(t, s.safeMode.Enable())
```
lance6716, line 103: but in fact this case (safe-mode = true & safe-mode-duration = 0s) should not happen?

dm/tests/sequence_safe_mode/run.sh
```diff
@@ -90,7 +90,7 @@ function run() {
 		fi
 
 		((i += 1))
-		echo "wait for one of DM-worker offine failed, retry later" && sleep 1
+		echo "wait for one of DM-worker offine failed, retry later" && sleep 2
```
lance6716, line 93:
```suggestion
		echo "wait for one of DM-worker offine failed, retry later" && sleep 1
```

lance6716 (commented):
rest lgtm

dm/syncer/safe_mode_test.go
okJiang: In short, `safe-mode` has a high priority to `safe-mode-duration`. So `safe-mode-duration = 0s` doesn't affect `safe-mode = true`.

okJiang: > but in fact this case (safe-mode = true & safe-mode-duration = 0s) should not happen?

This case may not happen.

The test case is just testing the priority.

GMHDBJD (approved):
LGTM

lance6716 (approved):
/unhold

PR #6243, opened 2022-07-11
cdc: change redo meta resolved timestamp correctly

### LINKED ISSUE #6189: ticdc redo apply shouldn't depend on meta

- [X] Googled your question

### What is your question?

In the current implement, when applying redo log on a given downstream sink, the event timestamp range is specified from `(meta.checkpointTs, meta.resolvedTs]`.

redo-log meta file is flushed periodically. However, events can be flushed into downstream sinks after **log file** is flushed instead of **redo-log meta**. It makes `cdc redo apply` not be able to ensure downstream integration. For example:

1. cdc table redo log is flushed with timestamp `ts1`;
2. however the current redo-meta resolved timestamp is `ts0`, which is less than `ts1`;
3. events before `ts1` have been flushed into downstream sinks;
4. cdc is killed and `cdc redo apply` is called with the current redo-meta.
5. after recovered, the downstream can contain events after the point `ts0`, which breaks the required integration.

I have prepared a [branch](https://github.com/hicqu/ticdc/tree/6.1.0-for-oncall-5117), and please notice the [fatal log](https://github.com/hicqu/ticdc/blob/6.1.0-for-oncall-5117/cdc/sink/mysql/mysql.go#L236). And the panic really happens when I run the branch:
```
[2022/07/05 16:00:27.117 +08:00] [FATAL] [mysql.go:237] ["[QP] mysqlSink.FlushRowChangedEvents is called with ts less than LogWriterResolved"] [stack="github.com/pingcap/tiflow/cdc/sink/mysql.(*mysqlSink).FlushRowChangedEvents\n\tgithub.com/pingcap/tiflow/cdc/sink/mysql/mysql.go:237\ngithub.com/pingcap/tiflow/cdc/sink.(*tableSink).flushResolvedTs\n\tgithub.com/pingcap/tiflow/cdc/sink/table_sink.go:109\ngithub.com/pingcap/tiflow/cdc/sink.(*tableSink).FlushRowChangedEvents\n\tgithub.com/pingcap/tiflow/cdc/sink/table_sink.go:86\ngithub.com/pingcap/tiflow/cdc/processor/pipeline.(*sinkNode).flushSink\n\tgithub.com/pingcap/tiflow/cdc/processor/pipeline/sink.go:155\ngithub.com/pingcap/tiflow/cdc/processor/pipeline.(*sinkNode).updateBarrierTs\n\tgithub.com/pingcap/tiflow/cdc/processor/pipeline/sink.go:338\ngithub.com/pingcap/tiflow/cdc/processor/pipeline.(*sinkNode).HandleMessage\n\tgithub.com/pingcap/tiflow/cdc/processor/pipeline/sink.go:329\ngithub.com/pingcap/tiflow/cdc/processor/pipeline.(*sinkNode).Receive\n\tgithub.com/pingcap/tiflow/cdc/processor/pipeline/sink.go:289\ngithub.com/pingcap/tiflow/pkg/pipeline.(*nodeRunner).run\n\tgithub.com/pingcap/tiflow/pkg/pipeline/runner.go:65\ngithub.com/pingcap/tiflow/pkg/pipeline.(*Pipeline).driveRunner\n\tgithub.com/pingcap/tiflow/pkg/pipeline/pipeline.go:97"]
``` 

The test shows that the above guess is correct.

To resolve the critical bug, we can consider to 
* always flush log meta when flushing log files, or
* remove log meta from the design, and put `checkpointTs` and `resolvedTs` into redo-log entry

### LINKED ISSUE #6240: redo log: redo_apply_cdc_scale_sync data checksum mismatch
### What did you do?

Case in test-infra: 
[redo_apply_cdc_scale_sync](510194)
[redo_enable_cdc_allnode_restart_sync](480260)
Both failed caused by data checksum mismatch

_No response_

### What did you see instead?

Minio log: seach testbed name
testbed: cdc-testbed-tps-871607-1-840

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
main
```

```console
(paste TiCDC version here)
[release-version=v6.2.0-alpha] [git-hash=1de2fb12b2b3633d4e816e7f92cf841350d8a4b5] [git-branch=heads/refs/tags/v6.2.0-alpha] [utc-build-time="2022-07-08 22:27:19"] [go-version="go version go1.18.2 linux/amd64"] [failpoint-build=false]
```

hicqu:

Issue Number: close #6240

* Put all table resolved timestamps into redo meta
* Calculate resolved timestamp correctly when apply redo log
* When flushing redo meta, merge memory content and disk content correctly
* Choose correct meta file when apply redo log

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


cdc/redo/manager.go
```diff
@@ -375,8 +375,20 @@ func (m *ManagerImpl) AddTable(tableID model.TableID, startTs uint64) {
 
 // RemoveTable removes a table from redo log manager
 func (m *ManagerImpl) RemoveTable(tableID model.TableID) {
```
nongfushanquan, line 377: Please add some comments here to explain why flush ts in the rtsmap is necessary

cdc/redo/manager.go
```diff
@@ -404,9 +416,9 @@ func (m *ManagerImpl) prepareForFlush() (tableRtsMap map[model.TableID]model.Ts,
 		unflushed := rts.getUnflushed()
 		flushed := rts.getFlushed()
 		if unflushed > flushed {
-			tableRtsMap[tableID] = unflushed
 			flushed = unflushed
```
nongfushanquan, line 419: if flushed <= unflushed , the tableRtsMap[tableID] = nil before ? which will cause some problems ?

cdc/redo/manager.go
```diff
@@ -447,7 +471,7 @@ func (m *ManagerImpl) flushLog(ctx context.Context, handleErr func(err error)) {
 		defer atomic.StoreInt64(&m.flushing, 0)
```
nongfushanquan, line 471: why a go routine is needed here?

cdc/redo/manager.go
```diff
@@ -420,9 +432,21 @@ func (m *ManagerImpl) prepareForFlush() (tableRtsMap map[model.TableID]model.Ts,
 }
 
 func (m *ManagerImpl) postFlush(tableRtsMap map[model.TableID]model.Ts, minResolvedTs model.Ts) {
-	if minResolvedTs > m.minResolvedTs {
-		atomic.StoreUint64(&m.minResolvedTs, minResolvedTs)
+	for {
```
nongfushanquan, line 435: why the for loop is needed here?

cdc/redo/manager.go
hicqu: In the patch, `prepareForFlush` returns all tables' resolved timestamps, even if they are not changed.

hicqu: The reason is LogWriter needs the information to construct a redo meta.

cdc/redo/manager.go
hicqu: flush should be performed in an individual goroutine. It's same with the master branch.

cdc/processor/pipeline/sink.go
```diff
@@ -137,7 +137,7 @@ func (n *sinkNode) flushSink(ctx context.Context, resolved model.ResolvedTs) (er
 		// 2. tableActor resolvedTs >= processor resolvedTs >= global resolvedTs >= barrierTs
 		redoTs := n.redoManager.GetMinResolvedTs()
 		if redoTs < currentBarrierTs {
-			log.Debug("redoTs should not less than current barrierTs",
+			log.Fatal("redoTs should not less than current barrierTs",
```
overvenus, line 140:
```suggestion
			log.Panic("redoTs should not less than current barrierTs",
```

CharlesCheung96: Whether `schedulerV2` might violate this constraint?

cdc/redo/common/redo.go
```diff
@@ -52,7 +57,54 @@ const (
...
+		}
+	}
+	if minResolvedTs == math.MaxUint64 {
+		return z.CheckPointTs
+	}
+	return minResolvedTs
+}
+
+// MarshalMsg implements msgp.Marshaler
+func (z LogMeta) MarshalMsg(b []byte) ([]byte, error) {
```
zhaoxinyu, line 81: There may exist a compatibility issue here, because previously we use msgpack to encode metadata, but now the metadata is encoded by this customized `MarshalMsg` function. So when users need to apply redolog, could the redo reader decode the old msgpack metadata correctly?

cdc/redo/common/redo.go
```diff
@@ -52,7 +57,54 @@ const (
 
 // LogMeta is used for store meta info.
 type LogMeta struct {
-	CheckPointTs   uint64           `msg:"checkPointTs"`
-	ResolvedTs     uint64           `msg:"resolvedTs"`
-	ResolvedTsList map[int64]uint64 `msg:"-"`
+	CheckPointTs uint64 `msg:"checkPointTs"`
```
zhaoxinyu, line 60: If msgpack is no longer used to encode metadata, the `msg` tag can be removed.

cdc/redo/common/redo.go
```diff
@@ -52,7 +57,54 @@ const (
...
+		binary.LittleEndian.PutUint64(buff[tStart:tStart+8], uint64(tID))
+		binary.LittleEndian.PutUint64(buff[tStart+8:tStart+16], ts)
+		tStart += 16
+	}
+	b = append(b, buff...)
+	return b, nil
+}
+
+// UnmarshalMsg implements msgp.Unmarshaler
+func (z *LogMeta) UnmarshalMsg(b []byte) ([]byte, error) {
```
zhaoxinyu, line 97: Also curious about why the msgpack `MashalMsg` and `UnmarshalMsg` are replaced with these two customized functions?

cdc/redo/common/redo.go
hicqu: Old msgpack data can't be decoded correctly.

The reason to avoid msgpack is seems its map can only support string keys:
> Maps must have string keys. This is intentional (as it preserves JSON interop.) Although non-string map keys are not forbidden by the MessagePack standard, many serializers impose this restriction. (It also means any well-formed struct can be de-serialized into a map[string]interface{}.) The only exception to this rule is that the deserializers will allow you to read map keys encoded as bin types, due to the fact that some legacy encodings permitted this. (However, those values will still be cast to Go strings, and they will be converted to str types when re-encoded. It is the responsibility of the user to ensure that map keys are UTF-8 safe in this case.) The same rules hold true for JSON translation.

cdc/redo/common/redo.go
hicqu: Seems msgp map only support string map key.

cdc/processor/pipeline/sink.go
hicqu: I'm not sure. I think it won't break the consitraint because
1. barrierTs is global min **redo-flushed** table resolved timestamp
2. redoManager.minResolvedTs can only regress when a table is moved in it

@overvenus PTAL thanks.

hicqu: The log level has been reverted to `Warn` because it can really happen.

cdc/redo/writer/writer.go
```diff
@@ -555,19 +521,71 @@ func (l *LogWriter) getMetafileName() string {
...
 		l.meta.CheckPointTs = checkPointTs
+		hasChange = true
+	} else if checkPointTs > 0 && checkPointTs != l.meta.CheckPointTs {
+		log.Panic("flushLogMeta with a regressed checkpoint ts",
+			zap.Uint64("currCheckPointTs", l.meta.CheckPointTs),
+			zap.Uint64("recvCheckPointTs", checkPointTs))
+	}
+
+	for tID, ts := range rtsMap {
+		if l.meta.ResolvedTsList[tID] == ts {
```
zhaoxinyu, line 539: Is it assumed that `rtsMap` contains all keys in `l.meta.ResolvedTsList`? Do we need to check the existence of `l.meta.ResolvedTsList[tID]`?

hicqu: nice catch!

hicqu: Oh, if `tID` doesn't exist in `l.meta.ResolvedTsList`, `l.meta.ResolvedTsList[tID]` will be 0. So seems it's still ok.

cdc/redo/common/redo.go
```diff
@@ -52,7 +59,59 @@ const (
 
 // LogMeta is used for store meta info.
 type LogMeta struct {
-	CheckPointTs   uint64           `msg:"checkPointTs"`
-	ResolvedTs     uint64           `msg:"resolvedTs"`
-	ResolvedTsList map[int64]uint64 `msg:"-"`
+	CheckPointTs   uint64
```
overvenus, line 62: Why remove the `msg` tag?

go.mod
```diff
@@ -104,6 +104,8 @@ require (
 	upper.io/db.v3 v3.7.1+incompatible
 )
 
+require github.com/pkg/errors v0.9.1
```
overvenus, line 107: Why do we require this package?

zhaoxinyu (approved):

cdc/redo/common/redo.go
hicqu: Because now marsual and unmarshal are customed instead of generated.

tests/integration_tests/consistent_replicate_gbk/run.sh
```diff
@@ -90,7 +90,7 @@ function run() {
 	run_sql "insert into consistent_replicate_gbk.GBKTABLE2 select * from consistent_replicate_gbk.GBKTABLE" ${UP_TIDB_HOST} ${UP_TIDB_PORT}
 
 	# to ensure row changed events have been replicated to TiCDC
-	sleep 5
```
CharlesCheung96, line 93: Better to add a comment here indicating that we need to change the wait time back in the future.

CharlesCheung96 (approved):

PR #6252, opened 2022-07-11
schema(cdc): remove useless cas in AdvanceResolvedTs

### LINKED ISSUE #6251: schema(cdc): remove useless cas when advance resolvedTs of schema storage

No

remove useless cas when advance resolvedTs of schema storage

_No response_

_No response_

maxshuang:

Issue Number: close #6251 

1. Remove the cas operation since the AdvanceResolvedTs is called in a single thread
2. Using `Store` to replace the `cas`

#### Tests 

 - Unit test
old ut has cover the case of fallback resolvedTs 

### Release note 

```release-note
`None`.
```


nongfushanquan (commented):
just code optimized or a bug happened ?

maxshuang:
> just code optimized or a bug happened ?

Just  a code optimized. 
When xiongjiwei implements the `concurrent DDL` for cdc, it confuses him on the `cas` operation which seems unnecessary.

3AceShowHand (approved):

cdc/entry/schema_storage.go
```diff
@@ -192,13 +192,8 @@ func (s *schemaStorageImpl) HandleDDLJob(job *timodel.Job) error {
...
-	var swapped bool
-	for !swapped {
-		oldResolvedTs := atomic.LoadUint64(&s.resolvedTs)
-		if ts < oldResolvedTs {
-			return
-		}
-		swapped = atomic.CompareAndSwapUint64(&s.resolvedTs, oldResolvedTs, ts)
+	if ts > s.ResolvedTs() {
+		atomic.StoreUint64(&s.resolvedTs, ts)
 	}
```
amyangfei, line 197: The optimization is correct only if `AdvanceResolvedTs` won't be called concurrently. Please add a comment about it.

Otherwise there will be write skew such as
- ts1 > ts2 > s.ResolvedTs
- Both `AdvanceResolvedTs(ts1)` and `AdvanceResolvedTs(ts2)` enters L196
- The value of `s.resolvedTs` will be either `ts1` or `ts`, which is not expected.

maxshuang: Done. It is optimized since we don't need to call it concurrently.

amyangfei (approved):

PR #6265, opened 2022-07-12
api(cdc): drain capture return 503 if not all captures initialized

3AceShowHand:

Issue Number: ref #4757 

When call the open api to drain the capture, all captures should be initialized, otherwise return 503 indicate that the client should retry.

#### Tests 

 - Unit test

### Release note 

```release-note
`None`
```


cdc/owner/owner.go
```diff
@@ -410,21 +410,36 @@ func (o *ownerImpl) clusterVersionConsistent(captures map[model.CaptureID]*model
...
-	changefeedWithTableCount := 0
-	totalTableCount := 0
+	var (
+		changefeedWithTableCount int
+		totalTableCount          int
+		err                      error
+	)
 	for _, changefeed := range o.changefeeds {
-		count := changefeed.scheduler.DrainCapture(query.CaptureID)
+		count, e := changefeed.scheduler.DrainCapture(query.CaptureID)
```
overvenus, line 419: Please check error right after `DrainCapture`

```suggestion
		count, e := changefeed.scheduler.DrainCapture(query.CaptureID)
		if e != nil {
			err = e
			break
		}
```

cdc/scheduler/internal/v3/coordinator.go
```diff
@@ -139,59 +140,61 @@ func (c *coordinator) Rebalance() {
...
+func (c *coordinator) DrainCapture(target model.CaptureID) (int, error) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	if !c.captureM.CheckAllCaptureInitialized() {
+		log.Info("tpscheduler: drain capture request ignored, "+
+			"since not all captures initialized",
+			zap.String("target", target))
+		// return false to let client retry.
+		return 0, cerror.ErrSchedulerRequestFailed.
```
overvenus, line 153: Please add a test to cover this line.

overvenus (approved):

asddongmen (approved):

PR #6268, opened 2022-07-13
cli(ticdc): cli check is old cdc cluster

### LINKED ISSUE #6267: cli check the if cdc server version is too old
if the new cli connect to a cdc server that before 6.2.0, we should throw an error

`
Can’t operate the TiCDC cluster, the cluster version is too old. Only the version of this cluster larger or equal than v6.2.0 can be operated, please make sure the server version and the cli-tools version are matched.
`

sdojjy:

Issue Number: close #6267

check the cdc server version by calling the status API

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```

### DIFF
--- a/pkg/cmd/factory/factory_impl.go
+++ b/pkg/cmd/factory/factory_impl.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	apiv1client "github.com/pingcap/tiflow/pkg/api/v1"
@@ -202,7 +203,11 @@ func (f *factoryImpl) APIV1Client() (apiv1client.APIV1Interface, error) {
 		return nil, errors.Trace(err)
 	}
 	log.Info(serverAddr)
-	return apiv1client.NewAPIClient(serverAddr, f.clientGetter.GetCredential())
+	client, err := apiv1client.NewAPIClient(serverAddr, f.clientGetter.GetCredential())
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	return client, checkCDCVersion(client)
 }
 
 // APIV2Client returns cdc api v2 client.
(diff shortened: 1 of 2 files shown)


pkg/cmd/factory/factory_impl.go
```diff
@@ -265,3 +277,21 @@ func (f *factoryImpl) findServerAddr() (string, error) {
...
+func checkCDCVersion(client apiv1client.APIV1Interface) error {
+	serverStatus, err := client.Status().Get(cmdconetxt.GetDefaultContext())
+	if err != nil {
+		return errors.Trace(err)
+	}
+	cdcVersion, err := version.GetTiCDCClusterVersion([]string{serverStatus.Version})
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if cdcVersion.LessThan(*semver.New("6.2.0-master")) {
```
nongfushanquan, line 290: 6.2.0-master is OK ？

sdojjy: fixed

pkg/cmd/factory/factory_impl.go
```diff
@@ -265,3 +276,21 @@ func (f *factoryImpl) findServerAddr() (string, error) {
...
+	if err != nil {
+		return errors.Trace(err)
+	}
+	cdcVersion, err := version.GetTiCDCClusterVersion([]string{serverStatus.Version})
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if !cdcVersion.ShouldRunCliWithOpenAPI() {
+		return errors.New("Can’t operate the TiCDC cluster, " +
+			"the cluster version is too old. Only the version of this cluster " +
```
asddongmen, line 291: Would it be better if we return a error like this?
```go
errors.New("Can’t use this cli-tools to operate the TiCDC cluster, " +
			 "the TiCDC cluster server's version '%s' is older than cli-tools '%s'" +
              "please make sure their versions are matched.", serverVersion, cliVersion)
```

sdojjy: the error message is from the feature  spec

pkg/cmd/factory/factory_impl.go
```diff
@@ -202,7 +202,11 @@ func (f *factoryImpl) APIV1Client() (apiv1client.APIV1Interface, error) {
 		return nil, errors.Trace(err)
 	}
 	log.Info(serverAddr)
-	return apiv1client.NewAPIClient(serverAddr, f.clientGetter.GetCredential())
+	client, err := apiv1client.NewAPIClient(serverAddr, f.clientGetter.GetCredential())
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	return client, checkCDCVersion(client)
```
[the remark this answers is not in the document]
sdojjy: older version will be ignored now

asddongmen (approved):

3AceShowHand (approved):

PR #6284, opened 2022-07-14
capture(cdc): do not pass context to the etcd client to prevent etcd operation does not take effects

3AceShowHand:

Issue Number: ref #4757

At the moment, we pass the `ctx` to create the etcd client, it may happen that the context was canceled when the client was processing some internal operation, such as resigning the campaign key;

This is fixed by remove the ctx passed to the etcd client.

Before

image

Now

image

#### Tests 

- manual test

### Release note 

```release-note
`None`.
```


overvenus (dismissed):

cdc/capture/capture.go
```diff
@@ -530,7 +570,10 @@ func (c *captureImpl) campaign(ctx cdcContext.Context) error {
 	failpoint.Inject("capture-campaign-compacted-error", func() {
 		failpoint.Return(errors.Trace(mvcc.ErrCompacted))
 	})
-	return cerror.WrapError(cerror.ErrCaptureCampaignOwner, c.election.Campaign(ctx, c.info.ID))
+
+	stdCtx, cancel := context.WithTimeout(ctx, time.Second)
```
overvenus, line 574: The context always timeout after 1 second, the capture can never be owner.

cdc/capture/capture.go
```diff
@@ -267,7 +268,8 @@ func (c *captureImpl) reset(ctx context.Context) error {
 func (c *captureImpl) Run(ctx context.Context) error {
 	defer log.Info("the capture routine has exited")
 	// Limit the frequency of reset capture to avoid frequent recreating of resources
-	rl := rate.NewLimiter(0.05, 2)
+	rl := rate.NewLimiter(rate.Every(time.Second), 1)
+	//rl := rate.NewLimiter(0.05, 2)
```
overvenus, line 272:
```suggestion
```

cdc/capture/capture.go
```diff
@@ -267,7 +268,8 @@ func (c *captureImpl) reset(ctx context.Context) error {
 func (c *captureImpl) Run(ctx context.Context) error {
 	defer log.Info("the capture routine has exited")
 	// Limit the frequency of reset capture to avoid frequent recreating of resources
-	rl := rate.NewLimiter(0.05, 2)
+	rl := rate.NewLimiter(rate.Every(time.Second), 1)
```
overvenus, line 271: Please also add comments about the frequency.

cdc/capture/capture.go
```diff
@@ -409,17 +418,33 @@ func (c *captureImpl) campaignOwner(ctx cdcContext.Context) error {
...
+		if resp, _ := c.election.Leader(ctx); resp != nil {
+			log.Warn("leader found", zap.Any("resp", resp))
+		} else {
+			log.Warn("leader error", zap.Error(err), zap.Any("resp", resp))
+		}
+
+		// campaign for the ownership just wait for 1 second, if timeout, retry it.
+		campaignCtx, cancel := context.WithTimeout(ctx, 1*time.Second)
+		err = c.campaign(campaignCtx)
+		cancel()
```
overvenus, line 431: Does the `cancel` drop etcd lease?

cdc/capture/capture.go
```diff
@@ -409,17 +417,26 @@ func (c *captureImpl) campaignOwner(ctx cdcContext.Context) error {
...
+				continue
 			case mvcc.ErrCompacted:
 				// the revision we requested is compacted, just retry
 				continue
 			}
-			log.Warn("campaign owner failed", zap.Error(err))
-			// if campaign owner failed, restart capture
+			if strings.Contains(err.Error(), "context deadline exceeded") {
+				continue
+			}
```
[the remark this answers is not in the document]
3AceShowHand: No, it's borrowed from https://github.com/pingcap/tiflow/blob/a4d2b787d34e7917d3992621a6b7c6392e8f8c0a/cdc/sorter/unified/merger.go#L515

I met the error that does not match `context.DeadlineExceeded`, but can be handled by this error string match, it was from gRPC I think.

This logic is removed, since campaign in a timeout way is useless.

cdc/capture/capture.go
```diff
@@ -458,12 +465,24 @@ func (c *captureImpl) campaignOwner(ctx cdcContext.Context) error {
...
+		if err != nil {
+			log.Warn("run owner exited with error",
+				zap.String("captureID", c.info.ID), zap.Int64("ownerRev", ownerRev),
+				zap.Error(err))
+		} else {
+			log.Info("run owner exited normally",
+				zap.String("captureID", c.info.ID), zap.Int64("ownerRev", ownerRev))
+		}
 		// if owner exits, resign the owner key
 		if resignErr := c.resign(ctx); resignErr != nil {
```
overvenus, line 477: Do we need to use a new context in case the `ctx` has cancelled?

cdc/capture/capture.go
```diff
@@ -600,7 +619,17 @@ func (c *captureImpl) Drain(ctx context.Context) <-chan struct{} {
 	const drainInterval = 100 * time.Millisecond
 	done := make(chan struct{})
 	go func() {
-		defer close(done)
+		defer func() {
+			lease := c.session.Lease()
+			_, err := c.EtcdClient.Client.Revoke(ctx, lease)
```
overvenus, line 624: Why not revoke lease in `capture.resign()`?

3AceShowHand: I  think `capture.resign` is called, if the capture win the `campaign`.

For capture which is not the owner, `resign` won't be called.

cdc/capture/capture.go
3AceShowHand: The capture is closed asynchronously, https://github.com/pingcap/tiflow/blob/3405b79aa2fa01076541322a8c92aea288d8c386/cdc/capture/capture.go#L553-L557

It has a chance that the `ctx` be canceled, looks have no negative effect on the `resign` call. 

How about using` context.Background()` to prevent from being canceled?

asddongmen (approved):

cdc/capture/capture.go
```diff
@@ -526,24 +545,24 @@ func (c *captureImpl) GetOwner() (owner.Owner, error) {
...
 	return cerror.WrapError(cerror.ErrCaptureCampaignOwner, c.election.Campaign(ctx, c.info.ID))
 }
 
 // resign lets an owner start a new election.
-func (c *captureImpl) resign(ctx cdcContext.Context) error {
+func (c *captureImpl) resign(ctx context.Context) error {
 	failpoint.Inject("capture-resign-failed", func() {
 		failpoint.Return(errors.New("capture resign failed"))
 	})
 	return cerror.WrapError(cerror.ErrCaptureResignOwner, c.election.Resign(ctx))
```
overvenus, line 560: err := cerror.WrapError(cerror.ErrCaptureResignOwner, c.election.Resign(ctx))
	if err != nil {
		return err
	}
	return lease.Revoke()

cdc/server.go
```diff
@@ -125,10 +125,17 @@ func (s *Server) Run(ctx context.Context) error {
 	logConfig := logutil.DefaultZapLoggerConfig
 	logConfig.Level = zap.NewAtomicLevelAt(zapcore.ErrorLevel)
 
+	// we does not pass a `context` to the etcd client,
+	// to prevent it's cancelled when the server is closed.
```
3AceShowHand, line 129:
```suggestion
	// to prevent it's cancelled when the server is closing.
```

cdc/capture/capture.go
```diff
@@ -526,24 +540,40 @@ func (c *captureImpl) GetOwner() (owner.Owner, error) {
...
 // resign lets an owner start a new election.
-func (c *captureImpl) resign(ctx cdcContext.Context) error {
+func (c *captureImpl) resign(ctx context.Context) error {
 	failpoint.Inject("capture-resign-failed", func() {
 		failpoint.Return(errors.New("capture resign failed"))
 	})
-	return cerror.WrapError(cerror.ErrCaptureResignOwner, c.election.Resign(ctx))
+
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
```
overvenus, line 557: It should be the caller to set the timeout.

overvenus (approved):

PR #6295, opened 2022-07-15
sink(ticdc): implement txn sink and backend interface

hicqu:

Issue Number: ref #5928 

Implement txn sink and backend interface.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/sinkv2/eventsink/txn/backend.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/pingcap/tiflow/cdc/sinkv2/eventsink"
+	"github.com/pingcap/tiflow/pkg/notify"
+)
+
+type backend interface {
+	onTxnEvent(*eventsink.TxnCallbackableEvent) error
+	onNotify() error
+	notifier() *notify.Receiver
```
liuzix, line 24: Why are these methods private?

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,197 @@
...
+
+	"github.com/pingcap/log"
 	"github.com/pingcap/tiflow/cdc/model"
 	"github.com/pingcap/tiflow/cdc/sinkv2/eventsink"
+	"github.com/pingcap/tiflow/pkg/causality"
+	"go.uber.org/zap"
+)
+
+const (
+	conflictDetectorSlots int64 = 1024 * 1024
```
liuzix, line 29: Is it better if we offer a hidden config parameter so that we can tuning the performance on released versions?

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,197 @@
...
 var _ eventsink.EventSink[*model.SingleTableTxn] = (*Sink)(nil)
 
 // Sink is the sink for SingleTableTxn.
-type Sink struct{}
+type Sink struct {
+	conflictDetector *causality.ConflictDetector[*worker, txnEvent]
+	workers          []*worker
+}
+
+func newSink(bes []backend) Sink {
```
liuzix, line 41:
```suggestion
func newSink(backends []backend) Sink {
```
"bes" looks a little bit confusing.

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,197 @@
...
+	txnCh   chan txnWithNotifier
+	stopped chan struct{}
+	wg      sync.WaitGroup
+
+	be backend
+}
+
+type txnWithNotifier struct {
+	txnEvent
+	wantMore func()
```
liuzix, line 98: What is the purpose of `wantMore`?

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,197 @@
...
+}
+
+func newWorker(ID int, be backend) *worker {
+	return &worker{
+		ID:      ID,
+		txnCh:   make(chan txnWithNotifier, 1),
+		stopped: make(chan struct{}, 1),
```
liuzix, line 105:
```suggestion
		stopped: make(chan struct{}),
```

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,197 @@
...
+		be:      be,
+	}
+}
+
+func (w *worker) Add(txn txnEvent, unlock func()) {
+	w.txnCh <- txnWithNotifier{txn, unlock}
+}
+
+func (w *worker) Close() {
+	w.stopped <- struct{}{}
```
liuzix, line 115:
```suggestion
	close(w.stopped)
```

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,197 @@
...
+			log.Panic("crc64 hasher fail")
+		}
+		sums = append(sums, int64(hasher.Sum64()))
+	}
+	return
+}
+
+type worker struct {
+	ID      int
+	txnCh   chan txnWithNotifier
```
liuzix, line 89: Please use an unbounded channel.

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,197 @@
...
+	s.conflictDetector.Close()
+	return nil
+}
+
+type txnEvent struct {
+	*eventsink.TxnCallbackableEvent
+}
+
+// ConflictKeys implements causality.txnEvent interface.
+func (e txnEvent) ConflictKeys() (sums []int64) {
```
liuzix, line 74: Can you cache the results?

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,197 @@
...
+	keys := genTxnKeys(e.TxnCallbackableEvent.Event)
+	sums = make([]int64, 0, len(keys))
+	for _, key := range keys {
+		hasher := crc64.New(crc64.MakeTable(crc64.ISO))
```
liuzix, line 78: Is it necessary to create a new hasher each time? A benchmark case is probably helpful in determining whether the performance is satisfactory.

cdc/sinkv2/eventsink/txn/backend.go
hicqu: All backends will be put in this package, and they won't be used out from the package.

cdc/sinkv2/eventsink/txn/txn_sink.go
hicqu: To notify the conflict detector that transactions blocked by this one can be scheduled.

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,204 @@
...
+func newSink(backends []backend, conflictDetectorSlots int64) sink {
+	workers := make([]*worker, 0, len(backends))
+	for i, backend := range backends {
+		w := newWorker(i, backend)
+		w.wg.Add(1)
+		go w.Run()
+		workers = append(workers, w)
+	}
+	detector := causality.NewConflictDetector[*worker, *txnEvent](workers, conflictDetectorSlots)
+	return sink{detector, workers}
```
liuzix, line 52: I think naming the fields in the initialization can make the code more maintainable.

cdc/sinkv2/eventsink/txn/backend.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/pingcap/tiflow/cdc/sinkv2/eventsink"
+)
+
+// backend indicates a transaction backend like MySQL, TiDB, ...
+type backend interface {
+	// onTxnEvent handles one TxnCallbackableEvent.
+	onTxnEvent(*eventsink.TxnCallbackableEvent) error
+	// timer gets a timer so that onTimeout can be called periodcally.
+	timer() *time.Timer
+	onTimeout() error
```
liuzix, line 28: I feels more nature to me if a `backend` handles timer ticks internally, as different backends may have different uses for timers. For example, a mqSink may need a timer for controlling latency while batching, and a mysql sink may need a timer to split transactions that are too big. Providing a one-size-fits-all solution for using the timer may prevent the evolution of the sink system.

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,204 @@
...
+
+// sink is the sink for SingleTableTxn.
+type sink struct {
+	conflictDetector *causality.ConflictDetector[*worker, *txnEvent]
+	workers          []*worker
+}
 
-// Sink is the sink for SingleTableTxn.
-type Sink struct{}
+func newSink(backends []backend, conflictDetectorSlots int64) sink {
```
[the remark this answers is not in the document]
hicqu: We will add something like `NewSink(opts ...Options)` later.

cdc/sinkv2/eventsink/txn/backend.go
hicqu: > backend handles timer ticks internally

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,204 @@
...
+		select {
+		case <-w.stopped:
+			return
+		case txn := <-w.txnCh.Out():
+			txn.wantMore()
+			if err := w.backend.onTxnEvent(txn.txnEvent.TxnCallbackableEvent); err != nil {
+				// TODO: handle errors.
+				return
+			}
+		case <-w.backend.timer().C:
```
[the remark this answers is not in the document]
hicqu: It returns a pointer. The backend implementation can decide whether to create a new one or not. Generally it should always return one same object.

`Ticker` is not suite here because `backend.onTxnEvent` may flush pending items internally, in which case the timer should be reset.

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,213 @@
...
+	return &worker{
+		ID:      ID,
+		txnCh:   chann.New[txnWithNotifier](chann.Cap(-1 /*unbounded*/)),
+		stopped: make(chan struct{}),
+		backend: backend,
+	}
+}
+
+func (w *worker) runBackground() {
+	go w.Run()
```
liuzix, line 119: I seems `runBackground` will create one goroutine from another.

cdc/sinkv2/eventsink/txn/txn_sink.go
```diff
@@ -14,24 +14,223 @@
...
+}
+
+func (w *worker) Add(txn *txnEvent, unlock func()) {
+	w.txnCh.In() <- txnWithNotifier{txn, unlock}
+}
+
+func (w *worker) Close() {
+	close(w.stopped)
+	w.wg.Wait()
+	w.txnCh.Close()
```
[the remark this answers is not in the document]
hicqu: As we discussed, the worker can only be closed when the processor is closed. So we don't need to consume pending items here.

hi-rustin (approved):
Changes looks good tome, better to add tests for `event.go` and `wroker.go`. No blocking, early approval.

cdc/sinkv2/eventsink/txn/backend.go
```diff
@@ -0,0 +1,32 @@
...
+// backend indicates a transaction backend like MySQL, TiDB, ...
+type backend interface {
+	// onTxnEvent handles one TxnCallbackableEvent.
+	onTxnEvent(*eventsink.TxnCallbackableEvent) (needFlush bool)
+
+	// Flush pending events in the backend.
+	flush() error
+
+	// To reduce latency for low throughput cases.
+	maxFlushInterval() time.Duration
```
liuzix, line 31: These methods are private, indicating that `backend` cannot be implemented by any other package. But I would imagine that you might want to implement different backend in different packages for better encapsulation.

cdc/sinkv2/eventsink/txn/worker.go
```diff
@@ -0,0 +1,95 @@
...
+	go func() {
+		defer w.wg.Done()
+		w.timer = time.NewTimer(w.backend.maxFlushInterval())
+		for {
+			select {
+			case <-w.stopped:
+				return
+			case txn, ok := <-w.txnCh.Out():
+				if !ok {
+					break
```
liuzix, line 70: This only breaks from the select statement. Is this intended?

cdc/sinkv2/eventsink/txn/worker.go
```diff
@@ -0,0 +1,98 @@
...
+		for {
+			select {
+			case <-w.stopped:
+				log.Info("transaction sink backend worker exits as expected",
+					zap.Int("workerID", w.ID))
+				return
+			case txn := <-w.txnCh.Out():
+				txn.wantMore()
+				if w.backend.OnTxnEvent(txn.txnEvent.TxnCallbackableEvent) && w.doFlush() {
+					log.Warn("transaction sink backend exits unexcepted")
```
liuzix, line 75:
```suggestion
					log.Warn("transaction sink backend exits unexceptedly")
```

cdc/sinkv2/eventsink/txn/worker.go
```diff
@@ -0,0 +1,98 @@
...
+					return
+				}
+			case <-w.timer.C:
+				if w.doFlush() {
+					log.Warn("transaction sink backend exits unexcepted")
```
liuzix, line 80:
```suggestion
					log.Warn("transaction sink backend exits unexceptedly")
```

cdc/sinkv2/eventsink/txn/worker.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (w *worker) Close() {
+	close(w.stopped)
+	w.wg.Wait()
+	w.txnCh.Close()
+}
+
+// Run a background loop.
+func (w *worker) Run() {
```
liuzix, line 61: One last suggestion: `Run()` method is usually blocking in the rest of the tiflow repo. It's better to keep it uniform.

liuzix: Or you can hide the method, and call it in the constructor.

liuzix (approved):
Rest LGTM

PR #6306, opened 2022-07-15
master(engine): add log allow list for some rpc methods

### LINKED ISSUE #6293: engine server should not filter log of several APIs

Currently server master has a global rate limiter for log, it was originally designed to avoid flood of logs caused by heartbeat request.

https://github.com/pingcap/tiflow/blob/93435aecf63d5c6bc0f899d2b96c4fe78ec5d109/engine/pkg/rpcutil/server.go#L155-L160

We should filter log sender, making a whitelist for some critical rpc requests, including
- SubmitJob
- CancelJob
- ScheduleTask

_No response_

_No response_

amyangfei:

Issue Number: close #6293

- Add rpc log whitelist

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


engine/pkg/rpcutil/server.go
```diff
@@ -88,6 +88,29 @@ func (l *LeaderClientWithLock[T]) Close() {
 	}
 }
 
+// rpcLimiter is a customed rate limiter, which delegates Allow of rate.Limiter,
```
CharlesCheung96, line 91:
```suggestion
// rpcLimiter is a customized rate limiter, which delegates Allow of rate.Limiter,
```

engine/pkg/rpcutil/server.go
```diff
@@ -88,6 +88,29 @@ func (l *LeaderClientWithLock[T]) Close() {
...
+// and provides an allow list with a higher priority.
+type rpcLimiter struct {
+	limiter   *rate.Limiter
+	allowList []string
```
CharlesCheung96, line 95: Why not use a `map` here?

amyangfei: In fact for small data size, slice search is faster than map search.
Given a sample benchmark: https://gist.github.com/amyangfei/aa9f3edb2785dafb86296b148da56c9d

```go
➜  go test -bench='Benchmark' .
goos: darwin
goarch: arm64
pkg: bench
BenchmarkSliceSearch-8           9815140               115.0 ns/op
BenchmarkMapSearch-8             7893643               151.5 ns/op
PASS
ok      bench   2.687s
```

amyangfei: There are similar tests, https://darkcoding.net/software/go-slice-search-vs-map-lookup/
I think both slice and map are ok here, since it is not a hot path.

CharlesCheung96 (approved):

maxshuang (approved):

PR #6344, opened 2022-07-19
sinkv2(ticdc): make event sink close waiting for all callbacks be called

hi-rustin:

- Make event sink close waiting for all callbacks be called. Otherwise we would have a double-writing problem.
- sinkv2(ticdc): remove useless replicaID

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```


cdc/sinkv2/tablesink/event_table_sink.go
```diff
@@ -103,6 +104,10 @@ func (e *eventTableSink[E]) GetCheckpointTs() model.ResolvedTs {
 
 func (e *eventTableSink[E]) Close() {
 	e.state.Store(pipeline.TableStateStopped)
+	for e.progressTracker.trackingCount() != 0 {
```
hicqu, line 107: Is it necessary? I think just ensure no more events will be flushed into downstreams after `Close` is called is enough.

cdc/sinkv2/tablesink/event_table_sink.go
```diff
@@ -103,6 +104,10 @@ func (e *eventTableSink[E]) GetCheckpointTs() model.ResolvedTs {
...
+		// Sleep for a while to prevent CPU spin.
+		time.Sleep(time.Second)
```
hicqu, line 109: Maybe 1s is too long. How about 100ms?

cdc/sinkv2/tablesink/progress_tracker.go
```diff
@@ -118,3 +118,11 @@ func (r *progressTracker) minTs() model.ResolvedTs {
 
 	return r.lastMinResolvedTs
 }
+
+// trackingCount returns the number of pending events and resolved tss.
+// Notice: must hold the lock.
+func (r *progressTracker) trackingCount() int {
```
hicqu, line 124: We can add a `closed` flag into `progressTracker` to simplify `remove` and make `minTs()` more safe.

hicqu (approved):

asddongmen (approved):

PR #6349, opened 2022-07-19
syncer(dm): handle `;` in SQL comment when load table structure

### LINKED ISSUE #6205: DM should handle escape db&table name from dump files
### What did you do?

create table with escape name. see https://github.com/pingcap/dumpling/issues/131

DM parse db&table as normal

### What did you see instead?

DM parse failed.
https://github.com/pingcap/tiflow/blob/06039a56f58400be5657138159396ee7421f6aa7/dm/pkg/utils/file.go#L141-L163

### Versions of the cluster

```console
```
```console
```
```console
```
```console
(leave TiUP or manually here)
```
```console
>
>
```

```console
```

lance6716:

Issue Number: close #6205

except for `\n` in identifier, we have confidence to treat `;\n` as the end of a statement. We hope user will not use `\n` in identifier.

#### Tests 

 - Integration test

### Release note 

```release-note
`None`.
```


amyangfei (commented):
> We hope user will not use \n in identifier.

Does it mean DM still doesn't support uesr to create a table with name like `t;\n2`. 😂

lance6716:
> > We hope user will not use \n in identifier.
> 
> Does it mean DM still doesn't support uesr to create a table with name like `t;\n2`. 🤔 @lance6716

yes. i guess loader of DM will also not work for this case. not sure about lightning.

dm/tests/downstream_more_column/data/db1.prepare.sql
```diff
@@ -2,6 +2,10 @@ drop database if exists `downstream_more_column`;
 drop database if exists `downstream_more_column1`;
 create database `downstream_more_column1`;
 use `downstream_more_column1`;
-create table t1 (c1 int, c2 int, c3 int, primary key(c1));
+create table t1 (c1 int, c2 int, c3 int comment '1;
```
GMHDBJD, line 5: Can we test with dumpling format, this stmt will be format by ShowCreateTable as
```
CREATE TABLE `t1` (
  `c1` int(11) NOT NULL,
  `c2` int(11) DEFAULT NULL,
  `c3` int(11) DEFAULT NULL COMMENT '1;\n2',
  PRIMARY KEY (`c1`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1
```

PR #6359, opened 2022-07-20
scheduler(ticdc): add changefeed to logs

overvenus:

Issue Number: ref #4757 

Add changefeed fields to scheduler logs.

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
None
```


cdc/scheduler/internal/v3/scheduler_balance.go
```diff
@@ -101,7 +100,8 @@ func buildBalanceMoveTables(
 		}
 	}
 
-	moves := newBalanceMoveTables(random, captures, replications, maxTaskConcurrency)
+	moves := newBalanceMoveTables(
```
lonng, line 103: Is it OK to pass an invalid changefeed ID?

Issue #6105, opened 2022-06-28
unstable test gtid


lance6716:
### Which jobs are flaking?

gtid

### Which test(s) are flaking?

DM-IT

### Jenkins logs or GitHub Actions link

https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/7280/pipeline

```
[2022-06-28T08:19:47.127Z] check diff failed 20-th time, retry later

[2022-06-28T08:19:49.025Z] check diff failed at last
```

```
[2022-06-28T08:19:49.838Z] [2022/06/28 16:19:06.457 +08:00] [ERROR] [subtask.go:377] ["unit process error"] [subtask=test] [unit=Sync] ["error information"="ErrCode:11121 ErrClass:\"functional\" ErrScope:\"internal\" ErrLevel:\"high\" Message:\"fail to restart streamer controller: no relay pos match gtid 05279c58-f6ba-11ec-b2b6-261c1a5f24f8:1-3\" "]
```

gtid test turned off auto-resume

### Anything else we need to know

- Does this test exist for other branches as well?

- Has there been a high frequency of failure lately?

dveeden:
Isn't this the same as #4174 ?

lance6716:
happens again https://ci2.pingcap.net/blue/organizations/jenkins/dm_ghpr_integration_test/detail/dm_ghpr_integration_test/7976/pipeline

Will take a look now

lance6716:
```
[2022-07-21T02:38:30.904Z] [2022/07/21 10:37:47.306 +08:00] [ERROR] [subtask.go:377] ["unit process error"] [subtask=test] [unit=Sync] ["error information"="ErrCode:11121 ErrClass:\"functional\" ErrScope:\"internal\" ErrLevel:\"high\" Message:\"fail to restart streamer controller: no relay pos match gtid 53c6992e-089d-11ed-8027-4e905788d272:1-3\" "]
``` 

caused by we disable auto-resume in this test case.

PR #6404, opened 2022-07-21
test(dm): fix some unstable test

lance6716:

Issue Number: close #6105

- if checking the resulte of resume too quickly, we will see the old error
- in `gtid` test case we manually removed relay log, so when relay log is download, the task may not be able to find the position in checkpoint. We rely on auto-resume to handle it.

#### Tests 

 - No code

### Release note 

```release-note
`None`.
```


buchuitoudegou (approved):

GMHDBJD (approved):
LGTM

PR #6412, opened 2022-07-21
sinkv2(ticdc): impl kafka producer

hi-rustin:

- impl kakfa producer
  - make sure no deadlock
  - make sure no double close panic
  - make sure not sends on closed prodcuer
  - add failpoint 
- add tests

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```


hi-rustin:
/hold

I will add more tests. Not ready for review.

cdc/sinkv2/eventsink/mq/mq_sink.go
```diff
@@ -33,7 +33,7 @@ import (
 )
 
 // newProducerImpl specifies the build method for the producer.
-var newProducerImpl producer.NewProducerFunc = producer.NewMockProducer
+var newProducerImpl producer.NewProducerFunc = producer.NewKafkaProducer
```
liuzix, line 36: Why is it necessary to use a global variable?

cdc/sinkv2/eventsink/mq/mq_sink_test.go
```diff
@@ -54,8 +54,10 @@ func TestWriteEvents(t *testing.T) {
 	t.Parallel()
 
 	kafkav1.NewAdminClientImpl = kafka.NewMockAdminClient
+	newProducerImpl = producer.NewMockProducer
 	defer func() {
 		kafkav1.NewAdminClientImpl = kafka.NewSaramaAdminClient
+		newProducerImpl = producer.NewKafkaProducer
```
liuzix, line 60: This pattern might prevent parallel test running if used in multiple test cases.

cdc/sinkv2/eventsink/mq/kafka_sink.go
```diff
@@ -70,7 +75,17 @@ func NewKafkaSink(
 		return nil, errors.Trace(err)
 	}
 
-	producer := newProducerImpl()
+	client, err := sarama.NewClient(baseConfig.BrokerEndpoints, saramaConfig)
+	if err != nil {
+		return nil, cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
+	}
+
+	log.Info("Try to create a producer",
```
asddongmen, line 83: Why do we need a log here?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,240 @@
...
+	errCh chan error,
+) (Producer, error) {
+	changefeedID := contextutil.ChangefeedIDFromCtx(ctx)
+	role := contextutil.RoleFromCtx(ctx)
+	log.Info("Starting kafka sarama producer ...",
+		zap.String("namespace", changefeedID.Namespace),
+		zap.String("changefeed", changefeedID.ID), zap.Any("role", role))
+
+	asyncProducer, err := sarama.NewAsyncProducerFromClient(client)
+	if err != nil {
```
asddongmen, line 211: we can just:
`return nil, cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)`

asddongmen: I am wrong , nerver mind.

asddongmen (approved):

cdc/sinkv2/eventsink/mq/kafka_sink.go
```diff
@@ -70,7 +75,17 @@ func NewKafkaSink(
...
+		zap.Any("baseConfig", baseConfig), zap.Any("saramaConfig", saramaConfig))
```
amyangfei, line 84: `saramaConfig` contains json unsupported type.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,241 @@
...
+	cerror "github.com/pingcap/tiflow/pkg/errors"
+	"github.com/pingcap/tiflow/pkg/util"
+	"go.uber.org/zap"
+)
+
+var _ Producer = (*kafkaProducer)(nil)
+
+// messageMetaData is used to store the callback function for the message.
+type messageMetaData struct {
+	callback eventsink.CallbackFunc
```
amyangfei, line 38: What logic will be in the callback function?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,241 @@
...
+			zap.String("namespace", k.id.Namespace),
+			zap.String("changefeed", k.id.ID), zap.Any("role", k.role))
+		return err
+	}
+	log.Info("sarama client closed", zap.Duration("duration", time.Since(start)),
+		zap.String("namespace", k.id.Namespace),
+		zap.String("changefeed", k.id.ID), zap.Any("role", k.role))
+
+	start = time.Now()
+	err := k.asyncProducer.Close()
```
amyangfei, line 150: I'm not sure whether `k.client.Close` or `k.asyncProducer.Close()` could block for a long time, if it could, will the hold of `k.closedMu` cause some deadlock

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: Update the table sink progress.

cdc/sinkv2/eventsink/mq/kafka_sink.go
hi-rustin: I don't understand what you mean. Could you please explain it?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: Yes, it will. So, how did the old one solve this problem?

hi-rustin: Perhaps it is acceptable that it causes the AsyncSendMessage to get stuck, since it does not affect the rest of the main goroutine.

cdc/sinkv2/eventsink/mq/kafka_sink.go
amyangfei: Apply the following diff and run case `go test -v -count=1 -run=TestWriteEvents`
```diff
diff --git a/cdc/sinkv2/eventsink/mq/kafka_sink.go b/cdc/sinkv2/eventsink/mq/kafka_sink.go
index 99f95b997..34ee8fee6 100644
--- a/cdc/sinkv2/eventsink/mq/kafka_sink.go
+++ b/cdc/sinkv2/eventsink/mq/kafka_sink.go
@@ -80,7 +80,7 @@ func NewKafkaSink(
                return nil, cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
        }

-       log.Info("Try to create a producer",
+       log.Warn("Try to create a producer",
```

[2022/07/25 14:10:08.975 +08:00] [WARN] [kafka_sink.go:83] ["Try to create a producer"] [baseConfig="{\"BrokerEndpoints\":[\"127.0.0.1:38625\"],\"PartitionNum\":1,\"ReplicationFactor\":1,
\"Version\":\"0.9.0.0\",\"MaxMessageBytes\":1048576,\"Compression\":\"gzip\",\"ClientID\":\"unit-test\",\"EnableTLS\":false,\"Credential\":{\"ca-path\":\"\",\"cert-path\":\"\",\"key-path\
":\"\",\"cert-allowed-cn\":null},\"SASL\":{\"sasl-user\":\"\",\"sasl-password\":\"\",\"sasl-mechanism\":\"\",\"sasl-gssapi\":{\"sasl-gssapi-auth-type\":0,\"sasl-gssapi-keytab-path\":\"\",
\"sasl-gssapi-kerberos-config-path\":\"\",\"sasl-gssapi-service-name\":\"\",\"sasl-gssapi-user\":\"\",\"sasl-gssapi-password\":\"\",\"sasl-gssapi-realm\":\"\",\"sasl-gssapi-disable-pafxfa
st\":false}},\"AutoCreate\":false,\"DialTimeout\":10000000000,\"WriteTimeout\":10000000000,\"ReadTimeout\":10000000000}"] [saramaConfigError="json: unsupported type: func() sarama.SCRAMCl
ient"]

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: I will try to make it asynchronous in here, and I think that way we can solve the problem.

amyangfei: > Yes, it will. So, how did the old one solve this problem?

Seems the old version has the same issue. I'm not sure whether we have met such deadlock problem.

amyangfei: Maybe we can leave enough comments and a TODO here, and observe whether we can meet such problem in the future tests.

cdc/sinkv2/eventsink/mq/kafka_sink.go
amyangfei: Besides running case `TestWriteEvents` alone, I met the following goroutine leak

```golang
--- PASS: TestWriteEvents (1.01s)
[2022/07/25 14:16:51.944 +08:00] [INFO] [worker.go:74] ["flushWorker exited"] [error="context canceled"] [errorVerbose="context canceled\ngithub.com/pingcap/errors.AddStack\n\t/home/apple/go/pkg/mod/github.com/pingcap/errors@v0.11.5-0.20211224045212-9687c2b0f87c/errors.go:174\ngithub.com/pingcap/errors.Trace\n\t/home/apple/go/pkg/mod/github.com/pingcap/errors@v0.11.5-0.20211224045212-9687c2b0f87c/juju_adaptor.go:15\ngithub.com/pingcap/tiflow/cdc/sinkv2/eventsink/mq.(*worker).run\n\t/home/apple/code/tiflow/cdc/sinkv2/eventsink/mq/worker.go:85\ngithub.com/pingcap/tiflow/cdc/sinkv2/eventsink/mq.newSink.func1\n\t/home/apple/code/tiflow/cdc/sinkv2/eventsink/mq/mq_sink.go:89\nruntime.goexit\n\t/home/apple/.g/go/src/runtime/asm_amd64.s:1571"] [namespace=] [changefeed=]
PASS
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 23 in state IO wait, with internal/poll.runtime_pollWait on top of the stack:
goroutine 23 [IO wait]:
internal/poll.runtime_pollWait(0x7f9f7c5f0028, 0x72)
        /home/apple/.g/go/src/runtime/netpoll.go:302 +0x89
internal/poll.(*pollDesc).wait(0xc000288200?, 0xc000556470?, 0x0)
        /home/apple/.g/go/src/internal/poll/fd_poll_runtime.go:83 +0x32
internal/poll.(*pollDesc).waitRead(...)
        /home/apple/.g/go/src/internal/poll/fd_poll_runtime.go:88
internal/poll.(*FD).ReadMsg(0xc000288200, {0xc000556470, 0x10, 0x10}, {0xc00055a020, 0x1000, 0x1000}, 0x16a950f?)
        /home/apple/.g/go/src/internal/poll/fd_unix.go:304 +0x31c
net.(*netFD).readMsg(0xc000288200, {0xc000556470?, 0x7f9f7ce11338?, 0x5?}, {0xc00055a020?, 0x16a81e5?, 0x6fb3800?}, 0xc00056fd08?)
        /home/apple/.g/go/src/net/fd_posix.go:78 +0x37
net.(*UnixConn).readMsg(0xc0004fa8d0, {0xc000556470?, 0xc00056fd80?, 0x16a12e5?}, {0xc00055a020?, 0xb?, 0x7f9f7eb13fff?})
        /home/apple/.g/go/src/net/unixsock_posix.go:115 +0x4f
net.(*UnixConn).ReadMsgUnix(0xc0004fa8d0, {0xc000556470?, 0xc00056fdf8?, 0x16a12e5?}, {0xc00055a020?, 0x0?, 0x0?})
        /home/apple/.g/go/src/net/unixsock.go:143 +0x3c
github.com/godbus/dbus.(*oobReader).Read(0xc00055a000, {0xc000556470?, 0xc00056fe20?, 0x16a1647?})
        /home/apple/go/pkg/mod/github.com/godbus/dbus@v0.0.0-20190726142602-4481cbc300e2/transport_unix.go:21 +0x45
io.ReadAtLeast({0x4c505e0, 0xc00055a000}, {0xc000556470, 0x10, 0x10}, 0x10)
        /home/apple/.g/go/src/io/io.go:331 +0x9a
io.ReadFull(...)
        /home/apple/.g/go/src/io/io.go:350
github.com/godbus/dbus.(*unixTransport).ReadMessage(0xc000128ea0)
        /home/apple/go/pkg/mod/github.com/godbus/dbus@v0.0.0-20190726142602-4481cbc300e2/transport_unix.go:91 +0x11e
github.com/godbus/dbus.(*Conn).inWorker(0xc0001f97a0)
        /home/apple/go/pkg/mod/github.com/godbus/dbus@v0.0.0-20190726142602-4481cbc300e2/conn.go:294 +0x3b
created by github.com/godbus/dbus.(*Conn).Auth
        /home/apple/go/pkg/mod/github.com/godbus/dbus@v0.0.0-20190726142602-4481cbc300e2/auth.go:118 +0x9ed

]
exit status 1
FAIL    github.com/pingcap/tiflow/cdc/sinkv2/eventsink/mq       1.545s
```

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: Please checkout https://github.com/pingcap/tiflow/pull/6412/commits/2b637aab0945afded7aecafcb547da7ffb0d8778.

cdc/sinkv2/eventsink/mq/kafka_sink.go
hi-rustin: Removed. 

amyangfei (approved):

cdc/sinkv2/eventsink/mq/producer/producer.go
```diff
@@ -27,8 +28,9 @@ type Producer interface {
...
 
 	// Close closes the producer and client(s).
-	Close() error
+	Close()
 }
 
-// NewProducerFunc is a function to create a producer.
-type NewProducerFunc func() Producer
+// Creator is a function to create a producer.
+type Creator func(ctx context.Context, client sarama.Client,
```
liuzix, line 35: - I think defining an interface here may be good for code extensibility. We might want the `Creator` to store some states.
- `Factory` can be a better name.

cdc/sinkv2/eventsink/mq/producer/producer.go
```diff
@@ -27,8 +28,9 @@ type Producer interface {
...
+	errCh chan error) (Producer, error)
```
liuzix, line 36: Please add comment on how `errCh` is supposed to be used.
- Who is responsible for closing it?
- What's the expectation for the buffer size?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+) error {
+	// We have to hold the lock to prevent write to closed producer.
+	// Close may be blocked for a long time.
+	k.closedMu.RLock()
+	defer k.closedMu.RUnlock()
+
+	// If the producer is closed, we should skip the message.
+	if k.closed {
+		// We return the context error directly rather than other error.
+		// Because if producer already closed, it means the context maybe is canceling or canceled.
```
liuzix, line 77: How is the statement guaranteed?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+		// So we shouldn't return other errors to the caller to mess up the shutdown process.
+		return ctx.Err()
```
liuzix, line 79: I think returning a special error for `ProducedClosed` may be better.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+		// which could lead to panic.
+		log.Warn("kafka producer already closed",
+			zap.String("namespace", k.id.Namespace),
+			zap.String("changefeed", k.id.ID),
+			zap.Any("role", k.role))
+		return
+	}
+	close(k.failpointCh)
+	// Notify the run loop to exit.
+	close(k.closedChan)
```
liuzix, line 129: This will make `run()` exit, meaning that nobody consumes from `k.asyncProducer.Successes()` and `<-k.asyncProducer.Errors()`. Will this be a problem? Is there a deadlock risk?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+	// Otherwise, we might get stuck with it in an unhealthy state of kafka.
+	go func() {
+		// `client` is mainly used by `asyncProducer` to fetch metadata and other related
+		// operations. When we close the `kafkaSaramaProducer`, TiCDC no need to make sure
+		// that buffered messages flushed.
+		// Consider the situation that the broker does not respond, If the client is not
+		// closed, `asyncProducer.Close()` would waste a mount of time to try flush all messages.
+		// To prevent the scenario mentioned above, close client first.
+		start := time.Now()
+		if err := k.client.Close(); err != nil {
```
liuzix, line 141: - Will `client.Close()` take long?
- Since the closing takes place asynchronously in a goroutine, how do we avoid a zombie problem? After `Close()` has returned, the developer of the caller may want to make sure that no new data will be written to the downstream. How can this be implemented? Is this a risk?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+		case err := <-k.failpointCh:
+			log.Warn("receive from failpoint chan", zap.Error(err),
+				zap.String("namespace", k.id.Namespace),
+				zap.String("changefeed", k.id.ID), zap.Any("role", k.role))
+			return err
+		case ack := <-k.asyncProducer.Successes():
+			if ack != nil {
+				callback := ack.Metadata.(messageMetaData).callback
+				if callback != nil {
+					callback()
```
liuzix, line 184: What's the guarantee for the `callback()`? Some callbacks will never be called. Will this be a problem?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case k.asyncProducer.Input() <- msg:
+	}
+	return nil
+}
+
+func (k *kafkaProducer) Close() {
+	// We have to hold the lock to prevent write to closed producer.
```
liuzix, line 114:
```suggestion
	// We have to hold the lock to synchronize closing with writing.
```

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+	k.closedMu.Lock()
+	defer k.closedMu.Unlock()
+	// If the producer was already closed, we should skip the close operation.
```
liuzix, line 117:
```suggestion
	// If the producer has already been closed, we should skip this close operation.
```

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+			zap.String("changefeed", k.id.ID),
+			zap.Any("role", k.role))
+		return
+	}
+	close(k.failpointCh)
+	// Notify the run loop to exit.
+	close(k.closedChan)
+	k.closed = true
+	// We need to close it asynchronously.
+	// Otherwise, we might get stuck with it in an unhealthy state of kafka.
```
liuzix, line 132:
```suggestion
	// Otherwise, we might get stuck with an unhealthy state of kafka.
```
Please specify what "unhealthy" means.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+	go func() {
+		// `client` is mainly used by `asyncProducer` to fetch metadata and other related
+		// operations. When we close the `kafkaSaramaProducer`, TiCDC no need to make sure
+		// that buffered messages flushed.
```
liuzix, line 136:
```suggestion
		// `client` is mainly used by `asyncProducer` to fetch metadata and perform other related
		// operations. When we close the `kafkaSaramaProducer`, there is no need for TiCDC to make sure
		// that all buffered messages are flushed.
```

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+		// Consider the situation that the broker does not respond, If the client is not
+		// closed, `asyncProducer.Close()` would waste a mount of time to try flush all messages.
```
liuzix, line 138:
```suggestion
		// Consider the situation where the broker is irresponsive. If the client were not
		// closed, `asyncProducer.Close()` would waste a mount of time to try flush all messages.
```
Please use subjunctive mood when discussing contra-factual hypotheses.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+		// To prevent the scenario mentioned above, close client first.
```
liuzix, line 139:
```suggestion
		// To prevent the scenario mentioned above, close the client first.
```

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,243 @@
...
+	partition int32, message *codec.MQMessage,
+) error {
+	// We have to hold the lock to prevent write to closed producer.
+	// Close may be blocked for a long time.
+	k.closedMu.RLock()
+	defer k.closedMu.RUnlock()
+
+	// If the producer is closed, we should skip the message.
+	if k.closed {
+		// We return the context error directly rather than other error.
```
liuzix, line 76:
```suggestion
		// We return the context error directly rather than any other error.
```

cdc/sinkv2/eventsink/mq/producer/producer.go
hi-rustin: 1. I don't know what kind of state we need to store. I think we can keep it for now. It's already expandable.
2. Good Idea. Please checkout https://github.com/pingcap/tiflow/pull/6412/commits/b038e88acdd6691fe3200234539dbcf59185e5c5.

cdc/sinkv2/eventsink/mq/producer/producer.go
hi-rustin: Added. Please checkout [6256b2e](https://github.com/pingcap/tiflow/pull/6412/commits/6256b2e50e3e3b0ff3fc1827448c7655d7e35216)

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin:
```go
// processor.go
	log.Info("processor closing ...",
		zap.String("changefeed", p.changefeed.ID.ID))
	for _, tbl := range p.tables {
		tbl.Cancel()
	}
	for _, tbl := range p.tables {
		tbl.Wait()
	}
	p.cancel()
	p.wg.Wait()

	if p.agent == nil {
		return nil
	}
	if err := p.agent.Close(); err != nil {
		return errors.Trace(err)
	}
	p.agent = nil

	// sink close might be time-consuming, do it the last.
	if p.sink != nil {
```

```go
// owner.go
	c.cancel()
	c.cancel = func() {}
	c.ddlPuller.Close()
	c.schema = nil
	c.cleanupRedoManager(ctx)
	c.cleanupServiceGCSafePoints(ctx)
	canceledCtx, cancel := context.WithCancel(context.Background())
	cancel()
	// We don't need to wait sink Close, pass a canceled context is ok
	if err := c.sink.close(canceledCtx); err != nil {
		log.Warn("Closing sink failed in Owner",
			zap.String("namespace", c.state.ID.Namespace),
			zap.String("changefeed", c.state.ID.ID),
			zap.Error(err))
	}
```

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: I was afraid that returning a non-cancel error would lead to strange behavior, since the original implementation didn't return an error directly.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: I think it's OK.
Because close code looks like:
```go
func (p *asyncProducer) Close() error {
	p.AsyncClose()

	if p.conf.Producer.Return.Successes {
		go withRecover(func() {
			for range p.successes {
			}
		})
	}

	var errors ProducerErrors
	if p.conf.Producer.Return.Errors {
		for event := range p.errors {
			errors = append(errors, event)
		}
	} else {
		<-p.errors
	}

	if len(errors) > 0 {
		return errors
	}
	return nil
}
```

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: 1. possible(maybe about 1-2 mins)
2. Yes, it's risk. But it's acceptable.
 - If the kafka cluster is running well, it will be shut down as soon as possible. Also, we cancel all tablepiples before shutdown, so it's safe.
 - If there is a problem with the kafka cluster, it will shut down the client first, which means no more data will be sent because the connection to the broker is dropped. Also, all our tablepiples will be closed before the shutdown, so it is safe.
 - For mq, duplicate data is acceptable.
 - There is a risk of goroutine leakage, but it is acceptable and our main goal is not to get stuck with the tick.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: If there is never an ack meaning the send fails, we first retry, and if the retry also fails, an error is reported eventually causing the processor to shut down and restart, so it is safe.

cdc/sinkv2/eventsink/mq/kafka_sink.go
```diff
@@ -70,7 +75,17 @@ func NewKafkaSink(
...
 
-	producer := newProducerImpl()
+	client, err := sarama.NewClient(baseConfig.BrokerEndpoints, saramaConfig)
+	if err != nil {
+		return nil, cerror.WrapError(cerror.ErrKafkaNewSaramaProducer, err)
+	}
+
+	log.Info("Try to create a producer",
+		zap.Any("baseConfig", baseConfig))
+	p, err := producerCreator(ctx, client, errCh)
```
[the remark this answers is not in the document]
hi-rustin: Good catch. Please checkout [a50983d](https://github.com/pingcap/tiflow/pull/6412/commits/a50983d4e8dd5eb44fb97d1b865768baf203eed7).

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: Added safety note. Please checkout [1b5cd1f](https://github.com/pingcap/tiflow/pull/6412/commits/1b5cd1ff0d6819c225e16660b319ef7e91a14441)

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
liuzix: I still think returning `ctx.Err()` introduces additional coupling, making error handling more complicated in the long term.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
liuzix: > 1. possible(maybe about 1-2 mins)
> 2. Yes, it's risk. But it's acceptable.
> 
> * If the kafka cluster is running well, it will be shut down as soon as possible. Also, we cancel all tablepiples before shutdown, so it's safe.
> * If there is a problem with the kafka cluster, it will shut down the client first, which means no more data will be sent because the connection to the broker is dropped. Also, all our tablepiples will be closed before the shutdown, so it is safe.
> * For mq, duplicate data is acceptable.
> * There is a risk of goroutine leakage, but it is acceptable and our main goal is not to get stuck with the tick.

- Can we guarantee that no data will be written to the downstream if the table pipelines have been canceled but the Kafka producer is not yet closed?
- Duplication is acceptable only in the sense that the table is never being replicated by more than one TiCDC node. My concern is that before the client on one TiCDC node is completely closed, the table is scheduled to another TiCDC node.
- The interactions with table pipelines you describe seem like very complex coupling. What is a mitigation strategy? Is there an integration test case available?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
liuzix: >  we first retry

Who initiates the retry?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
liuzix: There is a test case for this?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
```diff
@@ -0,0 +1,252 @@
...
+		Partition: partition,
+		Key:       sarama.StringEncoder(message.Key),
+		Value:     sarama.ByteEncoder(message.Value),
+		Metadata:  messageMetaData{callback: message.Callback},
+	}
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case k.asyncProducer.Input() <- msg:
```
liuzix, line 108: Is it possible for `k.asyncProducer.Input()` to be blocked by some error caused by the Kafka cluster?

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: Sarama producer.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: OK, I will remove it.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: > * Can we guarantee that no data will be written to the downstream if the table pipelines have been canceled but the Kafka producer is not yet closed?

I think it is impossible in most cases. 1. if the kafka cluster is normal, we shut down the client first, which will disconnect the connection so no data can be written. 2. if the kafka cluster is not working, the data cannot be written. 3. the kafka cluster suddenly recovers, but we also shut down the client first so no more data will be written.

But the shutdown of the client and the sending of data from the producer are asynchronous. So there is still a possibility that the table that is scheduled to go away is still being written during the process. I'm not quite sure how to fix it more directly. Unless we modify the tick mechanism.



> * The interactions with table pipelines you describe seem like very complex coupling. What is a mitigation strategy? Is there an integration test case available?

We have some tests that inject sink errors that should cover it. But I can't say exactly how it covers these code paths.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: Yes. TestProducerSendMsgFailed should cover it. I was able to get the error it returned in the previous test. Now it is not asserting this value because it is asynchronous.

I meaning I can get error from:
```go
if p.conf.Producer.Return.Errors {
		for event := range p.errors {
			errors = append(errors, event)
		}
	} else {
		<-p.errors
	}

```

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: I think it might be time consuming, but it won't get stuck because after fan-in all the way, it will go to the broker producer's buffer.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: Please checkout https://github.com/pingcap/tiflow/pull/6412/commits/1f7ea637d619cc8dafe0e325ad6bde515da26983

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: > So there is still a possibility that the table that is scheduled to go away is still being written during the process. I'm not quite sure how to fix it more directly. Unless we modify the tick mechanism.

I think the probability of this is very low, because the producer always needs to get the leader through the client before sending a message (in case of failure). So when the cluster recovers, it should also have a hard time sending out messages. But in general it is possible, for example, that the producer finds a leader that has just recovered and is available when the client's shutdown operation has just been executed and the lock has not yet been obtained. In general it is difficult to find such a broker to send messages.

hi-rustin: Offline discussion:

Maybe we need to fix this issue at sarama client. We can drop the retry buffers directly.

liuzix (approved):
LGTM. Remaining issues are not blocking and cannot be resolved without extensive discussion.

cdc/sinkv2/eventsink/mq/producer/kafka_producer.go
hi-rustin: Tracking in https://github.com/pingcap/tiflow/issues/6472

We'll discuss this issue on Wed.

PR #6439, opened 2022-07-22
meta(engine): decouple the model creation and client creation for framework meta

### LINKED ISSUE #6570: orm(engine): decouple the framework model creation and client creation
As the title said.

maxshuang:

Issue Number: close #6570 

1. decouple the model creation and client creation for framework meta
2. initialize the backend meta table when server master becomes a master 

#### Tests 

 - Unit test

### Release note 

```release-note
`None`.
```


engine/pkg/orm/util.go
```diff
@@ -35,6 +41,86 @@ func IsNotFoundError(err error) bool {
...
+		}
+	}
+
+	// check if a gorm.DB
+	if gormDB == nil {
+		gormDB, ok = conn.(*gorm.DB)
+		if !ok {
+			return errors.ErrMetaParamsInvalid.GenWithStackByArgs("client conn is not sql DB")
+		}
+	}
```
amyangfei, line 74: Can extract a function, and reuse it in `InitEpochModel`

engine/pkg/orm/util_test.go
```diff
@@ -14,13 +14,24 @@
...
+
+	metaMock "github.com/pingcap/tiflow/engine/pkg/meta/mock"
+	cerrors "github.com/pingcap/tiflow/pkg/errors"
 )
 
+func init() {
+	// default log level is info, enable debug here for easy debug
+	// log.SetLevel(zapcore.DebugLevel)
+}
+
```
amyangfei, line 34: Not used

engine/servermaster/server.go
```diff
@@ -504,13 +506,29 @@ func (s *Server) registerMetaStore(ctx context.Context) error {
...
+	if cfg.FrameMetaConf.StoreType == metaModel.StoreTypeSQL {
+		// Normally, a schema will be created in advance and we may have no privilege
+		// to create schema for framework meta. Just for easy test here. Ignore any error.
+		ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
+		defer cancel()
+		if err := meta.CreateSchemaIfNotExists(ctx, *(s.cfg.FrameMetaConf)); err != nil {
+			log.Warn("create schema for framework metastore fail, but it can be ignored.",
+				zap.String("schema", s.cfg.FrameMetaConf.Schema),
+				zap.Error(err))
+		}
```
amyangfei, line 518: If this schema doesn't exist and engine has no privilege here, when and where will the error be raised.

engine/pkg/orm/client.go
```diff
@@ -246,29 +175,7 @@ type metaOpsClient struct {
...
-// TODO: need test: change column definition/add column/drop column?
-// TODO: refine the Initialize to decouple the model creation and client
-func (c *metaOpsClient) Initialize(ctx context.Context) error {
-	failpoint.InjectContext(ctx, "initializedDelay", nil)
-	if err := c.db.WithContext(ctx).
-		AutoMigrate(globalModels...); err != nil {
-		return errors.ErrMetaOpFail.Wrap(err)
-	}
+	// DON NOT CLOSE the underlying connection
 	return nil
```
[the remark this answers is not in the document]
maxshuang: 1. orm client is designed to be a thin calculated client. 
The lifecycle of orm client is the same as `server master` now. But after we split client to different modules, the lifecycle of client is determined by modules and the lifecycle of connection is still the same as `server master` or `executor `.
2. Currently, when `server master` exits, the orm client's `Close` will be invoked.

engine/pkg/orm/client_test.go
```diff
@@ -31,13 +31,18 @@ import (
...
+	metaMock "github.com/pingcap/tiflow/engine/pkg/meta/mock"
 	metaModel "github.com/pingcap/tiflow/engine/pkg/meta/model"
 	"github.com/pingcap/tiflow/engine/pkg/orm/model"
 	derror "github.com/pingcap/tiflow/pkg/errors"
 )
 
+func init() {
+	// default log level is info, enable debug here for easy debug
+	// log.SetLevel(zapcore.DebugLevel)
+}
```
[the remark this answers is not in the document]
maxshuang: They may be used for debug in the future. 
Or some people have no idea how to enable trace log for this test.

engine/servermaster/server.go
maxshuang: The error will be raised when creating `framework meta client` and the server-master will exits correspondingly.

engine/pkg/orm/util_test.go
maxshuang: They may be used for debug in the future.
Or some people have no idea how to enable trace log for this test.

engine/pkg/orm/util.go
maxshuang: Fix it.

engine/pkg/meta/model/config.go
```diff
@@ -31,6 +31,9 @@ const (
 	StoreTypeEtcd = "etcd"
 	// StoreTypeSQL is the store type string for SQL
 	StoreTypeSQL = "sql"
+	// StoreTypeSQLite is the store type string for SQLite
+	// NOTE: ONLY for test
+	StoreTypeSQLite = "sqlite"
```
gozssky, line 36: Do we use any feature that is incompatible with MySQL?

maxshuang: I'm not quite sure.
`sqlite` has some tests in  https://github.com/pingcap/tiflow/blob/master/engine/pkg/orm/mock_test.go. It has cover most of our use cases.

maxshuang: But the `sqlite` backend uses `sqlite driver` actually, so It can't verify the `compatibility with MySQL` actually.
I don't have better scheme to use an embeded mysql-compatible db for flexible unit test.

Do you have any suggestion?

gozssky: MySQL and SQLite both provide SQL interfaces. So I guess the `sql` type can contain the `sqlite` type. The only difference is the sql driver we are using. I see a new `sqlite` type added here, so I'm curious if we are using some features that are supported by SQLite but not by MySQL.

engine/pkg/meta/model/interfaces.go
```diff
@@ -84,8 +84,9 @@ type ClientConn interface {
 	ClientType() ClientType
 
 	// GetConn gets the underlying connection object
-	// For storeTypeEtcd, it returns *clientv3.Client for the first return param
-	// For storeTypeSQL, it returns *sql.DB
+	// For StoreTypeEtcd, it returns *clientv3.Client for the first return param
+	// For StoreTypeSQL, it returns *sql.DB
+	// For StoreTypeSQLite, it return *gorm.DB
```
gozssky, line 89: Why `StoreTypeSQL` and `StoreTypeSQLite` returns different object? Both of them provide SQL interfaces.

Since we have used MySQL for metastore. Can we remove the ETCD implementation of this? Thus we can simplify the code.  e.g. Here we can return a concrete type instead of `interface{}`.

engine/pkg/orm/client_test.go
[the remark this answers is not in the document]
gozssky: Maybe we can use the environment variable to enable the debug log. It's unnecessary to set it in every test file.

gozssky (commented):
Almost finished reviewing. I will take a look tomorrow.

engine/pkg/meta/model/config.go
maxshuang: Only for test. 
`sql.DB` is not compatible with `sqlite`, so not all the abstract can be used for `sqlite` and `mysql`.
So I make a new conn type for mock test.

maxshuang: You are right.
I will move the `sqlite` connection intialization to the `ClientConn` layer and reuse the `sql.DB`.
But we still need `sqlite` type to indicate different driver.

engine/pkg/meta/model/interfaces.go
maxshuang: 1. I will adjust the implementation to reuse the `sql.DB`.
2. Prefer to keep the etcd implementation in case that we want to use `etcd` as business metastore.

engine/pkg/orm/client_test.go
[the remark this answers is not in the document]
maxshuang: Ok. I will fix it.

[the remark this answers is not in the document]
maxshuang: Recommend to keep the same form as cdc: https://github.com/pingcap/tiflow/blob/c1c163701622c304adca4863a7cafe0a74c5b853/cdc/sorter/unified/sorter_test.go#L373

If we want to enable debug level, just enable it on the test-case level.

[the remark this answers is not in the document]
maxshuang: Fix it.

engine/pkg/meta/model/config.go
gozssky: If we indeed need `sqlite` type, I suggest renaming `StoreTypeSQL` to `StoreTypeMySQL`, because the semantics of `sql` contains `MySQL` and `SQLite`.

maxshuang: `MySQL` is a too specific DB name, and I think it may confuses user that only `mysql` can be the metastore.

amyangfei (approved):

engine/pkg/meta/model/config.go
gozssky: We might support different databases, like `MySQL`, `PostgreSQL` or `SQLite`. Although they have a similar  `SQL` interface, we still need to know which `SQL` driver we can use. Usually, [`DSN`](https://en.wikipedia.org/wiki/Data_source_name) is used to represent a data source, the driver is placed at the beginning as a scheme.

maxshuang: Done. Has changed `StoreTypeSQL` to `StoreTypeMySQL`.

go.mod
```diff
@@ -108,6 +108,8 @@ require (
 	upper.io/db.v3 v3.7.1+incompatible
 )
 
+require github.com/mattn/go-sqlite3 v2.0.2+incompatible
```
gozssky, line 111: Please move this line to the first `require` group.

gozssky (approved):

go.mod
maxshuang: Fix it.

Issue #3607, opened 2021-11-24
[TiCDC Bug Report] lost a table when rename the table in the upstream


leoppro:
### What did you do?

1. create a `changefeed.toml`:
```
[filter]
rules = ['test.tt']
```
2. create a changefeed with the `changefeed.toml`:
`$ cdc cli changefeed create --sink-uri="blackhole://" --config changefeed.toml`
3. create table `tt` and insert a row in the upstream:
```
MySQL [test]> create table tt ( id int primary key );
Query OK, 0 rows affected (0.11 sec)

MySQL [test]> insert into tt value (1);
Query OK, 1 row affected (0.01 sec)
```
4.  create table `tt_2` and insert some rows in the upstream:
```
MySQL [test]> create table tt_2 ( id int primary key );
Query OK, 0 rows affected (0.10 sec)

MySQL [test]> insert into tt_2 value (1);
Query OK, 1 row affected (0.02 sec)

MySQL [test]> insert into tt_2 value (2);
Query OK, 1 row affected (0.00 sec)
```
5. add a column for table `tt_2` :
```
MySQL [test]> alter table tt_2 add column a int ;
Query OK, 0 rows affected (0.26 sec)
```
6. drop table `tt`:
```
MySQL [test]> drop table tt;
Query OK, 0 rows affected (0.24 sec)
```
7. rename `tt_2` to `tt`:
```
MySQL [test]> rename table tt_2 to tt;
Query OK, 0 rows affected (0.09 sec)
```
8. insert some rows to the new `tt`:
```
MySQL [test]> insert into tt value (3,1);
Query OK, 1 row affected (0.01 sec)

MySQL [test]> insert into tt value (4,1);
Query OK, 1 row affected (0.01 sec)
```

### What did you expect to see?

TiCDC will capture all the changelogs from table `tt`, whether this table has been renamed or not.

### What did you see instead?

The program only captures table `tt` which have not been renamed.

the BlockHole log:
```
# grep BlockHoleSink cdc-*/ticdc.log  | grep -v FlushRowChangedEvents | grep -v Checkpoint
[...]

PR #6446, opened 2022-07-24
ddl_puller(ticdc): discard DDLs that irrelevant with changefeed

asddongmen:

Issue Number: ref #6447 #3607 

After this PR, TiCDC filters out DDLs that are not related to changefeed in `ddl_puller`, which can reduces the latency and memory consumption of replicating DDLs.

It is worth noting that before this PR, for `rename table` DDL, cdc will filter it with the new table name. After this PR, cdc will filter `rename table` by its old table name, which is more in line with the semantics of the filter rule.
```
For example:
----------------------------------
filter:  test.t1(white list)
DDL1:  rename test.t1 to test.t11
DDL2:  rename test.t11 to test.t1
----------------------------------
Before this pr:
CDC will ignore DDL1, sink DDL2.
It will cause upstream and downstream inconsistent.

Result: 
1. Upstream has table 't11',  but downstream still holds table 't1'.
2. When sink DDL2, if downstream don't have the table 't11' before, it may raise error in CDC. 
(This changefeed don't care about the 't11' actually)
----------------------------------
After this pr:
CDC will sink DDL1,  report error DDL2.

Result: 
1. Both upstream and downstream will have table 't11'.
2. CDC reports error for DDL2. More details in following explaination.
(This changefeed don't care about the table 't11' before. User need confirm the error before sinking DDL2)
------------------------------------
```

And, if the user wants to rename a table with an old table name that does not match filter.rule to a new table name that matches filter.rule, then an error will be reported directly. 

For example, we have `filter.rule = ['test.t1']`, and there are tables name `test.t11` in the upstream TiDB:
1. excute `rename table 'test.t11' to 'test.t1'` will report an error.

In addition, if you rename multiple tables in a single one `rename table` DDL, TiCDC will apply the rule mention above to every single subordinate `rename table` DDL jobs.

For example, we have `filter.rule = ['test.t1', `test.t3`]`, and there are tables name `test.t1`, `test.t2` in the upstream TiDB:
1. excute `rename table 'test.t1' to 'test.t11', 'test.t2' to 'test.t22'`, the `rename table 'test.t1'` will be replicated and 'test.t2' to 'test.t22'` will be filtered out since `test.t1` match `filter.rule` but `test.t2` does not. 
3. excuet `rename table 'test.t2' to 'test.t3', 'test.t1' to 'test.t11'` will report an error, since `test.t2` does not match `filter.rule` but `test.t3` does, TiCDC do not allow such behavior. 

#### Tests 

- [x] Unit test
- [x] Integration test

1. start cdc server 
4. create a chengfeed with `filter.rule = ['test1.t1', 'test1.t2', 'test1.t3', 'test1.t4'] `
5. create t1, t2, t3, t4, t5, t6 in upstream tidb.
6. exec : `rename table t1 to t11` (replicated)
7. exec : `rename table t11 to t1` (error)
```shell
➜  bin git:(filter_ddl) ✗ ./cdc cli changefeed list
[
  {
    "id": "test-1",
    "namespace": "default",
    "summary": {
      "state": "error",
      "tso": 435389327982985217,
      "checkpoint": "2022-08-19 11:07:58.939",
      "error": {
        "addr": "127.0.0.1:8300",
        "code": "CDC:ErrSyncRenameTablesFailed",
        "message": "[CDC:ErrSyncRenameTablesFailed]table's old name is not in filter rule, and its new name in filter rule table id '108', ddl query: [rename table t11 to t1], it's an unexpected behavior, if you want to replicate this table, please add its old name to filter rule."
      }
    }
  }
]
```
8. exec: `rename table t2 to t22, t3 to t33` (replicated)
9. exec: `rename table t22 to t222, t33 to t333` (discard)
10. exec: `rename table  t222 to t2, t333 to t3` (error)
```shell
➜  bin git:(filter_ddl) ✗ ./cdc cli changefeed list
[
  {
    "id": "test-1",
    "namespace": "default",
    "summary": {
      "state": "error",
      "tso": 435389397673967617,
      "checkpoint": "2022-08-19 11:12:24.789",
      "error": {
        "addr": "127.0.0.1:8300",
        "code": "CDC:ErrSyncRenameTablesFailed",
        "message": "[CDC:ErrSyncRenameTablesFailed]table's old name is not in filter rule, and its new name in filter rule table id '110', ddl query: [rename table t222 to t2, t333 to t3], it's an unexpected behavior, if you want to replicate this table, please add its old name to filter rule."
      }
    }
  }
]
```
11. exec: `rename table t4 to t44, t5 to t55`
It will be splited into two parts, the `rename table t4 to t44` part is replicated and `t5 to t55` is ignored.

You can learn more discussion detail in this internal doc: https://pingcap.feishu.cn/wiki/wikcnJ6SaB5cGLeG5OrM4nf6MPg

### Release note 

```release-note
None
```


asddongmen:
I will fix the failed unit test cases ASAP and add some unit test cases to cover this change.

overvenus (commented):
What about rename table DDL? Rename a table to a name that matches filter.

asddongmen:
> What about rename table DDL? Rename a table to a name that matches filter.

1. If we create a changefeed with config like:
```toml
[filter]
rule = ["test.t1", "test.t2"]
```
2. All ddls about `test.t3` or `test.t4` etc. will be filter out in `ddl_puller`.
3. Then there will only be `t1` and `t2` existed in TiCDC `schemaStorage`.
4. So I think it is reasonable to filter out `test.t3` in a rename tables ddl like:
```sql
rename table `test.t1` to `test.t11`, `test.t2` to `test.t22`, `test.t3` to `test.t33`
```

cdc/processor/processor.go
```diff
@@ -775,6 +776,9 @@ func (p *processor) createAndDriveSchemaStorage(ctx cdcContext.Context) (entry.S
 			case <-ctx.Done():
 				return
 			case jobEntry = <-ddlPuller.Output():
+				if jobEntry.Job != nil {
+					log.Info("fizz:processor get ddl job", zap.Any("job", jobEntry.Job.Query))
```
maxshuang, line 780: What does it mean `fizz:` ?

pkg/filter/filter.go
```diff
@@ -125,6 +125,9 @@ func (f *filter) ShouldIgnoreDDLEvent(ddl *model.DDLEvent) (bool, error) {
 	case timodel.ActionCreateSchema, timodel.ActionDropSchema,
 		timodel.ActionModifySchemaCharsetAndCollate:
 		shouldIgnoreTableOrSchema = !f.tableFilter.MatchSchema(ddl.TableInfo.Schema)
+	case timodel.ActionRenameTable:
```
maxshuang, line 128: Would it be a problem for following scenary:
```
database:
we only have test.t2 on both upstream and downstream cluster.

filter:
[test.t1] // expect only sync the data of test.t1

Run DDL in upstream cluster:
rename test.t2 to test.t1
```
We only check the `PreTableInfo` here, so the downstream has no table `test.t1`.
Will it happen?

cdc/puller/ddl_puller.go
```diff
@@ -142,22 +193,201 @@ func (p *ddlJobPullerImpl) initJobTableMeta() error {
...
+		preTableInfo, ok := p.schemaSnapshot.PhysicalTableByID(tableInfo.ID)
+		if !ok {
+			// If we can not find a table by its id in schemaSnapshot,
+			// it means we didn't replicate the table before, skip it.
+			log.Info("table not found", zap.Int64("tableID", tableInfo.ID),
+				zap.String("tableName", tableInfo.Name.O))
+			continue
+		}
+
+		oldSchemaName := oldSchemaNames[i].L // fizz: 因为 filter 可能是大小写敏感的，这里需要测一下
```
maxshuang, line 255: Need remove.

pkg/filter/filter.go
```diff
@@ -32,7 +32,7 @@ type Filter interface {
 	// ShouldDiscardDDL returns true if this DDL should be discarded.
 	// If a ddl is discarded, it will neither be applied to cdc's schema storage
 	// nor sent to downstream.
-	ShouldDiscardDDL(ddlType timodel.ActionType) bool
+	ShouldDiscardDDL(ddlType timodel.ActionType, schema, table string) bool
```
maxshuang, line 35: `ShouldDiscardDDL` and `ShouldIgnoreDDLEvent` confuses me.
What's the difference?

cdc/puller/ddl_puller.go
```diff
@@ -142,22 +193,201 @@ func (p *ddlJobPullerImpl) initJobTableMeta() error {
...
+		remainNewTableNames, remainOldSchemaNames                 []*timodel.CIStr
+	)
+
+	err = job.DecodeArgs(&oldSchemaIDs, &newSchemaIDs,
+		&newTableNames, &oldTableIDs, &oldSchemaNames)
+	if err != nil {
+		return true, errors.Trace(err)
+	}
+
+	multiTableInfos := job.BinlogInfo.MultipleTableInfos
```
maxshuang, line 235: Seems that the `owner/schema.go` also has the similar logic, can we merge them?

cdc/puller/ddl_puller.go
```diff
@@ -142,22 +193,201 @@ func (p *ddlJobPullerImpl) initJobTableMeta() error {
...
+			log.Info("table not found", zap.Int64("tableID", tableInfo.ID),
+				zap.String("tableName", tableInfo.Name.O))
+			continue
+		}
+
+		oldSchemaName := oldSchemaNames[i].L // fizz: 因为 filter 可能是大小写敏感的，这里需要测一下
+		oldTableName := preTableInfo.Name.O
+
+		// We only filter out the rename tables by its old schema name and old table name.
+		if p.filter.ShouldDiscardDDL(job.Type, oldSchemaName, oldTableName) {
```
maxshuang, line 259: Same question as in `filter.go`.

cdc/puller/ddl_puller_test.go
```diff
@@ -108,28 +114,144 @@ func (m *mockPuller) appendResolvedTs(ts model.Ts) {
...
+		mockPuller.appendDDL(job)
+		mockPuller.appendResolvedTs(job.BinlogInfo.FinishedTS + 1)
+		waitResolvedTs(t, ddlJobPuller, job.BinlogInfo.FinishedTS+1)
+
+		job = helper.DDL2Job("create table test1.t3(id int)")
+		mockPuller.appendDDL(job)
+		mockPuller.appendResolvedTs(job.BinlogInfo.FinishedTS + 1)
+		waitResolvedTs(t, ddlJobPuller, job.BinlogInfo.FinishedTS+1)
+
+		job = helper.DDL2Job("rename table test1.t1 to test1.t11, test1.t2 to test1.t22, test1.t3 to test1.t33")
```
maxshuang, line 187: Add test for `rename table test1.t33 to test1.t1`.

cdc/puller/ddl_puller.go
asddongmen: thanks, I will do this in next commit.

cdc/processor/processor.go
asddongmen: only for debug, I will remove it later.

pkg/filter/filter.go
asddongmen: yep, it will happen. 
So I think this behavior need to discuss with PM.

maxshuang:
Please add some integration tests before merging this pr.
Can refer to: https://github.com/pingcap/tiflow/tree/release-6.0/tests/integration_tests/rename_tables

pkg/filter/filter.go
asddongmen: `Discard` means we discard a DDL in puller, a discarded DDL will not enter other logic of TiCDC, `Ignore` means we ignore a DDL before we sink it, a ignored DDL will be applied to `schemaStorage`.

cdc/puller/ddl_puller.go
asddongmen: I have tried, but it is hard to do, these places split a `renameTables` job for different purposes.

cdc/puller/ddl_puller_test.go
asddongmen: ok

maxshuang:
> Please add some integration tests before merging this pr. Can refer to: https://github.com/pingcap/tiflow/tree/release-6.0/tests/integration_tests/rename_tables

Have been merged to this test on master:
https://github.com/pingcap/tiflow/blob/master/tests/integration_tests/multi_tables_ddl/data/test.sql

pkg/filter/filter.go
maxshuang: If we have few `rename` DDL on production environment, may be we can get the snapshot back and find the table info  
of `t2`.

pkg/filter/filter.go
maxshuang: Seems that in logically, they only have different input params.
The whole ticdc don't care about the the ignored or discarded DDL.

maxshuang:
> ### What problem does this PR solve?
> Issue Number: ref #6447
> 
> ### What is changed and how it works?
> After this PR, TiCDC filters out DDLs that are not related to changefeed in `ddl_puller`, which can reduces the latency and memory consumption of replicating DDLs.
> 
> ### Check List
> #### Tests
> * [x]  Unit test
> * [x]  Integration test
> * [x]   Manual test (add detailed scripts or steps below)
> 
> 1. start cdc server
> 2. create a chengfeed with `filter.rule = ['test1.t1', 'test1.t2', 'test1.t3', 'test1.t4'] `
> 3. create t1, t2, t3, t4, t5, t6 in upstream tidb.
> 4. exec : `rename table t1 to t11` (replicated)
> 5. exec : `rename table t11 to t1` (dicarded)
> 6. exec: `rename table t2 to t22, t3 to t33` (replicated)
> 7. exec: `rename table  t22 to t2, t33 to t3` (discared)
> 8. exec: `rename table t4 to t44, t5 to t55` (error)
> 
> ```shell
> ➜  bin git:(filter_ddl) ./cdc cli changefeed list
> [
>   {
>     "id": "test-1",
>     "namespace": "default",
>     "summary": {
>       "state": "error",
>       "tso": 435306114673541122,
>       "checkpoint": "2022-08-15 18:57:25.355",
>       "error": {
>         "addr": "127.0.0.1:8300",
>         "code": "CDC:ErrRenameTablesTableNotFound",
>         "message": "[CDC:ErrRenameTablesTableNotFound]failed to find table id(s): '[117]' in cdc schema snapshot, ddl query: [rename table t4 to t44, t5 to t55],if you want to replicate these table(s), please add them to filter rule."
>       }
>     }
>   }
> ]
> ```
> 
> #### Questions
> ##### Will it cause performance regression or break compatibility?
> ##### Do you need to update user documentation, design documentation or monitoring documentation?
> ### Release note
> ```
> None
> ```

I think we need to do more explaination:
1. Declair the principle and reason in this pr, instead  of the final behavior. Show the comparation before and after this pr.
2. Some `DDL` behavior will impact the `DML`, so the behavior of `DDL` is not enough.
 e.g.  when we discard `rename table t11 to t1`, any `update t1 ` DML will fail, so we need to show the `DML` behavior correspondingly.
3. Add the case `Rename t1 to t3`. Show all possible behaviors.
4. Why case 8 error `rename table t4 to t44, t5 to t55`? 
     `t4 to t44` means we no longer care about `t4`, so we can sink to downstream and remove inner relative schema.
     `t5 to t55`, we can simply ignore both since we only care about `t1` to `t4`.
     So why user get an unexpected error?
5. Do we have any plan to add explaination on official docs and the action to fix the error ?

cdc/puller/ddl_puller.go
```diff
@@ -142,22 +181,198 @@ func (p *ddlJobPullerImpl) initJobTableMeta() error {
...
+// handleRenameTables gets all the tables that are renamed
+// in the DDL job out and filter them one by one,
+// if all the tables are filtered, skip it.
+func (p *ddlJobPullerImpl) handleRenameTables(job *timodel.Job) (skip bool, err error) {
+	var (
+		oldSchemaIDs, newSchemaIDs, oldTableIDs []int64
+		newTableNames, oldSchemaNames           []*timodel.CIStr
+	)
+
+	// var (
```
zhaoxinyu, line 211: Should these commented lines be removed?

cdc/puller/ddl_puller.go
```diff
@@ -142,22 +181,198 @@ func (p *ddlJobPullerImpl) initJobTableMeta() error {
...
+		if p.filter.ShouldDiscardDDL(job.Type, oldSchemaNames[i].O, preTableInfo.Name.O) {
+			log.Info("table is filtered",
+				zap.Int64("tableID", tableInfo.ID),
+				zap.String("schema", oldSchemaNames[i].O),
+				zap.String("table", preTableInfo.Name.O))
+			skipTablesCount++
+			continue
+		}
+
+		// newMultiTableInfos = append(newMultiTableInfos, tableInfo)
```
zhaoxinyu, line 251: ditto

cdc/puller/ddl_puller.go
```diff
@@ -142,22 +181,198 @@ func (p *ddlJobPullerImpl) initJobTableMeta() error {
...
+	default:
+		// nil means it is a schema ddl job, it's no need to fill the table name.
+		if job.BinlogInfo.TableInfo != nil {
+			job.TableName = job.BinlogInfo.TableInfo.Name.O
+		}
+		skip = p.filter.ShouldDiscardDDL(job.Type, job.SchemaName, job.TableName)
+	}
+
+	if skip {
+		log.Info("ddl job schema or table is not match, discard it",
```
zhaoxinyu, line 352:
```suggestion
		log.Info("ddl job schema or table does not match, discard it",
```

cdc/puller/ddl_puller.go
zhaoxinyu: Maybe we can extract the basic logic in `handleRenameTables()` here and `schemaWrap4Owner.parseRenameTables` to a public place?

cdc/puller/ddl_puller.go
```diff
@@ -186,11 +403,13 @@ func findColumnByName(cols []*timodel.ColumnInfo, name string) (*timodel.ColumnI
...
-
-	return nil, cerror.WrapError(cerror.ErrDDLSchemaNotFound, errors.Errorf("can't find column %s", name))
+	return nil, cerror.WrapError(
+		cerror.ErrDDLSchemaNotFound,
+		errors.Errorf("can't find column %s", name))
 }
 
-// NewDDLJobPuller creates a new NewDDLJobPuller, which fetches events starting event start from checkpointTs.
+// NewDDLJobPuller creates a new NewDDLJobPuller,
+// which fetches ddl events start from checkpointTs.
```
zhaoxinyu, line 412:
```suggestion
// which fetches ddl events starting from checkpointTs.
```

zhaoxinyu:
I think it would be better to add a configuration item in `sink-uri` or changefeed toml configuration file to enable/disable discarding DDLs.

errors.toml
```diff
@@ -941,6 +941,11 @@ error = '''
 regions not completely left cover span, span %v regions: %v
 '''
 
+["CDC:ErrRenameTablesTableNotFound"]
+error = '''
+failed to find table id(s): '%v' in cdc schema snapshot, ddl query: [%s], if you want to replicate these table(s), please add them to filter rule.
```
maxshuang, line 946: I think this error message is not precise. 
We fail for case 8 because the atomicity of `rename tables`,  but not the `can't find table schema`. Or we can't explain why  DDL `rename t5 to t55` will not raise an error.

tests/integration_tests/multi_tables_ddl/data/test.sql
```diff
@@ -66,4 +108,16 @@ insert into t3 select * from t1_7;
 insert into t4 select * from t2_7;
 drop table t3, t4;
 
+/* replicate successful */
+rename table t9 to t99;
```
maxshuang, line 112: Please add all possible cases to this test as we can since they are limited, like:
`rename t1 to t3`
`rename t44 to t4, t5 to t55`

maxshuang:
Rest LGTM.

cdc/puller/ddl_puller.go
```diff
@@ -142,22 +181,182 @@ func (p *ddlJobPullerImpl) initJobTableMeta() error {
...
+	}
+
+	switch job.Type {
+	case timodel.ActionRenameTables:
+		skip, err = p.handleRenameTables(job)
+		if err != nil {
+			return true, errors.Trace(err)
+		}
+	case timodel.ActionRenameTable:
+		// schema, ok := p.schemaSnapshot.SchemaByID(job.SchemaID)
```
zhaoxinyu, line 304: Remove these commented lines.

cdc/puller/ddl_puller_test.go
```diff
@@ -108,28 +113,394 @@ func (m *mockPuller) appendResolvedTs(ts model.Ts) {
...
+	ddlJobPullerImpl := ddlJobPuller.(*ddlJobPullerImpl)
+	cfg := config.GetDefaultReplicaConfig()
+	cfg.Filter.Rules = []string{
+		"test1.t1",
+		"test1.t2",
+	}
+	f, err := filter.NewFilter(cfg, "")
+	require.NoError(t, err)
+	ddlJobPullerImpl.filter = f
+	// ctx, cancel := context.WithCancel(context.Background())
```
zhaoxinyu, line 342: Remove these commented lines.

tests/integration_tests/multi_tables_ddl/conf/diff_config.toml
```diff
@@ -6,14 +6,17 @@ export-fix-sql = true
 
 check-struct-only = false
 
+
```
zhaoxinyu, line 9: remove these two redundant blank lines.

zhaoxinyu (approved):

maxshuang (approved):

PR #6454, opened 2022-07-25
version(cdc): allow version inconsistant since v6.3.0

### LINKED ISSUE #6450: relax version consistency check to support graceful upgrade
At the moment, multiple running CDC instances should have the same semantic version to make the whole CDC cluster make progress, this disallows CDC cluster to support graceful upgrade.

To support graceful upgrade, we have to relax the CDC version check strategy

* since v6.2.0, the owner node allows other higher version CDC instances in the cluster

3AceShowHand:

Issue Number: close #6450, ref #4757

Before the change, The CDC owner does not allow CDC instances with different versions in the cluster, the whole cluster cannot make any progress, and this prevented CDC from supporting the graceful rolling upgrade.

* TiCDC cluster allows at the most 2 different versions running instances in the cluster, and at least  `v6.3.0`
* p2p server and client does not check version
* PD / TiKV version range in [5.1.0-alpha, 8.0.0), for cdc version `x.y.z`, the upper open bound for PD / TiKV is `x+2.0.0`

### Release note 

```release-note
`None`.
```


pkg/p2p/server.go
```diff
@@ -832,26 +831,6 @@ func (m *MessageServer) verifyStreamMeta(streamMeta *p2p.StreamMeta) error {
 		)
 	}
 
-	if m.config.ServerVersion == "" || streamMeta.ClientVersion == "" {
```
overvenus, line 835: We should add the version check here too.

pkg/context/context.go
```diff
@@ -189,7 +188,8 @@ func NewContext4Test(baseCtx context.Context, withChangefeedVars bool) Context {
 		CaptureInfo: &model.CaptureInfo{
 			ID:            "capture-id-test",
 			AdvertiseAddr: "127.0.0.1:0000",
-			Version:       version.ReleaseVersion,
+			// suppose the current version is `v6.3.0`
+			Version: "v6.3.0",
```
overvenus, line 192: Why not `version.ReleaseVersion`?

cdc/owner/owner.go
```diff
@@ -402,17 +378,20 @@ func (o *ownerImpl) updateMetrics(state *orchestrator.GlobalReactorState) {
...
-					zap.Reflect("capture", capture), zap.String("ownerVer", myVersion))
-			}
-			return false
+		versions[capture.Version] = struct{}{}
+	}
+
+	ok := version.CheckTiCDCVersion(versions)
+	if !ok {
+		if o.logLimiter.Allow() {
+			log.Warn("more than two versions instance in the cdc cluster, it's not allowed",
```
overvenus, line 389:
```suggestion
			log.Warn("more than two versions instance in the cdc cluster, it's not allowed",
```
I think it's ambiguous. Does it mean there are more two TiCDC instances have different versions, eg, capture1(v6.0.0), capture2(v6.1.0) and capture3(v6.2.0)? Or does it mean major versions different, eg, v6.3.0 and v8.0.0?

pkg/context/context.go
3AceShowHand: it's `None`, cannot perform the ut easily.

pkg/p2p/server.go
3AceShowHand: `server` and `client` version is already guaranteed by the `owner` in the first place, we can remove it simply since this won't be hit I think.

overvenus: What about rolling upgrade? I think there are some CDC using old protocol while others using new protocol.

pkg/version/check_test.go
```diff
@@ -341,7 +355,35 @@ func TestCheckPDVersionError(t *testing.T) {
...
 		"[CDC:ErrCheckClusterVersionFromPD]failed to request PD 500 Internal Server Error , please try again later",
 	)
 }
+
+func TestCheckTiCDCVersion(t *testing.T) {
+	t.Parallel()
+
+	versions := make(map[string]struct{})
+
+	versions["v6.3.0"] = struct{}{}
```
overvenus, line 368: Please add some comments to each tests.

pkg/version/check_test.go
```diff
@@ -341,7 +355,35 @@ func TestCheckPDVersionError(t *testing.T) {
...
 		w.WriteHeader(http.StatusInternalServerError)
 	}
-	require.Contains(t, CheckPDVersion(context.TODO(), ts.URL, nil).Error(),
+	require.Contains(t, checkPDVersion(context.TODO(), ts.URL, nil).Error(),
 		"[CDC:ErrCheckClusterVersionFromPD]failed to request PD 500 Internal Server Error , please try again later",
 	)
 }
+
+func TestCheckTiCDCVersion(t *testing.T) {
+	t.Parallel()
```
hi-rustin, line 364: I think we can use a map to organize the test cases.

For example:

map[case-name]struct{}

pkg/version/check.go
```diff
@@ -218,13 +242,13 @@ func (v *TiCDCClusterVersion) ShouldRunCliWithOpenAPI() bool {
 	return !v.LessThan(*semver.New("6.2.0")) || (v.Major == 6 && v.Minor == 2 && v.Patch == 0)
 }
 
-// TiCDCClusterVersionUnknown is a read-only variable to represent the unknown cluster version
-var TiCDCClusterVersionUnknown = TiCDCClusterVersion{}
+// cdcClusterVersionUnknown is a read-only variable to represent the unknown cluster version
+var cdcClusterVersionUnknown = TiCDCClusterVersion{}
```
hi-rustin, line 246: Maybe we can keep the old one. Because TiKV also has a CDC component.

cdc/owner/changefeed.go
```diff
@@ -153,10 +153,12 @@ func newChangefeed4Test(
 		changefeed model.ChangeFeedID,
 	) (puller.DDLPuller, error),
 	newSink func() DDLSink,
+	newScheduler func(ctx cdcContext.Context, startTs uint64) (scheduler.Scheduler, error),
```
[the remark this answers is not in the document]
3AceShowHand: we can do it, but let's do it in other refactor PR.

overvenus (approved):

PR #6508, opened 2022-07-28
engine(dm): implement update config

GMHDBJD:

Issue Number: ref #5504 

1. add ModRevision in JobCfg
2. worker update status with CfgModRevison
4. increment ModRevision when update-job-cfg
5. put new config to metadata
6. checkpoint agent create checkpoint for new config
7. worker manager stop worker for old version config
8. worker manager start worker for new version config

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


lance6716 (commented):
switch back to fix CI for my PR 😄

engine/jobmaster/dm/api.go
```diff
@@ -147,6 +161,21 @@ func (jm *JobMaster) GetJobCfg(ctx context.Context) (*config.JobCfg, error) {
...
 
+// UpdateJobCfg updates job config.
+func (jm *JobMaster) UpdateJobCfg(ctx context.Context, cfg *config.JobCfg) error {
+	if err := jm.preCheck(ctx, cfg); err != nil {
+		return err
+	}
+	if err := jm.OperateTask(ctx, dmpkg.Update, cfg, nil); err != nil {
+		return err
+	}
+	if err := jm.checkpointAgent.Update(ctx, cfg); err != nil {
```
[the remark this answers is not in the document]
GMHDBJD: for a full mode task to an incremental mode task. https://github.com/pingcap/tiflow/blob/45e167d76a49e7b1e0f0f88634572498a5db7b16/engine/test/e2e/e2e_dm_test.go#L190
Also support change `dm_meta` or task name.

engine/jobmaster/dm/worker_manager.go
```diff
@@ -176,11 +187,33 @@ func (wm *WorkerManager) stopUnneededWorkers(ctx context.Context, job *metadata.
...
 }
 
+// stop outdated workers, usually happened when update job cfgs.
+func (wm *WorkerManager) stopOutdatedWorkers(ctx context.Context, job *metadata.Job) error {
+	var recordError error
+	wm.workerStatusMap.Range(func(key, value interface{}) bool {
+		taskID := key.(string)
+		workerStatus := value.(runtime.WorkerStatus)
+		task, ok := job.Tasks[taskID]
+		if !ok || task.Cfg.ModRevision == workerStatus.CfgModRevision {
```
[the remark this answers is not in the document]
GMHDBJD: This case is handled by `stopUneededWorker`

engine/jobmaster/dm/metadata/job.go
```diff
@@ -127,3 +127,30 @@ func (jobStore *JobStore) UpdateStages(ctx context.Context, taskIDs []string, st
...
+
+// UpdateConfig will be called if user update job config.
+func (jobStore *JobStore) UpdateConfig(ctx context.Context, jobCfg *config.JobCfg) error {
+	state, err := jobStore.Get(ctx)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	oldJob := state.(*Job)
+
+	// TODO: each task can have different version.
```
lance6716, line 139: the purpose of this TODO is to reduce worker shutdown and restart? need to leverage pros and cons

engine/jobmaster/dm/runtime/worker_status.go
```diff
@@ -89,17 +95,18 @@ func (w *WorkerStatus) RunAsExpected() bool {
...
 func InitWorkerStatus(taskID string, unit framework.WorkerType, id frameModel.WorkerID) WorkerStatus {
-	workerStatus := NewWorkerStatus(taskID, unit, id, WorkerCreating)
+	workerStatus := NewWorkerStatus(taskID, unit, id, WorkerCreating, 0)
 	workerStatus.createdTime = time.Now()
 	return workerStatus
 }
 
 // NewWorkerStatus creates a new WorkerStatus instance
-func NewWorkerStatus(taskID string, unit framework.WorkerType, id frameModel.WorkerID, stage WorkerStage) WorkerStatus {
+func NewWorkerStatus(taskID string, unit framework.WorkerType, id frameModel.WorkerID, stage WorkerStage, cfgModRevision uint64) WorkerStatus {
```
lance6716, line 104: maybe we can move taskStatus.Task, taskStatus.Unit, taskStatus.CfgModRevision into one parameter taskStatus

OK to do it in future

lance6716 (commented):
11 / 25 files viewed

engine/jobmaster/dm/api.go
```diff
@@ -55,14 +58,21 @@ func (jm *JobMaster) QueryJobStatus(ctx context.Context, tasks []string) (*JobSt
...
 		workerStatusMap = jm.workerManager.WorkerStatus()
 		wg              sync.WaitGroup
 		mu              sync.Mutex
 		jobStatus       = &JobStatus{
-			JobMasterID: jm.JobMasterID(),
-			WorkerID:    jm.ID(),
-			TaskStatus:  make(map[string]TaskStatus),
+			JobMasterID:           jm.JobMasterID(),
+			WorkerID:              jm.ID(),
+			ExpectedCfgModRevison: cfgModeRevsion,
```
lance6716, line 74: just a remind that this is conflict with the TODO "each task has its separate `cfgModeRevsion`",

engine/jobmaster/dm/api.go
```diff
@@ -102,6 +115,7 @@ func (jm *JobMaster) QueryJobStatus(ctx context.Context, tasks []string) (*JobSt
 				ExpectedStage: expectedStage,
 				WorkerID:      workerID,
 				Status:        queryStatusResp,
+				CfgModRevison: cfgModRevision,
```
lance6716, line 118: Now we added ExpectedCfgModRevison and CfgModRevison, the OpenAPI client needs to parse and check their equivalent if it should ignore some task status. Can we add a "outdated stage" or change the queryStatusResp to indicate it's outdated?

engine/jobmaster/dm/api.go
```diff
@@ -125,7 +139,7 @@ func (jm *JobMaster) QueryStatus(ctx context.Context, taskID string) *dmpkg.Quer
 // OperateTask operate task.
 func (jm *JobMaster) OperateTask(ctx context.Context, op dmpkg.OperateType, cfg *config.JobCfg, tasks []string) error {
```
lance6716, line 140: it's more safe to make this method private

GMHDBJD: Align with QueryStatus, BinlogSchema...

engine/jobmaster/dm/api.go
GMHDBJD: I think version is more intuitive than outdated.🤔

engine/jobmaster/dm/api.go
lance6716: It must be combined with other functions to fullfill functionality (for example, update checkpoint for "Update" operation), if we expose it as public function, other developer may misuse it. Other functions like QueryStatus, BinlogSchema may be more self-complete.

engine/jobmaster/dm/api.go
lance6716: the client is more complex to handle it. I'll write integration test soon, and for each response I must check the outer revision is same as inner revision. This is not easy as just read one line "queryStatusResp"

But on the other hands, user may need full information to his/her own client logic.

GMHDBJD: For integration test, I think we can just check the number of version, because it's increase one per update request now.

lance6716: My use case is when we need to check if there are two "Running" tasks, we still need to parse and compare revisions, otherwise we don't have confident the "Running" is not outdated. In fact I think all response check should contain this logic, very complicated

engine/jobmaster/dm/metadata/job.go
GMHDBJD: Yes,  changing the configuration of one task will not affect other tasks. Add more comment in [9a868af](https://github.com/pingcap/tiflow/pull/6508/commits/9a868afdd2d787e84b51c10165f535b5132af4d4)

lance6716 (commented):
14 / 25 files viewed

engine/jobmaster/dm/api.go
GMHDBJD: Is this ok?
```
{
    expect-mod-revison: 4
    source1: {
        cfgModRevision: 3(outdated)
    }
}
```

lance6716: can add another field, so (xxx-)cfg-mod-revision is all number.

Now I think the complexity is that task status can't be determinated within the scope of itself. It's OK for me when we don't need to check job status to determine task status.

GMHDBJD: remove cfg-mod-revison and add config-outdated

engine/jobmaster/dm/config/config.go
```diff
@@ -116,6 +116,8 @@ type JobCfg struct {
 	// BWList map[string]*filter.Rules `yaml:"black-white-list" toml:"black-white-list" json:"black-white-list"`
 	// EnableANSIQuotes bool `yaml:"ansi-quotes" toml:"ansi-quotes" json:"ansi-quotes"`
 	// RemoveMeta bool `yaml:"remove-meta"`
+
+	ModRevision uint64 `yaml:"mod_revision" toml:"mod_revision" json:"mod_revision"`
```
lance6716, line 120: mod-revision, same as other variables

lance6716 (commented):
19 / 25 files viewed

switch to take care of my CI 😂

engine/jobmaster/dm/dm_jobmaster.go
```diff
@@ -202,7 +204,11 @@ func (jm *JobMaster) OnWorkerStatusUpdated(worker framework.WorkerHandle, newSta
 		return nil
 	}
 	jm.Logger().Info("on worker status updated", zap.String(logutil.ConstFieldWorkerKey, worker.ID()), zap.String("extra bytes", string(newStatus.ExtBytes)))
-	return jm.handleOnlineStatus(worker)
+	if err := jm.handleOnlineStatus(worker); err != nil {
+		return err
+	}
+	jm.taskManager.SetNextCheckTime(time.Now())
```
lance6716, line 210: what do we plan to do in next tick?

lance6716 (approved):

engine/jobmaster/dm/dm_jobmaster.go
GMHDBJD: operate-task. when task status changed. e.g. running -> paused

engine/jobmaster/dm/metadata/job.go
```diff
@@ -127,3 +127,31 @@ func (jobStore *JobStore) UpdateStages(ctx context.Context, taskIDs []string, st
...
+	}
+	oldJob := state.(*Job)
+
+	// TODO: we may diff the config at task level in the future, that way different tasks will have different modify revisions.
+	// so that changing the configuration of one task will not affect other tasks.
+	var oldVersion uint64
+	for _, task := range oldJob.Tasks {
+		oldVersion = task.Cfg.ModRevision
+		break
+	}
```
amyangfei, line 145: Does each tasks share the same `job config` value currently?

GMHDBJD: No, each task has its own configuration, but use the same ModRevision now.

amyangfei (approved):

PR #6527, opened 2022-07-29
sinkv2(ticdc): use table sink state instead of table state

hi-rustin:

use table sink state instead of table state. This will simplify the implementation. After this, we will not need to move the table state into the table sink and keep it in the sink node.

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```


asddongmen (approved):

cdc/sinkv2/eventsink/mq/worker.go
```diff
@@ -169,7 +169,7 @@ func (w *worker) asyncSend(
 	for key, events := range partitionedRows {
 		for _, event := range events {
 			// Skip this event when the table is stopping.
-			if event.TableStatus.Load() == pipeline.TableStateStopping {
+			if event.SinkState.Load() == state.TableSinkStopping {
```
liuzix, line 172: Can you wrap `event.SinkState.Load()` in a method, probably called `event.GetTableState()`?

cdc/sinkv2/tablesink/state/state.go
```diff
@@ -0,0 +1,53 @@
...
+package state
+
+import "sync/atomic"
+
+// TableSinkState is state of the table sink.
+type TableSinkState int32
+
+// State for table sink
+const (
+	TableSinkStateUnknown TableSinkState = iota
```
liuzix, line 23:
```suggestion
	TableSinkStateUnknown TableSinkState = iota+1
```
Make zero values panic.

cdc/sinkv2/eventsink/mq/worker.go
hi-rustin: Good suggestion!

cdc/sinkv2/tablesink/state/state.go
hi-rustin: Is this a best practice? Why do we need this? Could you please explain a little bit?

nongfushanquan (approved):

hi-rustin:
/merge

@liuzix, I'm gonna merge this PR. If you have any other comments, please feel free to `/hold` this PR.

PR #6564, opened 2022-08-01
scheduler(ticdc): inline secondary to captures map 

overvenus:

Issue Number: ref #4757 

Inline `Secondary` to `ReplicationSet.Captures` and remove `Secondary` in `ReplicationSet`.

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
None
```


cdc/scheduler/internal/v3/replication/replication_set.go
```diff
@@ -90,16 +90,49 @@ func (r ReplicationSetState) MarshalJSON() ([]byte, error) {
...
-	// Primary set indicate that the table is replicating on this capture
+	// Primary is the capture ID that is currently replicating the table.
 	Primary model.CaptureID
-	// Secondary set indicate that the table is preparing on this capture
-	Secondary  model.CaptureID
-	Captures   map[model.CaptureID]struct{}
+	// Captures is a map of captures that has the table replica.
+	// NB: Invariant, 1) at most one primary, 2) primary capture must be in
+	//     CaptureRolePrimary.
+	Captures   map[model.CaptureID]CaptureRole
```
3AceShowHand, line 135: I'm quite curious, Why we have `Captures` field, which only have 2 items? 

Why not just directly use `Primary` and `Secondary` fields ?

cdc/scheduler/internal/v3/replication/replication_set.go
```diff
@@ -195,6 +241,75 @@ func NewReplicationSet(
 	return r, nil
 }
 
+func (r *ReplicationSet) hasRole(role CaptureRole) bool {
+	_, has := r.getRole(role)
+	return has
+}
+
+func (r *ReplicationSet) isInRole(captureID model.CaptureID, role CaptureRole) bool {
```
3AceShowHand, line 249: just name it `inRole` should be ok.

cdc/scheduler/internal/v3/replication/replication_set.go
```diff
@@ -374,16 +487,17 @@ func (r *ReplicationSet) pollOnPrepare(
 				zap.Stringer("tableState", input),
 				zap.String("captureID", captureID),
 				zap.Any("replicationSet", r))
-			r.Primary = ""
-			delete(r.Captures, captureID)
+			r.clearPrimary()
 			return nil, false, nil
-		} else if r.Secondary == captureID {
+		} else if r.isInRole(captureID, CaptureRoleSecondary) {
```
3AceShowHand, line 492: `else if` can be simplified to `if` directly.

cdc/scheduler/internal/v3/replication/replication_set.go
overvenus: I'm planning to support adding more replicas to tables, so TiCDC can handle unplanned shutdown.

cdc/scheduler/internal/v3/replication/replication_set.go
overvenus: The usual convention to name methods that return boolean is to prefix verbs such as 'is' or 'has' to the predicate as a question, or use the predicate as an assertion.

cdc/scheduler/internal/v3/replication/replication_set.go
overvenus: I am planning to add a new role in later PR, called `Standby` for handling unplanned shutdown.

asddongmen (approved):

3AceShowHand (approved):

PR #6565, opened 2022-08-01
sinkv2(ticdc): add sink factory

hi-rustin:

Because golang doesn't support generic arg on interface very well, so we have to remove it.

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```


overvenus (approved):

hicqu (approved):

cdc/sinkv2/eventsink/factory/factory.go
```diff
@@ -0,0 +1,120 @@
...
+	mqSink sinkType = iota + 1
+	txnSink
+)
+
+// SinkFactory is the factory of sink.
+// It is responsible for creating sink and closing it.
+// Because there is no way to convert the eventsink.EventSink[*model.RowChangedEvent]
+// to eventsink.EventSink[eventsink.TableEvent].
+// So we have to use this factory to create and store the sink.
+type SinkFactory struct {
```
lonng, line 44: IMO, it's not a clear abstract because we will `switch-case` everywhere. It's better to define the `SinkFactory` interface and implement the `MQSinkFactory` and `TxnSinkFactory` to make our code clean.

BTW, We just need to add a new implementation instead of change all the `switch-case` if we add more implementation for the interface `SinkFactory` interface.

hi-rustin: I think it's difficult to achieve it. Becase SinkFactory has to hold the real sink instead of `eventsink.EventSink[eventsink.TableSink]`. This is also the reason for this hacky PR.

PR #6569, opened 2022-08-02
framework(engine): maintain worker epoch in heartbeat and worker manager

### LINKED ISSUE #6497: Is it necessary to add a worker epoch in worker heartbeat message

Given the scenario in https://github.com/pingcap/tiflow/issues/6239, an outdated worker(job-master actually) may send stale heartbeat message to its master(job-manager actually). There are two cases
- A stale worker sends heartbeat to job master, since job master doesn't maintain this worker id any more, the heartbeat will be dropped silently.  https://github.com/pingcap/tiflow/blob/bd65b57883f1368f39ab5539e5189d0e013b22f8/engine/framework/internal/master/worker_manager.go#L231-L238
- A stale job master sends heartbeat to job manager, job manager can't recognize this and treat it comes from the active job master. There are also two cases
   - A bug like #6239, multiple job masters are running
   - Another case, the stale job master has exited but stale heartbeat message leaves on the wire, and reaches to job manager with a delay.

Question: is it necessary to maintain a worker epoch in both worker and worker manager of master, and adds it in heartbeat message, master can drop stale heartbeat message if worker epoch doesn't match.

_No response_

_No response_

amyangfei:

Issue Number: close #6497

- Add `worker-epoch` with following rules
   - It is first generated by job master before creating worker, and this epoch will be passed in `dispatchTask` rpc.
   - Worker manager of job master should maintain worker epoch for each worker
   - Each worker records its worker epoch in `master-client`, and adds it in each `HeartbeatPingMessage`
   - Worker manger of job master checks the `worker-epoch` in worker's heartbeat message should match `worker-epoch` maintained by itself.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


amyangfei:
chaos test passed: https://github.com/pingcap/tiflow/actions/runs/2781674397

maxshuang (approved):

engine/framework/internal/master/worker_manager.go
```diff
@@ -569,6 +578,25 @@ func (m *WorkerManager) checkMasterEpochMatch(msgEpoch frameModel.Epoch) (ok boo
 	return true
 }
 
+func (m *WorkerManager) checkWorkerEpochMatch(curEpoch, msgEpoch frameModel.Epoch) bool {
+	if msgEpoch > curEpoch {
+		log.Panic("We are a stale master still running",
+			zap.String("master-id", m.masterID), zap.Int64("own-epoch", m.epoch),
+			zap.Int64("own-worker-epoch", curEpoch),
+			zap.Int64("msg-worker-epoch", msgEpoch),
```
liuzix, line 586: - Is there a unit test case for this branch?
- Could panicking here make the system more unstable in production?

amyangfei: This case may happen because of
- worker manager maintains a smaller worker epoch than that it uses to create worker
- a stale master, another job master creates a job and worker sends heartbeat to stale master

Both of them are bugs and should not happen.

Besides when job master panics here, new job master will load worker from metastore and uses correct worker epoch, it won't continue panic.

amyangfei:
PTAL @liuzix

liuzix (approved):

Issue #4689, opened 2022-02-24
dm-worker keeps retrying to execute ddl when encounter "invalid connection" error


gozssky:
### What did you do?

1. Replicate data from one MySQL to TiDB. 
2. Execute ddl on upstream MySQL:
```sql
ALTER TABLE xxx MODIFY COLUMN xxx SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0'
```

### What did you expect to see?

No error is reported.

### What did you see instead?

DM encountered "invalid connection" error and keeps retrying to execute ddl for every 5 minutes.

### Versions of the cluster

DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):
```console
v2.0.6
```

### current status of DM cluster (execute `query-status <task-name>` in dmctl)

```console
(paste current status of DM cluster here)
```

jiyfhust:
I meet this problem a few days ago，i found it is ```
MaxDDLConnectionTimeoutMinute 
```
set 5min cause the problem.

jiyfhust:
can i make a pr to solve the problem?

lance6716:
> can i make a pr to solve the problem?

welcome! before writing codes, can you give some brief introduction of your fixing? we can discuss the effects.

jiyfhust:
1. https://github.com/pingcap/tiflow/blob/c34585737ab371126c2cdd0fe196dc23fdf68f78/dm/syncer/syncer.go#L3324

SetReadTimeout(maxDDLConnectionTimeout)

2.https://github.com/pingcap/tiflow/blob/c34585737ab371126c2cdd0fe196dc23fdf68f78/dm/pkg/conn/basedb.go#L97

			dsn += fmt.Sprintf("&readTimeout=%s", rawCfg.ReadTimeout)

3.https://github.com/go-sql-driver/mysql/blob/217d05049e5a88d529b9a2d5fe5675120831efab/dsn.go#L51

	Timeout          time.Duration     // Dial timeout
	ReadTimeout      time.Duration     // I/O read timeout
	WriteTimeout     time.Duration     // I/O write timeout
	
4.https://github.com/go-sql-driver/mysql/blob/217d05049e5a88d529b9a2d5fe5675120831efab/packets.go#L115

       conn.SetReadDeadline(time.Now().Add(mc.cfg.ReadTimeout))
       
Because ddl reorganization like add index,modify column may take a long time, is this proper to set ReadTimeout  ulimited?
[...]

PR #6573, opened 2022-08-02
syncer(dm): add the function GetDDLStatusFromTidb()

lyzx2001:

Issue Number: ref #4689 

To solve #4689, we first need to retrieve the synchronizing status of the specified DDLs, which is the goal of the function GetDDLStatusFromTidb().
First, use "ADMIN SHOW DDL JOBS 10" to retrieve the results in the current DDL job queue. For every row, compare its create time with that of the required DDL. If the former one is larger, use "ADMIN SHOW DDL JOB QUERIES LIMIT 1 OFFSET n" to retrieve its query content. If the content is the same as that of the required DDL, return its executing state. Else, increase offset by 1 and check whether the following DLLs meet the requirement.
If none of the results in "ADMIN SHOW DDL JOBS 10" meet the requirement, increase the range by 10 and do the above steps again.
If the create time is smaller than that the required DDL, break the loop and return that the required DDL cannot be found.

#### Tests 

 - Unit test

No.

No.

### Release note 

```release-note
None
```


dm/syncer/error.go
```diff
@@ -61,6 +64,93 @@ func isDropColumnWithIndexError(err error) bool {
...
+		for rows.Next() {
+			err = rows.Scan(scanArgs...)
+			if err != nil {
+				return "", err
+			}
+
+			createTimeStr := string(values[8])
+			timeLayout := "2006-01-02 15:04:05"
+			loc, _ := time.LoadLocation("Local")
+			theTime, _ := time.ParseInLocation(timeLayout, createTimeStr, loc)
```
lance6716, line 113: check if this value is related to downstream timezone, and left an comment

lyzx2001: Modified

dm/syncer/error.go
```diff
@@ -61,6 +64,103 @@ func isDropColumnWithIndexError(err error) bool {
 			strings.Contains(mysqlErr.Message, "with tidb_enable_change_multi_schema is disable"))
 }
 
+// GetDDLStatusFromTiDB() retrieves the synchronizing status of DDL from TiDB
```
lance6716, line 67:
```suggestion
// GetDDLStatusFromTiDB retrieves the synchronizing status of DDL from TiDB
```

Your IDE should display a warning for wrong format of function comment

dm/syncer/error.go
```diff
@@ -61,6 +64,103 @@ func isDropColumnWithIndexError(err error) bool {
...
+	for {
+		showJobs := fmt.Sprintf("ADMIN SHOW DDL JOBS %d", rowNum)
+		// every attempt try 10 history jobs
+		rows, err := db.QueryContext(ctx, showJobs)
+		if err != nil {
+			rows.Close()
+			return "", err
+		}
+
+		columns, err := rows.Columns()
```
lance6716, line 81: there is a function `export.GetSpecifiedColumnValuesAndClose` in `"github.com/pingcap/tidb/dumpling/export"` to implement read part of columns according to column names (not hardcoded column offset), maybe you can use it.

ForwardStar:
Hi I hope I can review this PR. May I know whether this time string ``"2006-01-02 15:04:05"`` has a special meaning? https://github.com/lyzx2001/tiflow/blob/issue4689/dm/syncer/error.go#L112

dm/syncer/error.go
```diff
@@ -61,6 +64,103 @@ func isDropColumnWithIndexError(err error) bool {
...
+					rows.Close()
+					return "", err
+				}
+
+				offset := rowNum + count - 10
+				for {
+					var DDLJob string
+					var jobIDForLimit int
+					showJob := fmt.Sprintf("ADMIN SHOW DDL JOB QUERIES LIMIT 1 OFFSET %d", offset)
+					err = db.QueryRowContext(ctx, showJob).Scan(&jobIDForLimit, &DDLJob)
```
lance6716, line 133: When it's related to interact with other system, I prefer we read -> save needed data to this system -> close -> read another data ...

it's more error-prone if we do the second reading before closing the first reading.

dm/syncer/error.go
```diff
@@ -61,6 +64,103 @@ func isDropColumnWithIndexError(err error) bool {
...
+						return "", err
+					}
+					if jobID == jobIDForLimit && ddl == DDLJob {
+						rows.Close()
+						return string(values[11]), err
+					}
+					if jobIDForLimit <= jobID {
+						break
+					}
+					offset++
```
lance6716, line 145: Since TiDB support we retrieve multiple rows from `ADMIN SHOW DDL JOB QUERIES LIMIT N`, we should use that so network commucation is reduced

dm/syncer/error.go
```diff
@@ -61,6 +64,103 @@ func isDropColumnWithIndexError(err error) bool {
...
+				}
+				count++
+			} else {
+				// requested DDL cannot be found
+				rows.Close()
+				return "", err
```
lance6716, line 151: should return the origin `err` when DDL is not found, but the `err` has been overwritten

dm/syncer/error.go
lance6716: and maybe this function doesn't need to be public

lyzx2001:
> Hi I hope I can review this PR. May I know whether this time string `"2006-01-02 15:04:05"` has a special meaning? https://github.com/lyzx2001/tiflow/blob/issue4689/dm/syncer/error.go#L112

It's just the standard time layout for the first parameter of the function time.Parse()

dm/syncer/error.go
lyzx2001: Modified

dm/syncer/error.go
```diff
@@ -61,6 +65,139 @@ func isDropColumnWithIndexError(err error) bool {
...
+
+	var columnsLimit []string
+	columnsLimit, err = rowsLimit.Columns()
+	if err != nil {
+		rowsLimit.Close()
+		return "", err
+	}
+
+	var resultsLimit [][]string
+	resultsLimit, err = export.GetSpecifiedColumnValuesAndClose(rowsLimit, columnsLimit...)
```
lance6716, line 90: `GetSpecifiedColumnValuesAndClose` can extract part of columns from result, so we can pass needed column names as arguments. Please check it's usage in TiDB project.

dm/syncer/error.go
```diff
@@ -61,6 +65,139 @@ func isDropColumnWithIndexError(err error) bool {
...
+// hence here db should be TiDB database.
+func getDDLStatusFromTiDB(ctx context.Context, db *sql.DB, ddl string, createTime int64) (string, error) {
+	rowNum := 10
+	count := 0
+	rowOffset := 0
+
+	showJobsLimit := fmt.Sprintf("ADMIN SHOW DDL JOB QUERIES LIMIT 10 OFFSET %d", rowOffset)
+	rowsLimit, err := db.QueryContext(ctx, showJobsLimit)
+	if err != nil {
+		rowsLimit.Close()
```
lance6716, line 78: When error is not nil, it's the callee functions's responsibility to close, not caller. here `rowsLimit` may be nil.

dm/syncer/error.go
```diff
@@ -61,6 +65,139 @@ func isDropColumnWithIndexError(err error) bool {
...
 }
 
+// getDDLStatusFromTiDB retrieves the synchronizing status of DDL from TiDB
+// hence here db should be TiDB database.
+func getDDLStatusFromTiDB(ctx context.Context, db *sql.DB, ddl string, createTime int64) (string, error) {
+	rowNum := 10
+	count := 0
+	rowOffset := 0
+
+	showJobsLimit := fmt.Sprintf("ADMIN SHOW DDL JOB QUERIES LIMIT 10 OFFSET %d", rowOffset)
```
lance6716, line 75: with line 144: can we query `ADMIN SHOW DDL JOB QUERIES` in one place?

dm/syncer/error.go
```diff
@@ -61,6 +65,139 @@ func isDropColumnWithIndexError(err error) bool {
...
+
+							var resultsLimitNext [][]string
+							resultsLimitNext, err = export.GetSpecifiedColumnValuesAndClose(rowsLimitNext, columnsLimitNext...)
+							if err != nil {
+								rowsLimitNext.Close()
+								return "", err
+							}
+
+							for k := 0; k < 10; k++ {
+								resultsLimit = append(resultsLimit, resultsLimitNext[k])
```
lance6716, line 167: resultsLimit = append(resultsLimit, resultsLimitNext...)

And downstream may get another DDL between two queries, so we should not directly concatenate them

dm/syncer/error.go
lyzx2001: I don't think it's very convenient. The ```ADMIN SHOW DDL JOB QUERIES``` in line 75 can be considered as the base for these queries and should be executed first, and the one in line 144 serves the purpose of expanding the slice. The latter one is inside 2 for loops, it may not be easy to combine those two queries.

dm/syncer/error.go
```diff
@@ -61,6 +65,139 @@ func isDropColumnWithIndexError(err error) bool {
...
+
+							var columnsLimitNext []string
+							columnsLimitNext, err = rowsLimitNext.Columns()
+							if err != nil {
+								rowsLimitNext.Close()
+								return "", err
+							}
+
+							var resultsLimitNext [][]string
+							resultsLimitNext, err = export.GetSpecifiedColumnValuesAndClose(rowsLimitNext, columnsLimitNext...)
```
ForwardStar, line 160: Is it possible to derive ``resultsLimitNext`` directly from ``results`` array? This may save the query time.

lyzx2001: ```results``` is obtained using ```ADMIN SHOW DDL JOBS```, ```resultsLimitNext``` is obtained using ```ADMIN SHOW DDL JOB QUERIES LIMITm OFFSET n```, ```results``` and ```ADMIN SHOW DDL JOBS``` are arrays containing different columns and information.

lyzx2001: ```results``` is obtained using ```ADMIN SHOW DDL JOBS```, ```resultsLimitNext``` is obtained using ```ADMIN SHOW DDL JOB QUERIES LIMIT m OFFSET n```, ```results``` and ```resultsLimitNext``` are arrays containing different columns and information.

dm/syncer/error.go
```diff
@@ -61,6 +65,114 @@ func isDropColumnWithIndexError(err error) bool {
...
+						var jobIDForLimit int
+						jobIDForLimit, err = strconv.Atoi(resultsLimit[currentOffset][0])
+						if err != nil {
+							return "", err
+						}
+						if jobID == jobIDForLimit && ddl == resultsLimit[currentOffset][1] {
+							return results[i][2], nil
+						}
+						if jobIDForLimit <= jobID {
+							flag = true
```
lance6716, line 155: I remember jobID is not monotonous when we concatenate running DDL and finished DDL, so we should not break here

dm/syncer/error.go
```diff
@@ -61,6 +65,89 @@ func isDropColumnWithIndexError(err error) bool {
...
+						rowOffset += 10
+					} else {
+						if ddl == ddlQuery {
+							return jobsResults[i][2], nil
+						}
+						break
+					}
+				}
+			} else {
+				// requested DDL cannot be found
```
lance6716, line 143: hope ddlCreateTime is monotonous in results. Can you check it?

dm/syncer/error.go
```diff
@@ -61,6 +65,89 @@ func isDropColumnWithIndexError(err error) bool {
...
+						resultsLimitNext, err = export.GetSpecifiedColumnValuesAndClose(rowsLimitNext, "JOB_ID", "QUERY")
+						if err != nil {
+							return "", err
+						}
+
+						// if new DDLs are written to TiDB after the last query 'ADMIN SHOW DDL JOB QUERIES LIMIT 10 OFFSET'
+						// we may get duplicate rows here, but it does not affect the checking
+						for k := 0; k < 10; k++ {
+							var jobIDForLimit int
+							jobIDForLimit, err = strconv.Atoi(resultsLimitNext[k][0])
```
lance6716, line 128: resultsLimitNext may not have 10 rows. And when it has 0 length we should break/return as well

dm/syncer/error.go
lyzx2001: According to the examples in the manual (https://docs.pingcap.com/tidb/stable/sql-statement-admin-show-ddl#examples) and some experiments I conducted in TiDB, ddlCreateTime is monotonous in results.

ForwardStar (approved):
LGTM

dm/syncer/error.go
```diff
@@ -61,6 +65,93 @@ func isDropColumnWithIndexError(err error) bool {
...
+							return "", err
+						}
+
+						var resultsLimitNext [][]string
+						resultsLimitNext, err = export.GetSpecifiedColumnValuesAndClose(rowsLimitNext, "JOB_ID", "QUERY")
+						if err != nil {
+							return "", err
+						}
+						if len(resultsLimitNext) == 0 {
+							break
```
lance6716, line 124: maybe we can `return "", nil` because if we use up JOB QUERIES, no progress would be made if continue.

dm/syncer/error.go
```diff
@@ -61,6 +65,93 @@ func isDropColumnWithIndexError(err error) bool {
...
+						}
+
+						// if new DDLs are written to TiDB after the last query 'ADMIN SHOW DDL JOB QUERIES LIMIT 10 OFFSET'
+						// we may get duplicate rows here, but it does not affect the checking
+						for k := 0; k < 10 && k < len(resultsLimitNext); k++ {
```
lance6716, line 129: for k := range resultsLimitNext

or 

for _, row := range resultsLimitNext

dm/syncer/error_test.go
```diff
@@ -208,3 +210,187 @@ func TestIsConnectionRefusedError(t *testing.T) {
...
+func TestGetDDLStatusFromTiDB(t *testing.T) {
+	var (
+		adminShowDDLJobsSQL1      string
+		adminShowDDLJobsSQL2      string
+		adminShowDDLJobsLimitSQL1 string
+		adminShowDDLJobsLimitSQL2 string
+	)
+
+	db, mock, err := sqlmock.New()
+	require.Nil(t, err)
```
lance6716, line 223:
```suggestion
	require.NoError(t, err)
```

lance6716 (commented):
will review tests later

dm/syncer/error_test.go
```diff
@@ -208,3 +210,187 @@ func TestIsConnectionRefusedError(t *testing.T) {
 	isConnRefusedErr = isConnectionRefusedError(errors.New("connect: connection refused"))
 	require.True(t, isConnRefusedErr)
 }
+
+func TestGetDDLStatusFromTiDB(t *testing.T) {
+	var (
+		adminShowDDLJobsSQL1      string
```
lance6716, line 216:
```suggestion
		adminShowDDLJobsSQL1      = fmt.Sprintf("ADMIN SHOW DDL JOBS 10")
```

and rest of variable initiation

dm/syncer/error_test.go
```diff
@@ -208,3 +210,187 @@ func TestIsConnectionRefusedError(t *testing.T) {
...
+		adminShowDDLJobsSQL2      string
+		adminShowDDLJobsLimitSQL1 string
+		adminShowDDLJobsLimitSQL2 string
+	)
+
+	db, mock, err := sqlmock.New()
+	require.NoError(t, err)
+
+	adminShowDDLJobsSQL1 = fmt.Sprintf("ADMIN SHOW DDL JOBS 10")
```
lance6716, line 225:
```suggestion
	adminShowDDLJobsSQL1 = "ADMIN SHOW DDL JOBS 10"
```

dm/syncer/error_test.go
```diff
@@ -208,3 +210,187 @@ func TestIsConnectionRefusedError(t *testing.T) {
...
+	adminShowDDLJobsSQL2 = fmt.Sprintf("ADMIN SHOW DDL JOBS 20")
+	adminShowDDLJobsLimitSQL1 = fmt.Sprintf("ADMIN SHOW DDL JOB QUERIES LIMIT 10 OFFSET %d", 0)
```
lance6716, line 227:
```suggestion
	adminShowDDLJobsLimitSQL1 = "ADMIN SHOW DDL JOB QUERIES LIMIT 10 OFFSET 0"
```

dm/syncer/error_test.go
```diff
@@ -208,3 +210,187 @@ func TestIsConnectionRefusedError(t *testing.T) {
...
+		AddRow(53, "CREATE TABLE IF NOT EXISTS many_tables_test.t4(i TINYINT, j INT UNIQUE KEY)").
+		AddRow(52, "CREATE TABLE IF NOT EXISTS many_tables_test.t3(i TINYINT, j INT UNIQUE KEY)"))
+
+	var createTime time.Time
+	var status string
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// test 1
+	createTime, err = time.Parse("2006-01-02 15:04:05", "2022-08-02 2:51:38")
```
lance6716, line 370: seems we can declare a constant for `"2006-01-02 15:04:05"` in error.go, and let other places use that constant

dm/syncer/error_test.go
```diff
@@ -208,3 +210,187 @@ func TestIsConnectionRefusedError(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, "synced", status)
+
+	// test 4
+	createTime, err = time.Parse("2006-01-02 15:04:05", "2022-08-03 12:35:00")
+	require.NoError(t, err)
+	status, err = getDDLStatusFromTiDB(ctx, db, "CREATE TABLE IF NOT EXISTS many_tables_test.t7(i TINYINT, j INT UNIQUE KEY)", createTime.Unix())
+	require.NoError(t, err)
+	require.Equal(t, "", status) // DDL does not exist
+}
```
lance6716, line 396: also check mock.ExpectationsWereMet()

dm/syncer/error_test.go
```diff
@@ -208,3 +210,187 @@ func TestIsConnectionRefusedError(t *testing.T) {
...
+		AddRow(61, "ALTER TABLE many_tables_test.t6 ADD x timestamp DEFAULT current_timestamp").
+		AddRow(60, "ALTER TABLE many_tables_test.t5 ADD x timestamp DEFAULT current_timestamp").
+		AddRow(59, "ALTER TABLE many_tables_test.t4 ADD x timestamp DEFAULT current_timestamp").
+		AddRow(58, "ALTER TABLE many_tables_test.t3 ADD x timestamp DEFAULT current_timestamp").
+		AddRow(57, "ALTER TABLE many_tables_test.t2 ADD x timestamp DEFAULT current_timestamp").
+		AddRow(56, "ALTER TABLE many_tables_test.t1 ADD x timestamp DEFAULT current_timestamp").
+		AddRow(55, "CREATE TABLE IF NOT EXISTS many_tables_test.t6(i TINYINT, j INT UNIQUE KEY)").
+		AddRow(54, "CREATE TABLE IF NOT EXISTS many_tables_test.t5(i TINYINT, j INT UNIQUE KEY)").
+		AddRow(53, "CREATE TABLE IF NOT EXISTS many_tables_test.t4(i TINYINT, j INT UNIQUE KEY)").
+		AddRow(52, "CREATE TABLE IF NOT EXISTS many_tables_test.t3(i TINYINT, j INT UNIQUE KEY)"))
```
lance6716, line 253: it's more readable that we put mock and expected result in near place, like

test1 mock result
test1 other test input
test1 expected result
test2...

lance6716 (commented):
rest lgtm

dm/syncer/error_test.go
lyzx2001: Modified

dm/syncer/error.go
```diff
@@ -39,6 +39,9 @@ import (
 	"github.com/pingcap/tiflow/pkg/errorutil"
 )
 
+// the constant layout for transferring time to timestamp.
+var timeLayout = "2006-01-02 15:04:05"
```
lance6716, line 43:
```suggestion
// the time layout for TiDB SHOW DDL statements.
const timeLayout = "2006-01-02 15:04:05"
```

lance6716 (approved):
rest lgtm

dm/syncer/error.go
```diff
@@ -61,6 +68,95 @@ func isDropColumnWithIndexError(err error) bool {
 			strings.Contains(mysqlErr.Message, "with tidb_enable_change_multi_schema is disable"))
 }
 
+// getDDLStatusFromTiDB retrieves the synchronizing status of DDL from TiDB
```
GMHDBJD, line 71: How about move to other file but not error.go?

dm/syncer/error.go
```diff
@@ -61,6 +68,95 @@ func isDropColumnWithIndexError(err error) bool {
...
+			return "", err
+		}
+
+		var jobsResults [][]string
+		jobsResults, err = export.GetSpecifiedColumnValuesAndClose(jobsRows, "JOB_ID", "CREATE_TIME", "STATE")
+		if err != nil {
+			return "", err
+		}
+
+		for i := rowNum - 10; i < rowNum; i++ {
```
GMHDBJD, line 93: How about make 10 as a const variable?

dm/syncer/error.go
```diff
@@ -61,6 +68,95 @@ func isDropColumnWithIndexError(err error) bool {
...
+			ddlCreateTimeStr := jobsResults[i][1]
```
GMHDBJD, line 94: Will lens of jobsResults small then 10?

GMHDBJD (commented):
rest LGTM

dm/syncer/error.go
lyzx2001: Which file do you suggest?
From my perspective, the next PR to solve this issue #4689 will call GetDDLStatusFromTidb in handleSpecialDDLError, and handleSpecialDDLError is in error.go, so I put GetDDLStatusFromTidb here as well

GMHDBJD: Maybe syncer/util.go

lyzx2001: Modified. Since the tests in syncer/util_test.go were using the order version of check ```pingcap/check```, I updated them using  ```stretchr/testify```.
Since there is no function in ```stretchr/testify``` has the same affect as ```HasKey```  and ```HasLen``` of Assert, I manually implement it using ```require.True``` and ```require.Equal```, respectively.

dm/syncer/error.go
lyzx2001: Modified.

dm/syncer/error.go
lyzx2001: I don't think it's possible. I checked by initiating a new TiDB, do nothing then type ```ADMIN SHOW DDL JOBS```, and it returned 10 rows, it seems that every initialized TiDB would have 33 rows of history jobs by default. But to be safe, I added the situation where the lens of jobsResults is small than 10

dm/syncer/error.go
GMHDBJD: You can use require.Len/require.Contains

lyzx2001: Modified.

GMHDBJD (approved):
lGTM

Issue #6491, opened 2022-07-27
Replace google/btree with tidwall/btree 


amyangfei:
### Is your feature request related to a problem?

Given benchmark in https://github.com/tikv/pd/issues/4455, `tidwall/btree` has better performance than `google/btree`

### Describe the feature you'd like

Since PD has replaced google/btree with tidwall/btree after https://github.com/tikv/pd/pull/5335, **we can observe the stability for sometime (maybe half a year) and determine whether to replace it in TiCDC**.

TiCDC uses `google/btree` in three packages, and all of them are quite hot path.

- cdc/entry/schema/snapshot.go
- pkg/causality/internal/node.go
- pkg/regionspan/region_range_lock.go

PR #6592, opened 2022-08-03
schema,regionspan(cdc): use generic version btree to improve performance

crelax:

Issue Number: ref #6491

According to the benchmark results listed in #6491 , the generic version of google/btree performs much better than the non-generic one. So I updated the code in `cdc/entry/schema/snapshot.go` and `pkg/regionspan/region_range_lock.go` to improve the performace.

#### Tests 

 - Unit test
 - Integration test

No
No
### Release note 

```release-note
`use generic version btree to improve performance`.
```


pkg/regionspan/region_range_lock.go
```diff
@@ -40,32 +40,32 @@ func rangeTsEntryWithKey(key []byte) *rangeTsEntry {
...
-	return bytes.Compare(e.startKey, than.(*rangeTsEntry).startKey) < 0
+func rangeTsEntryLess(a, b *rangeTsEntry) bool {
+	return bytes.Compare(a.startKey, b.startKey) < 0
 }
 
 // RangeTsMap represents a map from key range to a timestamp. It supports range set and calculating min value among a
 // a specified range.
 type RangeTsMap struct {
-	m *btree.BTree
+	m *btree.BTreeG[*rangeTsEntry]
```
liuzix, line 50: Please investigate whether using [NewFreeListG](https://pkg.go.dev/github.com/google/btree#NewFreeListG) will have any benefit here.

crelax: Benchmark results show having many btrees sharing one big single freelist does not bring significant performance improvements, especially when the number of btrees can be large. The reasons are 
1) The shared freelist in google/btree is protected by a mutex. Multiple irrelevant btrees waiting for the mutex of freelist does no good to performance. 
2) The size of freelist is hard to decide when we don't know how many btrees share one. Here, the number of RangeTsMap can be very big since there can be a large number of tables . We don't get benefits given the size of freelist is no greater than the number of btrees sharing it.

liuzix (approved):

hicqu (approved):

Issue #6493, opened 2022-07-27
Add help promote for makefile


hi-rustin:
### Is your feature request related to a problem?

For now, we have a bunch of make commands but no docs. It isn't easy to find the available commands.

### Describe the feature you'd like

Adapted from https://www.thapaliya.com/en/writings/well-documented-makefiles/ and add help docs for each command.

hi-rustin:
Introduced in https://github.com/pingcap/tiflow/pull/6568

We need to add help for other targets.

PR #6595, opened 2022-08-03
makefile(ticdc): set default goal correctly

hi-rustin:

Issue Number: ref https://github.com/pingcap/tiflow/issues/6493

set default goal correctly. Set it to build and buildsucc.

#### Tests 

 - No code

No
No
### Release note 

```release-note
None
```


Makefile
```diff
@@ -7,6 +7,7 @@
     build_mysql_integration_test_images clean_integration_test_images \
 	dm dm-master dm-worker dmctl dm-syncer dm_coverage \
 	engine tiflow tiflow-demo tiflow-chaos-case help
+.DEFAULT_GOAL := default
```
lonng, line 10: 1. Please refine the indent of L7 and keep the same with other lines.
L7 uses four spaces and other lines use tab.

2. It's better to insert a blank line before the `.DEFAULT_GOAL := default`.

Rest LGTM

hi-rustin: 1. It's an issue with GitHub.
image
2. I think it makes sense because of almost all PHONY in here.

lonng (approved):

Makefile
lonng: image

Issue #6356, opened 2022-07-20
Refactor the TiCDC project layout and package organization


3AceShowHand:
as title, should be filled

PR #6596, opened 2022-08-03
engine: rename the servermaster package to master

lonng:

Issue Number: ref #6356 

The sub-package `servermaster` in `github.com/pingcap/tiflow/engine` should have a symmetric name with the `executor` because they are the same in conceptual.

This PR rename the `servermaster` to `master` to make its name more consistent with `executor`. But I think we have too many `master/worker/executor` packages. We should clear them and choose a more proper name for each sub-package.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


maxshuang:
Still sounds strange to me if the `servermaster` being changed to `master` since `master` and `worker` are more common pair and we already has the concept of `master` inside the tiflow engine.

IMHO, I think `server-master` or `cluster-master` or `control-center` is proper.

lonng:
OK, I think we need to refine the package name of modules in the engine. And I will close the PR and delay it to the future while we reorganize the concepts and naming.

Issue #6609, opened 2022-08-04
Add readiness probe for servermaster


gozssky:
When servermaster is deployed in Kubernetes, we need to tell the load balancer whether a servermaster is ready for serving requests, so that the traffic won't be forwarded to a follower servermaster. Although we have implemented server failover in the client, we still don't want to configure multiple server addresses in the client, because the client can't persist with the dynamic server addresses.

gozssky:
See https://github.com/pingcap/tiflow/pull/6610#issuecomment-1206054564.

PR #6610, opened 2022-08-04
servermaster(engine): add readiness probe and advertise-leader-addr

gozssky:

Issue Number: close https://github.com/pingcap/tiflow/issues/6609

* Add a readiness probe with path `/readyz`.
* Add `advertise-leader-addr` for leader election and follower forwarding.

`advertise-leader-addr` is used for leader election and follower forwarding only. It is invisible to clients. 
After this modification, multiple servermasters can share the same `advertise-addr` in the Kubernetes environment.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```

### DIFF
--- a/engine/pkg/cmd/master/master.go
+++ b/engine/pkg/cmd/master/master.go
@@ -57,6 +57,8 @@ func newOptions() *options {
 func (o *options) addFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.masterConfig.Addr, "addr", o.masterConfig.Addr, "Set the listening address for server master")
 	cmd.Flags().StringVar(&o.masterConfig.AdvertiseAddr, "advertise-addr", o.masterConfig.AdvertiseAddr, "Set the advertise listening address for client communication")
+	cmd.Flags().StringVar(&o.masterConfig.AdvertiseLeaderAddr, "advertise-leader-addr", o.masterConfig.AdvertiseLeaderAddr,
+		"Set the advertise listening address for follower forwarding. If not set, use the value of `--advertise-addr` instead.")
 	cmd.Flags().StringSliceVar(&o.masterConfig.ETCDEndpoints, "etcd-endpoints", o.masterConfig.ETCDEndpoints, "etcd endpoints")
 
 	cmd.Flags().StringVar(&o.masterConfig.LogConf.File, "log-file", o.masterConfig.LogConf.File, "log file path")
@@ -137,6 +139,8 @@ func (o *options) complete(cmd *cobra.Command) error {
 			cfg.Addr = o.masterConfig.Addr
 		case "advertise-addr":
 			cfg.AdvertiseAddr = o.masterConfig.AdvertiseAddr
+		case "advertise-leader-addr":
+			cfg.AdvertiseLeaderAddr = o.masterConfig.AdvertiseLeaderAddr
 		case "etcd-endpoints":
 			cfg.ETCDEndpoints = o.masterConfig.ETCDEndpoints
 		case "log-file":
--- a/engine/servermaster/campaign.go
+++ b/engine/servermaster/campaign.go
@@ -141,7 +141,7 @@ func (s *Server) checkLeaderExists(ctx context.Context) (retry bool, err error) 
 		}
 
 		// step-2, case-b
-		if leader.Name == s.name() || leader.AdvertiseAddr == s.cfg.AdvertiseAddr {
+		if leader.Name == s.name() || leader.AdvertiseLeaderAddr == s.cfg.AdvertiseLeaderAddr {
 			// - leader.Name == s.name() means this server should be leader
 			// - leader.AdvertiseAddr == s.cfg.AdvertiseAddr means the old leader
 			//   has the same advertise addr as current server
@@ -160,7 +160,7 @@ func (s *Server) checkLeaderExists(ctx context.Context) (retry bool, err error) 
 		// step-2, case-c
 		leader.IsLeader = true
 		log.Info("start to watch server master leader",
-			zap.String("leader-name", leader.Name), zap.String("addr", leader.AdvertiseAddr))
+			zap.String("leader-name", leader.Name), zap.String("addr", leader.AdvertiseLeaderAddr))
 		s.watchLeader(ctx, leader, key, rev)
 		log.Info("server master leader changed")
 	}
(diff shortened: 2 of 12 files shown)


engine/servermaster/config.go
```diff
@@ -60,8 +60,9 @@ const (
 type Config struct {
 	LogConf logutil.Config `toml:"log" json:"log"`
 
-	Addr          string `toml:"addr" json:"addr"`
-	AdvertiseAddr string `toml:"advertise-addr" json:"advertise-addr"`
+	Addr                string `toml:"addr" json:"addr"`
+	AdvertiseAddr       string `toml:"advertise-addr" json:"advertise-addr"`
+	AdvertiseLeaderAddr string `toml:"advertise-leader-addr" json:"advertise-leader-addr"`
```
amyangfei, line 65: Could you please explain why we separate these two addresses. In K8s environment, maybe neither `AdvertiseAddr` nor `AdvertiseLeaderAddr` is exposed to end user directly.

gozssky: `advertise-leader-addr` is like etcd's `advertise-peer-addr`. It is used for followers forwarding requests to the leader.  In K8s environment, k8s operator will create a service for servermaster, all clients should use this address, instead of the pod address (`advertise-leader-addr`). Otherwise, clients can't obtain the benefits of k8s service load balancers.

gozssky: On the other hand, the service endpoint is usually fixed. Separating these two addresses can simplify the deployment of executors.

engine/pkg/cmd/master/master.go
```diff
@@ -57,6 +57,8 @@ func newOptions() *options {
 func (o *options) addFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.masterConfig.Addr, "addr", o.masterConfig.Addr, "Set the listening address for server master")
 	cmd.Flags().StringVar(&o.masterConfig.AdvertiseAddr, "advertise-addr", o.masterConfig.AdvertiseAddr, "Set the advertise listening address for client communication")
+	cmd.Flags().StringVar(&o.masterConfig.AdvertiseLeaderAddr, "advertise-leader-addr", o.masterConfig.AdvertiseLeaderAddr,
```
lonng, line 60: What happens if we always use the `advertise-addr`?

engine/servermaster/config.go
lonng: It's a command line parameter and we cannot change it once the process started. But the leadership of the cluster will be changed dynamically. How do we guarantee the address is correct all the time？

engine/pkg/cmd/master/master.go
gozssky: It means executors must configure at least one valid address for servermaster. But it may be not always work well scenario.

For example:
1. The operator deploys a cluster with a single servermaster with `--advertise-addr=servermaster-0`.
2. All executor is configured to connect `servermaster-0`.
3. The operator scales up servermaster to 2 replicas. The new servermaster's address is `servermaster-1`.
4. servermaster-0 is down and servermaster-1 is elected to become the leader.
5. All executors will crash, because they don't know the servermaster-1's address.

lonng: Why can the executor not detect the master topology? I think we can reference the interaction between TiDB and PD. TiDB will detect the latest PD information without the extra command line parameter.

engine/servermaster/config.go
gozssky: We can reference the pod ip or the pod name. The pod has the same lifetime as servermaster process. So the address is always correct.

engine/pkg/cmd/master/master.go
gozssky: There are some drawbacks of the interaction between TiDB and PD. Like PD client, executors can also detect the master topology. But it can't persist the topology. Once it restarts, or it encounters the scenario as above, it may be unable to connect to master.

gozssky: > It means executors must configure at least one valid address for servermaster. But it may not always work well in all scenarios.
> 
> For example:
> 
> 1. The operator deploys a cluster with a single servermaster with `--advertise-addr=servermaster-0`.
> 2. All executor is configured to connect `servermaster-0`.
> 3. The operator scales up servermaster to 2 replicas. The new servermaster's address is `servermaster-1`.
> 4. servermaster-0 is down and servermaster-1 is elected to become the leader.
> 5. All executors will crash, because they don't know the servermaster-1's address.

@lonng This example can be also used for the interaction between TiDB and PD.

engine/servermaster/config.go
amyangfei: Does it mean in K8s environment, given the following topology 
- A service of server master, such as `svc-server-master:10000`
- Three server master pods
   - Each of them binds `advertise-addr` to `svc-server-master:10000`
   - Each of them binds `advertise-leader-addr` to a headless service, such as `svc-server-master-<id>:10001`
- All executors communicate with `svc-server-master:10000`

engine/servermaster/campaign.go
```diff
@@ -198,7 +198,7 @@ func (s *Server) watchLeader(ctx context.Context, m *rpcutil.Member, key []byte,
 	m.IsLeader = true
 	s.leader.Store(m)
 
-	s.createLeaderClient(ctx, m.AdvertiseAddr)
+	s.createLeaderClient(m.AdvertiseLeaderAddr)
```
amyangfei, line 201: Since the `advertise-leader-addr` will receive forward RPC, `advertise-addr` will receive executor RPC, should the server binds to `0.0.0.0:<port>`, and both `advertise-leader-addr` and `advertise-addr` use this port in their address.

engine/servermaster/config.go
gozssky: > Does it mean in K8s environment, we can deploy as follows:
> 
> * A service, naming `svc-server-master:10000`, with selector of server master
> * Three server master pods
>   
>   * Each of them binds `advertise-addr` to `svc-server-master:10000`
>   * Each of them binds `advertise-leader-addr` to a headless service, such as `svc-server-master-<id>:10000`
> * All executor pods communicate with `svc-server-master:10000`

@amyangfei Yes, that's right.

engine/servermaster/campaign.go
gozssky: `advertise(-leader)-addr` port is not required to be the same as the bind port. The `advertise(-leader)-addr` is the address that advertises the client to connect, it may be a service endpoint or any normal L4 or L7 proxy address.

gozssky:
It may be hard to understand what `advertise-leader-addr` means. And it's not necessary for normal scenarios. So I close this PR.   If needed in the future, reopen it then.

PR #6616, opened 2022-08-04
test(engine): add an integration test for DM

### LINKED ISSUE #6608: wrong error message format
### What did you do?

```
[2022/08/04 05:44:46.353 +00:00] [ERROR] [worker_manager.go:117] ["get job state failed"] [job_id=10b1f5f6-eb2f-4d07-a3d4-fad979ecc97f] [component=worker_manager] [error="[DFLOW:ErrMetaOpFail]meta operation fail%!(EXTRA *errors.errorString=record not found)"] ...
```

https://github.com/pingcap/tiflow/blob/b85512aa2734f4b5046d251c14e620eddd6f6630/engine/pkg/meta/internal/sqlkv/util.go#L39

should not attach args

_No response_

### What did you see instead?

pretty print

### Versions of the cluster

```console
(paste Dataflow Engine version here)
```

lance6716:

Issue Number: close #6608

add a "full mode" test with one task, 500 tables.

#### Tests 

 - Integration test

### Release note 

```release-note
`None`.
```


dm/config/task.go
```diff
@@ -585,6 +585,9 @@ func (c *TaskConfig) adjust() error {
 	if c.TaskMode != ModeFull && c.TaskMode != ModeIncrement && c.TaskMode != ModeAll {
 		return terror.ErrConfigInvalidTaskMode.Generate()
 	}
+	if c.MetaSchema == "" {
```
GMHDBJD, line 588: Do we need copy or refactor the default value here for engine? https://github.com/pingcap/tiflow/blob/de8a5068e3cbb169b3ab18acbb4e278046712b96/dm/config/task.go#L483

engine/test/integration_tests/dm_full_mode/run.sh
```diff
@@ -0,0 +1,52 @@
...
+	wait_mysql_online.sh --port 3306
+	wait_mysql_online.sh --port 4000
+
+	# prepare data
+
+	run_sql 'DROP DATABASE IF EXISTS dm_full_mode'
+	run_sql 'CREATE DATABASE dm_full_mode;'
+	for i in $(seq $TABLE_NUM); do
+		run_sql "CREATE TABLE dm_full_mode.t$i(i TINYINT, j INT UNIQUE KEY);"
+		for j in $(seq 2); do
```
GMHDBJD, line 23: Maybe more data will be better.

lance6716: Will add more large size test as well as multiple jobs later. This is only a medium size test

dm/config/task.go
lance6716: JobCfg.adjust calls https://github.com/pingcap/tiflow/blob/5bd84aea209e9f962e080d2fb2a005a79773591e/engine/jobmaster/dm/config/config.go#L187-L189

we don't use NewTaskConfig now. Will check it soon

GMHDBJD: > Anyway, assigning default value in adjust is more safe in case we forget to use NewTaskConfig in future

GMHDBJD: > Anyway, assigning default value in adjust is more safe in case we forget to use NewTaskConfig in future

Agreed. What I mean is that did you check all default variable in NewTaskConfig is now copy to adjust? Such as CleanDumpFiles, OnlineDDL...

lance6716: we can't tell the default value and zero value. Will use NewTaskConfig in toDMTaskConfig instead.

engine/jobmaster/dm/config/config.go
```diff
@@ -184,7 +184,7 @@ func FromTaskCfgs(taskCfgs []*TaskCfg) *JobCfg {
 
 // toDMTaskConfig transform a jobCfg to DM TaskCfg.
 func (c *JobCfg) toDMTaskConfig() (*dmconfig.TaskConfig, error) {
-	dmTaskCfg := &dmconfig.TaskConfig{}
+	dmTaskCfg := dmconfig.NewTaskConfig()
```
GMHDBJD, line 187: Will it take effects at L194?

lance6716: yaml.Unmarshal should overwrite on old values

GMHDBJD (approved):
LGTM

lance6716:
PTAL @amyangfei @CharlesCheung96

lance6716:
PTAL @amyangfei @CharlesCheung96

Makefile
```diff
@@ -553,15 +553,22 @@ engine_image_arm64:
 	GOOS=linux GOARCH=arm64 $(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/tiflow ./cmd/tiflow/main.go
 	GOOS=linux GOARCH=arm64 $(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/tiflow-demoserver ./cmd/tiflow-demoserver
 	GOOS=linux GOARCH=arm64 $(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/tiflow-chaos-case ./engine/chaos/cases
-	docker build --platform linux/amd64 -f ./engine/deployments/docker/dev.Dockerfile -t dataflow:test ./ 
+	docker build --platform linux/amd64 -f ./engine/deployments/docker/dev.Dockerfile -t dataflow:test ./
+
+engine_image_from_local:
```
CharlesCheung96, line 558: `engine_image_amd64`  is equal to `engine_image_from_local`.

Makefile
```diff
@@ -553,15 +553,22 @@ engine_image_arm64:
 	GOOS=linux GOARCH=arm64 $(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/tiflow ./cmd/tiflow/main.go
 	GOOS=linux GOARCH=arm64 $(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/tiflow-demoserver ./cmd/tiflow-demoserver
 	GOOS=linux GOARCH=arm64 $(GOBUILD) -ldflags '$(LDFLAGS)' -o bin/tiflow-chaos-case ./engine/chaos/cases
-	docker build --platform linux/amd64 -f ./engine/deployments/docker/dev.Dockerfile -t dataflow:test ./ 
+	docker build --platform linux/amd64 -f ./engine/deployments/docker/dev.Dockerfile -t dataflow:test ./
```
CharlesCheung96, line 556:
```suggestion
	docker build --platform linux/arm64 -f ./engine/deployments/docker/dev.Dockerfile -t dataflow:test ./
```

Makefile
lance6716: Can we remove explicit 3 binaries building in engine_image_amd64/arm64? that's too slow and I already have latest binaries.

scripts/download-sync-diff.sh
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+SHA1=$(curl http://fileserver.pingcap.net/download/refs/pingcap/tidb-tools/master/sha1)
+echo "will download tidb-tools at ${SHA1} to get sync_diff_inspector"
+curl -C - --retry 3 -o /tmp/tidb-tools.tar.gz http://fileserver.pingcap.net/download/builds/pingcap/tidb-tools/${SHA1}/centos7/tidb-tools.tar.gz
```
amyangfei, line 19: Better to use an external address.

lance6716: DM integration test uses master build to shorten the response time when sync-diff fixes a bug. 

@CharlesCheung96 Can we let CI environment ro prepare the downloading?

CharlesCheung96: Sure, such binaries will be prepared in advance in the CI environmen. Besides, an external address can be used to allow developers to run integration tests locally.

amyangfei (approved):

Issue #5555, opened 2022-05-24
seems grpc name has collision


lance6716:
### What did you do?

2022/05/24 03:25:35 proto: duplicate proto type registered: pb.WorkerInfo

https://github.com/pingcap/tiflow/runs/6566411247?check_suite_focus=true

### What did you expect to see?

no error

### What did you see instead?

2022/05/24 03:25:35 proto: duplicate proto type registered: pb.WorkerInfo

### Versions of the cluster

master

amyangfei:
Yep,
https://github.com/pingcap/tiflow/blob/12317f4139996245d238fcaff7c15d18308607da/dm/dm/proto/dmmaster.proto#L349
https://github.com/pingcap/tiflow/blob/12317f4139996245d238fcaff7c15d18308607da/engine/proto/master.proto#L92

`WorkerInfo` defined in dataflow engine should be renamed.

PR #6617, opened 2022-08-04
engine: rename the enginepb package name to pb

lonng:

Issue Number: ref #6356 

The package `github.com/pingcap/tiflow/engine/enginepb` has a duplicated word `engine`, and it's better to keep the same package naming style with `github.com/pingcap/tiflow/dm/pb` because all imports use `pb` as its alias (see below picture).

3vqs59YDFr

This PR refactors the package name of `github.com/pingcap/tiflow/engine/enginepb` and eliminate the duplicated word `engine`.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


amyangfei (approved):

lonng:
I will address the conflicts after two reviewers LGTM.

maxshuang:
@lonng @amyangfei 
The reason why engine change pkg name from 'pb' to 'enginepb' is :
https://github.com/pingcap/tiflow/issues/5555
https://github.com/pingcap/tiflow/pull/5627

I think we need to do more work.

amyangfei:
> @lonng @amyangfei The reason why engine change pkg name from 'pb' to 'enginepb' is : #5555 #5627
> 
> I think we need to do more work.

Yes, I forget this issue. What about using the structure in pingcap/kvproto

https://github.com/pingcap/kvproto/tree/master/pkg

lonng:
> > @lonng @amyangfei The reason why engine change pkg name from 'pb' to 'enginepb' is : #5555 #5627
> > I think we need to do more work.
> 
> Yes, I forget this issue. What about using the structure in pingcap/kvproto, https://github.com/pingcap/kvproto/tree/master/pkg, such as
> 
> ```shell
> pb
> ├── cdc
> ├── dm
> └── engine
> ```
> 
> But we need package alias when using them.

Make sense, I will reorg the structure.

lonng:
@amyangfei Why not change the `enginepb.WorkerInfo` to `pb.ExecutorInfo`,

```
message WorkerInfo {
    string id = 1;
    string executor_id = 2;
    bytes  status = 3;
    bytes  config = 4;
    bool  is_tombstone = 5;
    int64  last_hb_time = 6;
    int64  workload = 7;
}
```

What the difference between `id` and `executor_id`?

amyangfei:
> @amyangfei Why not change the `enginepb.WorkerInfo` to `pb.ExecutorInfo`,
> 
> ```
> message WorkerInfo {
>     string id = 1;
>     string executor_id = 2;
>     bytes  status = 3;
>     bytes  config = 4;
>     bool  is_tombstone = 5;
>     int64  last_hb_time = 6;
>     int64  workload = 7;
> }
> ```
> 
> What the difference between `id` and `executor_id`?

`enginepb.WorkerInfo` here is the information of a worker in master-worker framework, where `worker` can be treated as a runtime task and runs real job such as DM task. So `id` is the worker id, `executor_id` means on which executor is the worker running.

lonng:
@amyangfei Got it.

gozssky:
https://github.com/pingcap/tiflow/pull/6669 has changed the proto inlcude directory from engine/proto to project directory. So I guess there will be no pb type collision like `pb.WorkerInfo`.

Now engine is using the official [golang protobuf](https://github.com/protocolbuffers/protobuf-go). Golang protobuf doesn't allow duplicate filename globally, see https://github.com/golang/protobuf/issues/1122. That's why I changed the include directory. Otherwise, it will conflict with https://github.com/pingcap/tipb.

To avoid such problems, I think we can refer to https://github.com/googleapis/googleapis structure. Not only the generated codes but also the proto files, need to have a unique import path. So that we don't need to change our message name, file name or package name, which looks weird. There is a simple example in https://github.com/juanfont/headscale/tree/main/proto/headscale/v1.

gozssky (approved):

lonng:
This PR needs further work to merge and I will close it.

PR #6619, opened 2022-08-04
framework(engine): remove 'JobMasterID' method from 'BaseJobMaster' interface

### LINKED ISSUE #6615: engine: Remove the unnecessary method 'JobMasterID' in JobMaster interface

No

Although 'JobMaster' is a combination of 'Master' and 'Slave' inside, for the user, 'JobMaster' is an independent  character which has the ability to manager its 'workers'. 

So if we offer both the 'JobMasterID' and 'ID' interface method in  'JobMaster', it will confuses user.
The normal doubt Is : 

1. What's the difference of  JobMasterID and ID? 
2. What do they mean respectively?
3. Are they equal?

So it's better to remove the unnecessary method 'JobMasterID' in JobMaster interface to avoid confusion.

Remove the unnecessary method 'JobMasterID' in JobMaster interface

_No response_

maxshuang:

Issue Number: close #6615 

'JobMasterID' and 'ID' return the same id to user and we don't need the `JobMasterID` to return `JobManagerID` actually,
so remove 'JobMasterID'  method to avoid confusion

### Release note 

```release-note
 `None`.
```

### DIFF
--- a/engine/framework/base_jobmaster.go
+++ b/engine/framework/base_jobmaster.go
@@ -23,7 +23,6 @@ import (
 	"github.com/pingcap/log"
 	"go.uber.org/zap"
 
-	runtime "github.com/pingcap/tiflow/engine/executor/worker"
 	frameModel "github.com/pingcap/tiflow/engine/framework/model"
 	"github.com/pingcap/tiflow/engine/model"
 	dcontext "github.com/pingcap/tiflow/engine/pkg/context"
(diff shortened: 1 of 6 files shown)


GMHDBJD (approved):
LGTM

engine/framework/base_jobmaster.go
```diff
@@ -275,13 +273,10 @@ func (d *DefaultBaseJobMaster) Workload() model.RescUnit {
 	return d.worker.Workload()
 }
 
-// ID delegates the ID of inner worker
-func (d *DefaultBaseJobMaster) ID() runtime.RunnableID {
-	return d.worker.ID()
-}
```
amyangfei, line 281: This function signature should not be changed.

amyangfei (approved):

engine/framework/base_jobmaster.go
maxshuang: Yes!! You are right. Fix it.
BTW, I think it's strange for the `ID()` of `Worker` to return a runtime-related type `runtime.RunnableID`, but not a `WorkerID`.

maxshuang:
@GMHDBJD Could you please help fix the unit test error in `tiflow/engine/jobmaster/dm`?  Just remove the meaningless ID in your cases.

maxshuang:
Hold this pr untill https://github.com/pingcap/tiflow/pull/6620 being merged to master.
#6620 will remove the usage of `JobMasterID` method of dm.

PR #6620, opened 2022-08-04
dm(engine): cache finished status

GMHDBJD:

Issue Number: ref #5504 

when task is finished, cache the finished status so user can use query the status of the job.
- cache finished status when worker is finished
- when master failover or job updated, new worker will be created and finished status will be reset.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


engine/jobmaster/dm/dm_jobmaster.go
```diff
@@ -51,6 +51,9 @@ type JobMaster struct {
 	// only use when init
 	// it will be outdated if user update job cfg.
 	initJobCfg *config.JobCfg
+	// taskID -> FinishedTaskStatus
+	// worker exists after finished, so we need record the finished status for QueryJobStatus
+	finishedStatus sync.Map
```
lance6716, line 55: Worth writing in comment that `finishedStatus` will be lost when jobmaster failover.

And can we use Task.Stage to mark a task is finished? So when failover, only the summary status is lost. And maybe we can also store summary status into Task

engine/jobmaster/dm/api.go
```diff
@@ -170,6 +174,11 @@ func (jm *JobMaster) UpdateJobCfg(ctx context.Context, cfg *config.JobCfg) error
 	if err := jm.checkpointAgent.Update(ctx, cfg); err != nil {
 		return err
 	}
+	// reset finished status, all tasks will be restarted now.
+	jm.finishedStatus.Range(func(key interface{}, value interface{}) bool {
```
lance6716, line 178: maybe create a new sync.Map

engine/jobmaster/dm/dm_jobmaster.go
```diff
@@ -161,6 +162,7 @@ func (jm *JobMaster) handleOnlineStatus(worker framework.WorkerHandle) error {
 		return err
 	}
 
+	jm.finishedStatus.Delete(taskStatus.Task)
```
lance6716, line 165: in which case, will worker become Offine with Finished then become Online?

GMHDBJD: when switch to another unit.

GMHDBJD:
ut failed https://github.com/pingcap/tiflow/issues/6665
/run-verify-ci

engine/jobmaster/dm/api.go
```diff
@@ -98,7 +98,13 @@ func (jm *JobMaster) QueryJobStatus(ctx context.Context, tasks []string) (*JobSt
 					// task finished
 					workerID = workerStatus.ID
 					cfgModRevision = workerStatus.CfgModRevision
-					queryStatusResp = &dmpkg.QueryStatusResponse{Unit: workerStatus.Unit, Stage: metadata.StageFinished}
+					finishedStatus, ok := jm.finishedStatus.Load(taskID)
+					if !ok {
+						queryStatusResp = &dmpkg.QueryStatusResponse{Unit: workerStatus.Unit, Stage: metadata.StageFinished, ErrorMsg: fmt.Sprintf("task %s finished but status not found", taskID)}
```
lance6716, line 103: “not found” may cause worry, maybe "task %s is finished and status has been deleted"

lance6716 (approved):

maxshuang (approved):

PR #6622, opened 2022-08-04
client(engine): implement tombstones in ExecutorGroup

### LINKED ISSUE #6621: [Engine] Long tail latency of CreateWorker
### What did you do?

Shutdown an executor when trying to CreateWorker and the worker is dispatched to that executor that has been shutdown.

Retry should happen quickly.

### What did you see instead?

Retry continued for at least 10 seconds.

### Versions of the cluster

```console
a26ab366deb3eb97f46456348dac964b8a22e511
```

liuzix:

Issue Number: close #6621 

- Reduced the retry timeout for non-existent executors.
- Added a tombstone mechanism in `ExecutorGroup` to make such `GetExecutorClient` calls fail fast. 

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
None.
```

### DIFF
--- a/engine/pkg/client/errors.go
+++ b/engine/pkg/client/errors.go
@@ -22,7 +22,8 @@ import (
 type ExecutorNotFoundErrInfo struct {
 	rpcerror.Error[rpcerror.Retryable, rpcerror.NotFound]
 
-	ExecutorID model.ExecutorID
+	ExecutorID  model.ExecutorID
+	IsTombstone bool
 }
 
 // ErrExecutorNotFound is used when an executor ID is not found for
--- a/engine/pkg/client/executor_group.go
+++ b/engine/pkg/client/executor_group.go
@@ -14,7 +14,9 @@
 package client
 
 import (
+	"container/list"
 	"context"
+	"errors"
 	"sync"
 	"time"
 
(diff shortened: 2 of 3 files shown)


engine/pkg/client/errors.go
```diff
@@ -22,7 +22,8 @@ import (
 type ExecutorNotFoundErrInfo struct {
 	rpcerror.Error[rpcerror.Retryable, rpcerror.NotFound]
 
-	ExecutorID model.ExecutorID
+	ExecutorID  model.ExecutorID
```
maxshuang, line 25: If we have limited online executors, why don't we check the `online` executor list? so we don't need to do some gc.

maxshuang: Anything by designed here?

liuzix: We need a list of "executors definitely not online" in addition to a list of "executors currently online". For example, if you want a client for `executor-1`, it is possible that

1. The executor has gone offline, or
2. The executor has just gotten online, but due to a consistency problem, we just don't have a record for it yet.

It is worthwhile distinguishing these two situations because in situation 1, waiting for a while will be futile, just leading to wasted latency.

engine/pkg/client/executor_group.go
```diff
@@ -202,8 +257,27 @@ func (g *DefaultExecutorGroup) RemoveExecutor(executorID model.ExecutorID) error
 	client.Close()
 
 	delete(g.clients, executorID)
```
maxshuang, line 259: Seems that `g.client` and `g.tomstoneList` is updated under a lock `g.mu`, checking the `online list` and `offline list` has no difference. 

Can you explain more about the `executor has gone offline`?

maxshuang: when an executor's heartbeat is timeout, I think we no longer need to send something to executor.
So `executor is offline` and `executor has gone offline` seems have no big difference.

liuzix: - For situation 1, we need to give up immediately and try another executor.
- For situation 2, we need to wait for a while and retry.

maxshuang: Get the point.
`executor online list` in servermaster and executor may be different for some network delay.
So when a jobmaster get a worker scheduler plan with an executor id in it, it may find the `oriented executor` is not in the its `online executor list`. In that case, we need a tomstone list to indicate which executor is dead certainly.

maxshuang: We can add some comments to avoid the unexpected modification on this mechanism in the future.

maxshuang (approved):

amyangfei (approved):

PR #6626, opened 2022-08-05
api(ticdc): status api add cluster id

### LINKED ISSUE #6625: status api add cluster id
add cluster into status api response, so we can get the cluster from cdc Open API

sdojjy:

Issue Number: close #6625

add ClusterID field
```
type ServerStatus struct {
	Version   string   `json:"version"`
	GitHash   string   `json:"git_hash"`
	ID        string   `json:"id"`
	ClusterID string   `json:"cluster_id"`
	Pid       int      `json:"pid"`
	IsOwner   bool     `json:"is_owner"`
	Liveness  Liveness `json:"liveness"`
}
```

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


cdc/api/v1/api_test.go
```diff
@@ -785,9 +789,11 @@ func TestServerStatus(t *testing.T) {
 	require.Nil(t, err)
 	require.Equal(t, "capture-for-test", resp.ID)
 	require.True(t, resp.IsOwner)
+	require.Equal(t, "abcd", resp.ClusterID)
```
lonng, line 792: Do we need to add `liveness` test cases?

sdojjy: liveness test cases is already added

3AceShowHand (approved):

asddongmen (approved):

PR #6627, opened 2022-08-05
owner(cdc): cdc healthy also check cluster consistency

### LINKED ISSUE #6623: cdc 6.2 cannot tolerant tikv 6.3.0 cause drain capture cannot work
Deploy a cluster in `v6.2.0`, and upgrade it to v6.3.0.

After TiKV in `v6.3.0`, TiCDC is still `v6.2.0`. `DrainCapture` accept the request, but cannot move tables to other target captures, because no capture can contact TiKV successfully.

By using `tiup cluster upgrade` a CDC cluster from `v6.2.0` to `v6.3.0`, this has 2 problems:
* latency goes up too high, because the whole cdc cluster cannot contact to tikv, which is known and acceptable.
* CDC instances cannot be stopped and then restart in time, always cost `--transfer-timeout` at least for one capture, which is 10 minutes, this would make replication latency goes up too high, we have to find a solution to make the CDC can be upgraded more quickly.

* If the cluster is not consistent, should be treated as `unhealthy`

3AceShowHand:

Issue Number: ref #4757,close #6623

also check cdc cluster consistency and tikv version for cdc healthy.

#### Tests 

 - Unit test

### Release note 

```release-note
`None`
```


cdc/owner/owner.go
```diff
@@ -599,27 +599,45 @@ func (o *ownerImpl) handleQueries(query *Query) error {
...
 	}
 	return nil
 }
 
+func (o *ownerImpl) isHealthy(ctx context.Context) (bool, error) {
+	if !o.clusterVersionConsistent(o.captures) {
+		return false, nil
+	}
+
+	upstreamM, err := o.upstreamManager.GetDefaultUpstream()
```
sdojjy, line 616: cdc support multi upstream, check all upstreams

cdc/owner/owner.go
```diff
@@ -599,27 +599,45 @@ func (o *ownerImpl) handleQueries(query *Query) error {
...
+	if err != nil {
+		return false, errors.Trace(err)
+	}
+	if err := version.CheckStoreVersion(ctx, upstreamM.PDClient, 0); err != nil {
```
sdojjy, line 620: do we need to check version consistency for every health API call?

3AceShowHand: Some API works even though the cluster is not consistent, but this can not work properly.

If we try to make every API check version, we have to investigate each one first.

sdojjy (approved):

cdc/owner/owner_test.go
```diff
@@ -755,18 +758,55 @@ func (h *heathScheduler) IsInitialized() bool {
...
+	ctx := context.Background()
+
+	// Unhealthy, cdc cluster version is inconsistent
+	o.captures = map[model.CaptureID]*model.CaptureInfo{
+		model.CaptureID("1"): {
+			Version: "v6.3.0",
+		},
+		model.CaptureID("2"): {
+			Version: "v8.0.0",
+		},
```
overvenus, line 778: Is it healthy if one v6.3.0 and the other v7.0.0 ?

overvenus: Please do not hard code version, use `version.maxTiCDCVersion`.

overvenus: Please also test a version that is smaller that `minTiCDCVersion`.

overvenus (approved):

PR #6642, opened 2022-08-07
makefile(all): add format and check targets for makefiles

hi-rustin:

Issue Number: ref https://github.com/pingcap/tiflow/issues/6493

See more: https://github.com/pingcap/tiflow/pull/6595#discussion_r936421163

Add format and check targets for makefiles. 

#### Tests 

 - No code

No
No
### Release note 

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,8 @@
 	mysql_docker_integration_test mysql_docker_integration_test_with_build \
 	build_mysql_integration_test_images clean_integration_test_images \
 	dm dm-master dm-worker dmctl dm-syncer dm_coverage \
-	engine tiflow tiflow-demo tiflow-chaos-case engine_image help
+	engine tiflow tiflow-demo tiflow-chaos-case engine_image help \
+	format-makefiles
 
 .DEFAULT_GOAL := default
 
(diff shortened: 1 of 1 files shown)


hi-rustin:
/hold 
We should also use SED in other targets.

lonng (approved):

Makefile
```diff
@@ -120,6 +130,14 @@ test: unit_test dm_unit_test engine_unit_test
 
 build: cdc dm engine
 
+check-makefiles: ## Check the makefiles format.
+check-makefiles: format-makefiles
+	@git diff --exit-code -- $(MAKE_FILES) || (echo "Please format Makefiles by running 'make format-makefiles'" && false)
+
+format-makefiles: ## Format all Makefiles.
+format-makefiles: $(MAKE_FILES)
+	$(SED) -i -e 's/^\(\t*\)  /\1\t/g' -e 's/^\(\t*\) /\1/' -- $?
```
amyangfei, line 139: No need to address. It works fine if the indent of tab is 2 or 4 spaces, but in 4 spaces scenario we need to run `make format-makefiles` multiple times to replace all spaces.

amyangfei (approved):

amyangfei:
> /hold We should use SED in other targets.

What does other targets mean

hi-rustin:
> > /hold We should use SED in other targets.
> 
> What does other targets mean

I meant the target which uses the sed. Like `dm_coverage`.

hi-rustin:
/unhold

Ready to merge.

Makefile
hi-rustin: Good catch! 

PR #6647, opened 2022-08-08
capture(ticdc): disable best effort graceful shutdown

### LINKED ISSUE #6646: Upgrade TiCDC from v6.2 to v6.3 takes a long time
### What did you do?

Upgrade v6.2 TiCDC cluster to v6.3, it takes about 100s for every TiCDC instance.

Should be with in 10s.

### What did you see instead?

Takes about 100s to shutdown a TiCDC instance.

### Versions of the cluster

```console
v6.2.0
```

overvenus:

Issue Number: close #6646

Because TiCDC release-6.2 does not support cross version upgrade and stops
service during upgrade, it is impossible to perform graceful shutdown.

#### Tests 

 - Unit test

No.

No.

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/capture/capture.go
+++ b/cdc/capture/capture.go
@@ -76,6 +76,7 @@ type captureImpl struct {
 	processorManager processor.Manager
 	liveness         model.Liveness
 	config           *config.ServerConfig
+	disableDrain     bool
 
 	pdEndpoints     []string
 	ownerMu         sync.Mutex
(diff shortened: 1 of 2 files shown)


cdc/capture/capture.go
```diff
@@ -114,13 +115,19 @@ type captureImpl struct {
 	newOwner func(upstreamManager *upstream.Manager) owner.Owner
 }
 
+// disableDrain is a boolean that disable capture drain feature.
+// Because TiCDC release-6.2 does not support cross version upgrade and stops
+// service during upgrade, it is impossible to perform graceful shutdown.
+const disableDrain = true
```
lonng, line 121: Seems it's not configurable.

3AceShowHand:
I've test the PR, it works, reduce CDC upgrade time cost

3AceShowHand (approved):

cdc/capture/capture.go
overvenus: Yes, because this PR targets on release-6.2 branch where the best effort graceful shutdown is impossible during upgrading.

sdojjy (approved):

PR #6710, opened 2022-08-11
version(ticdc): adjust cdc version range to 6.3.0-alpha ~ 8.0.0-alpha

3AceShowHand:

Issue Number: ref #4757 

change cdc version from [6.3.0, 8.0.0), to [6.3.0-alpha, 8.0.0-alpha), so that `6.3.0-master` in the range.

#### Tests 

 - Unit test

### Release note 

```release-note
None`
```

### DIFF
--- a/pkg/version/check_test.go
+++ b/pkg/version/check_test.go
@@ -210,6 +210,22 @@ func TestCompareVersion(t *testing.T) {
 		Compare(*semver.New("3.0.5-beta.12")), 0)
 	require.Equal(t, semver.New(removeVAndHash("v2.1.0-rc.1-7-g38c939f-dirty")).
 		Compare(*semver.New("2.1.0-rc.1")), 0)
+
+	// build on master branch, `vx.y.z-master`
+	masterVersion := semver.New(removeVAndHash("v6.3.0-master"))
+	require.Equal(t, 1, masterVersion.Compare(*MinTiCDCVersion))
+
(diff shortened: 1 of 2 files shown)


sdojjy (approved):

pkg/version/check_test.go
```diff
@@ -195,21 +195,22 @@ func TestCheckClusterVersion(t *testing.T) {
...
+	// pre-release version, `vx.y.z-alpha-nightly-yyyymmdd`
+	alphaVersion := semver.New(removeVAndHash("v6.3.0-alpha-nightly-20220202"))
+	require.Equal(t, 1, alphaVersion.Compare(*MinTiCDCVersion))
+
+	// release version, `vx.y.z.`
+	releaseVersion := semver.New(removeVAndHash("v6.3.0"))
+	require.Equal(t, 1, releaseVersion.Compare(*MinTiCDCVersion))
+
+	// build with uncommitted changes, `vx.y.z-dirty`
+	dirtyVersion := semver.New(removeVAndHash("v6.3.0-dirty"))
```
lonng, line 211: Which one is greater between `v6.3.0-dirty` and `v6.3.0`?

overvenus (approved):
Nice catch

pkg/version/check_test.go
3AceShowHand: they are equal, since `v6.3.0-dirty` will be changed to `v6.3.0` first before comparison.

PR #6721, opened 2022-08-12
sinkv2(ticdc): remove useless role in eventsink

hi-rustin:

remove useless role in eventsink, because we only use it in processor.

#### Tests 

 - Unit test
 - Integration test

No
No
### Release note 

```release-note
None
```

### DIFF
--- a/cdc/sinkv2/ddlsink/mq/ddlproducer/kafka_ddl_producer.go
+++ b/cdc/sinkv2/ddlsink/mq/ddlproducer/kafka_ddl_producer.go
@@ -178,13 +178,13 @@ func (k *kafkaDDLProducer) Close() {
 	go func() {
 		start := time.Now()
 		if err := k.client.Close(); err != nil {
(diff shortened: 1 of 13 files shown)


asddongmen (approved):

overvenus (approved):

cdc/sinkv2/ddlsink/mq/ddlproducer/kafka_ddl_producer.go
```diff
@@ -67,12 +67,14 @@ func NewKafkaDDLProducer(ctx context.Context, client sarama.Client,
 		go func() {
 			err := client.Close()
 			if err != nil {
-				log.Error("Close sarama client with error", zap.Error(err),
+				log.Error("Close sarama client with error in kafka "+
```
lonng, line 70: The width is less than 80 and why do we need split the `Close sarama client with error in kafka DDL producer` into two lines?

PR #6755, opened 2022-08-16
deployment(engine): fix invalid variable type in docker compose env

amyangfei:

Issue Number: ref #6554

Fix following error

```bash
$ docker-compose -f 1meta.yaml up -d
ERROR: The Compose file './1meta.yaml' is invalid because:
services.mysql-standalone.environment.MYSQL_ALLOW_EMPTY_PASSWORD contains true, which is an invalid type, it should be a string, number, or a null
```

Either `MYSQL_ALLOW_EMPTY_PASSWORD: "yes"` or `MYSQL_ALLOW_EMPTY_PASSWORD: 1` is valid

#### Tests 

    - Check docker compose runs normally 

### Release note 

```release-note
None
```

### DIFF
--- a/deployments/engine/docker-compose/1m1e.yaml
+++ b/deployments/engine/docker-compose/1m1e.yaml
@@ -52,7 +52,7 @@ services:
     image: mysql:5.7
     container_name: mysql_standalone
     environment:
-      MYSQL_ALLOW_EMPTY_PASSWORD: true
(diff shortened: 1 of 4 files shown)


GMHDBJD (approved):
LGTM

lance6716 (approved):

PR #6782, opened 2022-08-17
syncer: remove syncer dependency from shardddl

GMHDBJD:

Issue Number: ref #4287

- combine ddl.go and shardddl.go
- shardddl only manages its own state and does not rely on syncer
### Note: some code will be refactor in later pr
- some callback(handleJobFunc...) will be removed
- some state should only be use in shardddl but not in syncer, like onlineddl... 

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


dm/syncer/ddl.go
```diff
@@ -18,18 +18,935 @@ import (
...
+	"go.uber.org/atomic"
+	"go.uber.org/zap"
 )
 
+type ShardDDLStrategy interface {
+	preFilter(ddlInfo *ddlInfo, qec *queryEventContext, sourceTable *filter.Table, targetTable *filter.Table) (bool, error)
+	handleDDL(qec *queryEventContext) error
+}
+
+type ShardDDL struct {
```
lance6716, line 53: Maybe DDLWorker

dm/syncer/ddl.go
```diff
@@ -18,18 +18,935 @@ import (
...
+	case config.ShardOptimistic:
+		shardDDL.strategy = NewOptimistDDL(&shardDDL.logger, syncer)
+	default:
+		shardDDL.strategy = NewNormalDDL(&shardDDL.logger, syncer)
+	}
+	return shardDDL
+}
+
+type Normal struct {
+	ShardDDLStrategy
```
lance6716, line 118: why embedded this interface? we can assert Normal/Optimist/Pessimist implement `ShardDDLStrategy`

lance6716 (commented):
4 / 10 files viewed

dm/syncer/ddl.go
```diff
@@ -18,18 +18,912 @@ import (
...
+	"github.com/pingcap/tiflow/dm/syncer/metrics"
+	onlineddl "github.com/pingcap/tiflow/dm/syncer/online-ddl-tools"
+	sm "github.com/pingcap/tiflow/dm/syncer/safe-mode"
+	"github.com/pingcap/tiflow/dm/syncer/shardddl"
+	"go.uber.org/atomic"
+	"go.uber.org/zap"
 )
 
+type shardDDLStrategy interface {
+	preFilter(ddlInfo *ddlInfo, qec *queryEventContext, sourceTable *filter.Table, targetTable *filter.Table) (bool, error)
```
lance6716, line 49: when preFilter returns true, it means we should skip this DDL

dm/syncer/ddl.go
```diff
@@ -18,18 +18,912 @@ import (
...
+				}
+			})
+
+			ddl.logger.Info("execute DDL job",
+				zap.String("event", "query"),
+				zap.Stringer("queryEventContext", qec),
+				zap.String("sourceTableID", sourceTableID),
+				zap.Stringer("operation", shardOp))
+		} else {
+			ddl.logger.Info("ignore DDL job",
```
lance6716, line 662: seems this is the same with old logic, but why we print "ignore" and continue building the DDL job 😂

lance6716 (approved):

lance6716:
PTAL @D3Hunter @buchuitoudegou

D3Hunter (approved):

buchuitoudegou (approved):

PR #6669, opened 2022-08-09
servermaster(engine): implement openapi with grpc-gateway

gozssky:
### What problem does this PR solve?

Issue Number: close https://github.com/pingcap/tiflow/issues/6668

### What is changed and how it works?
* Change gogo protobuf to https://github.com/protocolbuffers/protobuf-go.
* Remove HTTP OpenAPI framework.
* Change proto inlcude directory from `engine/proto` to project directory to avoid proto file conflicts. See https://github.com/golang/protobuf/issues/1122.
* Refactor JobManager service to support gRPC-gateway.

Most of the changes are in the generated code. Please review the changes in the `engine/servermaster` directory and `engine/proto/master.proto`.

#### Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

### Release note 

```release-note
None
```

PR #6788, opened 2022-08-18
resource(engine): Improve RPC method error handling

liuzix:

Issue Number: ref #6208

- Adopted `engine/pkg/rpcerror` to standardize gRPC methods' return errors.
- Added appropriate retry mechanisms in the `engine/pkg/client` package.

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
None.
```


engine/pkg/client/resource_manager_client.go
```diff
@@ -37,21 +39,69 @@ func NewResourceManagerClient(cli enginepb.ResourceManagerClient) ResourceManage
...
+	})
+	if err != nil {
+		return err
+	}
+
+	// Check whether the previous request has actually succeeded.
+	if queryResp.CreatorWorkerId == request.CreatorWorkerId {
+		return nil
+	}
+	return originalErr
```
amyangfei, line 78: Does this happen when two workers create the same resource?

engine/pkg/externalresource/internal/errors.go
```diff
@@ -0,0 +1,66 @@
...
+	ResourceID resModel.ResourceID
+	Details    string
+}
+
+// ErrResourceMetastoreError indicates that there is a failure in querying the
+// metastore on a resource, but the error can be non-fatal.
+var ErrResourceMetastoreError = rpcerror.Normalize[ResourceMetastoreError]()
+
+// ResourceAlreadyExistsErrorInfo provides details of ErrResourceAlreadyExists
+type ResourceAlreadyExistsErrorInfo struct {
```
gozssky, line 46: This name with the `Info` suffix is inconsistent with other places, like `ResourceMetastoreError`, and errors defined in https://github.com/pingcap/tiflow/blob/master/engine/servermaster/scheduler/errors.go.

engine/pkg/externalresource/manager/service.go
```diff
@@ -43,46 +42,65 @@ type Service struct {
...
 	shouldRet, err := s.preRPCHook.PreRPC(ctx, request, &resp2)
 	if shouldRet {
 		return resp2, err
 	}
 
-	record, err := s.metaclient.GetResourceByID(ctx,
-		pkgOrm.ResourceKey{JobID: request.GetResourceKey().GetJobId(), ID: request.GetResourceKey().GetResourceId()})
+	defer func() {
+		if retErr != nil {
+			retErr = rpcerror.ToGRPCError(retErr)
```
gozssky, line 77: https://github.com/pingcap/tiflow/pull/6669 will add a server interceptor to convert error. So we don't need to convert the error manually.

engine/pkg/client/resource_manager_client.go
liuzix: No. It can only happen if the same worker has tried to create the same resource twice.

amyangfei (approved):

gozssky (approved):

PR #6790, opened 2022-08-18
engine(dm): unfresh sync unit no need to request resource 

### LINKED ISSUE #6784: DMWorkerSync should not request resource after metadata is loaded into checkpoint

to provide more freedom for scheduling

as title

_No response_

_No response_

GMHDBJD:

Issue Number: close #6784 

- if sync is unfresh, no need to request resource.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


engine/executor/openapi.go
```diff
@@ -43,13 +45,16 @@ func newJobAPIServer() *jobAPIServer {
 
 func (s *jobAPIServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if h, ok := s.match(r.URL.Path); ok {
+		log.Debug("serve http request", zap.String("path", r.URL.Path))
```
lance6716, line 48: @gozssky can the framework log OpenAPI request/response soon?

lance6716 (approved):
if we need to save table checkpoint for `loadTableStructureFromDump` in future, we can save it before saving global checkpoint. Please left a comment to remind us.

engine/executor/openapi.go
gozssky: I think it will be soon. This log can be kept here. If there is any better way, we can improve it later.

amyangfei (approved):

PR #6795, opened 2022-08-18
sinkv2(ticdc): move mysql config into sink pkg

hi-rustin:

- Move kafka into sink pkg

- Move MySQL config into pkg

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```


cdc/sinkv2/eventsink/txn/mysql/mysql_test.go
```diff
@@ -663,18 +620,15 @@ func TestMysqlSinkNotRetryErrDupEntry(t *testing.T) {
...
-	sinkOpts.dmlMaxRetry = 1
-
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	changefeed := "test-changefeed"
 	sinkURI, err := url.Parse("mysql://127.0.0.1:4000/?time-zone=UTC&worker-count=1&safe-mode=false")
 	require.Nil(t, err)
-	sink, err := NewMySQLBackend(ctx, model.DefaultChangeFeedID(changefeed), sinkURI, sinkOpts)
+	sink, err := NewMySQLBackend(ctx, model.DefaultChangeFeedID(changefeed), sinkURI,
+		config.GetDefaultReplicaConfig(), mockDBInsertDupEntry)
```
hicqu, line 629: `dmlMaxRetry` is customed above.

pkg/sink/mysql/config.go
```diff
@@ -0,0 +1,314 @@
...
+	defaultTiDBTxnMode                = txnModeOptimistic
+	defaultBatchReplaceEnabled        = true
+	defaultBatchReplaceSize           = 20
+	defaultReadTimeout                = "2m"
+	defaultWriteTimeout               = "2m"
+	defaultDialTimeout                = "2m"
+	defaultSafeMode                   = true
+	defaultTxnIsolationRC             = "READ-COMMITTED"
+	defaultCharacterSet               = "utf8mb4"
+	defaultDMLMaxRetry         uint64 = 8
```
hicqu, line 58: `defaultDMLMaxRetry` isn't a generic option for both `txn.MySQLBackand` and `ddl.MySQLBackend`. If we must seperate `MySQLBackend` into 2 things, I prefer to move `defaultDMLMaxRetry` into `cdc/sinkv2/eventsink/txn/...`.

hi-rustin: How about rename it to defaultExectueMaxRetry?

cdc/sinkv2/eventsink/txn/mysql/mysql_test.go
hi-rustin: I moved it to bottom. Because it only for test, so I think it's ok. We have no way to custom it before we create the sink.

pkg/sink/mysql/config.go
hicqu: Seems DDL and DML uses different `maxRetry`.

hi-rustin: Good catch. Updated.

hicqu (approved):

hicqu:
LGTM. Thank you!

hi-rustin (commented):
Self check.

nongfushanquan (approved):

PR #6807, opened 2022-08-19
sinkv2(ticdc): add MySQL DDL sink

hi-rustin:

- Add MySQL DDL sink.
- Add tests

#### Tests 

 - [x] Unit test

No
No
### Release note 

```release-note
None
```

### DIFF
--- a/cdc/sinkv2/ddlsink/factory/factory.go
+++ b/cdc/sinkv2/ddlsink/factory/factory.go
@@ -22,9 +22,11 @@ import (
 	"github.com/pingcap/tiflow/cdc/sinkv2/ddlsink/blackhole"
 	"github.com/pingcap/tiflow/cdc/sinkv2/ddlsink/mq"
 	"github.com/pingcap/tiflow/cdc/sinkv2/ddlsink/mq/ddlproducer"
+	"github.com/pingcap/tiflow/cdc/sinkv2/ddlsink/txn"
 	"github.com/pingcap/tiflow/pkg/config"
 	cerror "github.com/pingcap/tiflow/pkg/errors"
 	"github.com/pingcap/tiflow/pkg/sink"
+	pmysql "github.com/pingcap/tiflow/pkg/sink/mysql"
 )
 
 // New creates a new ddlsink.DDLEventSink by schema.
@@ -38,13 +40,14 @@ func New(
 		return nil, err
 	}
 	schema := strings.ToLower(sinkURI.Scheme)
-	// TODO: add more sink factory here.
 	switch schema {
 	case sink.KafkaSchema, sink.KafkaSSLSchema:
 		return mq.NewKafkaDDLSink(ctx, sinkURI, cfg,
 			kafka.NewAdminClientImpl, ddlproducer.NewKafkaDDLProducer)
 	case sink.BlackHoleSchema:
 		return blackhole.New(), nil
+	case sink.MySQLSSLSchema, sink.MySQLSchema, sink.TiDBSchema, sink.TiDBSSLSchema:
+		return txn.NewMySQLDDLSink(ctx, sinkURI, cfg, pmysql.CreateMySQLDBConn)
 	default:
 		return nil,
 			cerror.ErrSinkURIInvalid.GenWithStack("the sink scheme (%s) is not supported", schema)
--- a/cdc/sinkv2/ddlsink/factory/factory.go
+++ b/cdc/sinkv2/ddlsink/factory/factory.go
@@ -22,7 +22,7 @@ import (
 	"github.com/pingcap/tiflow/cdc/sinkv2/ddlsink/blackhole"
 	"github.com/pingcap/tiflow/cdc/sinkv2/ddlsink/mq"
 	"github.com/pingcap/tiflow/cdc/sinkv2/ddlsink/mq/ddlproducer"
-	"github.com/pingcap/tiflow/cdc/sinkv2/ddlsink/txn"
+	"github.com/pingcap/tiflow/cdc/sinkv2/ddlsink/mysql"
 	"github.com/pingcap/tiflow/pkg/config"
 	cerror "github.com/pingcap/tiflow/pkg/errors"
 	"github.com/pingcap/tiflow/pkg/sink"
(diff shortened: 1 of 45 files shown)


cdc/sinkv2/ddlsink/txn/mysql_ddl_sink.go
```diff
@@ -0,0 +1,200 @@
...
+			return err
+		}
+
+		return tx.Commit()
+	})
+	if err != nil {
+		return cerror.WrapError(cerror.ErrMySQLTxnError, err)
+	}
+
+	log.Info("Exec DDL succeeded", zap.String("sql", ddl.Query),
```
nongfushanquan, line 172: if the time used exceed a value , it's better to record the information for investigation later

hi-rustin: Added. Thanks!

cdc/sinkv2/ddlsink/txn/mysql_ddl_sink.go
```diff
@@ -0,0 +1,206 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package txn
```
liuzix, line 14: Is it better if this file is put into a different package?

cdc/sinkv2/ddlsink/txn/mysql_ddl_sink.go
```diff
@@ -0,0 +1,206 @@
...
+		if _, err = tx.ExecContext(ctx, ddl.Query); err != nil {
+			if rbErr := tx.Rollback(); rbErr != nil {
+				log.Error("Failed to rollback", zap.String("sql", ddl.Query),
+					zap.String("namespace", m.id.Namespace),
+					zap.String("changefeed", m.id.ID), zap.Error(err))
+			}
+			return err
+		}
+
+		return tx.Commit()
```
liuzix, line 167:
```suggestion
		return errors.Trace(tx.Commit())
```

cdc/sinkv2/ddlsink/txn/mysql_ddl_sink.go
hi-rustin: Renamed it to mysql.

cdc/sinkv2/ddlsink/txn/mysql_ddl_sink.go
hi-rustin: Good catch!

cdc/sinkv2/ddlsink/factory/factory.go
```diff
@@ -38,13 +40,14 @@ func New(
...
 	schema := strings.ToLower(sinkURI.Scheme)
-	// TODO: add more sink factory here.
 	switch schema {
 	case sink.KafkaSchema, sink.KafkaSSLSchema:
 		return mq.NewKafkaDDLSink(ctx, sinkURI, cfg,
 			kafka.NewAdminClientImpl, ddlproducer.NewKafkaDDLProducer)
 	case sink.BlackHoleSchema:
 		return blackhole.New(), nil
+	case sink.MySQLSSLSchema, sink.MySQLSchema, sink.TiDBSchema, sink.TiDBSSLSchema:
+		return mysql.NewMySQLDDLSink(ctx, sinkURI, cfg, pmysql.CreateMySQLDBConn)
```
hicqu, line 50: How about remove parameter `pmysql.CreateMySQLDBConn`?

cdc/sinkv2/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -0,0 +1,206 @@
...
+	// We use it to record the DDL count.
+	statistics *metrics.Statistics
+}
+
+// NewMySQLDDLSink creates a new mysqlDDLSink.
+func NewMySQLDDLSink(
+	ctx context.Context,
+	sinkURI *url.URL,
+	replicaConfig *config.ReplicaConfig,
+	dbConnFactory pmysql.Factory,
```
hicqu, line 62: I prefer to remove the parameter. And we can add a new function `newMySQLDDLSink` for tests if necessary.

cdc/sinkv2/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -0,0 +1,206 @@
...
+
+	db, err := dbConnFactory(ctx, dsnStr)
+	if err != nil {
+		return nil, err
+	}
+
+	m := &mysqlDDLSink{
+		id:         changefeedID,
+		db:         db,
+		statistics: metrics.NewStatistics(ctx, sink.TxnSink),
```
hicqu, line 84: It's better to use a sub-context `ctx1, cancel = context.WithCancel(ctx)`, and call `cancel` when close the sink.

hicqu:
Rest LGTM. @hi-rustin thank you!

cdc/sinkv2/ddlsink/mysql/mysql_ddl_sink.go
hi-rustin: It will be canceled at:
```go
func (s *ddlSinkImpl) close(ctx context.Context) (err error) {
	s.cancel()
	if s.sinkV1 != nil {
		err = s.sinkV1.Close(ctx)
	} else {
		err = s.sinkV2.Close()
	}
	if s.syncPointStore != nil {
		err = s.syncPointStore.Close()
	}
	s.wg.Wait()
		return err
	}
	return nil
}
```

cdc/sinkv2/ddlsink/factory/factory.go
hi-rustin: I think this is acceptable because it is the internal method. And kafka does the same thing. Creating a new constructor would allow us to maintain the exact same code every time.

cdc/sinkv2/ddlsink/mysql/mysql_ddl_sink.go
hi-rustin: See: https://github.com/pingcap/tiflow/pull/6807#discussion_r950029311

cdc/sinkv2/ddlsink/factory/factory.go
hi-rustin: The ability to walk the real code path during testing is also a major advantage of this approach.

hi-rustin: We got tons of config processes in the constructor, so if we can walk these codes it would be better. And we only need to maintain one copy of the code.

nongfushanquan (approved):

hicqu (approved):

hi-rustin:
/merge

Issue #6817, opened 2022-08-19
engien OpenAPI style improvement


lance6716:
### Is your feature request related to a problem?

(only an internal doc, TBD)

### Describe the feature you'd like

(only an internal doc, TBD)

lance6716:
I think we can define a new ProcessResult struct with correct json tags, and convert it from DM's pb.ProcessResult by toml/yaml serde

lance6716:
```
curl "http://127.0.0.1:10245/api/v1/jobs/6c256185-ec52-467b-bf1a-35fd351047c0/status"

{"code":9,"details":[{"@type":"type.googleapis.com/enginepb.ErrorV2","name":"JobNotRunningError","details":"eyJKb2JJRCI6IjZjMjU2MTg1LWVjNTItNDY3Yi1iZjFhLTM1ZmQzNTEwNDdjMCJ9","stack_trace":"\ngithub.com/pingcap/tiflow/engine/pkg/rpcerror.(*Prototype[...]).GenWithStack\n\tgithub.com/pingcap/tiflow/engine/pkg/rpcerror/normalize.go:101\ngithub.com/pingcap/tiflow/engine/servermaster.(*JobManagerImpl).GetJobMasterForwardAddress\n\tgithub.com/pingcap/tiflow/engine/servermaster/jobmanager.go:365\ngithub.com/pingcap/tiflow/engine/servermaster.(*Server).handleForwardJobAPI\n\tgithub.com/pingcap/tiflow/engine/servermaster/server.go:695\ngithub.com/pingcap/tiflow/engine/servermaster.(*Server).forwardJobAPI\n\tgithub.com/pingcap/tiflow/engine/servermaster/server.go:668\ngithub.com/pingcap/tiflow/engine/servermaster.registerRoutes.func1\n\tgithub.com/pingcap/tiflow/engine/servermaster/http.go:48\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2109\nnet/http.(*ServeMux).ServeHTTP\n\tnet/http/server.go:2487\nnet/http.serverHandler.ServeHTTP\n\tnet/http/server.go:2947\nnet/http.(*conn).serve\n\tnet/http/server.go:1991\nruntime.goexit\n\truntime/asm_amd64.s:1594"}]}
```

not sure this is expected for web frontend

gozssky:
Fixed by https://github.com/pingcap/tiflow/pull/7516.

PR #6818, opened 2022-08-19
openapi(engine): improve OpenAPI readablity

lance6716:

Issue Number: ref #6817

the visual change can be verified in `engine/jobmaster/dm/api_test.go`

#### Tests 

 - Unit tests

### Release note 

```release-note
 `None`.
```


engine/executor/dm/worker.go
```diff
@@ -215,7 +215,7 @@ func (w *dmWorker) tryUpdateStatus(ctx context.Context) error {
 	if currentStage == previousStage {
 		return nil
 	}
-	w.Logger().Info("task stage changed", zap.String("task-id", w.taskID), zap.Int("from", int(previousStage)), zap.Int("to", int(currentStage)))
+	w.Logger().Info("task stage changed", zap.String("task-id", w.taskID), zap.String("from", string(previousStage)), zap.String("to", string(currentStage)))
```
GMHDBJD, line 218: Can we use zap.Stringer?

GMHDBJD (approved):
LGTM

engine/executor/dm/worker.go
lance6716: I haven't added String() for this type

engine/jobmaster/dm/api.go
```diff
@@ -29,17 +29,17 @@ import (
...
+	ExpectedStage  metadata.TaskStage         `json:"expected_stage"`
+	WorkerID       frameModel.WorkerID        `json:"worker_id"`
+	ConfigOutdated bool                       `json:"config_outdated"`
+	Status         *dmpkg.QueryStatusResponse `json:"status"`
 }
 
 // JobStatus represents status of a job
 type JobStatus struct {
-	JobID frameModel.MasterID
+	JobID frameModel.MasterID `json:"job_id"`
```
amyangfei, line 40: Is there any regulation that we follow about using `hyphen` or `underscore` in json key.

amyangfei (approved):

PR #6836, opened 2022-08-22
schedulerV3(ticdc): burst add table in a batch way to delay resource allocation

### LINKED ISSUE #6828: all tables dispatched to one nodes after recover from out of memory
As you can see from the monitoring screen below, all tables dispatched to one cdc instance, even though there are 2 available.

This makes the node who carry all tables with heavy burden, more easily to be crashed again.

2 cdc instance，2 changefeed, 300 tables, 600 tables at total, only around 16G memory for each cdc instance.

image
image
image
image

3AceShowHand:

Issue Number: close #6828

add `batchSize` to the `basicScheduler`, each tick add table no more than the batch.

#### Tests 

 - Unit test

### Release note 

```release-note
`None`
```


cdc/scheduler/internal/v3/scheduler/scheduler.go
```diff
@@ -42,3 +42,14 @@ const (
...
+
+const (
+	// defaultBurstAddTableBatchSize is the default batch size of adding tables on each tick,
+	// used by the `BasicScheduler`.
+	// When the owner crashed, it may exist hundreds of tables need to be added by the new owner,
+	// the crashed old owner is going online and not available yet, add tables in batch instead of
+	// add all to captures to wait for the old owner join the cluster .
+	// When there are only 2 captures, and a large number of tables, this can be helpful to prevent
+	// oom caused by all tables dispatched to only one capture.
+	defaultBurstAddTableBatchSize = 50
```
overvenus, line 54: Please make it configurable.

overvenus (approved):

asddongmen (approved):

PR #6848, opened 2022-08-22
syncer(dm): Fix dm-worker keeps retrying to execute ddl when encountering "invalid connection" error 

### LINKED ISSUE #4689: dm-worker keeps retrying to execute ddl when encounter "invalid connection" error

lyzx2001:

Issue Number: close #4689 

What is changed to fix this issue:
To resolve the issue of dm-worker keeps retrying to execute ddl when encounter "invalid connection" error, dm-worker need to handle the error according to the retrieved ddl status from TiDB history jobs. According to https://github.com/pingcap/tidb/blob/2e16291b2d48279e709feebed7a29a53337dc74c/parser/model/ddl.go#L808, the status can be: "running", "rollingback", "rollback done", "rollback done", "cancelled", "cancelling", "synced", "queueing", and "none". Thus in my implementation, the error is handled according to these possible statuses.
For "running", "queueing", and "none", the solution is to use for loop to keep checking the status until it turns to "synced" or "cancelled" or "done", which can prevent dm-worker to re-execute the ddl every 5 minutes. For "cancelled", simply return the original error.

How the integration test works:
Since it is hard to simulate the real error of encountering the "invalid connection" error after 5 minutes (need to create a very large table), here in the integration test I use failpoint to assign mysql2.ErrInvalidConn to err. Since the handle method will depend on the status of ddl retrieved from the TiDB history jobs, I also use failpoint to simulate several possible status of the actual return status, including "synced", "running", "cancelled", "queueing", and "none".
Since failpoint is only checked when dm-worker is started, to test the above 5 situations, I need to repeat the process (export failpoint, start dm-worker, and stop dm-worker) for 5 times.

Why the integration test can prove that the issue has been fixed:
Before the issue is fixed, running the integration test would result in multiple re-executed ddl in the TiDB history job queues, indicating that dm-worker keeps retrying to execute ddl when encounter "invalid connection" error, and the following ddls are blocked:
```
mysql> ADMIN SHOW DDL JOB QUERIES LIMIT 6 OFFSET 0;
+--------+---------------------------------------------------------------------------------------------+
| JOB_ID | QUERY                                                                                       | 
+--------+---------------------------------------------------------------------------------------------+
|     58 | ALTER TABLE gbk.invalid_conn_test1 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
|     56 | ALTER TABLE gbk.invalid_conn_test1 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
|     54 | ALTER TABLE gbk.invalid_conn_test1 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
|     52 | ALTER TABLE gbk.invalid_conn_test1 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
|     50 | ALTER TABLE gbk.invalid_conn_test1 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
|     48 | ALTER TABLE gbk.invalid_conn_test1 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
+--------+---------------------------------------------------------------------------------------------+
5 rows in set (0.00 sec)
```
After the issue is fixed, running the integration test would result in no re-executed ddl in the TiDB history job queues, indicating that dm-worker has handled the error according to the ddl status, and the following ddls can be executed successfully:
```
mysql> ADMIN SHOW DDL JOB QUERIES LIMIT 6 OFFSET 0;
+--------+---------------------------------------------------------------------------------------------+
| JOB_ID | QUERY                                                                                       | 
+--------+---------------------------------------------------------------------------------------------+
|     58 | ALTER TABLE gbk.invalid_conn_test6 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
|     56 | ALTER TABLE gbk.invalid_conn_test5 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
|     54 | ALTER TABLE gbk.invalid_conn_test4 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
|     52 | ALTER TABLE gbk.invalid_conn_test3 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
|     50 | ALTER TABLE gbk.invalid_conn_test2 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
|     48 | ALTER TABLE gbk.invalid_conn_test1 MODIFY COLUMN n SMALLINT(4) NOT NULL DEFAULT _UTF8MB4'0' | 
+--------+---------------------------------------------------------------------------------------------+
5 rows in set (0.00 sec)
```
 A small issue exists that I tried to output the TiDB history jobs by executing ```ADMIN SHOW DDL JOB QUERIES LIMIT 20 OFFSET 0``` to confirm the results, but since the process of dm takes some time, the current integration test output can only show the earliest ddl instead of the all the following ddls. But when I manually check the history jobs in TiDB, the result is like the above displayed one, which meets my expectation.

#### Tests 

 - Unit test
 - Integration test

No.
No.
### Release note 

```release-note
Fix the issue that dm-worker keeps retrying to execute ddl for every 5 minutes when encountering "invalid connection" error.
```


dm/syncer/error.go
```diff
@@ -211,13 +212,66 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
 	}
 	// TODO: add support for downstream alter pk without schema
 
+	// it handles the operations of DDL when encountering `invalid connection`
+	invalidConnF2 := func(tctx *tcontext.Context, err error, ddls []string, index int, conn *dbconn.DBConn, createTime int64) error {
```
lance6716, line 216: there are two ways to let `invalidConnF2` be able to use `createTime`, current code is adding a parameter, the other way is directly using `createTime` because `invalidConnF2` is a closure thus it can use implicit variables names in its surrounding scope.

It's OK for current code

dm/syncer/error.go
```diff
@@ -72,7 +73,7 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
 	// if we have other methods to judge the DDL dispatched but timeout for executing, we can update this method.
 	// NOTE: we must ensure other PK/UK exists for correctness.
 	// NOTE: when we are refactoring the shard DDL algorithm, we also need to consider supporting non-blocking `ADD INDEX`.
-	invalidConnF := func(tctx *tcontext.Context, err error, ddls []string, index int, conn *dbconn.DBConn) error {
+	invalidConnF := func(tctx *tcontext.Context, err error, ddls []string, index int, conn *dbconn.DBConn, createTime int64) error {
```
lance6716, line 76: invalidConnF and invalidConnF2 can have a better name, like ignoreAddIndexTimeout, waitAsyncDDL

dm/syncer/error.go
```diff
@@ -211,13 +212,66 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
 	}
 	// TODO: add support for downstream alter pk without schema
 
+	// it handles the operations of DDL when encountering `invalid connection`
+	invalidConnF2 := func(tctx *tcontext.Context, err error, ddls []string, index int, conn *dbconn.DBConn, createTime int64) error {
+		if len(ddls) == 0 || index > len(ddls)-1 || errors.Cause(err) != mysql.ErrInvalidConn || createTime == -1 {
```
lance6716, line 217: the special value of `createTime == -1` should be put in comment of handleSpecialDDLError

dm/syncer/error.go
```diff
@@ -211,13 +212,66 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+			return err // return the original error
+		}
+
+		statusChan := make(chan string)
```
lance6716, line 221: the read and write of `statusChan` is in same goroutine so no need to use channel. channel is more often used by multiple goroutines communication

dm/syncer/error.go
```diff
@@ -211,13 +212,66 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+		var status string
+		var err2 error
+
+		for {
+			status, err2 = getDDLStatusFromTiDB(tctx, conn, ddls[index], createTime) // status is synced
```
lance6716, line 226:
```suggestion
			status, err2 := getDDLStatusFromTiDB(tctx, conn, ddls[index], createTime) // status is synced
```

no need to declare them outside the loop

dm/syncer/error.go
```diff
@@ -211,13 +212,66 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+			failpoint.Inject("TestStatusRunning", func() {
```
lance6716, line 227: you can use one failpoint with parameter, so different string type status can be passed in

```
failpoint.Inject("...", func(val failpoint.Value) {
		status := val.(string)
		...
	})
```

dm/syncer/error.go
```diff
@@ -211,13 +212,66 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+			})
+			failpoint.Inject("TestStatusNone", func() {
+				status = "none"
+			})
+			if err != nil {
+				return err2
+			}
+			statusChan <- status
+			select {
+			case <-tctx.Ctx.Done():
```
lance6716, line 244: generally we only need to check context.Done along with slow IO, such as network IO. No need to use `select` to check memory operations.

dm/syncer/error.go
```diff
@@ -211,13 +212,66 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+				return nil
+			case status = <-statusChan:
+				switch status {
+				case "running":
```
lance6716, line 248: case "running", "rollingback"

case can have multiple matching pattern

dm/syncer/error.go
```diff
@@ -211,13 +212,66 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+				case "rollingback":
```
lance6716, line 249: these error status can also return `err`, no need to wait them becoming "cancelled"

dm/syncer/error.go
```diff
@@ -211,13 +212,66 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+				case "rollback done":
+					return nil
+				case "done":
+					return nil
+				case "cancelled":
+					return err
```
lance6716, line 255: should not return `err` because `err` is "invalid connection", we should return something like "DDL %s executed in background and met error". And can we find the error from TiDB?

dm/syncer/error.go
lyzx2001: I see that other variables like `tctx`, `err`, `ddls`, `index`, `conn` are all passed through parameters, so here my thought is to conform with them by also passing `createTime` through parameter.

dm/tests/gbk/run.sh
```diff
@@ -67,6 +67,199 @@ function run() {
...
+
+	echo "prepare data for invalid connection test"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test1(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test2(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test3(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test4(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test5(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test6(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+
+	dmctl_stop_task $cur/conf/dm-task.yaml "--remove-meta"
```
lance6716, line 79: stop-task don't have "remove-meta", and no need to stop the task and source if we will add them again

lyzx2001: I tried to kill dm-worker without stopping the task, but the returned error said that dm-worker cannot be killed because there were still tasks running

lance6716: where do you see `returned error`? `kill` is a linux command, not related to DM's error

dm/syncer/syncer.go
```diff
@@ -1388,8 +1390,21 @@ func (s *Syncer) syncDDL(queueBucket string, db *dbconn.DBConn, ddlJobChan chan
...
 			affected, err = db.ExecuteSQLWithIgnore(s.syncCtx, s.metricsProxies, errorutil.IsIgnorableMySQLDDLError, ddlJob.ddls)
+			failpoint.Inject("TestHandleSpecialDDLError", func() {
+				err = mysql2.ErrInvalidConn
+				affected = 1
+				s.tctx.L().Info("test handle special DDL error", zap.Strings("DDL", ddlJob.ddls), zap.String("affected", fmt.Sprint(affected)), zap.String("failpoint", "TestHandleSpecialDDLError"))
+			})
 			if err != nil {
-				err = s.handleSpecialDDLError(s.syncCtx, err, ddlJob.ddls, affected, db)
+				row, err2 := db.QuerySQL(s.syncCtx, s.metricsProxies, "SELECT @@TIMESTAMP")
+				createTimeResults, err3 := export.GetSpecifiedColumnValuesAndClose(row, "@@TIMESTAMP")
```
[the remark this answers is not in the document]
lyzx2001: The result from `export.GetSpecifiedColumnValuesAndClose` is `string` whose content is `float`, and to compare it with the create time of the required DDL which is ` int64`, we need to convert the current result to `int64` as well. so here I first use `strconv.ParseFloat` to convert the result to `float64`, then use `strconv.Atoi` with `"%1.0f"` to further convert it to `int64`.

dm/tests/_utils/count_contains
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -eu
+
+echo "start check count"
+echo "$TEST_DIR/sql_res.$TEST_NAME.txt"
+count=$(cat "$TEST_DIR/sql_res.$TEST_NAME.txt" | grep -c "$1") || true
+echo "OUTPUT CONTAINS $count '$1'"
```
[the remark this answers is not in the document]
lyzx2001: The output `count` should be equal to 1, which means dm-worker is not re-executing ddls.

dm/tests/gbk/run.sh
```diff
@@ -67,6 +67,199 @@ function run() {
...
 
 	test_from_tidb
+
+	echo "prepare data for invalid connection test"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test1(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test2(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test3(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test4(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test5(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test6(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
```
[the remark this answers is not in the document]
lyzx2001: Maybe reduce to 2 DDL is good enough, since we need to verify that the following ddls can be executed instead of still re-executing the same ddl

dm/tests/gbk/run.sh
lyzx2001: Fixed. Have deleted the part of stopping task and source and creating them again, and can work correctly.

dm/syncer/syncer.go
```diff
@@ -1387,9 +1389,37 @@ func (s *Syncer) syncDDL(queueBucket string, db *dbconn.DBConn, ddlJobChan chan
 
 		if !ignore {
 			var affected int
+			row, err2 := db.QuerySQL(s.syncCtx, s.metricsProxies, "SELECT @@TIMESTAMP")
```
lance6716, line 1392: there are functions like `select UNIX_TIMESTAMP();` which returns a int32/int64 value.

And `GetSpecifiedColumnValuesAndClose` is used to get specific columns from result with unknown number of columns. In this function we can use `Scan` to avoid strconv.

And considering the convention of getDDLStatusFromTiDB

```
// createTime should be based on the timezone of downstream, and its unit is second.
```

You can also be careful about the timezone.

dm/tests/_utils/count_contains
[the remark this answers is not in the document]
lance6716: seems there's no code to check the count is 1

dm/tests/gbk/run.sh
```diff
@@ -67,6 +67,105 @@ function run() {
...
 
 	test_from_tidb
+
+	echo "prepare data for invalid connection test"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test1(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+	run_sql_source1 "CREATE TABLE IF NOT EXISTS gbk.invalid_conn_test2(i TINYINT, j INT UNIQUE KEY, k TINYINT, m TINYINT, n TINYINT)"
+
+	kill_dm_worker
+
+	# test invalid connection with status synced
```
lance6716, line 77: I came up with two simple tests that don't need to inject DDL status:

1. inject TestHandleSpecialDDLError, start DM-worker, and execute **one** DDL and insert some data. DM-worker will see injected error and run into `waitAsyncDDL`. Then it will see the DDL is "synced" so replication continues, the data is soon replicated and we can use sync-diff to check the data.
2. inject TestHandleSpecialDDLError, start DM-worker, make a column contains duplicated data and execute **one** DDL that add UNIQUE on the column. DM-worker will see the DDL is "rollbacked" and the task should be paused with ErrCancelledDDL. Then we use `run_dmctl_with_retry` to check the result of query-status.

lyzx2001: I have put the code for checking whether count is equal to 1 to run.sh instead of count_contains, because count_contains is meant to check the number of appearances instead of justifying whether the count is equal to a certain number. This part should be done in run.sh that call this function.

dm/syncer/error.go
```diff
@@ -211,13 +217,60 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+			failpoint.Inject("TestStatus", func(val failpoint.Value) {
+				status = val.(string)
+				s.tctx.L().Info("TestStatus:", zap.String("TestStatus", status))
+			})
+			fmt.Printf("status: %s \n", status)
+			s.tctx.L().Info("statusLog:", zap.String("statusLog", status))
+			select {
+			case <-tctx.Ctx.Done():
+				return nil
+			case <-ticker.C:
```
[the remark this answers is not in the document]
lyzx2001: Modified. All the tests work as expected now.

dm/tests/gbk/run.sh
```diff
@@ -67,6 +67,218 @@ function run() {
...
+
+	run_sql_tidb_with_retry "ADMIN SHOW DDL JOB QUERIES LIMIT 10 OFFSET 0" "ALTER TABLE \`gbk\`.\`invalid_conn_test1\` MODIFY COLUMN \`n\` SMALLINT(4) NOT NULL DEFAULT '0'"
+	echo "check count"
+	check_count "ALTER TABLE \`gbk\`.\`invalid_conn_test1\` MODIFY COLUMN \`n\` SMALLINT(4) NOT NULL DEFAULT '0'" 1
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+	echo "check test invalid connection with status none successfully"
+
+	kill_dm_worker
+
+	# test inserting data after invalid connection
```
lance6716, line 216: we can remove test case in line 77 since this one totally covers it.

dm/tests/gbk/run.sh
```diff
@@ -67,6 +67,218 @@ function run() {
...
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"resume-task gbk" \
+		"\"result\": true" 3
+	echo "check test adding UNIQUE on column with duplicate data successfully"
+
+	dmctl_stop_task $cur/conf/dm-task.yaml
+	dmctl_operate_source stop $WORK_DIR/source1.yaml $SOURCE_ID1
+	dmctl_operate_source stop $WORK_DIR/source2.yaml $SOURCE_ID2
+
+	kill_dm_worker
```
lance6716, line 281:
```suggestion
```

these line checks nothing

dm/syncer/syncer.go
```diff
@@ -1387,9 +1388,29 @@ func (s *Syncer) syncDDL(queueBucket string, db *dbconn.DBConn, ddlJobChan chan
...
+				err2 = row.Scan(&ddlCreateTime)
+				if err2 != nil {
+					s.tctx.L().Warn("getting ddlCreateTime failed", zap.String("error", err2.Error()))
+				}
+			}
+			row.Close()
 			affected, err = db.ExecuteSQLWithIgnore(s.syncCtx, s.metricsProxies, errorutil.IsIgnorableMySQLDDLError, ddlJob.ddls)
+			failpoint.Inject("TestHandleSpecialDDLError", func() {
+				err = mysql2.ErrInvalidConn
+				affected = 1
```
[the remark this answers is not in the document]
lyzx2001: I have changed it to, when the original index of `ddlJob.ddls` is larger than 2 (the status is injected like "running" instead of caused by real error like "cancelled"), assign `affected` to index minus 2 due to adding the SET SQL of timezone and timestamp.

dm/syncer/error.go
```diff
@@ -211,13 +216,53 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+
+		duration := 30
+		failpoint.Inject("ChangeDuration", func() {
+			duration = 1
+		})
+		ticker := time.NewTicker(time.Duration(duration) * time.Second)
+		defer ticker.Stop()
+
+		for {
+			status, err2 := getDDLStatusFromTiDB(tctx, conn, ddls[index], createTime)
```
lance6716, line 233: If downstream TiDB does not support this SQL syntax (maybe an old version of TiDB), we should not enter dead loop

dm/syncer/syncer.go
```diff
@@ -1387,9 +1388,31 @@ func (s *Syncer) syncDDL(queueBucket string, db *dbconn.DBConn, ddlJobChan chan
 
 		if !ignore {
 			var affected int
+			var ddlCreateTime int64
+			ddlCreateTime = -1 // default when scan failed
```
lance6716, line 1392:
```suggestion
			var ddlCreateTime int64 = -1
```

dm/tests/gbk/run.sh
lance6716: Don't forget this comment

dm/tests/gbk/run.sh
```diff
@@ -67,6 +67,137 @@ function run() {
...
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"query-status gbk" \
+		"origin SQL: \[ALTER TABLE gbk.invalid_conn_test1 ADD UNIQUE(i)\]: DDL ALTER TABLE \`gbk\`.\`invalid_conn_test1\` ADD UNIQUE(\`i\`) executed in background and met error" 1
+	echo "check test adding UNIQUE on column with duplicate data successfully"
+
+	dmctl_stop_task $cur/conf/dm-task.yaml
+	dmctl_operate_source stop $WORK_DIR/source1.yaml $SOURCE_ID1
+	dmctl_operate_source stop $WORK_DIR/source2.yaml $SOURCE_ID2
+
+	kill_dm_worker
```
lance6716, line 200:
```suggestion
```

lance6716 (commented):
rest lgtm

/cc @GMHDBJD

dm/syncer/error.go
```diff
@@ -211,13 +216,53 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+				status = val.(string)
+				s.tctx.L().Info("injected test status:", zap.String("TestStatus", status))
+			})
+			switch status {
+			case model.JobStateDone.String(), model.JobStateSynced.String():
+				return nil
+			case model.JobStateCancelled.String(), model.JobStateRollingback.String(), model.JobStateRollbackDone.String(), model.JobStateCancelling.String():
+				return terror.ErrCancelledDDL.Generate(ddls[index])
+			case model.JobStateRunning.String(), model.JobStateQueueing.String(), model.JobStateNone.String():
+			default:
```
lance6716, line 247: What's the value of `status` and what should we do if `getDDLStatusFromTiDB` does not find the DDL in downstream?

dm/pkg/terror/error_list.go
```diff
@@ -195,6 +195,9 @@ const (
 	// pkg/shardddl/optimism.
 	codeShardDDLOptimismNeedSkipAndRedirect
 	codeShardDDLOptimismAddNotFullyDroppedColumn
+
+	// syncer.
+	codeCancelledDDL
```
GMHDBJD, line 200: Will move to sync unit error code(36000+) be better?

dm/syncer/error.go
```diff
@@ -211,13 +216,53 @@ func (s *Syncer) handleSpecialDDLError(tctx *tcontext.Context, err error, ddls [
...
+			status, err2 := getDDLStatusFromTiDB(tctx, conn, ddls[index], createTime)
+			if err2 != nil {
+				s.tctx.L().Warn("error when getting DDL status fromTiDB", zap.String("error", err2.Error()))
+			}
+			failpoint.Inject("TestStatus", func(val failpoint.Value) {
+				status = val.(string)
+				s.tctx.L().Info("injected test status:", zap.String("TestStatus", status))
+			})
+			switch status {
+			case model.JobStateDone.String(), model.JobStateSynced.String():
```
GMHDBJD, line 242: Can we return JobState in getDDLStatusFromTiDB rather then string?

dm/syncer/syncer.go
```diff
@@ -1387,9 +1388,31 @@ func (s *Syncer) syncDDL(queueBucket string, db *dbconn.DBConn, ddlJobChan chan
...
+			row, err2 := db.QuerySQL(s.syncCtx, s.metricsProxies, "SELECT UNIX_TIMESTAMP()")
+			if err2 != nil {
+				s.tctx.L().Warn("selecting unix timestamp failed", zap.String("error", err2.Error()))
```
GMHDBJD, line 1395: use zap.Error()

dm/syncer/syncer.go
```diff
@@ -1387,9 +1388,31 @@ func (s *Syncer) syncDDL(queueBucket string, db *dbconn.DBConn, ddlJobChan chan
 
 		if !ignore {
 			var affected int
+			var ddlCreateTime int64
+			ddlCreateTime = -1 // default when scan failed
+			row, err2 := db.QuerySQL(s.syncCtx, s.metricsProxies, "SELECT UNIX_TIMESTAMP()")
```
GMHDBJD, line 1393: Can we use QueryRowContext?

GMHDBJD (commented):
rest LGTM

dm/syncer/error.go
lyzx2001: I changed it to if the retrieved status is not the other known possibilities, return the original error.

dm/syncer/error.go
lyzx2001: If `getDDLStatusFromTiDB` does not find the DDL in downstream, the retrieved status would be "", we can just return the original error. dm-worker would re-execute the DDL.

dm/pkg/terror/error_list.go
lyzx2001: I have renamed the error to `codeSyncerCancelledDDL` and moved it to sync unit error code.

dm/syncer/error.go
lyzx2001: I think it may not be necessary. In `getDDLStatusFromTiDB` the original retrieved status is in the form of string, we do not need to manually convert it to `JobState`, which may slow down the performance. Besides, if it cannot find the status in downstream, currently it would return "", but if we change the return type to `JobState`, we may not be able to find a suitable `JobState` to return.

dm/syncer/syncer.go
lyzx2001: Here `db` is `dbconn.DBConn`, it is not convenient to use `QueryRowContext`

GMHDBJD (approved):
LGTM

lance6716:
/merge

the follow-up is testing multi-schema change format records in SHOW DDL @lyzx2001

Issue #6749, opened 2022-08-15
Tracking list for cancel job feature


amyangfei:
### Is your feature request related to a problem?

Currently we have a simple pause job feature in server master, which relies on the availability of job master, but it is not robust in many scenarios.

### Describe the feature you'd like

- [ ] Design the independent job status layer, between upper logic and underlying master worker framework. The feature of job status layer contains
   - [x] Explicit job status, which will be exposed to OpenAPI
   - [x] Decouple its logic from master worker framework
   - [ ] Provide an intermedium to translate job status in master worker framework to job status layer.
- [x] Provide circuit breaker in case of job master restarts and fails for several times during short time.
- [x] Provide `StopImpl` callback in master worker framework.
- [ ] Add a way to cancel hanging job from the runtime perspective. (*hanging job* is a running job, it receives `OnCancel` callback but keeps running and doesn't exit.

#### Normal path

![image](https://user-images.githubusercontent.com/1527315/185560669-8de26b63-2ae9-4219-8f42-12535528c021.png)
source: https://www.plantuml.com/plantuml/uml/SyfFKj2rKt3CoKnELR1Io4ZDoSa70000

#### Fail path

![image](https://user-images.githubusercontent.com/1527315/185562387-92095a32-ae97-4318-aa08-b31c55385c9c.png)
source: https://www.plantuml.com/plantuml/uml/SyfFKj2rKt3CoKnELR1Io4ZDoSa70000

lance6716:
please improve priority @amyangfei

PR #6852, opened 2022-08-23
framework(engine): add StopImpl callback in master impl

amyangfei:

Issue Number: ref #6749

This PR implements the normal path of job canceling, and adds `StopImpl` in job master callback

#### Review guide

- changes in `engine/framework/internal/eventloop`, add `Stop` interface to task
- changes in `engine/framework`, add `StopImpl`interface to MasterImpl.
- changes in `engine/pkg/orm`, support orm for new `job_ops` table.
- `engine/servermaster/jobop/operator.go`, a simple job operator, it provides `MarkJobCanceling` API and `Tick` entrance.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/engine/framework/base_jobmaster.go
+++ b/engine/framework/base_jobmaster.go
@@ -24,6 +24,7 @@ import (
 	"go.uber.org/zap"
 
 	runtime "github.com/pingcap/tiflow/engine/executor/worker"
+	"github.com/pingcap/tiflow/engine/framework/internal/eventloop"
 	frameModel "github.com/pingcap/tiflow/engine/framework/model"
 	"github.com/pingcap/tiflow/engine/model"
 	dcontext "github.com/pingcap/tiflow/engine/pkg/context"
@@ -118,10 +119,9 @@ type JobMasterImpl interface {
 
 	// Workload return the resource unit of the job master itself
 	Workload() model.RescUnit
-
-	// OnJobManagerMessage is called when receives a message from jobmanager
-	OnJobManagerMessage(topic p2p.Topic, message interface{}) error
-
+	// OnCancel is triggered when a cancel message is received. It can be
+	// triggered multiple times.
+	OnCancel(ctx context.Context) error
 	// OnOpenAPIInitialized is called when the OpenAPI is initialized.
 	// This is used to for JobMaster to register its OpenAPI handler.
 	// The implementation must not retain the apiGroup. It must register
@@ -150,7 +150,7 @@ func NewBaseJobMaster(
 	baseMaster := NewBaseMaster(
 		ctx, &jobMasterImplAsMasterImpl{jobMasterImpl}, workerID, tp)
 	baseWorker := NewBaseWorker(
-		ctx, &jobMasterImplAsWorkerImpl{jobMasterImpl}, workerID, masterID, tp, workerEpoch)
+		ctx, &jobMasterImplAsWorkerImpl{inner: jobMasterImpl}, workerID, masterID, tp, workerEpoch)
 	errCenter := errctx.NewErrCenter()
 	baseMaster.(*DefaultBaseMaster).errCenter = errCenter
 	baseWorker.(*DefaultBaseWorker).errCenter = errCenter
@@ -250,8 +250,28 @@ func (d *DefaultBaseJobMaster) Close(ctx context.Context) error {
 	return nil
 }
 
+// Stop implements BaseJobMaster.Stop
+func (d *DefaultBaseJobMaster) Stop(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, time.Second*5)
+	defer cancel()
+
+	if err := d.impl.StopImpl(ctx); err != nil {
+		d.Logger().Error("Failed to stop JobMasterImpl", zap.Error(err))
+	}
+	d.master.doClose()
+	d.worker.doClose()
+	return nil
+}
+
 // NotifyExit implements BaseJobMaster interface
 func (d *DefaultBaseJobMaster) NotifyExit(ctx context.Context, errIn error) (retErr error) {
+	if eventloop.IsTerminatedError(errIn) {
+		// In terminate scenario job master should call StopImpl, and we don't
+		// call NotifyExit to advance the calling of StopImpl. The drawback of
+		// this choice is job manager has to hearbeat timeout of this job master.
+		return nil
+	}
+
 	d.closeOnce.Do(func() {
 		err := d.impl.CloseImpl(ctx)
 		if err != nil {
@@ -350,7 +370,8 @@ func (d *DefaultBaseJobMaster) TriggerOpenAPIInitialize(apiGroup *gin.RouterGrou
 func (d *DefaultBaseJobMaster) IsBaseJobMasterExt() {}
 
 type jobMasterImplAsWorkerImpl struct {
-	inner JobMasterImpl
+	inner          JobMasterImpl
+	onCancelCalled bool
 }
 
 func (j *jobMasterImplAsWorkerImpl) InitImpl(ctx context.Context) error {
@@ -367,8 +388,24 @@ func (j *jobMasterImplAsWorkerImpl) Workload() model.RescUnit {
 	return j.inner.Workload()
 }
 
-func (j *jobMasterImplAsWorkerImpl) OnMasterMessage(topic p2p.Topic, message interface{}) error {
-	return j.inner.OnJobManagerMessage(topic, message)
+func (j *jobMasterImplAsWorkerImpl) OnMasterMessage(
+	ctx context.Context, topic p2p.Topic, message interface{},
+) error {
+	switch msg := message.(type) {
+	case *frameModel.StatusChangeRequest:
+		switch msg.ExpectState {
+		case frameModel.WorkerStateStopped:
+			if !j.onCancelCalled {
+				j.onCancelCalled = true
+				return j.inner.OnCancel(ctx)
+			}
+		default:
+			log.Info("Ignore status change state", zap.Int32("state", int32(msg.ExpectState)))
+		}
+	default:
+		log.Info("unsupported message", zap.Any("message", message))
+	}
+	return nil
 }
 
 func (j *jobMasterImplAsWorkerImpl) CloseImpl(ctx context.Context) error {
@@ -416,5 +453,10 @@ func (j *jobMasterImplAsMasterImpl) OnWorkerMessage(worker WorkerHandle, topic p
 
 func (j *jobMasterImplAsMasterImpl) CloseImpl(ctx context.Context) error {
 	log.Panic("unexpected Close call")
+	return nil
+}
+
+func (j *jobMasterImplAsMasterImpl) StopImpl(ctx context.Context) error {
+	log.Panic("unexpected StopImpl call")
 	return nil
 }
--- a/engine/framework/internal/eventloop/runner.go
+++ b/engine/framework/internal/eventloop/runner.go
@@ -67,8 +67,15 @@ func (r *Runner[R]) Run(ctx context.Context) error {
 		r.doGracefulExit(ctx, err)
 	}
 
-	if closeErr := r.task.Close(context.Background()); closeErr != nil {
-		log.Warn("Closing task returned error", zap.String("label", r.task.ID()))
+	if IsTerminatedError(err) {
+		if stopErr := r.task.Stop(context.Background()); stopErr != nil {
+			log.Warn("Stop task returned error",
+				zap.String("label", r.task.ID()), zap.Error(stopErr))
+		}
+	} else {
+		if closeErr := r.task.Close(context.Background()); closeErr != nil {
+			log.Warn("Closing task returned error", zap.String("label", r.task.ID()))
+		}
 	}
 
 	return err
@@ -117,4 +124,12 @@ func isForcefulExitError(errIn error) bool {
 
 	// Suicides should result in a forceful exit.
 	return derrors.ErrWorkerSuicide.Equal(errIn)
+}
+
+// IsTerminatedError checks whether task enters a terminated state, which include
+// finished, canceled
+func IsTerminatedError(errIn error) bool {
+	return derrors.ErrWorkerCancel.Equal(errIn) ||
+		derrors.ErrWorkerFinish.Equal(errIn) ||
+		derrors.ErrWorkerFailed.Equal(errIn)
 }
--- a/engine/framework/internal/master/worker_manager.go
+++ b/engine/framework/internal/master/worker_manager.go
@@ -508,7 +508,9 @@ func (m *WorkerManager) checkWorkerEntriesOnce() error {
 			case frameModel.WorkerStateFinished:
 				offlineError = derror.ErrWorkerFinish.FastGenByArgs()
 			case frameModel.WorkerStateStopped:
-				offlineError = derror.ErrWorkerStop.FastGenByArgs()
+				offlineError = derror.ErrWorkerCancel.FastGenByArgs()
+			case frameModel.WorkerStateError:
+				offlineError = derror.ErrWorkerFailed.FastGenByArgs()
 			default:
 				offlineError = derror.ErrWorkerOffline.FastGenByArgs(workerID)
 			}
--- a/engine/framework/master.go
+++ b/engine/framework/master.go
@@ -59,6 +59,7 @@ type Master interface {
 	Poll(ctx context.Context) error
 	MasterID() frameModel.MasterID
 	Close(ctx context.Context) error
+	Stop(ctx context.Context) error
 	NotifyExit(ctx context.Context, errIn error) error
 }
 
@@ -93,6 +94,9 @@ type MasterImpl interface {
 
 	// CloseImpl is called when the master is being closed
 	CloseImpl(ctx context.Context) error
+
+	// StopImpl is called when the master is being canceled
+	StopImpl(ctx context.Context) error
 }
 
 const (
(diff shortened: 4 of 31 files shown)


engine/servermaster/jobop/operator.go
```diff
@@ -0,0 +1,102 @@
...
+func (oper *JobOperatorImpl) MarkJobCanceled(ctx context.Context, jobID string) error {
+	return oper.setJobOperate(ctx, jobID, ormModel.JobOpStatusCanceled)
+}
+
+// Tick implements JobOperator.Tick
+func (oper *JobOperatorImpl) Tick(ctx context.Context) error {
+	if !oper.isRunning.CAS(false, true) {
+		return nil
+	}
+	defer oper.isRunning.Swap(true)
```
liuzix, line 89: Why is this necessary?

engine/servermaster/jobop/operator.go
```diff
@@ -0,0 +1,102 @@
...
+// NewJobOperatorImpl creates a new JobOperatorImpl
+func NewJobOperatorImpl(cli pkgOrm.Client, router operateRouter) *JobOperatorImpl {
+	return &JobOperatorImpl{
+		frameMetaClient: cli,
+		router:          router,
+		isRunning:       *atomic.NewBool(false),
+	}
+}
+
+func (oper *JobOperatorImpl) setJobOperate(
```
liuzix, line 54: Is `updateJobOperationStatus` a better name?

engine/servermaster/jobmanager.go
```diff
@@ -459,6 +475,11 @@ func (jm *JobManagerImpl) Tick(ctx context.Context) error {
 		return err
 	}
 
+	// FIXME: Is this too frequent?
+	if err := jm.jobOperator.Tick(ctx); err != nil {
```
liuzix, line 479: Is it feasible to let `jobOperator` maintain an internal goroutine?

engine/pkg/orm/client.go
```diff
@@ -641,6 +651,73 @@ func (c *metaOpsClient) GetOneResourceForGC(ctx context.Context) (*resModel.Reso
...
+			return errors.WrapError(errors.ErrMetaOpFail, err)
+		}
+		result.rowsAffected = exec.RowsAffected
+		return nil
+	})
+	return result, errors.WrapError(errors.ErrMetaOpFail, err)
+}
+
+// SetJobCanceled sets a cancelled status if a cancelling op exists.
+// If job op status is not found, does nothing.
```
liuzix, line 690: Can you comment on the behavior `If job op status is not found, does nothing`?

engine/servermaster/jobop/operator.go
amyangfei: removed this logic. I was expect `Tick` can be called concurrently, but current it doesn't need

engine/framework/base_jobmaster.go
```diff
@@ -367,8 +377,24 @@ func (j *jobMasterImplAsWorkerImpl) Workload() model.RescUnit {
...
-	return j.inner.OnJobManagerMessage(topic, message)
+func (j *jobMasterImplAsWorkerImpl) OnMasterMessage(
+	ctx context.Context, topic p2p.Topic, message interface{},
+) error {
+	switch msg := message.(type) {
+	case *frameModel.StatusChangeRequest:
+		switch msg.ExpectState {
+		case frameModel.WorkerStatusStopped:
+			if !j.onCancelCalled {
+				j.onCancelCalled = true
```
maxshuang, line 388: We can add a `Info` log here to observe the framework `Cancel` behavior.

engine/framework/internal/master/worker_manager.go
```diff
@@ -508,7 +508,7 @@ func (m *WorkerManager) checkWorkerEntriesOnce() error {
 			case frameModel.WorkerStatusFinished:
 				offlineError = derror.ErrWorkerFinish.FastGenByArgs()
 			case frameModel.WorkerStatusStopped:
```
maxshuang, line 510: will we unify the `status` in other pr? Like:
`WorkerStatusStopped` -> `WorkerStatusCanceled`

engine/framework/master.go
```diff
@@ -93,6 +94,9 @@ type MasterImpl interface {
 
 	// CloseImpl is called when the master is being closed
 	CloseImpl(ctx context.Context) error
+
+	// StopImpl is called when the master is being canceled
```
maxshuang, line 98: I think the comment for `CloseImpl` and `StopImpl` is inaccurate.
* `CloseImpl` is called when the job is not the terminated state;
* `StopImpl` is called when the job is  the terminated state;
terminated state is finished/failed/canceled.

amyangfei: I think `StopImpl` is called when job is failed or canceled, `finished` job should use `CloseImpl` logic.

engine/pkg/orm/model/jobop.go
```diff
@@ -0,0 +1,53 @@
...
+	"op",
+	"job_id",
+}
+
+// JobOp stores job operation recoreds
+type JobOp struct {
+	Model
+	Op    JobOpStatus `gorm:"type:tinyint not null;index:idx_job_op,priority:2;comment:Canceling(1),Canceled(2)"`
+	JobID string      `gorm:"type:varchar(128) not null;uniqueIndex:uk_job_id"`
+}
```
maxshuang, line 46: What about adding `Op` to job meta table directly?
What the motivation to use an isolated table to store the `canceling` action?

amyangfei: Do you mean table `master_meta_kv_data`?
To use an independent table to avoid data race, since both job manager and job master could update rows in table `master_meta_kv_data`

amyangfei: As we have discussed before, cancel job is a high level concept, which is not needed to store in master meta data.

maxshuang: Reasonable.

engine/framework/internal/master/worker_manager.go
amyangfei: yes

engine/servermaster/jobmanager.go
```diff
@@ -613,3 +646,33 @@ func (jm *JobManagerImpl) WatchJobStatuses(
...
+		timer := time.NewTimer(jobOperateInterval)
+		for {
+			select {
+			case <-ctx.Done():
+				return errors.Trace(ctx.Err())
+			case <-jm.jobOperatorNotifier:
+			case <-timer.C:
+			}
+			if err := jm.jobOperator.Tick(ctx); err != nil {
+				inErr := jm.Exit(ctx, framework.ExitReasonFailed, err, "")
```
maxshuang, line 661: Why we expect the whole jobmanager exit when job operator is fail?
What about keep log error and try?

maxshuang:
I will review it carefully later.

engine/framework/internal/eventloop/runner.go
```diff
@@ -62,6 +62,14 @@ func (r *Runner[R]) Run(ctx context.Context) error {
 		panic(fmt.Sprintf("unexpected exiting with nil error: %s", r.task.ID()))
 	}
 
+	if isCanceledError(err) {
+		if stopErr := r.task.Stop(ctx); stopErr != nil {
+			log.Warn("Cancel task returned error",
+				zap.String("label", r.task.ID()), zap.Error(stopErr))
+		}
+		return err
```
maxshuang, line 70: Why doesn't `CanceledError`  trigger `GracefulExit`?

engine/servermaster/jobmanager.go
```diff
@@ -84,41 +90,57 @@ type JobManagerImpl struct {
...
-	if job == nil {
-		// Check if the job is not found.
-		if _, err := jm.frameMetaClient.GetJobByID(ctx, req.Id); pkgOrm.IsNotFoundError(err) {
+	meta, err := jm.frameMetaClient.GetJobByID(ctx, req.Id)
+	if err != nil {
+		if pkgOrm.IsNotFoundError(err) {
 			return nil, ErrJobNotFound.GenWithStack(&JobNotFoundError{JobID: req.Id})
 		}
-		return nil, ErrJobNotRunning.GenWithStack(&JobNotRunningError{JobID: req.Id})
+		return nil, err
```
maxshuang, line 103: Seems that this error should be wrap as the grpc error by status.Error()

engine/servermaster/jobmanager.go
```diff
@@ -84,41 +90,57 @@ type JobManagerImpl struct {
...
+	pbJob, err := buildPBJob(meta)
+	if err != nil {
+		return nil, err
+	}
+	if pbJob.Status == pb.Job_Finished || pbJob.Status == pb.Job_Canceled {
+		return pbJob, nil
 	}
+
+	if err := jm.jobOperator.MarkJobCanceling(ctx, req.Id); err != nil {
+		return nil, err
```
maxshuang, line 115: ditto

engine/framework/master.go
maxshuang: Have we agreed that `CloseImpl` is called `when the job is not the terminated state`?

engine/servermaster/jobmanager.go
```diff
@@ -579,7 +607,12 @@ func (jm *JobManagerImpl) OnWorkerStatusUpdated(worker framework.WorkerHandle, n
 // CloseImpl implements frame.MasterImpl.CloseImpl
 func (jm *JobManagerImpl) CloseImpl(ctx context.Context) error {
 	jm.notifier.Close()
-	return nil
+	return jm.wg.Wait()
+}
+
+// StopImpl implements frame.MasterImpl.StopImpl
```
maxshuang, line 613: `bgJobOperatorLoop` may call `Exit` and trigger `StopImpl` for the  jobmananger.
Will it be a problem?

engine/servermaster/jobmanager.go
amyangfei: If we don't wrap gRPC status, gRPC will automatically wrap it. Maybe we should add unified wrapper as a middleware instead of wrapping it everywhere. This PR won't do it.

```
plain error
apple@dataflow3:~/code/scripts » ./cancel_job.sh c3d79fcf-6e60-4b4d-9261-fcbb806dca62
{"code":2, "message":"test cancel job error"}%

normalized error
apple@dataflow3:~/code/scripts » ./cancel_job.sh c3d79fcf-6e60-4b4d-9261-fcbb806dca62
{"code":2, "message":"JobNotFoundError: {\"JobID\":\"c3d79fcf-6e60-4b4d-9261-fcbb806dca62\"}"}

rpc wrapped error
apple@dataflow3:~/code/scripts » ./cancel_job.sh c3d79fcf-6e60-4b4d-9261-fcbb806dca62
{"code":10, "message":"JobNotFoundError: {\"JobID\":\"c3d79fcf-6e60-4b4d-9261-fcbb806dca62\"}"}
```

engine/framework/internal/eventloop/runner.go
amyangfei: Because cancel is a force stop scenario, in which job master doesn't need to wait workers exit.

engine/servermaster/jobmanager.go
amyangfei: Yes, changed the implement of `StopImpl` in job manager.

engine/servermaster/jobmanager.go
amyangfei: addressed

engine/pkg/orm/client.go
```diff
@@ -33,6 +33,8 @@ import (
 var globalModels = []interface{}{
 	&model.ProjectInfo{},
 	&model.ProjectOperation{},
+	&model.LogicEpoch{},
```
maxshuang, line 36:
```suggestion
```

engine/servermaster/jobop/operator.go
```diff
@@ -0,0 +1,122 @@
...
+	meta, err := oper.frameMetaClient.GetJobByID(ctx, jobID)
+	if err != nil {
+		if pkgOrm.IsNotFoundError(err) {
+			log.Warn("found orphan job operation", zap.String("job-id", jobID))
+			return oper.MarkJobNoop(ctx, jobID)
+		}
+		return err
+	}
+	switch meta.StatusCode {
+	case frameworkModel.MasterStatusFinished, frameworkModel.MasterStatusStopped:
```
maxshuang, line 118: If a job is marked `canceling` and fail at last, I think we can also mark the op `Canceled` since `failed` is a terminated state.

engine/framework/internal/eventloop/runner.go
```diff
@@ -115,6 +122,20 @@ func isForcefulExitError(errIn error) bool {
 		return true
 	}
 
+	if derrors.ErrWorkerCancel.Equal(errIn) {
+		// In canceling scenario business logic promises that all workers have
+		// been canceled, which means we don't need a graceful exit.
+		return true
```
maxshuang, line 128: I think it' s only for the `jobmaster`.
For `Worker`, things become subtle here. 
DM may hope to maintain the worker's status in `OnWorkerOffline`. If we skip `GracefulExit` here, 
DM job master need to rely on the `OnWorkerStatusUpdate` to update its worker's status.

maxshuang:
RestLGTM.

engine/servermaster/jobop/operator.go
amyangfei: Add a comment, will be added in another PR.

engine/servermaster/jobop/operator.go
```diff
@@ -0,0 +1,122 @@
...
+	ops, err := oper.frameMetaClient.QueryJobOpsByStatus(ctx, ormModel.JobOpStatusCanceling)
+	if err != nil {
+		return err
+	}
+	for _, op := range ops {
+		err := oper.router.SendCancelJobMessage(ctx, op.JobID)
+		if err != nil {
+			log.Warn("send cancel message to job master failed",
+				zap.String("job-id", op.JobID), zap.Error(err))
+			if err := oper.checkJobStatus(ctx, op.JobID); err != nil {
```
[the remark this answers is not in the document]
amyangfei: Not a good way to keep trying, it would cause dead loop.

engine/servermaster/jobmanager.go
```diff
@@ -613,3 +649,31 @@ func (jm *JobManagerImpl) WatchJobStatuses(
...
+				return errors.Trace(ctx.Err())
+			case <-jm.jobOperatorNotifier:
+			case <-timer.C:
+			}
+			if err := jm.jobOperator.Tick(ctx); err != nil {
+				// error returns from Tick is only caused by metastore error, so
+				// only log it and retry later.
+				log.Warn("job operator tick with error", zap.Error(err))
+			}
+			timer.Reset(jobOperateInterval)
```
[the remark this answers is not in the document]
amyangfei: It is acceptable, canceling has no fast-response requirement

engine/framework/internal/eventloop/runner.go
amyangfei: Move the logic to NotifyExit function of job master

engine/servermaster/jobmanager.go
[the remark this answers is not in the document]
maxshuang: `15s` seems too long for user waiting a job being canceled.
What's the `force kill` timeout for cancel in your design?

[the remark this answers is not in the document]
amyangfei: quite long, such as 10minutes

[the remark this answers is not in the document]
amyangfei: Canceling is not a high frequent behavior.

engine/servermaster/jobop/operator.go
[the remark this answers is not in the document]
maxshuang: Sorry for the expression.  
`Keep trying`  here is `try all the ops as possible`, not `try the error op infinitely.`

engine/framework/internal/eventloop/runner.go
maxshuang: Ok. But in my mind, `Cancel` is just a normal operation like job finish.
`jobmaster` exiting directly has pros and cons:
* Pros: `jobmaster` can call `StopImpl` quickly
* Cons: `jobmanager` need to wait a timeout to update its jobmaster status

maxshuang (approved):

engine/servermaster/jobop/operator.go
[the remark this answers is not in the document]
amyangfei: addressed

engine/servermaster/jobmanager.go
```diff
@@ -613,3 +649,31 @@ func (jm *JobManagerImpl) WatchJobStatuses(
...
+
+func (jm *JobManagerImpl) bgJobOperatorLoop(ctx context.Context) {
+	jm.wg.Go(func() error {
+		timer := time.NewTimer(jobOperateInterval)
+		for {
+			select {
+			case <-ctx.Done():
+				return errors.Trace(ctx.Err())
+			case <-jm.jobOperatorNotifier:
+			case <-timer.C:
```
liuzix, line 661: Why do we need both `jm.jobOperatorNotifier` and a timer?

amyangfei: Changed to use `notify.Notifier`, which contains both manually trigger and periodic trigger

engine/pkg/orm/client.go
```diff
@@ -641,6 +652,98 @@ func (c *metaOpsClient) GetOneResourceForGC(ctx context.Context) (*resModel.Reso
 	return &ret, nil
 }
 
+// SetJobNoop sets a job noop status if a job op record exists and status is canceling
```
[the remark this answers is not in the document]
amyangfei: added

amyangfei:
ptal @liuzix

liuzix (approved):

PR #6886, opened 2022-08-25
server(ticdc): do not drain capture by signal handling

### LINKED ISSUE #6829: cdc graceful shutdown trigger drain capture make user experience depressing when try to stop a instance immediately.
It's too time cost, around 100s before the instance can be shut down, which makes `TiUP stop` quite slow, making users feel time flies so slow.

3AceShowHand:

Issue Number: close #6829 

shutdown the instance by signal, do not drain the capture, to make the instance restart as soon as possible.

* if the instance is the owner, resign the ownership by call `AsyncStop`
* set the capture status to stopping

#### Tests 

* Deploy at least 2 cdc instances, replicating some tables on each instances.
* reload all cdc instance by `tiup cluster reload`, set `--transfer-timeout` to `2`.

all cdc instances restart as soon as possible.

### Release note 

```release-note
`None`.
```


cdc/capture/capture.go
```diff
@@ -618,102 +620,18 @@ func (c *captureImpl) AsyncClose() {
...
-			case <-ctx.Done():
-				// Give up when the context cancels. In the current
-				// implementation, it is caused TiCDC receives a second signal
-				// and begins force shutdown.
-				return
-			case <-ticker.C:
-			}
-		}
-	}()
+	close(done)
```
overvenus, line 631: `done` is useless.

cdc/capture/capture.go
```diff
@@ -618,102 +620,18 @@ func (c *captureImpl) AsyncClose() {
...
 	return done
 }
 
-func (c *captureImpl) drainImpl(ctx context.Context) bool {
-	if !c.config.Debug.EnableSchedulerV3 {
-		// Skip drain as two phase scheduler is disabled.
-		return true
-	}
-
-	// Step 1, resign ownership.
```
overvenus, line 656: I think we should keep resign owner.

pkg/cmd/server/server.go
```diff
@@ -150,7 +150,12 @@ func (o *options) run(cmd *cobra.Command) error {
 		return errors.Trace(err)
 	}
 	// Drain the server before shutdown.
-	shutdownNotify := func() <-chan struct{} { return server.Drain(ctx) }
+	shutdownNotify := func() <-chan struct{} {
```
overvenus, line 153: shutdownNotify returns a channel, because it should be asynchronous. `InitSignalHandling` can abort drain when it receives a second signal.

cdc/capture/capture.go
```diff
@@ -580,100 +582,17 @@ func (c *captureImpl) AsyncClose() {
...
-				return false
-			}
-		}
-		if len(query.Data.([]*model.CaptureInfo)) <= 1 {
-			// There is only one capture, the owner itself. It's impossible to
-			// resign owner nor move out tables, give up drain.
-			log.Warn("there is only one capture, skip drain")
-			return true
-		}
+func (c *captureImpl) Drain() {
```
overvenus, line 585: How about adding a for loop to make sure it resigns ownership?

3AceShowHand: no need to wait for the ownership resigend, most of the time it's very fast.

When the user stop the cdc instance by sent signal, the normal `server.Close()` is guaranteed that the ownership resigned before exit.

overvenus (approved):

asddongmen (approved):

Issue #6867, opened 2022-08-24
Implement service discovery on servermaster


gozssky:
## Background
To simplify the architecture of our system, we plan to remove the dependency on ETCD. Currently, we depend on the ETCD to watch the changes of the master or executor nodes. To remove the dependency, we need to implement discovery on our server master. Server master provides list APIs to get the current master and executor nodes. Executors can watch the changes in the list result and update the cluster topology.

See more details in https://github.com/pingcap/tiflow/issues/6387.

## Development Task

- [x] Add list masters/executors API https://github.com/pingcap/tiflow/pull/6864
- [x] Implement service discovery based on the list API https://github.com/pingcap/tiflow/pull/6891
- [ ] Persistent executors information to framework metastore

PR #6891, opened 2022-08-25
discovery(engine): refactor agent to use list api

gozssky:

Issue Number: ref https://github.com/pingcap/tiflow/issues/6867

Refactor `discovery.Agent` to use `ListMasters` and `ListExecutors` API.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```

### DIFF
--- a/engine/executor/server.go
+++ b/engine/executor/server.go
@@ -15,7 +15,6 @@ package executor
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"net/http/pprof"
 	"strings"
@@ -42,6 +41,7 @@ import (
 	pkgOrm "github.com/pingcap/tiflow/engine/pkg/orm"
 	"github.com/pingcap/tiflow/engine/pkg/p2p"
 	"github.com/pingcap/tiflow/engine/pkg/promutil"
+	"github.com/pingcap/tiflow/engine/pkg/serverutil"
 	"github.com/pingcap/tiflow/engine/pkg/tenant"
 	"github.com/pingcap/tiflow/engine/test"
 	"github.com/pingcap/tiflow/engine/test/mock"
@@ -58,6 +58,16 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	// TODO since we introduced queuing in the TaskRunner, it is no longer
+	// easy to implement the capacity. Think of a better solution later.
+	// defaultRuntimeCapacity      = 65536
+	defaultRuntimeIncomingQueueLen   = 256
+	defaultRuntimeInitConcurrency    = 256
+	defaultTaskPreDispatchRequestTTL = 10 * time.Second
+	defaultDiscoveryAutoSyncInterval = 5 * time.Second
+)
+
 // Server is an executor server.
 type Server struct {
 	cfg     *Config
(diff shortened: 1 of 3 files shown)


engine/executor/server.go
```diff
@@ -751,13 +721,18 @@ func (s *Server) bgUpdateServerMasterClients(ctx context.Context) error {
 		select {
 		case <-ctx.Done():
 			return perrors.Trace(ctx.Err())
-		case info, ok := <-s.cliUpdateCh:
-			if !ok {
-				// s.cliUpdateCh is closed.
-				return nil
+		case <-time.After(defaultDiscoveryAutoSyncInterval):
```
liuzix, line 724: Please consider using a `Ticker` unless it's absolutely necessary to use `time.After`. The reason is that `time.After` can leak timers.

engine/internal/pkg/discovery/agent.go
```diff
@@ -38,182 +35,166 @@ const (
...
-				zap.Duration("duration", duration))
+func (a *agentImpl) autoSync(ctx context.Context) error {
+	snap := make(Snapshot)
+	for {
+		newSnap, err := a.getSnapshot(ctx)
+		if err != nil {
+			if errors.Cause(err) == context.Canceled || errors.Cause(err) == context.DeadlineExceeded {
+				return errors.Trace(err)
+			}
+			log.Warn("failed to get snapshot", zap.Error(err))
```
liuzix, line 144: When can this happen?

engine/internal/pkg/discovery/agent.go
```diff
@@ -38,182 +35,166 @@ const (
...
+	defer eventNotifier.Close()
 
-	discoveryCtx, cancel := context.WithCancel(ctx)
-	defer cancel()
+	syncErr := make(chan error, 1)
+	syncDone := make(chan struct{})
+	go func() {
+		defer close(syncDone)
+		syncErr <- a.autoSync(ctx)
+	}()
```
liuzix, line 108: Do you consider using errgroup?

engine/executor/server.go
gozssky: `Ticker` will ignore the request latency. Is it possible to leak the timer every time it is created, or just when clicking "<-ctx.Done()".

engine/internal/pkg/discovery/agent.go
gozssky: When encountering network error or servermaster failover.

engine/internal/pkg/discovery/agent.go
gozssky: Changed to use errgroup.

engine/internal/pkg/discovery/agent.go
```diff
@@ -38,182 +36,160 @@ const (
...
 
+// NodeType describes the type of the node.
+type NodeType string
+
+const (
+	// NodeTypeMaster indicates that the node is a master.
+	NodeTypeMaster = NodeType("master")
+	// NodeTypeExecutor indicates that the node is an executor.
+	NodeTypeExecutor = NodeType("executor")
+)
```
amyangfei, line 47: Why don't reuse `NodeType` defined in engine/model/cluster.go

gozssky: `NodeInfo` and `NodeType` defined in `engine/model/cluster.go` will be removed after we store executor info in MySQL.

amyangfei (approved):

liuzix (approved):

PR #6897, opened 2022-08-26
model(engine): support ext field in MasterMetaKVData

liuzix:

Issue Number: ref #6270

- Added an `ext` field in `MasterMetaKVData` for better extensibility.

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
None.
```

### DIFF
--- a/engine/framework/model/master.go
+++ b/engine/framework/model/master.go
@@ -14,6 +14,7 @@
 package model
 
 import (
+	"database/sql/driver"
 	"encoding/json"
 
 	"gorm.io/gorm"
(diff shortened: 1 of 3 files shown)


amyangfei (approved):

engine/framework/model/master.go
```diff
@@ -40,6 +43,56 @@ const (
 	// extend the status code here
 )
 
+// MasterMetaExt stores some attributes of job masters that do not need
```
maxshuang, line 46: Does sqlite mock client support this json?

maxshuang:
Rest LGTM

engine/framework/model/master.go
liuzix: It seems working for now. We may encounter problem if we were to use advanced JSON functions, but they do not seem needed for now.

maxshuang (approved):

PR #6910, opened 2022-08-29
engine(dm): remove job id prefix in metadata

GMHDBJD:

Issue Number: ref #5504 

- remove job id prefix in metadata

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/engine/jobmaster/dm/metadata/cluster_info_test.go
+++ b/engine/jobmaster/dm/metadata/cluster_info_test.go
@@ -29,12 +29,12 @@ func TestInfoStore(t *testing.T) {
 	clusterInfo := NewClusterInfo(ver1)
 	require.True(t, clusterInfo.Version.Equal(ver1))
 
-	clusterInfoStore := NewClusterInfoStore("cluster_info_test", mock.NewMetaMock())
+	clusterInfoStore := NewClusterInfoStore(mock.NewMetaMock())
 	key := clusterInfoStore.Key()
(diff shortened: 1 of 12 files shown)


engine/jobmaster/dm/metadata/cluster_info_test.go
```diff
@@ -29,12 +29,12 @@ func TestInfoStore(t *testing.T) {
 	clusterInfo := NewClusterInfo(ver1)
 	require.True(t, clusterInfo.Version.Equal(ver1))
 
-	clusterInfoStore := NewClusterInfoStore("cluster_info_test", mock.NewMetaMock())
+	clusterInfoStore := NewClusterInfoStore(mock.NewMetaMock())
 	key := clusterInfoStore.Key()
 	keys, err := adapter.DMInfoKeyAdapter.Decode(key)
 	require.NoError(t, err)
 	require.Len(t, keys, 1)
```
lance6716, line 36: should we return zero length?

lance6716 (commented):
rest lgtm

engine/jobmaster/dm/metadata/cluster_info_test.go
GMHDBJD: return [""] now, same as dm keyadater.

lance6716 (approved):

maxshuang (approved):

Issue #6776, opened 2022-08-17
Tracking issue for  migrating some DM tests to DFE


CharlesCheung96:
## P0
- [x] full_mode https://github.com/pingcap/tiflow/pull/6883
- [x] gbk https://github.com/pingcap/tiflow/pull/7186
- [x] new_collation_off https://github.com/pingcap/tiflow/pull/6857
- [x] print_status https://github.com/pingcap/tiflow/pull/7129
- [x] sql_mode https://github.com/pingcap/tiflow/pull/6951
- [x] sync_collation https://github.com/pingcap/tiflow/pull/6808
- [x] tls https://github.com/pingcap/tiflow/pull/6913

## P1

- [x] case_sensitive https://github.com/pingcap/tiflow/pull/6768
- [ ] check_task
- [ ] downstream_more_column
- [ ] gtid
- [ ] handle_error
- [ ] handle_error_2
- [ ] handle_error_3
- [ ] online_ddl
- [ ] s3_dumpling_lighting
- [ ] sequence_sharding
- [ ] sequence_sharding_removemeta
- [ ] shardddl1
- [ ] shardddl1_1
- [ ] shardddl2
- [ ] shardddl2_1
- [ ] shardddl3
- [ ] shardddl3_1
- [ ] shardddl4
- [ ] shardddl4_1
- [ ] shardddl_optimistic
- [ ] start_task
- [ ] tracker_ignored_ddl
- [ ] upstream_switch

## P2

- [ ] all_mode
  - [ ] test_expression_filter
  - [ ] test_regexpr_router
  - [ ] others（failpoint）
- [ ] binlog_999999
- [ ] downstream_diff_index
- [ ] drop_column_with_index
- [ ] expression_filter
- [ ] extend_column
- [ ] many_tables https://github.com/pingcap/tiflow/pull/6616
- [ ] validator_basic
- [ ] validator_checkpoint

## failpoint
- [ ] adjust_gtid（failpoint）
- [ ] async_checkpoint_flush（failpoint）
- [ ] checkpoint_transaction（failpoint）
- [ ] duplicate_event（failpoint）
- [ ] fake_rotate_event（failpoint）
- [ ] incremental_mode(failpoint)
- [ ] load_interrupt(failpoint)
- [ ] metrics(failpoint)
- [ ] safe_mode(failpoint)
- [ ] sequence_safe_mode(failpoint)
- [ ] sequence_sharding_optimistic(failpoint)
[...]

PR #6913, opened 2022-08-29
test(engine): add DM TLS test

### LINKED ISSUE #7006: Error 1406: Data too long for column 'meta_value' at row 1
### What did you do?

checkout https://github.com/pingcap/tiflow/pull/6913 and run `dm_tls` test case

other error than Meta error

### What did you see instead?

```
[2022/09/08 02:49:37.024 +00:00] [ERROR] [callbacks.go:134] ["trace log"] [component=gorm] [elapsed=1.347393ms] [sql="INSERT INTO `dm_tls_meta_kvs` (`created_at`,`updated_at`,`meta_key`,`meta_value`,`job_id`) VALUES ('2022-09-08 02:49:37.024','2022-09-08 02:49:37.024','/data-flow/dm/job','<binary>','58a9a1b9-732c-472a-a28c-b919d27d9a5c') ON DUPLICATE KEY UPDATE `updated_at`='2022-09-08 02:49:37.024',`meta_key`=VALUES(`meta_key`),`meta_value`=VALUES(`meta_value`),`job_id`=VALUES(`job_id`)"] [affected-rows=0] [error="Error 1406: Data too long for column 'meta_value' at row 1"]
```

[server-executor-2.log](https://github.com/pingcap/tiflow/files/9523214/server-executor-2.log)

### Versions of the cluster

```console
(paste Dataflow Engine version here)
```

lance6716:

Issue Number: close #7006 ref #6776

#### Tests 

 - Integration test

### Release note 

```release-note
 `None`.
```


lance6716:
blocked by https://github.com/pingcap/tiflow/issues/6922 https://github.com/pingcap/tiflow/issues/6921

maxshuang:
Will `job detail` exceed 65k? 
May be we also need to enlarge the `detail` and `ExtMsg` in job meta and worker meta.

maxshuang:
Rest LGTM.

lance6716:
> Will `job detail` exceed 65k? May be we also need to enlarge the `detail` and `ExtMsg` in job meta and worker meta.

This may happen when the jon detail contains a large error message, for example, a message contains many sub-errors and their stack.

Currently DM will not report errors like that. We have solved the long message for dmctl in the past.

amyangfei:
PTAL @gozssky

dm/config/security.go
```diff
@@ -28,8 +29,11 @@ type Security struct {
 	SSLKey        string   `toml:"ssl-key" json:"ssl-key" yaml:"ssl-key"`
 	CertAllowedCN strArray `toml:"cert-allowed-cn" json:"cert-allowed-cn" yaml:"cert-allowed-cn"`
 	SSLCABytes    []byte   `toml:"ssl-ca-bytes" json:"-" yaml:"ssl-ca-bytes"`
```
lonng, line 31: Do we expect users to specify the SSL-related configurations in bytes? If not, we should not expose this configuration items.

lance6716: Yes, currently user can send the bytes by OpenAPI, and after DM is integrated to engine certificate bytes can be loaded from k8s secret.

dm/pkg/conn/basedb.go
```diff
@@ -69,16 +69,14 @@ func (d *DefaultDBProviderImpl) Apply(config *config.DBConfig) (*BaseDB, error)
 		if loadErr := config.Security.LoadTLSContent(); loadErr != nil {
 			return nil, terror.ErrCtlLoadTLSCfg.Delegate(loadErr)
 		}
-		tlsConfig, err := util.ToTLSConfigWithVerifyByRawbytes(config.Security.SSLCABytes,
-			config.Security.SSLCertBytes, config.Security.SSLKEYBytes, config.Security.CertAllowedCN)
+		tlsConfig, err := util.NewTLSConfig(
```
amyangfei, line 72: I find `InsecureSkipVerify` is always true from API `util.NewTLSConfig`, is this as expected?

deployments/engine/docker-compose/dm_databases_with_tls.yaml
```diff
@@ -0,0 +1,68 @@
...
+    dm_upstream_mysql:
+        image: mysql:5.7
+        container_name: dm_upstream_mysql
+        command:
+            - --require_secure_transport=ON
+        ports:
+            - "3306:3306"
+        volumes:
+            - ./config/mysql.cnf:/etc/mysql/conf.d/mysql.cnf
+            - /tmp/mysql1:/var/lib/mysql
```
CharlesCheung96, line 13: Is this folder only used to save certification configuration? If so, better to use a more explicit path, such as `/tmp/certs/mysql1`.

dm/pkg/conn/basedb.go
lance6716: https://github.com/pingcap/tidb/pull/37479#discussion_r962687063

we use `VerifyPeerCertificate` to do explicit check

deployments/engine/docker-compose/dm_databases_with_tls.yaml
lance6716: under this folder MySQL also writes other files like binlog file, storage engine file (but we don't use them)

gozssky: Do we need to access this directory via the host machine? If not, I think we can avoid binding host path, because it may bring side effects when we run multiple tests.

gozssky (approved):
Rest LGTM.

deployments/engine/docker-compose/dm_databases_with_tls.yaml
lance6716: need to let DM in engine access the certificate

lance6716: maybe I should try `docker cp`

deployments/engine/docker-compose/dm_databases_with_tls.yaml
```diff
@@ -0,0 +1,65 @@
...
+        restart: on-failure
+
+    dm_downstream_tidb:
+        image: pingcap/tidb:latest
+        ports:
+            - "4000:4000"
+            - "10080:10080"
+        volumes:
+            - ./config/tidb_tls.toml:/tidb.toml
+            - /tmp/certs/downstream:/etc/certs
```
CharlesCheung96, line 57: Better to use `/tmp/tiflow_engine_test` ?

lance6716: it will be adjusted to wrong path

lance6716: The adjust logic will cause many unexpected behaviour, and it's really hard to locate. I don't like it.

engine/test/integration_tests/dm_tls/run.sh
```diff
@@ -0,0 +1,71 @@
...
+	sed -i "s,<mysql2-cert>,$(base64 -w0 $WORK_DIR/mysql2/client-cert.pem)," $WORK_DIR/job.yaml
+
+	create_job_json=$(base64 -w0 $WORK_DIR/job.yaml | jq -Rs '{ type: "DM", config: . }')
+	echo "create_job_json: $create_job_json"
+	job_id=$(curl -X POST -H "Content-Type: application/json" -d "$create_job_json" "http://127.0.0.1:10245/api/v1/jobs?tenant_id=dm_tls&project_id=dm_tls" | jq -r .id)
+	echo "job_id: $job_id"
+
+	# wait for dump and load finished
+
+	exec_with_retry --count 30 "curl \"http://127.0.0.1:10245/api/v1/jobs/$job_id/status\" | tee /dev/stderr | jq -e '.task_status.\"mysql-02\".status.unit == 12'"
```
CharlesCheung96, line 56:
```suggestion
	exec_with_retry --count 30 "curl \"http://127.0.0.1:10245/api/v1/jobs/$job_id/status\" | tee /dev/stderr | jq -e '.task_status.\"mysql-02\".status.unit == \"DMSyncTask\"'"
```

CharlesCheung96: ref: https://github.com/pingcap/tiflow/pull/6992/files#diff-1bb6fc9a08ca33732e56435267aa683fa3ff2bbf79a9ee353b3828514858e2c6R45

deployments/engine/docker-compose/dm_databases_with_tls.yaml
CharlesCheung96: From my point of view, the adjust logic was a compromise that allowed multiple integration tests to share a single deployment file while still handling log files appropriately. And using '/tmp/tiflow_engine_test' ensures that all configuration files for an integration test are clustered in the same directory, which can greatly reduce the burden of finding test-related files. 
So even though 'adjust_config' has some problems, we should first try to fix it rather than scrap it. Besides, can you list some unexpected behaviors?

lance6716: I wrote `/tmp/tiflow_engine_test/certs/downstream` here, and also align this path in for `generate_cert` and mysql client in`run.sh`. And as the result, mysql client to TiDB report "server does not support TLS". Very confusing.

Please fix it in another PR, this PR has been opened too long.

CharlesCheung96 (approved):

PR #6918, opened 2022-08-29
discovery(engine): store executor information to database

gozssky:

Issue Number: close https://github.com/pingcap/tiflow/issues/6867

Store executor information to the database. And deprecate `NodeInfo`. After this modification, executor discovery doesn't need to depend on ETCD anymore.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/executor/server.go
```diff
@@ -282,20 +281,14 @@ func (s *Server) Stop() {
 	}
 
 	if s.metastores.IsInitialized() {
-		etcdCli := s.metastores.ServiceDiscoveryStore()
-		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
-		defer cancel()
-		_, err := etcdCli.Delete(ctx, s.info.EtcdKey())
-		if err != nil {
-			log.L().Warn("failed to delete executor info", zap.Error(err))
-		}
```
liuzix, line 291: Do we need any operations equivalent to what you have deleted in order to remove the executor's information from the database as soon as possible?

engine/pkg/rpcerror/grpc.go
```diff
@@ -141,9 +141,16 @@ func UnaryServerInterceptor(ctx context.Context, req interface{}, info *grpc.Una
...
+			s = status.FromContextError(err)
+		}
+		switch s.Code() {
+		case codes.Unknown:
+			logger.Warn("request handled with an unknown error")
+		case codes.Internal:
 			logger.Warn("request handled with an internal error")
-		} else {
+		default:
 			logger.Debug("request handled with an error")
```
liuzix, line 154: Do we need to log more information on the request here?

engine/servermaster/executor_manager.go
```diff
@@ -39,16 +43,18 @@ func init() {
...
 	HasExecutor(executorID string) bool
-	ListExecutors() []*model.NodeInfo
+	ListExecutors() []*ormModel.Executor
 	GetAddr(executorID model.ExecutorID) (string, bool)
 	Start(ctx context.Context)
 	Stop()
 
+	// ResetExecutors reset all executors with the latest meta from database.
+	ResetExecutors(ctx context.Context) error
+
```
liuzix, line 57: Maybe this should be part of the initialization logic instead of part of the interface?

engine/servermaster/executor_manager.go
```diff
@@ -91,19 +99,26 @@ func NewExecutorManagerImpl(initHeartbeatTTL, keepAliveInterval time.Duration, c
...
-	addr := exec.Addr
+	addr := exec.Address
 	delete(e.executors, id)
 	e.rescMgr.Unregister(id)
 	log.Info("notify to offline exec")
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	if err := e.metaClient.DeleteExecutor(ctx, id); err != nil {
+		return perrors.Trace(err)
```
liuzix, line 119: What is the expectation here if an error is returned? Note that `e.notifier.Notify` will not be called, so it seems like the status of  the `ExecutorManagerImpl` has become inconsistent.

engine/servermaster/executor_manager.go
```diff
@@ -148,51 +163,56 @@ func (e *ExecutorManagerImpl) HandleHeartbeat(req *pb.HeartbeatRequest) (*pb.Hea
...
 // and then registers the executor.
-func (e *ExecutorManagerImpl) AllocateNewExec(req *pb.RegisterExecutorRequest) (*model.NodeInfo, error) {
-	executor := req.Executor
-	log.Info("allocate new executor", zap.Stringer("executor", executor))
+func (e *ExecutorManagerImpl) AllocateNewExec(ctx context.Context, req *pb.RegisterExecutorRequest) (*ormModel.Executor, error) {
+	pbExecutor := req.Executor
+	log.Info("allocate new executor", zap.Stringer("executor", pbExecutor))
 
 	e.mu.Lock()
+	defer e.mu.Unlock()
```
liuzix, line 193: Do you think it would be a good idea to use a cancellable lock here?

engine/executor/server.go
gozssky: Yes. I will add a unregister method on master later. (Maybe in the next PR)

engine/pkg/rpcerror/grpc.go
gozssky: `logger` already contains the method, error, and request info. Do we still need more information here?

engine/servermaster/executor_manager.go
gozssky: Good idea!

engine/servermaster/executor_manager.go
gozssky: Yes, it's possible to become inconsistent. Can we tolerate that an executor is removed from memory, but failed to delete from the database? If yes, I think I can move this operation to the last.

engine/servermaster/executor_manager.go
gozssky: Addressed in https://github.com/pingcap/tiflow/pull/6918/commits/e5d1fc51f98a045215e3edfd1a67a3efe0e93733.

engine/servermaster/executor_manager.go
gozssky: I move this operation to a goroutine. Since we use ttl mechanism to manage the executor's life cycle, I think we can tolerate that a tombstone executor may be left in the database.

Addressed in https://github.com/pingcap/tiflow/commit/e5d1fc51f98a045215e3edfd1a67a3efe0e93733.

engine/servermaster/executor_manager.go
gozssky: After a deep investigation, I found a cancellable lock would bring a lot of changes.

To reduce lock wait time, I move database operation out of lock held. If the database operation fails, the client shouldn't reuse the executor in memory map, which will be removed after ttl expires.

gozssky:
Reviewing notes:

Is it better rename `engine/servermaster/orm` to `engine/servermaster/executormeta`? (`orm` may be confused with `engine/pkg/orm`).

amyangfei:
> Reviewing notes:
> 
> Is it better to rename `engine/servermaster/orm` to `engine/servermaster/executormeta`? (`orm` may be confused with `engine/pkg/orm`).

+1 for `engine/servermaster/executormeta`

engine/servermaster/executor_manager.go
```diff
@@ -334,7 +371,35 @@ func (e *ExecutorManagerImpl) GetAddr(executorID model.ExecutorID) (string, bool
...
+func (e *ExecutorManagerImpl) resetExecutors(ctx context.Context) error {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	orphanExecutorIDs := make(map[model.ExecutorID]struct{})
+	for id := range e.executors {
+		orphanExecutorIDs[id] = struct{}{}
+	}
+
+	executors, err := e.metaClient.QueryExecutors(ctx)
```
amyangfei, line 386: If an executor is offline and remains metadata in metastore, it will be added into executor manager and removed after heartbeat timeout. During that time could tasks be dispatched to this nonexistent executor.

engine/servermaster/orm/model/executor.go
```diff
@@ -0,0 +1,102 @@
...
+			"Expected string or []byte, got %s", reflect.TypeOf(rawInput))
+	}
+
+	if err := json.Unmarshal(bytes, s); err != nil {
+		return errors.Annotate(err, "failed to unmarshal LabelSet")
+	}
+	return nil
+}
+
+// Executor records the information of an executor.
```
amyangfei, line 75: Will server matser information be recorded in metastore in the future?

gozssky: Yes, I will maintain the information in the leader election later.

engine/servermaster/executor_manager.go
gozssky: See https://github.com/pingcap/tiflow/pull/6918#discussion_r957970038.

liuzix (approved):

amyangfei (approved):

gozssky:
/hold

Merge after `engine/servermaster/orm` is renamed to engine/servermaster/executormeta`.

gozssky:
> /hold
> 
> Merge after `engine/servermaster/orm` is renamed to `engine/servermaster/executormeta`.

Done. See https://github.com/pingcap/tiflow/pull/6918/commits/5d8c700ae3ce39ad7a4bfe6d4c895e8913a656c3.
@amyangfei @liuzix  PTAL again

amyangfei (approved):

Issue #6915, opened 2022-08-29
When failed to create a job, engine's status is always "Created'


lance6716:
### What did you do?

```
# lance @ lance6716-nuc10i7fnh in /tmp/tiflow_engine_test [16:38:59] 
$ curl http://127.0.0.1:10245/api/v1/jobs/d2fb3ff4-a357-411d-ad45-d1063bcc13d3 

{"id":"d2fb3ff4-a357-411d-ad45-d1063bcc13d3", "type":"DM", "status":"Created", "config":"", "error":null}

                                                     
# lance @ lance6716-nuc10i7fnh in /tmp/tiflow_engine_test [16:49:28] 
$ curl http://127.0.0.1:10245/api/v1/jobs/d2fb3ff4-a357-411d-ad45-d1063bcc13d3/status

[...]

Issue #6806, opened 2022-08-19
When user submit a ill-formed DM job, should display the error


lance6716:
### What did you do?

job with wrong format

```
[...]

PR #6925, opened 2022-08-30
framework(engine): support fast fail when pre dispatch worker

amyangfei:

Issue Number: ref #6806, #6915, #6749

- Add one interface `IsRetryableError` in `WorkerFactory`, which will be implemented by business logic to check whether an error raised from business will fast fail job. In this way business logic doesn't need to modify any logic to achieve fast fail.
- Parse business error from plain message, and try to construct to an RFC error, which is used when we find a persisted error from metastore.
- The error sent from executor to server master is a `normalized error` and wrapped as an gRPC error. So gRPC client will use `FromGRPCError` to extract it into a `normalized error`. 

There are two fast fail scenarios, one is when a job is starting, the other one is job fails during running, this PR fixes the first scenario, the second scenario will follow the same mechanism, but requires to parse business error message from master meta. And the second scenario will be added in another PR.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


engine/framework/registry/errors.go
```diff
@@ -0,0 +1,35 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package registry
+
+import "github.com/pingcap/tiflow/engine/pkg/rpcerror"
+
+// CreateWorkerRetryableError provides details of ErrCreateWorkerRetryable, which
+// contains original error returned from CreateWorker
+type CreateWorkerRetryableError struct {
+	rpcerror.Error[rpcerror.NotRetryable, rpcerror.Aborted]
```
lance6716, line 20: this should be Retryable?

engine/framework/registry/factory.go
```diff
@@ -81,7 +87,15 @@ func (f *SimpleWorkerFactory[T, C]) DeserializeConfig(configBytes []byte) (Worke
...
 	if err := json.Unmarshal(configBytes, config); err != nil {
-		return nil, errors.Trace(err)
+		return nil, errors.Annotate(errDeserializeConfig, err.Error())
 	}
 	return config, nil
 }
+
+// IsRetryableError implements WorkerFactory.IsRetryableError
+func (f *SimpleWorkerFactory[T, C]) IsRetryableError(err error) bool {
+	if libErrors.Is(err, errDeserializeConfig) {
```
lance6716, line 97: can the error from all implementations of `DeserializeConfig` become non-retryable?

engine/executor/server.go
```diff
@@ -240,7 +241,7 @@ func (s *Server) PreDispatchTask(ctx context.Context, req *pb.PreDispatchTaskReq
 		// "Use Aborted if the client should retry at a higher-level".
 		// Failure to make task is usually a problem that the business logic
 		// should be notified of.
-		return nil, status.Error(codes.Aborted, err.Error())
+		return nil, registry.GlobalWorkerRegistry().AnalyzeError(err, workerType)
```
lance6716, line 244: Can we extract the raw business error from framework wrapped error and pass it to business `IsRetryableError`, rather than use regexp to convert to business error from error messsage?

And seems `makeTask` will not raise errors of DM type, because DM business logic begins to be called after InitImpl

engine/framework/registry/errors.go
amyangfei: `rpcerror.NotRetryable` is used in gRPC client, not the business logic

engine/framework/registry/factory.go
amyangfei: It depends how `DeserializeConfig` is implemented, whether it contains retryable operation such as network access etc.

engine/executor/server.go
amyangfei: > And seems makeTask will not raise errors of DM type, because DM business logic begins to be called after InitImpl

Yes, `makeTask` mainly contains API from registry factory, including `DeserializeConfig` and `NewWorkerImpl`, which covers the error in create worker part.

Other DM business logic such as `Init` and `Tick` are called from the runtime level asynchronously, we need to find a way to communicate between executor and master to reach an agreement about the `fast fail` of a job. That's why we reuse the regulation of DM error or PingCAP rfc error to ship this information.

amyangfei: > Can we extract the raw business error from framework wrapped error and pass it to business IsRetryableError, rather than use regexp to convert to business error from error messsage?
Of course, can do it here.

engine/framework/registry/factory.go
```diff
@@ -81,7 +87,15 @@ func (f *SimpleWorkerFactory[T, C]) DeserializeConfig(configBytes []byte) (Worke
 	var config C
 	config = reflect.New(reflect.TypeOf(config).Elem()).Interface().(C)
 	if err := json.Unmarshal(configBytes, config); err != nil {
-		return nil, errors.Trace(err)
+		return nil, errors.Annotate(errDeserializeConfig, err.Error())
```
liuzix, line 90: This does not seem to be the idiomatic way to use `Annotate`, which is typically used to provide additional context to an error, rather than provide cause.

amyangfei: fixed, ptal

engine/executor/server.go
amyangfei: addressed, ptal

engine/executor/server.go
```diff
@@ -226,21 +226,23 @@ func (s *Server) PreDispatchTask(ctx context.Context, req *pb.PreDispatchTaskReq
...
 		req.GetMasterId(),
-		frameModel.WorkerType(req.GetTaskTypeId()),
+		workerType,
 		req.GetTaskConfig(),
 		req.GetWorkerEpoch(),
 	)
 	if err != nil {
-		// We use the code Aborted here per the suggestion in gRPC's documentation
+		// AnalyzeError will return a gRPC error with code Aborted, which
+		// followsthe suggestion in gRPC's documentation.
```
maxshuang, line 241:
```suggestion
// follows the suggestion in gRPC's documentation.
```

engine/framework/registry/errors.go
maxshuang, line 20: Why a `CreateWorkerRetryableError` has the `rpcerror.NotRetryable` atrribute?

maxshuang (approved):

maxshuang:
Rest LGTM.

dm/pkg/terror/terror.go
```diff
@@ -155,14 +159,16 @@ type Error struct {
...
-	return &Error{
+	err := &Error{
 		code:       code,
 		class:      class,
 		scope:      scope,
 		level:      level,
 		message:    message,
 		workaround: workaround,
 	}
+	CodeToErrorMap[code] = err
```
lonng, line 170: Do we need to take data race into account?

engine/framework/registry/registry.go
```diff
@@ -43,6 +44,9 @@ type Registry interface {
 		config []byte,
 		epoch frameModel.Epoch,
 	) (framework.Worker, error)
+	// AnalyzeError parses error, checks whether this error is retryable and
+	// returns rpc error based on it.
+	AnalyzeError(err error, tp framework.WorkerType) error
```
lance6716, line 49: I slightly prefer we don't add this function to Registry. `AnalyzeError` is related to a specific WorkerType, and inside `CreateWorker` we resolved the WorkerType to WorkerFactory, we can call WorkerFactory.IsRetryableError at that place.

lance6716 (commented):
14/18 files reviewed

engine/framework/registry/registry.go
amyangfei: addressed. I have separated the logic of `AnalyzeError` and leave `IsRetryableError` API in registry only.

lance6716 (approved):

Issue #6255, opened 2022-07-11
Tracking issue for adding integration tests to dataflow engine


CharlesCheung96:
- [x] Infrastructure development in https://github.com/PingCAP-QE/ci
- [x] Optimize directories and documents to reduce the difficulty of deploy DFE locally
- [x] Develop integration test framework and some utils
- [x] https://github.com/pingcap/tiflow/issues/6776
- [x] https://github.com/pingcap/tiflow/issues/6554
- [x] https://github.com/pingcap/tiflow/pull/6672

maxshuang:
Close this issue.
Leave the https://github.com/pingcap/tiflow/issues/6776 to track transfer progress.

PR #6934, opened 2022-08-30
engine(test): add docker in docker config

CharlesCheung96:

Issue Number: ref #6255 

1. Add `docker-cli in docker` config
2. Since jenkins pipeline would lost some logs, we save output of engine_integration_tests to file.

#### Tests 

 - Integration test

### Release note 

```release-note
`None`.
```


deployments/engine/docker/dind.Dockerfile
```diff
@@ -0,0 +1,13 @@
+# In this file, we build an image with `docker-cli in docker` for engine integration tests.
```
maxshuang, line 1: Which line indicates we are using `docker:dind`?

Makefile
```diff
@@ -500,7 +500,9 @@ engine_unit_test: check_failpoint_ctl
 	$(call run_engine_unit_test,$(ENGINE_PACKAGES))
 
 engine_integration_test: check_third_party_binary_for_engine
```
maxshuang, line 502: Does it mean we can no longer see the output log in the `action` page directly?
image

maxshuang:
RestLGTM.

Makefile
CharlesCheung96: This pr does not affect tests in Github Action.

deployments/engine/docker/dind.Dockerfile
CharlesCheung96: image

CharlesCheung96: In jenkins pipeline, we rely on two images: `docker-cli in docker` is built in this file, and `dockerd in docker` is pulled from https://hub.docker.com/_/docker/tags?page=1&name=dind.

amyangfei (approved):

Makefile
maxshuang: After moving log to file, do we need to go to the `Artifect` page to get the log, or we can see the log in the first-tier page directly?

maxshuang: Since jenkins pipeline would lost some logs, we save output of engine_integration_tests to file.
--- For the user, what happens after this pr?

CharlesCheung96: If the log is missing, which means neither `Run all test success` nor `[Makefile:503: engine_integration_test] Error 1` occur in the output, you can find the full output in the Artifect page.

maxshuang: you mean `either ... or` ?

maxshuang (approved):

PR #6948, opened 2022-09-01
metrics: remove metrics proxy

GMHDBJD:

Issue Number: ref #4287

- remove metrics proxy
- upgrade promethus and use `DeletePartialMatch` if we want to delete by partial labels.

#### Tests 
Manaully test

```release-note
None
```


CharlesCheung96:
The integration test is fixed in this [pr](https://github.com/pingcap/tiflow/pull/6945/files#diff-bffb058b78b9842d5e4fdb46a7fd39537fe4a2ff843e57773db583da3d997392L1), please rebase this branch and trigger engine tests again.

lance6716 (approved):

GMHDBJD:
cc @D3Hunter @buchuitoudegou

buchuitoudegou (approved):
Would you mind introduce a bit below this comment about why we used this before and why we abandon it now?

GMHDBJD:
> Would you mind introduce a bit below this comment about why we used this before and why we abandon it now?

see https://github.com/pingcap/dm/pull/575 for more detail. We want to remove metrics when task/source stopped. However, our metrics don't quite fit the specification. e.g. [txnHistogram](https://github.com/pingcap/tiflow/blob/6f9c00d8602a965a7bd763a03081a8661c059d6b/dm/loader/metrics.go#L42) is created with  target_schema/target_table label, but we need to remove when its task/source is deleted, so we need delete the metrics when label is partial match. https://github.com/pingcap/dm/pull/575 wraps metrics and label with a Map, which has perfomance issue. Promethus support `DeletePartialMatch` now. https://github.com/prometheus/client_golang/pull/1013

Issue #6387, opened 2022-07-21
Tracking issue for engine removing dependency on etcd


gozssky:
## Background
Currently, our system relies on ETCD to support service discovery and leader election, and on MySQL to store metadata. To simplify the architecture of our system, we want to reduce the dependency on external services.

Our system is a typical master/worker architecture, all executors are connected to the server master. So we can implement the service discovery on the server master. And the leader election can also be implemented by MySQL. Therefore, we plan to remove the dependency on the ETCD.

## Development Task

- [x] https://github.com/pingcap/tiflow/issues/6388
- [x] https://github.com/pingcap/tiflow/issues/6867
- [x] Abstract leader election interface and support MySQL https://github.com/pingcap/tiflow/pull/6949

PR #6949, opened 2022-09-01
election(engine): replace implementation with mysql

gozssky:

Issue Number: ref https://github.com/pingcap/tiflow/issues/6387

* Abstract leader election interface.
* Replace servermaster's election implementation with MySQL.

After this PR, the engine will completely remove the dependency on ETCD.
To reduce the PR size, this PR doesn't remove the legacy code based on ETCD. Code clean will come in the next PR.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/pkg/election/elector.go
```diff
@@ -0,0 +1,357 @@
...
+		// changed, update the local observedRenews.
+		if !ok || !oldMember.RenewTime.Equal(m.RenewTime) {
+			e.observedRenews[m.ID] = time.Now()
+		}
+	}
+
+	// New leader is elected.
+	if record.LeaderID != "" && record.LeaderID != e.observedRecord.LeaderID {
+		leader, ok := record.FindMember(record.LeaderID)
+		if ok {
```
CharlesCheung96, line 279: Why is this judgment needed here? Is it possible for OK to be false in practice?

gozssky: It's impossible. This is a safety check. `!ok` means `leader == nil`.

engine/pkg/election/elector.go
```diff
@@ -0,0 +1,357 @@
...
+
+	return e.isLeaseExpiredLocked(memberID)
+}
+
+func (e *Elector) isLeaseExpiredLocked(memberID string) bool {
+	member, ok := e.observedRecord.FindMember(memberID)
+	if !ok {
+		return true
+	}
+	renewTime := e.observedRenews[memberID]
```
amyangfei, line 304: Can we ensure `observedRenews` always has the same members with `observedRecord`

engine/pkg/election/elector.go
```diff
@@ -0,0 +1,357 @@
...
+	e.observeLock.RLock()
+	defer e.observeLock.RUnlock()
+
+	for _, m := range e.observedRecord.Members {
+		if e.isLeaseExpiredLocked(m.ID) {
+			continue
+		}
+		if m.ID == e.observedRecord.LeaderID {
+			return m.Clone(), true
+		}
```
amyangfei, line 331: Is it better to check `m.ID == e.observedRecord.LeaderID` first, and break loop immediately if lease of leader id is expired.

engine/pkg/election/elector.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	if record.LeaderID == e.config.ID {
+		record.LeaderID = ""
+	}
+
+	if err := s.Update(ctx, record); err != nil {
+		return errors.Trace(err)
+	}
+	e.setObservedRecord(record)
```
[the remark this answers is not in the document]
gozssky: We rely on the first record to observe member renews, because the second record may not update successfully.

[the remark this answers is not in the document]
gozssky: Since the observed record is usually outdated, updating it once we got a new record would make it more accurate.

engine/pkg/election/elector.go
gozssky: Yes, it's guaranteed in `setObservedRecord`.

engine/pkg/election/elector.go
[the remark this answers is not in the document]
gozssky: `record.Clone` is used to avoid modifying `e.observedRecord` accidentally. Maybe cloning it inside `setObservedRecord` is better?

engine/pkg/election/elector.go
gozssky: Optimized in https://github.com/pingcap/tiflow/pull/6949/commits/69242db0678f279014e9043e3454e8a82cdd19d4.

engine/pkg/election/elector.go
```diff
@@ -0,0 +1,353 @@
...
+
+// NewElector creates a new Elector.
+func NewElector(config Config) (*Elector, error) {
+	if err := config.AdjustAndValidate(); err != nil {
+		return nil, err
+	}
+	return &Elector{
+		config:         config,
+		observedRenews: make(map[string]time.Time),
+		resignCh:       make(chan time.Duration),
```
amyangfei, line 59: Do we need to add one buffer for `resignCh`, and add a default branch at L350?
In some corner case, election could stuck in a deadloop of `renew`, which leads to no channel receiver of `resignCh` and block of `ResignLeader` API.

engine/pkg/election/elector.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	return nil, false
+}
+
+// GetMembers returns all members.
+func (e *Elector) GetMembers() []*Member {
+	e.observeLock.RLock()
+	defer e.observeLock.RUnlock()
+
+	var members []*Member
```
amyangfei, line 337:
```suggestion
	members := make([]*Member, 0, len(e.observedRecord.Members))
```

engine/pkg/election/storage.go
```diff
@@ -0,0 +1,103 @@
...
+	// or any other error encountered.
+	Update(ctx context.Context, record *Record) error
+}
+
+// Record holds the information of a leader election.
+type Record struct {
+	// LeaderID is the ID of the leader. If it is empty, it means
+	// there is no leader and all members may try to become leader.
+	LeaderID string `json:"leader_id"`
+	// Members is the members that are eligible to become leader.
```
amyangfei, line 43: `Members` also contains current leader if it exists, better to make the comment more clear.

engine/pkg/election/config.go
```diff
@@ -0,0 +1,101 @@
...
+	//
+	// It defaults to 2 seconds if not set.
+	RenewInterval time.Duration
+	// RenewDeadline is the duration that the current member waits before
+	// it gives up on the leadership when it can't renew the lease. To avoid
+	// two members from both acting as leader at the same time, RenewDeadline
+	// should be set shorter than LeaseDuration - RenewInterval.
+	//
+	// It defaults to 8 seconds if not set.
+	RenewDeadline time.Duration
```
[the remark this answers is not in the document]
gozssky: Currently, there are exposed to package users.

engine/pkg/election/elector.go
gozssky: There is a timeout for `renew` loop. So this API shouldn't be blocked for too long. No buffer is intended to make the resign duration more intuitive (It's calculated after API response).

gozssky: But there is indeed a bug here. When `elector` is closed, `ResignLeader` will be blocked forever. Maybe we should add a timeout here.

engine/pkg/election/config.go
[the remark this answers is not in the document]
amyangfei: Yep, I mean will they be exposed to end users later?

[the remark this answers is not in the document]
gozssky: Yes, I will add a config later. But may not reuse this struct.

amyangfei (approved):

amyangfei:
PTAL @maxshuang @CharlesCheung96

CharlesCheung96 (approved):

engine/pkg/election/elector.go
```diff
@@ -0,0 +1,384 @@
...
+	if !e.callbackIsRunning.Load() {
+		leaderCallback := e.config.LeaderCallback
+		leaderCtx, leaderCancel := context.WithCancel(ctx)
+		e.callbackWg.Add(1)
+		e.callbackIsRunning.Store(true)
+		go func() {
+			defer func() {
+				e.callbackIsRunning.Store(false)
+				e.callbackWg.Done()
+				leaderCancel()
```
amyangfei, line 218: Can we remove `leaderCancel()` here, since it is always triggered in `leaderCancel`.

gozssky:
/run-engine-integration-test

engine-integration-test is unstable due to MySQL being too slow. The root cause is
`[2022/09/21 08:18:18.659 +00:00] [ERROR] [server.go:792] ["framework metastore initializes all backend tables fail"] [error="[DFLOW:ErrMetaOpFail]meta operation fail: context deadline exceeded"]`

engine/pkg/election/elector.go
gozssky: When the callback is unexpectedly exited, the elector will not cancel the `leaderCtx`.

amyangfei (approved):

PR #6955, opened 2022-09-02
servermaster(engine): add backoff mechanism for job

amyangfei:

Issue Number: ref #6749

- Add a job backoff mechanism
   - Use `github.com/cenkalti/backoff/v4` underlying
   - Each time a job fails, the backoff interval will move forward
   - If a job is online for more than `backoff reset interval` (default 10 minutes), the backoff will be reset.
- A job won't be re-created until the job backoff manager `Allow` returns true.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


engine/servermaster/jobop/backoff.go
```diff
@@ -0,0 +1,191 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package jobop
```
CharlesCheung96, line 14: Jobop is somewhat confusing, please add some comments.

engine/servermaster/jobop/backoff.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+}
+
+type backoffEvent struct {
+	tp backoffEventType
+	ts time.Time
+}
+
+// NewJobBackoff creates a new job backoff
+func NewJobBackoff(jobID string, options ...BackoffOption) *JobBackoff {
```
CharlesCheung96, line 94:
```go
func NewJobBackoff(jobID string, options ...BackoffOption) *JobBackoff {
	opts := &backoffOpts{
		resetInterval:   defaultBackoffRestInterval,
		initialInterval: defaultBackoffInitInterval,
		maxInterval:     defaultBackoffMaxInterval,
		multiplier:      defaultBackoffMultiplier,
		clocker:         clock.New(),
	}
	for _, option := range options {
		option(opts)
	}
	errBackoff := backoff.NewExponentialBackOff()
	errBackoff.InitialInterval = opts.initialInterval
	errBackoff.MaxInterval = opts.maxInterval
	errBackoff.Multiplier = opts.multiplier
	errBackoff.Reset()

	return &JobBackoff{
		jobID:      jobID,
		opts:       opts,
		errBackoff: errBackoff,
	}
}
```

engine/servermaster/jobmanager.go
```diff
@@ -447,11 +450,18 @@ func (jm *JobManagerImpl) Tick(ctx context.Context) error {
...
 		}
+		if derrors.ErrMasterCreateWorkerBackoff.Equal(err) {
+			log.Debug("create worker is being backoff, retry later", zap.Error(err))
+			return true, nil
+		}
 		return false, err
 	}
 
 	err := jm.JobFsm.IterPendingJobs(
 		func(job *frameModel.MasterMetaKVData) (string, error) {
```
maxshuang, line 461: Every job has its own backoff inner state, why a dis-allowed job affect other job's retry process?

engine/servermaster/jobop/backoff.go
```diff
@@ -0,0 +1,191 @@
...
+	b.errBackoff = errBackoff
+
+	return b
+}
+
+// JobBackoff is a job backoff manager, it recoreds job online and offline events
+// and determines whether a job can be re-created based on backoff mechanism.
+type JobBackoff struct {
+	jobID           string
+	events          []backoffEvent
```
maxshuang, line 123: I don't really get the point of this `event queue`.
In my mind, we always get the seqence `OnWorkerOnline, OnWorkerOffline, OnWorkerOnline, OnWorkerOffline...`, so every `OnWorkerOffline` will update the `backoffInterval`, a long `online` interval will reset the `backoffInterval`.

So a `lastOnlineTs`  seems enough for me.

engine/servermaster/jobop/backoff.go
maxshuang: Yes. A similar `jobop` in `CancelJob`.

amyangfei: They will be in the same package actually, which contains high level job operations.

engine/servermaster/jobop/backoff.go
amyangfei: The sequence could be `Offline`, `Offline`, `Offline` without any Online.
Currently implement can only record the latest online ts, but it is flexible to record all events.

engine/servermaster/jobmanager.go
amyangfei: fixed, `ErrMasterCreateWorkerBackoff` should be ignored in job fsm.

engine/servermaster/jobop/backoff.go
amyangfei: added some comments

maxshuang (approved):

CharlesCheung96 (approved):

engine/servermaster/jobop/backoff.go
```diff
@@ -0,0 +1,194 @@
...
+	// reset the backoff
+	if len(b.events) > 0 {
+		lastEvent := b.events[len(b.events)-1]
+		if lastEvent.tp == backoffOnline &&
+			b.opts.clocker.Since(lastEvent.ts) >= b.opts.resetInterval {
+			b.events = make([]backoffEvent, 0)
+			b.resetErrBackoff()
+		}
+	}
+	b.events = append(b.events, event)
```
maxshuang, line 178: Actually, we only need `lastOnlineTs` to do this check.
```
 if lastOnlineTs >0 && b.opts.clocker.Since(lastEvent.ts) >= b.opts.resetInterval {
			b.resetErrBackoff()
}
```

Set lastOnlineTs = 0 every time we receive the `offline`  event. 
Set lastOnlineTs = xxx every time we receive the `online`  event.

PR #7024, opened 2022-09-08
pkg/pdutil(ticdc): improve pd time accuracy

overvenus:

Issue Number: ref #4757

Add duration to ts event time.

#### Tests 

 - Unit test

No.

No.

### Release note 

```release-note
None
```


pkg/pdutil/clock.go
```diff
@@ -41,32 +41,38 @@ type clock struct {
...
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	ret.mu.timeCache = oracle.GetTimeFromTS(oracle.ComposeTS(physical, 0))
+	ret.mu.tsEventTime = oracle.GetTimeFromTS(oracle.ComposeTS(physical, 0))
+	ret.mu.tsProcessingTime = time.Now()
 	return ret, nil
 }
 
 // Run will get time from pd periodically to cache in timeCache
```
asddongmen, line 71: `timeCache` in the comment should be remove.

pkg/pdutil/clock.go
```diff
@@ -100,10 +109,10 @@ func (c *clock) Run(ctx context.Context) {
 // CurrentTime returns current time from timeCache
```
asddongmen, line 109: ditto

pkg/pdutil/clock.go
```diff
@@ -100,10 +109,10 @@ func (c *clock) Run(ctx context.Context) {
...
 func (c *clock) CurrentTime() (time.Time, error) {
 	c.mu.RLock()
-	err := c.mu.err
-	cacheTime := c.mu.timeCache
-	c.mu.RUnlock()
-	return cacheTime, errors.Trace(err)
+	defer c.mu.RUnlock()
+	tsEventTime := c.mu.tsEventTime
+	current := tsEventTime.Add(time.Since(c.mu.tsProcessingTime))
```
asddongmen, line 114: why can we just return `tsEventTime.Add(time.Since(c.mu.tsProcessingTime))`?

overvenus: To make code easier to understand

3AceShowHand (approved):

asddongmen (approved):

pkg/pdutil/clock.go
```diff
@@ -41,32 +41,38 @@ type clock struct {
...
+	ret.mu.tsEventTime = oracle.GetTimeFromTS(oracle.ComposeTS(physical, 0))
+	ret.mu.tsProcessingTime = time.Now()
 	return ret, nil
 }
 
-// Run will get time from pd periodically to cache in timeCache
+// Run gets time from pd periodically.
 func (c *clock) Run(ctx context.Context) {
 	ctx, cancel := context.WithCancel(ctx)
 	c.cancel = cancel
```
3AceShowHand, line 74: `cancel` is called by `stop`, cause race detected in the test.

pkg/pdutil/clock.go
```diff
@@ -81,33 +89,38 @@ func (c *clock) Run(ctx context.Context) {
...
+	tsEventTime := c.mu.tsEventTime
+	current := tsEventTime.Add(time.Since(c.mu.tsProcessingTime))
+	return current, errors.Trace(c.mu.err)
 }
 
 // Stop clock.
 func (c *clock) Stop() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
 	c.cancel()
```
3AceShowHand, line 124: it looks that the `cancel` maybe nil?

PR #7034, opened 2022-09-09
externalresources(engine): Support s3 as storage (Part I)

liuzix:

Issue Number: ref #7114

- Implement `FileManager` for s3.

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
None.
```


engine/pkg/externalresource/s3/s3_file_manager.go
```diff
@@ -0,0 +1,254 @@
...
+			&internal.ResourceFilesNotFoundError{
+				Ident: ident,
+			})
+	}
+
+	return desc, nil
+}
+
+// RemoveTemporaryFiles removes all temporary resources (those that are not persisted).
+// It can only be used to clean up resources created by the local executor.
```
CharlesCheung96, line 125: Better to check executorID within this function.

liuzix: I think the caller can guarantee it. This requirement is here due to the risk of data race in s3.

engine/pkg/externalresource/s3/index.go
```diff
@@ -0,0 +1,309 @@
...
+
+	select {
+	case <-ctx.Done():
+		return errors.Trace(ctx.Err())
+	case err := <-done:
+		if err != nil {
+			return err
+		}
+		return nil
+	}
```
amyangfei, line 259: Could it block here if `indexManager` is closed and context is still alive.

engine/pkg/externalresource/s3/index.go
```diff
@@ -0,0 +1,309 @@
...
+		m.indexFiles[ident.WorkerID] = index
+	}
+
+	ok = index.SetPersisted(ident)
+	m.indexMu.Unlock()
+	if !ok {
+		return false, nil
+	}
+
+	if err := m.triggerFlush(ctx, ident.WorkerID); err != nil {
```
amyangfei, line 176: Each persisted file will trigger one file flush to s3, do we need to introduce some batch mechanism for index file flush?

engine/pkg/externalresource/s3/index.go
```diff
@@ -0,0 +1,309 @@
...
+		return nil
+	}
+
+	bytes, err := index.Serialize()
+	m.indexMu.Unlock()
+	if err != nil {
+		return err
+	}
+
+	storage, err := m.createStorageForIndexFile(ctx, internal.ResourceScope{
```
amyangfei, line 282: Is it necessary to create a new storage each time, could the index manager reuse one storage instance?

engine/pkg/externalresource/internal/file_manager.go
```diff
@@ -0,0 +1,72 @@
...
+	Name resModel.ResourceName
+}
+
+// Scope returns the Scope of the ResourceIdent.
+func (i ResourceIdent) Scope() ResourceScope {
+	return i.ResourceScope
+}
+
+// FileManager abstracts the operations on the underlying storage.
+type FileManager interface {
```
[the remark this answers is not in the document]
CharlesCheung96: local flie manager implment the interface in [this pr](https://github.com/pingcap/tiflow/pull/7127).

engine/pkg/externalresource/s3/index.go
```diff
@@ -0,0 +1,309 @@
...
+		return nil, errors.Annotate(err, "serialize index file to s3")
+	}
+	return bytes, nil
+}
+
+func (f *indexFile) GetPersistedFileSet() map[model.ResourceName]struct{} {
+	return f.PersistedFileSet
+}
+
+type indexManagerImpl struct {
```
amyangfei, line 104: Would it better to implement `indexManager` with metastore, which may be more flexible, and don't need to worry about rate limit of S3.

amyangfei:
Closed by #7127 and #7204.

PR #7035, opened 2022-09-09
framework(engine): record error message when job master closes

### LINKED ISSUE #7021: error message is not set in master meta when business job master meets error
### What did you do?

Hack a business job master, returns error in `Tick` API

Business job master fails and records `error_message` in master_meta table in metastore.

### What did you see instead?

Business job master fails but `error_message` is empty.

### Versions of the cluster

master@https://github.com/pingcap/tiflow/commit/3e43a18b34d8f052c33c83266c9f4a10f5abf666

amyangfei:

Issue Number: close #7021

- Update `error_message` field before job master calls `doClose`
- Change the interface signature of `UpdateJob` to explicit way, which means the parameter contains all fields that need to be updated.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/engine/framework/base_jobmaster.go
+++ b/engine/framework/base_jobmaster.go
@@ -201,6 +201,7 @@ func (d *DefaultBaseJobMaster) Init(ctx context.Context) error {
 
 	if isFirstStartUp {
 		if err := d.impl.InitImpl(ctx); err != nil {
+			d.errCenter.OnError(err)
 			return errors.Trace(err)
 		}
 		if err := d.master.markStateInMetadata(ctx, frameModel.MasterStateInit); err != nil {
@@ -208,6 +209,7 @@ func (d *DefaultBaseJobMaster) Init(ctx context.Context) error {
 		}
 	} else {
 		if err := d.impl.OnMasterRecovered(ctx); err != nil {
+			d.errCenter.OnError(err)
 			return errors.Trace(err)
 		}
 	}
@@ -234,6 +236,7 @@ func (d *DefaultBaseJobMaster) Poll(ctx context.Context) error {
 		return nil
 	}
 	if err := d.impl.Tick(ctx); err != nil {
+		d.errCenter.OnError(err)
 		return errors.Trace(err)
 	}
 	return nil
@@ -253,6 +256,7 @@ func (d *DefaultBaseJobMaster) Close(ctx context.Context) error {
 		}
 	})
 
+	d.master.persistMetaError()
 	d.master.doClose()
 	d.worker.doClose()
 	return nil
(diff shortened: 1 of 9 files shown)


engine/framework/base_jobmaster.go
```diff
@@ -201,13 +201,15 @@ func (d *DefaultBaseJobMaster) Init(ctx context.Context) error {
...
 		if err := d.impl.InitImpl(ctx); err != nil {
+			d.errCenter.OnError(err)
 			return errors.Trace(err)
 		}
 		if err := d.master.markStateInMetadata(ctx, frameModel.MasterStateInit); err != nil {
 			return errors.Trace(err)
 		}
 	} else {
 		if err := d.impl.OnMasterRecovered(ctx); err != nil {
+			d.errCenter.OnError(err)
```
maxshuang, line 212: 1. Would it be better to set the `errCenter` in `eventLoop`? Or we may lost some error message if we forget to set the `errCenter`.
2. Please add some comment here if we have specific strategy when record the error message.

maxshuang:
Rest LGTM.

engine/framework/base_jobmaster.go
amyangfei: 1. ErrCenter is logic oriented, instead of runtime work, it is better to record error in implement code, rather than in runtime layer.
2. Comments added. Only record error message when business API returns error.

amyangfei: PTAL @maxshuang

CharlesCheung96 (approved):

engine/framework/master.go
```diff
@@ -737,9 +747,8 @@ func (m *DefaultBaseMaster) exitWithoutSetErrCenter(ctx context.Context, exitRea
 		m.masterMeta.ErrorMsg = err.Error()
 	}
```
maxshuang, line 748: Please clear the `ErrorMsg` here if the input `err` is nil, or the old ErrorMsg will remain in the job meta table even the job is finished normally.

engine/framework/master.go
```diff
@@ -592,13 +593,22 @@ func (m *DefaultBaseMaster) markStateInMetadata(
...
+
+func (m *DefaultBaseMaster) persistMetaError() {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
+	defer cancel()
 
-	masterMeta.State = code
-	return metaClient.Update(ctx, masterMeta)
+	if err := m.errCenter.CheckError(); err != nil {
+		metaClient := metadata.NewMasterMetadataClient(m.id, m.frameMetaClient)
+		m.masterMeta.ErrorMsg = err.Error()
```
maxshuang, line 606: Be carefully and clear the `ErrorMsg` in other place to avoid error setting the job meta.

engine/framework/master.go
amyangfei: PTAL @maxshuang

maxshuang (approved):

PR #7036, opened 2022-09-09
syncer(dm): Fix follow-up bugs after fixing nondeterministic DDL

### LINKED ISSUE #6628: follow-up after fixing Nondeterministic DDL
### What did you do?

1. modification to `job.ddls` should be as late as possible
https://github.com/pingcap/tiflow/blob/5a0452d3fefb157244520a11f21f102de57a1937/dm/syncer/syncer.go#L1332-L1336

to avoid interfering the logic between modification and using. Like https://github.com/pingcap/tiflow/blob/5a0452d3fefb157244520a11f21f102de57a1937/dm/syncer/syncer.go#L1373

2. there're too many warnings in log, we know in most of time there's no timezone information in status vars, so should not raise warning so frequently
https://github.com/pingcap/tiflow/blob/5a0452d3fefb157244520a11f21f102de57a1937/dm/syncer/shardddl.go#L158

_No response_

### What did you see instead?

_No response_

### Versions of the cluster

master

```console
```

lyzx2001:

Issue Number: close #6628 

Modification to `job.ddls` is put to the place as late as possible, which is just before `db.ExecuteSQLWithIgnore`.

The appearing time of the Warn message is reduced by adding conditions to trigger it. Now only when no timezone information was retrieved from `GetTimezoneByStatusVars` and upstream timezone has not been previously set, the warn message would be triggered.

The changes in `engine/pkg/orm/client.go` and `engine/servermaster/jobop/backoff.go` are caused by `make fmt` of go version 1.19.

#### Tests 

 - Integration test

No.
No.
### Release note 

```release-note
Reduce the appearing time of the warning message "found error when getting timezone from binlog status_vars" in dm-worker log.
```

### DIFF
--- a/dm/syncer/syncer.go
+++ b/dm/syncer/syncer.go
@@ -1349,26 +1349,6 @@ func (s *Syncer) syncDDL(queueBucket string, db *dbconn.DBConn, ddlJobChan chan 
 			}
 		}
 
-		// set timezone
-		if ddlJob.timezone != "" {
-			s.timezoneLastTime = ddlJob.timezone
-			setTimezoneSQL := fmt.Sprintf("SET SESSION TIME_ZONE = '%s'", ddlJob.timezone)
-			ddlJob.ddls = append([]string{setTimezoneSQL}, ddlJob.ddls...)
-			setTimezoneSQLDefault := "SET SESSION TIME_ZONE = DEFAULT"
-			ddlJob.ddls = append(ddlJob.ddls, setTimezoneSQLDefault)
-		} else if s.timezoneLastTime != "" {
-			// use last time's time zone
-			setTimezoneSQL := fmt.Sprintf("SET SESSION TIME_ZONE = '%s'", s.timezoneLastTime)
-			ddlJob.ddls = append([]string{setTimezoneSQL}, ddlJob.ddls...)
-			setTimezoneSQLDefault := "SET SESSION TIME_ZONE = DEFAULT"
-			ddlJob.ddls = append(ddlJob.ddls, setTimezoneSQLDefault)
-		}
-		// set timestamp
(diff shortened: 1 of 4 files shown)


lyzx2001:
@lance6716

lyzx2001:
@lance6716

dm/syncer/syncer.go
```diff
@@ -1369,6 +1349,26 @@ func (s *Syncer) syncDDL(queueBucket string, db *dbconn.DBConn, ddlJobChan chan
...
+			ddlJob.ddls = append([]string{setTimezoneSQL}, ddlJob.ddls...)
+			setTimezoneSQLDefault := "SET SESSION TIME_ZONE = DEFAULT"
+			ddlJob.ddls = append(ddlJob.ddls, setTimezoneSQLDefault)
+		}
+		// set timestamp
+		setTimestampSQL := fmt.Sprintf("SET TIMESTAMP = %d", ddlJob.timestamp)
+		ddlJob.ddls = append([]string{setTimestampSQL}, ddlJob.ddls...)
+		setTimestampSQLDefault := "SET TIMESTAMP = DEFAULT"
+		ddlJob.ddls = append(ddlJob.ddls, setTimestampSQLDefault)
+
```
lance6716, line 1371: seems these line can be further moved into `if !ignore {}` in line 1378

lance6716 (approved):
rest lgtm

dm/syncer/syncer.go
lyzx2001: but when `ignore` is true, these lines would not be executed, and the timestamp and timezone would not be set back to default, I'm not sure whether this would cause some bugs

GMHDBJD (approved):
LGTM

PR #7059, opened 2022-09-13
syncer: log KeyNotFound error for update/delete

### LINKED ISSUE #6383: DM silently ignores update row and delete row events when the row doesn't exist on the target
### What did you do?

MySQL 8.0.29 → TiDB 6.1.0 with DM 6.1.0 with GTID.

On MySQL (source):
```
mysql> CREATE TABLE t1 (id int primary key, c1 varchar(255));
Query OK, 0 rows affected (0.09 sec)

mysql> INSERT INTO t1 VALUES (1, 'test 1');
Query OK, 1 row affected (0.02 sec)

mysql> SET sql_log_bin=OFF;
Query OK, 0 rows affected (0.00 sec)

mysql> INSERT INTO t1 VALUES (2, 'test 2');
Query OK, 1 row affected (0.01 sec)

mysql> INSERT INTO t1 VALUES (3, 'test 3');
Query OK, 1 row affected (0.02 sec)

mysql> SET sql_log_bin=ON;
Query OK, 0 rows affected (0.00 sec)

mysql> UPDATE t1 SET c1='test 2 again' WHERE id=2;
Query OK, 1 row affected (0.02 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> DELETE FROM t1 WHERE id=3;
Query OK, 1 row affected (0.02 sec)

mysql> INSERT INTO t1 VALUES (4, 'test 4');
Query OK, 1 row affected (0.01 sec)

mysql> TABLE t1;
+----+--------------+
| id | c1           |
+----+--------------+
|  1 | test 1       |
|  2 | test 2 again |
|  4 | test 4       |
+----+--------------+
3 rows in set (0.00 sec)
```

On TiDB (target)
```
tidb> TABLE t1;
+----+--------+
| id | c1     |
+----+--------+
|  1 | test 1 |
|  4 | test 4 |
+----+--------+
2 rows in set (0.01 sec)
```

An error (or at least a warning) about the fact that:

- The row with id=2 couldn't be updated
- The row with id=3 couldn't be deleted

### What did you see instead?

DM continueing to replicate data without any warning or error

### Versions of the cluster

MySQL 8.0.29 → TiDB 6.1.0 with DM 6.1.0

```console
» query-status demo-mysql8-tidb
{
    "result": true,
    "msg": "",
    "sources": [
        {
            "result": true,
            "msg": "",
            "sourceStatus": {
                "source": "mysql8",
                "worker": "worker1",
                "result": null,
                "relayStatus": null
            },
            "subTaskStatus": [
                {
                    "name": "demo-mysql8-tidb",
                    "stage": "Running",
                    "unit": "Sync",
                    "result": null,
                    "unresolvedDDLLockID": "",
                    "sync": {
                        "totalEvents": "31",
                        "totalTps": "0",
                        "recentTps": "0",
                        "masterBinlog": "(binlog.000001, 4914)",
                        "masterBinlogGtid": "e7d25aa3-0833-11ed-830d-52ff284f3c30:1-17",
                        "syncerBinlog": "(binlog.000001, 4914)",
                        "syncerBinlogGtid": "e7d25aa3-0833-11ed-830d-52ff284f3c30:1-17",
                        "blockingDDLs": [
                        ],
                        "unresolvedGroups": [
                        ],
                        "synced": true,
                        "binlogType": "remote",
                        "secondsBehindMaster": "0",
                        "blockDDLOwner": "",
                        "conflictMsg": ""
                    }
                }
            ]
        }
    ]
}

```

GMHDBJD:

Issue Number: close #6383 

- return affect when execute sql
- check affect equals lens of dml jobs

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
log ErrKeyNotFound for update row and delete row events when the row doesn't exist on the target
```


dm/pkg/conn/baseconn.go
```diff
@@ -169,23 +170,25 @@ func (conn *BaseConn) ExecuteSQLWithIgnoreError(tctx *tcontext.Context, hVec *pr
 		}
 
 		startTime = time.Now()
-		_, err = txn.ExecContext(tctx.Context(), query, arg...)
-		if err == nil {
+		result, err2 := txn.ExecContext(tctx.Context(), query, arg...)
+		if err2 == nil {
+			rows, _ := result.RowsAffected()
```
lance6716, line 175: not sure this value is correct for TiDB optimistic transaction 😂

dm/syncer/dml_worker.go
```diff
@@ -312,3 +320,15 @@ func (w *DMLWorker) genSQLs(jobs []*job) ([]string, [][]interface{}) {
 	}
 	return queries, args
 }
+
+func (w *DMLWorker) judgeKeyNotFound(affect int, jobs []*job) bool {
+	if w.compact || w.multipleRows {
```
lance6716, line 325: for multipleRows, we can count the row numbers rather than job number.

And compact should also be OK?

dm/tests/shardddl1/run.sh
```diff
@@ -711,6 +711,47 @@ function DM_MULTIPLE_ROWS() {
...
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+
+	run_sql_tidb "delete from ${shardddl}.${tb} where id=1;"
+	run_sql_tidb "delete from ${shardddl}.${tb} where id=2;"
+	run_sql_source1 "update ${shardddl1}.${tb1} set id=3 where id=1;"
+	run_sql_source2 "update ${shardddl1}.${tb1} set id=4 where id=2;"
+	run_sql_source1 "delete from ${shardddl1}.${tb1} where id=3;"
+	run_sql_source2 "delete from ${shardddl1}.${tb1} where id=4;"
+	check_log_contain_with_retry "no matching record is found to update/delete, ER_KEY_NOT_FOUND" $WORK_DIR/worker1/log/dm-worker.log $WORK_DIR/worker2/log/dm-worker.log
+	check_log_contains $WORK_DIR/worker1/log/dm-worker.log "no matching record is found to update/delete, ER_KEY_NOT_FOUND" 2
```
lance6716, line 724: why it will appear twice?

dm/tests/shardddl1/run.sh
```diff
@@ -711,6 +711,47 @@ function DM_MULTIPLE_ROWS() {
...
+function DM_KEY_NOT_FOUND_CASE() {
+	check_sync_diff $WORK_DIR $cur/conf/diff_config.toml
+
+	run_sql_tidb "delete from ${shardddl}.${tb} where id=1;"
+	run_sql_tidb "delete from ${shardddl}.${tb} where id=2;"
+	run_sql_source1 "update ${shardddl1}.${tb1} set id=3 where id=1;"
+	run_sql_source2 "update ${shardddl1}.${tb1} set id=4 where id=2;"
+	run_sql_source1 "delete from ${shardddl1}.${tb1} where id=3;"
+	run_sql_source2 "delete from ${shardddl1}.${tb1} where id=4;"
+	check_log_contain_with_retry "no matching record is found to update/delete, ER_KEY_NOT_FOUND" $WORK_DIR/worker1/log/dm-worker.log $WORK_DIR/worker2/log/dm-worker.log
```
lance6716, line 723: and we can check worker1/worker2 log one by one, that will make test more stable

dm/syncer/dml_worker.go
GMHDBJD: in multipleRows, update will be transform to insert on duplicate update, if downstream may lack the previous key and return wrong affect.

dm/tests/shardddl1/run.sh
GMHDBJD: update once and delete once

dm/syncer/dml_worker.go
GMHDBJD: for example
```
update 1,'aa'
update 1,'aaa'
```
will be compacted to one job and executed as followed
```
insert into xx(1,'aaa') on duplicate key update x=VALUES(x)
```
If downstream lack key 1, above sql will return affect=1. But acually it should return an error.

GMHDBJD: for example
```
insert 1
delete 1
```
will be compacted to one job and executed as followed
```
delete 1
```
which will return affect=0, but it should not be an error.

GMHDBJD: for example
```
delete 1
insert 1
```
will be compacted to one job and executed as followed
```
insert into xx(1) on duplicate key update x=VALUES(x)
```
If downstream lack key 1, above sql will return affect=1. But acually it should return an error when `delete 1` which has been compacted.

buchuitoudegou: Thx for your explanation.

Is this feature generally unavailable when users enable `compact` or `multipleRows`? In the situation demonstrated above, does DM just simply skip checking `affect`?

GMHDBJD: Yes, because we now it's a warning log and we don't want to add a new config item, such as 'ErrKeyNotFound', so the `compact` and `mutipleRows` has higher priority.

GMHDBJD:
cc @lance6716 @buchuitoudegou

dm/syncer/dml_worker.go
lance6716: 1. For multipleRows, we should be able to check the affected rows number is
   - len(job) for INSERT
   - 2*len(job) for INSERT ON DUPLIDATE KEY UDPATE. ref https://dev.mysql.com/doc/refman/5.6/en/insert-on-duplicate.html
   - len(job) for DELETE

2. For compacted job, we still can check that the compected DML should have expected affected rows number? DELETE + INSERT = UPDATE, and check it by the expected value of UPDATE or INSERT ON DUPLIDATE KEY UDPATE

GMHDBJD: Compact: there are still many cases, like update primary key will be transform to DELETE+INSERT, which actually should be 1 row affected
MultipleRows: we need to calculate expected affected rows base on the sql type. I leave a TODO😂

dm/tests/shardddl1/run.sh
```diff
@@ -711,6 +711,46 @@ function DM_MULTIPLE_ROWS() {
...
+	check_process_exit worker2 20
+	export GO_FAILPOINTS='github.com/pingcap/tiflow/dm/syncer/SafeModeInitPhaseSeconds=return(0)'
+	run_dm_worker $WORK_DIR/worker1 $WORKER1_PORT $cur/conf/dm-worker1.toml
+	run_dm_worker $WORK_DIR/worker2 $WORKER2_PORT $cur/conf/dm-worker2.toml
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER1_PORT
+	check_rpc_alive $cur/../bin/check_worker_online 127.0.0.1:$WORKER2_PORT
+
+	run_case KEY_NOT_FOUND "double-source-optimistic" \
+		"init_table 111 211; \
+	 run_sql_source1 \"insert into ${shardddl1}.${tb1} values(1);\"; \
```
lance6716, line 740: need fmt

lance6716 (approved):

dm/tests/shardddl1/run.sh
GMHDBJD: already format

D3Hunter (approved):

buchuitoudegou (approved):

PR #7061, opened 2022-09-14
executor(engine): support fast fail on business unretryable error

amyangfei:

Issue Number: ref #6915 

task-2

- Check error message in master meta before creating a new job, in executor side only, for two reasons
   - tiflow master is not business awareness, which means it doesn't handle business logic via registry.
   - Add this check in executor to unify the un-retryable message mechanism. ref: #6925

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


engine/executor/server.go
```diff
@@ -220,19 +229,46 @@ func (s *Server) makeTask(
 	return taskutil.WrapWorker(newWorker), nil
 }
 
+// precheckMasterMeta checks job master metadata before running it, stop task
+// creating if job master has met a business unretryable error.
+// Return error means meets failure in this function or job creation should be
+// terminated.
+func precheckMasterMeta(
```
maxshuang, line 236: Can we split the `IsRetryableError` from `registry` to make it more common?
Since we have the `GracefulExit` mechanism, we can check the busniess error in `jobmanager` to decide whether we need to redispatch this jobmaster.

maxshuang (approved):

engine/executor/server.go
```diff
@@ -212,6 +215,12 @@ func (s *Server) makeTask(
 		log.Error("Failed to create worker", zap.Error(err))
 		return nil, err
 	}
+	if _, ok := newWorker.(framework.BaseJobMaster); ok {
+		err := precheckMasterMeta(dctx, globalRegistry, workerID, workerType)
```
lance6716, line 219: Seems this PR is to solve `CreateWorker` return a `nil` `err` but the error is written to metastore. How could this happen? the jobmaster is not initialized, no business logic can run and no error can be written to metastore.

amyangfei: It solves the scenario that `Tick` or `InitImpl` of business logic returns error.

lance6716: CreateWorker will not call `Tick` or `InitImpl`? I noticed when code runs here, `OnOpenAPIInitialized` is not called, and from our doc `OnOpenAPIInitialized` is the first callback function.

amyangfei: error in metadata is not persisted here, it **was recorded** by last round of job master.

lance6716 (approved):

PR #7072, opened 2022-09-15
master(engine): exit operator goroutine after job manager closes

### LINKED ISSUE #7071: job operator goroutine doesn't exit if job manager meets error
### What did you do?

Hack the poll method of job manager, return an error

server master leader exits

### What did you see instead?

server master leader hangs, with following goroutine

```go
goroutine 240 [select]:
github.com/pingcap/tiflow/engine/servermaster.(*JobManagerImpl).bgJobOperatorLoop.func1()
    github.com/pingcap/tiflow/engine/servermaster/jobmanager.go:798 +0x105
golang.org/x/sync/errgroup.(*Group).Go.func1()
    golang.org/x/sync@v0.0.0-20220722155255-886fb9371eb4/errgroup/errgroup.go:75 +0x64
created by golang.org/x/sync/errgroup.(*Group).Go
    golang.org/x/sync@v0.0.0-20220722155255-886fb9371eb4/errgroup/errgroup.go:72 +0xa5
```

### Versions of the cluster

master@cf4c080db3dbcd978d150aaf5e1efbe09f299abe

amyangfei:

Issue Number: close #7071

- Close JobOperatorNotifier to exit job operator background goroutine.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


engine/servermaster/jobmanager_test.go
```diff
@@ -664,3 +630,40 @@ func TestOnWorkerDispatchedFastFail(t *testing.T) {
...
+
+func TestJobOperatorBgLoop(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	masterID := "job-operator-bg-loop-test"
+	mockMaster, mgr := prepareMockJobManager(ctx, t, masterID)
+	mockMaster.On("InitImpl", mock.Anything).Return(nil)
```
CharlesCheung96, line 642:
```suggestion
```

CharlesCheung96: Ref: https://github.com/pingcap/tiflow/pull/7072/files#diff-853e09825aa9d13f03821b40301ca876e7ad6b3c1e40badca6c48794d7220df4R54

amyangfei: Removed this line from `prepareMockJobManager`, since mock functions should be customized outside.

engine/servermaster/jobop/mock_operator.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"sync"
+
+	"github.com/stretchr/testify/mock"
+	"golang.org/x/net/context"
+)
+
+// MockJobOperator implements JobOperator
+type MockJobOperator struct {
```
gozssky, line 24: Why not generated mock by `mockgen`?

amyangfei: We have two styles to use mock interface, one is using `mockgen`, the other one is inheriting `testify.Mock`. Maybe we should unify to use one style in the future 😂

gozssky: `mockgen` supports mock method implementation on demand, which makes the mock structure more general. In this case. I'm not sure whether `testify.Mock` supports this feature.

See an example in https://github.com/pingcap/tiflow/blob/f6a740339057d4ac1ec047ea8402d64c8bcbee7e/engine/internal/pkg/discovery/agent_test.go#L39.

maxshuang (approved):

engine/servermaster/jobop/mock_operator.go
amyangfei: changed to use `gomock`, it is better than `testify.Mock`. ptal @gozssky

gozssky (approved):

PR #7079, opened 2022-09-15
puller(ticdc):  region worker batch process events

### LINKED ISSUE #7078: ticdc region worker batch process events
region worker batch process events to reduce switch context

sdojjy:

Issue Number: close #7078

batch process event to reduce context switch and some unnecessary object allocation,   as this is the first part of improving the performance of handling one large single table with lots of regions.  this pr is focused on batch 

profile before
image

after
image

CPU usage
image

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


overvenus:
Could you `sync.Pool` to reduce GC overhead?

cdc/kv/client.go
```diff
@@ -108,6 +108,11 @@ type regionStatefulEvent struct {
 	finishedCallbackCh chan struct{}
 }
 
+type resolvedTsEvent struct {
+	resolvedTs uint64
+	regions    []*regionFeedState
```
nongfushanquan, line 113: please add some comments about regions

nongfushanquan (commented):
could you add some comments  to explain the solution of the batch mode

cdc/puller/puller.go
```diff
@@ -198,48 +198,50 @@ func (p *pullerImpl) Run(ctx context.Context) error {
...
-				}
-				// Forward is called in a single thread
-				p.tsTracker.Forward(e.Resolved.Span, e.Resolved.ResolvedTs)
-				resolvedTs := p.tsTracker.Frontier()
-				if resolvedTs > 0 && !initialized {
-					// Advancing to a non-zero value means the puller level
-					// resolved ts is initialized.
-					atomic.StoreInt64(&p.initialized, 1)
-					initialized = true
+				for _, resolvedSpan := range e.Resolved {
```
3AceShowHand, line 201:
```suggestion
                 metricTxnCollectCounterResolved.Add(float64(len(e.Resolved)))
				for _, resolvedSpan := range e.Resolved {
```

cdc/puller/puller.go
```diff
@@ -198,48 +198,50 @@ func (p *pullerImpl) Run(ctx context.Context) error {
 			}
 
 			if e.Resolved != nil {
```
3AceShowHand, line 200:
```suggestion
			if e.Resolved != nil {
			    metricTxnCollectCounterResolved.Add(float64(len(e.Resolved)))
```

cdc/puller/puller.go
```diff
@@ -198,48 +198,50 @@ func (p *pullerImpl) Run(ctx context.Context) error {
...
-				// Forward is called in a single thread
-				p.tsTracker.Forward(e.Resolved.Span, e.Resolved.ResolvedTs)
-				resolvedTs := p.tsTracker.Frontier()
-				if resolvedTs > 0 && !initialized {
-					// Advancing to a non-zero value means the puller level
-					// resolved ts is initialized.
-					atomic.StoreInt64(&p.initialized, 1)
-					initialized = true
+				for _, resolvedSpan := range e.Resolved {
+					metricTxnCollectCounterResolved.Inc()
```
3AceShowHand, line 202:
```suggestion
```

cdc/puller/puller.go
```diff
@@ -198,48 +198,50 @@ func (p *pullerImpl) Run(ctx context.Context) error {
...
+							zap.Any("spans", p.spans),
+						)
+					}
+					// Forward is called in a single thread
+					p.tsTracker.Forward(resolvedSpan.Span, resolvedSpan.ResolvedTs)
+					resolvedTs := p.tsTracker.Frontier()
+					if resolvedTs > 0 && !initialized {
+						// Advancing to a non-zero value means the puller level
+						// resolved ts is initialized.
+						atomic.StoreInt64(&p.initialized, 1)
```
3AceShowHand, line 219: `Puller.IsInitialized()` is only used in the test now, we can remove it

cdc/kv/client.go
```diff
@@ -873,11 +878,11 @@ func (s *eventFeedSession) dispatchRequest(ctx context.Context) error {
 		// After this resolved ts event is sent, we don't need to send one more
 		// resolved ts event when the region starts to work.
 		resolvedEv := model.RegionFeedEvent{
-			RegionID: sri.verID.GetID(),
-			Resolved: &model.ResolvedSpan{
+			// RegionID: sri.verID.GetID(),
```
3AceShowHand, line 881:
```suggestion
```

cdc/kv/region_worker.go
```diff
@@ -470,7 +486,8 @@ func (w *regionWorker) eventHandler(ctx context.Context) error {
 			exitEventHandler = true
 			return
 		}
-		if event.state.isStopped() {
+		// event.state is nil when resolvedTsEvent is not nil
+		if event.state != nil && event.state.isStopped() {
```
3AceShowHand, line 490: how about add a method `event.isStopped()` to wrap this 2 condition check.

3AceShowHand (approved):

cdc/kv/region_worker.go
```diff
@@ -400,38 +412,42 @@ func (w *regionWorker) processEvent(ctx context.Context, event *regionStatefulEv
...
+			})
+		}
+		if err != nil {
+			return err
 		}
 	}
 
-	if event.resolvedTs != nil {
-		if err = w.handleResolvedTs(ctx, event.resolvedTs.Ts, event.state); err != nil {
-			err = w.handleSingleRegionError(err, event.state)
```
nongfushanquan, line 431: why need handleSingleRegionError function in the older version and remove it now ?

cdc/kv/region_worker.go
```diff
@@ -541,7 +558,7 @@ func (w *regionWorker) eventHandler(ctx context.Context) error {
 			// We measure whether the current worker is busy based on the input
 			// channel size. If the buffered event count is larger than the high
 			// watermark, we send events to worker pool to increase processing
-			// throughput. Otherwise we process event in local region worker to
+			// throughput. Otherwise, we process event in local region worker to
```
nongfushanquan, line 561: will  the value of the  highWatermarkMet be affected ?

nongfushanquan (approved):

cdc/kv/region_worker.go
sdojjy: handleResolvedTs throw error only if the context is done

cdc/kv/client.go
sdojjy: done

cdc/puller/puller.go
sdojjy: I will remove it in next PR

cdc/puller/puller.go
sdojjy: done

cdc/kv/client.go
lonng, line 113: It's better to use value type instead of pointer type to reduce memory allocation. Because we can determine the slice length (ref: https://github.com/pingcap/tiflow/pull/7079/files#diff-31a8a1f4c2bd92516aebe3c694c8ab06eb69f88aa09a95212f1f5e81a91fcf1cR1331).

cdc/puller/puller.go
sdojjy: done

cdc/kv/client.go
```diff
@@ -1323,11 +1328,17 @@ func (s *eventFeedSession) sendResolvedTs(
 	worker *regionWorker,
 	addr string,
 ) error {
-	statefulEvents := make([][]*regionStatefulEvent, worker.inputSlots)
+	statefulEvents := make([]*regionStatefulEvent, worker.inputSlots)
```
lonng, line 1331: It's better to use value type instead of pointer type to reduce memory allocation. Because we can determine the slice length.

cdc/kv/region_worker.go
sdojjy: yes , the batch will affect the highWatermarkMet, but i think it's ok to use old highWatermarkMet , because  data event is not batch for now

cdc/kv/client.go
```diff
@@ -1343,16 +1353,16 @@ func (s *eventFeedSession) sendResolvedTs(
...
-			statefulEvents[slot] = append(statefulEvents[slot], &regionStatefulEvent{
-				resolvedTs: resolvedTs,
-				regionID:   regionID,
-				state:      state,
-			})
+			statefulEvents[slot].resolvedTsEvent.regions = append(
+				statefulEvents[slot].resolvedTsEvent.regions, state,
+			)
+			// regionID is just an slot index
+			statefulEvents[slot].regionID = regionID
```
lonng, line 1360: Why not set the `regionID` first and then append it to the slice? Which can eliminate duplication slice index: `statefulEvents[slot]`.

cdc/kv/client.go
sdojjy: ok, I will do it in another PR if needed.

PR #7101, opened 2022-09-16
cdc: refactor protobuf packages

### LINKED ISSUE #7100: TiCDC TableState is defined twice
TableState is defined twice, one in protobuf and anther in processor. We should unify them.

overvenus:

Issue Number: close #7100

Split `table_schedule.proto` and move table definition into package tablepb.

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
None
```


cdc/scheduler/internal/v3/agent/table.go
```diff
@@ -43,48 +43,36 @@ func newTable(
...
-	return schedulepb.TableStatus{
-		TableID: t.id,
-		State:   state,
-		Checkpoint: schedulepb.Checkpoint{
-			CheckpointTs: meta.CheckpointTs,
-			ResolvedTs:   meta.ResolvedTs,
-		},
-	}
+func (t *table) getTableStatus() tablepb.TableStatus {
+	return t.executor.GetTableMeta(t.id)
```
3AceShowHand: how about rename `GetTableMeta` to `GetTableStatus`

3AceShowHand (approved):

asddongmen (approved):

cdc/processor/pipeline/sink.go
```diff
@@ -103,16 +104,37 @@ func (n *sinkNode) getCheckpointTs() model.ResolvedTs {
...
-	return n.closeTableSink(ctx)
+	defer n.state.Store(tablepb.TableStateStopped)
+	if n.sinkV1 != nil {
+		err = n.sinkV1.Close(ctx)
+		if err != nil {
+			return
+		}
+		log.Info("sinkV1 is closed",
+			zap.Int64("tableID", n.tableID),
+			zap.String("namespace", n.changefeed.Namespace),
```
lonng, line 115: Can we implement the `fmt.Stringer` interface for the `Changefeed` to avoid we will add `namespace` and `changefeed-name` everywhere.

For example:
```go
func (c *Changefeed) String() string {
    return fmt.Srpintf("%s:%s", c.Namesapce, c.ID)
}


log.Info("something", zap.Stringer("changefeed", n.changefeed))
```

overvenus: Yes, we can. Let's do it in another PR.

PR #7123, opened 2022-09-20
metastore(engine): add retry when initializing sql client

amyangfei:

Issue Number: close #xxx

debug only

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


amyangfei:
Possible related to https://bugs.mysql.com/bug.php?id=106560

PR #7130, opened 2022-09-20
jobmanager(engine): terminate job if it fails continuously more than several times

amyangfei:

Issue Number: ref #6749

- Mark job as failed if it fails continuously more than `max-try-time`
- `max-try-time` can be configured from config file

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


amyangfei:
PTAL @maxshuang @CharlesCheung96

lance6716 (approved):

amyangfei:
PTAL @CharlesCheung96

deployments/engine/docker-compose/config/master.toml
```diff
@@ -17,3 +17,4 @@ level = "debug"
 reset-interval = "2s"
 initial-interval = "1s"
 max-interval = "15s"
+max-try-time = 100
```
CharlesCheung96, line 20: How long the total retry time is？Can we reduce `max-try-time` appropriately?

amyangfei: This is used in test to prevent job from termination. In product environment we use a smaller value `8`

CharlesCheung96 (approved):

PR #7155, opened 2022-09-21
*(engine): add comments to interface

lance6716:

Issue Number: ref #4287

simply add comment to MasterImpl and WorkerImpl, to remind the developer about thier behaviours

#### Tests 

 - Pass old tests

### Release note 

```release-note
`None`.
```


engine/executor/dm/worker.go
```diff
@@ -164,33 +164,29 @@ func (w *dmWorker) Workload() model.RescUnit {
...
-
 // OnMasterMessage implements lib.WorkerImpl.OnMasterMessage
 func (w *dmWorker) OnMasterMessage(ctx context.Context, topic p2p.Topic, message p2p.MessageValue) error {
 	w.Logger().Info("dmworker.OnMasterMessage", zap.String("topic", topic), zap.Any("message", message))
 	return nil
 }
 
 // CloseImpl implements lib.WorkerImpl.CloseImpl
-func (w *dmWorker) CloseImpl(ctx context.Context) error {
+func (w *dmWorker) CloseImpl(ctx context.Context) {
```
lonng, line 174: Can we change the method name?

engine/executor/dm/worker.go
```diff
@@ -164,33 +164,29 @@ func (w *dmWorker) Workload() model.RescUnit {
...
 	w.Logger().Info("close the dm worker", zap.String("task-id", w.taskID))
-	var recordErr error
 	// unregister jobmaster client
-	if err := w.messageAgent.UpdateClient(w.masterID, nil); err != nil {
-		w.Logger().Error("failed to update message client", zap.Error(err))
-		recordErr = err
+	if w.messageAgent != nil {
```
lonng, line 177: Why not early return if `w.messageAgent == nil` to avoid nil check everywhere? See: L185.

engine/executor/dm/worker.go
lonng: I submitted a PR to fix it: https://github.com/pingcap/tiflow/pull/7158

engine/executor/dm/worker.go
lance6716: the old logic is 
1. delete message client to jobmaster
2. stop DM units
3. close message agent
 
Can we change the order of them? @GMHDBJD

GMHDBJD: yes, you can change the order

engine/framework/master.go
```diff
@@ -67,37 +67,96 @@ type Master interface {
...
+	Tick(ctx context.Context) error
+
+	// OnWorkerDispatched is called when the asynchronized action of CreateWorker
+	// is finished. Only after OnWorkerDispatched, OnWorkerOnline and OnWorkerStatusUpdated
+	// of the same worker may be called.
+	// Return:
+	// - error to let the framework call CloseImpl.
+	// Concurrent safety:
+	// - this function may be concurrently called with another worker's OnWorkerXXX,
+	//   Tick, CloseImpl, StopImpl, OnCancel.
```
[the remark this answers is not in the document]
lance6716: OK there's no limitation to forbid the concurrency with InitImpl, OnMasterRecovered, and I want to learn more about detailed behaviour which is related to business implementation.

- If business created worker `w1` in InitImpl and InitImpl returned error or framework failed to flush internal status. After master failover, second InitImpl will be concurrent with OnWorkerDispatched of `w1`. Is this all the case?
- like above case, if framework succeed to flush InitImpl and master failover, second OnMasterRecovered will be concurrent with OnWorkerDispatched of `w1`. Is this all the case?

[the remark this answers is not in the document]
lance6716: cc @GMHDBJD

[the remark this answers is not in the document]
lance6716: ptal @amyangfei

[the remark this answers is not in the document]
amyangfei: > If business created worker w1 in InitImpl and InitImpl returned error or framework failed to flush internal status. After master failover, second InitImpl will be concurrent with OnWorkerDispatched of w1. Is this all the case?

no

> like above case, if framework succeed to flush InitImpl and master failover, OnMasterRecovered will be concurrent with OnWorkerDispatched of w1. Is this all the case?

no

For the same job master, its `InitImpl` or `OnMasterRecovered` is called first, and only after `InitImpl` or `OnMasterRecovered` is called, `OnWorkerXXX` can be triggered.

amyangfei (approved):

GMHDBJD (approved):
LGTM

PR #7203, opened 2022-09-23
puller(ticdc): remove span hack func call from frontier Forward

### LINKED ISSUE #7202: remove span hack func call from frontier Forward
remove span hack func call from frontier Forward

sdojjy:

Issue Number: close #7202

the End key return by the PD API will be nil to represent the biggest key,  now we will try to Hanck the EndKey every time when forwarding resolvedTs,  but it's better to Hack when span is created.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


cdc/puller/frontier/frontier.go
```diff
@@ -68,7 +68,6 @@ func (s *spanFrontier) Frontier() uint64 {
 
 // Forward advances the timestamp for a span.
 func (s *spanFrontier) Forward(span regionspan.ComparableSpan, ts uint64) {
-	span = span.Hack()
```
asddongmen, line 71: Is it possible to remove `span.Hack()` at https://github.com/pingcap/tiflow/pull/7203/files#diff-01b25d9f5901400302bd5f2b44a7f5e754b0037fd5fcec92d2a7acf066d25acaL39 too? It is little bit confused.

asddongmen (approved):

3AceShowHand (approved):

Issue #7114, opened 2022-09-19
Engine supports using s3 to store file resources


liuzix:
- [ ] Implement `FileManager` for s3. https://github.com/pingcap/tiflow/pull/7034
- [ ] Refactor existing code and integration the `FileManager` into existing code #7127
- [ ] Integration and system tests.

CharlesCheung96:
TODO:

- [ ] Figure out the root cause of performance rollback in [integration tests](https://ci2.pingcap.net/blue/organizations/jenkins/engine_ghpr_integration_test/detail/engine_ghpr_integration_test/1240/pipeline/)
- [ ] Implements remove resource interface for framework
- [ ] Always create a dump task with resource constraints

PR #7204, opened 2022-09-23
externalresources(engine): support s3 as storage (Part 2) 

CharlesCheung96:

Issue Number: ref #7114

- broker
    1.  Create `localFileManager` and `s3FileManager` in [`NewBroker`](https://github.com/pingcap/tiflow/pull/7204/files#diff-442c2620783282472c7ba6bc483b3b82bb860904b5f6b8bb3d61a97a3f892188R85)
    2. If S3 is configured, a dummy resource will be [persisted by broker](https://github.com/pingcap/tiflow/pull/7204/files#diff-442c2620783282472c7ba6bc483b3b82bb860904b5f6b8bb3d61a97a3f892188R303) to indicate that its temporary files have not been cleaned, which is useful to prevent resource leaks. 
    3. Asynchronously remove temporary files created by the worker in [`OnWorkerClosed`](https://github.com/pingcap/tiflow/pull/7204/files#diff-442c2620783282472c7ba6bc483b3b82bb860904b5f6b8bb3d61a97a3f892188R170)
    4. If an Executor exits normally, the corresponding broker will [attempt to clean up all temporary files](https://github.com/pingcap/tiflow/pull/7204/files#diff-442c2620783282472c7ba6bc483b3b82bb860904b5f6b8bb3d61a97a3f892188R332) before exiting. If this step fails, the dummy record is stored in Meta, which will be [cleaned up by the GCCoordinator](https://github.com/pingcap/tiflow/pull/7204/files#diff-7e9cb8664c8a29d3d577a870059edd4b8e826a6244687c3eda5988d112e4a147R216).

- manager
    1. DefaultGCCoordinator: [SetGCPending](https://github.com/pingcap/tiflow/pull/7204/files#diff-7e9cb8664c8a29d3d577a870059edd4b8e826a6244687c3eda5988d112e4a147R205) when job offline. [GCExecutors](https://github.com/pingcap/tiflow/pull/7204/files#diff-7e9cb8664c8a29d3d577a870059edd4b8e826a6244687c3eda5988d112e4a147R221) when executor offline.
    2. DefaultGCRunner: implement [GCExecutors](https://github.com/pingcap/tiflow/pull/7204/files#diff-3707465866a91b7dcb1b1b26212a10f25e1b13f91f0fefa5b6ffd266e4c5851bR191)

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


CharlesCheung96:
Broker integration tests verified in https://ci2.pingcap.net/blue/organizations/jenkins/engine_ghpr_integration_test/detail/engine_ghpr_integration_test/902/pipeline/86

engine/pkg/externalresource/internal/s3/resource_controller.go
```diff
@@ -0,0 +1,89 @@
...
+// GCExecutor cleanups all temporary resources created by the executor.
+func (r *resourceController) GCExecutor(
+	ctx context.Context, resources []*resModel.ResourceMeta, executorID model.ExecutorID,
+) error {
+	persistedResSet := make(map[string]struct{})
+	for _, res := range resources {
+		if res.Executor != executorID {
+			log.Panic("unexpected resource which is persisted by other exeuctor",
+				zap.Any("resource", res), zap.Any("expectedeExecutor", executorID))
+		}
```
amyangfei, line 71: Is it necessary to check executor id here?

engine/pkg/externalresource/broker/broker.go
```diff
@@ -247,7 +299,74 @@ func (b *DefaultBroker) getPersistResource(
 	return fm.GetPersistedResource(ctx, ident)
 }
 
-// PreCheckConfig does a preflight check on the executor's storage configurations.
+func (b *DefaultBroker) createDummyS3Resource() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	// TODO: add retry
```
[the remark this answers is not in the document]
CharlesCheung96: We need to add a retry mechanism for `createDummyS3Resource`.

In addition, there are various network operations in the `externalresource` package, such as Meta, RPC and S3. Therefore, we need to design a unified retry policy to improve its stability

engine/pkg/externalresource/internal/s3/file_manager.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+	if !ok {
+		log.Warn("resource is already persisted",
+			zap.Any("ident", ident))
+	}
+	return nil
+}
+
+func (m *FileManager) validateExecutor(creator model.ExecutorID, res interface{}) {
+	// Defensive verification to ensure that local resources are not accessible across nodes.
```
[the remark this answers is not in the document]
CharlesCheung96: We added a comment where the method was called.

engine/pkg/externalresource/internal/s3/resource_controller.go
CharlesCheung96: This is just a defensive check. And it's not necessary under normal circumstances

engine/pkg/externalresource/internal/s3/file_manager.go
```diff
@@ -0,0 +1,347 @@
...
+			zap.String("currentExecutor", string(m.executorID)))
+	}
+}
+
+func (m *FileManager) removeFilesIf(
+	ctx context.Context,
+	scope internal.ResourceScope,
+	pred func(path string) bool,
+) error {
+	storage, err := m.storageFactory.newS3ExternalStorageForScope(ctx, scope)
```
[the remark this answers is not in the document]
CharlesCheung96: We have added a comment explaining that we will add this feature in the future.

engine/pkg/externalresource/broker/broker.go
```diff
@@ -247,7 +306,72 @@ func (b *DefaultBroker) getPersistResource(
...
+	b.s3dummyHandler = handler
+	return nil
+}
+
+// Close cleans up the broker.
+func (b *DefaultBroker) Close() {
+	b.cancel()
+
+	if fm, ok := b.fileManagers[resModel.ResourceTypeS3]; ok {
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Hour)
```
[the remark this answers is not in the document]
CharlesCheung96: This is a mistake. Already fix it, PTAL.

engine/servermaster/config.go
```diff
@@ -171,7 +174,11 @@ func GetDefaultMasterConfig() *Config {
 		KeepAliveTTLStr:      defaultKeepAliveTTL,
 		KeepAliveIntervalStr: defaultKeepAliveInterval,
 		RPCTimeoutStr:        defaultRPCTimeout,
-		JobBackoff:           jobop.NewDefaultBackoffConfig(),
+		Storage: resModel.Config{
+			Local: resModel.LocalFileConfig{BaseDir: ""},
```
gozssky, line 178: Do all executors share the same local directory?

engine/pkg/orm/client.go
```diff
@@ -587,35 +589,34 @@ func (c *metaOpsClient) QueryResourcesByJobID(ctx context.Context, jobID string)
...
-	if result.Error == nil {
-		return &ormResult{rowsAffected: result.RowsAffected}, nil
+// DeleteResourcesByTypeAndExecutorIDs delete a specific type of resources of executorID
+func (c *metaOpsClient) DeleteResourcesByTypeAndExecutorIDs(
+	ctx context.Context, resType resModel.ResourceType, executorIDs ...engineModel.ExecutorID,
+) (Result, error) {
+	var result *gorm.DB
+	if len(executorIDs) == 1 {
+		result = c.db.WithContext(ctx).
+			Where("executor_id = ? and id like ?", executorIDs[0], fmt.Sprintf("/%s%%", resType)).
```
gozssky, line 613: Prefer adding `fmt.Sprintf("/%s%%", resType)` format to `pkg/externalresource/model/model.go`.

engine/framework/mock_master_impl.go
```diff
@@ -101,7 +101,7 @@ func (m *MockMasterImpl) GetFrameMetaClient() pkgOrm.Client {
 }
 
 // Reset resets the mock data.
-func (m *MockMasterImpl) Reset() {
+func (m *MockMasterImpl) Reset() context.CancelFunc {
```
gozssky, line 104: It looks a little weird that a `Reset` function returns a `CancelFunc`.

engine/pkg/externalresource/broker/broker.go
```diff
@@ -138,21 +175,46 @@ func (b *DefaultBroker) createResource(
...
-			// problem, we do not take further actions.
-			log.Warn("Failed to remove temporary files for worker",
-				zap.String("worker-id", workerID),
-				zap.String("job-id", jobID),
-				zap.Error(err))
+}
+
+func (b *DefaultBroker) runGCClosedWorker(ctx context.Context) {
+	// We run a gc loop at the max frequency of once per second.
+	rl := ratelimit.New(1 /* once per second */)
```
gozssky, line 189: Why not use `ticker` directly? It seems more reasonable.

engine/pkg/externalresource/broker/broker.go
```diff
@@ -138,21 +175,46 @@ func (b *DefaultBroker) createResource(
...
 // OnWorkerClosed implements Broker.OnWorkerClosed
 func (b *DefaultBroker) OnWorkerClosed(ctx context.Context, workerID resModel.WorkerID, jobID resModel.JobID) {
-	scope := internal.ResourceScope{
-		Executor: b.executorID,
-		WorkerID: workerID,
+	timer := time.NewTimer(defaultTimeout)
+	select {
+	case b.closedWorkerCh <- closedWorker{workerID: workerID, jobID: jobID}:
+		return
+	case <-timer.C:
```
gozssky, line 182:
```suggestion
	case <-timer.After(defaultTimeout):
```

gozssky: Please also select for `ctx.Done`.

engine/pkg/externalresource/broker/broker.go
```diff
@@ -138,21 +175,46 @@ func (b *DefaultBroker) createResource(
...
+		case <-ctx.Done():
+			return
+		case w := <-b.closedWorkerCh:
+			scope := internal.ResourceScope{
+				Executor: b.executorID,
+				WorkerID: w.workerID,
+			}
+			for _, fm := range b.fileManagers {
+				err := fm.RemoveTemporaryFiles(ctx, scope)
+				if err != nil {
```
gozssky, line 202: If process crashes, are any temporary files leak?

engine/servermaster/config.go
CharlesCheung96: This is not necessary, but in practice it is better to configure it as the same local directory.

engine/pkg/externalresource/broker/broker.go
CharlesCheung96: There are no resource leaks, since all temporary resources created by the crashed executor would be cleaned by [gcRunner](https://github.com/pingcap/tiflow/pull/7204/files#diff-3707465866a91b7dcb1b1b26212a10f25e1b13f91f0fefa5b6ffd266e4c5851bR228).

engine/pkg/orm/client.go
CharlesCheung96: Done, PTAL

engine/pkg/externalresource/internal/s3/file_manager.go
```diff
@@ -0,0 +1,347 @@
...
+			// remove all files within this scope since no persisted files are found
+			return true
+		})
+	}
+
+	return m.removeFilesIf(ctx, scope, func(path string) bool {
+		resName, _, ok := strings.Cut(path, "/")
+		if !ok {
+			return false
+		}
```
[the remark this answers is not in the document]
CharlesCheung96: Fixed, PTAL

engine/pkg/externalresource/internal/s3/integration/main_test.go
```diff
@@ -0,0 +1,21 @@
...
+// limitations under the License.
+package integration
+
+// import (
+// 	"testing"
+// )
+
+// func TestMain(m *testing.M) {
+// 	// leakutil.SetUpLeakTest(m)
+// }
```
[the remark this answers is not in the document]
CharlesCheung96: Ref: https://github.com/pingcap/tiflow/pull/7204/files#diff-d73ed6b421de43084347e8eed178ffa9865f0e7b2ed25ca99ce9cd810c00d3dcR23

engine/servermaster/config.go
gozssky: There is only one storage config in servermaser. Does it mean all executors are forced to use the same config?

engine/pkg/externalresource/model/model.go
```diff
@@ -133,6 +137,12 @@ const (
 	ResourceTypeS3        = ResourceType("s3")
 )
 
+// BuildPrefix returns the prefix of the resource type.
+func (r ResourceType) BuildPrefix() string {
+	// For local file, the prefix is `/local%`. For S3, the prefix is `/s3%`.
+	return fmt.Sprintf("/%s%%", r)
```
gozssky, line 143: If I understand correctly, the suffix `%` is used in SQL queries. It's not a good idea to coupling with SQL here.

engine/pkg/externalresource/model/model.go
```diff
@@ -133,6 +137,12 @@ const (
 	ResourceTypeS3        = ResourceType("s3")
 )
 
+// BuildPrefix returns the prefix of the resource type.
+func (r ResourceType) BuildPrefix() string {
+	// For local file, the prefix is `/local%`. For S3, the prefix is `/s3%`.
```
gozssky, line 142:
```suggestion
	// For local file, the prefix is `/local`. For S3, the prefix is `/s3`.
```

engine/servermaster/config.go
CharlesCheung96: The local directory needs to be configured directly for each executor.

engine/pkg/externalresource/manager/gc_coordinator.go
```diff
@@ -175,68 +172,59 @@ func (c *DefaultGCCoordinator) gcByStatusSnapshots(
...
 			// The resource belongs to a deleted job.
-			toGC = append(toGC, resMeta.Job)
+			toGCJobs = append(toGCJobs, resMeta.Job)
 			continue
 		}
 
 		if _, exists := executorSet[resMeta.Executor]; !exists {
 			// The resource belongs to an offlined executor.
-			toRemove = append(toRemove, resMeta.Executor)
+			tp, resName, err := resModel.PasreResourceID(resMeta.ID)
```
gozssky, line 187: Since we query resources after getting the executor's snapshot, is it possible that a new executor creates a resource but the executor is not in the `executorSet`?

engine/pkg/externalresource/manager/gc_coordinator.go
```diff
@@ -175,68 +172,59 @@ func (c *DefaultGCCoordinator) gcByStatusSnapshots(
...
+			if err != nil {
+				return err
+			}
+			if tp == resModel.ResourceTypeLocalFile ||
+				tp == resModel.ResourceTypeS3 && resName == s3.DummyResourceName {
```
gozssky, line 192: I'm not clear about why we need a dummy resource for s3 resources. What if we remove the `resName == s3.DummyResourceName`?

engine/pkg/externalresource/manager/gc_coordinator.go
```diff
@@ -175,68 +172,59 @@ func (c *DefaultGCCoordinator) gcByStatusSnapshots(
...
+				toGCExecutors = append(toGCExecutors, resMeta.Executor)
```
gozssky, line 193: Do multiple resources reference the same executor?

engine/pkg/externalresource/manager/gc_coordinator.go
CharlesCheung96: Yes, it is possible, so we need to query meta before getting the snapshot.

engine/pkg/externalresource/manager/gc_coordinator.go
CharlesCheung96: If S3 is configured, a dummy resource will be persisted by broker to indicate that its temporary files have not been cleaned, which is useful to prevent resource leaks.

Normally a broker will attempt to clean up temporary files and dummy resources before exiting. If this step fails, the dummy record is stored in Meta, which will be cleaned up by GCCoordinator eventually.

Hence, we need the dummy resource's meta to trigger gc of the offlined executor.

gozssky: Do other s3 normal resources store in meta too?

engine/pkg/externalresource/manager/gc_coordinator.go
CharlesCheung96: fix it, PTAL!

engine/pkg/externalresource/manager/gc_coordinator.go
CharlesCheung96: Yes, only persistent resources are stored in meta.

engine/pkg/externalresource/internal/s3/file_manager.go
```diff
@@ -0,0 +1,320 @@
...
+)
+
+const (
+	// TODO: use the contents of placeholder to indicate persistent status
+	placeholderFileName = ".keep"
+)
+
+type persistedResources map[resModel.ResourceName]struct{}
+
+func newpersistedResources() persistedResources {
```
gozssky, line 38:
```suggestion
func newPersistedResources() persistedResources {
```

gozssky (approved):

engine/test/integration_tests/external_resource/run.sh
```diff
@@ -0,0 +1,33 @@
...
+	export ENGINE_S3_ACCESS_KEY=engine
+	export ENGINE_S3_SECRET_KEY=engineSecret
+	export ENGINE_S3_ENDPOINT="http://127.0.0.1:9000/"
+
+	CGO_ENABLED=0 go test -timeout 300s -cover -count 1 -v -run ^TestIntegrationS3 github.com/pingcap/tiflow/engine/pkg/externalresource/internal/s3
+
+	CGO_ENABLED=0 go test -timeout 600s -cover -count 1 -v -run ^TestIntegrationBroker github.com/pingcap/tiflow/engine/pkg/externalresource/broker
+
+	echo -e "\ncheck external storage integration tests result..."
+	if cat $OUT_DIR/engine_it.log | grep "SKIP:"; then
```
amyangfei, line 25: Which log contains `SKIP:`

CharlesCheung96: [s3 integration related tests](https://github.com/pingcap/tiflow/pull/7204/files#diff-c6b5d7e31f3c3675f70a4bd422d792771fa9443ed16b065d2f3d4192d0aa4a97R75)

amyangfei (approved):

PR #7208, opened 2022-09-23
util(dm): auto split the transaction when it's too large

### LINKED ISSUE #5010: DM may meet `Error 8004: Transaction is too large, size: xxx` when update checkpoint
### What did you do?

https://asktug.com/t/topic/603512

seems when there're many tables and big TableInfo, DM will meet this problem. And don't know why v5.4 is more error-prone then earlier versions.

This problem may auto-resume but in some very bad cases I guess auto-resume will not help. In that case, user can split task to many smaller tasks that contain part of tables to be replicated.

_No response_

### What did you see instead?

_No response_

### Versions of the cluster

v5.4.0

```console
```

lance6716:

Issue Number: ref #5010

#### Tests 

 - Unit test

### Release note 

```release-note
`None`.
```


lance6716:
added a simplt util function
/cc @buchuitoudegou @D3Hunter @GMHDBJD

buchuitoudegou (approved):

lance6716:
ptal @D3Hunter @GMHDBJD

dm/pkg/conn/baseconn.go
```diff
@@ -223,6 +225,34 @@ func (conn *BaseConn) ExecuteSQL(tctx *tcontext.Context, hVec *prometheus.Histog
...
+	if !ok {
+		return err
+	}
+
+	if mysqlErr.Number != errno.ErrTxnTooLarge || len(queries) == 1 {
+		return err
+	}
+
+	mid := len(queries) / 2
+	err = conn.ExecuteSQLsAutoSplit(tctx, hVec, task, queries[:mid], args[:mid]...)
```
D3Hunter, line 249: this pr is part 1? seems no one calls `ExecuteSQLsAutoSplit`.

if checkpoint is half saved, will it cause `global pos`(save success) > `some table pos` which failed to save

D3Hunter (approved):
lgtm

dm/pkg/conn/baseconn.go
lance6716: yes, I'll let syncer checkpoint uses `ExecuteSQLsAutoSplit` in another PR.

And DML of global checkpoint should be the last DML, so if checkpoint is half saved only table checkpoint is safely forwarded.

Issue #7201, opened 2022-09-23
Unstable test pdutil.TestTimeFromPD


gozssky:
### What did you do?

```
go test -count=1 -v -run TestTimeFromPD  github.com/pingcap/tiflow/pkg/pdutil
```
Ref https://ci.pingcap.net/blue/organizations/jenkins/ut-check/detail/ut-check/36222/pipeline/.

### What did you expect to see?

Test pass.

### What did you see instead?

Test stuck.
```
=== RUN   TestTimeFromPD
=== PAUSE TestTimeFromPD
=== CONT  TestTimeFromPD
[2022/09/23 11:13:38.120 +08:00] [WARN] [clock.go:99] ["get time from pd failed, will use local time as pd time"]
```

### Versions of the cluster
```
master
```

PR #7219, opened 2022-09-26
pkg/pdutil(ticdc): fix a deadlock issue

hi-rustin:

Issue Number: close    https://github.com/pingcap/tiflow/issues/7201

Fix a deadlock issue. Release the lock as soon as possible, otherwise `Run` and `Close` will cause a deadlock.

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```

### DIFF
--- a/pkg/pdutil/clock.go
+++ b/pkg/pdutil/clock.go
@@ -43,7 +43,7 @@ type clock struct {
 		sync.RWMutex
 		// The time encoded in PD ts.
 		tsEventTime time.Time
-		// The time we receives PD ts.
+		// The time we receive PD ts.
 		tsProcessingTime time.Time
 		err              error
 	}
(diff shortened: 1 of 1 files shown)


gozssky (approved):

lance6716 (approved):

pkg/pdutil/clock.go
```diff
@@ -120,8 +120,8 @@ func (c *clock) CurrentTime() (time.Time, error) {
 // Stop clock.
 func (c *clock) Stop() {
 	c.mu.Lock()
-	defer c.mu.Unlock()
 	c.cancel()
+	c.mu.Unlock()
```
amyangfei, line 124: Do we need to add a unit test

hi-rustin: Actually, this will fix that unstable test. So I don't think we need to add a new test.

PR #7222, opened 2022-09-26
etcd (ticdc): fix cdc cluster has two owner at the same time

### LINKED ISSUE #4051: A scenario that TiCDC could have more than one owner
### What did you do?

1. Create two ticdc servers
2. Delete the owner key of the ticdc owner in etcd. `/tidb/cdc/owner/<lease-id>`
3. Observe another ticdc server will become owner, and the old owner is still running.

The old owner could detect this behavior.

### What did you see instead?

Two owner exist.

**NOTE: this is an unexpected user operation, if we don't think it is a bug, we must forbid it.**

### Versions of the cluster

```console
```

master@aa6f3c6e5ccf1dd5578de35a4a946712e9ef4ef9

asddongmen:

Issue Number: close #4051

Add a checking logic inside `EtcdWorker`, which will check if the owner key of current `capture` and `Etcdworker` was deleted by etcd watch mechanism, if we found the owner key has gone, return the `EtcdWorker` loop and the `capture` will **restart**.
#### Tests 

 - Integration test
The reproduce steps was articulated in #4051

### Release note 

```release-note
Fix a bug that can cause one TiCDC cluster has two owner at the same time. (when we delete the owner key by etcdctl directly).
```


nongfushanquan (approved):

pkg/orchestrator/etcd_worker.go
```diff
@@ -129,7 +133,13 @@ func (worker *EtcdWorker) initMetrics() {
 // A tick is generated either on a timer whose interval is timerInterval, or on an Etcd event.
 // If the specified etcd session is Done, this Run function will exit with cerrors.ErrEtcdSessionDone.
 // And the specified etcd session is nil-safety.
-func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session, timerInterval time.Duration, role string) error {
+func (worker *EtcdWorker) Run(
+	ctx context.Context,
+	session *concurrency.Session,
+	timerInterval time.Duration,
+	role string,
```
3AceShowHand, line 140: how about `role util.Role`

pkg/orchestrator/etcd_worker.go
```diff
@@ -261,6 +276,29 @@ func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session,
 			if !rl.Allow() {
 				continue
 			}
+
+			// we need to check if this the owner key is delete before we tick the reactor
+			if role == pkgutil.RoleOwner.String() {
+				var id model.CaptureID
+
+				err = retry.Do(ctx, func() error {
+					id, err = worker.client.GetOwnerID(ctx)
```
3AceShowHand, line 285: Is `GetOwnerID` already retried internally? If yes, it looks that no need to retry here ?

3AceShowHand (approved):

pkg/orchestrator/etcd_worker.go
asddongmen: good

tests/integration_tests/region_merge/run.sh
```diff
@@ -45,8 +46,6 @@ function run() {
 		run_kafka_consumer $WORK_DIR "kafka://127.0.0.1:9092/$TOPIC_NAME?protocol=open-protocol&partition-num=4&version=${KAFKA_VERSION}&max-message-bytes=10485760"
 	fi
 
-	# set max_execution_time to 30s, because split region could block even region has been split.
-	run_sql "SET @@global.MAX_EXECUTION_TIME = 30000;" ${UP_TIDB_HOST} ${UP_TIDB_PORT}
```
hicqu, line 49: Why remove these lines?

asddongmen: `split region` DDL would timeout if we set it to `30s`.

asddongmen:
The `merge_region` integration test always fails, can't find the reason.

Issue #7068, opened 2022-09-14
Auto resume is not aware of lightning error


lance6716:
### What did you do?

```
[...]

PR #7242, opened 2022-09-28
lightning(dm): all lightning errors are not auto-resumeable

lance6716:

Issue Number: ref #7068

lightning itself is a long time runnning import job, so it already has handled retry. It's no need to let DM add retry mechanism.

#### Tests 

 - Unit test

### Release note 

```release-note
`None`.
```


buchuitoudegou (approved):
lgtm

PR #7251, opened 2022-09-28
pipeline(ticdc): Add rate limit for some repeated logs

### LINKED ISSUE #7252: Add log rate limit for table actor

sdojjy:

Issue Number: close #7252

Add log rate limit for table actor handle barrierTis, region worker resolve lock 

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
 `None`.
```


pkg/txnutil/lock_resolver.go
```diff
@@ -131,7 +131,7 @@ func (r *resolver) Resolve(ctx context.Context, regionID uint64, maxVersion uint
 		}
 		bo = tikv.NewGcResolveLockMaxBackoffer(ctx)
 	}
-	log.Info("resolve lock successfully",
+	log.Debug("resolve lock successfully",
```
3AceShowHand, line 134: I think this would be helpful to check if the region is locked, so I would prefer to keep it at the `Info` level.

sdojjy, line 134:
```suggestion
	log.Info("resolve lock successfully",
```

sdojjy: done

3AceShowHand (approved):

cdc/kv/region_worker.go
```diff
@@ -298,6 +299,9 @@ type rtsUpdateEvent struct {
 	resolvedTs uint64
 }
 
+// too many repeated logs when TiDB OOM
+var resolveLockLogRateLimiter = rate.NewLimiter(rate.Every(time.Millisecond*500), 10)
```
sdojjy, line 303:
```suggestion
var resolveLockLogRateLimiter = rate.NewLimiter(rate.Every(time.Millisecond*2000), 10)
```

asddongmen (approved):

PR #7268, opened 2022-09-29
sqlmodel(dm): support generating UPDATE CASE WHEN

lance6716:

Issue Number: close https://github.com/pingcap/tiflow/issues/3408

as title

#### Tests 

 - Unit test

### Release note 

```release-note
`None`.
```


lance6716:
ptal @buchuitoudegou

dm/syncer/dml.go
```diff
@@ -355,6 +355,8 @@ func genSQLMultipleRows(op sqlmodel.DMLType, dmls []*sqlmodel.RowChange) (querie
 	switch op {
 	case sqlmodel.DMLInsert, sqlmodel.DMLReplace, sqlmodel.DMLInsertOnDuplicateUpdate:
 		return sqlmodel.GenInsertSQL(op, dmls...)
+	case sqlmodel.DMLUpdate:
+		return sqlmodel.GenUpdateSQL(dmls...)
```
buchuitoudegou, line 359: > // Compared to GenInsertSQL with DMLInsertOnDuplicateUpdate, this function is
// slower and more complex, we should only use it when PK/UK is updated.

I noticed this comment in `GenUpdateSQL`, which indicates we should only use it when updating the primary keys, but it seems we didn't consider it in this PR. Would there be any problems?

lance6716: https://github.com/pingcap/tiflow/blob/5140371b5990c91845f79d16f0b2988785083586/dm/syncer/dml.go#L464-L471

this is already implemented

buchuitoudegou (approved):

lance6716:
ptal @GMHDBJD @dsdashun

lance6716:
ptal @GMHDBJD @dsdashun

dm/syncer/dml.go
```diff
@@ -419,9 +420,24 @@ func genDMLsWithSameTable(op sqlmodel.DMLType, jobs []*job) ([]string, [][]inter
...
+			}
+			if lastTable != j.dml.TargetTableID() {
+				query, arg := genDMLsWithSameCols(op, groupDMLs)
+				queries = append(queries, query...)
+				args = append(args, arg...)
+
+				groupDMLs = groupDMLs[0:0]
+				lastTable = j.dml.TargetTableID()
+			}
+			groupDMLs = append(groupDMLs, j.dml)
```
GMHDBJD, line 435: Can we combine these lines with L447?

GMHDBJD (approved):
LGTM

GMHDBJD:
Have not measured the syntax, not sure if the performance has improved and there are no other side effects

dm/syncer/dml.go
lance6716: Will refine the code in future. In this sprint plan to let CDC use this library with @asddongmen

PR #7279, opened 2022-09-30
redo (ticdc): only delete changefeed related file when changefeed was removed

### LINKED ISSUE #6413: redo manager should only clean files of one specified changefeed instead of all changefeeds

- [X] Googled your question

### What is your question?

Test shows there is a bug about cleaning redo-log files when removing a changefeed:

* when removing a changefeed, `managerImpl.Cleanup` will be called;
* then `LogWriter.DeleteAllLogs` will be called, and all files in the redo-log directory and S3 path will be deleted;
* so if **several changefeeds share one redo-log directory or S3 bucket**, their redo files will be deleted incorrectly.

The bug should effect all versions.

asddongmen:

Issue Number: close #6413 

Only delete the file belong to a changefeed when it's release, rather than the whole directory that used by it.

#### Tests 

 - Unit test
1. create changefeed test-1, and other two changefeed. Enable redo log for all of them, using the same storage uri:
```
s3://guagua/test1?access-key=xxxx&secret-access-key=xxxx&endpoint=http://xxxx/
```
2. Replicate some data, and remove one changefeed.
3. Inspect the `s3://guagua/test1` dir, the files of removed changefeed has been deleted but other two changefeeds' files remain.
image

### Release note 

```release-note
Fix a bug that may cause changefeed's redo log files be deleted wrongly.
```


cdc/redo/common/util.go
```diff
@@ -160,3 +162,23 @@ func defaultS3Retryer() request.Retryer {
 		},
 	}
 }
+
+// GetChangefeedFiles return the files that belong to the changefeed.
+func GetChangefeedFiles(files []string, changefeedID model.ChangeFeedID) []string {
```
nongfushanquan, line 167: Maybe the name such as FilterChangefeedFiles  is more proper for this method

nongfushanquan (commented):
please correct the title "Redo"

cdc/redo/writer/file.go
```diff
@@ -446,6 +446,7 @@ func (w *Writer) newPageWriter() error {
 	return nil
 }
 
+// why we call it rotate?
```
zhaoxinyu, line 449: Because this function will close the current file and open a new file.

cdc/redo/writer/writer.go
```diff
@@ -51,7 +51,7 @@ type RedoLogWriter interface {
 	// Regressions on them will be ignored.
 	FlushLog(ctx context.Context, checkpointTs, resolvedTs model.Ts) error
 
-	// Get the current meta.
+	// GetMeta get current meta.
```
zhaoxinyu, line 54:
```suggestion
	// GetMeta gets current meta.
```

cdc/redo/writer/writer.go
```diff
@@ -93,7 +93,9 @@ type LogWriter struct {
 	cfg       *LogWriterConfig
 	rowWriter fileWriter
 	ddlWriter fileWriter
-	storage   storage.ExternalStorage
+	// storage in LogWriter is used to clean up the
```
zhaoxinyu, line 96: Maybe this comment is not accurate. Because use also use `storage` to save metadata. See function `writeMetaToS3()`

zhaoxinyu:
Rest LGTM

pkg/uuid/uuid.go
```diff
@@ -15,7 +15,7 @@ package uuid
 
 import guuid "github.com/google/uuid"
 
-// Generator defines an interface that can generate a uuid
+// Generator defines an interface that can generate an uuid
```
zhaoxinyu, line 18:
```suggestion
// Generator defines an interface that can generate a uuid
```

zhaoxinyu: The original code comment is correct.

zhaoxinyu (approved):

cdc/redo/common/util.go
```diff
@@ -21,6 +21,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pingcap/tiflow/cdc/model"
+
```
hicqu, line 25: How about removing the empty line?

cdc/redo/writer/writer.go
```diff
@@ -340,27 +342,50 @@ func (l *LogWriter) DeleteAllLogs(ctx context.Context) (err error) {
...
+	}
+	filteredFiles := common.FilterChangefeedFiles(fileNames, l.cfg.ChangeFeedID)
+
+	if len(filteredFiles) == len(fileNames) {
+		if err = os.RemoveAll(l.cfg.Dir); err != nil {
+			return cerror.WrapError(cerror.ErrRedoFileOp, err)
+		}
+	} else {
+		for _, file := range filteredFiles {
+			if err = os.RemoveAll(filepath.Join(l.cfg.Dir, file)); err != nil {
```
hicqu, line 367: If the error is `NotExists`, it should be ignored.

cdc/redo/common/util.go
asddongmen: good catch

nongfushanquan (approved):

hicqu (approved):

PR #7280, opened 2022-09-30
frontier(ticdc):  fast check if region is split or merged when forwarding resolvedTs

### LINKED ISSUE #7281: fast check if region is split or merged when forwarding resolvedTs
fast check if region is split or merged when forwarding resolvedTs

sdojjy:

Issue Number: close #7281

using map cached all region spans, when forwarding a span, check if it is split or merged by comparing the span start, span end, and region ID.

server  CPU and memory:
image

profile before:
image

profile after
image

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
 `None`.
```


cdc/puller/frontier/frontier.go
```diff
@@ -33,15 +34,27 @@ type Frontier interface {
...
+	metricResolvedRegionCachedCounter prometheus.Counter
 }
 
 // NewFrontier creates Frontier from the given spans.
 // spanFrontier don't support use Nil as the maximum key of End range
 // So we use set it as util.UpperBoundKey, the means the real use case *should not* have an
 // End key bigger than util.UpperBoundKey
-func NewFrontier(checkpointTs uint64, spans ...regionspan.ComparableSpan) Frontier {
+func NewFrontier(checkpointTs uint64,
+	metricResolvedRegionCachedCounter prometheus.Counter,
```
3AceShowHand, line 49: we have to pay attention to this metrics, since it's on the hot path.

cdc/puller/frontier/frontier.go
```diff
@@ -63,33 +76,50 @@ func (s *spanFrontier) Frontier() uint64 {
...
 // Forward advances the timestamp for a span.
-func (s *spanFrontier) Forward(span regionspan.ComparableSpan, ts uint64) {
-	s.insert(span, ts)
+func (s *spanFrontier) Forward(regionID uint64, span regionspan.ComparableSpan, ts uint64) {
+	// it's the fast part to detect if the region is split or merged,
+	// if not we can update the minTsHeap with use new ts directly
+	if n, ok := s.cachedRegions[regionID]; ok && n.regionID > 0 && n.end != nil {
+		if regionID == n.regionID && bytes.Equal(n.Key(), span.Start) && bytes.Equal(n.End(), span.End) {
+			s.minTsHeap.UpdateKey(n.Value(), ts)
+			s.metricResolvedRegionCachedCounter.Inc()
```
3AceShowHand, line 85: If we make the assumption that `region merge or split is relatively rare`, L82 ~ L85 should be hit frequently, which makes L85 always be called.

Can we just record `cache miss` case to reduce the overhead caused by updating the metrics?

3AceShowHand (approved):

cdc/puller/frontier/frontier.go
```diff
@@ -63,33 +76,50 @@ func (s *spanFrontier) Frontier() uint64 {
...
+	seekRes := s.spanList.Seek(span.Start, s.seekTempResult)
 	// if there is no change in the region span
 	// We just need to update the ts corresponding to the span in list
 	next := seekRes.Node().Next()
 	if next != nil {
-		cmpStart := bytes.Compare(seekRes.Node().Key(), span.Start)
-		cmpEnd := bytes.Compare(next.Key(), span.End)
-		if cmpStart == 0 && cmpEnd == 0 {
+		if bytes.Equal(seekRes.Node().Key(), span.Start) && bytes.Equal(next.Key(), span.End) {
 			s.minTsHeap.UpdateKey(seekRes.Node().Value(), ts)
```
nongfushanquan, line 103: Dose this line implement the same logic with the line 82~86 ?

cdc/puller/frontier/frontier.go
```diff
@@ -63,33 +76,50 @@ func (s *spanFrontier) Frontier() uint64 {
 }
 
 // Forward advances the timestamp for a span.
-func (s *spanFrontier) Forward(span regionspan.ComparableSpan, ts uint64) {
-	s.insert(span, ts)
+func (s *spanFrontier) Forward(regionID uint64, span regionspan.ComparableSpan, ts uint64) {
+	// it's the fast part to detect if the region is split or merged,
+	// if not we can update the minTsHeap with use new ts directly
+	if n, ok := s.cachedRegions[regionID]; ok && n.regionID > 0 && n.end != nil {
```
asddongmen, line 82: What is the meaning is regionID == 0? Could you please use a `commented const` instead of use `0` for the special case?

cdc/puller/frontier/frontier.go
```diff
@@ -63,33 +76,50 @@ func (s *spanFrontier) Frontier() uint64 {
...
+		if regionID == n.regionID && bytes.Equal(n.Key(), span.Start) && bytes.Equal(n.End(), span.End) {
```
asddongmen, line 83: When is `regionID != n.regionID` ? Seems that they would always be equal.

cdc/puller/frontier/frontier.go
sdojjy: done

cdc/puller/frontier/frontier.go
sdojjy: yes, if region is not merged or split, just update the resolved ts

cdc/puller/frontier/frontier.go
sdojjy: done

cdc/puller/frontier/frontier.go
sdojjy: done

asddongmen (approved):

PR #7288, opened 2022-10-08
kv(ticdc): use customized region state map to maintain region states

### LINKED ISSUE #7307: refactor region state manager use rw lock to reduce lock contention

3AceShowHand:

Issue Number: close #7307 

* for `singleRegionInfo` struct, rename `ts` to `resolvedts`
* add a new file `region_state` to maintain some related code
* use `syncRegionStateMap` to replace `sync.Map` in the `regionStateManager`

region state manager in the kv client component, use `sync.Map` to maintain regions state, which use `sync.Mutext` for mutual access.

We can replace `sync.Map` by `syncRegionStateMap` to use `sync.RWLock`, which should be helpful to reduce lock contention.

After:
`getRegionState` cost around "0.63"
image

Before:

`getRegionState` cost around `2.32`

image

Benchmark

`sync.Map`
Y3dhneOaiB

customized `syncRegionStateMap`
Screen Shot 2022-10-11 at 16 10 07

As we can see from the benchmark result, our customized `regionStateMap` struct access `getState` is twice fast than the `sync.Map`.

#### Tests 

### Release note 

```release-note
`None`
```


cdc/kv/client.go
```diff
@@ -730,37 +630,28 @@ func (s *eventFeedSession) requestRegionToStore(
...
-			if !ok {
-				// Should never happen
-				log.Panic("pending regions is not found for store",
-					zap.String("namespace", s.changefeed.Namespace),
-					zap.String("changefeed", s.changefeed.ID),
-					zap.Int64("tableID", s.tableID), zap.String("tableName", s.tableName),
-					zap.String("store", rpcCtx.Addr))
-			}
-		} else {
+		storeAddr := rpcCtx.Addr
```
zhaoxinyu, line 633: Can `rpcCtx` or `rpcCtx.Peer` be nil?

cdc/kv/region_state.go
```diff
@@ -0,0 +1,281 @@
...
+// into several buckets to reduce lock contention
+type regionStateManager struct {
+	bucket int
+	states []*syncRegionFeedStateMap
+}
+
+func newRegionStateManager(bucket int) *regionStateManager {
+	if bucket <= 0 {
+		bucket = runtime.NumCPU()
+		if bucket > maxRegionStateBucket {
```
zhaoxinyu, line 246: What if we just use the `NumCPU()` buckets?

cdc/kv/region_worker.go
```diff
@@ -200,11 +146,14 @@ func newRegionWorker(
 		concurrent:    s.client.config.WorkerConcurrent,
 		metrics:       metrics,
 		inputPending:  0,
-		inputCalcSlot: func(regionID uint64) int { return int(regionID) % s.client.config.WorkerConcurrent },
 		inputSlots:    s.client.config.WorkerConcurrent,
 	}
 }
 
+func (w *regionWorker) inputCalcSlot(regionID uint64) int {
+	return int(regionID) % w.concurrent
```
zhaoxinyu, line 154: Maybe `w.concurrency` is more appropriate.

zhaoxinyu (approved):

zhaoxinyu:
Rest LGRM

cdc/kv/region_state.go
```diff
@@ -0,0 +1,281 @@
...
+		rpcCtx:       rpcCtx,
+	}
+}
+
+type regionFeedState struct {
+	sri       singleRegionInfo
+	requestID uint64
+	stopped   int32
+
+	lock           sync.RWMutex
```
zhaoxinyu, line 57: Maybe we can use atomic values to avoid using the lock if the lock contention is really that severe.

cdc/kv/region_state.go
```diff
@@ -0,0 +1,283 @@
...
+	states []*syncRegionFeedStateMap
+}
+
+func newRegionStateManager(bucket int) *regionStateManager {
+	if bucket <= 0 {
+		bucket = runtime.NumCPU()
+		if bucket > maxRegionStateBucket {
+			bucket = maxRegionStateBucket
+		}
+		if bucket < minRegionStateBucket {
```
asddongmen, line 251: It seems that the `maxRegionStateBucket` and `minRegionStateBucket` if statement should be outside the `if bucket <= 0` if statement block, otherwise `bucket` argument can less than this two constraint constant.

cdc/kv/region_state.go
3AceShowHand: Atomic value access also has some cost, maybe it's less than locks.

Since there are so many fields need to be accessed exclusively, so maybe a `RWMutex` shared by all fields is ok.

overvenus:
Please add a benchmark about `getRegionState`.

cdc/kv/region_state.go
```diff
@@ -0,0 +1,289 @@
...
+
+const (
+	minRegionStateBucket = 4
+	maxRegionStateBucket = 16
+)
+
+type singleRegionInfo struct {
+	verID        tikv.RegionVerID
+	span         regionspan.ComparableSpan
+	checkpointTs uint64
```
sdojjy, line 34: it's no checkpointTs, resolvedTs is better?

cdc/kv/resolvedts_heap.go
```diff
@@ -80,9 +80,9 @@ func newRegionTsManager() *regionTsManager {
 	}
 }
 
-// Upsert implements insert	and update on duplicated key
```
sdojjy, line 83: should not update this comment

overvenus (approved):

Issue #7271, opened 2022-09-30
Upgrade golangci-lint and fix new lint errors


amyangfei:
### Is your feature request related to a problem?

Current version golangci-lint doesn't support generic features, and many linters don't work.

```go
cd tools/check && GO111MODULE=on go build -mod=mod -o ../bin/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
tools/bin/golangci-lint run --timeout 10m0s --skip-files kv_gen --skip-dirs dm,tests
WARN [linters context] bodyclose is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [linters context] gosimple is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [linters context] noctx is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [linters context] rowserrcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [linters context] sqlclosecheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [linters context] staticcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [linters context] structcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [linters context] stylecheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [linters context] tparallel is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [linters context] unparam is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [linters context] unused is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
WARN [runner/nolint] Found unknown linters in //nolint directives: will be used in next pr
```

### Describe the feature you'd like

[...]

PR #7293, opened 2022-10-08
lint(engine): fix all lint errors in engine module

amyangfei:

Issue Number: ref #7271

- Fix all errors under the engine module, with `golangci-lint v1.50.0`

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/engine/test/e2e/e2e_dm_test.go
+++ b/engine/test/e2e/e2e_dm_test.go
@@ -105,7 +105,7 @@ func TestDMJob(t *testing.T) {
 	cli, err := httputil.NewClient(nil)
 	require.NoError(t, err)
 	for _, metricsURL := range metricsURLs {
-		resp, err := cli.Get(ctx, metricsURL) //nolint:bodyclose
+		resp, err := cli.Get(ctx, metricsURL)
 		require.NoError(t, err)
 		content, err := io.ReadAll(resp.Body)
 		require.NoError(t, err)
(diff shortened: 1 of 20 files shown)


lance6716 (approved):

engine/test/e2e/e2e_dm_test.go
```diff
@@ -515,7 +517,7 @@ func getBinlogSchema(client *http.Client, jobID string, task string, schema stri
 		v.Add("table", table)
 	}
 	u += "?" + v.Encode()
-	resp, err := client.Get(u)
+	resp, err := client.Get(ctx, u) //nolint:bodyclose
```
gozssky, line 520: We can close `resp.Body` after error check.

pkg/httputil/httputil_test.go
```diff
@@ -70,25 +70,55 @@ func TestHttputilNewClient(t *testing.T) {
 	require.Equal(t, httputilServerMsg, string(body))
 }
 
+func TestStatusCodeCreated(t *testing.T) {
+	t.Parallel()
+
+	port := 8305
```
gozssky, line 76: It seems we can use `httptest.NewServer` to avoid port conflict.

gozssky (approved):
Rest LGTM.

pkg/httputil/httputil_test.go
amyangfei: addressed.

Issue #7304, opened 2022-10-10
create local storage met `inconsistent executor ID of local file` error and panic


amyangfei:
### What did you do?

run engine integration test
ref: https://ci2.pingcap.net/blue/organizations/jenkins/engine_ghpr_integration_test/detail/engine_ghpr_integration_test/1120/pipeline

detail logs: 
[Uploading log-dm_case_sensitive-dm_full_mode.tar.gz…]()

### What did you see instead?

```go
[2022/10/10 02:38:11.406 +00:00] [PANIC] [file_manager.go:308] ["inconsistent executor ID of local file"] [fileScope="{"Executor":"server-executor-2-0a5789b4","WorkerID":"4fc9df31-d6d5-4704-a2db-c5d110932e5e","Name":"31788357-36c4-49a9-907f-39ac23856517/mysql-01"}"] [creator=""server-executor-2-0a5789b4""] [currentExecutor=server-executor-0-7ae07045] [stack="github.com/pingcap/tiflow/engine/pkg/externalresource/internal/local.(*FileManager).validateIdent
        github.com/pingcap/tiflow/engine/pkg/externalresource/internal/local/file_manager.go:308
github.com/pingcap/tiflow/engine/pkg/externalresource/internal/local.(*FileManager).GetPersistedResource
        github.com/pingcap/tiflow/engine/pkg/externalresource/internal/local/file_manager.go:84
github.com/pingcap/tiflow/engine/pkg/externalresource/broker.(*DefaultBroker).getPersistResource
        github.com/pingcap/tiflow/engine/pkg/externalresource/broker/broker.go:247
github.com/pingcap/tiflow/engine/pkg/externalresource/broker.(*DefaultBroker).OpenStorage
        github.com/pingcap/tiflow/engine/pkg/externalresource/broker/broker.go:106
github.com/pingcap/tiflow/engine/framework.(*DefaultBaseWorker).OpenStorage
        github.com/pingcap/tiflow/engine/framework/worker.go:514
github.com/pingcap/tiflow/engine/executor/dm.(*dmWorker).setupStorage
        github.com/pingcap/tiflow/engine/executor/dm/worker.go:195
github.com/pingcap/tiflow/engine/executor/dm.(*dmWorker).InitImpl
        github.com/pingcap/tiflow/engine/executor/dm/worker.go:139
github.com/pingcap/tiflow/engine/framework.(*DefaultBaseWorker).Init
[...]

PR #7319, opened 2022-10-11
worker(engine/dm): no need extStorage if sync is not fresh

GMHDBJD:

Issue Number: ref #7304 

- add NoNeedExtStorage in TaskCfg(true when sync is not fresh)
- do not OpenStorage if NoNeedExtStorage
- use TaskCfg rather than SubTaskCfg to CreateWorker

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


lance6716:
related to https://github.com/pingcap/tiflow/issues/6784#issuecomment-1220377608 ?

GMHDBJD:
> related to [#6784 (comment)](https://github.com/pingcap/tiflow/issues/6784#issuecomment-1220377608) ?

Yes, but not completely resolved it, we need discard storage.

engine/executor/dm/worker.go
```diff
@@ -135,7 +137,7 @@ func (w *dmWorker) InitImpl(ctx context.Context) error {
 	if err := w.messageAgent.UpdateClient(w.masterID, w); err != nil {
 		return err
 	}
-	if w.cfg.Mode != dmconfig.ModeIncrement {
+	if w.cfg.Mode != dmconfig.ModeIncrement && !w.noNeedExtStorage {
```
lance6716, line 140: can we use needExtStorage? double negative is a bit hard to think

lance6716 (approved):

amyangfei (approved):

amyangfei:
Will there be more PRs to fix #7304 for scenario that error happens between dump and load tasks?

PR #7336, opened 2022-10-12
engine(dm, externalresouce): enable s3 for dm

### LINKED ISSUE #7114: Engine supports using s3 to store file resources

CharlesCheung96:

Issue Number: close #7114, ref #7304

1. s3 resources are used by default in dm tasks.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
`None`.
```


GMHDBJD:
can you also deploy minio in helm? https://github.com/pingcap/tiflow/blob/master/deployments/engine/helm/tiflow/values.yaml

CharlesCheung96:
> can you also deploy minio in helm? https://github.com/pingcap/tiflow/blob/master/deployments/engine/helm/tiflow/values.yaml

We do not need to deploy minio directly in helm, since a public minio service is available on test_infra. Here we provide a [configuration file](https://github.com/pingcap/tiflow/pull/7336/files#diff-0a8fd3e8ce548987a30624a56f4d9f93c8079a5a2dd378dd516de6d964309852) to deploy minio in k8s, which can be deployed manually in githuc action.

engine/jobmaster/dm/utils.go
```diff
@@ -19,5 +19,5 @@ import (
 
 // NewDMResourceID returns a ResourceID in DM's style. Currently only support local resource.
```
GMHDBJD, line 20: Please update the comment.

engine/jobmaster/dm/utils.go
```diff
@@ -19,5 +19,5 @@ import (
...
 func NewDMResourceID(taskName, sourceName string) resModel.ResourceID {
-	return "/" + string(resModel.ResourceTypeLocalFile) + "/" + taskName + "/" + sourceName
+	return "/" + string(resModel.ResourceTypeS3) + "/" + taskName + "-" + sourceName
```
GMHDBJD, line 22: Is this a definite or configurable option? If it's definite, what happens if the executor is not configured with s3?

CharlesCheung96: The current implementation is deterministic, which would cause errors when s3 is not configured. To solve this problem, we need to figure out: 
1. Add a configurable parameter for the dm job, or get the information about whether the framework is configured with s3 through the BaseWorker interface?
2. Restore the failover logic of the dm so that all dump jobs are created with resource Constraint.

GMHDBJD: cc @amyangfei @lance6716

CharlesCheung96:
chao tests verified in https://github.com/CharlesCheung96/tiflow/actions/runs/3247512899

engine/jobmaster/dm/utils.go
CharlesCheung96: Maybe we could fix this in another pr.

deployments/engine/docker-compose/3m3e_with_s3.yaml
```diff
@@ -0,0 +1,194 @@
+version: "2.3"
+services:
+  server-master-0:
```
amyangfei, line 3: Can add `restart` policy to server-master and executor containers.

engine/jobmaster/dm/utils.go
amyangfei: What about exposing an option to decide using s3 or local file in DM job config. 
- Maybe in the future this is decided by user or dataflow service
- Besides adding an option makes it easy to debug in local environment, without of setup a minio server.

deployments/engine/docker-compose/config/master_with_s3.toml
```diff
@@ -0,0 +1,27 @@
...
+initial-interval = "1s"
+max-interval = "15s"
+max-try-time = 100
+
+[storage.s3]
+bucket = "engine-it"
+endpoint = "http://minio-standalone:9000/"
+access-key = "engine"
+secret-access-key = "engineSecret"
+force-path-style = true
```
amyangfei, line 27: Do we always set `force-path-style` to true, if so what about setting it in code instead of configuration file.

deployments/engine/docker-compose/3m3e_with_s3.yaml
lance6716: For normal test I prefer we don't restart on failure, in case a panic bug is ignored by restart.

deployments/engine/docker-compose/config/executor_with_s3.toml
```diff
@@ -0,0 +1,13 @@
+keepalive-ttl = "20s"
+keepalive-interval = "500ms"
+session-ttl = 20
```
lance6716, line 3: Can we use default value? don't specify them in config files

deployments/engine/docker-compose/config/master_with_s3.toml
```diff
@@ -0,0 +1,27 @@
...
+password = ""
+
+[log]
+level = "debug"
+
+[job-backoff]
+reset-interval = "2s"
+initial-interval = "1s"
+max-interval = "15s"
+max-try-time = 100
```
lance6716, line 20: ditto

engine/pkg/externalresource/model/model.go
```diff
@@ -165,10 +168,31 @@ func ParseResourceID(rpath ResourceID) (ResourceType, ResourceName, error) {
...
+func BuildResourceID(rtype ResourceType, resName ResourceName) ResourceID {
+	name, err := DecodeResourceName(resName)
+	if err != nil {
+		log.Panic("invalid resource name", zap.Error(err))
+	}
 	return path.Join("/"+string(rtype), name)
 }
+
+// EncodeResourceName encodes raw resource name to a valid resource name.
+func EncodeResourceName(resNameStr string) ResourceName {
```
lance6716, line 184: why we need to repeat `resNameStr` twice in the return value?

engine/pkg/externalresource/model/model.go
```diff
@@ -165,10 +168,31 @@ func ParseResourceID(rpath ResourceID) (ResourceType, ResourceName, error) {
...
+	prefix := hex.EncodeToString([]byte(resNameStr))
+	suffix := strings.ReplaceAll(resNameStr, "/", "-")
+	return ResourceName(prefix + "-" + suffix)
+}
+
+// DecodeResourceName decodes resource name to raw resource name.
+func DecodeResourceName(encodedName ResourceName) (string, error) {
```
lance6716, line 191: good to try fuzzing test for `EncodeResourceName` and `DecodeResourceName`

lance6716 (commented):
rest lgtm

lance6716:
Can some TODO in chaos test also be fixed? @GMHDBJD

engine/pkg/externalresource/model/model.go
CharlesCheung96: This is used for debugging and only the part of hex encode is kept in the design

deployments/engine/docker-compose/config/executor_with_s3.toml
CharlesCheung96: Maybe we can solve this problem in another pr, ref: https://github.com/pingcap/tiflow/issues/7380

lance6716 (approved):

engine/jobmaster/dm/utils.go
CharlesCheung96: This feature will be implemented in other pr.

amyangfei (approved):

PR #7339, opened 2022-10-12
codec(ticdc): optimize canal-json encoder by reduce allocate small objects

### LINKED ISSUE #7354: canal-json encoder memory allocation optimization
current `canal-json` encoder allocates a huge number of small objects, such as `canal.Header` and `canal.Column`, `canal.RowData`, also `canalJsonMessage`, we can reduce them by avoid produce those intermediate products.

3AceShowHand:

Issue Number: close #7354

do not generate intermediate products to avoid allocating a huge number of small objects, to reduce overhead on GC.

K9MHfQ33aB

On the master branch, 23 ~ 25k; On the PR branch, 26 ~ 28K, throughput increased around 10 ~ 15% 

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


cdc/sink/codec/canal/canal_json_message.go
```diff
@@ -82,7 +82,7 @@ func (c *canalJSONMessage) getTable() *string {
 	return &c.Table
 }
 
-// for canalJSONMessage, we lost the commitTs.
+// for CanalJSONMessage, we lost the commitTs.
```
zhaoxinyu, line 85: Is this a typo? It seems that the original `canalJSONMessage` is correct.

cdc/sink/codec/canal/canal_json_encoder.go
```diff
@@ -24,133 +24,162 @@ import (
...
+	baseMessage.BuildTime = time.Now().UnixNano() / 1e6 // ignored by both Canal Adapter and Flink
+	baseMessage.Query = ""
+	baseMessage.SQLType = sqlTypeMap
+	baseMessage.MySQLType = mysqlTypeMap
+	baseMessage.Old = nil
+	baseMessage.tikvTs = e.CommitTs
+
 	if e.IsDelete() {
-		msg.Data = append(msg.Data, oldData)
+		baseMessage.Data[0] = oldData
```
zhaoxinyu, line 133: What if we make `Data` field just a `map[string]interface{}` since it only contains one item?

3AceShowHand: it's represented as `an array with only one item` officially.

zhaoxinyu (approved):

cdc/sink/codec/canal/canal_json_encoder.go
```diff
@@ -24,133 +24,147 @@ import (
...
-	builder     *canalEntryBuilder
-	messageBuf  []canalJSONMessageInterface
-	callbackBuf []func()
+	builder *canalEntryBuilder
+
 	// When it is true, canal-json would generate TiDB extension information
 	// which, at the moment, only includes `tidbWaterMarkType` and `_tidb` fields.
 	enableTiDBExtension bool
+
+	messageHolder canalJSONMessageInterface
```
zhaoxinyu, line 39: It would be better to add some comments about this field because it is relevant to memory efficiency.

asddongmen (approved):

PR #7358, opened 2022-10-14
etcd(ticdc): not skip etcd worker tick if nothing is changed

### LINKED ISSUE #7357: not skip etcd worker tick if nothing is changed
https://github.com/pingcap/tiflow/blob/c5b1dd0edcf2d1b03838619fdb27acab908b1f27/pkg/orchestrator/reactor_state.go#L262-L282

always add a pending patch, but nothing is changed,  that will make etcd worker skip one tick

sdojjy:

Issue Number: close #7357

https://github.com/pingcap/tiflow/blob/c5b1dd0edcf2d1b03838619fdb27acab908b1f27/pkg/orchestrator/reactor_state.go#L262-L282
always add a pending patch even if nothing is changed
in this PR, 
1. check if something is really committed to etcd server, if not, continue the current tick loop.
2. try to commit changes at the end of tick 

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


overvenus (approved):
Could you add a test?

pkg/orchestrator/etcd_worker.go
```diff
@@ -163,9 +163,11 @@ func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session,
 	}
 
 	var (
-		pendingPatches [][]DataPatch
-		exiting        bool
-		sessionDone    <-chan struct{}
+		pendingPatches  [][]DataPatch
+		commitedChanges int
```
3AceShowHand, line 167:
```suggestion
		committedChanges int
```

sdojjy: fixed

pkg/orchestrator/etcd_worker.go
```diff
@@ -229,58 +231,75 @@ func (worker *EtcdWorker) Run(ctx context.Context, session *concurrency.Session,
...
 			}
+			return false, nil
+		}
+		if retry, err = tryCommitPendingPatches(); err != nil {
+			return err
+		}
+		if retry {
+			continue
+		}
+		if exiting {
```
3AceShowHand, line 258: it looks, we should check `exsting` before `retry`

sdojjy: done

asddongmen (approved):

PR #7367, opened 2022-10-14
scheduler(ticdc): agent always sends the current table status to the owner when ticking

### LINKED ISSUE #7366: agent always send current table status to owner when tick
agent always sends the current table status to the owner when ticking to fast forward resolved ts

sdojjy:

Issue Number: close #7366

agent always sends the current table status to the owner when ticking to fast forward resolved ts

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
 `None`.
```


overvenus:
Processor heartbeats will be buffered in memory before owner handling these messages. So I'm worried about:

1. Owner handles heartbeats periodically, we should notify owner to handle buffered heartbeats somehow.
2. Can it oom when there are many TiCDC instance and owner is slow?
3. Does it exhaust processor CPU when there are many tables?

3AceShowHand (approved):

sdojjy:
> Processor heartbeats will be buffered in memory before owner handling these messages. So I'm worried about:
> 
> 1. Owner handles heartbeats periodically, we should notify owner to handle buffered heartbeats somehow.
> 2. Can it oom when there are many TiCDC instance and owner is slow?
> 3. Does it exhaust processor CPU when there are many tables?

 1. Owner handles heartbeats periodically, we should notify owner to handle buffered heartbeats somehow.
    The logic of handling heartbeats on the owner side does not need to be changed, it handles heartbeat every tick in `c.scheduler.Tick` 
 2. Can it oom when there are many TiCDC instance and owner is slow?
    The message size is small and only sends one more heartbeat for the default processor-flush-interval, If we change the owner and the processor tick interval both to 50ms in the future, no more messages will be added.
 3. Does it exhaust processor CPU when there are many tables?
    No,  heartbeat logic on the processor side is fast, and scheduler v2 sends resolved ts to the owner every processor tick interval.

overvenus:
How about setting `debug.heartbeat-tick = 1`?

nongfushanquan (approved):

PR #7372, opened 2022-10-14
pkg(dm): improve the backward compatibility of “SHOW SLAVE HOSTS“ command

### LINKED ISSUE #5017: improve the backward compatibility of “SHOW SLAVE HOSTS“ command

lyzx2001:

Issue Number: close #5017 

I used `export.GetSpecifiedColumnValueAndClose` and `export.GetSpecifiedColumnValuesAndClose` to retrieve the requested columns, instead of restricting the return format of the columns, which is not robust.

#### Tests 

 - Unit test

No.
No.
### Release note 

```release-note
None
```


lance6716:
DM unit test is failed

lance6716:
the failed case is 

```
[2022-10-26T13:29:09.902Z]     server_test.go:2888: 
[2022-10-26T13:29:09.902Z]         	Error Trace:	/tmp/TestMaster1418414956/001/server_test.go:2888
[2022-10-26T13:29:09.902Z]         	Error:      	Should be true
[2022-10-26T13:29:09.902Z]         	Test:       	TestMaster
```

dm/pkg/conn/utils.go
```diff
@@ -144,23 +141,63 @@ func GetMasterStatus(ctx *tcontext.Context, db *BaseDB, flavor string) (
...
+		rowsResult, err = export.GetSpecifiedColumnValuesAndClose(rows, "File", "Position", "Binlog_Do_DB", "Binlog_Ignore_DB", "Executed_Gtid_Set")
+		if err != nil {
+			err = terror.DBErrorAdapt(err, terror.ErrDBDriverError)
+			return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
+		}
+
+		if len(rowsResult) == 0 {
+			err = terror.ErrNoMasterStatus.Generate()
+			return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
+		} else if len(rowsResult[0]) != 0 {
```
lance6716, line 156: will `len(rowsResult[0]) != 5` happen?

dm/pkg/conn/utils.go
```diff
@@ -144,23 +141,63 @@ func GetMasterStatus(ctx *tcontext.Context, db *BaseDB, flavor string) (
...
+			binlogName = rowsResult[0][0]
+			var posInt int
+			posInt, err = strconv.Atoi(rowsResult[0][1])
+			if err != nil {
+				err = terror.DBErrorAdapt(err, terror.ErrDBDriverError)
+				return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
+			}
+			pos = uint32(posInt)
```
lance6716, line 164: convert int to uint32 is a bit dangerous because int type can be 32 or 64 bits.
We can parse it to type of 64 bit first and then convert it to uint32

and suddenly I found this should be int64 😂  https://github.com/mysql/mysql-server/blob/a246bad76b9271cb4333634e954040a970222e0a/sql/rpl_source.cc#L1225-L1226 we can open an issue to record

dm/pkg/conn/utils.go
lyzx2001: According to the implementation of GetSpecifiedColumnValuesAndClose, `len(rowsResult[0])` is always equal to the number of the column names that are passed in, which is 5 here. It is possible that some of the retrieved values are "".

dm/pkg/conn/utils.go
lyzx2001: Modified

dm/pkg/conn/utils.go
lance6716: then you can remove it or should check `!= 5` rather than `!= 0` and print a log when not so

dm/pkg/conn/utils.go
```diff
@@ -144,23 +141,63 @@ func GetMasterStatus(ctx *tcontext.Context, db *BaseDB, flavor string) (
...
+		rowsResult, err = export.GetSpecifiedColumnValuesAndClose(rows, "File", "Position", "Binlog_Do_DB", "Binlog_Ignore_DB")
+		if err != nil {
+			err = terror.DBErrorAdapt(err, terror.ErrDBDriverError)
+			return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
+		}
+
+		if len(rowsResult) == 0 {
+			err = terror.ErrNoMasterStatus.Generate()
+			return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
+		} else if len(rowsResult[0]) != 0 {
```
lance6716, line 179: ditto

dm/pkg/conn/utils.go
```diff
@@ -144,23 +141,63 @@ func GetMasterStatus(ctx *tcontext.Context, db *BaseDB, flavor string) (
...
+			binlogName = rowsResult[0][0]
+			var posInt int
+			posInt, err = strconv.Atoi(rowsResult[0][1])
+			if err != nil {
+				err = terror.DBErrorAdapt(err, terror.ErrDBDriverError)
+				return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
+			}
+			pos = uint32(posInt)
```
lance6716, line 187: ditto

dm/pkg/conn/utils_test.go
```diff
@@ -59,3 +60,71 @@ func TestGetBinlogDB(t *testing.T) {
 	require.Equal(t, binlogIgnoreDB, "ignore_db")
 	require.Nil(t, mock.ExpectationsWereMet())
 }
+
+func TestGetMasterStatus(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), DefaultDBTimeout)
```
lance6716, line 65: seems can directly use Background

dm/pkg/utils/db_test.go
```diff
@@ -389,3 +389,42 @@ func TestCreateTableSQLToOneRow(t *testing.T) {
...
+
+	for _, ca := range cases {
+		mock.ExpectQuery("SHOW SLAVE HOSTS").WillReturnRows(ca.rows)
+		results, err2 := GetSlaveServerID(context.Background(), db)
+		require.NoError(t, err2)
+		require.Equal(t, ca.results, results)
+		require.NoError(t, mock.ExpectationsWereMet())
+	}
+
+	require.NoError(t, mock.ExpectationsWereMet())
```
lance6716, line 429:
```suggestion
```

dm/pkg/conn/utils_test.go
```diff
@@ -59,3 +60,71 @@ func TestGetBinlogDB(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, ca.binlogName, binlogName)
+		require.Equal(t, ca.pos, pos)
+		require.Equal(t, ca.binlogDoDB, binlogDoDB)
+		require.Equal(t, ca.binlogIgnoreDB, binlogIgnoreDB)
+		require.Equal(t, ca.gtidStr, gtidStr)
+		require.NoError(t, mock.ExpectationsWereMet())
+	}
+
+	require.NoError(t, mock.ExpectationsWereMet())
```
lance6716, line 129:
```suggestion
```

dm/pkg/conn/utils.go
```diff
@@ -144,23 +141,69 @@ func GetMasterStatus(ctx *tcontext.Context, db *BaseDB, flavor string) (
...
+			err = terror.DBErrorAdapt(err, terror.ErrDBDriverError)
+			return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
+		}
+
+		switch {
+		case len(rowsResult) == 0:
+			err = terror.ErrNoMasterStatus.Generate()
+			return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
+		case len(rowsResult[0]) != 5:
+			ctx.L().Warn("The number of columns that SHOW MASTER STATUS returns for MySQL is not equal to 5, will not use the retrieved information")
```
lance6716, line 158:
```suggestion
			ctx.L().DPanic("The number of columns that SHOW MASTER STATUS returns for MySQL is not equal to 5, will not use the retrieved information")
```
and we should return an error for this case, like `err = terror.ErrNoMasterStatus.Generate()`. Otherwise caller may has no ability to handle that case

dm/pkg/conn/utils.go
```diff
@@ -144,23 +141,69 @@ func GetMasterStatus(ctx *tcontext.Context, db *BaseDB, flavor string) (
...
+			err = terror.DBErrorAdapt(err, terror.ErrDBDriverError)
+			return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
+		}
+
+		switch {
+		case len(rowsResult) == 0:
+			err = terror.ErrNoMasterStatus.Generate()
+			return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
+		case len(rowsResult[0]) != 4:
+			ctx.L().Warn("The number of columns that SHOW MASTER STATUS returns for MariaDB is not equal to 4, will not use the retrieved information")
```
lance6716, line 184: ditto

dm/pkg/utils/db.go
```diff
@@ -132,37 +128,20 @@ func GetSlaveServerID(ctx context.Context, db *sql.DB) (map[uint32]struct{}, err
...
-			err = rows.Scan(&serverID, &host, &port, &masterID)
-		}
+	var rowsResult []string
+	rowsResult, err = export.GetSpecifiedColumnValueAndClose(rows, "Server_id")
+	if err != nil {
+		return nil, terror.DBErrorAdapt(err, terror.ErrDBDriverError)
+	}
+	for _, serverID := range rowsResult {
+		// serverID will not be null
+		serverIDInt, err := strconv.Atoi(serverID)
```
lance6716, line 139: may raise error when int is 32-bit but `serverID` is in the upper-half of uint32 values?

lance6716 (commented):
rest lgtm

dm/pkg/utils/db_test.go
```diff
@@ -389,3 +390,60 @@ func TestCreateTableSQLToOneRow(t *testing.T) {
...
+	}
+
+	casesWithErr := []struct {
+		rows *sqlmock.Rows
+		err  error
+	}{
+		// For MariaDB, when invalid conversion of Server_id (converting int32 in the upper-half of uint32 values to uint32)
+		{
+			sqlmock.NewRows([]string{"Server_id", "Host", "Port", "Master_id"}).
+				AddRow(-1, "iconnect2", 3306, 192168011).
```
lance6716, line 431: -1 is not a valid server ID?

https://dev.mysql.com/doc/refman/5.7/en/replication-options.html#sysvar_server_id

dm/pkg/utils/db.go
lance6716: You can use https://pkg.go.dev/strconv#ParseUint, don't use int because its range is undetermined

lance6716 (approved):

dm/pkg/conn/utils_test.go
```diff
@@ -59,3 +60,67 @@ func TestGetBinlogDB(t *testing.T) {
...
+		},
+		// For MariaDB
+		{
+			sqlmock.NewRows([]string{"File", "Position", "Binlog_Do_DB", "Binlog_Ignore_DB"}).
+				AddRow("mariadb-bin.000016", 475, "", ""),
+			"mariadb-bin.000016",
+			475,
+			"",
+			"",
+			"0-1-2",
```
lichunzhu, line 102: Nit: It won't appear in mariaDB
```suggestion
			"",
```

lichunzhu (approved):
LGTM

dm/pkg/conn/utils_test.go
lyzx2001: I conformed to this example

https://github.com/pingcap/tiflow/blob/c982fca9d18e295c20b826e4f81490007ea96e80/dm/pkg/conn/utils.go#L133-L146

Here its `gtidStr` is `"0-1-2"` in MariaDB.
I have changed my tests according to your suggestion, do I need to modify this comment as well?

lichunzhu: Sorry, I make a mistake. Current test is correct. Can you revert https://github.com/pingcap/tiflow/pull/7372/commits/1c100ab8fe57c490c246ecf8ff5552a0ccf03b3f? After that we can merge this PR

lyzx2001: Modified

PR #7388, opened 2022-10-17
sinkmanager(ticdc): init

hi-rustin:

 init sink manager implement.

#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```


cdc/processor/sinkmanager/manager_impl.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// generateTableSinkFetchTask generates a fetch tableSinkTask to fetch data from the source manager.
+func (m *ManagerImpl) generateTableSinkFetchTask() error {
+	for {
+		select {
+		case <-m.closedChan:
+			return nil
+		case <-m.taskTicker.C:
+			slowestTableProgress := m.progressHeap.pop()
```
hicqu, line 85: It's better to check whether there are any free workers or not. If there are no free workers, it's not necessary to generate a task.

cdc/processor/sinkmanager/manager_impl.go
```diff
@@ -0,0 +1,183 @@
...
+			tableSink, ok := m.tableSinks.Load(slowestTableProgress.tableID)
```
hicqu, line 86: Seems it's better to store `state`s instead of `progress`es in the heap. So we can check `state.removed` to know whether the table is still valid or not.

cdc/processor/sinkmanager/manager_impl.go
```diff
@@ -0,0 +1,183 @@
...
+				continue
+			}
+			// If redo log is disabled, we use the barrier ts as the upper bound of the fetch tableSinkTask.
+			// Because can not exceed the barrier ts.
+			barrierTs := m.lastBarrierTs.Load()
+			upperBoundPos := sorter.Position{
+				StartTs:  barrierTs - 1,
+				CommitTs: barrierTs,
+			}
+			canBeAdvance := slowestTableProgress.nextLowerBoundPos.Less(upperBoundPos)
```
hicqu, line 100: Seems should use `LessOrEqual`, because upper bound is included.

cdc/processor/sinkmanager/manager_impl.go
```diff
@@ -0,0 +1,183 @@
...
+			}
+		}
+	}
+}
+
+// UpdateTableResolvedTs updates the resolved ts of the table.
+func (m *ManagerImpl) UpdateTableResolvedTs(tableID model.TableID, ts model.Ts) {
+	tableSink, ok := m.tableSinks.Load(tableID)
+	if !ok {
+		panic("table sink not found")
```
hicqu, line 135: Seems we use log.Panic.

cdc/processor/sinkmanager/manager_impl.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	tableSink.(*tableSinkWrapper).updateCurrentSorterResolvedTs(ts)
+}
+
+// UpdateBarrierTs updates the barrier ts of all tables in the sink manager.
+func (m *ManagerImpl) UpdateBarrierTs(ts model.Ts) {
+	if ts <= m.lastBarrierTs.Load() {
```
hicqu, line 142: If the check is really necessary, we should use
```go
for {
    old := lastBarrierTs.Load()
    if old < new && lastBarrierTs.CompareAndSwap(old, new) {
        break
    }
}
```

cdc/processor/sinkmanager/mem_quota.go
```diff
@@ -0,0 +1,39 @@
...
+package sinkmanager
+
+import "github.com/pingcap/tiflow/cdc/model"
+
+//nolint:varcheck,deadcode
+const defaultMemoryUsage = 10 * 1024 * 1024 // 10MB
+
+type memQuota interface {
+	// TryAcquire try to acquire n bytes of memory.
+	TryAcquire() bool
```
hicqu, line 23: How about add a parameter to specify required bytes?

cdc/processor/sinkmanager/mem_quota.go
```diff
@@ -0,0 +1,39 @@
...
+	// ForceAcquire force to acquire n bytes of memory.
+	// It will exceed the total memory limit.
+	ForceAcquire()
```
hicqu, line 26: Seems we need
```go
func ForceAcquire(clientID int, bytes int)
```

`MemoryQuota` should ensure there is at most 1 client can acquire memory exceeds the limit.

hicqu, line 26: Seems we need to comment `there will at most 1 client can acquire memory exceeds the limit`.

pkg/sorter/sorter.go
```diff
@@ -0,0 +1,125 @@
...
+	StartTs  model.Ts
+}
+
+// Valid indicates whether the position is valid or not.
+func (p Position) Valid() bool {
+	return p.StartTs != 0 && p.CommitTs != 0
+}
+
+// Less returns true if p is less than others.
+func (p Position) Less(other Position) bool {
```
hicqu, line 106: Seems `func (p Position) Compare(other Position) int` is better.

cdc/processor/sinkmanager/worker_impl.go
```diff
@@ -0,0 +1,136 @@
...
+				events = append(events, e)
+				// We meet a finished transaction.
+				if pos.Valid() {
+					lastPos = pos
+					if w.splitTxn {
+						w.memQuota.ResetBatchID(t.tableID)
+					}
+					// Always emit the events to the sink.
+					// Whatever splitTxn is true or false, we should emit the events to the sink as soon as possible.
+					err := w.emitEventsToTableSink(events, t, e.CRTs)
```
hicqu, line 72: If there are lots of small transactions, it can call `memQuota.Record` too much times. Maybe it's a problem.
How about just writeEvents into the sink, but update resolved ts later? For example, we can update resolved ts per 256K.

hicqu: So we can call `memQuota.Record` with `sink.UpdateResolvedTs`, to reduce the frequency.

cdc/processor/sinkmanager/manager_impl.go
hicqu: So we can use a small ticker like `10ms`. Or we can do some refine to remove the ticker at all.

I'm worry about `100ms` is too large. It can increase latency.

hi-rustin: If there is no worker, it blocks on the send task.

cdc/processor/sinkmanager/manager_impl.go
hi-rustin: What kind of state do you mean?

cdc/processor/sinkmanager/manager_impl.go
hi-rustin: It seems we only use this method in the processor, so maybe this is acceptable. Other threads only read it.

cdc/processor/sinkmanager/mem_quota.go
hi-rustin: I guess it's OK to have some workers exceed the limitation. Compared to the old implementation, we exceeded the limits to a manageable degree. Originally it was the number of tables that determined this, but now it is up to n workers.

cdc/processor/sinkmanager/worker_impl.go
hi-rustin: Do we send too infrequently if it causes delays? For example, it can't reach 256k in a short time? Do we have to introduce a timeout for this?

cdc/processor/sinkmanager/manager_impl.go
hi-rustin: I'll change it.

cdc/processor/sinkmanager/worker_impl.go
hicqu: Timeout seems unnecessary. We always do a flush when this task is finished.

cdc/processor/sinkmanager/mem_quota.go
hicqu: We can pass 10MB as a parameter into these functions.

cdc/processor/sinkmanager/manager_impl.go
hicqu: If so it's better to comment it on the function.

cdc/processor/sinkmanager/manager_impl.go
hicqu: If so, it's better to load barrierTs when workers receive a task, instead of manager sends the task.

cdc/processor/sinkmanager/manager_impl.go
hicqu: The table's state, include `stopped`, `barrierTs`, `resolvedTs` and some other things.

hi-rustin: These things are stored in the table sink wrapper.

cdc/processor/sinkmanager/mem_quota.go
hicqu: If table count is less than worker count, this implementation is same bad with the old one. I guess it shouldn't be very hard to make it better.

cdc/processor/sinkmanager/worker_impl.go
hi-rustin: I see. I'll try it.

cdc/processor/sinkmanager/mem_quota.go
hi-rustin: > I guess it shouldn't be very hard to make it better.

Now it's a multi-table share, in which case competing workers are coordinated and they can affect each other. For example, what if the slowest kind of table doesn't get the only quota? This would still affect latency.

hicqu: If memory quota full really happens, I think it's more important to avoid OOM instead of take care about latency.

hicqu: Seems even if at most 1 client can acquire memory on quota full, OOM is still possible to happen.

Let's ignore OOM problem here. We can deal with it in another place.

cdc/processor/sinkmanager/manager_impl.go
hi-rustin: Make sense.

hi-rustin: Please check out https://github.com/pingcap/tiflow/pull/7388/commits/9883f4573a646cb7ff784959adadad24ea8d7bf3.

cdc/processor/sinkmanager/manager_impl.go
hi-rustin: Please check out https://github.com/pingcap/tiflow/pull/7388/commits/9883f4573a646cb7ff784959adadad24ea8d7bf3

cdc/processor/sinkmanager/manager_impl.go
hi-rustin: Please checkout https://github.com/pingcap/tiflow/pull/7388/commits/ffd4e8836844d37c6e427eb2892a27a1ee8b82d2

cdc/processor/sinkmanager/manager_impl.go
hi-rustin: Added. Please check out https://github.com/pingcap/tiflow/pull/7388/commits/87ed02fbf93089d1dfa8101c5f0a097a63b48f18

cdc/processor/sinkmanager/mem_quota.go
hi-rustin: Updated. Please checkout https://github.com/pingcap/tiflow/pull/7388/commits/6fbf0762fb2b36dde8d50efcb7d88a470d769df6

cdc/processor/sinkmanager/worker_impl.go
hi-rustin: Added. Please check out https://github.com/pingcap/tiflow/pull/7388/commits/09151209d27f18bfa0b7b6bfd97d8966b5081909

cdc/processor/sinkmanager/worker_impl.go
```diff
@@ -0,0 +1,170 @@
...
+	t.tableSink.emitRowChangedEvent(rowChangeEvents...)
+	return uint64(size), nil
+}
+
+func (w *workerImpl) updateTableSinkResolvedTs(t *tableSinkTask, commitTs model.Ts, barrierTs model.Ts, size uint64) error {
+	tableSinkResolvedTs := commitTs
+	if barrierTs < tableSinkResolvedTs {
+		tableSinkResolvedTs = barrierTs
+	}
+	if w.redoManager != nil {
```
hicqu, line 153: How about removing redoManager from this function? Let's focus on table sink here. We can handle redo-related things in another function.

And parameter `barrierTs` can be removed, too.

hicqu:
Rest LGTM.

cdc/processor/sinkmanager/worker_impl.go
hi-rustin: Oh. I am stupid. It never over the barrier(aka upper bound).

hi-rustin: Thanks!

hicqu (approved):

hi-rustin:
/hold 
This PR will be split into multi PRs.

PR #7390, opened 2022-10-17
engine(externalresource): configure storage only on master

CharlesCheung96:

Issue Number: ref #7304

1. configure storage only on master
2. fix some teses in gcrunner
3. add opts for `broker.OpenStorage`

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


CharlesCheung96:
chaos test verified in :https://github.com/CharlesCheung96/tiflow/actions/runs/3270307081

engine/pkg/externalresource/broker/interfaces.go
```diff
@@ -46,3 +47,17 @@ type Broker interface {
 
 	Close()
 }
+
+type openStorageOptions struct {
+	needEmpty bool
```
amyangfei, line 52: Does `needEmpty` mean `cleanBeforeOpen`?

engine/proto/master.proto
```diff
@@ -274,3 +275,11 @@ message DeleteJobRequest {
     string tenant_id = 2;
     string project_id = 3;
 }
+
+message QueryStorageConfigRequest {
+}
+
+message QueryStorageConfigResponse {
+    Error err = 1;
+    string address = 2;
```
amyangfei, line 284: `address` is not accurate, maybe rename it to `config` is better.

engine/pkg/externalresource/broker/interfaces.go
CharlesCheung96: Yes, seems `cleanBeforeOpen`  is clearer.

engine/executor/dm/worker.go
```diff
@@ -193,11 +193,18 @@ func (w *dmWorker) CloseImpl(ctx context.Context) {
 // setupStorage opens and configs external storage
 func (w *dmWorker) setupStorage(ctx context.Context) error {
 	rid := dm.NewDMResourceID(w.cfg.Name, w.cfg.SourceID)
-	h, err := w.OpenStorage(ctx, rid)
+	openStorageFunc := func() (broker.Handle, error) {
```
GMHDBJD, line 196: how about
```
opts:=make([]opts)
if wt==Dump {
    opts=append(opts, broker.WithEmptyOption())
}
h, err = w.OpenStorage(ctx,rid, opts...)
```

amyangfei (approved):

PR #7391, opened 2022-10-18
workload (ticdc): add workload

asddongmen:

Issue Number: close #xxx

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note

```


pkg/workload/main.go
```diff
@@ -0,0 +1,203 @@
+// Copyright 2022 PingCAP, Inc.
```
lonng, line 1: Seems this is a binary for the test. Why not put it into the `tests/workload` directory?

asddongmen: Since QE team want to use this workload to do the TiCDC perfromance testing, and I think it's natural to put it in `/pkg` if another repo will import it.  Do you think it is better to put it into `tests/workload`?

sdojjy (approved):

Issue #6985, opened 2022-09-07
Tracking issues for the grafana and prometheus metric


maxshuang:
### Is your feature request related to a problem?

No

### Describe the feature you'd like

- [ ] Add basic grafana page for Tiflow
- [ ] Add a dm job list table
- [ ] Add some k8s variables to grafana dashboard for cloud deployment
- [ ] Add metrics for tiflow components

PR #7402, opened 2022-10-19
engine(metrics): remove jobtype prefix in metrics

CharlesCheung96:

Issue Number: ref #6985

1. remove jobtype prefix in metrics

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


lance6716 (commented):
what's the effect before and after this PR?

engine/framework/worker.go
```diff
@@ -236,8 +236,9 @@ func NewBaseWorker(
 		errCenter:            errctx.NewErrCenter(),
 		clock:                clock.New(),
 		businessMetaKVClient: cli,
-		metricFactory:        promutil.NewFactory4Worker(ctx.ProjectInfo, MustConvertWorkerType2JobType(tp), masterID, workerID),
-		logger:               frameLog.WithWorkerID(frameLog.WithMasterID(logger, masterID), workerID),
+		metricFactory: promutil.NewFactory4Worker(ctx.ProjectInfo,
```
amyangfei, line 239: Is it better to change the function signature to 
```go
func NewFactory4Worker(info tenant.ProjectInfo, prefix string, jobID engineModel.JobID, workerID frameModel.WorkerID,
```
Then we don't need to add the job type `JobTypeNameNone`

CharlesCheung96:
> what's the effect before and after this PR?

@lance6716  Removing the type prefix makes it easier to reuse the existing monitoring panel of the dm.

amyangfei (approved):

lance6716 (approved):

PR #7408, opened 2022-10-20
owner(ticdc): not use global resolvedTs as a part of barrier ts  if redo is not enabled

### LINKED ISSUE #7407: use ddl barrier as the global resolvedTs if redo is not enabled

sdojjy:

Issue Number: close #7407

global resolvedTs not be used as a part of barrier ts  if redo is not enabled

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


cdc/owner/changefeed.go
```diff
@@ -339,11 +339,18 @@ func (c *changefeed) tick(ctx cdcContext.Context, captures map[model.CaptureID]*
 		return nil
 	}
 
+	// if redo is not enabled, use the barrierTs as the `newResolvedTs`
+	// not need to wait the slowest table
+	if !c.redoManager.Enabled() {
+		newResolvedTs = barrierTs
+	}
+
```
nongfushanquan: global resolved ts and barrier ts have different meaning, so I suggest use two variables instead of  the one in the following codes

asddongmen (approved):

overvenus:
What about partition tables? What happens when a DDL affects two partitions?

nongfushanquan (approved):

PR #7427, opened 2022-10-21
dep(*): update dependency to update dumpling and gc tuner

lance6716:

Issue Number: ref https://github.com/pingcap/tiflow/issues/7428

#### Tests 

 - Manual test [TODO]

### Release note 

```release-note
`None`.
```


engine/executor/server.go
```diff
@@ -390,6 +392,15 @@ func (s *Server) isReadyToServe() bool {
 // Run drives server logic in independent background goroutines, and use error
 // group to collect errors.
 func (s *Server) Run(ctx context.Context) error {
+	if s.cfg.ExclusivelyDeployed {
+		limit, err := memory.MemTotal()
+		if err != nil {
+			log.L().Warn("get memory failed", zap.Error(err))
+		}
```
amyangfei, line 399: Is it better to reset `limit` to zero if err is not nil

engine/executor/server.go
```diff
@@ -390,6 +392,20 @@ func (s *Server) isReadyToServe() bool {
 // Run drives server logic in independent background goroutines, and use error
 // group to collect errors.
 func (s *Server) Run(ctx context.Context) error {
+	if s.cfg.ExclusivelyDeployed {
+		limit, err := memory.MemTotal()
+		if err != nil {
+			log.L().Warn("get memory failed", zap.Error(err))
```
CharlesCheung96, line 398:
```suggestion
			log.Warn("get memory failed", zap.Error(err))
```

engine/executor/server.go
```diff
@@ -390,6 +392,20 @@ func (s *Server) isReadyToServe() bool {
...
+			limit = 0
+		}
+		threshold := limit * 7 / 10
+		log.L().Info("set memory threshold to GC tuner",
```
CharlesCheung96, line 402:
```suggestion
		log.Info("set memory threshold to GC tuner",
```

deployments/engine/docker-compose/1m1e_with_s3.yaml
```diff
@@ -0,0 +1,104 @@
+version: "2.3"
```
CharlesCheung96, line 1: Why do we need this config file?

engine/executor/config.go
```diff
@@ -144,6 +147,7 @@ func GetDefaultExecutorConfig() *Config {
 		Join:                 defaultJoinAddr,
 		Addr:                 defaultExecutorAddr,
 		AdvertiseAddr:        "",
+		ExclusivelyDeployed:  true, // currently 1 container 1 executor
```
CharlesCheung96, line 150: Is this parameter used elsewhere?

deployments/engine/docker-compose/1m1e_with_s3.yaml
lance6716: it's my local benchmark config. Maybe other can also re-use it

CharlesCheung96 (approved):

lance6716:
ptal @GMHDBJD @amyangfei

engine/executor/config.go
```diff
@@ -50,6 +50,9 @@ type Config struct {
 	AdvertiseAddr string `toml:"advertise-addr" json:"advertise-addr"`
 
 	Labels map[string]string `toml:"labels" json:"labels"`
+	// ExclusivelyDeployed means only one executor is deployed in the container/machine,
+	// which means it can consume almost all the resources of the container/machine.
+	ExclusivelyDeployed bool `toml:"exclusively-deployed" json:"exclusively-deployed"`
```
amyangfei, line 55: What about using a more explicit config, such as `EnableGCTuning`, it is difficult to determine whether enable gc tuner is safe even under exclusively deployment.

PR #7436, opened 2022-10-24
 filter,mounter (ticdc): support bidireactional replication

asddongmen:

Issue Number: close #xxx

1. Add `ShouldIgnoreReplicationEvent ` for `filter` interface to filter out dml events written to upstream by other changefeeds if `ignore-rows-written-by-ticdc = true`.
2. Decode meta column in mounter and check if we should ignore a dml event if `ignore-rows-written-by-ticdc = true`.

**Note**: 
1. When downstream is TiDB, TiCDC will exctue `SET SESSION tidb_write_by_ticdc = true;` before replicating, which make every single row written by TiCDC has a hidden value `written_by_ticdc = 1`. This can help other changefeed to decide whether it should ignore a row event.
2. If the **upstream** TiDB does not support `tidb_write_by_ticdc` system variable and the user set `ignore-rows-written-by-ticdc = true` for a changefeed, the changefeed will **ignore** this error and runs normally, and will not ignore those rows written by old version TiDB.
3.  If the **downstrem** TiDB does not support `tidb_write_by_ticdc` system variable and the user set `ignore-rows-written-by-ticdc = true` for a changefeed, the changefeed will **ignore** this error and runs normally. This behavior is by design since filter only depends on upstream to work.

#### Tests 
- [ ] Unit test
- [ ] Integration test

### Release note 

```release-note

```


nongfushanquan (commented):
can we do a precheck in the downstream when  the  changefeed is configured with ignore-cdc-replication

cdc/entry/mounter.go
```diff
@@ -550,11 +591,7 @@ func getDDLDefaultDefinition(col *timodel.ColumnInfo) interface{} {
...
-// DecodeTableID decodes the raw key to a table ID
-func DecodeTableID(key []byte) (model.TableID, error) {
-	_, physicalTableID, err := decodeTableID(key)
-	if err != nil {
-		return 0, errors.Trace(err)
-	}
-	return physicalTableID, nil
+// checkMetaColumn return true if the meta column tidb_write_by_ticdc=1.
+func checkMetaColumn() bool {
+	return false
```
xiongjiwei, line 596: what is this

cdc/entry/codec.go
```diff
@@ -301,3 +309,11 @@ func unflatten(datum types.Datum, ft *types.FieldType, loc *time.Location) (type
 	}
 	return datum, nil
 }
+
+func decodeRowMeta(rowByte []byte, tz *time.Location) (types.Datum, error) {
+	datums, err := tablecodec.DecodeRowToDatumMap(rowByte, rowMetaMap, tz)
+	if err != nil {
+		return types.Datum{}, cerror.WrapError(cerror.ErrDecodeRowToDatum, err)
+	}
+	return datums[metaColID], nil
```
xiongjiwei, line 318: `datums[metaColID]` could be nil if the data does not contains meta column

cdc/entry/mounter.go
asddongmen: I will delete it.

cdc/entry/codec.go
asddongmen: fixed, PTAL

PR #7445, opened 2022-10-25
sinkv2(cdc): improve conflict detector throughput in some cases

### LINKED ISSUE #7446: sinkv2 conflict detector can makes some workers hungry
### What did you do?

We have done some tests. The result shows for some workloads, it's possible that conflict detector dispatches tasks to workers unfairly. Some workers are very busy but other workers are hungry. It reduces throughput and increases latency.

Conflict detector should work fine.

### What did you see instead?

图片
图片

### Versions of the cluster

Both 6.3 and master have this problem.

hicqu:

Issue Number: close #7446 .

Firstly let's take a look at how conflict detector handles some cases **without the patch**:

Suppose a transaction A with 2 conflict keys, x and y. If x doesn't conflict with any other transactions but y conflicts with another transaction B, conflict detector will think that A only conflicts with B, and dispatches A to B's worker directly. We thought **this is an enhancement**. But tests shows it will cause sink workers unbalance.

This PR changes the behavior. Let's get back to the case again. Now conflict detector will think A conflict with 2 other transactions, B and X. It's not important that who is B, and the key point is A won't be dispatched to B's worker directly. So hungry should never happend again.

#### Tests 

 - Integration test

### Release note 

```release-note
None
```


pkg/causality/internal/node.go
```diff
@@ -105,10 +105,21 @@ func (n *Node) NodeID() int64 {
...
-func (n *Node) DependOn(others map[int64]*Node) {
+func (n *Node) DependOn(unresolvedDeps map[int64]*Node, resolvedDeps int) {
 	resolvedDependees, removedDependees := int32(0), int32(0)
 
 	depend := func(target *Node) {
+		if target == nil {
+			// For a given Node, every dependency corresponds to a target.
+			// If target is nil it means the dependency doesn't conflict
+			// with any other nodes. However it's still necessary to track
+			// it because it makes effect on Node.tryResolve.
```
hi-rustin, line 116: Better to describe what kind of effect.

hi-rustin (approved):

asddongmen (approved):

PR #7456, opened 2022-10-26
checker(dm): prepare to check on both upstream and downstream table structures

lance6716:

Issue Number: ref #4287

the struct `{Schema string, Name string}` is passed to TablesChecker, so it can use downstream table of that struct to fetch table structure

#### Tests 

 - Pass old tests

### Release note 

```release-note
`None`.
```


lance6716:
ptal

GMHDBJD (approved):
LGTM

lichunzhu (approved):
LGTM

PR #7463, opened 2022-10-27
mq(ticdc): mq dml sink employ multiple workers to accelerate production rate, double the throughput

3AceShowHand:

Issue Number: close #xxx

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note

```

### DIFF
--- a/cdc/sink/codec/encoder_group.go
+++ b/cdc/sink/codec/encoder_group.go
@@ -1,0 +1,24 @@
+// Copyright 2022 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package codec
+
+type EncoderGroup interface {
+}
+
+type encoderGroup struct {
+}
+
+func NewEncoderGroup(builder EncoderBuilder) *encoderGroup {
+	return &encoderGroup{}
+}
--- a/cdc/sinkv2/eventsink/mq/worker.go
+++ b/cdc/sinkv2/eventsink/mq/worker.go
@@ -76,12 +76,13 @@ func newWorker(
 	protocol config.Protocol,
 	encoder codec.EventBatchEncoder,
 	producer dmlproducer.DMLProducer,
+	inputChan *chann.Chann[mqEvent],
 	statistics *metrics.Statistics,
 ) *worker {
 	w := &worker{
 		changeFeedID:                id,
 		protocol:                    protocol,
-		msgChan:                     chann.New[mqEvent](),
+		msgChan:                     inputChan,
 		ticker:                      time.NewTicker(flushInterval),
 		encoder:                     encoder,
 		producer:                    producer,
@@ -290,11 +291,11 @@ func (w *worker) asyncSend(
 }
 
 func (w *worker) close() {
-	w.msgChan.Close()
-	// We must finish consuming the data here,
-	// otherwise it will cause the channel to not close properly.
-	for range w.msgChan.Out() {
-		// Do nothing. We do not care about the data.
-	}
-	w.producer.Close()
+	//w.msgChan.Close()
+	//// We must finish consuming the data here,
+	//// otherwise it will cause the channel to not close properly.
+	//for range w.msgChan.Out() {
+	//	// Do nothing. We do not care about the data.
+	//}
+	//w.producer.Close()
 }
--- a/cdc/sink/codec/builder/encoder_builder.go
+++ b/cdc/sink/codec/builder/encoder_builder.go
@@ -10,6 +10,7 @@
 // distributed under the License is distributed on an "AS IS" BASIS,
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
 package builder
 
 import (
--- a/cdc/sinkv2/eventsink/mq/dmlproducer/kafka_dml_producer.go
+++ b/cdc/sinkv2/eventsink/mq/dmlproducer/kafka_dml_producer.go
@@ -120,13 +120,15 @@ func NewKafkaDMLProducer(
 			case <-ctx.Done():
 				return
 			case errCh <- err:
-				log.Error("Kafka DML producer run error", zap.Error(err),
+				log.Error("Kafka DML producer run error",
 					zap.String("namespace", k.id.Namespace),
-					zap.String("changefeed", k.id.ID))
+					zap.String("changefeed", k.id.ID),
+					zap.Error(err))
 			default:
-				log.Error("Error channel is full in kafka DML producer", zap.Error(err),
+				log.Error("Error channel is full in kafka DML producer",
 					zap.String("namespace", k.id.Namespace),
-					zap.String("changefeed", k.id.ID))
+					zap.String("changefeed", k.id.ID),
+					zap.Error(err))
 			}
 		}
 	}()
@@ -249,9 +251,10 @@ func (k *kafkaDMLProducer) run(ctx context.Context) error {
 			return nil
 		case err := <-k.failpointCh:
 			log.Warn("Receive from failpoint chan in kafka "+
-				"DML producer", zap.Error(err),
+				"DML producer",
 				zap.String("namespace", k.id.Namespace),
-				zap.String("changefeed", k.id.ID))
+				zap.String("changefeed", k.id.ID),
+				zap.Error(err))
 			return errors.Trace(err)
 		case ack := <-k.asyncProducer.Successes():
 			if ack != nil {
--- a/cdc/sinkv2/eventsink/mq/mq_dml_sink.go
+++ b/cdc/sinkv2/eventsink/mq/mq_dml_sink.go
@@ -20,7 +20,6 @@ import (
 	"github.com/pingcap/log"
 	"github.com/pingcap/tiflow/cdc/contextutil"
 	"github.com/pingcap/tiflow/cdc/model"
-	"github.com/pingcap/tiflow/cdc/sink/codec"
 	"github.com/pingcap/tiflow/cdc/sink/codec/builder"
 	"github.com/pingcap/tiflow/cdc/sink/codec/common"
 	mqv1 "github.com/pingcap/tiflow/cdc/sink/mq"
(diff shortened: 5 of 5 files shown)


cdc/sinkv2/metrics/mq/metrics.go
```diff
@@ -19,18 +19,28 @@ import (
...
+		prometheus.HistogramOpts{
+			Namespace: "ticdc",
+			Subsystem: "sinkv2",
+			Name:      "mq_worker_flush_duration",
+			Help: "Flush duration(s) for MQ worker. " +
+				"It contains the grouping, encoding, and asynchronous sending off a batch of data.",
+			Buckets: prometheus.ExponentialBuckets(0.001, 2, 20), // 1ms~1000s
+		}, []string{"namespace", "changefeed"})
+
+	WorkerDispatchedEventCount = prometheus.NewCounterVec(prometheus.CounterOpts{
```
3AceShowHand, line 33: todo: remove this

cdc/sink/codec/canal/canal_json_encoder.go
```diff
@@ -234,13 +234,14 @@ func (c *JSONBatchEncoder) AppendRowChangedEvent(
...
 // Build implements the EventJSONBatchEncoder interface
+// Build should be called immediately after call AppendRowChangedEvent
 func (c *JSONBatchEncoder) Build() []*common.Message {
 	if len(c.messages) == 0 {
 		return nil
 	}
 
 	result := c.messages
-	c.messages = c.messages[:0]
+	c.messages = nil
```
asddongmen, line 244: Please add a comment about why we need to set messages to nil here.

cdc/sinkv2/eventsink/mq/worker.go
```diff
@@ -289,6 +298,48 @@ func (w *worker) asyncSend(
...
+	}()
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.Trace(ctx.Err())
+		case <-ticker.C:
+			metric.Set(float64(len(inputCh)))
+		case future, ok := <-inputCh:
+			if !ok {
+				log.Warn("MQ sink encode response promise channel closed",
```
asddongmen, line 319: future

cdc/sinkv2/eventsink/mq/worker.go
```diff
@@ -289,6 +298,48 @@ func (w *worker) asyncSend(
 	return nil
 }
 
+func (w *worker) sendMessages(ctx context.Context) error {
+	inputCh := w.encoderGroup.Output()
```
asddongmen, line 302: Why call it `input` rather than `output`?

cdc/sinkv2/eventsink/mq/worker.go
```diff
@@ -289,6 +298,48 @@ func (w *worker) asyncSend(
...
+		codec.EncoderGroupResponseChanSizeGauge.
+			DeleteLabelValues(w.changeFeedID.Namespace, w.changeFeedID.ID)
+	}()
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.Trace(ctx.Err())
+		case <-ticker.C:
+			metric.Set(float64(len(inputCh)))
+		case future, ok := <-inputCh:
```
asddongmen, line 317: Will batch get event from this channel bring better performance?

cdc/sinkv2/eventsink/mq/worker.go
3AceShowHand: It's encoder group's output, but `sendMessages` input.

cdc/sinkv2/eventsink/mq/worker.go
3AceShowHand: batch can improve performance by reducing frequent function calls to avoid the cost of context switch.

If we want to batch multiple futures here to improve performance, the `AsyncSendMessage` method should also be changed to take a batch a messages as parameters.

cdc/sink/codec/metrics.go
```diff
@@ -0,0 +1,40 @@
...
+			Subsystem: "sink",
+			Name:      "encoder_group_input_chan_size",
+			Help:      "The size of input channel of encoder group",
+		}, []string{"namespace", "changefeed", "index"})
+	// EncoderGroupResponseChanSizeGauge tracks the size of response channel of encoder group
+	EncoderGroupResponseChanSizeGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "ticdc",
+			Subsystem: "sink",
+			Name:      "encoder_group_response_chan_size",
```
sdojjy, line 31: output_chan_size is better base on the name  `encoder_group_input_chan_size`

cdc/sink/codec/encoder_group.go
```diff
@@ -0,0 +1,171 @@
...
+	builder EncoderBuilder
+	count   int
+	inputCh []chan *future
+	index   uint64
+
+	outputCh chan *future
+}
+
+// NewEncoderGroup creates a new EncoderGroup instance
+func NewEncoderGroup(builder EncoderBuilder, number int, changefeedID model.ChangeFeedID) *encoderGroup {
```
sdojjy, line 57: `number` is not a good name

cdc/sink/codec/encoder_group.go
```diff
@@ -0,0 +1,171 @@
...
+		Topic:     topic,
+		Partition: partition,
+		Event:     event,
+		Callback:  callback,
+
+		done: make(chan struct{}),
+	}
+}
+
+// Ready waits until the response is ready, should be called before consume the future.
```
zhaoxinyu, line 163:
```suggestion
// Ready waits until the response is ready, should be called before consuming the future.
```

sdojjy (approved):

sdojjy:
please create an issue to enable the encoder group for batch encoders

cdc/sink/codec/encoder_group.go
```diff
@@ -0,0 +1,171 @@
...
+			if err := encoder.AppendRowChangedEvent(ctx, future.Topic, future.Event, future.Callback); err != nil {
+				return err
+			}
+			future.Messages = encoder.Build()
+			close(future.done)
+		}
+	}
+}
+
+func (g *encoderGroup) AddEvent(ctx context.Context, topic string, partition int32, event *model.RowChangedEvent, callback func()) error {
```
zhaoxinyu, line 119: The input parameter list is relatively long. Maybe this function definition can be replaced with 
```
func (g *encoderGroup) AddEvent(ctx context.Context, fu *future)
```

pkg/config/config_test_data.go
```diff
@@ -32,6 +32,7 @@ const (
     "worker-num": 3
   },
   "sink": {
+	"encoder-concurrency": 16,
```
zhaoxinyu, line 35: Is the indentation here consistent with the next line?

zhaoxinyu (approved):

pkg/config/config_test_data.go
3AceShowHand: yes, at the same level .

pkg/config/config_test_data.go
```diff
@@ -160,6 +161,7 @@ const (
...
```
hi-rustin, line 164: Please refine this format.

cdc/sink/codec/encoder_group.go
hi-rustin, line 119: I guess we can combine `topic string, partition int32, event *model.RowChangedEvent, callback func()`, because we copy it again and again.

hi-rustin (commented):
/hold

pkg/config/config_test_data.go
```diff
@@ -210,6 +212,7 @@ const (
...
```
hi-rustin, line 215: ditto

cdc/sinkv2/eventsink/mq/worker.go
```diff
@@ -59,8 +60,15 @@ type worker struct {
...
 	// ticker used to force flush the messages when the interval is reached.
 	ticker *time.Ticker
+
+	// builder is used to build encoders
+	builder codec.EncoderBuilder
 	// encoder is used to encode the messages.
 	encoder codec.EventBatchEncoder
+	// encoderGroup only support non batch encode protocol at the moment.
+	encoderGroup       codec.EncoderGroup
+	encoderConcurrency int
```
hi-rustin, line 70: Why do we need to keep it on struct? It seems needless because we already have the group.

pkg/config/replica_config.go
```diff
@@ -105,7 +106,7 @@ func (c *ReplicaConfig) Marshal() (string, error) {
 	return string(cfg), nil
 }
 
-// UnmarshalJSON unmarshals into *ReplicationConfig from json marshal byte slice
```
hi-rustin, line 108: I guess this grammar is correct.

pkg/config/sink.go
```diff
@@ -80,12 +80,13 @@ var ForceEnableOldValueProtocols = []string{
...
-	ColumnSelectors []*ColumnSelector `toml:"column-selectors" json:"column-selectors"`
-	SchemaRegistry  string            `toml:"schema-registry" json:"schema-registry"`
-	TxnAtomicity    AtomicityLevel    `toml:"transaction-atomicity" json:"transaction-atomicity"`
+	DispatchRules      []*DispatchRule   `toml:"dispatchers" json:"dispatchers"`
+	CSVConfig          *CSVConfig        `toml:"csv" json:"csv"`
+	Protocol           string            `toml:"protocol" json:"protocol"`
+	ColumnSelectors    []*ColumnSelector `toml:"column-selectors" json:"column-selectors"`
+	SchemaRegistry     string            `toml:"schema-registry" json:"schema-registry"`
+	TxnAtomicity       AtomicityLevel    `toml:"transaction-atomicity" json:"transaction-atomicity"`
+	EncoderConcurrency int               `toml:"encoder-concurrency" json:"encoder-concurrency"`
```
hi-rustin, line 89: Do we need to do some basic checks for this config? Like less than zero.

cdc/sink/codec/encoder_group.go
3AceShowHand: they were extracted from `mqEvent`. we cannot pass `mqEvent` directly since it introduces cyclic import problem.

If we move mqEvent to codec package, it becomes more complicated and weird

cdc/sinkv2/eventsink/mq/worker.go
3AceShowHand: `encoderGroup` is initialized in the `run` method only for non-batch protocol at the moment.

`encoderConcurrency` can be passed to the `run` method explicitly or by the current way.

I will make encoder group support batched protocols, after that we can initialize the encoder group in the `newWorker` method, after that we can just keep one `encoderBuilder` field in the worker struct, and remove others.

pkg/config/replica_config.go
3AceShowHand: yes, it's correct grammer, it was altered by IDE typo detector.

3AceShowHand:
* 300 sysbench tables, throughput goes up to more than 200k/s
image

cdc/sink/codec/encoder_group.go
hi-rustin: In any case, it is not good to copy the same relevant parameters four times :(

Issue #7380, opened 2022-10-17
engine docker-composition configuration needs to be optimized


CharlesCheung96:
1. It is necessary to to determine the restart strategy of the node. Restarting the node when an error occurs can improve the stability of the test, but at the same time, some errors [may be ignored](https://github.com/pingcap/tiflow/pull/7336#discussion_r996577491).
2. Some configuration parameters need to be simplified. Ref: [Replace the configuration with the default parameters](https://github.com/pingcap/tiflow/pull/7336#discussion_r996578290)

PR #7466, opened 2022-10-27
ci(engine): update restart policy in docker compose tpl

amyangfei:

Issue Number: ref #7380

- Change docker compose restart policy to `unless-stopped`
- Check whether panic happens in master and executor logs, if it does, fail the test directly.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


engine/test/utils/check_log.sh
```diff
@@ -1,3 +1,12 @@
 #!/bin/bash
 
-grep "Run all test success" /tmp/tiflow_engine_test/engine_it.log && exit 0 || exit 1
+grep "Run all test success" /tmp/tiflow_engine_test/engine_it.log || exit 1
+
+grep -rE "\[PANIC\]|panic:" /tmp/tiflow_engine_test
+
+if [ $? -eq 0 ]; then
```
lance6716, line 7: should be not equal 0 for "not found panic"?

lance6716 (approved):

deployments/engine/docker-compose/1m1e_with_s3.yaml
```diff
@@ -42,7 +42,6 @@ services:
       - "10241:10241"
     depends_on:
       - "server-master"
-    restart: "on-failure"
```
CharlesCheung96, line 45: Why not use `restart: unless-stopped` policy here?

amyangfei: L49 has one.

CharlesCheung96 (approved):

Issue #7465, opened 2022-10-27
Job reponse should hide sensitive information


gozssky:
Config in job struct may contain sensitive information such as plain password and tls key content. We shouldn't always return them to clients in every job api.
https://github.com/pingcap/tiflow/blob/6baa8bb9c84c92b2fbea1232b9dc22ec1d11be19/engine/proto/master.proto#L233
Since config are store as raw bytes, there is no perfect way to hide sensitive information exactly. As a workaround, we can add a parameter to control whether to return config or not.

gozssky:
cc @D3Hunter

PR #7471, opened 2022-10-28
openapi(engine): ignore config in job response by default

gozssky:

Issue Number: close https://github.com/pingcap/tiflow/issues/7465

Config in job struct may contain sensitive information such as plain password and tls key content.

Add a parameter to control whether to return config or not. By default, the job response won't include config content.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


lance6716:
> Add a parameter to control whether to return config or not

what's our expected behaviour for this security problem? if we mistakenly configurated the parameter the password is still printed

gozssky:
> what's our expected behaviour for this security problem? if we mistakenly configurated the parameter the password is still printed

@lance6716 In most cases, I think we don't need config from the client side, since it is from the client side when the job is first created. By default, the config will not return to the clients, so the parameter should always not be configurated. If a client really needs the config, he can add the parameter and be responsible for the security issues.

lance6716:
> #7465

I can't get why this is a security problem, maybe we don't want to leak the password to client? If the client to OpenAPI is all authenticated we don't need to worry about it. If we should not trust the client, we should also not let client control itself to read the password.

gozssky:
@lance6716 It is not related to authentication. We can trust the clients.
There are two concerns here:
1. Config is not used often after job is created.
2. All job apis except DeleteJob will return config in response, clients may not handle it carefully.

lance6716 (approved):

amyangfei (approved):

PR #7472, opened 2022-10-28
capture(cdc): fix cdc `NewSession` stuck when send SIGSTOP to pd leader

### LINKED ISSUE #7470: cdc maybe stuck when send SIGSTOP to pd leader
### What did you do?

1. tiup start one cluster with two cdc & three cdc
2. `kill -19` to pd leader
3. `cdc cli capture list` can't list all capture, if `cdc cli capture list` sucessfully, repeat step 2.

`cdc cli capture list` list all capture

### What did you see instead?

`cdc cli capture list` can't list all capture

### Versions of the cluster

```console
v6.3.0
```

```console
v6.3.0 
```

```console
v6.3.0
```

zeminzhou:
Signed-off-by: zeminzhou <zhouzemin@pingcap.com>

fix cdc stuck when send SIGSTOP to pd leader

Issue Number: close #7470

1. get lease id by `Grant` before `NewSession` instead of using to `NewSesion` get lease id, so we can add timeout when call `Grant`
2. use `WithRequireLeader`

#### Tests 

 - Unit test
 - Integration test

NO
NO
### Release note 

```release-note
`None`
```


overvenus (commented):
What happens when upgrading? Will there be two owners at the same time?

zeminzhou:
> What happens when upgrading? Will there be two owners at the same time?
I think there will not be wo owners at the same time. 
1. In the first modification, I just forwarded the `Grant` in `NewSession` to the outside, so that we could set timeout the `grant` by ourselves instead of going through etcd. Therefore, when the pd leader enters the STOP state, we can ensure that this place will not get stuck by setting a timeout.
2. Below the second modification, `WithRequireLeader` only requires that the `watch` operation requesting server must be a `leader`. If a network partition occurs, the current `leader` finds that it is no longer a `leader`, and will put this `watch stream` closure.

overvenus (approved):

3AceShowHand (approved):

zeminzhou:
From debug, `WithRequreLeader` is useless for `campagin` stuck. I add TODO first.

zeminzhou:
Please PTAL~, again.

PR #7474, opened 2022-10-28
sinkv2(cdc): implement EventSortEngine based on pebble

hicqu:

Issue Number: ref #5928 

Add an implementation of EventSortEngine based on Pebble.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


pkg/sorter/pebble/db.go
```diff
@@ -0,0 +1,150 @@
...
+	userProps[maxTableCRTsLabel] = fmt.Sprintf("%d", t.maxTs)
+	return nil
+}
+
+func (t *tableCRTsCollector) Name() string {
+	return tableCRTsCollectorName
+}
+
+// NOTE: both lowerBound and upperBound are included.
+func iterTable(
```
nongfushanquan, line 66: why are both bound included instead of one bound?

pkg/sorter/pebble/event_sorter.go
```diff
@@ -0,0 +1,404 @@
...
+		if _, err = s.serde.Unmarshal(event, value); err != nil {
+			return
+		}
+		if s.headItem != nil {
+			break
+		}
+		s.headItem, event = event, nil
+	}
+	if s.headItem != nil {
+		if event == nil || s.headItem.CRTs != event.CRTs || s.headItem.StartTs != event.StartTs {
```
nongfushanquan, line 256: if event != nil  & s.headItem.CRTs == event.CTRs & s.headItem.StartTs == event.StartTs ,  what is the value of pos?

hicqu: In this case `pos` won't be updated. `pos` will only be filled when meets a finished transaction.

pkg/sorter/pebble/db.go
hicqu: Generally `upperBound` comes from `resolvedTs` or `barrierTs`, which are included.
`lowerBound` can be included or excluded, I think both are OK.

nongfushanquan (approved):

pkg/sorter/pebble/event_sorter.go
```diff
@@ -0,0 +1,404 @@
...
+		if err := s.cleanTable(state, tableID); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Next implements sorter.EventIterator.
+func (s *EventIter) Next() (event *model.PolymorphicEvent, pos sorter.Position, err error) {
+	var valid bool = s.iter != nil && s.iter.Valid()
```
hi-rustin, line 242:
```suggestion
	bool := s.iter != nil && s.iter.Valid()
```

pkg/sorter/pebble/event_sorter.go
```diff
@@ -0,0 +1,404 @@
...
+	var value []byte = nil
```
hi-rustin, line 243:
```suggestion
	var value []byte
```

hi-rustin (approved):
rest LGTM

pkg/sorter/manager/pebble.go
```diff
@@ -0,0 +1,83 @@
...
+	return dbs, writeStalls, nil
+}
+
+type pebbleLogger struct{ id int }
+
+var _ pebble.Logger = (*pebbleLogger)(nil)
+
+func (logger *pebbleLogger) Infof(format string, args ...interface{}) {
+	// Do not output low-level pebble log to TiCDC log.
+	log.Debug(fmt.Sprintf(format, args...), zap.Int("db", logger.id))
```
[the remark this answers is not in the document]
hicqu: DB ID can't be passed into the log if use `Debugf`.

hi-rustin:
@hicqu Your code has some lint issues.

PR #7484, opened 2022-10-29
sinkmanager(ticdc): impl worker and mem quota

hi-rustin:

Split from https://github.com/pingcap/tiflow/pull/7388.

- Impl worker to fetch data.
- Impl mem quota to control the mem usage. 

#### Tests 

 - [x] Unit test

No
No
### Release note 

```release-note
None
```


hi-rustin (commented):
Not ready for review. Please do not review this PR now. Thanks!

cdc/processor/sinkmanager/worker_impl.go
```diff
@@ -0,0 +1,236 @@
...
+			var lastPos sorter.Position
+			currentTotalSize := uint64(0)
+			batchID := uint64(1)
+			batchGroupCount := uint64(1)
+			// We have to get the latest barrier ts, this is because the barrier ts may be updated.
+			// We always want to get the latest value.
+			currentBarrierTs := task.lastBarrierTs.Load()
+			upperBound := sorter.Position{
+				CommitTs: currentBarrierTs - 1,
+				StartTs:  currentBarrierTs,
```
nongfushanquan, line 98: why start ts is less than commits

cdc/processor/sinkmanager/worker.go
```diff
@@ -0,0 +1,45 @@
...
+// It only considers how to control the table sink.
+type tableSinkTask struct {
+	tableID model.TableID
+	// lowerBound indicates the lower bound of the table sink.
+	// It is a closed interval.
+	lowerBound sorter.Position
+	// lastBarrierTs indicates the upper bound of the table sink.
+	// It is a closed interval.
+	// Use atomic to get the latest value.
+	lastBarrierTs *atomic.Uint64
```
nongfushanquan, line 37: why not upper bound instead of lastBarrierTs  here?

cdc/processor/sinkmanager/worker_impl.go
hi-rustin: Oops! This is a mistake.

cdc/processor/sinkmanager/worker.go
hi-rustin: I can change it. But we still need to use a pointer, not a value because we want the latest value.

cdc/processor/sinkmanager/mem_quota.go
```diff
@@ -0,0 +1,149 @@
...
+			return nil
+		}
+		m.blockAcquireCond.Wait()
+	}
+}
+
+func (m *memQuota) refund(nBytes uint64) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.usedBytes -= nBytes
```
hicqu, line 97: Should call `m.blockAcquireCond.Signal()`?

cdc/processor/sinkmanager/worker_impl.go
```diff
@@ -0,0 +1,215 @@
...
+							return errors.Trace(err)
+						}
+					}
+					break
+				}
+				for availableMem-e.Row.ApproximateBytes() < 0 {
+					if !w.splitTxn {
+						w.memQuota.forceAcquire(defaultRequestMemSize)
+					} else {
+						err := w.memQuota.blockAcquire(defaultRequestMemSize)
```
hicqu, line 137: How about doing a flush before this line?

hicqu:
Rest LGTM.

cdc/processor/sinkmanager/mem_quota.go
hi-rustin: Good catch!

cdc/processor/sinkmanager/worker_impl.go
hi-rustin: I think we should not flush it. Because we only if the split txn enabled, we only flush it when we reach a batch(50MB).

hicqu (approved):

cdc/processor/sinkmanager/worker_impl.go
```diff
@@ -0,0 +1,221 @@
...
+					}
+					// If no more available memory, we should put the table
+					// back to the SinkManager and wait for the next round.
+					if !w.memQuota.hasAvailable(defaultRequestMemSize) {
+						break
+					}
+				} else {
+					if w.splitTxn {
+						// If we enable splitTxn, we should emit the events to the sink when the batch size is exceeded.
+						if currentTotalSize >= maxUpdateIntervalSize*50 {
```
nongfushanquan, line 173: I think 50M is too big for the flushing-unit

nongfushanquan (approved):

Issue #7237, opened 2022-09-27
discovery agent fails to get snapshot during master starts up


amyangfei:
### What did you do?

- Setup one tiflow master and one executor
- Create a long running job
- Restart tiflow master

Given the following diff

```diff
apple in ~/code/tiflow on master ● λ git diff

--- a/engine/executor/server.go
+++ b/engine/executor/server.go
@@ -483,7 +483,7 @@ func (s *Server) Run(ctx context.Context) error {
                }

                for _, node := range snap {
-                       log.Debug("update p2p msg router by snapshot", zap.Any("node", node))
+                       log.Info("update p2p msg router by snapshot", zap.Any("node", node))
                        s.p2pMsgRouter.AddPeer(node.ID, node.Addr)
                }

@@ -495,7 +495,7 @@ func (s *Server) Run(ctx context.Context) error {
                        case event = <-receiver.C:
                        }

-                       log.Debug("update p2p msg router", zap.Any("event", event))
+                       log.Info("update p2p msg router", zap.Any("event", event))
                        if event.Tp == discovery.EventTypeDel {
                                s.p2pMsgRouter.RemovePeer(event.Node.ID)
                        } else if event.Tp == discovery.EventTypeAdd {
```

- Observe the discovery agent behavior

### What did you see instead?

`update p2p msg router` with new tiflow master is triggered after leader is ready

```go
[2022/09/27 22:36:48.631 +08:00] [INFO] [elector.go:302] ["new leader elected"] [leader-id=master-127.0.0.1:10240-fc6f9d92] [leader-name=master-127.0.0.1:10240] [leader-address=127.0.0.1:10240]

[2022/09/27 22:37:09.049 +08:00] [INFO] [server.go:939] ["leader is initialized"] [took=20.417796252s]
```

tiflow master leader is initialized slowly because it is waiting for all persisted workers' heartbeat or timeout, but existing job master on executors are not able to aware new tiflow masters via `getSnapshot` of discovery agent.

```go
[2022/09/27 22:37:10.397 +08:00] [INFO] [server.go:498] ["update p2p msg router"] [event="{\"Tp\":\"add\",\"Node\":{\"Tp\":\"master\",\"ID\":\"master-127.0.0.1:10240-fc6f9d92\",\
"Addr\":\"127.0.0.1:10240\"}}"]
[...]

PR #7502, opened 2022-10-31
servermaster(engine): support API level availability check

amyangfei:

Issue Number: ref #7237

- Introduce a new interface `FeatureChecker` in `rpcutil` package, which can check whether an API is available by name.
- Implement the `FeatureChecker` in servermaster as a service degrader, separate the executor management and master worker framework management as two independent services, in this way some APIs can be serving independently. Such as when worker manger is not ready, server master can still serve `ListExecutors` API if executor manager is ready.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


gozssky (approved):

engine/servermaster/server.go
```diff
@@ -814,21 +801,44 @@ func (s *Server) runLeaderService(ctx context.Context) (err error) {
...
 	errg.Go(func() error {
-		return s.gcCoordinator.Run(errgCtx)
+		updater := executorInfoUpdater{
+			msgRouter:     s.p2pMsgRouter,
+			executorGroup: executorClients,
+		}
+		return serverutil.WatchExecutors(errgCtx, s.executorManager, updater)
 	})
 
+	s.leaderDegrader.updateExecutorManager(true)
```
CharlesCheung96, line 819: Whether this field should be updated inside `s.executorManager.Run`？

engine/servermaster/service_util.go
```diff
@@ -56,3 +57,41 @@ func generateNodeID(name string) string {
 	id := fmt.Sprintf("%s-%08x", name, val)
 	return id
 }
+
```
CharlesCheung96, line 60:
```suggestion
var _ rpcutil.FeatureChecker = &featureDegrader{}
```
Besides, perhaps leaderFeatureChecker would be a better name?

engine/servermaster/server.go
amyangfei: Not necessary, `executorManager` can serve after it is created. API of `executorManager` can run concurrently against logic in `Run`

CharlesCheung96 (approved):

engine/servermaster/service_util.go
amyangfei: I use `featureDegrader` here because we may add more downgrade policy in server master, `rpcutil.FeatureChecker` is part of them.

amyangfei:
wait chaos test.

PR #7523, opened 2022-11-03
ShardDDL(engine/dm): implement shard ddl in dm jobmaster

GMHDBJD:

Issue Number: ref #5504 

- add ddl coordinator to coordinate DDL

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


dm/pkg/conn/basedb.go
```diff
@@ -249,7 +259,7 @@ func CloseBaseConnWithoutErr(d *BaseDB, conn *BaseConn) {
 
 // Close release *BaseDB resource.
 func (d *BaseDB) Close() error {
-	if d == nil || d.DB == nil {
+	if d == nil || d.DB == nil || d.doNotClose {
```
niubell, line 262: Why just don't call `Close` in unit tests?

GMHDBJD: In the unit test we need to mock the db operation, if it is closed, go-sql-mock will report the error that database is already closed. So we don't close here, and let the person who writing the test decide when to close

lance6716: Will the `mock.ExpectClose()` help?

engine/jobmaster/dm/api.go
```diff
@@ -46,6 +46,23 @@ type JobStatus struct {
 	FinishedUnitStatus map[string][]*metadata.FinishedTaskStatus `json:"finished_unit_status,omitempty"`
 }
 
+// ShardTable represents create table statements of a source table.
+type ShardTable struct {
+	Current string
+	Pending string
+}
```
lance6716, line 53: When I read it I can't make sure `Pending` is a DDL or CREATE TABLE. Maybe `Current` and `Next`.
And the type name `ShardTable` can be better reflect it's a chaning, has the previous and post data, not a static table data. If the type name is showed in API we may ask PM's opinion

engine/jobmaster/dm/checkpoint/agent.go
```diff
@@ -156,6 +169,99 @@ func (c *AgentImpl) UpgradeFuncs() []bootstrap.UpgradeFunc {
...
+	}
+	var ti *model.TableInfo
+	if bytes.Equal(tiBytes, []byte("null")) {
+		return "", errors.Errorf("table info not found %v", sourceTable)
+	}
+	if err := json.Unmarshal(tiBytes, &ti); err != nil {
+		return "", err
+	}
+	t := schemacmp.Encode(ti)
+	return t.String(), nil
```
lance6716, line 262: this is not the real table structure (ConstructResultOfShowCreateTable). Can you add a test to show the difference so reviewer can better understand?

lance6716 (commented):
2 / 14 files viewed

engine/jobmaster/dm/metadata/ddl.go
```diff
@@ -14,35 +14,195 @@
...
+type SourceTable struct {
+	Source string
+	Schema string
+	Table  string
+}
+
+// MarshalText implements json.MarshalText.
+func (st SourceTable) MarshalText() (text []byte, err error) {
+	type s SourceTable
+	return json.Marshal(s(st))
```
lance6716, line 36: Can we use

```suggestion
	return json.Marshal(st)
```

or remove this function to use default marshaller?

engine/jobmaster/dm/metadata/ddl.go
```diff
@@ -14,35 +14,195 @@
...
+		if _, ok := tbs[sourceTable]; ok {
+			return true
+		}
+	}
+	return false
+}
+
+// AddDroppedColumns adds dropped columns to the state
+func (s *DroppedColumnsStore) AddDroppedColumns(ctx context.Context, cols []string, sourceTable SourceTable) error {
+	s.mu.Lock()
```
lance6716, line 135: can implement something like `(f *UnitStateStore) ReadModifyWrite` to avoid the repeated get/release lock of below functions

engine/jobmaster/dm/dm_jobmaster.go
```diff
@@ -117,6 +118,7 @@ func (jm *JobMaster) initComponents() error {
 	jm.taskManager = NewTaskManager(taskStatus, jm.metadata.JobStore(), jm.messageAgent, jm.Logger(), jm.MetricFactory())
 	jm.workerManager = NewWorkerManager(jm.ID(), workerStatus, jm.metadata.JobStore(), jm.metadata.UnitStateStore(),
 		jm, jm.messageAgent, jm.checkpointAgent, jm.Logger(), jm.IsS3StorageEnabled())
+	jm.ddlCoordinator = NewDDLCoordinator(jm.ID(), jm.MetaKVClient(), jm.checkpointAgent, jm.metadata.JobStore(), jm.Logger())
```
lance6716, line 121: Can we not enable this member if job is using sharding mode? To reduce the chance we meet error.

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,832 @@
...
+			c.tables[targetTable][sourceTable] = struct{}{}
+		}
+	}
+	return nil
+}
+
+// Coordinate coordinates ddls.
+func (c *DDLCoordinator) Coordinate(ctx context.Context, item *metadata.DDLItem) ([]string, optimism.ConflictStage, error) {
+	c.mu.Lock()
+	c.pendings.Add(1)
```
lance6716, line 207: the problem is the race accessing of c.pendings. Even if we use lock to avoid simultaneously accessing, will it happen that `Coordinate` is called after `Clear`?

dm/pkg/conn/basedb.go
GMHDBJD: ExpectClose just mock the close result, the database will still be closed

engine/jobmaster/dm/metadata/ddl.go
GMHDBJD: No, we want to use SourceTable as key https://github.com/pingcap/tiflow/blob/7521c5107d8a5f551520a9b7d2afcbca0ded9606/engine/jobmaster/dm/metadata/ddl.go#L85
From [json doc](https://pkg.go.dev/encoding/json#Marshal), `Map values encode as JSON objects. The map's key type must either be a string, an integer type, or implement encoding.TextMarshaler.`

engine/jobmaster/dm/metadata/ddl.go
GMHDBJD: Some behaviors are not easy to combine. For example, if the state is not found, AddDroppedColumn creates a new one, DeleteDropColumn will do noting.

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: Clear only can be called when cancel job, in that time all the worker should be closed, or the jobmaster will be forcely close

engine/jobmaster/dm/checkpoint/agent.go
```diff
@@ -156,6 +169,99 @@ func (c *AgentImpl) UpgradeFuncs() []bootstrap.UpgradeFunc {
...
+		return "", err
+	}
+	defer db.Close()
+
+	query := `SELECT table_info FROM ` + syncTableName(jobID, cfg) + ` WHERE id = ? AND cp_schema = ? AND cp_table = ?`
+	row := db.DB.QueryRowContext(ctx, query, sourceTable.Source, sourceTable.Schema, sourceTable.Table)
+	if row.Err() != nil {
+		return "", row.Err()
+	}
+	var tiBytes []byte
```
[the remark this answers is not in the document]
GMHDBJD:
```
sql: RawBytes isn't allowed on Row.Scan
```

engine/jobmaster/dm/dm_jobmaster.go
GMHDBJD: still enable it, but bypassing in the internal 
00f0094

engine/jobmaster/dm/checkpoint/agent.go
```diff
@@ -156,6 +168,99 @@ func (c *AgentImpl) UpgradeFuncs() []bootstrap.UpgradeFunc {
...
+			return "", errors.Errorf("table info not found %v", sourceTable)
+		}
+		return "", err
+	}
+	var ti *model.TableInfo
+	if bytes.Equal(tiBytes, []byte("null")) {
+		return "", errors.Errorf("table info not found %v", sourceTable)
+	}
+	if err := json.Unmarshal(tiBytes, &ti); err != nil {
+		return "", err
```
lance6716, line 258: add a trace 😂 or something that we can quickly identify the error

engine/jobmaster/dm/checkpoint/agent_test.go
```diff
@@ -269,3 +274,38 @@ func TestIsFresh(t *testing.T) {
...
+
+	tbInfo, err := tracker.GetTableInfo(&filter.Table{Schema: database, Name: table})
+	require.NoError(t, err)
+	bs, err := json.Marshal(tbInfo)
+	require.NoError(t, err)
+
+	mockDB.ExpectQuery(regexp.QuoteMeta("SELECT table_info FROM `meta`.`job_id_syncer_checkpoint` WHERE id = ? AND cp_schema = ? AND cp_table = ?")).WithArgs(source, database, table).WillReturnRows(sqlmock.NewRows([]string{"table_info"}).AddRow(bs))
+	tbStmt, err := agent.FetchTableStmt(context.Background(), "job_id", &config.JobCfg{MetaSchema: "meta"}, metadata.SourceTable{Source: source, Schema: database, Table: table})
+	require.NoError(t, err)
+	require.Equal(t, "CREATE TABLE `tbl`(`id` INT(11)) CHARSET LATIN1 COLLATE LATIN1_BIN", tbStmt)
```
lance6716, line 310: if this content is displayed to user, we should make sure it does not missing noticeable elements (like missing comments?) which makes user confused about the displayed table is at which history version.

And I noticed that the Table.Restore in Table.String has comment "for debug use only", maybe it will not generate a legal CREATE TABLE statement in some case, I prefer we use `ConstructResultOfShowCreateTable`

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,841 @@
...
+		kvClient:    kvClient,
+		shardGroups: make(map[metadata.TargetTable]*shardGroup),
+		jobStore:    jobStore,
+		logger:      pLogger.With(zap.String("component", "ddl_coordinator")),
+	}
+}
+
+// Clear clears metadata.
+func (c *DDLCoordinator) Clear(ctx context.Context) error {
+	c.mu.Lock()
```
lance6716, line 170: this locking is in fact not needed because as the lifetime of jobmaster, Coordinate will not be called concurrent with Clear/Reset?

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,841 @@
...
+		if err := g.clear(ctx); err != nil {
+			c.logger.Error("clear shard group failed", zap.Error(err))
+		}
+	}
+	delete(c.shardGroups, targetTable)
+}
+
+type shardGroup struct {
+	mu                  sync.RWMutex
+	normalTables        map[metadata.SourceTable]string
```
lance6716, line 303: please left some comment about the meaning of `string` value and what's a normal/conflict table

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,841 @@
...
+	conflictTables      map[metadata.SourceTable]string
+	tableAgent          TableAgent
+	droppedColumnsStore *metadata.DroppedColumnsStore
+	id                  frameModel.MasterID
+	cfg                 *config.JobCfg
+	deleted             bool
```
lance6716, line 309: why we need it? will empty `shardGroup` be deleted by DDLCoordinator.removeShardGroup?

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,841 @@
...
+	}
+	for sourceTable := range sourceTables {
+		stmt, err := g.tableAgent.FetchTableStmt(ctx, id, cfg, sourceTable)
+		// NOTE: There are cases where the source table exists but the table info does not,
+		// such as a table is created upstream when the coordinator starts, but the dm-worker is not yet synchronized
+		// we skip these tables in handleDDL.
+		// TODO: better error handling
+		if err != nil && !strings.HasPrefix(err.Error(), "table info not found") {
+			return nil, err
+		}
```
lance6716, line 329: should we continue the rest of tables?

engine/jobmaster/dm/checkpoint/agent.go
```diff
@@ -156,6 +168,99 @@ func (c *AgentImpl) UpgradeFuncs() []bootstrap.UpgradeFunc {
...
+			}
+			return nil
+		})
+
+	}
+	err := g.Wait()
+	return result, err
+}
+
+// FetchTableStmt fetch create table statement from checkpoint.
```
lance6716, line 231: the comments can be moved or repeated to interface, so mouse hovering on invocations will see it.

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,841 @@
...
+	case metadata.DropTable:
+		needDeleted = g.handleDropTable(ctx, item)
+		ddls = append(ddls, item.DDLs...)
+	case metadata.OtherDDL:
+		ddls, conflictStage, err = g.handleDDLs(ctx, item)
+	default:
+		return nil, optimism.ConflictError, false, errors.Errorf("unknown ddl type %v", item.Type)
+	}
+
+	return ddls, conflictStage, needDeleted || g.isResolved(ctx), err
```
lance6716, line 367: what's the definition of resolved? without knowing it developers can't understand why it can be combined with needDeleted?

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,841 @@
...
+}
+
+// handleCreateTable handles create table ddl.
+// add new source table to shard group.
+func (g *shardGroup) handleCreateTable(item *metadata.DDLItem) {
+	stmt, ok := g.normalTables[item.SourceTable]
+	if ok && stmt != "" {
+		log.L().Warn("create table already exists", zap.Any("source table", item.SourceTable))
```
lance6716, line 375: seems these logging are useful when debugging production problems (or make then DPanic level), please use the framework logger with job_id attached

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: // if all source table is deleted, we should remove the shard group.

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: we just return the error. when new ddl comes, we will try to load tables again

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: I do it in https://github.com/pingcap/tiflow/pull/7858

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,839 @@
...
+		if err := g.droppedColumnsStore.AddDroppedColumns(ctx, dropCols, item.SourceTable); err != nil {
+			return nil, optimism.ConflictError, err
+		}
+	}
+
+	return newDDLs, optimism.ConflictNone, nil
+}
+
+func (g *shardGroup) handleDDL(sourceTable metadata.SourceTable, prevTableStmt, postTableStmt string) (bool, optimism.ConflictStage) {
+	delete(g.conflictTables, sourceTable)
```
lance6716, line 448: why we delete it?

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,839 @@
...
+
+	currTable := genCmpTable(g.normalTables[sourceTable])
+	// handle idempotent ddl
+	idempotent := false
+	if cmp, err := prevTable.Compare(currTable); err != nil || cmp != 0 {
+		if cmp, err := postTable.Compare(currTable); err == nil && cmp == 0 {
+			idempotent = true
+		}
+		log.L().Warn("prev-table not equal table saved in master", zap.Stringer("master-table", currTable), zap.Stringer("prev-table", prevTable))
+		g.normalTables[sourceTable] = prevTableStmt
```
lance6716, line 461: the definition of normalTables is table structure at checkpoint? why overwrite it to an old value

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,839 @@
...
+//   ├────┐                       │                              │                                │
+//   │    │                       │                              │                                │
+//   │    │ Redirect              │                              │                                │
+//   │    │                       │                              │                                │
+//   │◄───┘                       │                              │                                │
+//   │                            │                              │                                │
+
+type tableType int
+
+const (
```
lance6716, line 130: what's the meaning of these type?

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,839 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	// wait all pending ddls coordinated.
+	c.pendings.Wait()
+
+	return metadata.DelAllDroppedColumns(ctx, c.kvClient)
+}
+
+// Reset resets the shard group.
+func (c *DDLCoordinator) Reset(ctx context.Context) error {
```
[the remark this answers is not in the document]
GMHDBJD: I think Reset makes more sense. It reset the ddl coordinator, include tables and shard groups, I update the comment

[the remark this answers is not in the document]
lance6716: My old change request comes from that Clear and Reset looks similar

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,839 @@
...
+	var (
+		ddls          []string
+		needDeleted   bool
+		err           error
+		conflictStage optimism.ConflictStage = optimism.ConflictNone
+	)
+
+	switch item.Type {
+	case metadata.CreateTable:
+		g.handleCreateTable(item)
```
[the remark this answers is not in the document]
GMHDBJD: Ignore just like the original logic.

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,843 @@
...
+	p, _ := conn.GetParserFromSQLModeStr(mysql.DefaultSQLMode)
+	stmtNode, _ := p.ParseOneStmt(createStmt, "", "")
+	ti, _ := ddl.BuildTableInfoFromAST(stmtNode.(*ast.CreateTableStmt))
+	ti.State = model.StatePublic
+	return schemacmp.Encode(ti)
+}
+
+// getColumnNames and return columns' names for create table stmt.
+func getColumnNames(createStmt string) []string {
+	p, _ := conn.GetParserFromSQLModeStr(mysql.DefaultSQLMode)
```
lance6716, line 834: Can directly use `parser.New()`?

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,843 @@
...
+	stmtNode, _ := p.ParseOneStmt(createStmt, "", "")
```
lance6716, line 835: not sure when upstream has unsupported table structure it will cause jobmaster panic

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,843 @@
...
+			Current: stmt,
+			Next:    g.conflictTables[sourceTable],
+		}
+	}
+	// show dropped columns if needed.
+	return tables
+}
+
+func genCmpTable(createStmt string) schemacmp.Table {
+	p, _ := conn.GetParserFromSQLModeStr(mysql.DefaultSQLMode)
```
lance6716, line 825: ditto

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,843 @@
...
+
+// isResoved means all tables in the group are resolved.
+// 1. no conflict ddls waiting
+// 2. all dropped column has done
+// 3. all shard tables stmts are same.
+func (g *shardGroup) isResolved(ctx context.Context) bool {
+	if len(g.conflictTables) != 0 {
+		return false
+	}
+	if _, err := g.droppedColumnsStore.Get(ctx); err == nil || errors.Cause(err) != metadata.ErrStateNotFound {
```
lance6716, line 783:
```suggestion
	if _, err := g.droppedColumnsStore.Get(ctx); errors.Cause(err) != metadata.ErrStateNotFound {
```

seems `errors.Cause` is clear enough

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,843 @@
...
+	var previousTable string
+	for _, tbStmt := range g.normalTables {
+		if tbStmt == "" {
+			continue
+		}
+		if previousTable == "" {
+			previousTable = tbStmt
+			continue
+		}
+		lhs := genCmpTable(previousTable)
```
lance6716, line 796: seems previousTable does not need to be changed after initialized and we can save half `genCmpTable`

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,843 @@
...
+	state, err := g.droppedColumnsStore.Get(ctx)
+	if err != nil {
+		if errors.Cause(err) == metadata.ErrStateNotFound {
+			return nil
+		}
+		return err
+	}
+
+	droppedColumns := state.(*metadata.DroppedColumns)
+	cacheStmts := make(map[metadata.SourceTable]string)
```
lance6716, line 740: no write to cache?

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,843 @@
...
+		for _, tbStmt := range g.normalTables {
+			if tbStmt == "" {
+				continue
+			}
+			if cols := getColumnNames(tbStmt); slices.Contains(cols, col) {
+				continue OutLoop
+			}
+		}
+
+		for sourceTable := range g.normalTables {
```
lance6716, line 752: why there're two loops?

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: the createStmt was get from downstream checkpoint or from worker message, so no error here

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,848 @@
...
+	jobCfg, err := c.jobStore.GetJobCfg(ctx)
+	if err != nil {
+		return nil, optimism.ConflictError, err
+	} else if jobCfg.ShardMode == "" {
+		return nil, optimism.ConflictError, errors.New("coordinate error with non-shard-mode")
+	}
+
+	for {
+		// create shard group if not exists.
+		g, err := c.loadOrCreateShardGroup(ctx, item.TargetTable, jobCfg)
```
lance6716, line 229: concurrent support for `Coordinate` brings many troubles, we need extra logic for protecting c.pending, g.deleted,  for-loop to optimistically imitate no-locking, maybe we should use simple implementation next time.

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,848 @@
...
+		ddls, conflictStage, err := g.handle(ctx, item)
+		if err != nil {
+			g.mu.Unlock()
+			return ddls, conflictStage, err
+		}
+
+		g.deleted = g.isResolved(ctx)
+		// if all source table is deleted or the shard group is resolved, we should remove the shard group.
+		if g.deleted {
+			c.removeShardGroup(ctx, item.TargetTable)
```
lance6716, line 250: seems a dead lock, here we acquired g.mu then c.mu, but in ShowDDLLocks we reverted

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: Unless we use distributed locks, there doesn't seem to be a particularly good way to do it

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: [7622504](https://github.com/pingcap/tiflow/pull/7523/commits/76225045c73a09d8d566c3938f2c07119395e8a8)

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: Maybe we can change to synchronous in the future.

GMHDBJD: [2562bd7](https://github.com/pingcap/tiflow/pull/7523/commits/2562bd758df1bd386391d8fde07b75f5f747c7a0) synchronous

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,812 @@
...
+		if cmp, err := joined.Compare(postTable); err != nil || cmp < 0 {
+			continue
+		}
+
+		return true
+	}
+	return false
+}
+
+func (g *shardGroup) allTableSmaller(tp tableType) bool {
```
lance6716, line 567: it checks all current table structure of conflicting tables are smaller than joined table by tableType? why does it not check `cmp < 0`?

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,812 @@
...
+	normalTables map[metadata.SourceTable]string
+	// conflictTables represents upstream table info after executing conflict DDL.
+	conflictTables      map[metadata.SourceTable]string
+	tableAgent          TableAgent
+	droppedColumnsStore *metadata.DroppedColumnsStore
+	id                  frameModel.MasterID
+	cfg                 *config.JobCfg
+}
+
+func newShardGroup(ctx context.Context, id frameModel.MasterID, cfg *config.JobCfg, targetTable metadata.TargetTable, sourceTables map[metadata.SourceTable]struct{}, kvClient metaModel.KVClient, tableAgent TableAgent) (*shardGroup, error) {
```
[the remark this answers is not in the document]
GMHDBJD: do in #7858

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,812 @@
...
+	// handle ddls one by one
+	for idx, ddl := range item.DDLs {
+		prevTableStmt, postTableStmt := g.getTableForOneDDL(item, idx)
+		schemaChanged, conflictStage := g.handleDDL(item.SourceTable, prevTableStmt, postTableStmt)
+
+		switch conflictStage {
+		case optimism.ConflictDetected:
+			return nil, optimism.ConflictDetected, errors.Errorf("conflict detected for table %v", item.SourceTable)
+		case optimism.ConflictSkipWaitRedirect:
+			return newDDLs, optimism.ConflictSkipWaitRedirect, nil
```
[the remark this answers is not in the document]
GMHDBJD: no, they will be executed

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,812 @@
...
+		case optimism.ConflictNone:
+			if col, err := g.checkAddDroppedColumn(ctx, item.SourceTable, ddl, prevTableStmt, postTableStmt, dropCols); err != nil {
+				return nil, optimism.ConflictError, err
```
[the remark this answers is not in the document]
GMHDBJD: it should report an error to user because it may cause data inconsistent

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,812 @@
...
+			// return schema changed in 2 cases
+			// oldJoined != newJoined
+			// prevTable < postTable
+			return (joinedErr != nil || joinedCmp != 0) || tableCmp < 0, optimism.ConflictNone
+		}
+	}
+
+	log.L().Info("found conflict for DDL", zap.Any("source table", sourceTable), zap.String("prevTable", prevTableStmt), zap.String("postTable", postTableStmt), log.ShortError(tableErr))
+
+	if idempotent || g.noConflictWithOneNormalTable(sourceTable, prevTable, postTable) {
```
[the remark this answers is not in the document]
GMHDBJD: just no conflict with one normal table. e.g.
tb1(a,b), tb2(a,c), tb3(a,b)
receive a ddl tb3(a,b) -> tb3(a,c)
though it's a conflict DDL(rename column) and conflict with tb1, but it's not conflict with tb3, which has been executed(save in checkpoint), so this DDL can be executed.

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: just copy from https://github.com/pingcap/tiflow/blob/master/dm/pkg/shardddl/optimism/lock.go#L965-L974
it check all tables doesn't have column x

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,812 @@
...
+		if err := g.clear(ctx); err != nil {
+			c.logger.Error("clear shard group failed", zap.Error(err))
+		}
+	}
+	delete(c.shardGroups, targetTable)
+}
+
+type shardGroup struct {
+	// normalTables represents upstream table info record in checkpoint.
+	normalTables map[metadata.SourceTable]string
```
lance6716, line 287: can we say that all normalTables can be joined without error?

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,812 @@
...
+func (g *shardGroup) checkAddDroppedColumn(ctx context.Context, sourceTable metadata.SourceTable, ddl string, prevTableStmt, postTableStmt string, newDroppedColumns []string) (string, error) {
+	currTable := g.normalTables[sourceTable]
+	defer func() {
+		g.normalTables[sourceTable] = currTable
+	}()
+
+	g.normalTables[sourceTable] = prevTableStmt
+	oldJoined, err := g.joinTables(normal)
+	if err != nil {
+		// nolint:nilerr
```
lance6716, line 655: will it happen? Seems the function is called only when current DDL is not conflicting

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,812 @@
...
+	}
+
+	cmp, err := oldJoined.Compare(newJoined)
+	if err != nil {
+		// nolint:nilerr
+		return "", nil
+	}
+
+	if cmp <= 0 {
+		if col, err2 := optimism.AddDifferentFieldLenColumns("", ddl, oldJoined, newJoined); err2 != nil {
```
lance6716, line 674: should we mention that checkAddDroppedColumn will also check field length?

And why this case is "add column with a larger field len"? it can be anything that modifying the table structure with the only longest field length column

engine/jobmaster/dm/ddl_coordinator.go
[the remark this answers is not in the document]
GMHDBJD: Yes, but we don't know whether the other source will drop it or not.

engine/jobmaster/dm/ddl_coordinator.go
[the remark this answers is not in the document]
GMHDBJD: sry. it's no conflict with tb2. We always return true for a conflict DDL, such as rename column/modify column, and rely on the worker to handle the DDL error.

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: Usually it's true, except for:
1. start a task with different and conflict shard tables
2. user manually operate-schema or force skip/execute the ddl
3. worker1 rename column done, worker2 rename column not done but restarted, then jobmaster will receive the conflict normal tables

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: see above reply

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: I don't want to change the original name😂

engine/jobmaster/dm/ddl_coordinator.go
GMHDBJD: [08ad8c1](https://github.com/pingcap/tiflow/pull/7523/commits/08ad8c1c45c73ef4b60b430273f2da5ae05a6686)

lance6716 (approved):

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,814 @@
...
+	return ShowDDLLocksResponse{Locks: ddlLocks}
+}
+
+func (c *DDLCoordinator) loadOrCreateShardGroup(ctx context.Context, targetTable metadata.TargetTable, jobCfg *config.JobCfg) (*shardGroup, error) {
+	if g, ok := c.shardGroups[targetTable]; ok {
+		return g, nil
+	}
+
+	newGroup, err := newShardGroup(ctx, c.jobID, jobCfg, targetTable, c.tables[targetTable], c.kvClient, c.tableAgent)
+	if err != nil {
```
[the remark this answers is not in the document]
GMHDBJD: https://github.com/pingcap/tiflow/blob/08ad8c1c45c73ef4b60b430273f2da5ae05a6686/engine/jobmaster/dm/ddl_coordinator.go#L210 jobCfg gets from the jobstore, which will be updated when user update-config

engine/jobmaster/dm/ddl_coordinator.go
```diff
@@ -0,0 +1,814 @@
...
+		}
+		joined, err = joined.Join(table)
+		if err != nil {
+			return joined, err
+		}
+	}
+	return joined, nil
+}
+
+func (g *shardGroup) noConflictForTables(tp tableType) bool {
```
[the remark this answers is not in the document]
GMHDBJD: added

engine/jobmaster/dm/ddl_coordinator.go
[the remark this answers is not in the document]
okJiang: But during the reconciliation process, the `jobCfg` will be not updated in `ShardGroup`.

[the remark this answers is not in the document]
GMHDBJD: That's expected because you don't know if you'll get the ddl or update-config command first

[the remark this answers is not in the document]
GMHDBJD: So we just get the copy of jobCfg when we receive the ddl request

lichunzhu (approved):
LGTM

PR #7524, opened 2022-11-03
checker(dm): add precheck on upstream/downstream table structure

lance6716:

Issue Number: ref #4287

add precheck when downstream table is created before the task, it will cause warning when
- charset is different and downstream is not utf8mb4
- collation is different
- PK/UK is different
- upstream has more columns
- downstream has more columns and they must specify values when inserting

#### Tests 

 - Unit test

### Release note 

```release-note
Add more precheck about downstream tables that are created before task starts
```

### DIFF
--- a/dm/pkg/checker/table_structure.go
+++ b/dm/pkg/checker/table_structure.go
@@ -373,6 +373,64 @@ func (c *TablesChecker) checkTableStructurePair(upstream, downstream *ast.Create
 		})
 	}
 
+	// check PK/UK
+	upstreamPKUK := getPKAndNotNullUK(upstream)
+	downstreamPKUK := getPKAndNotNullUK(downstream)
+	// the number of PK/UK should be small, we use a simple but slow algorithm for now
+	for idxNameUp, s := range upstreamPKUK {
+		for idxNameDown, s2 := range downstreamPKUK {
+			if stringSetEqual(s, s2) {
+				delete(upstreamPKUK, idxNameUp)
+				delete(downstreamPKUK, idxNameDown)
+				break
+			}
+		}
+	}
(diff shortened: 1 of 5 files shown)


dm/pkg/checker/table_structure.go
```diff
@@ -299,19 +343,96 @@ func (c *TablesChecker) checkUnique(cst *ast.Constraint) bool {
...
+			state: StateWarning,
+			errMessage: fmt.Sprintf("downstream has more PK or NOT NULL UK than upstream, table name: %s, index name: %s, columns: %v",
+				downstream.Table.Name.O, idxName, utils.SetToSlice(cols)),
+		})
+	}
+
+	// check columns
+	upstreamCols := getColumnsAndIgnorable(upstream)
+	downstreamCols := getColumnsAndIgnorable(downstream)
+	// TODO: handle extended column
```
lichunzhu, line 408: Maybe we can't be released in 6.4 until this is fixed?

lichunzhu (approved):
LGTM

dm/pkg/checker/table_structure.go
lance6716: The user should be able to use that feature according to the document, we just can't find the error in advance because precheck is not implemented. I think it's OK.

lichunzhu: Extended columns feature downstream will always have more columns. Will this check always return a warning?

okJiang (approved):
nice job!

dm/pkg/checker/table_structure.go
lance6716: Currently yes, I'll fix this TODO before v6.5.0 release

PR #7537, opened 2022-11-07
sinkv2(cdc): add a memory table based sort engine for tests

hicqu:

Issue Number: ref #5928 

Add a memory table based sort engine for tests

#### Tests 

 - Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/sorter/sorter.go
+++ b/pkg/sorter/sorter.go
@@ -102,7 +102,7 @@ type Position struct {
 
 // Valid indicates whether the position is valid or not.
 func (p Position) Valid() bool {
-	return p.StartTs != 0 && p.CommitTs != 0
+	return p.CommitTs != 0
 }
 
 // Next can only be called on a valid Position.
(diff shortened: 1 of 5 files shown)


pkg/sorter/sorter.go
```diff
@@ -102,7 +102,7 @@ type Position struct {
 
 // Valid indicates whether the position is valid or not.
 func (p Position) Valid() bool {
-	return p.StartTs != 0 && p.CommitTs != 0
```
hi-rustin, line 105: Why do we remove start ts?

hi-rustin (approved):

pkg/sorter/sorter.go
hicqu: In some tests transactions' `StartTs` can be 0, so just check `CommitTs` is OK.

nongfushanquan (approved):

Issue #7343, opened 2022-10-13
DM: more detailed query-status


okJiang:
### Is your feature request related to a problem?

no

### Describe the feature you'd like

add some status:
- duration(from unit start)
- instantaneous speed
	- for dump: bytes/s https://github.com/pingcap/tidb/pull/38456
	- for load: bytes/s https://github.com/pingcap/tiflow/pull/7212
	- ~~for sync: rows/s(now is tps)~~, in fact, tps is the rps(rows per sec), a naming problem
- progress
	- for dump: calculate by chunks https://github.com/pingcap/tidb/pull/38625
	- ~~for load: calculate by rows~~(bytes is accurate)
- completed phase's detail https://github.com/pingcap/tiflow/pull/7386

### Teachability, Documentation, Adoption, Migration Strategy

display in TiDB cloud

lance6716:
Duration (start-time) should be kept across failover, for DM in engine scenario only the metaDB is survived across failover, so we might only add this field to the response of engine job's status.

PR #7538, opened 2022-11-07
syncer(dm): try to use driver API to calculate IO 

lance6716:

Issue Number: ref #7343

add a atomic variable to let IO bytes be added into it

#### Tests 

 - Integration test

### Release note 

```release-note
`None`.
```

### DIFF
--- a/metrics/grafana/ticdc.json
+++ b/metrics/grafana/ticdc.json
@@ -83,7 +83,7 @@
       {
         "datasource": "${DS_TEST-CLUSTER}",
         "enable": false,
-        "expr": "delta(up{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", job=~\"tikv|ticdc\"}[30s]) < BOOL 0",
+        "expr": "delta(up{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", job=~\"tikv|ticdc|tidb|pd\"}[30s]) < BOOL 0",
         "hide": false,
         "iconColor": "#FF9830",
         "limit": 100,
@@ -13169,7 +13169,7 @@
           "targets": [
             {
               "exemplar": true,
-              "expr": "sum(rate(ticdc_kvclient_batch_resolved_event_size_bucket{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}[1m])) by (le)",
+              "expr": "sum(rate(ticdc_kvclient_batch_resolved_event_size_bucket{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$ticdc_instance\"}[1m])) by (le)",
               "format": "heatmap",
               "instant": false,
               "interval": "",


dm/syncer/dbconn/utils.go
```diff
@@ -96,3 +98,30 @@ func getSessionVariable(tctx *tcontext.Context, conn *DBConn, variable string) (
...
+type TCPConnWithIOCounter struct {
+	*net.TCPConn
+	c *atomic.Uint64
+}
+
+// NewTCPConnWithIOCounter creates a new TCPConnWithIOCounter.
+func NewTCPConnWithIOCounter(conn *net.TCPConn, c *atomic.Uint64) net.Conn {
+	return &TCPConnWithIOCounter{
+		TCPConn: conn,
+		c:       c,
```
amyangfei, line 113: Does it require to separate read and write traffic?

lance6716: confirmed with @buchuitoudegou that we can add them together

buchuitoudegou (commented):
wonderful solution! But I'm wondering how we retrieve the network consumption from the `TCPConnWithIOCounter`?

I'm also worried about how much difference is there between the data size we write to the downstream (or fetch from the upstream) and that we collect in this manner.

dm/syncer/dbconn/utils.go
buchuitoudegou: I think calculating the whole stuff is preferable since both read and write are billed by the AWS/GCP. But I'm worried about the network consumption collected in this way would be far larger than the data size (i.e. the total size of binlog and SQLs we write to the downstream) so that users would suspect our stats.

dm/syncer/dbconn/utils.go
```diff
@@ -96,3 +98,30 @@ func getSessionVariable(tctx *tcontext.Context, conn *DBConn, variable string) (
 	}
 	return value, nil
 }
+
+// TCPConnWithIOCounter is a wrapper of net.TCPConn with counter that accumulates
+// the bytes this connection reads/writes.
+type TCPConnWithIOCounter struct {
+	*net.TCPConn
+	c *atomic.Uint64
```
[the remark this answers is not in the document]
lance6716: > I'm wondering how we retrieve the network consumption from the TCPConnWithIOCounter?

this is only internally used so short name is tolerated. The caller owns the atomic variable and pass the reference to `NewTCPConnWithIOCounter`, and caller will read atomic variable later.

So engine can `NewAtomicUint64`, set it to config, start a worker and read the value from it later

> how much difference is there between the data size we write to the downstream

This should take effect at the bottom-most layer I can find 🤔

dm/config/subtask.go
```diff
@@ -291,6 +293,8 @@ type SubTaskConfig struct {
 	ExtStorage      extstorage.ExternalStorage `toml:"-" json:"-"`
 	MetricsFactory  promutil.Factory           `toml:"-" json:"-"`
 	FrameworkLogger *zap.Logger                `toml:"-" json:"-"`
+	UUID            string                     `toml:"-" json:"-"`
```
amyangfei, line 296: Could you add some comments about how does it work.

amyangfei (approved):

dm/config/subtask.go
lance6716: Do you have a better name @GMHDBJD I'll rename it and add some comments in next commit

dm/syncer/dbconn/utils.go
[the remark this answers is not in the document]
buchuitoudegou: > this is only internally used so short name is tolerated. The caller owns the atomic variable and pass the reference to NewTCPConnWithIOCounter, and caller will read atomic variable later.

lgtm

> This should take effect at the bottom-most layer I can find

Maybe we can try on this approach in this version since it seems accessible to collect the stats in this way. If we received any suspicion or complaint from the user, then we should consider a different way.

buchuitoudegou (approved):

lance6716:
```
denied: adding 28.8 MiB of storage resource, which when updated to current usage of 15.0 GiB will exceed the configured upper limit of 15.0 GiB.
```

amyangfei:
> ```
> denied: adding 28.8 MiB of storage resource, which when updated to current usage of 15.0 GiB will exceed the configured upper limit of 15.0 GiB.
> ```

Have cleaned docker registry manually, seems the clean mechanism doesn't work 😕

PR #7548, opened 2022-11-08
sink(ticdc): Use dm sqlmodel pkg to generate sql in batch mode

asddongmen:

Issue Number: ref #7653 

After this PR, when`batch-dml-enable = false` is not explicitly specifed in `sink-uri`, `mysqlBackend` will convert `RowChangedEvent` in a single txn event in batch mode, which means all `RowChangedEvent` in a txn event that has the same `type` would be converted into one SQL. 
Let say we've got a table `create `test.t1` (id int primary key, name varchar(20));` and a 
`SingleTableTxn = [delete-1(id=1), delete-2(id=2), delete-3(id=3), insert-1(id=4, name="dd"), update-1(id=5, name="ff"), insert-2(id=6, name="gg"), update-2(id=7, name="hh")]`
(the number 1,2,3 is only a fake sequence number, do not represent anything more), it wil be convert into 3 sqls in `prepareDMLs()`:
```sql 
1. DELETE FROM `test`.`t1` WHERE `id` in ((1), (2));
2. INSERT INTO `test`.`t1` (`id`, `name`) VALUES (4, "dd"), (6, "gg");
3. UPDATE `test`.`t1` SET `id`=CASE WHEN `id`= 5 THEN 5 WHEN `id`=7 THEN 7 END, `name`=CASE WHEN `id`=5 THEN "ff" WHEN `id`=7 THEN "gg" END WHERE `id` IN (5,7);
```
 
#### Tests 

 - Unit test
 - Integration test
 Now we turn on `batch-dml-enable` by default, so all exists it case go throught this logic.
1. delete 500000 rows in a single txn (workload bank)
image
2. replicate 2000000 rows in incremental scan (workload bank)
image
3. sysbench workload
image

### Release note 

```release-note
`None`
```


cdc/sinkv2/eventsink/txn/mysql/mysql.go
```diff
@@ -191,6 +200,167 @@ type preparedDMLs struct {
...
+		if col.Charset != "" && col.Charset != charset.CharsetBin {
+			colValBytes, ok := col.Value.([]byte)
+			if ok {
+				row.Columns[i].Value = string(colValBytes)
+			}
+		}
+	}
+}
+
+// 对 delete event 在这里做一些特殊判断 如 id<? and id>? 这种条件的sql语句
```
nongfushanquan, line 248: pls write the comments in English

asddongmen: ok, I will refine all comments after all integration test cases are passed.

cdc/sinkv2/eventsink/txn/mysql/mysql.go
```diff
@@ -191,6 +198,163 @@ type preparedDMLs struct {
...
+	}
+
+	// handle insert
+	if len(insertRows) > 0 {
+		if translateToInsert {
+			sql, value := sqlmodel.GenInsertSQL(sqlmodel.DMLInsert, insertRows...)
+
+			sqls = append(sqls, sql)
+			values = append(values, value)
+
```
zhaoxinyu, line 326:
```suggestion
```

cdc/sinkv2/eventsink/txn/mysql/mysql.go
```diff
@@ -191,6 +198,163 @@ type preparedDMLs struct {
...
+		} else {
+			sql, value := sqlmodel.GenInsertSQL(sqlmodel.DMLReplace, insertRows...)
+			sqls = append(sqls, sql)
+			values = append(values, value)
+
```
zhaoxinyu, line 331:
```suggestion
```

cdc/sinkv2/tablesink/table_sink_impl.go
```diff
@@ -96,12 +96,12 @@ func (e *EventTableSink[E]) UpdateResolvedTs(resolvedTs model.ResolvedTs) error
 		return nil
 	}
 	resolvedEvents := e.eventBuffer[:i]
-	e.eventBuffer = append(make([]E, 0, len(e.eventBuffer[i:])), e.eventBuffer[i:]...)
 
 	// We have to create a new slice for the rest of the elements,
 	// otherwise we cannot GC the flushed values as soon as possible.
-	resolvedCallbackableEvents := make([]*eventsink.CallbackableEvent[E], 0, len(resolvedEvents))
+	e.eventBuffer = append(make([]E, 0, len(e.eventBuffer[i:])), e.eventBuffer[i:]...)
```
zhaoxinyu, line 102: Why line:99 and line:103 should be exchanged?

zhaoxinyu:
Rest LGTM

cdc/sinkv2/eventsink/txn/mysql/mysql.go
zhaoxinyu, line 326:
```suggestion
```

cdc/sinkv2/eventsink/txn/mysql/mysql.go
zhaoxinyu, line 331:
```suggestion
```

cdc/sinkv2/eventsink/txn/mysql/mysql.go
```diff
@@ -214,33 +378,49 @@ func (s *mysqlBackend) prepareDMLs() *preparedDMLs {
...
+				tableColumns = firstRow.PreColumns
+			}
+			// only use batch dml when the table has a handle key
+			if hasHandleKey(tableColumns) {
+				// TODO(dongmen): find a better way to get table info.
+				tableInfo := model.BuildTiDBTableInfo(tableColumns, firstRow.IndexColumns)
+				sql, value := batchSingleTxnDmls(event, tableInfo, translateToInsert)
+				sqls = append(sqls, sql...)
+				values = append(values, value...)
+				rowCount += len(sql)
```
zhaoxinyu, line 415: Maybe this variable is not accurate anymore. Should it be replaced with `statementCount`?

cdc/sinkv2/tablesink/table_sink_impl.go
asddongmen: I did not exchange them, what I really do is to move the comment to a correct place.

nongfushanquan (approved):

Issue #7475, opened 2022-10-28
Standardize openapi error format


gozssky:
### Is your feature request related to a problem?

Currently, API error returns are inconsistent and lack uniform standards and accuracy, which may confuse the users.

1. The framework API (known as the basic job API) returns [google/rpc/status.proto](https://github.com/googleapis/googleapis/blob/923a4b9a6538087654f02c089cbbaa7c45fd06bd/google/rpc/status.proto#L35) as the error format, and the details field is defined as [engine/proto/errorv2.proto](https://github.com/pingcap/tiflow/blob/9abde064c9a50aef77ffd3aa8be609d3265e4ed0/engine/proto/errorv2.proto#L7). errorv2 was originally designed for passing different structured errors between internal components of engine, and its detail part is actually the result of `json.Marshal` result of arbitrary Go struct. When it is exposed to external services, users may not understand the meaning of this field.

2. The framework has no constraints on the error return of the business API, which is freely implemented by the business (DM) and lacks documentation.

### Describe the feature you'd like

Define a uniform api error format. The following content is an example of an error.
```json
{
  "code": "DFLOW:ErrJobAlreadyExists",
  "message": "job job-1 alreadys exists"
}
```
The uniform api error format has two fields. 
- `code` is a simple error code that can be easily handled by the client. 
- `message` describe the details of the error and offer an actionable resolution to it.

The `code` is a string, we attach different prefixes for different components. 
The code format is `{component}:{error_code}`, where component may be `DM`, `CDC` or `DFLOW`.

PR #7565, opened 2022-11-09
lint(engine): update lint scripts and config

gozssky:

Issue Number: ref https://github.com/pingcap/tiflow/issues/7475

* Add depguard config to check errors import path for engine.
* Fix incorrect `--skip-dirs` in golangci-lint script.
* Fix several log import paths.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/jobmaster/dm/metadata/unit_state.go
```diff
@@ -28,13 +28,14 @@ import (
 
 // UnitState represents the state of units.
 type UnitState struct {
-	state
+	state //nolint:unused
```
lance6716, line 31: I forget why we must embed it into a customized State. Can we remove it now? @GMHDBJD

gozssky: `state` is an empty interface. From my understanding, it is not much different from a raw interface.

amyangfei (approved):

.golangci.yml
```diff
@@ -60,6 +60,18 @@ linters-settings:
...
+        include-go-root: true
+        packages:
+          - errors
+          - github.com/pingcap/errors
+        packages-with-error-message:
+          - errors: "error handling is allowed only by github.com/pingcap/tiflow/pkg/errors"
+          - github.com/pingcap/errors: "error handling is allowed only by github.com/pingcap/tiflow/pkg/errors"
+        ignore-file-rules:
+          - "!**/tiflow/engine/**"
+          - "**/engine/pkg/rpcutil/*.go"
```
amyangfei, line 74: This rule is confused, I wonder why `!**/tiflow/cdc/**` is not configured, and it is not checked.

gozssky: 1. `"!**/tiflow/engine/**"` means all files except `engine` directory should be excluded.
2. `"**/engine/pkg/rpcutil/*.go"` means `engine/pkg/rpcutil` directory should be excluded.

`tiflow/cdc` is excluded in rule 1.

lance6716 (approved):

engine/jobmaster/dm/metadata/unit_state.go
GMHDBJD: We can remove it.

gozssky: Removed in https://github.com/pingcap/tiflow/pull/7565/commits/ff2ac803e5b92130b448c43c47a29b9b10f3ddb4. @lance6716 @GMHDBJD PTAL again.

lance6716 (commented):
lgtm

PR #7580, opened 2022-11-10
relay: add async/batch relay writer

GMHDBJD:

- async write relay log
- write relay log in batch(1MB)

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


dm/relay/binlog_writer.go
```diff
@@ -61,6 +74,71 @@ func NewBinlogWriter(logger log.Logger, relayDir string) *BinlogWriter {
...
+			}
+			errOccurs = true
+			return
+		}
+		buf.Reset()
+	}
+
+	for {
+		select {
+		case bs, ok := <-w.input:
```
lance6716, line 118: when we can read something from the buffered channel, we can read all of them then do a flush. This can achieve low latency because no `waitTime` or `bufferSize` is needed

GMHDBJD: I have try with in [dc9edc5](https://github.com/pingcap/tiflow/pull/7580/commits/dc9edc540e47abbfea1407b5e55b5517364af2c4) originally, but the performance is even worse than the original....
```
for bs := <- range w.input{
     ...
     if len(w.input)==0{
          flush
     }
}
```

lance6716: 😂 ptal @D3Hunter

buchuitoudegou: Does @lance6716 mean reading and writing as long as the channel is full?

lance6716: I think we should be able to do a block reading on channel, and when the goroutine continues it reads all data from the channel to flush them. This is more adaptive.

Maybe we need to take a look at profling to see why this idea does not work.

GMHDBJD: performance issue fixed after remove the lock in relay writer [20c870a](https://github.com/pingcap/tiflow/pull/7580/commits/20c870a54ddc43b52dbc5567c06fb7b8eec3cbff)

dm/relay/binlog_writer.go
```diff
@@ -105,32 +176,46 @@ func (w *BinlogWriter) Close() error {
 
 	w.file = nil
 	w.offset.Store(0)
-	w.uuid = ""
-	w.filename = ""
+	w.uuid.Store("")
+	w.filename.Store("")
+	w.input = nil
 
+	if writeErr := w.Error(); writeErr != nil {
```
lance6716, line 183: return w.Error()

dm/relay/binlog_writer.go
```diff
@@ -105,32 +176,46 @@ func (w *BinlogWriter) Close() error {
...
 func (w *BinlogWriter) Write(rawData []byte) error {
-	w.mu.RLock()
-	defer w.mu.RUnlock()
-
 	if w.file == nil {
 		return terror.ErrRelayWriterNotOpened.Delegate(errors.New("file not opened"))
 	}
+	w.input <- rawData
+	w.offset.Add(int64(len(rawData)))
+	return w.Error()
```
lance6716, line 195: since every Write need to check Error, Error is a hot function and we'd better using fast implementation. I prefer atomic variable like using `atomic.Error` and CAS to replace nil to an error

dm/relay/binlog_writer.go
```diff
@@ -61,6 +73,59 @@ func NewBinlogWriter(logger log.Logger, relayDir string) *BinlogWriter {
...
+		}
+		buf.Reset()
+	}
+
+	for bs := range w.input {
+		if errOccurs {
+			continue
+		}
+		if bs != nil {
+			buf.Write(bs)
```
lance6716, line 118: this will cause the byte content copied from `bs` to `buf`? not sure if use `[][]bytes` to take ownership of `bs` is faster

dm/relay/binlog_writer.go
```diff
@@ -27,17 +29,27 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	bufferSize = 1 * 1024 * 1024 // 1MB
+	chanSize   = 1024
+	waitTime   = 10 * time.Millisecond
```
lance6716, line 35:
```suggestion
```

lance6716 (commented):
no need to accept all comments if it's negligible in performance

dm/relay/binlog_writer.go
GMHDBJD: If w.Error()!=nil, we should return the error of file close, L174

dm/relay/binlog_writer.go
```diff
@@ -61,6 +73,59 @@ func NewBinlogWriter(logger log.Logger, relayDir string) *BinlogWriter {
...
+	}
+
+	for bs := range w.input {
+		if errOccurs {
+			continue
+		}
+		if bs != nil {
+			buf.Write(bs)
+		}
+		if bs == nil || buf.Len() > bufferSize || len(w.input) == 0 {
```
D3Hunter, line 120:
```suggestion
        // we use bs = nil to mean flush
		if bs == nil || buf.Len() > bufferSize || len(w.input) == 0 {
```

dm/relay/binlog_writer.go
```diff
@@ -105,32 +176,46 @@ func (w *BinlogWriter) Close() error {
...
 
 func (w *BinlogWriter) Write(rawData []byte) error {
-	w.mu.RLock()
-	defer w.mu.RUnlock()
-
 	if w.file == nil {
 		return terror.ErrRelayWriterNotOpened.Delegate(errors.New("file not opened"))
 	}
+	w.input <- rawData
+	w.offset.Add(int64(len(rawData)))
```
D3Hunter, line 194: `BinlogReader` will use this offset to check whether there's need to re-parse binlog, maybe move it after we actual write data.

D3Hunter (commented):
rest lgtm

dm/relay/binlog_writer.go
GMHDBJD: we use offset to check whether hole exist, so cannot update it later.

GMHDBJD: https://github.com/pingcap/tiflow/blob/d98f885b51eb8d3cb35a960598dbe025281cf733/dm/relay/relay_writer.go#L304

dm/relay/binlog_writer.go
GMHDBJD: 0.43s/30s in frame graph, leave it for better code reading

dm/relay/binlog_writer.go
D3Hunter: use another offset?

dm/relay/binlog_writer.go
```diff
@@ -64,6 +74,53 @@ func NewBinlogWriter(logger log.Logger, relayDir string) *BinlogWriter {
...
+		}
+
+		if w.file == nil {
+			w.err.CompareAndSwap(nil, terror.ErrRelayWriterNotOpened.Delegate(errors.New("file not opened")))
+			errOccurs = true
+			return
+		}
+		n, err := w.file.Write(buf.Bytes())
+		if err != nil {
+			w.err.CompareAndSwap(nil, terror.ErrBinlogWriterWriteDataLen.Delegate(err, n))
```
lance6716, line 97: sorry seems the `nil` for CAS is not what I expected

```
func TestCAS(t *testing.T) {
	var nilErr error
	e := atomic.NewError(nil)
	e.Store(nilErr)
	require.NoError(t, e.Load())
	ok := e.CompareAndSwap(nilErr, errors.New("test"))
	require.True(t, ok)
}
```

Can we use the `nilErr` variable or simply `Store`? but in the latter we will store the last error not the first one.

dm/relay/binlog_writer.go
```diff
@@ -27,17 +28,31 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	bufferSize = 1 * 1024 * 1024 // 1MB
+	chanSize   = 1024
+)
+
+var nilErr error
```
lance6716, line 36: I suspect we should set it to BinlogWriter.err at Init, otherwise CAS will always failed. Please add an unit test to cover write IO fail case.
Also the binlog_write_test.go is very small, please help migrate this file to testify!

dm/relay/binlog_writer.go
```diff
@@ -64,6 +79,54 @@ func NewBinlogWriter(logger log.Logger, relayDir string) *BinlogWriter {
...
+	)
+
+	// writeToFile writes buffer to file
+	writeToFile := func() {
+		if buf.Len() == 0 {
+			return
+		}
+
+		if w.file == nil {
+			w.err.CompareAndSwap(nilErr, terror.ErrRelayWriterNotOpened.Delegate(errors.New("file not opened")))
```
lance6716, line 96: can we simply use `Generate()` rather than `Delegate`?

dm/relay/relay_writer.go
```diff
@@ -145,6 +152,10 @@ func (w *FileWriter) handleFormatDescriptionEvent(ev *replication.BinlogEvent) (
 		if err != nil {
 			return WResult{}, terror.Annotatef(err, "write binlog file header for %s", fullName)
 		}
+		err = w.Flush()
+		if err != nil {
+			return WResult{}, terror.Annotatef(err, "write binlog file header for %s", fullName)
```
lance6716, line 157:
```suggestion
			return WResult{}, terror.Annotatef(err, "flush binlog file for %s", fullName)
```

lance6716 (approved):

D3Hunter (approved):

PR #7581, opened 2022-11-11
dockerfile: change base image to rockylinux

GMHDBJD:

- change base image to rockylinux

#### Tests 

 - chaos test

### Release note 

```release-note
None
```

### DIFF
--- a/dm/Dockerfile
+++ b/dm/Dockerfile
@@ -1,5 +1,4 @@
 FROM golang:1.19.1-alpine3.15 as builder
-MAINTAINER siddontang
 
 RUN apk add --no-cache git make
 
@@ -18,7 +17,7 @@ RUN apk update && apk add bash
 
 RUN make dm-worker dm-master dmctl
 
-FROM alpine:3.15
+FROM rockylinux:9.0
 
 # keep compatibility
 COPY --from=builder /go/src/github.com/pingcap/dm/bin/dm-worker /dm-worker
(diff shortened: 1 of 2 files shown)


lichunzhu (approved):
LGTM

dm/Dockerfile
```diff
@@ -1,5 +1,4 @@
 FROM golang:1.19.1-alpine3.15 as builder
```
lance6716, line 1: do we need to update this image? not sure if the building stage should also be put in safer image

dm/Dockerfile
```diff
@@ -18,7 +17,7 @@ RUN apk update && apk add bash
 
 RUN make dm-worker dm-master dmctl
 
-FROM alpine:3.15
+FROM rockylinux:9.0
```
CharlesCheung96, line 20: Why should this change be made? Should the relevant dockerfile in engine be changed to this as well?

dm/Dockerfile
lijie0123: If runtime use glibc, builder use glibc

lijie0123: here we just parallel produce another docker image, directly change may be dangerous

lijie0123 (changes requested):

lijie0123:
here we just parallel produce another docker image, directly change may be dangerous， so you can add this dockerfile to https://pingcap.feishu.cn/wiki/wikcnYlK7ZLs709jXDFSBlUHm6f

dm/Dockerfile
amyangfei: Yes, engine will replace alpine with rockylinux too.

Issue #7555, opened 2022-11-08
dm job master accesses an invalid memory address


CharlesCheung96:
### What did you do?

1. submit a dm job to engine
2. kill executor, kill master
3. restart master and executor
4. query the previous dm job

### What did you expect to see?

dm job resume normally

### What did you see instead?

```
http: panic serving 172.31.52.193:53156: runtime error: invalid memory address or nil pointer dereference
goroutine 375 [running]:
net/http.(*conn).serve.func1()
        net/http/server.go:1850 +0xbf
panic({0x3393cc0, 0x68133e0})
        runtime/panic.go:890 +0x262
github.com/pingcap/tiflow/engine/jobmaster/dm/metadata.(*MetaData).JobStore(...)
        github.com/pingcap/tiflow/engine/jobmaster/dm/metadata/metadata.go:49
github.com/pingcap/tiflow/engine/jobmaster/dm.(*JobMaster).QueryJobStatus(0xc00075c600, {0x4265de8?, 0xc000ed7dc0}, {0x0, 0x0, 0x0})
        github.com/pingcap/tiflow/engine/jobmaster/dm/api.go:50 +0x65
github.com/pingcap/tiflow/engine/jobmaster/dm.(*JobMaster).DMAPIGetJobStatus(0x391ad17?, 0xc000d2ab00, {0x450001?})
        github.com/pingcap/tiflow/engine/jobmaster/dm/openapi.go:60 +0x74
github.com/pingcap/tiflow/engine/jobmaster/dm/openapi.(*ServerInterfaceWrapper).DMAPIGetJobStatus(0xc000eacc30, 0xc000d2ab00)
        github.com/pingcap/tiflow/engine/jobmaster/dm/openapi/gen.server.go:258 +0x24a
github.com/gin-gonic/gin.(*Context).Next(...)
        github.com/gin-gonic/gin@v1.7.4/context.go:165
github.com/pingcap/tiflow/engine/jobmaster/dm.terrorHTTPErrorHandler.func1(0xc000d2ab00)
        github.com/pingcap/tiflow/engine/jobmaster/dm/openapi.go:72 +0x42
github.com/gin-gonic/gin.(*Context).Next(...)
        github.com/gin-gonic/gin@v1.7.4/context.go:165
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest(0xc000e036c0, 0xc000d2ab00)
        github.com/gin-gonic/gin@v1.7.4/gin.go:489 +0x63e
[...]

PR #7587, opened 2022-11-11
dm(engine): openapi returns 503 when job master is not initialized

gozssky:

Issue Number: ref https://github.com/pingcap/tiflow/issues/7555

Add middleware to check whether job master is initialized or not. If job master is not initialized, return 503 directly.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/jobmaster/dm/dm_jobmaster.go
```diff
@@ -128,7 +131,11 @@ func (jm *JobMaster) InitImpl(ctx context.Context) error {
 	if err := jm.checkpointAgent.Create(ctx, jm.initJobCfg); err != nil {
 		return err
 	}
-	return jm.taskManager.OperateTask(ctx, dmpkg.Create, jm.initJobCfg, nil)
+	if err := jm.taskManager.OperateTask(ctx, dmpkg.Create, jm.initJobCfg, nil); err != nil {
+		return err
+	}
+	jm.initialized.Store(true)
```
amyangfei, line 137: `OnMasterRecovered` should also set `initialized`, because either `InitImpl` or `OnMasterRecovered` is called during job master starts up

https://github.com/pingcap/tiflow/blob/04173f64ea66bd4e709f8274e5bd1ffe60e70625/engine/framework/base_jobmaster.go#L202-L217

gozssky: Is it possible that `InitImpl` and `OnMasterRecovered` are called on the same `JobMasterImpl` struct?

lance6716 (commented):
rest lgtm, please address existing comment

engine/jobmaster/dm/dm_jobmaster.go
amyangfei: No, they are mutually exclusive.

gozssky: Addressed in https://github.com/pingcap/tiflow/pull/7587/commits/fee52ba544791a0b863d51137cc5d20c82e47528.

amyangfei (approved):

lance6716 (approved):

PR #7584, opened 2022-11-11
*: upgrade tidb version

okJiang:
### What problem does this PR solve?

Issue Number: ref  https://github.com/pingcap/tiflow/issues/4287

#### Tests 

 - No code

### Release note 

```release-note
`None`.
```

PR #7590, opened 2022-11-12
syncer(dm): update dep to remove clone for TableInfo

lance6716:

Issue Number: ref #4287

~in the old PR, I'm afraid that TableInfo is changing by SchemaTracker when DMLWorker is accesing, so I clone it before assign it to DMLWorker. But seems DML and DDL job will not be processed simutaneously, this clone is not needed.~

#### Tests 

 - pass original test

### Release note 

```release-note
`None`.
```


okJiang (commented):
I don't find where changes the tableInfo😂, can you give an example? BTW, related problem about version updating is fixed in https://github.com/pingcap/tiflow/pull/7584.

lance6716:
TableInfo is changed in SchemaTracker when handling DDL, https://github.com/pingcap/tidb/pull/39103

An example: TableInfo is read by checkpoint component as a "flushed TableInfo", and when task has error, checkpoint will use flushed TableInfo to rollback some changes. If the TableInfo is mutable, flushed TableInfo may be modified after a DDL
 is applied in SchemaTracker, rolling back logic is wrong.

lance6716:
ptal @CharlesCheung96 @GMHDBJD @okJiang

okJiang (approved):

D3Hunter (approved):

Issue #7115, opened 2022-09-19
dm code adjustment to adapt tidb cloud


D3Hunter:
### Is your feature request related to a problem?

no

### Describe the feature you'd like

to make DM as a service, we need some adjustment to some apis.

PR #7594, opened 2022-11-14
metrics(engine/dm): add job task stage metrics

okJiang:

Issue Number: ref https://github.com/pingcap/tiflow/issues/7115

- add a metrics, that we can count the stage(e.g. init, unscheduled...) duration for each job when we should alert.

#### Tests 

 - Integration test

### Release note 

```release-note
`None`.
```


engine/executor/dm/metrics.go
```diff
@@ -0,0 +1,17 @@
+package dm
```
lance6716, line 1: copyright

engine/executor/dm/worker.go
```diff
@@ -302,6 +308,7 @@ func (w *dmWorker) setStage(stage metadata.TaskStage) {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 	w.stage = stage
+	w.stageGauge.Set(float64(stage))
```
lance6716, line 311: maybe we can use the metrics factory `w.MetricFactory()`

lance6716 (commented):
there is a metrics test in e2e_dm_test.go, around line 92

engine/executor/dm/worker.go
okJiang: used `w.MetricFactory()`

engine/executor/dm/worker.go
```diff
@@ -141,6 +144,14 @@ func (w *dmWorker) InitImpl(ctx context.Context) error {
 		}
 	}
 	w.cfg.MetricsFactory = w.MetricFactory()
+	w.stageGauge = w.MetricFactory().NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: "dataflow",
```
lance6716, line 149: not sure what's the namespace

/cc @CharlesCheung96

engine/executor/dm/worker.go
```diff
@@ -257,6 +268,7 @@ func (w *dmWorker) tryUpdateStatus(ctx context.Context) error {
 		return err
 	}
 
+	w.stageGauge.Set(0)
```
lance6716, line 271: Should the Finish stage be kept?

lance6716 (commented):
Please note that if the jobmaster failed to create worker, the stage will be unavailable.

okJiang:
> Please note that if the jobmaster failed to create worker, the stage will be unavailable.

I see `taskManager.UpdateTaskStatus` is the place `jobmaster` store `stage`, we can record metrics here? And it is only called when worker online/offline, we also need record metrics in worker internal. It is strange to record metrics two different place.🤔

engine/executor/dm/worker.go
CharlesCheung96: Perhaps dm is a more appropriate namespace because `task-state` is a business-scope monitor.

engine/jobmaster/dm/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+
+package dm
+
+import (
+	"github.com/pingcap/tiflow/engine/pkg/promutil"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	jobFactory        = promutil.NewFactory4Framework()
```
CharlesCheung96, line 22: Note that `promutil.NewFactory4Framework()`  can only be used inside the framework.
Should `jm.metricFactory ()` be used to create the monitor item here?

okJiang: I tried to use `jm.MetricFactory()` to create, but I failed. It seems that using `metricFactory()` function need a timing that after some initialization. Are you sure we can use `jm.MetricFactory()` in `NewTaskManager`?

CharlesCheung96:
> > Please note that if the jobmaster failed to create worker, the stage will be unavailable.
> 
> I see `taskManager.UpdateTaskStatus` is the place `jobmaster` store `stage`, we can record metrics here? And it is only called when worker online/offline, we also need record metrics in worker internal. It is strange to record metrics two different place.🤔 Do you have any better method? @lance6716

@lance6716 @GMHDBJD Please check whether all stage changes in the worker can be fed back to jobMaster via `OnWorkerStatusUpdated`? If so, does that mean we can set up monitoring only in jobMaster?

engine/jobmaster/dm/api_test.go
```diff
@@ -177,7 +178,7 @@ func TestQueryStatusAPI(t *testing.T) {
 	messageAgent := &dmpkg.MockMessageAgent{}
 	jm.messageAgent = messageAgent
 	jm.workerManager = NewWorkerManager(mockBaseJobmaster.ID(), nil, jm.metadata.JobStore(), jm.metadata.UnitStateStore(), nil, nil, nil, jm.Logger(), false)
-	jm.taskManager = NewTaskManager(nil, nil, nil, jm.Logger())
+	jm.taskManager = NewTaskManager(nil, nil, nil, jm.Logger(), promutil.NewFactory4Test(t.TempDir()))
```
okJiang, line 181: use `t.TempDir()` to get an unique string

engine/jobmaster/dm/task_manager.go
```diff
@@ -38,21 +40,33 @@ var (
 type TaskManager struct {
 	*ticker.DefaultTicker
 
+	jobId        string
```
CharlesCheung96, line 43:
```suggestion
```

engine/jobmaster/dm/task_manager.go
```diff
@@ -44,15 +46,26 @@ type TaskManager struct {
...
-func NewTaskManager(initTaskStatus []runtime.TaskStatus, jobStore *metadata.JobStore, messageAgent dmpkg.MessageAgent, pLogger *zap.Logger) *TaskManager {
+func NewTaskManager(initTaskStatus []runtime.TaskStatus, jobStore *metadata.JobStore, messageAgent dmpkg.MessageAgent, pLogger *zap.Logger, metricFactory promutil.Factory) *TaskManager {
 	taskManager := &TaskManager{
 		DefaultTicker: ticker.NewDefaultTicker(taskNormalInterval, taskErrorInterval),
 		jobStore:      jobStore,
 		logger:        pLogger.With(zap.String("component", "task_manager")),
 		messageAgent:  messageAgent,
+		gaugeVec: metricFactory.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Namespace: "job_master",
```
CharlesCheung96, line 63: Does the monitor name here need to be compatible with `dm_worker_task_state`? This makes it easier for the dm and engine to share the monitor panel.

okJiang: It is ok for me.

engine/jobmaster/dm/task_manager.go
```diff
@@ -105,6 +118,10 @@ func (tm *TaskManager) UpdateTaskStatus(taskStatus runtime.TaskStatus) {
 		zap.Uint64("config_modify_revison", taskStatus.CfgModRevision),
 	)
 	tm.tasks.Store(taskStatus.Task, taskStatus)
+	if _, ok := tm.stageGauge[taskStatus.Task]; !ok {
+		tm.stageGauge[taskStatus.Task] = tm.gaugeVec.WithLabelValues(taskStatus.Task)
+	}
+	tm.stageGauge[taskStatus.Task].Set(float64(taskStatus.Stage))
```
GMHDBJD, line 124: Where do you clear the metrics?

okJiang: I think it is cleared after the taskManager/jobMaster close. For example, full job finished in `testSimpleAllModeTask`, it is not shown when we call `metrics`.

engine/jobmaster/dm/task_manager.go
```diff
@@ -44,15 +46,26 @@ type TaskManager struct {
...
 	taskManager := &TaskManager{
 		DefaultTicker: ticker.NewDefaultTicker(taskNormalInterval, taskErrorInterval),
 		jobStore:      jobStore,
 		logger:        pLogger.With(zap.String("component", "task_manager")),
 		messageAgent:  messageAgent,
+		gaugeVec: metricFactory.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Namespace: "dm",
+				Subsystem: "job",
+				Name:      "task_stage",
```
D3Hunter, line 65:
```suggestion
				Subsystem: "task",
				Name:      "stage",
```

D3Hunter (approved):
rest lgtm

engine/jobmaster/dm/task_manager.go
lance6716: The behaviour you stated is done by framework's metrics factory, and there may exist another case that we should manually clear the metrics, which is user remove an upstream by changing the job config. We can only left a TODO here

lance6716 (approved):

engine/jobmaster/dm/task_manager.go
okJiang: added clear https://github.com/pingcap/tiflow/pull/7594/commits/21db7aae9a06c2703640e71304cdcbf5ff3b2267

engine/jobmaster/dm/task_manager.go
```diff
@@ -105,6 +119,10 @@ func (tm *TaskManager) UpdateTaskStatus(taskStatus runtime.TaskStatus) {
 		zap.Uint64("config_modify_revison", taskStatus.CfgModRevision),
 	)
 	tm.tasks.Store(taskStatus.Task, taskStatus)
+	if _, ok := tm.stageGauge[taskStatus.Task]; !ok {
+		tm.stageGauge[taskStatus.Task] = tm.gaugeVec.WithLabelValues(taskStatus.Task)
+	}
+	tm.stageGauge[taskStatus.Task].Set(float64(taskStatus.Stage))
```
GMHDBJD, line 125: Will this cause problems with concurrent map reads and writes?

okJiang: After using `withLableValues` directly instead of map, do not exist concurrent problem https://github.com/pingcap/tiflow/pull/7594/commits/3d6be7831b89b453431b905a50cbe8a58cb27022

GMHDBJD (approved):
LGTM

Issue #4954, opened 2022-03-20
Reove dead code


hi-rustin:
If we find dead code, we need to clean it up.

asddongmen:
![image](https://user-images.githubusercontent.com/20351731/191652005-1d0e6528-420d-411f-8d3c-b76aa298df53.png)

PR #7595, opened 2022-11-14
scheduler(ticdc): remove v2 scheduler

hi-rustin:

Issue Number: ref https://github.com/pingcap/tiflow/issues/4954

remove v2 scheduler and its tests.

#### Tests 

 - Unit test
 - Integration test

No
No
### Release note 

```release-note
None
```


cdc/scheduler/internal/info_provider.go
```diff
@@ -32,18 +32,4 @@ type InfoProvider interface {
 	// TODO: Remove the API, as capture no longer has local checkpoint ts and
```
hi-rustin, line 32: @overvenus Do we need to remove it this time?

cdc/owner/changefeed.go
```diff
@@ -41,7 +41,7 @@ import (
 	"go.uber.org/zap"
 )
 
-// newSchedulerFromCtx creates a new schedulerV2 from context.
+// newSchedulerFromCtx creates a new schedulerV3 from context.
 // This function is factored out to facilitate unit testing.
```
3AceShowHand, line 45: after `schedulerV2` removed, there is only one scheduler implementation in the cluster, so no need to name it `schedulerV3`, just name it `scheduler` should be ok?

3AceShowHand (approved):

cdc/scheduler/internal/info_provider.go
overvenus: Yes, it's dead code now.

cdc/owner/changefeed.go
hi-rustin: yes. I'll update it. Thanks!

cdc/scheduler/internal/info_provider.go
hi-rustin: Removed! Thanks!

cdc/model/http_model.go
```diff
@@ -190,16 +190,7 @@ type ProcessorCommonInfo struct {
 
 // ProcessorDetail holds the detail info of a processor
 type ProcessorDetail struct {
-	// The maximum event CommitTs that has been synchronized.
-	CheckPointTs uint64 `json:"checkpoint_ts"`
-	// The event that satisfies CommitTs <= ResolvedTs can be synchronized.
-	ResolvedTs uint64 `json:"resolved_ts"`
-	// all table ids that this processor are replicating
```
overvenus, line 197: Please keep the comment.

hi-rustin: Reverted!

asddongmen (approved):

hi-rustin (commented):
Self check

PR #7600, opened 2022-11-14
framework(engine): refine worker factory errors

gozssky:

Issue Number: ref https://github.com/pingcap/tiflow/issues/7475

`SimpleWorkerFactory` is a common worker factory. It should also use common errors.

* Remove `fake.JobUnRetryableError` and define `ErrDeserializeConfig` for `SimpleWorkerFactory`.
* Move out fake worker registration from the registry package.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/framework/registry/factory.go
```diff
@@ -84,13 +84,18 @@ func (f *SimpleWorkerFactory[T, C]) DeserializeConfig(configBytes []byte) (Worke
...
 }
 
 // IsRetryableError implements WorkerFactory.IsRetryableError
 func (f *SimpleWorkerFactory[T, C]) IsRetryableError(err error) bool {
-	var errOut *fake.JobUnRetryableError
-	return !errors.As(err, &errOut)
+	if errors.Is(err, errors.ErrDeserializeConfig) {
+		return false
+	}
+	if strings.Contains(err.Error(), string(errors.ErrDeserializeConfig.RFCCode())) {
```
lance6716, line 97: when will this case be triggered?

lance6716 (approved):

amyangfei (approved):

engine/framework/registry/factory.go
gozssky: https://github.com/pingcap/tiflow/blob/e3cd4604d518f0474c7fe2d0f8fa9688dc18189f/engine/executor/server.go#L255
We may create an error from the error message and check whether it is retryable or not. However, I think it's not a good way to depend on the error message. This may be refactored in the future.

PR #7618, opened 2022-11-15
engine: unit status use cached source status

D3Hunter:

Issue Number: ref #7343

- df-service will call `get job api` which calls `dm status`, but current impl will get source status every time, if the source is down, it won't return after 30s timeout, which will make df-service `get/list job` timeout and fails. so change `unit.Status` to use cached source status.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
`None`.
```

### DIFF
--- a/engine/executor/dm/unitholder.go
+++ b/engine/executor/dm/unitholder.go
@@ -235,6 +235,11 @@ func (u *unitHolderImpl) Stage() (metadata.TaskStage, *pb.ProcessResult) {
 // Status implement UnitHolder.Status. Each invocation will try to query upstream
 // once and calculate the status.
 func (u *unitHolderImpl) Status(ctx context.Context) interface{} {
+	// nil sourceStatus is supported
+	return u.unit.Status(u.getSourceStatus())
+}
+
+func (u *unitHolderImpl) updateSourceStatus(ctx context.Context) interface{} {
 	sourceStatus, err := binlog.GetSourceStatus(
 		tcontext.NewContext(ctx, u.logger),
 		u.upstreamDB,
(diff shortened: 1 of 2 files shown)


engine/executor/dm/unitholder.go
```diff
@@ -243,22 +248,28 @@ func (u *unitHolderImpl) Status(ctx context.Context) interface{} {
...
 // CheckAndUpdateStatus implement UnitHolder.CheckAndUpdateStatus.
 func (u *unitHolderImpl) CheckAndUpdateStatus(ctx context.Context) {
-	u.sourceStatusMu.RLock()
-	sourceStatus := u.sourceStatus
-	u.sourceStatusMu.RUnlock()
+	sourceStatus := u.getSourceStatus()
 
 	if sourceStatus == nil || time.Since(sourceStatus.UpdateTime) > sourceStatusRefreshInterval {
-		u.Status(ctx)
+		u.updateSourceStatus(ctx)
```
lance6716, line 272: Still need to call Status here, because the metrics related variables are updated in Status.

D3Hunter: which metric?

lance6716: for example

```
			s.metricsProxies.Metrics.RemainingTimeGauge.Set(float64(remainingSeconds))
```

this line is contained in (*Syncer).printStatus <- (*Syncer).Status

D3Hunter: fixed here https://github.com/pingcap/tiflow/pull/7618/commits/c55d42df659a1026549397229f95600a2bc45937

bad idea to have a status getter to have such side effects.

okJiang (approved):
LGTM

lance6716 (approved):

Issue #7613, opened 2022-11-15
list job api stuck when jobMaster OOM


CharlesCheung96:
### What did you do?

1. Submit a dm job to engine
2. OOM occurs on the executor node where jobMaster resides
3. Query list_job api in master

### What did you expect to see?

The request returns result or timeout.

### What did you see instead?

The request takes 7 minutes to return:

```shell
$ curl -X GET  http://127.0.0.1:10240/api/v1/jobs/37cb223a-3230-47fd-9c9e-ca88ed4c8b07/status
{"code":"DFLOW:ErrJobNotRunning","message":"job 37cb223a-3230-47fd-9c9e-ca88ed4c8b07 is not running"}

$ time curl -X GET http://127.0.0.1:10240/api/v1/jobs
{"jobs":[{"id":"37cb223a-3230-47fd-9c9e-ca88ed4c8b07", "type":"DM", "state":"Running", "config":"", "detail":"", "error":{"code":"", "message":"[DFLOW:ErrJobManagerGetJobDetailFail]failed to get job detail from job master: Get \"http://172.31.63.16:10241/api/v1/jobs/37cb223a-3230-47fd-9c9e-ca88ed4c8b07/status\": read tcp 172.31.52.193:55466->172.31.63.16:10241: read: connection reset by peer"}, "selectors":[]}], "next_page_token":"37cb223a-3230-47fd-9c9e-ca88ed4c8b07"}curl -X GET http://127.0.0.1:10240/api/v1/jobs  
0.00s user 0.02s system 0% cpu 7:23.03 total

$  curl -X GET  http://127.0.0.1:10240/api/v1/jobs --max-time 10 --connect-timeout 2
curl: (28) Operation timed out after 10001 milliseconds with 0 bytes received
```

### Versions of the cluster

Dataflow Engine version (run `tiflow version`):

```console
Release Version: v6.4.0-master
Git Commit Hash: 5aaf1ddf1f05e33ec35706936f8e52cec8e2563d
Git Branch: remove-clone
UTC Build Time: 2022-11-15 05:51:39
Go Version: go version go1.19.2 linux/amd64
Failpoint Build: false
```

CharlesCheung96:
[...]

PR #7620, opened 2022-11-16
servermaster(engine): add check and timeout for querying job detail

gozssky:

Issue Number: close https://github.com/pingcap/tiflow/issues/7613

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/servermaster/jobmanager.go
```diff
@@ -381,7 +383,7 @@ func (jm *JobManagerImpl) ListJobs(ctx context.Context, req *pb.ListJobsRequest)
 		}
 
 		// if job status is running, forward the request to jobmaster openapi
-		if canQueryJobDetail(masterMetas[i].State) {
+		if jm.canQueryJobDetail(masterMetas[i]) {
 			detail, errJob := jm.jobHTTPClient.GetJobDetail(ctx, masterMetas[i].Addr, masterMetas[i].ID)
```
CharlesCheung96, line 387: Maybe we should use context to control the timeout? Besides, what is the difference between `http client timeout` and `context timeout`?

gozssky: The `http client timeout` is applied for all requests, while `context timeout` only affects one request. There is not much difference between using context timeout or client timeout to control the request timeout. But it's easy to forget to set the context timeout in every place, which can lead to some security issues. This is why I prefer to set the client timeout.

CharlesCheung96 (approved):

amyangfei (approved):

PR #7636, opened 2022-11-17
[DNM]codec(ticdc): fix avro

3AceShowHand:

Issue Number: close #xxx

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note

```

### DIFF
--- a/cdc/sink/codec/avro/avro_test.go
+++ b/cdc/sink/codec/avro/avro_test.go
@@ -17,7 +17,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"math"
 	"math/big"
 	"testing"
 
@@ -61,8 +60,7 @@ func setupEncoderAndSchemaRegistry(
 		namespace:                  model.DefaultNamespace,
 		valueSchemaManager:         valueManager,
 		keySchemaManager:           keyManager,
-		resultBuf:                  make([]*common.Message, 0, 4096),
-		maxMessageBytes:            math.MaxInt,
+		result:                     make([]*common.Message, 0, 4096),
 		enableTiDBExtension:        enableTiDBExtension,
 		decimalHandlingMode:        decimalHandlingMode,
 		bigintUnsignedHandlingMode: bigintUnsignedHandlingMode,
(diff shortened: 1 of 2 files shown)


cdc/sink/codec/avro/avro_test.go
```diff
@@ -61,8 +60,7 @@ func setupEncoderAndSchemaRegistry(
 		namespace:                  model.DefaultNamespace,
 		valueSchemaManager:         valueManager,
 		keySchemaManager:           keyManager,
-		resultBuf:                  make([]*common.Message, 0, 4096),
-		maxMessageBytes:            math.MaxInt,
+		result:                     make([]*common.Message, 0, 4096),
```
hi-rustin, line 63: Do we also need to update here?

hi-rustin:
> * no need to preallocate `4096` for messages, since `Build` will always be called after `AppendRowChangedEvent`, only one message is enough.

I was wondering if we always know that it only has one event, can we use an array instead of a slice? like `[1] *message`

3AceShowHand:
> > * no need to preallocate `4096` for messages, since `Build` will always be called after `AppendRowChangedEvent`, only one message is enough.
> 
> I was wondering if we always know that it only has one event, can we use an array instead of a slice? like `[1] *message`

I've tried to use the array instead of the slice, but it make the code more complicated, so use slice is preferred.

When we call the build method, and consider `a.result` is typed in `[1]*common.Message`:
```
result = a.result[:]
a.result = [1]*common.Message{}
return result
```

This makes the returned value `result` become the `nil`, to fix the problem we have to use `copy`, I think it's no necessary, and to keep the code simple and unified with other codec encoder implementation, so keep it to use slice.

cdc/sink/codec/avro/avro_test.go
```diff
@@ -899,3 +897,47 @@ func TestGetAvroNamespace(t *testing.T) {
 		getAvroNamespace("N-amespace", &model.TableName{Schema: "S.chema", Table: "normalTable"}),
 	)
 }
+
+func TestArvoAppendRowChangedEventWithCallback(t *testing.T) {
+	encoder, err := setupEncoderAndSchemaRegistry(true, "precise", "long")
```
hi-rustin, line 902:
```suggestion
func TestArvoAppendRowChangedEventWithCallback(t *testing.T) {
    t.Parallel()

	encoder, err := setupEncoderAndSchemaRegistry(true, "precise", "long")
```

hi-rustin (approved):

zhaoxinyu (approved):

PR #7640, opened 2022-11-18
dm(engine): declarative command should consider time

lance6716:

Issue Number: ref #4287

If the running "error" stage is updated after expected "running" stage, we skip action.

#### Tests 

 - Integration test

### Release note 

```release-note
 `None`.
```


engine/executor/dm/worker.go
```diff
@@ -263,9 +263,15 @@ func (w *dmWorker) tryUpdateStatus(ctx context.Context) error {
...
-		stage       = w.getStage()
-		code        frameModel.WorkerState
-		taskStatus  = &runtime.TaskStatus{Unit: w.workerType, Task: w.taskID, Stage: stage, CfgModRevision: w.cfgModRevision}
+		stage      = w.getStage()
+		code       frameModel.WorkerState
+		taskStatus = &runtime.TaskStatus{
+			Unit:             w.workerType,
+			Task:             w.taskID,
+			Stage:            stage,
+			StageUpdatedTime: time.Now(),
```
GMHDBJD, line 272: Seems this assign tasks no effect

engine/jobmaster/dm/task_manager.go
```diff
@@ -115,6 +115,7 @@ func (tm *TaskManager) UpdateTaskStatus(taskStatus runtime.TaskStatus) {
 		zap.Stringer("unit", taskStatus.Unit),
 		zap.Uint64("config_modify_revison", taskStatus.CfgModRevision),
 	)
+	taskStatus.StageUpdatedTime = time.Now()
```
GMHDBJD, line 118: Maybe we can save it somewhere other than taskStatus, so we don't have to add unnecessary properties to TaskStatus and change the unit test

lance6716: I prefer Stage and the StageUpdateTime to be together because they have tight relationship, will think about it soon

engine/executor/dm/worker.go
lance6716: I prefer to left this one, remove the write to StageUpdatedTime at jobmaster side

CharlesCheung96 (approved):

engine/executor/dm/worker.go
GMHDBJD: We only update task status when task stage changed, so the StageUpdatedTime may not change for a long time, does this meet user expectations

GMHDBJD (commented):
rest LGTM

GMHDBJD (approved):
LGTM

lance6716:
/test-engine-integration-test


I'll take a look at failing log

lance6716:
/hold

waiting to find a better solution

lance6716:
/run-engine-integration-test

ptal @GMHDBJD

lance6716:
I can say for this time, falling CI should really be a flaky test. The case can be passed in my local PC

```
[Mon Nov 21 18:35:02 CST 2022] <<<<<< run test case e2e_worker_error success! >>>>>>
```

GMHDBJD (approved):
LGTM

amyangfei:
failure caused by slow MySQL

```
server-master-2.log
275:[2022/11/21 11:15:59.732 +00:00] [INFO] [server.go:845] ["leader is initialized"] [took=207.391438ms]
667:[2022/11/21 11:16:39.607 +00:00] [INFO] [server.go:845] ["leader is initialized"] [took=4.55990789s]

server-master-0.log
67:[2022/11/21 11:13:52.868 +00:00] [INFO] [server.go:845] ["leader is initialized"] [took=127.998996ms]

server-master-1.log
404:[2022/11/21 11:15:04.314 +00:00] [INFO] [server.go:845] ["leader is initialized"] [took=169.515992ms]
1027:[2022/11/21 11:16:21.163 +00:00] [INFO] [server.go:845] ["leader is initialized"] [took=133.602022ms]
```

PR #7655, opened 2022-11-21
openapi(engine): add filter and limit to list job

gozssky:

Issue Number: close https://github.com/pingcap/tiflow/issues/7654

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/servermaster/jobmanager.go
```diff
@@ -224,13 +226,25 @@ func (jm *JobManagerImpl) GetJob(ctx context.Context, req *pb.GetJobRequest) (*p
...
-		setDetailToMasterMeta(masterMeta, detail, err)
+		if err != nil {
+			job.Error = &pb.Job_Error{
+				Code:    "", // TODO: extract error code from err.
+				Message: err.Error(),
+			}
+		} else {
+			job.Detail = detail
+		}
 	}
```
amyangfei, line 245: Can extract this logic into a function and reuse it at L412

gozssky: Addressed in latest commits.

engine/servermaster/jobmanager.go
```diff
@@ -482,7 +486,7 @@ func (jm *JobManagerImpl) GetJobMasterForwardAddress(ctx context.Context, jobID
 		}
 		return "", err
 	}
-	if masterMeta.State != frameModel.MasterStateInit || jm.JobFsm.QueryOnlineJob(jobID) == nil {
+	if masterMeta.State != frameModel.MasterStateInit && jm.JobFsm.QueryOnlineJob(jobID) == nil {
```
CharlesCheung96, line 489: Why change the logic here?

gozssky: This's a bug. We shouldn't forward requests to the job master when the job is not online.
It's also the root cause of https://github.com/pingcap/tiflow/issues/7555.

amyangfei: Confused, does `job online` mean
`masterMeta.State == frameModel.MasterStateInit && jm.JobFsm.QueryOnlineJob(jobID) != nil`
Seems the original check is correct

gozssky, line 489:
```suggestion
	if masterMeta.State != frameModel.MasterStateInit || jm.JobFsm.QueryOnlineJob(jobID) == nil {
```

gozssky: Fixed.

amyangfei (approved):

CharlesCheung96 (approved):

Issue #7619, opened 2022-11-16
Tracking issue for improving code structure and abstraction


gozssky:
## Motivation

The current structure and abstraction of engine are not good enough, and it is hard to maintain and extend. This issue is used to track the improvement. It is a long-term issue, and we will split it into multiple issues. We will focus on the following aspects:
* master/worker abstraction
* executor runtime enhancement
* better structure for maintaining and testing

## Task List

- [x] #7695
- [ ] Do not embed `BaseXXX` in `XXXImpl`
- [ ] Refine master/worker interface naming
- [ ] Recheck [dig](https://github.com/uber-go/dig) usage, remove or refactor

PR #7663, opened 2022-11-21
proto(engine): rename and remove some useless rpc

CharlesCheung96:

Issue Number: ref #7619

1. rename and remove some useless rpc 

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


engine/proto/master.proto
```diff
@@ -121,11 +114,10 @@ message Selector {
 
 message HeartbeatRequest {
     string executor_id = 1;
-    int32 resource_usage = 2;
-    int32 status = 3;
+    int32 status = 2;
```
CharlesCheung96, line 117: Should the status field be removed?

gozssky, line 117: `status` could also be removed.

engine/proto/master.proto
```diff
@@ -136,8 +128,7 @@ message Executor {
     // name is the readable name of the executor.
     string name = 2;
     string address = 3;
-    int64  capability = 4;
```
CharlesCheung96, line 139: Should the capability field be removed?

engine/proto/master.proto
gozssky: +1

engine/proto/master.proto
gozssky: +1

engine/proto/master.proto
amyangfei: Better to remove it, it is not used now.

engine/proto/master.proto
```diff
@@ -169,8 +160,8 @@ message ListMastersResponse {
 
 message ScheduleTaskRequest {
     string task_id = 1;
-    int64 cost = 2;
-    repeated ResourceKey resource_requirements = 3;
+    // resources required by the task.
+    repeated ResourceKey resources = 3;
     repeated Selector selectors = 4;
```
amyangfei, line 165: repeated ResourceKey resources = 2;
    repeated Selector selectors = 3;

gozssky (approved):

amyangfei (approved):

PR #7697, opened 2022-11-23
engine/dm: period update status for dm worker

GMHDBJD:

Issue Number: ref #5504 

- period update status for dm worker

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


engine/executor/dm/worker.go
```diff
@@ -165,9 +174,15 @@ func (w *dmWorker) Tick(ctx context.Context) error {
 	// update unit status periodically to update metrics
 	w.unitHolder.CheckAndUpdateStatus()
 	w.discardResource4Syncer(ctx)
+	w.DoTick(ctx)
 	return w.messageAgent.Tick(ctx)
 }
 
+func (w *dmWorker) TickImpl(ctx context.Context) error {
+	status := w.workerStatus(ctx)
+	return w.UpdateStatus(ctx, status)
```
lance6716, line 183: this is to avoid UpdateStatus message is dropped so ExtBytes always empty?
Can we change the logic of `tryUpdateStatus`?

GMHDBJD: 1. Yes
2. I think we don't need change tryUpdateStatus, still update it when stage changed

lance6716: OK we can separate trigger actions and periodical actions, but the name of `tryUpdateStatus` can be changed to `updateStatusWhenStageChange` or something.

please help update the comment, now OnWorkerOffline can appear without OnWorkerOnline https://github.com/pingcap/tiflow/blob/f7b84c7683f31f7b856d2c65bec5ec1b53c6065b/engine/framework/master.go#L107

amyangfei (approved):

lance6716 (approved):

PR #7707, opened 2022-11-24
sink(cdc): cancel sink operations when network becomes unavailable

### LINKED ISSUE #7706: Sink operations can't be canceled correctly if network becomes unavailable
### What did you do?

1. Create a cdc capture with a MySQL downstream;
2. Drop all network packages to the downstream;
3. Run a workload on the upstream cluster.

CDC can detect that network between the downstream and itself becomes broken. And then try to re-establish connections.

### What did you see instead?

CDC can't detect the network problem. All sink workers are blocked.
图片

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

hicqu:

Issue Number: close #7706 .

Set a timeout for `context`s used in sink operations.

Another alternative is set [wait-timeout](https://dev.mysql.com/doc/refman/5.6/en/server-system-variables.html#sysvar_wait_timeout) on mysql sessions. But I'm not sure whether both TiDB and MySQL support it or not.
So I fix it in golang layer.

#### Why `writeTimeout` and `readTimeout` doesn't work on network partition?
These parameters make server side possible to shut connections which are busy at read-write operations. But when network partition happens, client side can know nothing about this.

#### Tests 

### Release note 

```release-note
None
```


cdc/sinkv2/eventsink/txn/mysql/mysql.go
```diff
@@ -482,15 +485,20 @@ func (s *mysqlBackend) prepareDMLs() *preparedDMLs {
...
 	}
 
 	start := time.Now()
-	return retry.Do(ctx, func() error {
+	return retry.Do(pctx, func() error {
+		writeTimeout, _ := time.ParseDuration(s.cfg.WriteTimeout)
+		writeTimeout += networkDriftDuration
+		ctx, cancelFunc := context.WithTimeout(pctx, writeTimeout)
+		defer cancelFunc()
+
```
nongfushanquan, line 501: why not inject some timeout  fail point

cdc/sink/mysql/mysql.go
```diff
@@ -51,6 +51,9 @@ const (
 	backoffBaseDelayInMs = 500
 	// in previous/backoff retry pkg, the DefaultMaxInterval = 60 * time.Second
 	backoffMaxDelayInMs = 60 * 1000
+
+	// networkDriftDuration is used to construct a context timeout for database operations.
+	networkDriftDuration = 5 * time.Second
```
nongfushanquan, line 56: Is 5 enough since the normal duration often exceed 8s ?

hicqu: It will be added with `writeTimeout`.

hi-rustin (approved):

PR #7715, opened 2022-11-25
binlog_filter: use table name instead of table id as cache key

GMHDBJD:

Issue Number: ref #4287 

- use schema/table name as cache key to filter binlog

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


lance6716 (approved):
daily upstream_switch GitHub CI will test the result

buchuitoudegou (approved):

Issue #7720, opened 2022-11-25
Tracking issue for scale-out single large table to multiple TiCDC instances


overvenus:
This project add the ability of scale-out of a single table to multiple TiCDC nodes. Let TiCDC support dynamically split and merge spans based on statistical data, including region count, write throughput, etc. Split and merge are just table scheduling, and with the help of two-phase scheduling, they have minimal impact on replication latency.

## Span Replication

- [x] https://github.com/pingcap/tiflow/pull/7722
- [x] https://github.com/pingcap/tiflow/pull/7723
- [x] https://github.com/pingcap/tiflow/pull/7731
- [x] https://github.com/pingcap/tiflow/pull/7793
- [x] https://github.com/pingcap/tiflow/pull/7806
- [x] https://github.com/pingcap/tiflow/pull/7832
- [x] https://github.com/pingcap/tiflow/pull/7833
- [x] https://github.com/pingcap/tiflow/pull/7838
- [x] https://github.com/pingcap/tiflow/pull/7880
- [x] https://github.com/pingcap/tiflow/pull/7909
- [x] https://github.com/pingcap/tiflow/pull/7916
- [x] https://github.com/pingcap/tiflow/pull/7978
- [x] https://github.com/pingcap/tiflow/pull/8010
- [ ] Unify variable names, rename table to tableSpan
- [x] https://github.com/pingcap/tiflow/pull/8021
- [x] https://github.com/pingcap/tiflow/pull/8070
- [x] https://github.com/pingcap/tiflow/pull/8100

### Support split span by write throughput

- [x] https://github.com/pingcap/tiflow/pull/8496
- [x] https://github.com/pingcap/tiflow/pull/8505

### Document

- [x] https://github.com/pingcap/docs-cn/pull/12693
- [x] https://github.com/pingcap/docs-cn/pull/13341

PR #7722, opened 2022-11-25
processor,pkg(ticdc): add Span and package spanz

overvenus:

* Add protobuf message `Span`, it will replace table ID in TiCDC code base, and will become the ID of table pipeline.
* Add scaffolds around `Span`, they are dead code for now, will be used in later PRs.
  *  A ordered map of `Span`.
  * A set of `Span`.

Issue Number: ref #7720

#### Tests 

 - Unit test

No.

No.

### Release note 

```release-note
None
```


pkg/spanz/cov.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2022 PingCAP, Inc.
```
3AceShowHand, line 1: what does the file name `cov` mean ?

3AceShowHand (approved):

pkg/spanz/cov.go
overvenus: It means convert. Will update the file name.

asddongmen (approved):

PR #7723, opened 2022-11-25
 scheduler(ticdc): add compat layer

overvenus:
### What problem does this PR solve?

The compact layer makes span replication be compatible with table replication.

Before sending messages from span replication to table replication:

 - span agent -> table scheduler
   - sets tableID base on span.TableID
 - span scheduler -> table agent
   - sets tableID base on span.TableID

After receiving messages from table replication:
  
 - table scheduler -> span agent
   - Fill span based on table ID if span is empty
 - table agent -> span scheduler
   - Fill span based on table ID if span is empty

Issue Number: ref #7720

#### Tests 

 - Unit test

##### Will it cause performance regression or break compatibility?

No.

##### Do you need to update user documentation, design documentation or monitoring documentation?

No.

### Release note 

```release-note
None
```

PR #7793, opened 2022-12-02
cdc: replace tableID with Span in TableExecutor and agent 

overvenus:

Issue Number: ref #7720

* Replace TableID with tablepb.Span in TableExecutor
* Replace TableID with tablepb.Span in agent

This PR is compatible with older TiCDC, because #7723 convert table ID and span in message automatically.

#### Tests 

 - Unit test
 - Integration test

No

No

### Release note 

```release-note
None
```


cdc/processor/processor.go
```diff
@@ -407,53 +410,49 @@ func (p *processor) IsRemoveTableFinished(tableID model.TableID) (model.Ts, bool
...
 		zap.String("changefeed", p.changefeedID.ID),
-		zap.Int64("tableID", tableID),
+		zap.Int64("tableID", span.TableID),
 		zap.Uint64("checkpointTs", checkpointTs))
 	return checkpointTs, true
 }
 
 // GetAllCurrentTables implements TableExecutor interface.
-func (p *processor) GetAllCurrentTables() []model.TableID {
+func (p *processor) GetAllCurrentTables() int {
```
3AceShowHand, line 451: Is it possible that we care about all tableID ?

How about adding a new method `GetAllTableCount` here?

cdc/processor/processor.go
```diff
@@ -135,17 +137,17 @@ func (p *processor) AddTable(
 			zap.String("captureID", p.captureInfo.ID),
 			zap.String("namespace", p.changefeedID.Namespace),
 			zap.String("changefeed", p.changefeedID.ID),
-			zap.Int64("tableID", tableID),
+			zap.Int64("tableID", span.TableID),
```
3AceShowHand, line 140: It looks a lot of modification about replace the `tableID` to `span.TableID`.

How about add a local variable `tableID := span.TableID` first to reduce modifications ?

cdc/processor/processor.go
overvenus: I checked code base, no one needs tableID. I will rename the method.

cdc/processor/processor.go
overvenus: The log field will be replaced `zap.Stringer("span", span)` in later PR. I think it's okay for now, and ease my work in later PR (just search and replace).

cdc/scheduler/internal/table_executor.go
```diff
@@ -30,33 +30,33 @@ type TableExecutor interface {
...
+		ctx context.Context, span tablepb.Span, startTs model.Ts, isPrepare bool,
 	) (done bool, err error)
 
 	// IsAddTableFinished make sure the requested table is in the proper status
-	IsAddTableFinished(tableID model.TableID, isPrepare bool) (done bool)
+	IsAddTableFinished(span tablepb.Span, isPrepare bool) (done bool)
 
 	// RemoveTable remove the table, return true if the table is already removed
-	RemoveTable(tableID model.TableID) (done bool)
+	RemoveTable(span tablepb.Span) (done bool)
```
sdojjy, line 40: rename the function name?  it's span related func now

overvenus: Good point, thanks!

overvenus:
kafka tests timeout, but there is no ERROR or WARN log from schedulerv3.

PR #7795, opened 2022-12-02
sinkv2(cdc): reduce latency for the case with lots of tables

### LINKED ISSUE #7778: cdc lag >1min
### What did you do?

/cdc  cli  changefeed  create --pd http://xxxx --changefeed-id xxxxx--sink-uri tidb://xxxxxx

_No response_

### What did you see instead?

checkpoint lag >1min
image

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

/ # /cdc version
Release Version: v6.5.0-alpha
Git Commit Hash: bc176c35e40c6dfb3a386097a4e189b37639ef4a
Git Branch: heads/refs/tags/v6.5.0-alpha
UTC Build Time: 2022-12-02 02:55:53
Go Version: go version go1.19.3 linux/amd64
Failpoint Build: false

```console
(paste TiCDC version here)
```

hicqu:

Issue Number: close #7778 .

Reduce latency for the case with lots of tables.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/processor/sinkmanager/manager_test.go
```diff
@@ -401,11 +401,6 @@ func TestGetTableStats(t *testing.T) {
 		return checkpointTS.ResolvedMark() == 3
 	}, 5*time.Second, 10*time.Millisecond)
 
-	cleanPos := engine.Position{
```
hi-rustin, line 404: Please keep this assertion. Otherwise, this test would useless.

cdc/processor/sinkmanager/manager.go
```diff
@@ -287,21 +282,47 @@ func (m *SinkManager) backgroundGC() {
...
+							zap.Int64("tableID", tableID),
 							zap.Error(err))
+						select {
+						case m.errChan <- err:
+						default:
+							log.Error("Failed to send error to error channel, error channel is full",
+								zap.String("namespace", m.changefeedID.Namespace),
+								zap.String("changefeed", m.changefeedID.ID),
+								zap.Error(err))
+						}
```
amyangfei, line 317: What is the best behavior when error happens in the background goroutine
- to keep the goroutine running
- or let it return.

In other word, if the error is sent to the `errChan` or there exists an error in it already, the parent context will be canceled later, the opportunity of the logic in these background goroutines is indeterminate.

hicqu:
/run-integration-test tidb=f5487e3f2596cdd63ae9840606c7fff3fb539e03

cdc/processor/sinkmanager/manager_test.go
hicqu: `GetTableStats` won't drive GC now. So we can remove this from the case.

cdc/processor/sinkmanager/manager.go
```diff
@@ -310,20 +328,34 @@ func (m *SinkManager) backgroundGC() {
...
+	}
+
+	dispatchTasks := func() error {
+		tables := make([]*tableSinkWrapper, 0, sinkWorkerNum)
+		progs := make([]*progress, 0, sinkWorkerNum)
+
+		// Collect some table progresses.
+		for len(tables) < sinkWorkerNum && m.sinkProgressHeap.len() > 0 {
 			slowestTableProgress := m.sinkProgressHeap.pop()
 			tableID := slowestTableProgress.tableID
```
nongfushanquan, line 356: would be slowestTableProgress empty?

hicqu: The `for` condition has tested it.

cdc/processor/sinkmanager/manager.go
```diff
@@ -345,90 +379,108 @@ func (m *SinkManager) generateSinkTasks() error {
...
 			case m.sinkTaskChan <- t:
+				log.Debug("Generate sink task",
+					zap.String("namespace", m.changefeedID.Namespace),
+					zap.String("changefeed", m.changefeedID.ID),
+					zap.Int64("tableID", tableSink.tableID),
+					zap.Any("lowerBound", lowerBound),
+					zap.Any("currentUpperBound", upperBound))
+			default:
+				m.memQuota.refund(requestMemSize)
+				break LOOP
```
amyangfei, line 440: `m.sinkTaskChan` is an unbuffered channel, if sink worker is running in `handleTask`, this sink task will be ignored, is it as expected or something will be lost?

cdc/processor/sinkmanager/manager_test.go
hi-rustin: So we need to change or add a test case to cover this logic.

hi-rustin: If you don't have time to add it. At least we need to delete this test. Because it's just for GC logic.

cdc/processor/sinkmanager/tasks.go
```diff
@@ -38,7 +38,7 @@ var (
 type writeSuccessCallback func(lastWrittenPos engine.Position)
 
 // Used to get an upper bound.
-type upperBoundGetter func() engine.Position
+type upperBoundGetter func(*tableSinkWrapper) engine.Position
```
hi-rustin, line 41: Could you please make this func be a method of task? Because it is wried to pass take sink back to this func, it already has table sink pointer in task struct. We can just add `getUpperBoundForSink` and `getUpperBoundForRedo` to the task struct to get the upper bound ts.

cdc/processor/sinkmanager/manager.go
hicqu: It's expected. The task will be return into `progressHeap`.

amyangfei (approved):

cdc/processor/sinkmanager/manager_test.go
hicqu: Let's delete the case for now.

cdc/processor/sinkmanager/tasks.go
hi-rustin: It can be a pointer. Atomic pointer for barrier ts.

nongfushanquan (approved):

hi-rustin (approved):

PR #7796, opened 2022-12-02
[WIP]checker(dm): add a worker pool

lance6716:

Issue Number: close #xxx

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note

```

### DIFF
--- a/dm/pkg/checker/worker_pool.go
+++ b/dm/pkg/checker/worker_pool.go
@@ -1,0 +1,114 @@
+// Copyright 2022 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checker
+
+import (
+	"context"
+
+	"golang.org/x/sync/errgroup"
+)
+
+// WorkerPool is a easy-to-use worker pool that can start workers by Go then use
+// PutJob to send jobs to worker. After worker finished a job, the result is
+// sequentially called by resultHandler function which is the parameter of
+// NewWorkerPool or NewWorkerPoolWithContext. After caller send all jobs, it can
+// call Wait to make sure all jobs are finished.
+// The type parameter J means job, R means result. Type J MUST only share
+// concurrent-safe member like *sql.DB.
+type WorkerPool[J, R any] struct {
+	ctx context.Context
+
+	// the closing order is inCh -> outCh -> done
+	inCh     chan J
+	outCh    chan R
+	done     chan struct{}
+	errGroup *errgroup.Group
+}
+
+// NewWorkerPool creates a new worker pool.
+// The type parameter J means job, R means result. Type J MUST only share
+// concurrent-safe member like *sql.DB.
+func NewWorkerPool[J, R any](resultHandler func(R)) *WorkerPool[J, R] {
+	return NewWorkerPoolWithContext[J, R](context.Background(), resultHandler)
+}
+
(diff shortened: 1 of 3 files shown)


dm/pkg/checker/table_structure.go
```diff
@@ -110,11 +107,97 @@ func NewTablesChecker(
...
 
+type tablesCheckerWorker struct {
+	c                *TablesChecker
+	downstreamParser *parser.Parser
+
+	lastSourceID   string
+	upstreamParser *parser.Parser
+}
+
+func (w *tablesCheckerWorker) handle(ctx context.Context, checkItem *checkItem) ([]*incompatibilityOption, error) {
```
lance6716, line 121: moved from `(*TablesChecker).startWorker`

dm/pkg/checker/worker_pool.go
```diff
@@ -0,0 +1,115 @@
...
+	return NewWorkerPoolWithContext[J, R](context.Background(), resultHandler)
+}
+
+// NewWorkerPoolWithContext creates a new worker pool with a context which may
+// be canceled from caller.
+// The type parameter J means job, R means result. Type J MUST only share
+// concurrent-safe member like *sql.DB.
+func NewWorkerPoolWithContext[J, R any](
+	ctx context.Context,
+	resultHandler func(R),
```
D3Hunter, line 52: - maybe add `concurrency` param here
- no need a public `Go`, loop and call `Go` is common pattern
- maybe let J impl this interface
```go
type JobI interface {
    Run(ctx) R
}
```

lance6716: > maybe add concurrency param here
> no need a public Go, loop and call Go is common pattern

seems now the worker pool will spawn workers, but if one worker need init logic for multiple jobs, like create a downstream parser, the init logic should also be passed to `NewWorkerPoolWithContext` which is hard to design.

Now we let the user go init logic in for-loop then call `Go`

> maybe let J impl this interface

it's also related to above thing. Choosing let worker owns `run` rather than job owns `run` can have worker init logic for multiple jobs

D3Hunter: > but if one worker need init logic for multiple jobs, , like create a downstream parser

why let worker-pool care about job init? let job do it or we init the parser for job before `PutJob`

lance6716: it's worker init logic, not job init. For example, there're 16 workers and each should get a downstream parser. The number of job is not determined, we only know the worker numer is 16.

Also job has its own init logic, such as upstream DB connection pool, but worker init is another thing

dm/pkg/checker/table_structure.go
```diff
@@ -131,30 +214,30 @@ func (c *TablesChecker) Check(ctx context.Context) *Result {
...
+	dispatchTableItemWithDownstreamTable(c.tableMap, pool.PutJob)
+
+	if err := pool.Wait(); err != nil {
 		markCheckError(r, err)
-		c.reMu.Unlock()
+		return r
 	}
-	close(c.optCh)
-	optWg.Wait()
+	finalHandler()
```
D3Hunter, line 240: call it with `defer`?

dm/pkg/checker/table_structure.go
```diff
@@ -949,23 +929,23 @@ func dispatchTableItem(ctx context.Context, tableMap map[string][]filter.Table,
 }
 
 func dispatchTableItemWithDownstreamTable(
-	ctx context.Context,
 	tableMaps map[string]map[filter.Table][]filter.Table,
-	inCh chan *checkItem,
+	putJobFn func(*checkItem) bool,
```
D3Hunter, line 933: why not pass worker-pool directly

D3Hunter (commented):
rest lgtm

dm/pkg/checker/table_structure.go
lance6716: though its name is "finalXXX", it is a normal logic rather than some cleanup logic. It will append common message to the result where the message is calculated from all jobs, so if the result has been marked failed we can skip it.

buchuitoudegou (approved):

PR #7815, opened 2022-12-06
pkg(dm): modify the type of pos returned by function GetMasterStatus

### LINKED ISSUE #7483: Returned pos of function GetMasterStatus should be int64 instead of uint32
### What did you do?

https://github.com/pingcap/tiflow/blob/6ad967659ba98065994f8cb79415912a408abb28/dm/pkg/conn/utils.go#L101-L103

The second returned parameter `pos` should be `int64` instead of `uint32`, to conform to MySQL's behavior:
https://github.com/mysql/mysql-server/blob/a246bad76b9271cb4333634e954040a970222e0a/sql/rpl_source.cc#L1225-L1226

### What did you see instead?

The second returned parameter `pos` is `uint32`.

### Versions of the cluster

```console
v6.3.0-master-dirty
```

_No response_

liumengya94:

Issue Number: close #7483

Returned pos of function GetMasterStatus should be `uint64` instead of `uint32`
#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```

### DIFF
--- a/dm/pkg/conn/utils.go
+++ b/dm/pkg/conn/utils.go
@@ -100,11 +100,11 @@ func getVariable(ctx *tcontext.Context, conn *BaseConn, variable string, isGloba
 // GetMasterStatus gets status from master.
 // When the returned error is nil, the gtid must be not nil.
 func GetMasterStatus(ctx *tcontext.Context, db *BaseDB, flavor string) (
-	string, uint32, string, string, string, error,
+	string, uint64, string, string, string, error,
 ) {
 	var (
 		binlogName     string
-		pos            uint32
+		pos            uint64
 		binlogDoDB     string
 		binlogIgnoreDB string
 		gtidStr        string
@@ -166,7 +166,7 @@ func GetMasterStatus(ctx *tcontext.Context, db *BaseDB, flavor string) (
 				err = terror.DBErrorAdapt(err, terror.ErrDBDriverError)
 				return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
 			}
-			pos = uint32(posInt)
+			pos = uint64(posInt)
 			binlogDoDB = rowsResult[0][2]
 			binlogIgnoreDB = rowsResult[0][3]
 			gtidStr = rowsResult[0][4]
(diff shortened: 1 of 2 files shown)


dm/pkg/conn/utils.go
```diff
@@ -166,7 +166,7 @@ func GetMasterStatus(ctx *tcontext.Context, db *BaseDB, flavor string) (
 				err = terror.DBErrorAdapt(err, terror.ErrDBDriverError)
 				return binlogName, pos, binlogDoDB, binlogIgnoreDB, gtidStr, err
 			}
-			pos = uint32(posInt)
+			pos = uint64(posInt)
```
lance6716, line 169: in above lines, we can parse it into uint64

```
var posInt int64
posInt, err = strconv.ParseInt(rowsResult[0][1], 10, 64)
```

dm/pkg/conn/utils.go
```diff
@@ -194,7 +194,7 @@ func GetMasterStatus(ctx *tcontext.Context, db *BaseDB, flavor string) (
...
```
lance6716, line 197: also here

dm/pkg/conn/utils.go
```diff
@@ -234,7 +234,7 @@ func GetPosAndGs(ctx *tcontext.Context, db *BaseDB, flavor string) (
 	}
 	binlogPos = gmysql.Position{
 		Name: binlogName,
-		Pos:  pos,
+		Pos:  uint32(pos),
```
lance6716, line 237: maybe we can log a warning here, there's a logger in `ctx *tcontext.Context`

dm/pkg/conn/utils_test.go
```diff
@@ -72,7 +72,7 @@ func TestGetMasterStatus(t *testing.T) {
 	cases := []struct {
 		rows           *sqlmock.Rows
 		binlogName     string
-		pos            uint32
+		pos            uint64
```
lance6716, line 75: please add a test for values in uint64 range

dm/pkg/conn/utils.go
```diff
@@ -234,9 +234,9 @@ func GetPosAndGs(ctx *tcontext.Context, db *BaseDB, flavor string) (
...
 	}
-
+	ctx.L().Warn("Convert the pos type returned by GetMasterStatus to uint32 to fit binlog position")
```
lance6716, line 239: seems we can only log a warning when the value is larger than max value of uint32 of error here 🤔 I prefer warning because it may not in a critical path

lance6716 (commented):
rest lgtm

lance6716 (approved):

D3Hunter (approved):

PR #7838, opened 2022-12-07
sinkv2,processor: replace table ID with Span

overvenus:

Issue Number: ref #7720 

* Replace TableID with tablepb.Span in sink v2

#### Tests 

 - Unit test
 - Integration test

No

No

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/config/debug.go
+++ b/pkg/config/debug.go
@@ -15,6 +15,7 @@ package config
 
 import (
 	"github.com/pingcap/errors"
+	cerror "github.com/pingcap/tiflow/pkg/errors"
 	cerrors "github.com/pingcap/tiflow/pkg/errors"
 )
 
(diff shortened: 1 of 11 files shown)


3AceShowHand (approved):

sdojjy (approved):

pkg/config/debug.go
```diff
@@ -61,6 +62,16 @@ func (c *DebugConfig) ValidateAndAdjust() error {
 	if err := c.Scheduler.ValidateAndAdjust(); err != nil {
 		return errors.Trace(err)
 	}
+	if c.Scheduler.RegionPerSpan != 0 {
+		if c.EnablePullBasedSink || !c.EnableNewSink {
```
hi-rustin, line 66: The `debug.enable-pull-based-sink` default is true. Does this mean you want to implement it in the table pipeline first?

overvenus: Yes, for now, I think table pipeline is more stable for testing this feature.

Issue #7376, opened 2022-10-17
SQL connection should tolerate being killed by idle too long


lance6716:
### What did you do?

Currently when the connection is idle for a long time and be killed, DM will raise error when use it afterwards and triggers an alert

https://asktug.com/t/topic/993973

### What did you see instead?

Auto recover from this kind of error

### Versions of the cluster

master

### current status of DM cluster (execute `query-status <task-name>` in dmctl)

```console
(paste current status of DM cluster here)
```

Never-NG:
This situation alse exists in the dumping and loading stage

lance6716:
another report https://github.com/pingcap/tiflow/issues/5804

Issue #3734, opened 2021-08-25
Unnecessary rollback and Fail to drop table from schema tracker when pause task


GMHDBJD:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
```
insert into tb1 values (1)
insert into tb2 values (2)
stop-task
start-task
insert into tb1 values(3)
pause-task
```
2. What did you expect to see?
no need to rollback, no error log
3. What did you see instead?
table 2 failed to rollback
```
[2021/08/24 23:01:30.377 -04:00] [DEBUG] [checkpoint.go:691] ["try to rollback checkpoint"] [task=test] [unit="binlog replication"] [component="remote checkpoint"] [schema=shardddl1] [table=tb1] [checkpoint="position: (mysql-bin.000001, 1990), gtid-set: (flushed position: (mysql-bin.000001, 1990), gtid-set: )"]
[2021/08/24 23:01:30.378 -04:00] [DEBUG] [checkpoint.go:691] ["try to rollback checkpoint"] [task=test] [unit="binlog replication"] [component="remote checkpoint"] [schema=shardddl1] [table=tb2] [checkpoint="position: (mysql-bin.000001, 1735), gtid-set: (flushed position: (mysql-bin.000001, 1735), gtid-set: )"]
[2021/08/24 23:01:30.378 -04:00] [INFO] [checkpoint.go:694] ["rollback checkpoint"] [task=test] [unit="binlog replication"] [component="remote checkpoint"] [schema=shardddl1] [table=tb2] [from="position: (mysql-bin.000001, 1735), gtid-set: "] [to="position: (mysql-bin.000001, 1735), gtid-set: "]
[2021/08/24 23:01:30.378 -04:00] [WARN] [checkpoint.go:697] ["failed to drop table from schema tracker"] [task=test] [unit="binlog replication"] [component="remote checkpoint"] [schema=shardddl1] [table=tb2] [error="[schema:1146]Table 'shardddl1.tb2' doesn't exist"]
```
4. Root Cause. 
[...]

PR #7852, opened 2022-12-08
metric(dm): add resumable_err label to unit error count metric

D3Hunter:

Issue Number: ref #7115, #7376

- we need alert customer on dump/load/sync error, but some-error can be resumed automatically, we don't want alert too frequently, so i add a `resumable_err` label to all `XXX_exit_with_error_count` metric
- uses `changes(dm_syncer_exit_with_error_count{resumable_err="false"}[1m]) > 0 or on(source_id, task) increase(dm_syncer_exit_with_error_count{resumable_err="true"}[1m]) > 3` to alert, other `XXX_exit_with_error_count` too

left is un-resumable err, right is resumable
image

#### Tests 

 - Unit test
 - Integration test
   - alert expression is tested manually, see prev image
 - No code

### Release note 

```release-note
adjust alert expression of `DM_XXX_process_exits_with_error` alerts to reduce false alarms
```


dm/syncer/syncer.go
```diff
@@ -3165,10 +3175,12 @@ func (s *Syncer) Resume(ctx context.Context, pr chan pb.ProcessResult) {
 	var err error
 	defer func() {
 		if err != nil {
+			processError := unit.NewProcessError(err)
+			s.handleExitErrMetric(processError)
```
D3Hunter, line 3179: syncer will prepare db before `Process` on resume, the error should be counted

dm/loader/lightning.go
```diff
@@ -431,7 +438,9 @@ func (l *LightningLoader) Process(ctx context.Context, pr chan pb.ProcessResult)
 
 	if err := l.restore(ctx); err != nil && !utils.IsContextCanceledError(err) {
 		l.logger.Error("process error", zap.Error(err))
-		errs = append(errs, unit.NewProcessError(err))
+		processError := unit.NewProcessError(err)
+		l.handleExitErrMetric(processError)
```
D3Hunter, line 442: add metric here too

D3Hunter:
@lance6716 @buchuitoudegou ptal

dm/syncer/syncer.go
```diff
@@ -745,9 +754,10 @@ func (s *Syncer) Process(ctx context.Context, pr chan pb.ProcessResult) {
 			s.tctx.L().Info("filter out error caused by user cancel", log.ShortError(err))
 		} else {
 			s.tctx.L().Debug("unit syncer quits with error", zap.Error(err))
-			s.metricsProxies.Metrics.SyncerExitWithErrorCounter.Inc()
+			processError := unit.NewProcessError(err)
+			s.handleExitErrMetric(processError)
```
lance6716, line 758: maybe move to line 771, before we sent to `pr`

lance6716 (approved):

dm/metrics/alertmanager/dm_worker.rules.yml
```diff
@@ -24,11 +24,11 @@ groups:
       summary: DM remain storage of relay log
 
   - alert: DM_relay_process_exits_with_error
-    expr: changes(dm_relay_exit_with_error_count[1m]) > 0
+    expr: changes(dm_relay_exit_with_error_count{resumable_err="false"}[1m]) > 0 or on(instance, job) increase(dm_relay_exit_with_error_count{resumable_err="true"}[1m]) > 3
```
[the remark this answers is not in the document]
D3Hunter: change `expr` to `increase(dm_relay_exit_with_error_count{resumable_err="true"}[2m]) > 3` in case the resumeable error is a `connection timeout` which takes 30s, @lance6716 ptal

dm/metrics/alertmanager/dm_worker.rules.yml
```diff
@@ -68,33 +68,33 @@ groups:
...
       summary: DM load process exists with error
 
   - alert: DM_sync_process_exists_with_error
-    expr: changes(dm_syncer_exit_with_error_count[1m]) > 0
+    expr: changes(dm_syncer_exit_with_error_count{resumable_err="false"}[1m]) > 0 or on(source_id, task) increase(dm_syncer_exit_with_error_count{resumable_err="true"}[2m]) > 3
     labels:
       env: ENV_LABELS_ENV
       level: critical
-      expr: changes(dm_syncer_exit_with_error_count[1m]) > 0
+      expr: changes(dm_syncer_exit_with_error_count{resumable_err="false"}[1m]) > 0 or on(source_id, task) increase(dm_syncer_exit_with_error_count{resumable_err="true"}[2m]) > 3
```
D3Hunter, line 97: change `expr` to `increase(xxxxxxx[2m]) > 3` in case the resumeable error is a `connection timeout` which takes 30s, @lance6716 ptal

lance6716: I'm not sure the pause and resume action run quickly enough so that 3 resuming can happen in 2 minutes. There's another rule "DM_binlog_file_gap_between_master_syncer", will the service use it as well?

Also our document should be updated https://docs.pingcap.com/zh/tidb/stable/dm-handle-alerts

D3Hunter: if we met resumable error for the first time or the first time after long `running` state, `pause/resume` will happen every 5 seconds in default check interval. which type of `pause/resume` is even slower than `connection timeout`?

lance6716: https://github.com/pingcap/tiflow/issues/3734#issuecomment-929070540

D3Hunter:
@okJiang ptal

D3Hunter:
@buchuitoudegou ptal

PR #7902, opened 2022-12-13
test(dm): add foreign key test

liumengya94:

Issue Number: ref #7598

- append `foreign_key_checks=0` in `dsn` if user didn't configure
- add a test to check the inconsistency of upstream and downstream data when tidb set `foreign_key_checks=0` 

#### Tests 

 - Integration test

No
No
### Release note 

```release-note
use foreign_key_checks=0 for incremental replication phase, which may cause data inconsistency
```


lance6716 (commented):
rest lgtm. please update the release note part of PR description

lance6716 (approved):

D3Hunter (approved):

PR #7962, opened 2022-12-22
tls(dm): fix the error when only ssl-ca is configured

### LINKED ISSUE #7941: dm failed to run when only `ssl-ca` is configured
### What did you do?

- target db is tidbcloud tidb cluster
- use ssl to connect target db, tidbcloud only has `ca` part
- start-task, and fail with `"Message": "[Lightning:DB:ErrDBConnect]failed to connect database: could not load client key pair: tls: failed to find any PEM data in certificate input",`

bug was introduced in https://github.com/pingcap/tiflow/pull/3260, if user didn't specify some `ssl-xxx`(like in here, there's only `ssl-ca` config, no `ssl-cert` and `ssl-key`), we should not dump it.
https://github.com/pingcap/tiflow/blob/45ee743ccbea727a3c5d360a1295e51d61acd1c4/dm/config/security/security.go#L105-L109

task cfg
```
name: test
task-mode: all

target-database:
  host: "xxxxxx.clusters.dev.tidb-cloud.com"
  port: 4000
  user: "root"
  password: "xxx"
  security:
    ssl-ca: "tidbcloud-ca.pem"

mysql-instances:
  - source-id: "mysql-3306"
    block-allow-list: "ba-rule1"
    syncer-config-name: "global"

block-allow-list:
  ba-rule1:
    do-dbs: ["test"]

syncers:
  global:
    worker-count: 1
    batch: 1
```

success

### What did you see instead?

fail

### Versions of the cluster

tested using 6.1.3, ~~but is should fail on current master too~~, master branch change lightning tls config to `sslCABytes`, so no such problem

```console
```

liumengya94:

Issue Number: close #7941

don't dump if user didn't specify some `ssl-xxx`

#### Tests 

 - Unit test
 - Integration test

No
No
### Release note 

```release-note
None
```


D3Hunter (approved):

lance6716 (approved):

lance6716:
release-6.5 should also be fixed

D3Hunter:
> release-6.5 should also be fixed

after 6.3, we're passing SSLCaBytes to lightning, no such problem

PR #8010, opened 2023-01-03
scheduler(ticdc): optimize owner memory allocation

overvenus:

Issue Number: ref #7720 

* pkg/spanz(ticdc): increase btree degree to speed up Ascend
* *(ticdc): cache []tablepb.Span
* pkg/spanz(ticdc): do not let variables move to heap 
* scheduler(ticdc): do not let span move to heap

#### Tests 

 - Unit test
 - Integration test

 Replicate 72k tables. Left master branch, right this PR.
* `alloc_objects` are reduced from 46% to less than 1% 
* `cpu` are almost the same, from 3.44% to 3.19%
image
image

No.

No.

### Release note 

```release-note
None
```


pkg/spanz/map.go
```diff
@@ -109,52 +117,65 @@ func (m *Map[T]) AscendRange(start, end tablepb.Span, iterator ItemIterator[T])
...
 // FindHoles returns an array of Span that are not covered in the range
 // [start, end).
-// Note: Table ID is not set in returned holes.
-func (m *Map[T]) FindHoles(start, end tablepb.Span) (coveredSpans, holes []tablepb.Span) {
+// Note:
+// * Table ID is not set in returned holes.
+// * Returned slice is read only and will be changed on next FindHoles.
+func (m *Map[T]) FindHoles(start, end tablepb.Span) ([]tablepb.Span, []tablepb.Span) {
 	if bytes.Compare(start.StartKey, end.StartKey) >= 0 {
 		log.Panic("start muse be larger than end",
```
zhaoxinyu, line 125:
```suggestion
		log.Panic("start must be larger than end",
```

pkg/spanz/map.go
```diff
@@ -36,12 +37,19 @@ func lessSpanItem[T any](a, b spanItem[T]) bool {
...
+	cache *struct {
+		coveredSpans, holes []tablepb.Span
+		lastGC              time.Time
+	}
 }
 
 // NewMap returns a new SpanMap.
 func NewMap[T any]() *Map[T] {
+	// Map is read heavy map, many Ascend.
+	const defaultDegree = 256
```
zhaoxinyu, line 50: Maybe it's better to move this constant variable out of `NewMap()`

pkg/spanz/map.go
```diff
@@ -109,52 +117,65 @@ func (m *Map[T]) AscendRange(start, end tablepb.Span, iterator ItemIterator[T])
...
+			zap.String("end", end.String()))
+	}
+	if m.cache == nil || time.Since(m.cache.lastGC) > time.Minute {
+		m.cache = &struct {
+			coveredSpans []tablepb.Span
+			holes        []tablepb.Span
+			lastGC       time.Time
+		}{lastGC: time.Now()}
 	}
+	m.cache.coveredSpans = m.cache.coveredSpans[:0]
```
3AceShowHand, line 136: How about add a new method named `CleanCache` to replace this 2 lines.

3AceShowHand (approved):

pkg/spanz/map.go
overvenus: I file a task in https://github.com/pingcap/tiflow/issues/7720, the btree map will only be used in scheduler package which has to deal with a larger number of tables and lots of Ascend. I will address your comment in the task.

zhaoxinyu (approved):

PR #8055, opened 2023-01-11
sinkv2(cdc): add busy ratio metric for conflict detector

hicqu:

Issue Number: ref #5928 

* Add a metrics for conflict detector background goroutine busy ratio
* Use slice instead of BTree for depender list, because all events are already sorted

#### Tests 

 - Unit test
 - Manual bench
    `With the patch     : BenchmarkHighConflictsWithManyKeys-16               1230            855992 ns/op`
    `Without the patch: BenchmarkHighConflictsWithManyKeys-16               1501            961344 ns/op` 

### Release note 

```release-note
None
```


hicqu:
The PR seems not necessary. Test result shows the conflict detector background thread is always free enough.

PR #8070, opened 2023-01-12
scheduler(ticdc): split more evenly

overvenus:

Issue Number: ref #7720 

* ~Enable split span by default.~
* Split span evenly.

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
None
```


nongfushanquan (commented):
I recommend not enabling split span by default .the user can enable this feature in the  changefeed configure  if necessary

3AceShowHand:
> I recommend not enabling split span by default .the user can enable this feature in the changefeed configure if necessary

I agreed. In the most use case, there are multiple tables in the changefeed so that split span is not necessary.

We can update the documents to tell users this feature, and let them to decide whether to enable it or not.

3AceShowHand (approved):

overvenus:
> I recommend not enabling split span by default .the user can enable this feature in the changefeed configure if necessary

How about setting a larger default value? E.g., 100K regions. It's will disable the feature for most cases (smaller than 100k regions).

asddongmen (approved):

PR #8147, opened 2023-01-31
*(all): remove the deprecated build tag and use the correct context

0xPoe:

Issue Number: ref https://github.com/pingcap/tiflow/issues/4954

remove the deprecated build tag and use the correct context. Please try `go fix ./...`.
See more at https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives
#### Tests 

 - Unit test
 - Integration test

No
No
### Release note 

```release-note
None
```

### DIFF
--- a/tests/mq_protocol_tests/framework/docker_compose_op.go
+++ b/tests/mq_protocol_tests/framework/docker_compose_op.go
@@ -14,12 +14,12 @@
 package framework
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"os"
 	"os/exec"
 
-	"context"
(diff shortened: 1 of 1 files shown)


lance6716 (approved):

tests/mq_protocol_tests/framework/docker_compose_op.go
```diff
@@ -19,13 +19,13 @@ import (
 	"os"
 	"os/exec"
 
+	"context"
```
gozssky, line 22: Imports are not sorted.

gozssky (approved):

PR #8190, opened 2023-02-07
tracker(dm): remove NO_AUTO_CREATE_USER from sql_mode used by dm

### LINKED ISSUE #8149: DM uses a SQL mode which is removed in MySQL 8.0
### What did you do?

https://asktug.com/t/topic/1000577/6

```
[2023/01/31 11:40:41.469 +08:00] [ERROR] [baseconn.go:184] [“execute statement failed”] [task=alydm] [unit=“binlog replication”] [query=“SET SESSION SQL_MODE = ‘ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION’”] [argument=”[]”] [error=“Error 1231: Variable ‘sql_mode’ can’t be set to the value of ‘NO_AUTO_CREATE_USER’”]
```

https://github.com/pingcap/tiflow/blob/b633cef010a52a152c3ab6d2bd27cb1da563d365/dm/pkg/schema/tracker.go#L477-L478

Although we don't support MySQL 8.0 as downstream, this is easy to fix so maybe we should consider it.

_No response_

### What did you see instead?

_No response_

### Versions of the cluster

```console
```
```console
```
```console
```
```console
(leave TiUP or manually here)
```
```console
>
>
```

```console
```

liumengya94:

Issue Number: close #8149

`NO_AUTO_CREATE_USER` is removed in MySQL8.0([doc](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-11.html#mysqld-8-0-11-deprecation-removal))，so it should not be used in dm.

#### Tests 

 - Unit test
 - Integration test

No
No
### Release note 

```release-note
None
```

### DIFF
--- a/dm/pkg/schema/tracker.go
+++ b/dm/pkg/schema/tracker.go
@@ -89,7 +89,7 @@ type executorContext struct {
 
 var _ sqlexec.RestrictedSQLExecutor = executorContext{}
 
-const defaultSQLMode = "ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION"
+const DefaultSQLMode = "ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION"
 
 func (se executorContext) ParseWithParams(context.Context, string, ...interface{}) (ast.StmtNode, error) {
 	return nil, nil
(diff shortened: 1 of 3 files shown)


dm/pkg/schema/tracker.go
```diff
@@ -89,6 +89,8 @@ type executorContext struct {
 
 var _ sqlexec.RestrictedSQLExecutor = executorContext{}
 
+const DefaultSQLMode = "ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION"
```
lance6716, line 92:
```suggestion
const DefaultSQLModeMySQL80 = "ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION"
```

lance6716 (commented):
rest lgtm

dm/pkg/schema/tracker.go
D3Hunter: why add suffix 80, seems we didn't check mysql version and used it directly.

lance6716: in case other developers "correct" it to MySQL v5

PR #8249, opened 2023-02-14
tests(ticdc): add some integration tests to cover kafka sink v2

### LINKED ISSUE #8250: enable kafka sink v2 for integration test

3AceShowHand:

Issue Number: close #8250 

* migrate some kafka integration tests and enable kafka sink v2. The following cases are selected:
  * multi_topics
  * multi_tables_ddl
  * kafka_big_message
  * big_txn 
* fix some issue which cause cased failed

#### Tests 

 - Integration test

### Release note 

```release-note
`None`
```


pkg/sink/kafka/options.go
```diff
@@ -424,15 +432,15 @@ var (
 )
 
 // NewKafkaClientID generates kafka client id
-func NewKafkaClientID(role, captureAddr string,
+func NewKafkaClientID(captureAddr string,
```
zhaoxinyu, line 435: Why `role` is not needed anymore?

pkg/sink/kafka/v2/admin.go
```diff
@@ -233,7 +221,7 @@ func (a *admin) CreateTopic(
 	}
 
 	for _, err := range response.Errors {
-		if err != nil {
+		if err != nil && !errors.Is(err, kafka.TopicAlreadyExists) {
```
zhaoxinyu, line 224: We should have a manual test to check if `errors.Is()` is effective. In earlier release, we made a wrong assumption and we finally adopt the string matching method. 
Ref: https://github.com/pingcap/tiflow/blob/release-6.5/cdc/sink/mq/manager/kafka_manager.go#L283

zhaoxinyu: /cc @hi-rustin

tests/integration_tests/_utils/server.toml
```diff
@@ -0,0 +1,3 @@
+[debug]
+# todo: remove this after v2 is enabled as the default
+enable-kafka-sink-v2 = true
```
sdojjy, line 3: add some test that `enable-kafka-sink-v2=false` to test the sarama logic

zhaoxinyu (approved):

sdojjy (approved):

tests/integration_tests/_utils/server.toml
hi-rustin: Will this break other tests? Because some tests use this arg.

Also, why don't we just enable it in the code?  The release6.6 has been created.

hi-rustin: > Will this break other tests? Because some tests use this arg.

I mean `./cdc server --config xxx`

pkg/sink/kafka/sarama_factory.go
```diff
@@ -14,44 +14,44 @@
...
-		config:          saramaConfig,
+		changefeedID: changefeedID,
+		option:       o,
+		config:       saramaConfig,
 	}, nil
 }
 
 func (f *saramaFactory) AdminClient() (ClusterAdminClient, error) {
-	return newAdminClient(f.brokerEndpoints, f.config, f.changefeedID)
+	f.config.Producer.MaxMessageBytes = f.option.MaxMessageBytes
```
hi-rustin, line 46: Could you please add some tests for this code?

hi-rustin: Can we just apply it when we new the factory?

3AceShowHand: after the factory created, it's option would be altered by `AdjustOptions` later.

hi-rustin: It's difficult to maintain it. If we change any options in `AdjustOptions `, then we need to update it there times. How about we create the config from options when we using it? Or pass the `f.config` into AdjustOptions?

hi-rustin (approved):
Maybe we need to find a better way to set the client ID. Instead of just passing the ctx to the options. 

Let's move forward for now.

zhaoxinyu (approved):

PR #8268, opened 2023-02-15
cdc: add changefeed epoch to prevent unexpected state

### LINKED ISSUE #7657: Delete and create the same changefeed in a short time may cause deadlock and data loss
### What did you do?

Because of `model.ChangeFeedID` is non-unique, owner and process has no way to
find out whether they are running the same changefeed. After appling delete and
create operation on the same changefeed name, there is a race condition between
owner and processor that they are running two different changefeeds. This may
cause deadlock and data loss.

To fix the issue, we should make sure:
1. When owner receives a delete changefeed request, it can only return a
response after the changefeed is closed completely on every TiCDC node. It
prevents data loss.
2. Add a unique field to `model.ChangeFeed`. It prevents deadlock and other 
illegal states.

No deadlock and no data loss.

### What did you see instead?

Changefeed is a deadlock state, and lag is increasing.

### Versions of the cluster

```console
All TiCDC released version
```

### LINKED ISSUE #8242: Changefeed lag keeps increasing after a changefeed restart
### What did you do?

1. Create a changefeed with storage sink.
2. The downstream storage system (e.g. minio) experiences a temporary unavailability.

Changefeed is restarted and the tables are replicated normally.

### What did you see instead?

Changefeed is restarted, but some of the tables are not being replicated normally. And a large amount of logs are printed as follows:
```
[2023/02/12 01:39:38.441 +08:00] [WARN] [replication_set.go:741] ["schedulerv3: ignore input, unexpected replication set state"] [tableState="table_id:36489 span:<{table_id:36489,start_key:748000000000008eff895f720000000000fa,end_key:748000000000008eff895f730000000000fa}> state:Absent checkpoint:<> stats:<> "] [captureID=ee34951b-2232-4afc-a557-73349adea798] [replicationSet="{\"Changefeed\":{\"Namespace\":\"default\",\"ID\":\"longrun-s3-test-02\"},\"Span\":\"{table_id:36489,start_key:748000000000008eff895f720000000000fa,end_key:748000000000008eff895f730000000000fa}\",\"State\":\"Replicating\",\"Primary\":\"ee34951b-2232-4afc-a557-73349adea798\",\"Captures\":{\"ee34951b-2232-4afc-a557-73349adea798\":\"Primary\"},\"Checkpoint\":{\"checkpoint_ts\":439387383142023858,\"resolved_ts\":439387383194452934},\"Stats\":{\"stage_checkpoints\":null}}"]
```

### Versions of the cluster

```console
v6.6.0
```

overvenus:

Issue Number: close #7657
Also close #8242

Add changefeed epoch, so we can distinguish messages after changefeed restart.

#### Tests 

 - Unit test

No

No

### Release note 

```release-note
Fix an issue that may cause data lost or checkpoint stuck when changefeed restarting.
```


cdc/model/changefeed.go
```diff
@@ -146,6 +146,8 @@ type ChangeFeedInfo struct {
 	Error  *RunningError         `json:"error"`
 
 	CreatorVersion string `json:"creator-version"`
+	// Epoch is the epoch of a changefeed, changes on every restart.
+	Epoch uint64 `json:""`
```
overvenus, line 150: We should add a valid JSOn tag.

cdc/model/changefeed.go
```diff
@@ -146,6 +146,8 @@ type ChangeFeedInfo struct {
 	Error  *RunningError         `json:"error"`
 
 	CreatorVersion string `json:"creator-version"`
+	// Epoch is the epoch of a changefeed, changes on every restart.
```
3AceShowHand, line 149: how about refine this comment, what an `epoch` stand for ?

cdc/owner/changefeed.go
```diff
@@ -583,6 +584,7 @@ LOOP:
 	c.initialized = true
 	log.Info("changefeed initialized",
 		zap.String("namespace", c.state.ID.Namespace),
+		zap.Uint64("changefeedEpoch", epoch),
```
3AceShowHand, line 587: reorder the fields, changefeed first, and rename this to epoch should be ok.

cdc/owner/feed_state_manager.go
```diff
@@ -370,6 +359,18 @@ func (m *feedStateManager) patchState(feedState model.FeedState) {
 			info.AdminJobType = adminJobType
 			changed = true
 		}
+		if updateEpoch {
+			previous := info.Epoch
+			ctx, cancel := context.WithTimeout(context.TODO(), 5*time.Second)
```
3AceShowHand, line 364: Is `context.TODO()` ok here ?

3AceShowHand (approved):

cdc/owner/feed_state_manager.go
overvenus: Tried, it will change a lot of unrelated code, I'll add a TODO here.

cdc/model/changefeed.go
zhaoxinyu: Is `epoch` similar to `term`?

cdc/owner/feed_state_manager.go
```diff
@@ -370,6 +359,18 @@ func (m *feedStateManager) patchState(feedState model.FeedState) {
...
+			defer cancel()
+			info.Epoch = GenerateChangefeedEpoch(ctx, m.upstream.PDClient)
+			changed = true
+			log.Info("update changefeed epoch",
+				zap.String("namespace", m.state.ID.Namespace),
+				zap.String("changefeed", m.state.ID.ID),
+				zap.Uint64("perviousEpoch", previous),
+				zap.Uint64("changefeedEpoch", info.Epoch))
```
zhaoxinyu, line 372:
```suggestion
				zap.Uint64("currentEpoch", info.Epoch))
```

cdc/owner/feed_state_manager.go
```diff
@@ -498,3 +499,13 @@ func (m *feedStateManager) handleError(errs ...*model.RunningError) {
 			zap.Duration("newInterval", m.backoffInterval))
 	}
 }
+
+// GenerateChangefeedEpoch generates a unique changefeed epoch.
+func GenerateChangefeedEpoch(ctx context.Context, pdClient pd.Client) uint64 {
+	phyTs, logical, err := pdClient.GetTS(ctx)
+	if err != nil {
+		log.Warn("generate epoch using local timestamp due error", zap.Error(err))
```
zhaoxinyu, line 507:
```suggestion
		log.Warn("generate epoch using local timestamp due to error", zap.Error(err))
```

zhaoxinyu (approved):

Issue #8106, opened 2023-01-18
Optimize TiCDC handling of DDL event and reduce the amount of memory consumed by table schemas.


asddongmen:
### Is your feature request related to a problem?

Currently, TiCDC has two main problems when dealing with DDL events:
1. The DDL event of any table will block the progress of all tables in the changefeed, resulting in increased checkpoint lag.
2. DDL events are processed serially in TiCDC, while TiDB supports concurrent execution of ddl. If the upstream executes a large number of ddls concurrently in a short period of time, checkpoint lag of a changefeed is likely to increase due to slow processing of DDL event.

In terms of table schema storage, TiCDC mainly has the following problems:
1. Each changefeed stores the meta-information of all tables in the upstream TiDB cluster. If there are a large number of tables in the upstream cluster, this will cause a large amount of waste.

### Describe the feature you'd like

1. Make ddl event block only the progression of related tables. #8343 #8535 #8265 
2. Make TiCDC support concurrent DDL processing.
3. Make `ddlPuller` share one `schemaStorage` with `changefeed` or `processor`. #8633 
4. Delete unrelate schem infos of a chengfeed in `schemaStorage`.https://github.com/pingcap/tiflow/pull/8697

asddongmen:
All PRs has been merged.

PR #8343, opened 2023-02-22
changefeed (ticdc): add ddl manager

asddongmen:

Issue Number: ref #8106 

Only add a new struct and some new methods, no any logic changes. 
Test will be added in next PR.

### Release note 

```release-note
None
```


cdc/model/changefeed.go
```diff
@@ -453,3 +453,40 @@ func (info *ChangeFeedInfo) fixMemoryQuota() {
 func (info *ChangeFeedInfo) fixScheduler() {
 	info.Config.FixScheduler()
 }
+
+type Barrier struct {
+	GlobalBarrierTs   Ts
```
overvenus, line 458: Do they need json tags?

cdc/model/changefeed.go
```diff
@@ -453,3 +453,40 @@ func (info *ChangeFeedInfo) fixMemoryQuota() {
...
+	}
+}
+
+type DownStreamType int
+
+const (
+	// DB is the type of sink for database.
+	DB DownStreamType = iota
+	// Other is the type of kafka or storage.
+	Other
```
overvenus, line 480: What about `MQ`?

cdc/model/changefeed.go
```diff
@@ -453,3 +453,40 @@ func (info *ChangeFeedInfo) fixMemoryQuota() {
...
+		GlobalBarrierTs: ts,
+	}
+}
+
+type DownStreamType int
+
+const (
+	// DB is the type of sink for database.
+	DB DownStreamType = iota
+	// Other is the type of kafka or storage.
```
amyangfei, line 479: s/storage/cloud storage

cdc/owner/ddl_manager.go
```diff
@@ -0,0 +1,480 @@
...
+	"github.com/pingcap/tiflow/cdc/model"
+	"github.com/pingcap/tiflow/cdc/puller"
+	"github.com/pingcap/tiflow/cdc/redo"
+	"go.uber.org/zap"
+)
+
+// globalDDLs is the DDLs that affect all tables in the changefeed.
+// we need to wait all tables checkpointTs reach the DDL commitTs
+// before we can execute the DDL.
+var globalDDLs = []timodel.ActionType{
```
amyangfei, line 30: Should `truncate table` be in `globalDDLs`?

cdc/owner/ddl_manager.go
```diff
@@ -0,0 +1,480 @@
...
+	m.justSentDDL = nil
+	if m.executingDDL == nil {
+		return nil
+	}
+	done, err := m.ddlSink.emitDDLEvent(ctx, m.executingDDL)
+	if err != nil {
+		return err
+	}
+	if done {
+		m.pendingDDLs[m.executingDDL.TableInfo.TableName] = m.pendingDDLs[m.executingDDL.TableInfo.TableName][1:]
```
amyangfei, line 268: Can set `m.pendingsDDLs[0]` to nil before this line to accelerate GC.

cdc/owner/ddl_manager.go
```diff
@@ -0,0 +1,480 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package owner
+
+import (
+	"context"
```
amyangfei, line 17: Can separate the import into two groups.

cdc/model/changefeed.go
asddongmen: Yes,thanks!

cdc/model/changefeed.go
asddongmen: MQ and cloud storage both belong to `Other` type.

cdc/owner/ddl_manager.go
asddongmen: I think that when handling a `truncate table` ddl, you only need to stop its `preTable` progress at ddl commiTs before executing it. So there is no need to put it in `globalDDLs`. 
I will do more testing with this ddl.

asddongmen: Based on my understanding, the `truncate table` DDL has two table information sets: `preTableInfo` and `tableInfo`. The `ddl_manager` uses the `preTableInfo` to prevent the checkpointTs from progressing, ensuring that the DDL is executed without any disruption. After the DDL is executed, the scheduler will use the `tableID` contained in the `tableInfo`, along with the `commiTs` as `startTs`, to create a new table task. 

So I think that the `truncate table` DDL does not need to be included in `globalDDLs`. However, I will conduct more tests on this DDL to ensure that our understanding is correct. Thank you for reminding me of this important consideration.

cdc/model/changefeed.go
```diff
@@ -453,3 +453,40 @@ func (info *ChangeFeedInfo) fixMemoryQuota() {
...
+	}
+}
+
+type DownStreamType int
+
+const (
+	// DB is the type of Database.
+	DB DownStreamType = iota
+	// Other is the type of MQ or Cloud Storage.
+	Other
```
CharlesCheung96, line 480: Better to reuse the definitions in [sink pkg](https://github.com/pingcap/tiflow/blob/master/pkg/sink/sink_type.go#L74).

asddongmen: This is a bit complicated. I prefer keeping things simple, so I don't want to pass sink-uri into newDDLManager().

cdc/model/changefeed.go
overvenus: Prefer explicit, what about adding a new `Storage` type?

amyangfei (approved):

cdc/owner/ddl_manager.go
```diff
@@ -0,0 +1,483 @@
...
+		// minTableBarrierTs is the min commitTs of all tables DDLs,
+		// it is used to prevent the checkpointTs from advancing too fast.
+		if ddl.CommitTs < minTableBarrierTs {
+			minTableBarrierTs = ddl.CommitTs
+		}
+	}
+
+	m.justSentDDL = nil
+	return model.Barrier{
+		GlobalBarrierTs:   globalBarrierTs,
```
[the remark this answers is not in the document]
asddongmen: `globalBarrierTs` >= `minTableBarrierTs`.
When a DDL is being executed, the `minTableBarrierTs = ddlCommitTs = changefeed checkpointTs`. This field is used to determine whether a DDL was being executed when a changefeed is resumed. If so, we should start the `ddlPuller` from `checkpointTs-1`.

cdc/model/changefeed.go
asddongmen: done

PR #8362, opened 2023-02-23
redo(ticdc): support for applying ddl event in applier

### LINKED ISSUE #8361: Make redo applier support consuming DDL events

CharlesCheung96:

Issue Number: close #8361

1. Support for applying ddl event in applier
2. Read and decode log events in background goroutines in logReader. 
3. [Add TableName field to RedoDDLEvent](https://github.com/pingcap/tiflow/pull/8362/files#diff-6d2f9b5931ca742dd70204701ec3baa8cce6350a61e29d55a0e85176b7618fcaR112).

#### Tests 

 - Unit test
 - Integration test
 
### Release note 

```release-note
`support for applying ddl event in redo log`.
```

### DIFF
--- a/cdc/redo/reader/reader.go
+++ b/cdc/redo/reader/reader.go
@@ -22,30 +22,30 @@ import (
 	"path/filepath"
 	"sync"
 
-	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	"github.com/pingcap/tiflow/cdc/model"
 	"github.com/pingcap/tiflow/cdc/redo/common"
-	cerror "github.com/pingcap/tiflow/pkg/errors"
+	"github.com/pingcap/tiflow/pkg/errors"
 	"github.com/pingcap/tiflow/pkg/redo"
+	"github.com/pingcap/tiflow/pkg/sink/mysql"
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	emitBatch             = mysql.DefaultMaxTxnRow
+	defaultReaderChanSize = mysql.DefaultWorkerCount * emitBatch
 )
 
 // RedoLogReader is a reader abstraction for redo log storage layer
 type RedoLogReader interface {
-	io.Closer
-
-	// ResetReader setup the reader boundary
-	ResetReader(ctx context.Context, startTs, endTs uint64) error
-
-	// ReadNextLog reads up to `maxNumberOfMessages` messages from current cursor.
-	// The returned redo logs sorted by commit-ts
-	ReadNextLog(ctx context.Context, maxNumberOfEvents uint64) ([]*model.RowChangedEvent, error)
-
-	// ReadNextDDL reads `maxNumberOfDDLs` ddl events from redo logs from current cursor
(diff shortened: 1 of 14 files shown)


cdc/redo/reader/reader.go
```diff
@@ -145,203 +141,140 @@ func (l *LogReader) ResetReader(ctx context.Context, startTs, endTs uint64) erro
...
-	if err != nil {
-		return err
-	}
-
+func (l *LogReader) runDDLReader(egCtx context.Context) error {
+	defer close(l.ddlCh)
 	ddlCfg := &readerConfig{
-		startTs:            startTs,
-		endTs:              endTs,
+		startTs:            l.meta.CheckpointTs - 1,
```
nongfushanquan, line 171: why is -1 needed?

CharlesCheung96: If there is a ddl event in the check point, there is no way to determine whether the event was executed. Therefore, it is a safe practice to always assume that the event did not execute successfully and try to execute this DDL event in applier, which is similar to initializing DDLPuller with `checkpointTs - 1` when owner restarts.

pkg/applier/main_test
```diff
@@ -0,0 +1,24 @@
+// Copyright 2023 PingCAP, Inc.
```
amyangfei, line 1: Filename should be `main_test.go`?

pkg/applier/redo.go
```diff
@@ -103,101 +117,207 @@ func (ra *RedoApplier) catchError(ctx context.Context) error {
 	}
 }
 
+func (ra *RedoApplier) initSink(ctx context.Context) (err error) {
+	// MySQL sink will use the following replication config
+	// - EnableOldValue: default true
+	// - ForceReplicate: default false
+	// - filter: default []string{"*.*"}
```
amyangfei, line 124: Seems these comments are not necessary and can remove them.

cdc/redo/reader/reader.go
```diff
@@ -22,30 +22,30 @@ import (
...
-	// The returned redo logs sorted by commit-ts
-	ReadNextLog(ctx context.Context, maxNumberOfEvents uint64) ([]*model.RowChangedEvent, error)
-
-	// ReadNextDDL reads `maxNumberOfDDLs` ddl events from redo logs from current cursor
-	ReadNextDDL(ctx context.Context, maxNumberOfEvents uint64) ([]*model.DDLEvent, error)
-
+	// Run read and decode redo logs in background.
+	Run(ctx context.Context) error
+	// ReadNextRow read one row event from redo logs.
+	ReadNextRow(ctx context.Context) (*model.RowChangedEvent, error)
```
amyangfei, line 46: Before this pr, what is the batch size that `ReadNextLog` return in most cases? Is there any performance issue that changing it from batch to one row?

CharlesCheung96: Currently we use [buffered channel](https://github.com/pingcap/tiflow/pull/8362/files#diff-7194f7523941d867fef7ed626c77db0337996b8bc35063b9fa595099215f51beR108) to cache logs with [the same capacity](https://github.com/pingcap/tiflow/pull/8362/files#diff-7194f7523941d867fef7ed626c77db0337996b8bc35063b9fa595099215f51beR38) as the previous batchsize.

amyangfei (approved):

Issue #8056, opened 2023-01-11
Tracking issue for optimizing redo log


CharlesCheung96:
- [x] In owner, separate DDL dataflow to redoManager and DDLsink. And calculate resolvedTs of redo module based on DDL and DML progress.
- [x] #8361
- [x] Refactor redo applier based on sinkV2 to improve downstream write concurrency and reduce RTO. 
- [x] [Support use memory to cache temporary files in redo log](https://github.com/pingcap/tiflow/issues/8086).
- [x] https://github.com/pingcap/tiflow/issues/8361
- [x] https://github.com/pingcap/tiflow/pull/8407
- [x] #8963
- [x] In sinkManager, ensure that redo log is not blocked by the sink related error. 
- [x] #9313
- [ ] Refactor redo log module based on the cloud storage sink.

nongfushanquan:
Support use memory to cache temporary files in redo log. --- this feature has been done in the master and 6.5

PR #8407, opened 2023-03-01
redo(ticdc): simplify reader initialization

CharlesCheung96:

Issue Number: ref #8056

1. After downloading a file, sort the file before writing it to the local disk to [avoid multiple disk access](https://github.com/pingcap/tiflow/pull/8407/files#diff-702fc02e577e50bc8109e25c72a283bd7cb82f4f9ce76e34aac437be0bedcd59R116).

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/redo/reader/file.go
+++ b/cdc/redo/reader/file.go
@@ -243,18 +243,17 @@ func sortAndWriteFile(
 		return err
 	}
 
-	data, err := extStorage.ReadFile(egCtx, fileName)
+	fileContent, err := extStorage.ReadFile(egCtx, fileName)
 	if err != nil {
 		return cerror.WrapError(cerror.ErrExternalStorageAPI, err)
 	}
-	if len(data) == 0 {
+	if len(fileContent) == 0 {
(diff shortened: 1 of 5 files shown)


cdc/redo/reader/file.go
```diff
@@ -121,134 +113,99 @@ func newReader(ctx context.Context, cfg *readerConfig) ([]fileReader, error) {
...
 		select {
-		case <-ctx.Done():
-			return ctx.Err()
+		case <-eCtx.Done():
+			return nil, eCtx.Err()
 		case limit <- struct{}{}:
 		}
-		f := file
+
+		fileName := file
```
hicqu, line 144: It's better to comment why we need `fileName` instead of using `file` directly.

hicqu (approved):

cdc/redo/reader/file.go
CharlesCheung96: Since we need to use this variable in the closure:
image

asddongmen (approved):

Issue #8132, opened 2023-01-30
Remove table pipeline and sink v1


hi-rustin:
Because we already using pull-based-sink for a while. And it works well. So we can remove the old sink and table pipeline.

- [x] [remove table pipeline](https://github.com/pingcap/tiflow/pull/8133)
- [x] remove old sorter
- [x] remove sinkv1
- [ ] fix all unit tests

PR #8419, opened 2023-03-02
sink(ticdc): use a global variable to ensure that the batch ID is monotonically increasing

hi-rustin:

Issue Number: ref https://github.com/pingcap/tiflow/issues/8132

use a global variable to ensure that the batch ID is monotonically increasing. See more in the code comments.
#### Tests 

 - Unit test

No
No
### Release note 

```release-note
None
```


cdc/processor/sinkmanager/table_sink_worker.go
```diff
@@ -109,7 +131,9 @@ func validateAndAdjustBound(changefeedID model.ChangeFeedID,
 }
 
 func (w *sinkWorker) handleTask(ctx context.Context, task *sinkTask) (finalErr error) {
-	advancer := newTableSinkAdvancer(task, w.splitTxn, w.sinkMemQuota, requestMemSize)
+	// We need to use a new batch ID for each task.
+	batchID := batchID.Add(1)
+	advancer := newTableSinkAdvancer(task, w.splitTxn, w.sinkMemQuota, requestMemSize, batchID)
```
hicqu, line 136: Why not access global `batchID` in Advancer directly?

hi-rustin: It breaks a lot of code and is not easy to test and maintain.

hi-rustin: Also, it reduces a lot of atomic operations.

amyangfei (approved):

PR #8485, opened 2023-03-09
capture(ticdc): fix graceful shutdown in k8s

### LINKED ISSUE #8484: TiCDC can not graceful upgrade in k8s
#7400 breaks graceful shutdown in k8s by deleting `CaptureInfo` immediately after a node is drained.  
And #7519 makes TiCDC always returns 503 for every API call which makes it even worse.

overvenus:

Issue Number: close #8484

1. Allow access etcd client after a capture has been drained.
2. Do not return error in `campaignOwner` after a capture has been drained.

#### Tests 

 - Unit test
	Run TiCDC cluster in a k8s cluster and rolling restart all TiCDC nodes.

No.

No.

### Release note 

```release-note
Fix an issue that TiCDC cluster can not graceful upgrade in k8s
```

### DIFF
--- a/cdc/api/owner/owner.go
+++ b/cdc/api/owner/owner.go
@@ -227,18 +227,13 @@ func (h *ownerAPI) handleChangefeedQuery(w http.ResponseWriter, req *http.Reques
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
-	etcdClient, err := h.capture.GetEtcdClient()
-	if err != nil {
-		api.WriteError(w, http.StatusBadRequest, err)
-		return
-	}
-	cfInfo, err := etcdClient.GetChangeFeedInfo(ctx, changefeedID)
+	cfInfo, err := h.capture.GetEtcdClient().GetChangeFeedInfo(ctx, changefeedID)
 	if err != nil && cerror.ErrChangeFeedNotExists.NotEqual(err) {
 		api.WriteError(w, http.StatusBadRequest,
 			cerror.ErrAPIInvalidParam.GenWithStack("invalid changefeed id: %s", changefeedID))
 		return
 	}
-	cfStatus, _, err := etcdClient.GetChangeFeedStatus(ctx, changefeedID)
+	cfStatus, _, err := h.capture.GetEtcdClient().GetChangeFeedStatus(ctx, changefeedID)
 	if err != nil && cerror.ErrChangeFeedNotExists.NotEqual(err) {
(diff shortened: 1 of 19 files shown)


cdc/capture/capture.go
```diff
@@ -181,9 +178,6 @@ func (c *captureImpl) GetUpstreamManager() (*upstream.Manager, error) {
 func (c *captureImpl) GetEtcdClient() (etcd.CDCEtcdClient, error) {
 	c.captureMu.Lock()
 	defer c.captureMu.Unlock()
-	if !c.initialized {
-		return nil, cerror.ErrCaptureNotInitialized.GenWithStackByArgs()
-	}
 	return c.EtcdClient, nil
```
sdojjy, line 181: this func does not return error now , change the func definition

cdc/api/owner/owner.go
```diff
@@ -227,18 +227,13 @@ func (h *ownerAPI) handleChangefeedQuery(w http.ResponseWriter, req *http.Reques
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
-	etcdClient, err := h.capture.GetEtcdClient()
-	if err != nil {
-		api.WriteError(w, http.StatusBadRequest, err)
-		return
-	}
-	cfInfo, err := etcdClient.GetChangeFeedInfo(ctx, changefeedID)
+	cfInfo, err := h.capture.GetEtcdClient().GetChangeFeedInfo(ctx, changefeedID)
```
nongfushanquan, line 230: Is it not safe to cache etcdclient ?

overvenus: Yes, etcd client is initialized before creating `Capture`, and will not be closed before exiting.

cdc/capture/capture.go
```diff
@@ -326,15 +315,13 @@ func (c *captureImpl) run(stdCtx context.Context) error {
 		// an unrecoverable serious errors (recoverable errors are intercepted in the owner tick)
 		// so we should restart the capture.
 		err := c.campaignOwner(ctx)
-		if err != nil {
+		if err != nil || c.liveness.Load() != model.LivenessCaptureStopping {
 			log.Error("campaign owner routine exited with error, restart the capture",
```
sdojjy, line 319: the log  is not correct when `c.liveness.Load() != model.LivenessCaptureStopping `

asddongmen (approved):

3AceShowHand (approved):

PR #8494, opened 2023-03-09
redo(ticdc): limit memory usage in applier

CharlesCheung96:

Issue Number: ref #8056

1. Limit [total memory usage](https://github.com/pingcap/tiflow/pull/8494/files#diff-a4f2a9712d65b01443a6b810b343317221039c2558b867f82d3d2c3c1fe18e0dR32) in reader.
2. Add memory quota in redo applier.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`Fix the issue that redo apply may cause OOM`.
```

### DIFF
--- a/cdc/processor/memquota/mem_quota.go
+++ b/cdc/processor/memquota/mem_quota.go
@@ -68,9 +68,11 @@ func NewMemQuota(changefeedID model.ChangeFeedID, totalBytes uint64, comp string
 		changefeedID:     changefeedID,
 		totalBytes:       totalBytes,
 		blockAcquireCond: sync.NewCond(&sync.Mutex{}),
-		metricTotal:      MemoryQuota.WithLabelValues(changefeedID.Namespace, changefeedID.ID, "total", comp),
-		metricUsed:       MemoryQuota.WithLabelValues(changefeedID.Namespace, changefeedID.ID, "used", comp),
-		closeBg:          make(chan struct{}, 1),
+		metricTotal: MemoryQuota.WithLabelValues(changefeedID.Namespace,
+			changefeedID.ID, "total", comp),
+		metricUsed: MemoryQuota.WithLabelValues(changefeedID.Namespace,
+			changefeedID.ID, "used", comp),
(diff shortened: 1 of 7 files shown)


cdc/processor/memquota/mem_quota.go
```diff
@@ -51,30 +52,28 @@ type MemQuota struct {
 
 	// blockAcquireCond is used to notify the blocked acquire.
 	blockAcquireCond *sync.Cond
-	// condMu protects nothing, but sync.Cond needs a mutex.
-	condMu sync.Mutex
```
hicqu, line 55: `blockAcquireCond` shouldn't share one mutex with `tableMemory`. It will increase lock conflicts.

pkg/applier/redo.go
```diff
@@ -201,6 +223,42 @@ func (ra *RedoApplier) consumeLogs(ctx context.Context) error {
...
+func (ra *RedoApplier) resetQuota(rowSize uint64) error {
+	if rowSize >= config.DefaultChangefeedMemoryQuota || rowSize < ra.pendingQuota {
+		log.Panic("row size exceeds memory quota",
+			zap.Uint64("rowSize", rowSize),
+			zap.Uint64("memoryQuota", config.DefaultChangefeedMemoryQuota))
+	}
+
+	// flush all tables before acquire new quota
+	for tableID, tableRecord := range ra.tableResolvedTsMap {
+		if !tableRecord.ResolvedTs.IsBatchMode() {
```
hicqu, line 235: If all records are not in batch mode, Why we need `MemConsumedRecord`?

cdc/processor/memquota/mem_quota.go
CharlesCheung96: `blockAcquireCond` maintains a separate lock internally that can be accessed through `blockAcquireCond.L`.

pkg/applier/redo.go
CharlesCheung96: If batch mode is not used, the applier will panic.

hicqu (approved):

PR #8501, opened 2023-03-10
dm: support schema isolation for lightning checkpoint when writing to downstream DB

### LINKED ISSUE #8500: dm: dm need support schema isolation when writing lightning checkpoint to downstream DB
### What did you do?

Using downstream DB to save lightning checkpoint for Cloud DM

Different DM job has different lightning checkpoint schema

### What did you see instead?

678792a6-6b16-4b30-a567-e277f01f165f

All dm jobs share the same task id and checkpoint related tables

### Versions of the cluster

```console
[ea92bc1](https://github.com/pingcap/tiflow/commit/ea92bc1833e1a57df9b511db7c5e49b4bfb1fb75)
```
```console
```
```console
```
```console
(leave TiUP or manually here)
```
```console
>
>
```

```console
```

maxshuang:

Issue Number: close #8500 

1. Using JobID+soureID as prefix to isolate the schema for lightning checkpoint when writing to downstream DB
2. Set 'Checkpoint.KeepAfterSuccess' to 'CheckpointRemove' to remove the lightning checkpoint file or schema after import.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
support schema isolation for lightning checkpoint when writing to downstream DB for cloud dm.
```


dm/loader/lightning.go
```diff
@@ -320,13 +321,13 @@ func GetLightningConfig(globalCfg *lcfg.GlobalConfig, subtaskCfg *config.SubTask
...
-		// NOTE: if we use bucket as dumper storage, write lightning checkpoint to downstream DB to avoid bucket ratelimit
-		// it's the default path for cloud dm.
+		// NOTE: If we use bucket as dumper storage, write lightning checkpoint to downstream DB to avoid bucket ratelimit
 		// since we will use check Checkpoint in 'ignoreCheckpointError', MAKE SURE we have assigned the Checkpoint config properly here
 		if err := cfg.Security.BuildTLSConfig(); err != nil {
 			return nil, err
 		}
-		cfg.Checkpoint.Schema = "tidb_lightning_checkpoint"
+		// To enable the loader worker failover, we need to use jobID+sourceID to isolate the checkpoint schema
+		cfg.Checkpoint.Schema = cputil.LightningCheckpointSchema(workerName, subtaskCfg.SourceID)
```
lance6716, line 330: Should be task name+source id.
Worker is dynamically bound to source. If the dump files are stored in s3, another worker can continue the task from previous worker's checkpoint

maxshuang: workerName is assigned to the  `JobID` outside as the comment says.

dm/loader/lightning.go
```diff
@@ -340,7 +341,7 @@ func GetLightningConfig(globalCfg *lcfg.GlobalConfig, subtaskCfg *config.SubTask
 		}
 		cfg.Checkpoint.DSN = cpPath
 	}
-	cfg.Checkpoint.KeepAfterSuccess = lcfg.CheckpointOrigin
+	cfg.Checkpoint.KeepAfterSuccess = lcfg.CheckpointRemove
```
maxshuang, line 344: PTAL~ 
Pls help review whether it's ok to remove the lightning checkpoint after the import.
@GMHDBJD @lance6716

dm/loader/lightning.go
```diff
@@ -308,7 +309,7 @@ func GetTaskInfoSchemaName(dmMetaSchema, taskName string) string {
 }
 
 // GetLightningConfig returns the lightning task config for the lightning global config and DM subtask config.
-func GetLightningConfig(globalCfg *lcfg.GlobalConfig, subtaskCfg *config.SubTaskConfig) (*lcfg.Config, error) {
+func GetLightningConfig(globalCfg *lcfg.GlobalConfig, subtaskCfg *config.SubTaskConfig, workerName string) (*lcfg.Config, error) {
```
lance6716, line 312: task name can be get from subtaskCfg.Name, so we don't need to change the signature

dm/loader/lightning.go
lance6716: The problem is if the worker process crashes between line 430 and line 432, after failover data will be imported twice. The candidate actions can be
- keep CheckpointOrigin, but user may complain that a strange database is left in thier cluster
- change to CheckpointRemove and ignore the risk
- add a callback function to lightning, update the DM load unit checkpoint before removing lightning's checkpoint
- refactor lightning checkpoint schema, put all tables under DM task meta database.

I'm neutral about them

dm/loader/lightning.go
lance6716: not sure if lightning escape the database name.

dm/pkg/cputil/table.go
```diff
@@ -23,6 +31,16 @@ func LightningCheckpoint(task string) string {
 	return task + "_lightning_checkpoint_list"
 }
 
+// LightningCheckpointSchema returns lightning's checkpoint schema name.
+func LightningCheckpointSchema(task string, sourceID string) string {
+	if task == "" {
+		return "tidb_lightning_checkpoint"
+	}
+	return utils.TruncateStringQuiet(
+		fmt.Sprintf("%s_%d_tidb_lightning_checkpoint", task, utils.GenHashKey(sourceID)),
```
lance6716, line 40: it's OK to left sourceID in database name? why task is not hashed but sourceID is hashed

engine/test/integration_tests/dm_lightning_checkpoint/run.sh
```diff
@@ -0,0 +1,42 @@
...
+function run() {
+	start_engine_cluster $CONFIG
+	wait_mysql_online.sh --port 3306
+	wait_mysql_online.sh --port 3307
+	wait_mysql_online.sh --port 4000
+
+	# prepare MySQL global variables and data
+	run_sql_file $CUR_DIR/data/db1.prepare.sql
+
+	# create job & wait for job finished
```
lance6716, line 21: if there a test that two lightning are running simultaneously?

dm/loader/lightning.go
maxshuang: I will check if the `subtaskCfg.Name`  can be unchanged during the whole job process.

dm/pkg/cputil/table.go
maxshuang: -- not sure if lightning escape the database name.
image

-- it's OK to left sourceID in database name? why task is not hashed but sourceID is hashed
For cloud DM, 'task' alway be the number like '12123', so the previous table name use the 'task' directly. 
'sourceID' may be the user-defined identifier which can contains some illegal chars for 'schema' and 'table'.

engine/test/integration_tests/dm_lightning_checkpoint/run.sh
maxshuang: I will add two jobs in this test.

dm/loader/lightning.go
maxshuang: `subtaskCfg.Name` is the `jobID`, I will fix it.

dm/pkg/cputil/table.go
maxshuang: After test, current implement doesn't escape the schema name.

engine/test/integration_tests/dm_lightning_checkpoint/run.sh
maxshuang: Fix it.

dm/loader/lightning.go
maxshuang: Already add the 'todo' and wait for another pr to fix it.
May be 'way 3' is the easier way.

dm/pkg/cputil/table.go
maxshuang: Currently, 'sourceID' is like  '{\"SourceID\":\"source-990001\"'. May be we can directly escape the name.

lance6716: Will task contain special character? we might also need to hash it

lance6716 (commented):
rest lgtm

lance6716 (approved):

dm/pkg/cputil/table.go
D3Hunter: there's no task param in engine `create-job` api, it's generated by engine

PR #8502, opened 2023-03-13
mounter(ticdc): mount float32 value correctly to avoid the precision lost.

3AceShowHand:

Issue Number: close #8490 

* For float32 types, fetch it from the`datum` by use the corresponding method.

#### Tests 

 - Unit test

### Release note 

```release-note
Fix float32 types precision lost
```

### DIFF
--- a/cdc/entry/mounter_test.go
+++ b/cdc/entry/mounter_test.go
@@ -16,7 +16,7 @@ package entry
 import (
 	"bytes"
 	"context"
-	"math"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


cdc/entry/mounter_test.go
```diff
@@ -1267,3 +1267,18 @@ func TestNewDMRowChange(t *testing.T) {
 		require.Equal(t, []interface{}{1, 2, 1, 2}, argsGot)
 	}
 }
+
+func TestFormatColVal(t *testing.T) {
+	t.Parallel()
+
+	var datum types.Datum
+	datum.SetFloat32(123.99)
```
amyangfei, line 1275: Can also add a test for `+Inf`, `-Inf` and `Nan`

sdojjy (approved):

overvenus (approved):

Issue #8189, opened 2023-02-07
Tracking issue for improving storage consumer capabilites


zhaoxinyu:
This issue tracks several enhancements for storage consumer which is very important for test cases in integration tests and test-infra.

- [x] https://github.com/pingcap/tiflow/issues/7720

PR #8530, opened 2023-03-15
sink(ticdc): reuse index number in storage sink

CharlesCheung96:

Issue Number: ref #8189

1. Reuse index number in storage sink if the last file does not exists.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```

### DIFF
--- a/pkg/sink/cloudstorage/path.go
+++ b/pkg/sink/cloudstorage/path.go
@@ -152,32 +152,12 @@ func (f *FilePathGenerator) GenerateDataFilePath(
 	date string,
 ) (string, error) {
 	var elems []string
-
 	elems = append(elems, f.generateDataDirPath(tbl, date))
 	if idx, ok := f.fileIndex[tbl]; !ok {
-		var fileIdx uint64
-
-		indexFile := f.GenerateIndexFilePath(tbl, date)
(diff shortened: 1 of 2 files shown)


zhaoxinyu (approved):

nongfushanquan (approved):

pkg/sink/cloudstorage/path.go
```diff
@@ -207,3 +188,58 @@ func (f *FilePathGenerator) GenerateIndexFilePath(tbl VersionedTable, date strin
...
+
+func (f *FilePathGenerator) getNextFileIdxFromIndexFile(
+	ctx context.Context, tbl VersionedTable, date string,
+) (fileIdx uint64, err error) {
+	indexFile := f.GenerateIndexFilePath(tbl, date)
+	exist, err := f.storage.FileExists(ctx, indexFile)
+	if err != nil {
+		return fileIdx, err
+	}
+	if exist {
```
asddongmen, line 200: May be we can do this first:
```go
if !exist {
  return 0,nil
}
```
So that the complicated logic below can move outside the if block.

asddongmen (approved):

Issue #8508, opened 2023-03-13
CDC changefeed not advance "Can't create more than max_prepared_stmt_count statements (current value: 16382)"


fubinzh:
### What did you do?

1. Create changeeed
2. Run workload (sysbench --db-driver=mysql --mysql-host=upstream-tidb.cdc-testbed-tps-1656790-1-776 --mysql-port=4000 --mysql-user=root --mysql-db=workload --tables=100 --table-size=500000 --create_secondary=off --time=3600 --threads=100 oltp_update_non_index prepare)
3. Run workload (sysbench --db-driver=mysql --mysql-host=upstream-tidb.cdc-testbed-tps-1656790-1-776 --mysql-port=4000 --mysql-user=root --mysql-db=workload --tables=100 --table-size=500000 --create_secondary=off --time=3600 --threads=100 oltp_update_non_index run)
4. After workload running for 30m, inject pod failure for CDC owner
5. After workload running for 60m, inject pod failure again for CDC owner

### What did you expect to see?

cdc checkpoint should advance

### What did you see instead?

1. CDC changefeed not advance during sysbench prepare (no failure injection yet)
2. CDC changefeed resolved ts not advances after cdc owner failure

![image](https://user-images.githubusercontent.com/7403864/224657773-2d9a0fab-179d-4a65-a435-128a1c278fb9.png)

```
[2023/03/12 06:20:32.069 +00:00] [WARN] [mysql.go:728] ["execute DMLs with error, retry later"] [error="[CDC:ErrMySQLTxnErro
r]MySQL txn error: Error 1461 (42000): Can't create more than max_prepared_stmt_count statements (current value: 16382)"] [e
rrorVerbose="[CDC:ErrMySQLTxnError]MySQL txn error: Error 1461 (42000): Can't create more than max_prepared_stmt_count state
ments (current value: 16382)\ngithub.com/pingcap/errors.AddStack\n\tgithub.com/pingcap/errors@v0.11.5-0.20221009092201-b66cd
db77c32/errors.go:174\ngithub.com/pingcap/errors.(*Error).GenWithStackByArgs\n\tgithub.com/pingcap/errors@v0.11.5-0.20221009
092201-b66cddb77c32/normalize.go:164\ngithub.com/pingcap/tiflow/pkg/errors.WrapError\n\tgithub.com/pingcap/tiflow/pkg/errors
[...]

PR #8387, opened 2023-02-28
sink(ticdc): make the size of prep stmt cache adaptive based on the value of max_prepared_stmt_count and worker count

dbsid:
### What problem does this PR solve?

Issue Number: close #8379

### What is changed and how it works?

lower the default size of `prep-stmt-cache-size` based on the default value of MySQL max_prepared_stmt_count, to make sure  when there is single one changfeed, there is no error `Error 1461 (42000): Can't create more than max_prepared_stmt_cou
nt statements (current value: 16382)`

```
	// defaultPrepStmtCacheSize is the default size of prepared statement cache
	// default size = (default max_prepared_stmt_count) / (default worker count + 1)
	// 16382 / (16 + 1) = 963
	defaultPrepStmtCacheSize = 963
```

#### Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

### Release note 

```release-note
Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

If you don't think this PR needs a release note then fill it with `None`.
```

PR #8532, opened 2023-03-15
sink(cdc): fallback when preparing statements meets error

hicqu:

Issue Number: ref #8508 

Fallback to normal SQL executor when preparing statements meets error.

#### Tests 

 - No code

### Release note 

```release-note
None
```


cdc/sink/dmlsink/txn/mysql/mysql.go
```diff
@@ -656,23 +653,25 @@ func (s *mysqlBackend) sequenceExecute(
...
-				}
 
+		var prepStmt *sql.Stmt
+		if s.cachePrepStmts {
+			if stmt, ok := s.stmtCache.Get(query); ok {
+				prepStmt = stmt.(*sql.Stmt)
+			} else if stmt, err := s.db.Prepare(query); err == nil {
+				prepStmt = stmt
 				s.stmtCache.Add(query, stmt)
+			} else {
```
asddongmen, line 664: Could you please add a comment to tell why we need to do this in the second else block?

asddongmen (approved):

cdc/sink/dmlsink/txn/mysql/mysql.go
```diff
@@ -138,13 +137,11 @@ func NewMySQLBackends(
 		// Because each connection can not hold at lease one prepared statement.
 		if maxPreparedStmtCount == 0 || maxPreparedStmtCount/(cfg.WorkerCount+1) == 0 {
 			cachePrepStmts = false
-		} else if maxPreparedStmtCount/(cfg.WorkerCount+1) < prepStmtCacheSize {
-			// if maxPreparedStmtCount/(cfg.WorkerCount+1) < prepStmtCacheSize,
-			// it means that the prepared statement cache is too large on clientsize.
-			// adjust the size of the prepared statement cache on clientsize.
-			// to avoid error `Can't create more than max_prepared_stmt_count statements`
-			prepStmtCacheSize = maxPreparedStmtCount / (cfg.WorkerCount + 1)
```
amyangfei, line 146: https://github.com/pingcap/tiflow/pull/8387 has tried to fixed related issue, seems the `prepStmtCacheSize` is not set to a proper value.

hicqu: The solution can't resolve the situation with more than one CDC instances. So I remove `cacheSize`. Now cache evict is triggered by errors returned by `db.Prepare`.

cdc/sink/dmlsink/txn/mysql/mysql.go
```diff
@@ -122,29 +127,28 @@ func NewMySQLBackends(
...
 		}
+	}
+
+	var stmtCache *lru.Cache
+	if cachePrepStmts {
 		stmtCache, err = lru.NewWithEvict(prepStmtCacheSize, func(key, value interface{}) {
```
dbsid, line 152: I think we can remove the const prepStmtCacheSize, and replace it with maxPreparedStmtCount

cdc/sink/dmlsink/txn/mysql/mysql.go
```diff
@@ -122,29 +127,28 @@ func NewMySQLBackends(
...
-	var stmtCache *lru.Cache
 	if cachePrepStmts {
 		// query the size of the prepared statement cache on serverside
 		maxPreparedStmtCount, err := pmysql.QueryMaxPreparedStmtCount(ctx, db)
 		if err != nil {
 			return nil, err
 		}
+		if maxPreparedStmtCount == -1 {
+			// NOTE: seems TiDB doesn't follow MySQL's specification.
+			maxPreparedStmtCount = math.MaxInt
```
dbsid, line 138: Should we change `math.MaxInt` to 1048576?

hicqu: I guess it's unnecessary. Even if in MySQL we still can set it bigger than 1048576.
So `-1` just means nolimit, so let's just keep it.

cdc/sink/dmlsink/txn/mysql/mysql.go
dbsid: although this does not fix the situation with more than one CDC instances. It set the reasonable limit for the cache size. This value is much lower than `16*1024` when there is more than one worker count. If we hard code the cache size with `16*1024`, we might need to keep try db.Prepare() then got error and fallback to query.

dbsid (approved):

amyangfei (approved):

Issue #8565, opened 2023-03-17
The `canal_json_storage_basic` test failed


hi-rustin:
### Which jobs are flaking?

`canal_json_storage_basic` test

### Which test(s) are flaking?

`canal_json_storage_basic` test

### Jenkins logs or GitHub Actions link

None

### Anything else we need to know

None

PR #8566, opened 2023-03-17
cmd/storage-consumer(ticdc): fix broken tests

hi-rustin:

Issue Number: close https://github.com/pingcap/tiflow/issues/8565

Always use the tidb extension for storage consumers when it using `canal-protocl`.
If we don't enable it, then the commit ts of every row is `0`. It can not advance the table sink. So we can not pass the test.
The reason why it is working before is we use `sinkv1`, so it can use `0` to flush the MySQL sink(all rows got `0` commit ts). After https://github.com/pingcap/tiflow/pull/8188, it does not work anymore.

#### Tests 

 - Integration test

No
No
### Release note 

```release-note
None
```


nongfushanquan (approved):

tests/integration_tests/csv_storage_multi_tables_ddl/conf/error-2.toml
```diff
@@ -0,0 +1,5 @@
+[filter]
+rules = ["multi_tables_ddl_test.t9", "multi_tables_ddl_test.t10"]
+
+[sink.csv]
+include-commit-ts = true
```
CharlesCheung96, line 5:
```suggestion
include-commit-ts = true

```

tests/integration_tests/csv_storage_multi_tables_ddl/conf/normal.toml
```diff
@@ -0,0 +1,5 @@
+[filter]
+rules = ["multi_tables_ddl_test.t1", "multi_tables_ddl_test.t2", "multi_tables_ddl_test.t3", "multi_tables_ddl_test.t4" ,"multi_tables_ddl_test.t1_7","multi_tables_ddl_test.t2_7", "multi_tables_ddl_test.finish_mark"]
+
+[sink.csv]
+include-commit-ts = true
```
CharlesCheung96, line 5:
```suggestion
include-commit-ts = true

```

tests/integration_tests/csv_storage_multi_tables_ddl/run.sh
CharlesCheung96, line 1: Is it necessary to add the DDL test for canal-json?

CharlesCheung96 (approved):

PR #8567, opened 2023-03-17
codec(ticdc): avro decode float by using float32

3AceShowHand:

Issue Number: close #8490

* parse `mysql.TypeFloat` by using `float32`

#### Tests 

 - Unit test

### Release note 

```release-note
Fix avro float type precision by using float32 to parse the data correctly
```


3AceShowHand:
> Do we need to update any docs?

yes, I will update the doc and link the issue

sdojjy (approved):

overvenus (commented):
Without updating schema registry, does it compatible with old schema?

PR #8633, opened 2023-03-22
ddl_puller (ticdc): only create a schemStorage in a changefeed or processor

asddongmen:

Issue Number: ref #8106 

Before this PR, `ddlPuller` held an exclusive `schemaSnapshot` that contained the metadata of all tables in the upstream TiDB. `ddlPuller` needed to use the information in this `schemaSnapshot` to process the received DDL events. At the same time, the `owner` or `processor` also held a struct named `schemaStorage`, which contained the same information as `ddlPuller`'s `schemaSnapshot`.

Therefore, by simply allowing `ddlPuller` and `owner` or `processor` to share a `schemaStorage`, the memory consumption of the original `schemaSnapshot` can be reduced.

This pull request removes the exclusive `schemaSnapshot` from `ddlPuller`, which should decrease the amount of memory required to store tableInfo. The hope is that this change will reduce memory consumption for storing table Infos by up to 50%.
Below are the memory usage statistics for TiCDC, both with and without this PR, in a scenario with 100,000 tables.

**Without this PR:**
image

**With this PR:**
image

#### Tests 

### Release note 

```release-note
None
```


asddongmen:
@purelind  要用这个命令 `/run-kafka-integration-test`

nongfushanquan (commented):
could you add some comments  to describe the main idea about how  to reduce the memory consumed

asddongmen:
> could you add some comments to describe the main idea about how to reduce the memory consumed
@nongfushanquan 
Done, it's added in this PR's description. PTAL. Thank you!

hicqu (approved):

sdojjy (approved):

PR #8671, opened 2023-03-24
DM: fix default physical import mode disk quota

lichunzhu:

Issue Number: close https://github.com/pingcap/tiflow/issues/8672

Modify lightnig's disk quota when we set DM's `disk-quota-physical` a non-negative value.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


lance6716 (approved):

gozssky (approved):

niubell (approved):
LGTM

PR #8676, opened 2023-03-24
server(ticdc): add server memory limit

### LINKED ISSUE #8675: Introduce gomemlimit

Go1.19 introduces [debug.SetMemoryLimit](https://pkg.go.dev/runtime/debug#SetMemoryLimit) to set soft memory limit of a running process, which can help to avoid OOM in some scenarios.

More references:
- https://tip.golang.org/doc/gc-guide
- https://weaviate.io/blog/gomemlimit-a-game-changer-for-high-memory-applications
- https://dzone.com/articles/how-we-solved-an-oom-issue-in-tidb-with-gomemlimit

Add a server config to control memory limit

_No response_

_No response_

amyangfei:

Issue Number: close #8675

- Set default value of `max-memory-percentage` to 0 to disable memory limit, because in most scenarios TiCDC has frequent memory allocation, we don't set soft limit to maximize performance
- When OOM happens, user can config `max-memory-percentage` to avoid it.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


pkg/config/server_config.go
```diff
@@ -44,6 +44,10 @@ const (
 
 	// DefaultChangefeedMemoryQuota is the default memory quota for each changefeed.
 	DefaultChangefeedMemoryQuota = 1024 * 1024 * 1024 // 1GB.
+
+	// DefaultMaxMemoryPercentage is the default max memory percentage
+	// cdc server will set memory limit to 70% of the memory limit of this process.
+	DefaultMaxMemoryPercentage = 70
```
amyangfei, line 50:
```suggestion
	// cdc server will disable soft memory limit by default.
	DefaultMaxMemoryPercentage = 0
```
Change it later.

asddongmen (approved):

pkg/config/server_config.go
```diff
@@ -166,6 +171,7 @@ type ServerConfig struct {
 	KVClient            *KVClientConfig `toml:"kv-client" json:"kv-client"`
 	Debug               *DebugConfig    `toml:"debug" json:"debug"`
 	ClusterID           string          `toml:"cluster-id" json:"cluster-id"`
+	MaxMemoryPercentage int             `toml:"max-memory-percentage" json:"max-memory-percentage"`
```
CharlesCheung96, line 174: Why not use the `uint` type?

nongfushanquan (approved):

pkg/config/server_config.go
amyangfei: Not very necessary, we treat both negative value and zero as disabled option.

Issue #8657, opened 2023-03-23
Improve processor management and error handling


hicqu:
### Is your feature request related to a problem?

* If `sink` error occurs, it's better to let `puller` and `redo` keep running;
* If `puller` error occurs, it's better to let `sink` and `redo` keep running;
    - although generally `puller` doesn't emit errors, seems the only error can be "etcd context canceled"
* If  `redo` error occurs, it's better to let `sink` keep running;
* If etcd session is broken, `processor.Close` should be called correctly, or some stale context will be left. For example, `SortEngine` instances.
* Many sub components of `Processor` use `errCh` directly. When pushing an error into the channel, some different styles are used:
    - with a `default` branch in `select`; so the error can be dropped silenctly;
    - with a `case ctx.Done()` branch in `select`, which means the processor has been closed correctly.
    - Personally I prefer the second style, but code needs to be simplified.

### Describe the feature you'd like

I want to
* Seperate `Processor` lifetime from etcd session. `Processor` can only be closed explicitly instead of canceling the etcd session's context;
* Ensure `Processor.Close` will always be called correctly when destroying a `Processor` instance;
* Unify and simplify error handling in all sub components of `Processor`;
* Isolate error handling of `sink`, `puller` and `redo`, so that one component failure won't break whole system.

PR #8695, opened 2023-03-28
capture(cdc): always close processor if etcd sesison is broken

hicqu:

Issue Number: ref #8657 

Without the PR if a capture's etcd session is broken, `processor.Close` won't be called correctly. It will leak some resources of `processor`. For now it won't cause critical results but we should fix the behavior, which will make code more clear.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/capture/capture.go
```diff
@@ -599,7 +599,7 @@ func (c *captureImpl) AsyncClose() {
 	c.captureMu.Lock()
 	defer c.captureMu.Unlock()
 	if c.processorManager != nil {
-		c.processorManager.AsyncClose()
+		c.processorManager.Close()
```
asddongmen, line 602: May be we also need to change the `capture.AsyncClose` to  `capture.Close`.

cdc/processor/manager.go
```diff
@@ -183,16 +183,13 @@ func (m *managerImpl) closeProcessor(changefeedID model.ChangeFeedID, ctx cdcCon
...
-// AsyncClose sends a signal to Manager to close all processors.
-func (m *managerImpl) AsyncClose() {
-	timeout := 3 * time.Second
-	ctx, cancel := context.WithTimeout(context.TODO(), timeout)
-	defer cancel()
-	done := make(chan error, 1)
-	err := m.sendCommand(ctx, commandTpClose, nil, done)
-	if err != nil {
-		log.Warn("async close failed", zap.Error(err))
+// Close the manager itself and all processors. Can't be called with `Tick`.
```
asddongmen, line 186:
```suggestion
// Close the manager itself and all processors. 
// Note: This method must not be called with `Tick`. Please be careful.
```

asddongmen (approved):

cdc/capture/capture.go
hicqu: We will consider it, when examine `owner.AsyncStop`.

cdc/processor/manager.go
```diff
@@ -159,11 +159,11 @@ func (m *managerImpl) Tick(stdCtx context.Context, state orchestrator.ReactorSta
 	return state, nil
 }
 
-func (m *managerImpl) closeProcessor(changefeedID model.ChangeFeedID, ctx cdcContext.Context) {
+func (m *managerImpl) closeProcessor(changefeedID model.ChangeFeedID) {
```
CharlesCheung96, line 162: Would you please add some comment to explain that this function may stuck the tick function?

CharlesCheung96 (approved):

cdc/processor/manager.go
hicqu: I add comment for `processor.Tick` and `processor.Close`, which can be called in `processor.Manager.Tick`.

PR #8697, opened 2023-03-29
schemaStorage (ticdc): Filter out schema information that is irrelevant to changefeed.

asddongmen:

Issue Number: ref #8106

#### Tests 

 - Unit test
1. Create two databases, `test` and `ntest`, in the upstream TiDB. `test` has 100,000 tables and `ntest` has 1 table.
2. Create a changefeed with the following configuration:
```toml
[filter]
rules = ["ntest.*"] # only replicate tables in `ntest` DB. 
```
Memory consumption without this PR:
image
Memory consumption with this PR:
image

From the above tests, it can be concluded that in such extreme situations, TiCDC server's memory can **be reduced by up to 99%**. After this PR, TiCDC can support run more changefeeds.

### Release note 

```release-note
Reduce memory consumption by filter out schema information that is not relevant to a changefeed.
```


cdc/entry/schema/snapshot.go
```diff
@@ -139,6 +144,10 @@ func NewSnapshotFromMeta(meta *timeta.Meta, currentTs uint64, forceReplicate boo
 	tag := negative(currentTs)
 
 	for _, dbinfo := range dbinfos {
+		if filter.ShouldIgnoreSchema(dbinfo.Name.O) {
+			log.Info("ignore database", zap.String("db", dbinfo.Name.O))
```
amyangfei, line 148: Will this log too much, ditto for L165, if there exist many schemas or tables

cdc/entry/schema_storage.go
```diff
@@ -78,9 +79,9 @@ func NewSchemaStorage(
 	)
 	if meta == nil {
 		snap = schema.NewEmptySnapshot(forceReplicate)
-		snap.InitPreExistingTables()
+		// snap.InitPreExistingTables()
```
amyangfei, line 82: Why is this init phase commented?

asddongmen: I forgot to remove this line. Actually, `InitPreExistingTables()` is removed in this PR.

cdc/entry/schema/snapshot.go
asddongmen: It may be so, I will change it to debug level.

amyangfei (approved):

cdc/puller/ddl_puller_test.go
```diff
@@ -647,15 +657,22 @@ func TestResolvedTsStuck(t *testing.T) {
 	mockPuller := newMockPuller(t, startTs)
 	ctx := cdcContext.NewBackendContext4Test(true)
 	up := upstream.NewUpstream4Test(nil)
+	f, err := filter.NewFilter(config.GetDefaultReplicaConfig(), "")
+	require.Nil(t, err)
 	schemaStorage, err := entry.NewSchemaStorage(nil,
 		startTs,
-		true,
+		ctx.ChangefeedVars().Info.Config.ForceReplicate,
```
hicqu, line 664: Why we change `true` to `Config.ForceReplicate`? It was a bug?

cdc/entry/schema/snapshot.go
```diff
@@ -203,60 +215,6 @@ func NewEmptySnapshot(forceReplicate bool) *Snapshot {
...
-)
-
-// InitPreExistingTables initializes the pre-existing tables in an empty Snapshot.
-// Since v6.2.0, tables of concurrent DDL will be directly written as meta KV in
-// TiKV, without being written to history DDL jobs. So the Snapshot which is not
-// build from meta needs this method to handle history DDL.
-// Since v6.5.0, Backfill tables is written as meta KV in TiKV, so the Snapshot
-// which is not build from meta needs this method to handle history DDL.
-// See:https://github.com/pingcap/tidb/pull/39616
-func (s *Snapshot) InitPreExistingTables() {
```
hicqu, line 220: Could you explain why we remove this?

cdc/puller/ddl_puller_test.go
asddongmen: There was no bug; I just tried to keep it the same with line 671. There is a changefeedCtx that contains the configuration of a changefeed, so using the changefeed's configuration instead of a magic true seems more reasonable to me.

cdc/entry/schema/snapshot.go
asddongmen: Based on our analysis, it appears that TiCDC does not rely on the pre-existing system table. As such, I recommend removing this function and discontinuing its use.

hicqu (approved):

Issue #8700, opened 2023-03-30
DM integrate with physical backend config name is not accurate


lichunzhu:
### Is your feature request related to a problem?

1. DM-worker's configuration name needs to be changed.
2. DM-worker's dist-scan-concurrency, build-stats-concurrency, index-serial-scan-concurrency will affect checksum time.
3. DM-worker's embedded lightning precheck can't be skipped, which may cause problems.
4. Lightning's progress status is not accurate. It will be updated after an engine is finished. 
```log
[INFO] [lightning.go:601] ["progress status of lightning"] [task=test] [unit=lightning-load] [finished_bytes=0] [total_bytes=775181295408] [progress="0.00 %"] ["current speed (bytes / seconds)"=0]
```

### Describe the feature you'd like

fix abobe problems

PR #8701, opened 2023-03-30
dm: fix lightning config name and several fix

lichunzhu:

Issue Number: close https://github.com/pingcap/tiflow/issues/8700

1. fix loader config name, remove physical suffix for arguments only exist in physical mode.
2. ignore lightning precheck when all checking items are ignored, to avoid lightning precheck block import
3. add more arguments to improve checksum speed
4. improve lightning status's accuracy

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
remove physical suffix for arguments only exist in physical mode, add distsql-scan-concurrency, index-serial-scan-concurrency, checksum-table-concurrency arguments to accelerate checksum
```


dm/config/task.go
```diff
@@ -286,21 +286,30 @@ const (
...
+	// deprecated, use DiskQuota instead.
+	DiskQuotaPhysical config.ByteSize `yaml:"disk-quota-physical" toml:"disk-quota-physical" json:"disk-quota-physical"`
+	DiskQuota         config.ByteSize `yaml:"disk-quota" toml:"disk-quota" json:"disk-quota"`
+	// deprecated, use Checksum instead.
+	ChecksumPhysical           PhysicalPostOpLevel `yaml:"checksum-physical" toml:"checksum-physical" json:"checksum-physical"`
+	Checksum                   PhysicalPostOpLevel `yaml:"checksum" toml:"checksum" json:"checksum"`
+	Analyze                    PhysicalPostOpLevel `yaml:"analyze" toml:"analyze" json:"analyze"`
+	RangeConcurrency           int                 `yaml:"range-concurrency" toml:"range-concurrency" json:"range-concurrency"`
+	CompressKVPairs            string              `yaml:"compress-kv-pairs" toml:"compress-kv-pairs" json:"compress-kv-pairs"`
+	PDAddr                     string              `yaml:"pd-addr" toml:"pd-addr" json:"pd-addr"`
```
lance6716, line 309: why we need this?

dm/config/task.go
```diff
@@ -286,21 +286,30 @@ const (
...
+	DistSQLScanConcurrency     int                 `yaml:"distsql-scan-concurrency" toml:"distsql-scan-concurrency" json:"distsql-scan-concurrency"`
```
lance6716, line 310: how do user know these fields are not taken effect in logical import

lichunzhu: Maybe adding a physical structure? In Lightning, there are also some arguments that won't take effect in logical mode.

dm/config/task.go
lance6716: as we discussed before, will TiKV address is also a private address so following ingestion still fails?

lichunzhu: PD's address is from TiDB's path setting. It's filled by users and may be a local network address. TiKV's address is its advertise address. We may be able to enter this address.

lance6716 (approved):

PR #8780, opened 2023-04-12
mounter(ticdc): support row level checksum verification

### LINKED ISSUE #8777: Mounter support checksum calculation

### LINKED ISSUE #8791: Add new unit test in the mounter to cover checksum calculation

3AceShowHand:

Issue Number: close #8777, close #8791

* Extract the checksum and compare it.

#### Tests 

 - Unit test

### Release note 

```release-note
Mounter support row level checksum verification
```


cdc/entry/mounter.go
```diff
@@ -332,11 +474,29 @@ func (m *mounter) mountRowKVEntry(tableInfo *model.TableInfo, row *rowKVEntry, d
...
 			return nil, rawRow, errors.Trace(err)
 		}
 
+		preChecksum, matched, err = m.verifyChecksum(columnIDs, preRawCols, true)
+		if err != nil {
+			return nil, rawRow, errors.Trace(err)
+		}
+
+		if !matched {
+			log.Error("pre columns checksum mismatch",
```
overvenus, line 488: What does `pre` mean?

3AceShowHand: I mean the previous columns, I will refine the log.

overvenus (approved):
Could you add an integration test?

3AceShowHand:
> Could you add an integration test?

I will add an integration test to cover checksum functionality after tidb fully support the feature.

cdc/entry/mounter.go
```diff
@@ -314,16 +383,94 @@ func datum2Column(tableInfo *model.TableInfo, datums map[int64]types.Datum, fill
...
 	}
-	return cols, rawCols, nil
+
+	var decoder *rowcodec.DatumMapDecoder
+	if isPreRow {
+		decoder = m.preDecoder
+	} else {
+		decoder = m.decoder
+	}
+	if decoder == nil {
```
asddongmen, line 412: seems like this will never happen? can use panic instead of error here, which will make debugging easier.

errors.toml
```diff
@@ -186,6 +186,11 @@ error = '''
 failed to convert ddl '%s' to filter event type
 '''
 
+["CDC:ErrCorruptedDataMutation"]
```
asddongmen, line 189: Please add this error to `var changefeedUnRetryableErrors `.

cdc/entry/mounter.go
3AceShowHand: I would prefer error, to avoid one single changefeed makes the whole server crashed

asddongmen (approved):

cdc/entry/mounter.go
```diff
@@ -314,16 +383,94 @@ func datum2Column(tableInfo *model.TableInfo, datums map[int64]types.Datum, fill
...
+	if !ok {
+		log.Error("cannot found the extra checksum, the first checksum mismatched",
+			zap.Uint32("checksum", checksum),
+			zap.Uint32("first", first),
+			zap.Uint32("extra", extra))
+		return checksum, false, errors.New("cannot found the extra checksum from the event")
+	}
+
+	if checksum == extra {
+		log.Warn("extra checksum matched, this may happen the upstream TiDB is during the DDL"+
```
zhangjinpeng1987, line 444: Please change it to debug level, or the log will make ticdc out of disk when there are many ddls in TiDB.

3AceShowHand: Get it.

Issue #8798, opened 2023-04-17
Some bugs when TiCDC setting the timezone


hi-rustin:
### What did you do?

Set the timezone to TiCDC server and sink-uri. [How does TiCDC set the time zone internally?](https://pingcap.feishu.cn/docx/DcGMd41xsoo2NWxZRrzcCQKknfb#Q2cid8YCgoKkeWxsfyMc4DjLnZb)

### What did you expect to see?

TiCDC uses timezone correctly.

### What did you see instead?

  1. [ ] We unescape the timezone value twice.
  2. [ ] We use the wrong ctx when we get the timezone.
  3. [ ] We should not suggest users use CST to create the changefeed in docs.
  4. [ ] We should not suggest users use time-zone="" in sink-uri.
  5. [ ] Should we only keep the time-zone from the sink-uri? Considering multi-tenancy, we should not use server configurations.
  6. [ ] TiCDC can not handle the timezone for some DDLs correctly. We need to add docs for it.

### Versions of the cluster

TiCDC version (execute `cdc version`):

```console
v6.5.1
```

PR #8799, opened 2023-04-17
pkg/tz(ticdc): use correct ctx to get timezone

hi-rustin:

Issue Number: ref https://github.com/pingcap/tiflow/issues/8798

- Fix a bug that unescape the timezone value twice
- Fix a bug we use the wrong ctx when we get the timezone.
- Warning when users use time-zone="" in sink-uri.
- Warning when users do not set the timezone in sink-uri

#### Tests 

 - Unit test
     - https://github.com/pingcap/tiflow/pull/8799#issuecomment-1512417945
     - https://github.com/pingcap/tiflow/pull/8799#issuecomment-1512438022

None
None

### Release note 

```release-note
Fix the bug that TiCDC always uses `UTC` when creating changefeed
```


hi-rustin:
Test it locally to make sure we use the correct timezone from the ctx.
1. install the MySQL with homebrew, it would use the system timezone as MySQL timezone
2. start upstream tidb cluster, build and ticdc with this PR
3. start ticdc server without timezone config(it would use system)
4. try to create changefeed, it would use the Asia/Shanghai instead of UTC.
```console
➜  bin git:(rustin-patch-timezone) ✗ ./cdc cli changefeed create --sink-uri="mysql://root:123456@127.0.0.1:3306" -c "test1"
Error: [CDC:ErrMySQLConnectionError]fail to open MySQL connection: Error 1298 (HY000): Unknown or incorrect time zone: 'Asia/Shanghai'
```

hi-rustin:
Test it locally to make sure it will bail out when the changefeed timezone is different from server.
1. install the MySQL with homebrew, it would use the system timezone as MySQL timezone
2. start upstream tidb cluster, build and ticdc with this PR
3. start ticdc server without timezone config(it would use system)
4. try to create changefeed, it would bail out an error.
```console
➜  bin git:(rustin-patch-timezone) ./cdc cli changefeed create --sink-uri="mysql://root:123456@127.0.0.1:3306?time-zone=UTC" -c "test3"
Error: [CDC:ErrMySQLInvalidConfig]MySQL config invalid: the timezone of the TiCDC server and the sink-uri are inconsistent. TiCDC server timezone: Asia/Shanghai, sink-uri timezone: UTC. Please make sure that the timezone of the TiCDC server, sink-uri and the downstream database are consistent.
```

pkg/sink/mysql/config.go
```diff
@@ -330,28 +331,48 @@ func getSafeMode(values url.Values, safeMode *bool) error {
...
 	if err != nil {
 		return cerror.WrapError(cerror.ErrMySQLInvalidConfig, err)
 	}
-	_, err = time.LoadLocation(value)
-	if err != nil {
-		return cerror.WrapError(cerror.ErrMySQLInvalidConfig, err)
+	*timezone = fmt.Sprintf(`"%s"`, changefeedTimezone.String())
+	// We need to check whether the timezone of the TiCDC server and the sink-uri are consistent.
+	// If they are inconsistent, it may cause the data to be inconsistent.
+	if changefeedTimezone.String() != serverTimezone.String() {
```
nongfushanquan, line 367: If the timezone sink in the uri is set by the user , the timezone of the server should be ignored

nongfushanquan (approved):

pkg/sink/mysql/config.go
hi-rustin: We can do it in the future. Because we need to update the init order for the mounter.

amyangfei (approved):

PR #8836, opened 2023-04-23
pkg/config(ticdc): hide fields that are not required for specific protocols

### LINKED ISSUE #8824: Refine the content of the configure file when executing the query command by cli
### What did you do?

When the user executes the query command with CLI, some irrelevant information remains.

only useful information will be showed.

### What did you see instead?

as title described

### Versions of the cluster

6.1 & 6.5

charleszheng44:

Issue Number: close #8824 

Hide fields that are not required for specific protocols and downstream types. Since we utilize a common changefeed configuration template, certain fields may not be utilized for certain protocols.

#### Tests 

Execute a manual test on a TiDB cluster generated with TiUP playground, and include a partial output of the `cdc cli changefeed query` command. The output has been modified to exclude any unnecessary fields, such as `csv` and `schema_registry`, as well as other fields that are not utilized.
```json
{
  "upstream_id": 7225670347271394026,
  "namespace": "default",
  "id": "simple-replication-task",
  "sink_uri": "kafka://127.0.0.1:9092/quickstart-events?protocol=canal-json",
  "config": {
    "memory_quota": 1073741824,
    "case_sensitive": true,
    "enable_old_value": true,
    "force_replicate": false,
    "ignore_ineligible_table": false,
    "check_gc_safe_point": true,
    "enable_sync_point": false,
    "bdr_mode": false,
    "sync_point_interval": 600000000000,
    "sync_point_retention": 86400000000000,
    "filter": {
      "rules": [
        "*.*"
      ],
      "event_filters": null
    },
    "mounter": {
      "worker_num": 16
    },
    "sink": {
      "protocol": "canal-json",
      "transaction_atomicity": "",
      "encoder_concurrency": 16,
      "terminator": "\r\n",
      "date_separator": "none",
      "enable_partition_separator": true,
      "enable_kafka_sink_v2": false,
      "only_output_updated_columns": false
    },
    "consistent": {
      "level": "none",
      "max_log_size": 64,
      "flush_interval": 2000,
      "storage": "",
      "use_file_backend": false
    },
    "scheduler": {
      "enable_table_across_nodes": false,
      "region_threshold": 100000,
      "write_key_threshold": 0
    },
    "integrity": {
      "integrity_check_level": "none",
      "corruption_handle_level": "warn"
    }
  },
  "create_time": "2023-04-24 10:33:13.965",
  "start_ts": 441019949069107204,
  "resolved_ts": 441019961560268803,
  "target_ts": 0,
  "checkpoint_tso": 441019961560268803,
  "checkpoint_time": "2023-04-24 10:34:01.450",
  "state": "normal",
  "creator_version": "v7.1.0-master-dirty",
  "task_status": [
    {
      "capture_id": "90d43a75-9047-4b3f-a7ab-37e6a3bff357",
      "table_ids": [],
      "table_operations": null
    }
  ]
}
```

#### Questions 
N/A

N/A

N/A

### Release note 

```release-note
`None`
```


asddongmen:
Could you please add a unit test to cover this change? Thank you!

charleszheng44:
> Could you please add a unit test to cover this change? Thank you!

Test cases added

pkg/util/pointer.go
```diff
@@ -0,0 +1,27 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+// GetValueOrDefault returns the value pointed to by p, or a zero value of
+// its type if p is nil.
+func GetValueOrDefault[T any](p *T) T {
```
CharlesCheung96, line 18: Would it be more appropriate to name it `GetOrZero`.

tests/integration_tests/api_v2/model.go
```diff
@@ -234,16 +234,16 @@ type Table struct {
 // SinkConfig represents sink config for a changefeed
 // This is a duplicate of config.SinkConfig
 type SinkConfig struct {
-	Protocol                 string            `json:"protocol"`
-	SchemaRegistry           string            `json:"schema_registry"`
-	CSVConfig                *CSVConfig        `json:"csv"`
+	Protocol                 string            `json:"protocol,omitempty"`
```
CharlesCheung96, line 237: Whether all fields should use Pointer?

pkg/config/replica_config.go
```diff
@@ -56,11 +58,12 @@ var defaultReplicaConfig = &ReplicaConfig{
 			Delimiter:  Comma,
 			NullString: NULL,
 		},
-		EncoderConcurrency:       16,
+		EncoderConcurrency:       util.AddressOf(16),
 		Terminator:               CRLF,
 		DateSeparator:            DateSeparatorNone.String(),
-		EnablePartitionSeparator: true,
-		EnableKafkaSinkV2:        false,
+		EnablePartitionSeparator: util.AddressOf(true),
```
CharlesCheung96, line 64: Should the default value here be set in `validateAndAdjust` according to sink type?  eg, set `EnablePartitionSeparator` to true only when using storage sink.

tests/integration_tests/api_v2/model.go
charleszheng44: I think we do not need to use pointer for a string, as an empty string `var s string = ""` will be ignored automatically.

pkg/util/pointer.go
charleszheng44: Agree

sdojjy (approved):

asddongmen (approved):

asddongmen (approved):

CharlesCheung96 (approved):

Issue #8686, opened 2023-03-27
Optimize the synchronization behavior of the lossy DDL operation 


nongfushanquan:
### Is your feature request related to a problem?

When a customer runs certain DDLs on a big table, such as modifying a column that affects all rows of the table, all rows will replicate to the downstream

### Describe the feature you'd like

 the ticdc will ignore the dml and only sync ddl to the downstream.

PR #8912, opened 2023-05-08
tests(ticdc): add exchange partition test to lossy DDL case

hi-rustin:

Issue Number: close https://github.com/pingcap/tiflow/issues/8686

add exchange partition test to lossy DDL case

#### Tests 

 - Integration test
No
No
### Release note 

```release-note
None
```


tests/integration_tests/lossy_ddl/data/prepare.sql
```diff
@@ -164,6 +164,21 @@ INSERT INTO example15 (id, b)
 VALUES (1, -1);
 ALTER TABLE example15 MODIFY COLUMN b INT UNSIGNED;
 
+-- exchange partition
+CREATE TABLE example16
+(
+    a INT PRIMARY KEY
```
CharlesCheung96, line 170: Whether it is better to insert one piece of data into each partition?

CharlesCheung96 (approved):

PR #8923, opened 2023-05-09
p2p(ticdc): remove message peer when capture is removed

### LINKED ISSUE #8922: p2p(ticdc): remove message peer when capture is removed
p2p(ticdc): remove message peer when capture is removed

sdojjy:

If an owner is killed by "kill -19",  other CDC nodes will remove that capture, but the peer in the message server will not be removed, so the message server still sends ack message to that peer, until the write buffer is full.

Issue Number: close #8922

remove message peer when capture is removed

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
 `None`.
```


pkg/p2p/server.go
```diff
@@ -287,6 +291,9 @@ func (m *MessageServer) run(ctx context.Context) error {
 					}
 					return errors.Trace(err)
 				}
+			case taskOnDeregisterPeer:
+				log.Debug("taskOnDeregisterPeer", zap.String("peerID", task.peerID))
```
nongfushanquan, line 295: Log.Info level is suggested here

cdc/capture/capture.go
```diff
@@ -459,6 +459,11 @@ func (c *captureImpl) campaignOwner(ctx cdcContext.Context) error {
 		})
 		globalState.SetOnCaptureRemoved(func(captureID model.CaptureID) {
 			c.MessageRouter.RemovePeer(captureID)
+			if err := c.MessageServer.ScheduleDeregisterPeerTask(ctx, captureID); err != nil {
```
overvenus, line 462: Could you add some comments?

overvenus (approved):

asddongmen (approved):

pkg/p2p/server.go
sdojjy: done

cdc/capture/capture.go
sdojjy: done

PR #8930, opened 2023-05-10
scheduler (ticdc): add region number limit for table span 

### LINKED ISSUE #8980: When enabling the table split feature, limit the number of regions in each tableSpan.

Currently, the scheduler determines the number of spans based on the number of captures when splitting tables. This can lead to too many regions in a single span, which can cause an uneven distribution of spans when scaling out the ticdc cluster.

This problem can be mitigated by limiting the number of regions in each span.

_No response_

_No response_

asddongmen:

Issue Number: close #8980 

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/scheduler/internal/v3/keyspan/reconciler.go
```diff
@@ -30,6 +30,8 @@ import (
 	"go.uber.org/zap"
 )
 
+const spanRegionLimit = 50000
```
3AceShowHand, line 33: why you choose 5000 ? please add some comment.

cdc/scheduler/internal/v3/keyspan/reconciler.go
```diff
@@ -69,6 +71,7 @@ func NewReconciler(
 		changefeedID: changefeedID,
 		config:       config,
 		splitter: []splitter{
+			// write splitter has the highest priority.
```
3AceShowHand, line 74: why it is the highest one ?

cdc/scheduler/internal/v3/keyspan/reconciler.go
asddongmen: We don't want a span to be too large or too small. Currently, this value is inferred based on user scenarios.

cdc/scheduler/internal/v3/keyspan/reconciler.go
asddongmen: This is by design. When the user enables the feature to split tables based on input volume, it will have higher priority.

3AceShowHand (approved):

CharlesCheung96 (approved):

PR #8949, opened 2023-05-11
sink(cdc): handle sink errors more fast and light

hicqu:
### What problem does this PR solve?

Issue Number: ref #8657 

### What is changed and how it works?

Handle sink errors more fast and light, without rebuilding the whole changefeed.

#### Tests 

 - Unit test
 - Manual test (add detailed scripts or steps below)

### Release note 

```release-note
None
```

PR #8989, opened 2023-05-18
(sink/cdc): fix stop table dead lock introduced in #8949

hicqu:

Issue Number: ref #8657 

The dead lock is caused by `tableSinkWrapper.tableSinkMu`:
* one goroutine is calling `tableSinkWrapper.close`, which requires the lock;
* `tableSinkWrapper.close` can be blocked for a long while if the backend sink is hang;
* now another goroutine wants to close the processor, but it can't get the lock;
* so the processor can't be closed until the backend sink hang is over.

This should fix capture_suicide_while_balance_table failures.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/processor/sinkmanager/table_sink_worker.go
+++ b/cdc/processor/sinkmanager/table_sink_worker.go
@@ -180,7 +180,7 @@ func (w *sinkWorker) handleTask(ctx context.Context, task *sinkTask) (finalErr e
 			// events have been reported. Then we can continue the table
 			// at the checkpoint position.
 			case tablesink.SinkInternalError:
-				task.tableSink.clearTableSink()
+				task.tableSink.clearTableSink(true)
 				// Restart the table sink based on the checkpoint position.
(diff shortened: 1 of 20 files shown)


cdc/processor/sinkmanager/table_sink_worker.go
```diff
@@ -180,7 +180,7 @@ func (w *sinkWorker) handleTask(ctx context.Context, task *sinkTask) (finalErr e
 			// events have been reported. Then we can continue the table
 			// at the checkpoint position.
 			case tablesink.SinkInternalError:
-				task.tableSink.clearTableSink()
+				task.tableSink.clearTableSink(false)
```
nongfushanquan, line 183: please add some comments here to describe  the meaning of  "false"

nongfushanquan (approved):

sdojjy (approved):

PR #9055, opened 2023-05-25
ticdc(integration-test): use wget instead to curl to fetch binaries 

charleszheng44:

Issue Number: None

use wget instead to curl to fetch binaries  and some minor fix for community binaries fetch

#### Tests 

 - Unit test
 - Integration test
 - No code

No

No

### Release note 

```release-note
`None`.
```


CharlesCheung96 (approved):

hi-rustin (approved):
Thanks

PR #9163, opened 2023-06-09
puller(cdc): share grpc streams and goroutines in tables

### LINKED ISSUE #9222: cdc: unbound resources from changefeed tables

Currently many resources are bounded with changefeed tables:
* goroutines
* memory allocation
* CPU usage

If there are lots of idle tables, resources can be wasted.

This issue tracks steps to relax the bound:

- [x] reduce memory usage of unbounded channels
    * #9223 

_No response_

_No response_

hicqu:

Issue Number: close #9222 

Add a new configuration `server.kvclient.enable-multiplexing`. When it's enabled:
* tables in one changefeed will share grpc streams and region workers;
* tables in one changefeed will share goroutines.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


cdc/kv/regionlock/region_range_lock.go
```diff
@@ -492,4 +524,5 @@ type LockedRangeValue struct {
 	RegionID     uint64
 	CheckpointTs uint64
```
CharlesCheung96, line 525: Better to reuse `LockedRange` in `LockedRangeValue `.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1184 @@
...
+	defer gcTicker.Stop()
+LOOP:
+	for {
+		var task resolveLockTask
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-gcTicker.C:
+			gcResolveLastRun()
+			continue LOOP
```
CharlesCheung96, line 1056: There is only one layer of loop, why LOOP is needed?

cdc/kv/regionlock/region_range_lock.go
hicqu: `LockedRange` contains atomic things. I prefer to avoid atomic variables in the return value of `CheckLockedRanges`.

cdc/kv/shared_client.go
hicqu: I'm not sure whether `continue` can jump out from `select` block or not.

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,423 @@
...
+}
+
+func (p *tableProgress) resolveLock() {
+	if !p.initialized || time.Since(p.resolvedTsUpdated) < resolveLockFence {
+		return
+	}
+	resolvedTs := p.resolvedTs.Load()
+	resolvedTime := oracle.GetTimeFromTS(resolvedTs)
+	currentTime := p.client.GetPDClock().CurrentTime()
+	if !currentTime.After(resolvedTime) || currentTime.Sub(resolvedTime) < resolveLockFence {
```
CharlesCheung96, line 400:
```suggestion
	if currentTime.Sub(resolvedTime) < resolveLockFence {
```

cdc/kv/regionlock/region_range_lock.go
CharlesCheung96: From an implementation point of view, `resolveLock` is a relatively [low-frequency operation](https://github.com/pingcap/tiflow/pull/9163/files#diff-c82af6de18a9d7ddc2bba61db2c56944891008ec6f1cc76eb1d2b15fc68b4178R394), so the cost of reusing atomic variables is almost negligible. 

By contrast, creating multiple duplicate entities will add unnecessary complexity to the code, which is likely to be detrimental to future maintenance. At present, the concept of `LockedRange`, `LockRangeResult` and `LockedRangeValue` has been confused.

cdc/kv/shared_client.go
CharlesCheung96: continue takes effect only for the for loop, ref: https://github.com/golang/go/issues/43889.

Besides, only one select branch is executed in each for loop, so perhaps continue is not needed here?

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1184 @@
...
+
+type resolveLockTask struct {
+	regionID   uint64
+	maxVersion uint64
+	state      *regionlock.LockedRange
+	enter      time.Time
+}
+
+type rangeTask struct {
+	span           tablepb.Span
```
CharlesCheung96, line 107: Is `rangeTask.span` equal to `requestedTable .span`?

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1184 @@
...
+
+func hashRegionID(regionID uint64, slots int) int {
+	b := make([]byte, 8)
+	binary.LittleEndian.PutUint64(b, regionID)
+	return int(seahash.Sum64(b) % uint64(slots))
+}
+
+var (
+	// To generate a requestID in `newRequestedTable`.
+	requestIDGen atomic.Uint64
```
CharlesCheung96, line 1163:
```suggestion
	subscriptionIDGen atomic.Uint64
```
And please fix the comment on this field.

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,423 @@
...
+
+	x.inputChs = make([]chan kv.MultiplexingEvent, 0, workers)
+	for i := 0; i < workers; i++ {
+		x.inputChs = append(x.inputChs, make(chan kv.MultiplexingEvent, 1024))
+	}
+	return x
+}
+
+// Subscribe some spans. They will share one same resolved timestamp progress.
+func (p *MultiplexingPuller) Subscribe(spans []tablepb.Span, startTs model.Ts, tableName string) []kv.SubscriptionID {
```
CharlesCheung96, line 124: The return value of `Subscribe` has never been processed by its caller, so would it be more appropriate to change it to a method with no return value?

cdc/kv/regionlock/region_range_lock.go
hicqu: The reason to avoid atomics in `CheckLockedRangeResult` is not about performance. Just to make it more clear.

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,423 @@
...
+		defer progress.consume.RUnlock()
+		if !progress.consume.removed {
+			return p.consume(ctx, raw, spans)
+		}
+		return nil
+	}
+
+	subIDs := make([]kv.SubscriptionID, len(spans))
+	for i, span := range spans {
+		subID := p.client.AllocSubscriptionID()
```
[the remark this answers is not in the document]
hicqu: `SubscriptionID` should be exposed to puller so puller can finds `tableProgress` from it.

cdc/kv/shared_client.go
hicqu: `requestedTable.span` contains `rangeTask.span`.

cdc/kv/regionlock/region_range_lock.go
hicqu: I have renamed this function and its return value. Please take a look again.

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,428 @@
...
+	for _, span := range spans {
+		if subID, ok := p.client.Unsubscribe(span); ok {
+			progress := p.subscriptions.m[subID]
+			progress.consume.Lock()
+			progress.consume.removed = true
+			progress.consume.Unlock()
+
+			delete(p.subscriptions.m, subID)
+			p.subscriptions.n.Delete(span)
+			continue
```
CharlesCheung96, line 187: Perhaps it would be clearer to use the else branch here.

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,428 @@
...
+		defer func() {
+			progress.scheduled.Store(false)
+			if len(progress.resolvedSpans) > 0 {
+				p.schedule(ctx, progress)
+			}
+		}()
+
+		var event kv.MultiplexingEvent
+		var spans *model.ResolvedSpans
+		for i := 0; i < 128; i++ {
```
CharlesCheung96, line 309: Why handle resolvedTs events asynchronously? And what happens if consumption is slower here?

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,428 @@
...
+			p.pullerEventCounterKv.Inc()
+			if err := progress.consume.f(ctx, e.Val, progress.spans); err != nil {
+				return errors.Trace(err)
+			}
+		} else if e.Resolved != nil {
+			p.pullerEventCounterResolved.Add(float64(len(e.Resolved.Spans)))
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case progress.resolvedSpans <- e:
```
CharlesCheung96, line 260: Is it possible for the worker to get stuck here?

cdc/puller/multiplexing_puller.go
hicqu: By default we uses more threads to handle resolved timestamps. It's helpful for large tables because `Frontier` utilizes too much CPU.

cdc/puller/multiplexing_puller.go
hicqu: I guess no. frontier threads just do some CPU intensive works. If we move those works into this goroutine, it will also stuck.

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,428 @@
...
+			log.Panic("unexist unsubscription",
+				zap.String("namespace", p.changefeed.Namespace),
+				zap.String("changefeed", p.changefeed.ID),
+				zap.String("span", span.String()))
+		}
+	}
+}
+
+// Run the puller.
+func (p *MultiplexingPuller) Run(ctx context.Context) (err error) {
```
CharlesCheung96, line 197: Is it necessary to print a log when exiting?

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,428 @@
...
+	consume func(context.Context, *model.RawKVEntry, []tablepb.Span) error,
+	workers int,
+	hasher func(tablepb.Span, int) int,
+	frontiers int,
+) *MultiplexingPuller {
+	x := &MultiplexingPuller{
+		changefeed: changefeed,
+		client:     client,
+		consume:    consume,
+		workers:    workers,
```
CharlesCheung96, line 108: len(x.inputChs) equals to x.workers, so s.workers should be removed from MultiplexingPuller.

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,428 @@
...
+	spans      []tablepb.Span
+	startTs    model.Ts
+	tableName  string
+
+	initialized          bool
+	resolvedTsUpdated    time.Time
+	resolvedTs           atomic.Uint64
+	maxIngressResolvedTs atomic.Uint64
+
+	resolvedSpans chan kv.MultiplexingEvent
```
CharlesCheung96, line 52:
```suggestion
	resolvedEventsCache chan kv.MultiplexingEvent
```
Maybe resolvedEventsCache is makes more sense. Also, please add comments to explain the reason for caching resolved events and processing them asynchronously.

cdc/puller/multiplexing_puller.go
hicqu: yes, it's reasonable.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1202 @@
...
+			zap.String("namespace", s.changefeed.Namespace),
+			zap.String("changefeed", s.changefeed.ID))
+	}
+
+	s.totalSpans.Lock()
+	var subID SubscriptionID
+	if subID, success = s.totalSpans.m.Get(span); success {
+		rt := s.totalSpans.v[subID]
+		s.totalSpans.Unlock()
+		rt.updateStaleLocks(s, maxVersion)
```
CharlesCheung96, line 273: It is more appropriate to return `resolveLockTask slice` in `updateStaleLocks` to avoid using complex callback function (postUpdateRegionResolvedTs).

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,435 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-resolveLockTicker.C:
+		}
+		for progress := range p.getAllProgresses() {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case progress.resolvedEventsCache <- kv.MultiplexingEvent{}:
```
CharlesCheung96, line 364: Why not call `progress.resolveLock` directly?

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1202 @@
...
+func (s *SharedClient) AllocSubscriptionID() SubscriptionID {
+	return SubscriptionID(subscriptionIDGen.Add(1))
+}
+
+// Subscribe the given table span.
+// NOTE: `span.TableID` must be set correctly.
+func (s *SharedClient) Subscribe(
+	subID SubscriptionID, span tablepb.Span, startTs uint64,
+	eventCh chan<- MultiplexingEvent,
+) (existSubID SubscriptionID, success bool) {
```
CharlesCheung96, line 201: The SubID returned by the function should be used, which avoids calling AllocSubscriptionID externally.

cdc/puller/multiplexing_puller.go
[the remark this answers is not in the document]
CharlesCheung96: Ref: https://github.com/pingcap/tiflow/pull/9163/files#r1301390865

cdc/puller/ddl_puller.go
```diff
@@ -561,6 +582,28 @@ func NewDDLJobPuller(
...
+
+		rawDDLCh := make(chan *model.RawKVEntry, defaultPullerOutputChanSize)
+		mp.sortedDDLCh = memorysorter.SortOutput(ctx, changefeed, rawDDLCh)
+
+		client := kv.NewSharedClient(
+			changefeed, cfg, ddlPullerFilterLoop,
+			pdCli, grpcPool, regionCache, pdClock,
+			txnutil.NewLockerResolver(kvStorage.(tikv.Storage), changefeed),
+		)
+		consume := func(ctx context.Context, raw *model.RawKVEntry, _ []tablepb.Span) error {
```
CharlesCheung96, line 595: It is recommended to retain the previous output channel, which can reduce the coupling degree of different modules.

cdc/puller/multiplexing_puller.go
```diff
@@ -0,0 +1,435 @@
...
+			return ctx.Err()
+		case progress = <-p.advanceCh:
+			if err := handleProgress(ctx, progress); err != nil {
+				return errors.Trace(err)
+			}
+		}
+	}
+}
+
+func (p *MultiplexingPuller) checkResolveLock(ctx context.Context) error {
```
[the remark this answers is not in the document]
hicqu: So we will need another track for slowest resolved timestamp of requestedTable. It is redundant with `Frontier`.

cdc/kv/shared_client.go
hicqu: It's better to call `AllocSubscriptionID` before `Subscribe` so that `MultiplexingPuller` can install some handlers for the allocatedID before `Subscribe`.

cdc/puller/ddl_puller.go
hicqu: It's possible that the output is not a channel. In SourceManager, `consume` will write events into `SortEngine` directly.

cdc/puller/multiplexing_puller.go
hicqu: Good idea.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1233 @@
...
+
+	workers []*sharedRegionWorker
+
+	// only modified in requestRegionToStore so lock is unnecessary.
+	requestedStores map[string]*requestedStore
+
+	totalSpans struct {
+		sync.RWMutex
+		m *spanz.HashMap[SubscriptionID]
+		v map[SubscriptionID]*requestedTable
```
CharlesCheung96, line 96: Fields in the struct are confused, so it is preferred to use meaningful names.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1233 @@
...
+
+func (s *SharedClient) setTableStopped(rt *requestedTable) {
+	log.Info("event feed starts to stop table",
+		zap.String("namespace", s.changefeed.Namespace),
+		zap.String("changefeed", s.changefeed.ID),
+		zap.Any("subscriptionID", rt.subscriptionID),
+		zap.String("span", rt.span.String()))
+
+	// Set stopped to true so we can stop handling region events from the table.
+	// Then send a spetial singleRegionInfo to regionRouter to deregister the table
```
CharlesCheung96, line 347:
```suggestion
	// Then send a special singleRegionInfo to regionRouter to deregister the table
```

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1233 @@
...
+	}
+
+	s.totalSpans.Lock()
+	if subID, success = s.totalSpans.m.Get(span); !success {
+		s.totalSpans.Unlock()
+		return
+	}
+
+	s.totalSpans.m.Delete(span)
+	rt := s.totalSpans.v[subID]
```
CharlesCheung96, line 246: Please add a comment explaining why m and v are handled differently.

cdc/kv/regionlock/region.go
```diff
@@ -44,3 +44,27 @@ func CheckRegionsLeftCover(regions []*metapb.Region, span tablepb.Span) bool {
 	}
 	return true
 }
+
+// CutRegionsLeftCoverSpan cuts regions at the position which doesn't cover span
+// or is incontinuous with the previous one.
+func CutRegionsLeftCoverSpan(regions []*metapb.Region, span tablepb.Span) []*metapb.Region {
```
CharlesCheung96, line 50: Please add test function to CutRegionsLeftCoverSpan.

cdc/kv/regionlock/region.go
```diff
@@ -44,3 +44,27 @@ func CheckRegionsLeftCover(regions []*metapb.Region, span tablepb.Span) bool {
 	}
 	return true
```
CharlesCheung96, line 45:
```suggestion
	subRegions := CutRegionsLeftCoverSpan(regions, span)
	return len(subRegions) == len(regions)
```
It is better to reuse code and tests.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1233 @@
...
+		sync.RWMutex
+		// map[subscriptionID]map[regionID]*regionFeedState
+		m map[SubscriptionID]map[uint64]*regionFeedState
+	}
+}
+
+type requestedTable struct {
+	subscriptionID SubscriptionID
+
+	span      tablepb.Span
```
CharlesCheung96, line 141:
```suggestion
	totalSpan      tablepb.Span
```

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1233 @@
...
+
+	s.totalSpans.Lock()
+	defer s.totalSpans.Unlock()
+	delete(s.totalSpans.v, rt.subscriptionID)
+}
+
+func (s *SharedClient) onRegionFail(ctx context.Context, errInfo regionErrorInfo) {
+	select {
+	case s.errCh.In() <- errInfo:
+	case <-ctx.Done():
```
CharlesCheung96, line 372:
```suggestion
```
ctx detection is no longer necessary when sending data to an unbounded channel.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1233 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case task := <-s.requestRangeCh.Out():
+			g.Go(func() error { return s.divideAndRequestRegions(ctx, task.span, task.requestedTable) })
+		}
+	}
+}
+
+func (s *SharedClient) divideAndRequestRegions(
```
CharlesCheung96, line 857:
```suggestion
func (s *SharedClient) divideAndScheduleRegions(
```

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1233 @@
...
+
+	rangeLock := sri.requestedTable.rangeLock
+	res := rangeLock.LockRange(ctx, sri.span.StartKey, sri.span.EndKey, sri.verID.GetID(), sri.verID.GetVer())
+	if res.Status == regionlock.LockRangeStatusWait {
+		res = res.WaitFn()
+	}
+	handleResult(res)
+}
+
+func (s *SharedClient) scheduleDivideRegionAndRequest(
```
CharlesCheung96, line 960:
```suggestion
func (s *SharedClient) scheduleRangeRequest(
```

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+func (s *SharedClient) clearStream(r *requestedStore, stream *requestedStream) {
+	if stream.conn != nil {
+		s.grpcPool.ReleaseConn(stream.conn, r.storeAddr)
+		stream.conn = nil
+	}
+}
+
+func (s *SharedClient) runStream(
```
CharlesCheung96, line 733: It is recommended to move stream-related logic to a separate file.

cdc/puller/ddl_puller.go
```diff
@@ -561,6 +582,28 @@ func NewDDLJobPuller(
...
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case rawDDLCh <- raw:
+				return nil
+			}
+		}
+		slots, hasher := 1, func(tablepb.Span, int) int { return 0 }
```
3AceShowHand, line 603: please add some comment to explain why the `slots` is set to 1, and hasher always return the 0.

cdc/puller/ddl_puller.go
```diff
@@ -69,6 +71,10 @@ type ddlJobPullerImpl struct {
 	puller       struct {
```
3AceShowHand, line 71: Why the puller type is wrapped as a struct ?

cdc/kv/region_state.go
```diff
@@ -76,12 +91,33 @@ func (s *regionFeedState) start() {
...
+	if s.state.CompareAndSwap(stateNormal, stateStopped) {
+		if err != nil {
+			s.err.Lock()
+			defer s.err.Unlock()
+			s.err.e = err
+		}
+	}
+}
+
+// masrk regionFeedState as removed if possible.
```
3AceShowHand, line 105:
```suggestion
// mark regionFeedState as removed if possible.
```

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1233 @@
...
+
+func (s *SharedClient) requestRegionToStore(ctx context.Context, g *errgroup.Group) error {
+	for {
+		var sri singleRegionInfo
+		select {
+		case <-ctx.Done():
+			return errors.Trace(ctx.Err())
+		case sri = <-s.regionRouter.Out():
+		}
+		// If lockedRange is nil it means it's a spetial task from stopping the table.
```
3AceShowHand, line 415:
```suggestion
		// If lockedRange is nil it means it's a special task from stopping the table.
```

3AceShowHand (approved):

cdc/kv/shared_client.go
hicqu: `m` is removed now.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1233 @@
...
+			return ctx.Err()
+		case task := <-s.requestRangeCh.Out():
+			g.Go(func() error { return s.divideAndRequestRegions(ctx, task.span, task.requestedTable) })
+		}
+	}
+}
+
+func (s *SharedClient) divideAndRequestRegions(
+	ctx context.Context,
+	span tablepb.Span,
```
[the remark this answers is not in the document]
hicqu: Span will always covered by requestedTable.span, but they can be not equal.

cdc/kv/shared_client.go
hicqu: I prefer `span` because `totalSpan` seems like a collection of table spans.

cdc/kv/shared_client.go
hicqu: I prefer to keep it as a method of `SharedClient`.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1196 @@
...
+}
+
+func (s *SharedClient) newStream(ctx context.Context, g *errgroup.Group, r *requestedStore) *requestedStream {
+	stream := &requestedStream{
+		streamID: streamIDGen.Add(1),
+		requests: chann.NewAutoDrainChann[singleRegionInfo](),
+	}
+
+	stream.requestedRegions.m = make(map[SubscriptionID]map[uint64]*regionFeedState)
+	stream.init = func(ctx context.Context) (err error) {
```
CharlesCheung96, line 637: init callback and preFetchForConnecting are unnecessary since stream could be initialized before runStream([L662](https://github.com/pingcap/tiflow/pull/9163/files#diff-76a50fd4045a2fd26a077fde7fa6f1e77ed9cce5871eb09a0588c74434a44c51R662)).

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1196 @@
...
+		defer waitReqTicker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return sri, ctx.Err()
+			case sri = <-stream.requests.Out():
+				return sri, nil
+			case <-waitReqTicker.C:
+				if stream.countStates() == 0 {
+					return sri, errors.New("closed as idle")
```
CharlesCheung96, line 488: Please add some comments here to explain that the stream will be reinitialized when the next request is received.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1196 @@
...
+			return err
+		}
+	}
+}
+
+func (s *SharedClient) receiveFromStream(ctx context.Context, rs *requestedStore, stream *requestedStream) (err error) {
+	for {
+		cevent, err := stream.client.Recv()
+		if err != nil {
+			code := grpc.Code(err)
```
CharlesCheung96, line 551: image
grpc.Code is deprecated: use status.Code instead.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1196 @@
...
+			if code != grpcerror.OK && code != grpcerror.Canceled {
+				log.Warn("event feed receive from grpc stream failed",
+					zap.String("namespace", s.changefeed.Namespace),
+					zap.String("changefeed", s.changefeed.ID),
+					zap.Uint64("storeID", rs.storeID),
+					zap.String("addr", rs.storeAddr),
+					zap.Error(err))
+				return errors.Trace(err)
+			}
+			return context.Canceled
```
CharlesCheung96, line 561: Please add some comments here to explain why the external stream needs to exit rather than restart in this case.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1196 @@
...
+		}
+	}
+}
+
+func (s *SharedClient) receiveFromStream(ctx context.Context, rs *requestedStore, stream *requestedStream) (err error) {
+	for {
+		cevent, err := stream.client.Recv()
+		if err != nil {
+			code := grpc.Code(err)
+			if code != grpcerror.OK && code != grpcerror.Canceled {
```
CharlesCheung96, line 552:
```suggestion
			if code = grpcerror.OK || code == grpcerror.Canceled {
				return context.Canceled
			}
			log.Warn("event feed receive from grpc stream failed",
					zap.String("namespace", s.changefeed.Namespace),
					zap.String("changefeed", s.changefeed.ID),
					zap.Uint64("storeID", rs.storeID),
					zap.String("addr", rs.storeAddr),
					zap.Error(err))
			return errors.Trace(err)
```

cdc/kv/shared_client.go
hicqu: `init` can be removed but seems `preFetchForConnecting` is still necessary.

cdc/kv/shared_client.go
```diff
@@ -0,0 +1,1194 @@
...
+	// `RequestId` in region requests of the table.
+	subscriptionIDGen atomic.Uint64
+
+	// To generate a streamID in `newStream`.
+	streamIDGen atomic.Uint64
+)
+
+const (
+	rpcMetaFeaturesKey               string = "features"
+	rpcMetaFeatureStreamMultiplexing string = "stream-multiplexing"
```
CharlesCheung96, line 1181: Please add comments to explain what this metadata does.

asddongmen (approved):

CharlesCheung96 (approved):

PR #9164, opened 2023-06-09
codec(ticdc): open-protocol support encode only handle key column if the message size large than the max message bytes

### LINKED ISSUE #9165: add changefeed level configuration to support only encode handle key columns

### LINKED ISSUE #9188: open-protocol encoder support only encode handle key columns

3AceShowHand:

Issue Number: close #9165, close #9188

#### Tests 

 - Unit test

### Release note 

```release-note
`None`
```

### DIFF
--- a/pkg/sink/codec/internal/message_key.go
+++ b/pkg/sink/codec/internal/message_key.go
@@ -22,13 +22,14 @@ import (
 
 // MessageKey defines the key for a message.
 type MessageKey struct {
-	Ts                               uint64            `json:"ts"`
-	Schema                           string            `json:"scm,omitempty"`
-	Table                            string            `json:"tbl,omitempty"`
-	RowID                            int64             `json:"rid,omitempty"`
-	Partition                        *int64            `json:"ptn,omitempty"`
-	Type                             model.MessageType `json:"t"`
-	LargeMessageOnlyHandleKeyColumns bool              `json:"loh,omitempty"`
+	Ts        uint64            `json:"ts"`
+	Schema    string            `json:"scm,omitempty"`
+	Table     string            `json:"tbl,omitempty"`
+	RowID     int64             `json:"rid,omitempty"`
+	Partition *int64            `json:"ptn,omitempty"`
+	Type      model.MessageType `json:"t"`
+	// Only Handle Key Columns encoded in the message's value part.
+	OnlyHandleKey bool `json:"ohk,omitempty"`
 }
 
 // Encode encodes the message key to a byte slice.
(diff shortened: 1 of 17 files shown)


pkg/sink/codec/open/open_protocol_encoder.go
```diff
@@ -37,14 +37,47 @@ type BatchEncoder struct {
...
+	}
+
+	length := len(key) + len(value) + common.MaxRecordOverhead + 16 + 8
+	if length > d.config.MaxMessageBytes {
+		log.Warn("Single message is too large for open-protocol",
+			zap.Int("maxMessageBytes", d.config.MaxMessageBytes),
+			zap.Int("length", length),
+			zap.Any("table", e.Table),
+			zap.Any("key", key))
+		return nil, nil, cerror.ErrMessageTooLarge.GenWithStackByArgs()
```
nongfushanquan, line 61: Why are two nils being used here?

3AceShowHand: since return error here, 2 nils means that key and value are nil. It's ok to return nils here.

pkg/sink/codec/internal/message_key.go
```diff
@@ -28,6 +28,8 @@ type MessageKey struct {
 	RowID     int64             `json:"rid,omitempty"`
 	Partition *int64            `json:"ptn,omitempty"`
 	Type      model.MessageType `json:"t"`
+	// Only Handle Key Columns encoded in the message's value part.
+	OnlyHandleKey bool `json:"ohk,omitempty"`
```
sdojjy, line 32: `ohk` is a not good name,  why not use full name?

sdojjy (approved):

pkg/sink/codec/internal/message_key.go
3AceShowHand: follow the open-protocol naming convention.

pkg/sink/codec/open/open_protocol_encoder.go
```diff
@@ -37,14 +37,47 @@ type BatchEncoder struct {
...
+func (d *BatchEncoder) buildMessageOnlyHandleKeyColumns(e *model.RowChangedEvent) ([]byte, []byte, error) {
+	keyMsg, valueMsg, err := rowChangeToMsg(e, d.config.DeleteOnlyHandleKeyColumns, true)
+	if err != nil {
+		return nil, nil, errors.Trace(err)
+	}
+	key, err := keyMsg.Encode()
+	if err != nil {
+		return nil, nil, errors.Trace(err)
+	}
+	value, err := valueMsg.encode(d.config.OnlyOutputUpdatedColumns)
```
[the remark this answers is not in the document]
3AceShowHand: this method encodes the `valueMsg` into JSON format, so we have to call it.

PR #9224, opened 2023-06-14
config(ticdc): deprecate  changefeed level config enable-old-value

### LINKED ISSUE #9667: deprecate the changefeed level `enable-old-value` configuration
`enable-old-value` controls whether to output previous column values for the update events, also adjust the output content of the delete event previously.

Mostly, It's set to true. There is no need to make it configurable, so remove it.

3AceShowHand:

Issue Number: close #9667 

* remove the `enable-old-value` from the changefeed configuration.
* remove all reference about this variable

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
deprecate the changefeed level `enable-old-value` configuration, always is true
```


asddongmen (approved):

asddongmen:
Finally.

sdojjy (approved):

PR #9228, opened 2023-06-14
engine: support new task mode for dm

lichunzhu:

Issue Number: close https://github.com/pingcap/tiflow/issues/9230

support new task mode for dm to get prepare for cloud import speeding up:
* dump: contains only dump progress
* load&sync: contains both load and sync progress

#### Tests 

 - Unit test
 - Integration test
  test using engine to deploy new task mode task. it works and passes sync-diff check.

### Release note 

```release-note
None
```


dm/checker/checker.go
lance6716, line 1: maybe we can define hasDump/hasLoad/hasSync for cfg.Mode

dm/config/subtask.go
```diff
@@ -348,7 +350,11 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 			return terror.ErrConfigLoaderDirInvalid.Delegate(err, c.LoaderConfig.Dir)
 		}
 		c.LoaderConfig.Dir = newDir
+	}
 
+	// adjust sorting dir
+	if c.Mode != ModeIncrement {
```
lance6716, line 356: also skip for ModeDump?

dm/config/task.go
```diff
@@ -775,11 +777,11 @@ func (c *TaskConfig) adjust() error {
 		instanceIDs[inst.SourceID] = i
 
 		switch c.TaskMode {
-		case ModeFull, ModeAll:
+		case ModeFull, ModeAll, ModeDump:
 			if inst.Meta != nil {
 				log.L().Warn("metadata will not be used. for Full mode, incremental sync will never occur; for All mode, the meta dumped by MyDumper will be used", zap.Int("mysql instance", i), zap.String("task mode", c.TaskMode))
 			}
-		case ModeIncrement:
+		case ModeIncrement, ModeLoadSync:
```
lance6716, line 784: do we require the data source files of load unit having a metadata file? or what if the metadata file and task metadata both have binlog starting position?

lichunzhu: We require the data source files of load unit having a metadata file. If the metadata file and task metadata both have binlog starting position we should use the data in metadata file.
BTW, these two task modes will only be used in cloud in our design. They are not planned to be exposed to other users.

engine/test/integration_tests/dm_dump_sync_mode/run.sh
```diff
@@ -0,0 +1,28 @@
...
+	wait_mysql_online.sh --host 127.0.0.1 --port 3306
+	wait_mysql_online.sh --host 127.0.0.1 --port 4000
+
+	echo -e "\nstart to run dump/load&sync mode integration tests...\n"
+
+	export ENGINE_S3_ACCESS_KEY=engine
+	export ENGINE_S3_SECRET_KEY=engineSecret
+	export ENGINE_S3_ENDPOINT="http://minio-standalone:9000/"
+
+	CGO_ENABLED=0 go test -count=1 -v -run ^TestDMDumpSyncJob$ github.com/pingcap/tiflow/engine/test/e2e
```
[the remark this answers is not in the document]
lichunzhu: Don't understand what you mean 🤔 This is integration test written in go framework.

dm/checker/checker.go
lichunzhu: Do you mean using `hasDump/hasLoad/hasSync` to replace `cfg.Mode`?

lance6716: helper functions

engine/test/integration_tests/dm_dump_sync_mode/run.sh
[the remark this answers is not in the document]
lance6716: It's OK to write in go. I see the framework is called by HTTP API so language is not important.

dm/config/subtask.go
lichunzhu: addressed in https://github.com/pingcap/tiflow/pull/9228/commits/f18d7217d2a76a9ed00a6a8f3cfb4a9aedd97484

engine/jobmaster/dm/task_manager.go
```diff
@@ -282,7 +282,13 @@ func (tm *TaskManager) allFinished(ctx context.Context) bool {
...
 		}
 		// update if we add new task mode
-		if runningTask.Unit != frameModel.WorkerDMLoad || task.Cfg.TaskMode != dmconfig.ModeFull {
+		if runningTask.Unit == frameModel.WorkerDMDump && task.Cfg.TaskMode != dmconfig.ModeDump {
+			return false
+		}
+		if runningTask.Unit == frameModel.WorkerDMLoad && task.Cfg.TaskMode != dmconfig.ModeFull {
+			return false
+		}
+		if runningTask.Unit == frameModel.WorkerDMSync {
```
D3Hunter, line 291: wound be easier to understand if we switch on mode first, like this
```go
switch mode {
case full:
   if current-unit != load {
        return false
    }
case dump:
    return true // current unit must be dump
default:
    return false // all/incr/load&sync never finishes
}
```

engine/test/e2e/e2e_dm_test.go
```diff
@@ -123,6 +127,82 @@ func TestDMJob(t *testing.T) {
...
+			for _, m := range matched {
+				jobIDs[string(m[1])] = struct{}{}
+			}
+			require.NoError(t, resp.Body.Close())
+		}
+		require.Equal(t, 2, len(jobIDs))
+	}
+
+	// test metrics for syncer
+	re := regexp.MustCompile(`syncer.*\{job_id="(.{36})"`)
```
D3Hunter, line 199: our alert on cloud requires all underlying jobs of a single DM job(from dataflow-service view) has the same job-id(dataflow-service side), if we give each underlying job a different id, even with same prefix, like `123-dump` `123-load&sync`, we need to change it's label, like in here to make current alert works. https://stackoverflow.com/questions/70177067/prometheus-substring-or-split-functionality

D3Hunter (commented):
rest lgtm

engine/test/e2e/e2e_dm_test.go
lichunzhu: What about using the same job_id for dump and load&sync? Two tasks shouldn't run at the same time

D3Hunter: fow now, the job is not delete physically for meter purpose, there's still meta left on tiflow-metadb, so same id cannot be used twice

D3Hunter (approved):

dm/checker/checker.go
```diff
@@ -393,7 +389,7 @@ func (c *Checker) Init(ctx context.Context) (err error) {
 	// Because the table schema obtained from `show create table` is not the schema at the point of binlog.
 	_, checkingShardID := c.checkingItems[config.ShardAutoIncrementIDChecking]
 	_, checkingShard := c.checkingItems[config.ShardTableSchemaChecking]
-	if checkingShard && instance.cfg.ShardMode != "" && instance.cfg.Mode != config.ModeIncrement {
+	if checkingShard && instance.cfg.ShardMode != "" && config.HasLoad(instance.cfg.Mode) {
```
lance6716, line 392: I'm not sure, just remember below checkers checks that current upstream table structure can be handled by sharding strategy. So here we should check HasDump & HasLoad?

dm/config/task.go
```diff
@@ -775,7 +777,7 @@ func (c *TaskConfig) adjust() error {
 		instanceIDs[inst.SourceID] = i
 
 		switch c.TaskMode {
-		case ModeFull, ModeAll:
+		case ModeFull, ModeAll, ModeDump:
 			if inst.Meta != nil {
 				log.L().Warn("metadata will not be used. for Full mode, incremental sync will never occur; for All mode, the meta dumped by MyDumper will be used", zap.Int("mysql instance", i), zap.String("task mode", c.TaskMode))
 			}
```
lance6716, line 783: should we add ModeLoadSync? seems ModeLoadSync will also use the metadata file in source directory.

engine/jobmaster/dm/task_manager.go
```diff
@@ -282,7 +282,14 @@ func (tm *TaskManager) allFinished(ctx context.Context) bool {
 			return false
 		}
 		// update if we add new task mode
-		if runningTask.Unit != frameModel.WorkerDMLoad || task.Cfg.TaskMode != dmconfig.ModeFull {
+		switch task.Cfg.TaskMode {
+		case dmconfig.ModeFull:
+			if runningTask.Unit != frameModel.WorkerDMLoad {
+				return false
+			}
+		case dmconfig.ModeDump:
```
lance6716, line 290: in old behaviour, we don't return true before finish checking all `job.Tasks`, so I'm not sure we should directly `return true` here

lance6716 (commented):
rest lgtm

dm/config/task.go
lichunzhu: ModeLoadSync can use metadata from syncers and it can use metadata file. So `inst.Meta` is empty or not is both okay?

lance6716: it's about how the upper layer uses engine, and I'm not sure about the behaviour of upper layer. lgtm now

PR #9252, opened 2023-06-16
ci: add component management to `.codecov.yaml` and migrate flag manage

wuhuizuo:
Codecov's `Flags` should be better be used to split test types and the `components` feature better for components management.

The comonent management is a new feature in codecov, currently we used it in codecov page of `pingcap/tidb`. 

Ref:
- https://docs.codecov.com/docs/flags
- https://docs.codecov.com/docs/components
- PingCAP-QE/ci#2171

- migrate flag mangement to Automatic Flag Management: only only with the components flags, also we need to support unit and integration and other test type's covrage data.
- add component management configuration, it no need to set when upload coverage, it will auto display in codecov.io page.

#### Tests 

 - No code

### Release note 

```release-note
None
```

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -50,9 +50,9 @@ component_management:
     - component_id: component_cdc # this is an identifier that should not be changed
       name: cdc # this is a display name, and can be changed freely
       paths:
-        - "cdc/**"
-        - "cmd/**"
-        - "pkg/**"
+        - cdc/**
+        - cmd/**
+        - pkg/**
     - component_id: component_dm
       name: dm
       paths:
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -24,7 +24,7 @@ parsers:
       macro: no
 
 comment:
-  layout: "header, flags, diff"
+  layout: "header, components, flags, diff"
   behavior: default
   require_changes: no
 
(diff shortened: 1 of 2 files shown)


.codecov.yml
```diff
@@ -41,20 +41,49 @@ ignore:
...
+    statuses:
+      - type: project # in this case every component that doens't have a status defined will have a project type one
+        target: auto
+  individual_components:
+    - component_id: component_cdc # this is an identifier that should not be changed
+      name: cdc # this is a display name, and can be changed freely
+      paths:
+        - "cdc/**"
+        - "cmd/**"
+        - "pkg/**"
```
wuhuizuo, line 55:
```suggestion
        - cdc/**
        - cmd/**
        - pkg/**
```

Makefile
```diff
@@ -417,7 +417,7 @@ dm_unit_test_in_verify_ci: check_failpoint_ctl tools/bin/gotestsum tools/bin/goc
 	|| { $(FAILPOINT_DISABLE); exit 1; }
 	tools/bin/gocov convert "$(DM_TEST_DIR)/cov.unit_test.out" | tools/bin/gocov-xml > dm-coverage.xml
 	$(FAILPOINT_DISABLE)
-	@bash <(curl -s https://codecov.io/bash) -F dm -f $(DM_TEST_DIR)/cov.unit_test.out -t $(TICDC_CODECOV_TOKEN)
+	@bash <(curl -s https://codecov.io/bash) -F unit,dm -f $(DM_TEST_DIR)/cov.unit_test.out -t $(TICDC_CODECOV_TOKEN)
```
lance6716, line 420: what's the meaning of "unit" 🤔

lance6716 (approved):

Makefile
wuhuizuo: Use flags management for test type dim.
- unit: means coverage from unit testing.
- integration: means coverage from integration testing.
- e2e: means coverage from end to end testing.
Ref: https://github.com/PingCAP-QE/ci/issues/2171

wuhuizuo: Just make the word short and simple. 😆

hi-rustin (approved):

PR #9262, opened 2023-06-19
mounter(ticdc): enum type default value set explicitly to avoid wrong enum value fetched by the encoder

### LINKED ISSUE #9259: type enum default value is ignored when mount the event
### What did you do?

```
create table t(a int primary key, state enum ("created", "working", "finished", "error") collate utf8mb4_general_ci not null default "created");
insert into t values (1, "working");
```

_No response_

### What did you see instead?

when drop the column `state` and insert new data at the delete-only phase, use old schema to parse the raw value, "created" should be used as the default value to fill the column.

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

3AceShowHand:

Issue Number: close #9259 

* mounter set the default value for the enum type explicitly, to make `enum.Value` set correctly.

#### Tests 

 - Unit test

### Release note 

```release-note
`None`
```

### DIFF
--- a/cdc/entry/mounter_test.go
+++ b/cdc/entry/mounter_test.go
@@ -999,7 +999,10 @@ func TestGetDefaultZeroValue(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		_, val, _, _, _ := getDefaultOrZeroValue(&tc.ColInfo)
+		datum, val, _, _, _ := getDefaultOrZeroValue(&tc.ColInfo)
+		if datum.Kind() == types.KindMysqlEnum {
+			require.Equal(t, uint64(1), datum.GetMysqlEnum().Value)
+		}
 		require.Equal(t, tc.Res, val, tc.Name)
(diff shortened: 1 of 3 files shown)


cdc/entry/mounter_test.go
```diff
@@ -957,7 +957,7 @@ func TestGetDefaultZeroValue(t *testing.T) {
 			ColInfo: timodel.ColumnInfo{FieldType: *ftTypeEnumNotNull},
 			// TypeEnum value will be a string and then translate to []byte
 			// NotNull && no default will choose first element
-			Res:     uint64(0),
+			Res:     uint64(1),
```
CharlesCheung96, line 960: please add more tests and comments

charleszheng44 (approved):
LGTM

CharlesCheung96 (approved):

asddongmen (approved):

Issue #9273, opened 2023-06-20
`cdc cli changefeed list` should return error message.


asddongmen:
### Is your feature request related to a problem?

When changefeed encounters certain errors internally, it will automatically retry.

Error retry may consume some time, causing the changefeed to stall. However, when users query the status of the changefeed, they will still see it in normal state and will not see any error messages.

This behavior can be confusing for users and does not allow them to promptly participate in error handling.

### Describe the feature you'd like

Users can view error information of changefeed through command line tools, and the error message should indicate the time when the error occurred and the time for retry.

asddongmen:
close by: https://github.com/pingcap/tiflow/pull/9626

PR #9358, opened 2023-07-07
changefeed (ticdc): remove error state and add pending, warning state

asddongmen:

Issue Number: ref #9273

After this PR, the error status of changefeed will be removed, and the warning status will be added.
The changefeed status transition is shown in the following figure.

image

1. The user pauses changefeed, and changefeed enters the `stopped` state.
2. The user resumes changefeed, and changefeed returns to the `normal` state.
3. Changefeed encounters an error and enters the `warning` state. Changefeed automatically retries in the `warning` state.
4. If the changefeed retry is successful and the `CheckpointTs` advances beyond the `CheckpointTs` when the error occurred, the changefeed enters the `normal` state.
5. If the changefeed error retry exceeds the maximum retry time limit, it enters the `failed` state. After entering the `failed` state, the changefeed will block the GC of the upstream TiDB cluster. The default blocking time is `gc-ttl`. Users can manually recover the changefeed within the `gc-ttl` time.
6. If the changefeed encounters an unrecoverable error during the running process, it directly enters the `failed` state.
7. The changefeed  is complete, and the its progress has reached the preset `TargetTs`.
8. If the changefeed is in the stopped state for a time exceeding the duration specified by `gc-ttl`, it cannot be recovered.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/api/v2/model.go
+++ b/cdc/api/v2/model.go
@@ -102,6 +102,27 @@ type ChangefeedCommonInfo struct {
 	RunningError   *model.RunningError `json:"error"`
 }
 
+// MarshalJSON marshal changefeed common info to json
+// we need to set feed state to normal if it is uninitialized and pending to warning
+// to hide the detail of uninitialized and pending state from user
+func (c ChangefeedCommonInfo) MarshalJSON() ([]byte, error) {
+	// alias the original type to prevent recursive call of MarshalJSON
+	type Alias ChangefeedCommonInfo
+
+	if c.FeedState == model.StateUnInitialized {
+		c.FeedState = model.StateNormal
+	}
+	if c.FeedState == model.StatePending {
+		c.FeedState = model.StateWarning
+	}
+
+	return json.Marshal(struct {
+		Alias
+	}{
+		Alias: Alias(c),
+	})
+}
+
 // ChangefeedConfig use by create changefeed api
 type ChangefeedConfig struct {
 	Namespace     string         `json:"namespace"`
--- a/cdc/model/changefeed.go
+++ b/cdc/model/changefeed.go
@@ -135,6 +135,11 @@ func (s FeedState) IsNeeded(need string) bool {
 	return need == string(s)
 }
 
+// IsRunning return true if the feedState represents a running state.
+func (s FeedState) IsRunning() bool {
+	return s == StateNormal || s == StateWarning
+}
+
 // ChangeFeedInfo describes the detail of a ChangeFeed
 type ChangeFeedInfo struct {
 	UpstreamID uint64    `json:"upstream-id"`
--- a/cdc/model/changefeed.go
+++ b/cdc/model/changefeed.go
@@ -83,7 +83,7 @@ type FeedState string
 // All FeedStates
 const (
 	StateNormal   FeedState = "normal"
-	StateError    FeedState = "error"
+	StateWarning  FeedState = "warning"
 	StateFailed   FeedState = "failed"
 	StateStopped  FeedState = "stopped"
 	StateRemoved  FeedState = "removed"
@@ -95,7 +95,7 @@ func (s FeedState) ToInt() int {
 	switch s {
 	case StateNormal:
 		return 0
-	case StateError:
+	case StateWarning:
 		return 1
 	case StateFailed:
 		return 2
@@ -418,7 +418,7 @@ func (info *ChangeFeedInfo) fixState() {
 				if cerror.IsChangefeedFastFailErrorCode(errors.RFCErrorCode(info.Error.Code)) {
 					state = StateFailed
 				} else {
-					state = StateError
+					state = StateWarning
 				}
 			}
 		case AdminStop:
--- a/cdc/model/changefeed.go
+++ b/cdc/model/changefeed.go
@@ -83,11 +83,12 @@ type FeedState string
 // All FeedStates
 const (
 	StateNormal   FeedState = "normal"
-	StateWarning  FeedState = "warning"
+	StatePending  FeedState = "pending"
 	StateFailed   FeedState = "failed"
 	StateStopped  FeedState = "stopped"
 	StateRemoved  FeedState = "removed"
 	StateFinished FeedState = "finished"
+	StateWarning  FeedState = "warning"
 )
 
 // ToInt return an int for each `FeedState`, only use this for metrics.
@@ -95,7 +96,7 @@ func (s FeedState) ToInt() int {
 	switch s {
 	case StateNormal:
 		return 0
-	case StateWarning:
+	case StatePending:
 		return 1
 	case StateFailed:
 		return 2
@@ -105,6 +106,8 @@ func (s FeedState) ToInt() int {
 		return 4
 	case StateRemoved:
 		return 5
+	case StateWarning:
+		return 6
 	}
 	// -1 for unknown feed state
 	return -1
@@ -122,6 +125,10 @@ func (s FeedState) IsNeeded(need string) bool {
 		case StateStopped:
 			return true
 		case StateFailed:
+			return true
+		case StateWarning:
+			return true
+		case StatePending:
 			return true
 		}
 	}
--- a/cdc/model/changefeed.go
+++ b/cdc/model/changefeed.go
@@ -81,14 +81,18 @@ const (
 type FeedState string
 
 // All FeedStates
+// Only `StateNormal` and `StatePending` changefeed is running,
+// others are stopped.
 const (
-	StateNormal        FeedState = "normal"
-	StatePending       FeedState = "pending"
-	StateFailed        FeedState = "failed"
-	StateStopped       FeedState = "stopped"
-	StateRemoved       FeedState = "removed"
-	StateFinished      FeedState = "finished"
-	StateWarning       FeedState = "warning"
+	StateNormal   FeedState = "normal"
+	StatePending  FeedState = "pending"
+	StateFailed   FeedState = "failed"
+	StateStopped  FeedState = "stopped"
+	StateRemoved  FeedState = "removed"
+	StateFinished FeedState = "finished"
+	StateWarning  FeedState = "warning"
+	// StateUnInitialized is used for the changefeed that has not been initialized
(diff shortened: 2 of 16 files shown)


cdc/model/changefeed.go
```diff
@@ -82,20 +82,22 @@ type FeedState string
...
-	StateStopped  FeedState = "stopped"
-	StateRemoved  FeedState = "removed"
-	StateFinished FeedState = "finished"
+	StateNormal        FeedState = "normal"
+	StatePending       FeedState = "pending"
+	StateFailed        FeedState = "failed"
+	StateStopped       FeedState = "stopped"
+	StateRemoved       FeedState = "removed"
+	StateFinished      FeedState = "finished"
+	StateWarning       FeedState = "warning"
```
nongfushanquan, line 91: Please add some comments about each state , such as whether it is visible  to the user or is running in the ticdc

cdc/owner/feed_state_manager.go
```diff
@@ -238,7 +269,8 @@ func (m *feedStateManager) handleAdminJob() (jobsPending bool) {
 			zap.Uint64("checkpointTs", checkpointTs))
 	case model.AdminResume:
 		switch m.state.Info.State {
-		case model.StateFailed, model.StateError, model.StateStopped, model.StateFinished:
```
nongfushanquan, line 241: if the state is  warning or pending ,whether  it can be resumed ?

asddongmen: yes,it can be resumed.

cdc/owner/feed_state_manager.go
```diff
@@ -110,40 +111,81 @@ func (m *feedStateManager) shiftStateWindow(state model.FeedState) {
...
+					m.errBackoff.MaxElapsedTime,
+				),
+			)
 			m.shouldBeRunning = false
 			m.patchState(model.StateFailed)
 			return
 		}
+
+		m.lastErrorRetryTime = time.Now()
+		if m.state.Status != nil {
```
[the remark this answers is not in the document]
asddongmen: `checkAndInitLastRetryCheckpointTs` only initializes it: from `0` to `checkpointTs` if the changefeed was just create in the owner. 
We still need to update it whenever retry on an error.

cdc/model/changefeed.go
asddongmen: ok

nongfushanquan (approved):

cdc/api/v2/model.go
```diff
@@ -102,6 +102,27 @@ type ChangefeedCommonInfo struct {
...
 }
 
+// MarshalJSON marshal changefeed common info to json
+// we need to set feed state to normal if it is uninitialized and pending to warning
+// to hide the detail of uninitialized and pending state from user
+func (c ChangefeedCommonInfo) MarshalJSON() ([]byte, error) {
+	// alias the original type to prevent recursive call of MarshalJSON
+	type Alias ChangefeedCommonInfo
+
+	if c.FeedState == model.StateUnInitialized {
```
CharlesCheung96, line 112: Is this compatible with the old logic?

cdc/owner/feed_state_manager.go
```diff
@@ -35,8 +35,8 @@ const (
 	//	 640s, 1280s, 1800s, ...).
```
CharlesCheung96, line 35: Please update this comment.

cdc/owner/feed_state_manager.go
```diff
@@ -77,16 +78,16 @@ func newFeedStateManager(up *upstream.Upstream) *feedStateManager {
 	// backoff will stop once the defaultBackoffMaxElapsedTime has elapsed.
 	f.errBackoff.MaxElapsedTime = defaultBackoffMaxElapsedTime
 
-	f.resetErrBackoff()
-	f.lastErrorTime = time.Unix(0, 0)
-
+	f.resetErrRetry()
+	f.lastErrorRetryTime = time.Unix(0, 0)
```
CharlesCheung96, line 82:
```suggestion
	f.lastErrorRetryTime = time.Unix(0, 0)
```
This is redundant since lastErrorRetryTime is initialized in `f.resetErrRetry()`.

cdc/owner/feed_state_manager.go
```diff
@@ -248,9 +279,9 @@ func (m *feedStateManager) handleAdminJob() (jobsPending bool) {
 		}
 		m.shouldBeRunning = true
 		// when the changefeed is manually resumed, we must reset the backoff
-		m.resetErrBackoff()
+		m.resetErrRetry()
 		// The lastErrorTime also needs to be cleared before a fresh run.
-		m.lastErrorTime = time.Unix(0, 0)
+		m.lastErrorRetryTime = time.Unix(0, 0)
```
CharlesCheung96, line 284:
```suggestion
		m.lastErrorRetryTime = time.Unix(0, 0)
```

cdc/owner/feed_state_manager_test.go
```diff
@@ -62,8 +62,8 @@ func newFeedStateManager4Test(
 	f.errBackoff.Multiplier = multiplier
 	f.errBackoff.RandomizationFactor = 0
 
-	f.resetErrBackoff()
-	f.lastErrorTime = time.Unix(0, 0)
+	f.resetErrRetry()
+	f.lastErrorRetryTime = time.Unix(0, 0)
```
CharlesCheung96, line 66:
```suggestion
	f.lastErrorRetryTime = time.Unix(0, 0)
```

cdc/owner/feed_state_manager.go
```diff
@@ -474,7 +509,7 @@ func (m *feedStateManager) handleError(errs ...*model.RunningError) {
 		}
 	}
 
-	//  changefeed state from stopped to failed is allowed
+	// changefeed state from stopped to failed is allowed
```
CharlesCheung96, line 512:
```suggestion
	// Changing changefeed state from stopped to failed is allowed
```

cdc/owner/feed_state_manager.go
```diff
@@ -484,96 +519,46 @@ func (m *feedStateManager) handleError(errs ...*model.RunningError) {
...
-				return info, true, nil
-			})
-			m.shouldBeRunning = false
-			m.patchState(model.StateError)
-			return
-		}
-	}
-
+	lastError := errs[len(errs)-1]
+	// patch the last error to changefeed info
```
CharlesCheung96, line 523:
```suggestion
	// if any error is occurred in this tick, we should patch the last error to changefeed info and
	// set the changefeed state to warning to stop the changefeed.
	if lastError != nil {
		log.Warn("changefeed meets an error", zap.Any("error", lastError))
		m.state.PatchInfo(func(info *model.ChangeFeedInfo) (*model.ChangeFeedInfo, bool, error) {
			if info == nil {
				return nil, false, nil
			}
			info.Error = lastError
			return info, true, nil
		})

		m.shouldBeRunning = false
		m.patchState(model.StatePending)
	}
```

cdc/owner/feed_state_manager.go
```diff
@@ -484,96 +519,46 @@ func (m *feedStateManager) handleError(errs ...*model.RunningError) {
...
 	m.state.PatchInfo(func(info *model.ChangeFeedInfo) (*model.ChangeFeedInfo, bool, error) {
 		if info == nil {
 			return nil, false, nil
 		}
-		for _, err := range errs {
-			info.Error = err
-		}
-		return info, len(errs) > 0, nil
+		info.Error = lastError
+		return info, lastError != nil, nil
```
CharlesCheung96, line 529: Patch a nil error is meaningless.

cdc/owner/feed_state_manager.go
```diff
@@ -484,96 +519,46 @@ func (m *feedStateManager) handleError(errs ...*model.RunningError) {
...
-				info.Error = err
-				return info, true, nil
-			})
-			m.shouldBeRunning = false
-			m.patchState(model.StateError)
-			return
-		}
-	}
-
+	lastError := errs[len(errs)-1]
```
CharlesCheung96, line 522: Perhaps we should iterate through errs in reverse order until a non-nil error is found?

cdc/owner/feed_state_manager.go
```diff
@@ -110,40 +111,81 @@ func (m *feedStateManager) shiftStateWindow(state model.FeedState) {
...
+
 	if m.handleAdminJob() {
 		// `handleAdminJob` returns true means that some admin jobs are pending
 		// skip to the next tick until all the admin jobs is handled
 		adminJobPending = true
 		return
 	}
+
 	switch m.state.Info.State {
+	case model.StateUnInitialized:
```
CharlesCheung96, line 133: Is it possible to enter this branch?

cdc/owner/feed_state_manager.go
```diff
@@ -484,96 +519,46 @@ func (m *feedStateManager) handleError(errs ...*model.RunningError) {
...
+		m.resetErrRetry()
 	}
 }
 
 func (m *feedStateManager) handleWarning(errs ...*model.RunningError) {
+	if len(errs) == 0 {
+		return
+	}
+	lastError := errs[len(errs)-1]
+	m.patchState(model.StateWarning)
```
CharlesCheung96, line 555: Perhaps lastRetryCheckpointTs should be updated before pathch warning state?

cdc/owner/feed_state_manager.go
asddongmen: It would never be `nil` here, `errs` is checked at the first line of this func.

cdc/owner/feed_state_manager.go
asddongmen: Yes, in the first tick.

cdc/owner/feed_state_manager.go
asddongmen: Changefeed will not retry on a warning, warning will be retried on sink level.

asddongmen: Changefeed will not retry on a warning, warning will be retried on sink level.

CharlesCheung96 (approved):

PR #9435, opened 2023-07-26
consumer(ticdc): kafka consumer refactor

### LINKED ISSUE #9452: refactor the kafka consumer
as title

3AceShowHand:

Issue Number: close #9452 

#### Tests 

 - Integration test

### Release note 

```release-note
`None`
```


asddongmen (approved):

cmd/kafka-consumer/main.go
```diff
@@ -57,17 +57,25 @@ import (
...
+		maxBatchSize:    math.MaxInt64,
+	}
+}
+
+type consumerOption struct {
+	address      []string
+	version      string
+	topic        string
+	partitionNum int32
+	groupID      string
```
CharlesCheung96, line 75: Is `groupID ` uniquely identifying a consumer group?

CharlesCheung96 (approved):

cmd/kafka-consumer/main.go
3AceShowHand: yes, it's used to identify the consumer group, but it's generated by the uuid, which means if the consumer restart, a new consumer group is added to consume message from the beginning.

PR #9437, opened 2023-07-27
sink(ticdc):  split RowChangeEvent if unique key is updated

### LINKED ISSUE #9430: Disordering events in a transaction with TiCDC may cause conflicts in downstream execution
### What did you do?

create table cptesttable (k INT NOT NULL PRIMARY KEY, v INT NOT NULL, UNIQUE(v));
MySQL [cptestdb]> BEGIN;
MySQL [cptestdb]> INSERT INTO cptesttable VALUES (1, 1);
MySQL [cptestdb]> INSERT INTO cptesttable VALUES (2, 2);
MySQL [cptestdb]> COMMIT;
MySQL [cptestdb]> BEGIN;
MySQL [cptestdb]> UPDATE cptesttable SET v = 3 WHERE k = 2;
MySQL [cptestdb]> UPDATE cptesttable SET v = 2 WHERE k = 1;
MySQL [cptestdb]> COMMIT;

MySQL [cptestdb]> UPDATE cptesttable SET v = 3 WHERE k = 2;
MySQL [cptestdb]> UPDATE cptesttable SET v = 2 WHERE k = 1;

### What did you see instead?

MySQL [cptestdb]> UPDATE cptesttable SET v = 2 WHERE k = 1;
MySQL [cptestdb]> UPDATE cptesttable SET v = 3 WHERE k = 2;

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

6.5 & 7.1
```console
(paste TiCDC version here)
```

### LINKED ISSUE #9436: split RowChangeEvent if unique key is updated
split RowChangeEvent if unique key is updated

sdojjy:

Issue Number: close #9430

split and sort RowChangeEvent if unique key is updated
split the update to delete and insert, then sort the whole txn with the order delete>update>insert

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


cdc/model/sink.go
```diff
@@ -719,6 +748,115 @@ func (t *SingleTableTxn) GetCommitTs() uint64 {
...
+	if err != nil {
+		return errors.Trace(err)
+	}
+	sort.Sort(RowChangedEvents(newRows))
+	t.Rows = newRows
+	return nil
+}
+
+// convertRowChangedEvents uses to convert RowChangedEvents to TableSinkRowChangedEvents.
+// It will deal with the old value compatibility.
```
asddongmen, line 766: plz remove this line

cdc/model/sink.go
```diff
@@ -332,11 +339,33 @@ type RowChangedEvent struct {
...
+// RowChangedEvents represents a set of events
+type RowChangedEvents []*RowChangedEvent
+
+// Len is the number of elements in the collection.
+func (e RowChangedEvents) Len() int {
+	return len(e)
+}
+
+func (e RowChangedEvents) Less(i, j int) bool {
+	return len(e[i].Columns)-len(e[i].PreColumns) <
```
3AceShowHand, line 351: This is not easy to understand directly, please add some comments about how the condition comes up.

cdc/model/sink.go
```diff
@@ -719,6 +748,115 @@ func (t *SingleTableTxn) GetCommitTs() uint64 {
 	return t.CommitTs
 }
 
+// TrySplitAndSortUpdateEvent split update events if unique key is updated
+func (t *SingleTableTxn) TrySplitAndSortUpdateEvent() error {
+	if len(t.Rows) < 2 {
```
3AceShowHand, line 753: is it possible there is only an update event, shall we split it here ?

cdc/model/sink.go
```diff
@@ -719,6 +748,115 @@ func (t *SingleTableTxn) GetCommitTs() uint64 {
...
+	newRows, err := convertRowChangedEvents(t.Rows)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	sort.Sort(RowChangedEvents(newRows))
+	t.Rows = newRows
+	return nil
+}
+
+// convertRowChangedEvents uses to convert RowChangedEvents to TableSinkRowChangedEvents.
```
3AceShowHand, line 765: `TableSinkRowChangedEvents` sounds like a new term but not exist in the code base, how about remove it ?

cdc/model/sink.go
```diff
@@ -719,6 +748,115 @@ func (t *SingleTableTxn) GetCommitTs() uint64 {
...
+		// begin; insert into t (id) values (1); delete from t where id=1; commit;
+		// Just ignore these row changed events.
+		if colLen == 0 && preColLen == 0 {
+			log.Warn("skip emit empty row event",
+				zap.Any("event", e))
+			continue
+		}
+
+		// This indicates that it is an update event,
+		// and after enable old value internally by default(but disable in the configuration).
```
3AceShowHand, line 790: do you mean that `enable-old-value` is always false in all conditions ?

cdc/model/sink.go
sdojjy: no, it's not needed to split a single update event.

cdc/model/sink.go
sdojjy: yes,  we should always split the update events if unique key is updated

cdc/model/sink.go
sdojjy: done

cdc/model/sink.go
sdojjy: changed to another implementation

asddongmen (approved):

cdc/model/sink.go
3AceShowHand: What happen if this update event has unique key columns changed ?

3AceShowHand: This method is almost identical to the `convertRowChangedEvents` method. 

After the enable-old-value is removed, this 2 method can be merged into one.

cdc/model/sink.go
```diff
@@ -719,6 +765,112 @@ func (t *SingleTableTxn) GetCommitTs() uint64 {
...
+	return nil
+}
+
+// trySplitAndSortUpdateEvent try to split update events if unique key is updated
+// returns true if some updated events is split
+func trySplitAndSortUpdateEvent(
+	events []*RowChangedEvent,
+) ([]*RowChangedEvent, bool, error) {
+	rowChangedEvents := make([]*RowChangedEvent, 0, len(events))
+	needSplit := false
```
3AceShowHand, line 791: `needSplit` is not a good variable name, since event is already splitter by this method.

cdc/model/sink.go
```diff
@@ -719,6 +765,112 @@ func (t *SingleTableTxn) GetCommitTs() uint64 {
...
+func (t *SingleTableTxn) TrySplitAndSortUpdateEvent() error {
+	if len(t.Rows) < 2 {
+		return nil
+	}
+	newRows, split, err := trySplitAndSortUpdateEvent(t.Rows)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	// some updated events is split, need to sort
+	if split {
```
3AceShowHand, line 778: This logic can looks can be put inside the `trySplitAndSortUpdateEvent`, and `split` no need to be returned

cdc/model/sink.go
```diff
@@ -332,11 +346,43 @@ type RowChangedEvent struct {
 	ReplicatingTs Ts `json:"-" msg:"-"`
 }
 
+// txnRows represents a set of events that belong to the same transaction.
+type txnRows []*RowChangedEvent
```
3AceShowHand, line 350: How about renaming it to `rowArray` ? There is similar type declaration in the Sarama named as `recordArray` to represent all message records in the same batch.

At the moment, all rows come from the same transaction, but I thinks it's no need to mention this in the variable name.

cdc/model/sink_test.go
```diff
@@ -396,3 +398,168 @@ func TestExchangeTablePartition(t *testing.T) {
...
+			Columns: []*Column{{}},
+		},
+		{
+			PreColumns: []*Column{{}},
+		},
+	}
+	assert.True(t, events[0].IsUpdate())
+	assert.True(t, events[1].IsInsert())
+	assert.True(t, events[2].IsDelete())
+	sort.Sort(txnRows(events))
```
3AceShowHand, line 418: I think the sort functionality is covered by other case, no need to have this case.

If put the sort into the `trySplitAndSortUpdateEvent`, only test this method with multiple difference cases should be ok.

cdc/model/sink.go
```diff
@@ -719,6 +765,112 @@ func (t *SingleTableTxn) GetCommitTs() uint64 {
...
+func shouldSplitUpdateEvent(updateEvent *RowChangedEvent) bool {
+	// nil event will never be split.
+	if updateEvent == nil {
+		return false
+	}
+
+	for i := range updateEvent.Columns {
+		col := updateEvent.Columns[i]
+		preCol := updateEvent.PreColumns[i]
+		if col != nil && (col.Flag.IsUniqueKey() || col.Flag.IsHandleKey()) &&
```
nongfushanquan, line 838: what does this condition " col.Flag.IsHandleKey()) && preCol != nil && (preCol.Flag.IsUniqueKey()" mean?

cdc/model/sink.go
sdojjy: fixed

cdc/model/sink.go
sdojjy: fixed

cdc/model/sink.go
sdojjy: here we check if the column is a part of the handle key or a unique key,   if so we should check if the value has been changed,
for the condition " col.Flag.IsHandleKey()) && preCol != nil && (preCol.Flag.IsUniqueKey()" I think it's safe to split the row because we can always split an update to delete + insert.

cdc/model/sink.go
sdojjy: this pr focuses on the duplicated key case, the pre-condition is more than two update events emitted.

sdojjy: convertRowChangedEvents checks "enable-old-value" and handle key.

cdc/model/sink_test.go
sdojjy: yes, but let keep this test case to just the test order b logic

nongfushanquan (approved):

PR #9494, opened 2023-08-03
dm: remove hard-coded secret key and support using customized key

### LINKED ISSUE #9492: remove hard-coded secret key and support using customized key

no

- remove hard-coded secret key in source code, as it's taken as compromised and might cause security issue.
- support using customized key to encrypt/decrypt password by add a config and a cli flag `secret-key-path` to `dm-master` to specify a secret key file, the key should be a valid hex encoded AES-256 key.
- remove dependency of secret-key from dmctl, now it calls api of `dm-master` to do encrypt/decrypt related work
- remove `decrypt` sub-command, to avoid user without access to secret-key to be able to see decrypted password.

this feature breaks backward compatibility
- when upgrade from old dm clusters which uses encrypted password, need to upgrade master with `secret-key-path` set first, then upgrade worker.
- when exporting config using `config export`, the exported password will encrypted if dm-master is started with a key, else it's plaintext.

we also remove deprecated `--encrypt/--decrypt` flags, user can use `encrypt` sub-command directly.

_No response_

_No response_

D3Hunter:

Issue Number: ref #9492

- see detailed info in issue.
- most code change is due to replacing encrypted password and change of terror string

#### Tests 

 - [x] Unit test
 - [x] Integration test
   - test upgrade manually
   - test newer version worker/master is compatible with older worker/master
 - No code

break compatibility
need to update user doc
### Release note 

```release-note
`None`.
```


dm/config/source_config_test.go
```diff
@@ -76,15 +79,6 @@ func TestConfigFunctions(t *testing.T) {
...
 
-	// test decrypt password
-	clone1.From.Password = "1234"
-	// fix empty map after marshal/unmarshal becomes nil
-	clone1.From.Session = cfg.From.Session
-	clone1.Tracer = map[string]interface{}{}
-	clone1.Filters = []*bf.BinlogEventRule{}
-	clone2 := cfg.DecryptPassword()
-	require.Equal(t, clone1, clone2)
-
```
D3Hunter, line 87: merged into TestConfigVerify

dm/tests/dmctl_basic/conf/key.txt
D3Hunter, line 1: after this pr, only `dmctl_basic`/`import_v10x`/`new_relay` has tests for custom cipher

D3Hunter:
@lance6716 @GMHDBJD

dm/common/config/config.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return terror.ErrConfigSecretKeyPath.Generate(err)
+	}
+	contentStr := strings.TrimSpace(string(content))
+	decodeContent, err := hex.DecodeString(contentStr)
+	if err != nil {
+		return terror.ErrConfigSecretKeyPath.Generate(err)
+	}
+	if len(decodeContent) != 32 {
+		return terror.ErrConfigSecretKeyPath.Generate("the secret key must be a hex string of length 64")
```
lance6716, line 47:
```suggestion
		return terror.ErrConfigSecretKeyPath.Generate("the secret key must be a hex AES-256 key of length 64")
```

dm/ctl/common/config.go
```diff
@@ -72,10 +68,7 @@ func DefineConfigFlagSet(fs *pflag.FlagSet) {
 	fs.String("ssl-ca", "", "Path of file that contains list of trusted SSL CAs for connection.")
 	fs.String("ssl-cert", "", "Path of file that contains X509 certificate in PEM format for connection.")
 	fs.String("ssl-key", "", "Path of file that contains X509 key in PEM format for connection.")
-	fs.String(EncryptCmdName, "", "Encrypts plaintext to ciphertext.")
-	fs.String(DecryptCmdName, "", "Decrypts ciphertext to plaintext.")
-	_ = fs.MarkHidden(EncryptCmdName)
-	_ = fs.MarkHidden(DecryptCmdName)
+	fs.String("secret-key-path", "", "Path of file that contains secret key for encrypting and decrypting password, the secret key should be a hex string of length 64.")
```
lance6716, line 71:
```suggestion
	fs.String("secret-key-path", "", "Path of file that contains secret key for encrypting and decrypting password, the secret key should be a hex AES-256 key of length 64.")
```

dm/master/config.go
```diff
@@ -88,6 +89,7 @@ func NewConfig() *Config {
 	fs.Var(&cfg.CertAllowedCN, "cert-allowed-cn", "the trusted common name that allowed to visit")
 
 	fs.StringVar(&cfg.V1SourcesPath, "v1-sources-path", "", "directory path used to store source config files when upgrading from v1.0.x")
+	fs.StringVar(&cfg.SecretKeyPath, "secret-key-path", "", "path of file that contains secret key for encrypting and decrypting password, the secret key should be a hex string of length 64")
```
lance6716, line 92:
```suggestion
	fs.StringVar(&cfg.SecretKeyPath, "secret-key-path", "", "path of file that contains secret key for encrypting and decrypting password, the secret key should be a hex AES-256 key of length 64")
```

lance6716 (commented):
can you point out which integration test cover the new secret key config? there're too many changes

D3Hunter:
> can you point out which integration test cover the new secret key config? there're too many changes

https://github.com/pingcap/tiflow/pull/9494#discussion_r1452900397

dm/master/server.go
```diff
@@ -347,7 +347,8 @@ func (s *Server) RegisterWorker(ctx context.Context, req *pb.RegisterWorkerReque
 	}
 	log.L().Info("register worker successfully", zap.String("name", req.Name), zap.String("address", req.Address))
 	return &pb.RegisterWorkerResponse{
-		Result: true,
+		Result:    true,
+		SecretKey: s.cfg.SecretKey,
```
lance6716, line 351: 😂 Do you have a clear boundary that which communication or file access should be protected? If a malicious dm-worker connects to the cluster it will read the secret key.

dm/pkg/encrypt/encrypt.go
```diff
@@ -18,31 +18,61 @@ import (
...
 	"crypto/rand"
-	"encoding/hex"
 
+	"github.com/pingcap/tiflow/dm/pkg/log"
 	"github.com/pingcap/tiflow/dm/pkg/terror"
+	"github.com/pingcap/tiflow/pkg/errors"
 )
 
-var (
-	secretKey, _ = hex.DecodeString("a529b7665997f043a30ac8fadcb51d6aa032c226ab5b7750530b12b8c1a16a48")
```
lance6716, line 27: where's the logic that new version DM cluster can decrypt the encrypted password of old version cluster?

lance6716 (commented):
rest lgtm

dm/master/server.go
D3Hunter: user should use inter-master-worker TLS to protect this case

dm/pkg/encrypt/encrypt.go
D3Hunter: you mean existing sources/tasks? they're stored in decrypted form actually, dm-worker don't handle encrypt/decrypt.

to upgrade, user should use the same secret key as before to upgrade master first, then worker

lance6716: > you mean existing sources/tasks? they're stored in decrypted form actually, dm-worker don't handle encrypt/decrypt.

I found some `subTaskCfg.DecryptPassword()` call from worker, like https://github.com/pingcap/tiflow/blob/b1c22cfa07e7717e941fd4ccf626bccaf1862c9d/dm/worker/server.go#L931

I'll check if it has no effect soon.

lance6716 (approved):

D3Hunter:
@GMHDBJD

GMHDBJD (commented):
LGTM

GMHDBJD (approved):

PR #9512, opened 2023-08-07
kafka(ticdc): kafka sink support compression after encoding event into message

### LINKED ISSUE #9596: kafka sink support compress the message value

3AceShowHand:

Issue Number: close #9596 

* After encoding the event into message, compress the value before compare against the message size limit.
* Update the monitoring to track the compression ratio, also record the number of records carried by the request.

image

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
Kafka sink support compress the message value before compare against the message size limit to reduce large message count.
```


pkg/compression/compress.go
```diff
@@ -0,0 +1,85 @@
...
+func Supported(cc string) bool {
+	switch cc {
+	case None, Snappy, LZ4:
+		return true
+	}
+	return false
+}
+
+// Encode the given data by the given compression codec.
+func Encode(cc string, data []byte) ([]byte, error) {
```
nongfushanquan, line 45: Why not use decompress and compress  to name the methods directly?

3AceShowHand: It's conventional to use `Encode` and `Decode`.

nongfushanquan (approved):

asddongmen (approved):

Issue #9488, opened 2023-08-03
rename ddl cannot be synchronized when the new schema is not in filter


CharlesCheung96:
### What did you do?

1. create changefeed with the following config:
```toml
[filter]
rules = ['*.*', '!ignore1.*']
```
2. create database test; create table test.a and test.a_new;
3. create database ignore1;
5. RENAME TABLE test.a TO ignore1.a, test.a_new TO ignore1.a_new;
6. create table test.b;
7. RENAME TABLE test.b to ignore1.b;

### What did you expect to see?

rename ddl is synchronized.

### What did you see instead?

step 5 rename tables ddl: "[CDC:ErrSnapshotSchemaNotFound]schema
step 7 rename table ddl is skipped

### Versions of the cluster

Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

```console
(paste TiDB cluster version here)
```

Upstream TiKV version (execute `tikv-server --version`):

```console
(paste TiKV version here)
```

TiCDC version (execute `cdc version`):

```console
(paste TiCDC version here)
```

PR #9521, opened 2023-08-08
ddlPuller (ticdc): fix rename table ddl

asddongmen:

Issue Number: ref #9488 

This PR make rename table ddl like: `rename table test.t1 to ignore.t1` can be replicated. (The schema name `ignore` means it does not match the changefeed filter rule.)

#### Tests 

 - Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/puller/ddl_puller_test.go
+++ b/cdc/puller/ddl_puller_test.go
@@ -165,6 +165,9 @@ func TestHandleRenameTable(t *testing.T) {
 		"test1.t66",
 		"test1.t99",
 		"test1.t100",
+		"test1.t20230808",
+		"test1.t202308081",
+		"test1.t202308082",
 
 		"test2.t4",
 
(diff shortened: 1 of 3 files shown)


3AceShowHand (approved):

cdc/puller/ddl_puller_test.go
```diff
@@ -340,6 +357,22 @@ func TestHandleRenameTable(t *testing.T) {
 		skip, err = ddlJobPullerImpl.handleJob(job)
 		require.NoError(t, err)
 		require.True(t, skip)
+
+		// since test1.t20230808 is  in filter rule, replicate it
```
CharlesCheung96, line 361:
```suggestion
		// since test1.t20230808 is in filter rule, replicate it
```

CharlesCheung96 (approved):

PR #9522, opened 2023-08-08
dm: fix validator deadlock and enhance retry

### LINKED ISSUE #9257: dm validator should retry on transient error and there should be no deadlock when bad thing happens
### What did you do?

there are 2 issues,
1. `Error 9005: Region is unavailable` is not retryable even though it could/should be transient error https://github.com/pingcap/tiflow/blob/master/dm/syncer/validate_worker.go#L170
2. there could be deadlock when errors happen, which will block any `dmctl` operation on the task, particularly, one cannot restart the task by `dmctl stop-task/start-task` because of the deadlock.

in data validator thread,

- the data validator tries to stop when encounter the first error https://github.com/pingcap/tiflow/blob/v7.1.0/dm/syncer/data_validator.go#L462
- in `stopInner`, it waits for all workers to be done https://github.com/pingcap/tiflow/blob/v7.1.0/dm/syncer/data_validator.go#L743
- thus, when it waits for the `wg`, and no one will drain the `errChan`

meantime, in worker threads,

- it reports error to the `errChan` https://github.com/pingcap/tiflow/blob/v7.1.0/dm/syncer/validate_worker.go#L170
- the bounded `errChan` will be filled fully because the validator doesn't drain the errors
- thus the workers waits on `errChan` holding the `wg` unreleased, which blocks the `innerStop` to finish

1. region is not available could be retried
2. there is no deadlock and the task could be restart by `stop-task/start-task` in any cases

### What did you see instead?

1. region is not available stops the validator
2. deadlock happens when a lot of workers report error

### Versions of the cluster

master

```
{
    "result": true,
    "msg": "",
    "sources": [
        {
            "result": true,
            "msg": "",
            "sourceStatus": {
                "source": "tidbmigration-sample-src",
                "worker": "test-single-cell-dm-worker-0",
                "result": null,
                "relayStatus": null
            },
            "subTaskStatus": [
                {
                    "name": "tidbmigration-sample-task",
                    "stage": "Running",
                    "unit": "Sync",
                    "result": null,
                    "unresolvedDDLLockID": "",
                    "sync": {
                        "totalEvents": "19635499",
                        "totalTps": "41072",
                        "recentTps": "12483",
                        "masterBinlog": "(mysql-bin-changelog.061817, 57798824)",
                        "masterBinlogGtid": "",
                        "syncerBinlog": "(mysql-bin-changelog.061817, 55744546)",
                        "syncerBinlogGtid": "00000000-0000-0000-0000-000000000000:0",
                        "blockingDDLs": [
                        ],
                        "unresolvedGroups": [
                        ],
                        "synced": false,
                        "binlogType": "remote",
                        "secondsBehindMaster": "0",
                        "blockDDLOwner": "",
                        "conflictMsg": "",
                        "totalRows": "19635499",
                        "totalRps": "41072",
                        "recentRps": "12483"
                    },
                    "validation": {
                        "task": "tidbmigration-sample-task",
                        "source": "tidbmigration-sample-src",
                        "mode": "full",
                        "stage": "Running",
                        "validatorBinlog": "(mysql-bin-changelog.061762, 77247898)",
                        "validatorBinlogGtid": "00000000-0000-0000-0000-000000000000:0",
                        "result": {
                            "isCanceled": false,
                            "errors": [
                                {
                                    "ErrCode": 43005,
                                    "ErrClass": "validator",
                                    "ErrScope": "internal",
                                    "ErrLevel": "high",
                                    "Message": "failed to validate row change",
                                    "RawCause": "Error 9005: Region is unavailable",
                                    "Workaround": ""
                                }
                            ],
                            "detail": null
                        },
                        "processedRowsStatus": "insert/update/delete: 1673826/1673561/371420",
                        "pendingRowsStatus": "insert/update/delete: 17259/15936/3015",
                        "errorRowsStatus": "new/ignored/resolved: 0/0/0"
                    }
                }
            ]
        }
    ]
}
```

D3Hunter:

Issue Number: close #9257

- fix deadlock between error-process-routine and worker routine when err channel is full
- remove object lock when `fillResult`
- enhance retry by reusing the resumable errors in syncer

#### Tests 

 - [x] Unit test
 - Integration test
 - No code

### Release note 

```release-note
dm: fix validator deadlock and enhance retry
```

### DIFF
--- a/dm/syncer/data_validator.go
+++ b/dm/syncer/data_validator.go
@@ -156,7 +156,7 @@ func (vs *tableValidateStatus) stopped(msg string) {
 	vs.message = msg
 }
 
-// DataValidator
+// DataValidator is used to continuously validate incremental data migrated to downstream by dm.
 // validator can be start when there's syncer unit in the subtask and validation mode is not none,
 // it's terminated when the subtask is terminated.
 // stage of validator is independent of subtask, pause/resume subtask doesn't affect the stage of validator.
@@ -164,8 +164,13 @@ func (vs *tableValidateStatus) stopped(msg string) {
 // validator can be in running or stopped stage
 // - in running when it's started with subtask or started later on the fly.
 // - in stopped when validation stop is executed.
+//
+// for each subtask, before it's closed/killed, only one DataValidator object is created,
+// on "dmctl validation stop/start", will call Stop and Start on the same object.
 type DataValidator struct {
+	// used to sync Stop and Start operations.
 	sync.RWMutex
+
 	cfg    *config.SubTaskConfig
 	syncer *Syncer
 	// whether validator starts together with subtask
@@ -199,7 +204,7 @@ type DataValidator struct {
 
 	// fields in this field block are guarded by stateMutex
 	stateMutex  sync.RWMutex
-	stage       pb.Stage
+	stage       pb.Stage // only Running or Stopped is allowed for validator
 	flushedLoc  *binlog.Location
 	result      pb.ProcessResult
 	tableStatus map[string]*tableValidateStatus
@@ -352,7 +357,7 @@ func (v *DataValidator) Start(expect pb.Stage) {
 	}
 
 	if err := v.initialize(); err != nil {
-		v.fillResult(err, false)
+		v.fillResult(err)
 		return
 	}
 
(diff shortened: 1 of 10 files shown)


D3Hunter:
@lance6716 @GMHDBJD

dm/syncer/data_validator.go
```diff
@@ -199,7 +204,7 @@ type DataValidator struct {
 
 	// fields in this field block are guarded by stateMutex
```
lance6716, line 205: seems we can update the comment about which members this mutex guards

lance6716 (commented):
will review later

dm/unit/unit.go
```diff
@@ -146,6 +147,35 @@ func IsResumableError(err *pb.ProcessError) bool {
...
+func IsResumableDBError(err error) bool {
+	if err == nil {
+		return true
+	}
+
+	err = errors.Cause(err)
+	if err == context.Canceled {
+		return false
+	}
+
```
lance6716, line 163: as its name, should we check it's a DB error? like a error instance from mysql.MySQLError

dm/syncer/data_validator.go
D3Hunter: which field do you mean? the range is not changed

GMHDBJD (approved):
LGTM

lance6716 (approved):

PR #9530, opened 2023-08-09
sink(cdc): fix internal retry algothrim

### LINKED ISSUE #9518: TiCDC changefeed stucks after cdc owner to Kafka network partition recovered
### What did you do?

1. create kafka changefeed
2. run workload, and at the same time inject cdc owner to Kafka network partition 40 seconds for every 5 minutes, stop the chaos injection after 1 hours.
3. Check CDC lag after chaos injection is finished.

CDC lag should be normal

### What did you see instead?

Chaos injection recoveryed at 03:28:53, checkpoint stucks at 03:30
CDC changefeed stucks
image

### Versions of the cluster

TiCDC version (execute `cdc version`): master

```console
["Welcome to Change Data Capture (CDC)"] [release-version=v7.4.0-alpha] [git-hash=e61669614f12a3e1f6bfc6af38ec7c297e7ede93] [git-branch=heads/refs/tags/v7.4.0-alpha] [utc-build-time="2023-08-07 11:36:10"] [go-version="go version go1.20.7 linux/amd64"] [failpoint-build=false]
```

hicqu:

Issue Number: close #9518 

Term Definition

- Same Error: The changefeed has not been advanced after the error occurred.

Prior to this PR, if the `DDLSink` or `DMLSink` returned the same error continuously, the `SinkManager `would retry it for 30 minutes. If different errors occurred, the `SinkManager` would reset the retry time counter. Once the 30 minutes were up, the sink manager would report an `UnretryableError` to the `owner` and the changefeed would fail immediately. However, the `SinkManager` couldn't accurately determine whether the error returned by the `DDLSink` or `DMLSink` was the same error (It does not reset the retry time counter when correctly), leading to incorrect `UnretryableError` reports and unexpected changefeed failures.

With this PR, if the `DDLSink` or `DMLSink` returns the same error continuously, the `SinkManager` will continue retrying indefinitely but will report a warning to the `owner`. Since the `owner` has a broader view of the changefeed, it can determine if the error reported by the `sinkManager` is the same error. This PR takes the checkpoint into account to make the decision. The underlying logic is simple: if the checkpoint advanced since the last error was reported, the subsequent error is not the same as the previous error. Thus, the retry time is reset to ensure that the subsequent error will be also retry for 30 minutes.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


cdc/owner/feed_state_manager.go
```diff
@@ -562,6 +572,28 @@ func (m *feedStateManager) handleWarning(errs ...*model.RunningError) {
...
 	}
 	lastError := errs[len(errs)-1]
+
+	if m.state != nil && m.state.Status != nil {
+		currTime := m.upstream.PDClock.CurrentTime()
+		ckptTime := oracle.GetTimeFromTS(m.state.Status.CheckpointTs)
+		// Conditions:
+		// 1. checkpoint lag is large enough;
+		// 2. checkpoint hasn't been advanced for a long while;
+		// 3. the changefeed has been initialized.
```
hicqu, line 582: @asddongmen please help to check whether the condition is correct or not.

cdc/owner/ddl_sink.go
```diff
@@ -184,6 +193,22 @@ func (s *ddlSinkImpl) retrySinkActionWithErrorReport(ctx context.Context, action
...
 }
 
+func (s *ddlSinkImpl) observedRetrySinkAction(ctx context.Context, name string, action func() error) (err error) {
+	errCh := make(chan error, 1)
+	go func() { errCh <- s.retrySinkAction(ctx, name, action) }()
+	for {
+		select {
+		case err := <-errCh:
+			return err
+		case <-time.After(30 * time.Second):
```
asddongmen, line 203: Is it better to use a ticker instead of timer here?

cdc/owner/feed_state_manager.go
```diff
@@ -110,6 +113,13 @@ func (m *feedStateManager) shiftStateWindow(state model.FeedState) {
 }
 
 func (m *feedStateManager) Tick(state *orchestrator.ChangefeedReactorState) (adminJobPending bool) {
+	if state != nil && state.Status != nil {
```
asddongmen, line 116: state should never be `nil` here, ie is a redundant check.

cdc/owner/feed_state_manager.go
```diff
@@ -562,6 +572,28 @@ func (m *feedStateManager) handleWarning(errs ...*model.RunningError) {
 		return
 	}
 	lastError := errs[len(errs)-1]
+
+	if m.state != nil && m.state.Status != nil {
```
asddongmen, line 576: m.state should never be `nil` here.

cdc/owner/feed_state_manager.go
```diff
@@ -562,6 +572,28 @@ func (m *feedStateManager) handleWarning(errs ...*model.RunningError) {
...
+		currTime := m.upstream.PDClock.CurrentTime()
+		ckptTime := oracle.GetTimeFromTS(m.state.Status.CheckpointTs)
+		// Conditions:
+		// 1. checkpoint lag is large enough;
+		// 2. checkpoint hasn't been advanced for a long while;
+		// 3. the changefeed has been initialized.
+		if currTime.Sub(ckptTime) > defaultBackoffMaxElapsedTime &&
+			time.Since(m.checkpointTsAdvanced) > defaultBackoffMaxElapsedTime &&
+			m.state.Status.MinTableBarrierTs > m.state.Info.StartTs {
```
asddongmen, line 585: `MinTableBarrierTs` cannot be used to indicate if the changefeed has finished incremental scanning, since it is equal to `ddlPuller` resolvedTs if there are no DDL events.

asddongmen (approved):

pkg/retry/error_retry.go
```diff
@@ -71,15 +80,15 @@ func (r *ErrorRetry) GetRetryBackoff(err error) (time.Duration, error) {
 	// it means the last error is retry success, and the sink is running well for some time
 	if r.lastInternalError == nil ||
```
CharlesCheung96, line 81: Whether lastInternalError should be deleted?

hicqu: I prefer to keep the logic because maybe it will be useful for some future cases. Now we can just use an infinite retry duration to avoid this.

CharlesCheung96 (approved):

asddongmen:
/need-cherry-pick-release-6.5

Issue #9571, opened 2023-08-15
CDC changefeed error "[CDC:ErrConvertDDLToEventTypeFailed]failed to convert ddl "


fubinzh:
### What did you do?

1. create kafka changefeed 
```
cdc  cli  changefeed  create "--server=127.0.0.1:8301" "--sink-uri=kafka://downstream-kafka.cdc-testbed-tps-1891749-1-356:9092/broker_fail?max-message-bytes=1048576&replication-factor=2&enable-tidb-extension=true&protocol=canal-json" "--changefeed-id=kafka-random-broker-fail-canal-json"
```
3. Run sysbench workload
```
sysbench --db-driver=mysql --mysql-host=`nslookup upstream-tidb.cdc-testbed-tps-1891749-1-356 | awk -F: '{print $2}' | awk 'NR==5' | sed s/[[:space:]]//g`  --mysql-port=4000 --mysql-user=root --mysql-db=workload --tables=100 --table-size=100000 --create_secondary=off --debug=true --threads=100 --mysql-ignore-errors=2013,1213,1105,1205,8022,8027,8028,9004,9007,1062 oltp_update_non_index prepare
```

### What did you expect to see?

CDC changefeed should not fail

### What did you see instead?

Changefeed report error 
```
[...]

PR #9577, opened 2023-08-15
filter (ticdc): add a lock to protect the parser from being called concurrently.

asddongmen:

Issue Number: close https://github.com/pingcap/tiflow/issues/9571

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/scripts/check-diff-line-width.sh
+++ b/scripts/check-diff-line-width.sh
@@ -26,7 +26,6 @@ ERROR_THRESHOLD=140
 git --no-pager diff $BASE_HASH -U0 -- cdc pkg cmd \
 	-- ':(exclude)*_gen.go' \
 	-- ':(exclude)*_gen_test.go' \
-	-- ':(exclude)*_test.go' \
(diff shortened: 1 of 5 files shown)


scripts/check-diff-line-width.sh
```diff
@@ -26,6 +26,7 @@ ERROR_THRESHOLD=140
 git --no-pager diff $BASE_HASH -U0 -- cdc pkg cmd \
 	-- ':(exclude)*_gen.go' \
 	-- ':(exclude)*_gen_test.go' \
+	-- ':(exclude)*_test.go' \
```
CharlesCheung96, line 29: It is better to use raw strings in the test and delete this line.

3AceShowHand (approved):

scripts/check-diff-line-width.sh
asddongmen: ok

CharlesCheung96 (approved):

Issue #9413, opened 2023-07-20
Support Ticdc output data to pulsar


yumchina:
### Is your feature request related to a problem?

Ticdc can output data to pulsar

### Describe the feature you'd like

Ticdc output data to pulsar like kafka.
1 Support some config with a pulsar client.
2 Ticdc output DDL/DML messages to pulsar with Canal/Canal-Json/Maxwell protocol. 
I will make a pulsar sink to output data to pulsar. All data formats will be similar to Kafka's, but because the client is different, the details will be slightly different.
This new function will provide ticdc with the ability to output data to pulsar, allowing users to reuse existing pulsar clusters.

### Describe alternatives you've considered

No 

### Teachability, Documentation, Adoption, Migration Strategy

[ticdc-pulsar-sink ](https://github.com/pingcap/tiflow/blob/master/docs/design/2023-07-04-ticdc-pulsar-sink.md)

asddongmen:
All development are finished.

PR #9578, opened 2023-08-15
ticdc(sink) move pulsar sink-uri configs to sink of replica config

yumchina:

move pulsar sink-uri configs to sink of replica config
Issue Number: ref https://github.com/pingcap/tiflow/issues/9413

#### Tests 

 - Unit test
 

No

No

### Release note 
```release-note
None
```


pkg/config/sink.go
```diff
@@ -331,6 +333,74 @@ type KafkaConfig struct {
...
+func (p *PulsarCompressionType) Value() pulsar.CompressionType {
+	if p == nil {
+		return 0
+	}
+	switch strings.ToLower(string(*p)) {
+	case "lz4":
+		return pulsar.LZ4
+	case "zlib":
+		return pulsar.ZLib
+	case "zstd":
```
nongfushanquan, line 349: whether is the  snappy type supported ?

yumchina: Pulsar client doesn't support snappy type

nongfushanquan (approved):

asddongmen (approved):

PR #9688, opened 2023-09-06
sink(ticdc): split the updat events all at once

### LINKED ISSUE #9689: split the update event when resolve the transaction.
as title

3AceShowHand:

Issue Number: close #9689

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```

### DIFF
--- a/cdc/model/sink.go
+++ b/cdc/model/sink.go
@@ -768,9 +768,6 @@ func (t *SingleTableTxn) GetCommitTs() uint64 {
 
 // TrySplitAndSortUpdateEvent split update events if unique key is updated
 func (t *SingleTableTxn) TrySplitAndSortUpdateEvent() error {
-	if len(t.Rows) < 2 {
-		return nil
-	}
 	newRows, err := trySplitAndSortUpdateEvent(t.Rows)
 	if err != nil {
 		return errors.Trace(err)
--- a/cdc/model/sink.go
+++ b/cdc/model/sink.go
@@ -271,7 +271,7 @@ func (r *RedoLog) GetCommitTs() Ts {
 }
 
 // TrySplitAndSortUpdateEvent redo log do nothing
-func (r *RedoLog) TrySplitAndSortUpdateEvent() error {
+func (r *RedoLog) TrySplitAndSortUpdateEvent(_ bool) error {
 	return nil
 }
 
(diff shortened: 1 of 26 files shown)


cdc/model/sink.go
```diff
@@ -779,6 +780,20 @@ func (t *SingleTableTxn) TrySplitAndSortUpdateEvent() error {
 	return nil
 }
```
CharlesCheung96, line 782:
```suggestion
// Whether split a single update event into delete and insert events？
//
// For the MySQL Sink, there is no need to split a single unique key changed update event, this
// is also to keep the backward compatibility, the same behavior as before.
//
// For the Kafka and Storage sink, always split a single unique key changed update event, since:
// 1. Avro and CSV does not output the previous column values for the update event, so it would
// cause consumer missing data if the unique key changed event is not split.
// 2. Index-Value Dispatcher cannot work correctly if the unique key changed event isn't split.
```

cdc/sink/dmlsink/mq/mq_dml_sink.go
```diff
@@ -185,3 +185,8 @@ func (s *dmlSink) Close() {
 func (s *dmlSink) Dead() <-chan struct{} {
 	return s.dead
 }
+
+// Scheme returns the scheme of this sink.
+func (s *dmlSink) Scheme() string {
+	return sink.KafkaScheme
```
CharlesCheung96, line 191: Better to return the real scheme here.

cdc/sink/dmlsink/txn/txn_dml_sink.go
```diff
@@ -173,3 +173,7 @@ func (s *dmlSink) Close() {
 func (s *dmlSink) Dead() <-chan struct{} {
 	return s.dead
 }
+
+func (s *dmlSink) Scheme() string {
+	return sink.MySQLScheme
```
CharlesCheung96, line 178: ditto

cdc/model/sink.go
CharlesCheung96: Better to use a method comment.

CharlesCheung96 (approved):

CharlesCheung96 (approved):

asddongmen (approved):

PR #9701, opened 2023-09-08
ddl(ticdc): async exec add index ddl

### LINKED ISSUE #9644: Reduce the latency when replicating the "add index" ddl to the downstream(tidb)
### What did you do?

When an "ADD INDEX" DDL is executed at the upstream, TiCDC will execute this DDL SQL at the downstream (TiDB) and hold the following DMLs. 

The latency is increasing.

### What did you see instead?

DML will not be blocked by this ddl and can continue to execute.

### Versions of the cluster

master

sdojjy:

Issue Number: close #9644

After this pull request, the "add index" DDL will be executed asynchronously, and it will no longer block the progression of DML replication. There is a risk associated with executing the "add index" DDL in this manner: If an error occurs during the execution of the "add index" DDL in downstream TiDB, the DDL may be lost. However, the likelihood of this situation occurring is low since the DDL was executed successfully upstream, and it will not impact the data correctness downstream.

#### Tests 

 - Unit test

### Release note 

```release-note
TiCDC exec adding index DDL  asynchronously and does not block checkpoint ts
```

### DIFF
--- a/cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
+++ b/cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
@@ -259,42 +259,46 @@ func (m *DDLSink) Close() {
 // asyncExecAddIndexDDLIfTimeout executes ddl in async mode.
 func (m *DDLSink) asyncExecAddIndexDDLIfTimeout(ctx context.Context, ddl *model.DDLEvent) error {
 	done := make(chan error, 1)
-	// wait for 1 second at most
-	tick := time.NewTimer(1 * time.Second)
+	// wait for 2 seconds at most
+	tick := time.NewTimer(2 * time.Second)
 	defer tick.Stop()
-	log.Info("async exec ddl start",
+	log.Info("async exec add index ddl start",
 		zap.String("changefeedID", m.id.String()),
 		zap.Uint64("commitTs", ddl.CommitTs),
 		zap.String("ddl", ddl.Query))
 	go func() {
 		if err := m.execDDLWithMaxRetries(ctx, ddl); err != nil {
-			log.Error("async exec ddl failed",
+			log.Error("async exec add index ddl failed",
 				zap.String("changefeedID", m.id.String()),
 				zap.Uint64("commitTs", ddl.CommitTs),
 				zap.String("ddl", ddl.Query))
 			done <- err
 			return
 		}
-		log.Info("async exec ddl done",
+		log.Info("async exec add index ddl done",
 			zap.String("changefeedID", m.id.String()),
-			zap.Uint64("commitTs", ddl.CommitTs))
+			zap.Uint64("commitTs", ddl.CommitTs),
+			zap.String("ddl", ddl.Query))
 		done <- nil
 	}()
-	for {
-		select {
-		case <-ctx.Done():
-			log.Info("async ddl exits as canceled",
-				zap.String("changefeedID", m.id.String()),
-				zap.Uint64("commitTs", ddl.CommitTs),
-				zap.String("ddl", ddl.Query))
-		case err := <-done:
-			return err
-		case <-tick.C:
-			log.Info("async ddl is still running",
-				zap.String("changefeedID", m.id.String()),
-				zap.Uint64("commitTs", ddl.CommitTs),
-				zap.String("ddl", ddl.Query))
-			return nil
-		}
+
+	select {
+	case <-ctx.Done():
+		log.Info("async add index ddl exits as canceled",
+			zap.String("changefeedID", m.id.String()),
+			zap.Uint64("commitTs", ddl.CommitTs),
+			zap.String("ddl", ddl.Query))
+		return nil
+	case err := <-done:
+		return err
+	case <-tick.C:
+		// if the ddl is still running, we just return nil,
+		// then if the ddl is failed, the downstream ddl is lost.
+		// because the checkpoint ts is forwarded.
+		log.Info("async add index ddl is still running",
+			zap.String("changefeedID", m.id.String()),
+			zap.Uint64("commitTs", ddl.CommitTs),
+			zap.String("ddl", ddl.Query))
+		return nil
 	}
 }
--- a/cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
+++ b/cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
@@ -84,6 +84,11 @@ func NewDDLSink(
 		return nil, err
 	}
 
+	cfg.IsTiDB, err = pmysql.CheckIsTiDB(ctx, db)
+	if err != nil {
+		return nil, err
+	}
+
 	m := &DDLSink{
 		id:         changefeedID,
 		db:         db,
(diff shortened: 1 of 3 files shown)


cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -247,3 +250,46 @@ func (m *DDLSink) Close() {
 		}
 	}
 }
+
+// asyncExecAddIndexDDLIfTimeout executes ddl in async mode.
+func (m *DDLSink) asyncExecAddIndexDDLIfTimeout(ctx context.Context, ddl *model.DDLEvent) error {
```
nongfushanquan, line 255: what will happen if return error ?

sdojjy: cdc will retry this changefeed

sdojjy: if the DDL execution time is more than 1 second,  this function always returns a nil.

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -247,3 +250,46 @@ func (m *DDLSink) Close() {
...
+	go func() {
+		if err := m.execDDLWithMaxRetries(ctx, ddl); err != nil {
+			log.Error("async exec ddl failed",
+				zap.String("changefeedID", m.id.String()),
+				zap.Uint64("commitTs", ddl.CommitTs),
+				zap.String("ddl", ddl.Query))
+			done <- err
+			return
+		}
+		log.Info("async exec ddl done",
```
asddongmen, line 278: Should also print this ddl

asddongmen (approved):

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -84,6 +84,11 @@ func NewDDLSink(
 		return nil, err
 	}
 
+	cfg.IsTiDB, err = pmysql.CheckIsTiDB(ctx, db)
+	if err != nil {
+		return nil, err
+	}
```
3AceShowHand, line 90: If the downstream is MySQL, an error is returned here, so the changefeed cannot work properly ?

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -247,3 +255,46 @@ func (m *DDLSink) Close() {
...
 	}
 }
+
+// asyncExecAddIndexDDLIfTimeout executes ddl in async mode.
+func (m *DDLSink) asyncExecAddIndexDDLIfTimeout(ctx context.Context, ddl *model.DDLEvent) error {
+	done := make(chan error, 1)
+	// wait for 1 second at most
+	tick := time.NewTimer(1 * time.Second)
+	defer tick.Stop()
+	log.Info("async exec ddl start",
```
3AceShowHand, line 265: the ddl must be add index, how about rewrite this to `async exec add index ddl start`

3AceShowHand, line 265:
```suggestion
	log.Info("async exec add index ddl start",
```

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -247,3 +255,46 @@ func (m *DDLSink) Close() {
...
+		zap.String("changefeedID", m.id.String()),
+		zap.Uint64("commitTs", ddl.CommitTs),
+		zap.String("ddl", ddl.Query))
+	go func() {
+		if err := m.execDDLWithMaxRetries(ctx, ddl); err != nil {
+			log.Error("async exec ddl failed",
```
3AceShowHand, line 271:
```suggestion
			log.Error("async exec add index ddl failed",
```

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -247,3 +255,46 @@ func (m *DDLSink) Close() {
...
+				zap.String("changefeedID", m.id.String()),
+				zap.Uint64("commitTs", ddl.CommitTs),
+				zap.String("ddl", ddl.Query))
+			done <- err
+			return
+		}
+		log.Info("async exec ddl done",
```
3AceShowHand, line 278:
```suggestion
		log.Info("async exec add index ddl done",
```

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -247,3 +255,46 @@ func (m *DDLSink) Close() {
...
+			zap.String("changefeedID", m.id.String()),
+			zap.Uint64("commitTs", ddl.CommitTs))
+		done <- nil
+	}()
+	for {
```
3AceShowHand, line 283: The DDL is executing in another goroutine, and this `for` loop looks useless, since the ticker will be triggered 1s later, so the method return, and the error channel `done` is not correctly checked.

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -247,3 +255,46 @@ func (m *DDLSink) Close() {
...
+		select {
+		case <-ctx.Done():
```
3AceShowHand, line 285: What happens if the `ctx.Done` triggered ?

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -247,3 +255,46 @@ func (m *DDLSink) Close() {
...
+			log.Info("async ddl exits as canceled",
+				zap.String("changefeedID", m.id.String()),
+				zap.Uint64("commitTs", ddl.CommitTs),
+				zap.String("ddl", ddl.Query))
+		case err := <-done:
+			return err
+		case <-tick.C:
```
CharlesCheung96, line 292: Is it possible to report an error due to tcp connection timeout during each retry? If possible, and the ticker time limit is exceeded, then the downstream ddl is lost. 

In addition, if this is the expected behavior, please add a comment to explain this problem.

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
sdojjy: fixed

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
sdojjy: MySQL downstream do not return an error

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
sdojjy: done

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
sdojjy: fixed

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
sdojjy: tasks are canceled

3AceShowHand (approved):

PR #9742, opened 2023-09-14
sink(cdc): only check sink stuck for MQ sinks

### LINKED ISSUE #9695: TiCDC sink re-establishes in some unnecessary cases
### What did you do?

After TiCDC sink is stuck for a while, TiCDC will destroy and re-establish the sink. However currently the stuck detection is not good enough. In such cases stuck will be reported mistakely:

* after a long time incremental scan;
* after a sync point is passed;
* if a table hasn't been scheduled for a while.

_No response_

### What did you see instead?

In those cases, sink will be re-established, which is not expected.

### Versions of the cluster

Master branch.

### LINKED ISSUE #9736: stuck check can still make some mistake reports
### What did you do?

Start one TiCDC instance, and run a changefeed with 30 tables. Then

1. pause the changefeed, and prepare some data in the upstream cluster;
2. inject a latency into the sink;
3. resume the changefeed;
4. after a while, we can get the log "stuck too long".

### Why "stuck too long" happens?
It's because the table is the last one to finish incremental scan. There are already `29` tables scheduled into sink. After the last table gets scheduled, its `resolvedTs` has been updated, but its checkpoint can only be advanced after all other tables
pending data has been consumed. So, the stuck check is triggered, which is unexpected.

_No response_

### What did you see instead?

As decribed above.

### Versions of the cluster

Master branch.

hicqu:

Issue Number: close #9736 .
Issue Number: close #9695 .

Other sink categories like `Txn` or `CloudStorage` can handles table sink progress callbacks good enough: callbacks won't be lost forever if sink factory fails internally. Otherwise an error will be reported instead, which can cause sink factory re-established correctly.

We only have met the callback lose issue for MQ sinks, so only check sink stuck for MQ sink is enough.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/sink/dmlsink/factory/factory.go
+++ b/cdc/sink/dmlsink/factory/factory.go
@@ -39,14 +39,24 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+type Category = int
+
+const (
+	CategoryTxn          Category = 1
+	CategoryMQ                    = 2
+	CategoryCloudStorage          = 3
+	CategoryBlackhole             = 4
+)
+
 // SinkFactory is the factory of sink.
 // It is responsible for creating sink and closing it.
 // Because there is no way to convert the eventsink.EventSink[*model.RowChangedEvent]
 // to eventsink.EventSink[eventsink.TableEvent].
 // So we have to use this factory to create and store the sink.
 type SinkFactory struct {
-	rowSink dmlsink.EventSink[*model.RowChangedEvent]
-	txnSink dmlsink.EventSink[*model.SingleTableTxn]
+	rowSink  dmlsink.EventSink[*model.RowChangedEvent]
+	txnSink  dmlsink.EventSink[*model.SingleTableTxn]
+	category Category
 }
 
 // New creates a new SinkFactory by schema.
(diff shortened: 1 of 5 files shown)


sdojjy (approved):

cdc/sink/dmlsink/factory/factory.go
```diff
@@ -39,14 +39,29 @@ import (
...
 // It is responsible for creating sink and closing it.
 // Because there is no way to convert the eventsink.EventSink[*model.RowChangedEvent]
 // to eventsink.EventSink[eventsink.TableEvent].
 // So we have to use this factory to create and store the sink.
 type SinkFactory struct {
-	rowSink dmlsink.EventSink[*model.RowChangedEvent]
-	txnSink dmlsink.EventSink[*model.SingleTableTxn]
+	rowSink  dmlsink.EventSink[*model.RowChangedEvent]
+	txnSink  dmlsink.EventSink[*model.SingleTableTxn]
+	category Category
```
CharlesCheung96, line 64: The scheme type can be obtained from EventSink, so the category is redunant, ref:

https://github.com/pingcap/tiflow/pull/9688/files#diff-afe94bc54d98b2f898bdce770bba8ee919e3643466203ce4d7087136708e1371R22-R24

hicqu: Different `Schema`s can belong to one same `Category`. Use the latter one can make logic more simple and clear.

CharlesCheung96 (approved):

PR #9755, opened 2023-09-15
kafka(ticdc): event router allow hard code topics and set the schema optional in the topic expression

### LINKED ISSUE #9763: topic exression allows the schema part to be optional and accept hard code topic name
as title

3AceShowHand:

Issue Number: close #9763 

* schema is allowed to be optional in the topic expression.
* allow user hard code the topic in the topic expression.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```

### DIFF
--- a/cdc/sink/dmlsink/mq/dispatcher/topic/expression.go
+++ b/cdc/sink/dmlsink/mq/dispatcher/topic/expression.go
@@ -21,6 +21,9 @@ import (
 )
 
 var (
+	// hardCodeTopicNameRe is used to match a topic name which is hard code in the config
+	hardCodeTopicNameRe = regexp.MustCompile(`^([A-Za-z0-9\._\-]+)$`)
+
 	// topicNameRE is used to match a valid topic expression
 	topicNameRE = regexp.MustCompile(
 		`^[A-Za-z0-9\._\-]*(\{schema\})?([A-Za-z0-9\._\-]*\{table\})?[A-Za-z0-9\._\-]*$`,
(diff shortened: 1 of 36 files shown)


cdc/sink/dmlsink/mq/dispatcher/topic/expression.go
```diff
@@ -21,9 +21,12 @@ import (
 )
 
 var (
+	// hardCodeTopicNameRe is used to match a topic name which is hard code in the config
+	hardCodeTopicNameRe = regexp.MustCompile(`^([A-Za-z0-9\._\-]+)$`)
+
 	// topicNameRE is used to match a valid topic expression
 	topicNameRE = regexp.MustCompile(
-		`^[A-Za-z0-9\._\-]*\{schema\}([A-Za-z0-9\._\-]*\{table\})?[A-Za-z0-9\._\-]*$`,
+		`^[A-Za-z0-9\._\-]*(\{schema\})?([A-Za-z0-9\._\-]*\{table\})?[A-Za-z0-9\._\-]*$`,
```
nongfushanquan, line 29: whether the empty case is covered by this expression?

cdc/sink/dmlsink/mq/dispatcher/topic/expression.go
```diff
@@ -21,9 +21,12 @@ import (
 )
 
 var (
+	// hardCodeTopicNameRe is used to match a topic name which is hard code in the config
+	hardCodeTopicNameRe = regexp.MustCompile(`^([A-Za-z0-9\._\-]+)$`)
```
nongfushanquan, line 25: why use another expression since this case is a subset of the next one ?

cdc/sink/dmlsink/mq/dispatcher/topic/expression.go
3AceShowHand: what do you mean by the `empty` case ?

3AceShowHand: Yes it also covers the empty case. But if user enter "" as the topic, the default topic is used directly, won't hit this logic.

nongfushanquan (approved):

asddongmen (approved):

PR #9813, opened 2023-09-26
mysql (ticdc): fix batch dml delete sql gennerate error

### LINKED ISSUE #9812: TiCDC did not correctly use the primary key as the where clause for the Delete statement.
### What did you do?

I use TiCDC to synchronize a table with one primary key (consisting of two columns of int values) and a non-null unique index (consisting of 3 columns). I observed from the logs that the where statement of the Delete statement generated by TiCDC did not correctly use the primary key as an execution condition.

TiCDC uses primary key to generate the where clause of delete statements.

### What did you see instead?

as above.

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

asddongmen:

Issue Number: close #9812 

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/entry/mounter_test.go
+++ b/cdc/entry/mounter_test.go
@@ -1518,13 +1518,11 @@ func TestDecodeEventIgnoreRow(t *testing.T) {
 
 func TestBuildTableInfo(t *testing.T) {
 	cases := []struct {
-		caseNumber          int
 		origin              string
 		recovered           string
 		recoveredWithNilCol string
 	}{
 		{
-			1,
 			"CREATE TABLE t1 (c INT PRIMARY KEY)",
 			"CREATE TABLE `BuildTiDBTableInfo` (\n" +
 				"  `c` int(0) NOT NULL,\n" +
@@ -1536,7 +1534,6 @@ func TestBuildTableInfo(t *testing.T) {
 				") ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin",
 		},
 		{
-			2,
 			"CREATE TABLE t1 (" +
 				" c INT UNSIGNED," +
 				" c2 VARCHAR(10) NOT NULL," +
@@ -1558,7 +1555,6 @@ func TestBuildTableInfo(t *testing.T) {
 				") ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin",
 		},
 		{
-			3,
 			"CREATE TABLE t1 (" +
 				" c INT UNSIGNED," +
 				" gen INT AS (c+1) VIRTUAL," +
@@ -1584,7 +1580,6 @@ func TestBuildTableInfo(t *testing.T) {
 				") ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin",
 		},
 		{
-			4,
 			"CREATE TABLE `t1` (" +
 				"  `a` int(11) NOT NULL," +
 				"  `b` int(11) DEFAULT NULL," +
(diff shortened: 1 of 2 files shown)


cdc/sink/dmlsink/txn/mysql/mysql.go
```diff
@@ -567,8 +567,15 @@ func (s *mysqlBackend) prepareDMLs() *preparedDMLs {
 			}
 			// only use batch dml when the table has a handle key
 			if hasHandleKey(tableColumns) {
-				// TODO(dongmen): find a better way to get table info.
-				tableInfo := model.BuildTiDBTableInfo(tableColumns, firstRow.IndexColumns)
+				var tableInfo *timodel.TableInfo
+				if firstRow.TableInfo != nil && firstRow.TableInfo.TableInfo != nil {
+					tableInfo = firstRow.TableInfo.TableInfo
+				} else {
+					// firstRow.TableInfo is nil only when the row change event was generated by
```
asddongmen, line 574:
```suggestion
					// firstRow.TableInfo is nil only when the row change event was **not** generated by
```

nongfushanquan (approved):

cdc/sink/dmlsink/txn/mysql/mysql.go
```diff
@@ -567,8 +567,15 @@ func (s *mysqlBackend) prepareDMLs() *preparedDMLs {
...
-				// TODO(dongmen): find a better way to get table info.
-				tableInfo := model.BuildTiDBTableInfo(tableColumns, firstRow.IndexColumns)
+				var tableInfo *timodel.TableInfo
+				if firstRow.TableInfo != nil && firstRow.TableInfo.TableInfo != nil {
+					tableInfo = firstRow.TableInfo.TableInfo
+				} else {
+					// firstRow.TableInfo is nil only when the row change event was **not** generated by
+					// mounter, which only occurs when mysql sink was used in redo consumer or
+					// in the other consumers.
+					tableInfo = model.BuildTiDBTableInfo(tableColumns, firstRow.IndexColumns)
```
sdojjy, line 577: `BuildTiDBTableInfo` generates incorrect index information, it's better to fix it or add a comment

cdc/entry/mounter_test.go
```diff
@@ -1615,7 +1653,18 @@ func TestBuildTableInfo(t *testing.T) {
 		handle := sqlmodel.GetWhereHandle(recoveredTI, recoveredTI)
 		require.NotNil(t, handle.UniqueNotNullIdx)
 		require.Equal(t, c.recovered, showCreateTable(t, recoveredTI))
-
+		// make sure BuildTiDBTableInfo indentify the correct primary key
+		if c.caseNumber == 5 {
```
lance6716, line 1657: caseNumber can be replaced by `i` of `for i, c := range cases {`

cdc/model/sink.go
```diff
@@ -604,7 +604,10 @@ func BuildTiDBTableInfo(columns []*Column, indexColumns [][]int) *model.TableInf
 			continue
 		}
 		if firstCol.Flag.IsPrimaryKey() {
```
lance6716, line 606: maybe we can postpone the recover of PK in line 617's loop

lance6716 (commented):
rest lgtm

cdc/model/sink.go
asddongmen: Thank you for your advice!

cdc/entry/mounter_test.go
asddongmen: sure

lance6716 (approved):

PR #9852, opened 2023-10-10
sink(cdc): fix the bug that mq sink can lost callbacks

### LINKED ISSUE #9855: TiCDC MQ sinks can lost callbacks
### What did you do?

TiCDC MQ sinks can lost callbacks, which makes checkpoint can be stuck forever.

_No response_

### What did you see instead?

TiCDC MQ sinks can lost callbacks, which makes checkpoint can be stuck forever.

And, it's the root cause of #9542 .

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
master
```

hicqu:

Issue Number: close #9855 

When errors happen in `MQSink.WriteEvents`, the backend sink should be canceled correctly. Otherwise the table can lost callbacks forever.

#### Tests 

 - Integration test

### Release note 

```release-note
None
```


3AceShowHand (approved):

cdc/sink/dmlsink/mq/mq_dml_sink.go
```diff
@@ -144,10 +147,16 @@ func (s *dmlSink) WriteEvents(txns ...*dmlsink.CallbackableEvent[*model.SingleTa
...
+
 		for _, row := range txn.Event.Rows {
 			topic := s.alive.eventRouter.GetTopicForRowChange(row)
 			partitionNum, err := s.alive.topicManager.GetPartitionNum(s.ctx, topic)
+			failpoint.Inject("MQSinkGetPartitionError", func() {
+				log.Info("failpoint MQSinkGetPartitionError injected", zap.String("changefeedID", s.id.ID))
+				err = errors.New("MQSinkGetPartitionError")
+			})
 			if err != nil {
+				s.cancel(err)
```
CharlesCheung96, line 159: Why could this cancel prevent callback from being lost?

CharlesCheung96: In addition, when WriteEvents returns error, the external caller should also close dml_sink?

hicqu: Here we cancel the sink backend. Then when checking whether the table sink has been closed or not, we can know the sink backend is dead, and then we don't need to be blocked on table sink callbacks.

CharlesCheung96 (approved):

PR #9856, opened 2023-10-10
ddl: support report error when replicate incompatible ddl

### LINKED ISSUE #9850: dm: support report error when replicate incompatible ddl

Some customers have experienced issues with upstream tools like pt-osc, which truncate the range of values during DDL changes like modifying Varchar(64) to Varchar(32), resulting in data loss. While these DDL changes might only produce warnings, there is a concern that they could execute successfully in TiDB under certain SQL modes and potentially cause downstream failures. It is advisable for DM to intercept and handle these DDL changes proactively. This approach also prevents prolonged scanning of large tables where eventual errors might occur due to data length mismatches.

```
filters:
  rule-1:
    schema-pattern: "*"   # The name of the database for which this binlog event filter should be applied. Use "*" to indicate all databases, as needed by the user.
    table-pattern: "*"    # The name of the table for which this binlog event filter should be applied. Use "*" to indicate all tables, as needed by the user.
    events: ["incompatible_ddl_changes"]  # The binlog events to be parsed. Set to "incompatible_ddl_changes" to represent DDL changes that cause incompatibility and data inconsistency.
    action: error  # The action to be taken when this DDL is encountered. "error" indicates an error should be raised, and synchronization should stop. Other possible values are "ignore" and "do". The default value is "do".

  rule-2:
    schema-pattern: "test1"   # The name of the database for which this binlog event filter should be applied, set to "test1".
    table-pattern: "t*"    # The name of the table for which this binlog event filter should be applied, for tables starting with "t".
    events: ["Modify Column", "Value Range Decrease"]  # Represents the event types to be parsed and blocked. Here, it indicates modifying column types and reducing the value range.
    action: error  # The action to be taken when this DDL is encountered. "error" indicates an error should be raised, and synchronization should stop. Other possible values are "ignore" and "do". The default value is "do".

```

_No response_

_No response_

GMHDBJD:

Issue Number: close #9850 

- judge incompatible ddl changes for alter table

#### Tests 

 - Integration test

### Release note 

```release-note

```


lichunzhu (approved):
LGTM

dm/syncer/ddl.go
```diff
@@ -955,14 +965,228 @@ func parseOneStmt(qec *queryEventContext) (stmt ast.StmtNode, err error) {
...
+			return bf.AlterTable
+		case mysql.TypeString:
+			// Due to the behavior of padding \x00 at binary type, always change column data when binary length changed
+			if types.IsBinaryStr(&oldCol.FieldType) {
+				if newCol.GetFlen() != oldCol.GetFlen() {
+					return bf.PrecisionDecrease
+				}
+			}
+			return bf.AlterTable
+		}
```
okJiang, line 1011: These are all cases of inequality, not cases of reduced precision?

dm/tests/incompatible_ddl_changes/run.sh
```diff
@@ -0,0 +1,260 @@
...
+	dmctl_start_task_standalone $cur/conf/$1
+
+	run_sql_file $cur/data/db1.increment.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
+
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"query-status test" \
+		"\"result\": true" 2 \
+		"\"synced\": true" 1
+
+	# value range decrease
```
okJiang, line 31: can you add the original column type in the comment? It is convenient.

okJiang (commented):
rest lgtm

dm/tests/incompatible_ddl_changes/run.sh
GMHDBJD: you can see from c_mediumint smallint(6), the column name descripe that it's a mediumint column, also can see the revert ddl at L39, which revert it to mediumint(7)

okJiang (approved):

PR #9911, opened 2023-10-17
codec(ticdc): kafka simple encoding protocol support encoding checkpoint event.

### LINKED ISSUE #9900: encoding & decoding checkpoint event
as title

3AceShowHand:

Issue Number: close #9900 

#### Tests 

 - Unit test

### Release note 

```release-note
`None`
```

### DIFF
--- a/pkg/config/sink_protocol.go
+++ b/pkg/config/sink_protocol.go
@@ -38,6 +38,7 @@ const (
 	ProtocolCraft
 	ProtocolOpen
 	ProtocolCsv
+	ProtocolSimple
 )
 
 // IsBatchEncode returns whether the protocol is a batch encoder.
@@ -66,6 +67,8 @@ func ParseSinkProtocolFromString(protocol string) (Protocol, error) {
 		return ProtocolOpen, nil
 	case "csv":
 		return ProtocolCsv, nil
+	case "simple":
+		return ProtocolSimple, nil
 	default:
 		return ProtocolUnknown, cerror.ErrSinkUnknownProtocol.GenWithStackByArgs(protocol)
 	}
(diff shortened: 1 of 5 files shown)


asddongmen:
Please add a issue to track the whole feature and link every related PR to it.

asddongmen (approved):

pkg/config/sink_protocol.go
```diff
@@ -38,6 +38,7 @@ const (
 	ProtocolCraft
 	ProtocolOpen
 	ProtocolCsv
+	ProtocolSimple
```
zhangjinpeng1987, line 41: Why use "Simple" as the new protocol name? We will add claim-check large message feature for this protocol, we will add water mark, checkpoint, schema tracking & bootstrap and more other features, can we use a proper name that can represent its functionalities? How about SufficientJson since it is json, but not each message contains the full schema information and is more sufficient than CanalJson?

zhangjinpeng1987: Or ExtensibleJson?

3AceShowHand: The name is not decided yet, use `simple` temporarily.

`Json` should not be part of the protocol, since this new protocol does not only support JSON format, we may make it support other encoding formats such as avro, cbor.

Sufficient sounds good to me. Extensible is also a good name, but it's not easy to explain the `customized`, since this new protocol belongs to the TiCDC itself, all things is customized.

3AceShowHand: For the concrete name, further discussion is required.

nongfushanquan (approved):

PR #9936, opened 2023-10-20
mysql(ticdc): use string join to build multi statement sql.

### LINKED ISSUE #9935: mysql sink allocate a huge number of memory when build multi statement sql.
Screenshot 2023-10-20 at 15 21 58

From the above image, we can see that the MySQL sink allocates a huge number of memory when trying to build the multi-statement SQL, this can be reduced by using a more efficient string concatenation method.

3AceShowHand:

Issue Number: close #9935 

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```

### DIFF
--- a/cdc/sink/dmlsink/txn/mysql/mysql.go
+++ b/cdc/sink/dmlsink/txn/mysql/mysql.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"math"
 	"net/url"
+	"strings"
 	"time"
 
 	dmysql "github.com/go-sql-driver/mysql"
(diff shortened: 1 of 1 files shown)


cdc/sink/dmlsink/txn/mysql/mysql.go
```diff
@@ -652,15 +653,29 @@ func (s *mysqlBackend) multiStmtExecute(
...
-		multiStmtSQL += query
+		size += len(query)
 		if i != len(dmls.sqls)-1 {
-			multiStmtSQL += ";"
+			size += 1
+		}
+	}
+	sb.Grow(size)
+
+	for i, query := range dmls.sqls {
```
hicqu, line 670: Why not use `strings.Join`?

asddongmen (approved):

cdc/sink/dmlsink/txn/mysql/mysql.go
3AceShowHand: it looks that `strings.Join` also use the string builder, it can be replaced by using it, let me fix it.

3AceShowHand: It looks that the `strings.Join` does not consider the `separator` when calculate the string builder capacity before the `Grow` method, it may cause reallocation.

nongfushanquan (approved):

cdc/sink/dmlsink/txn/mysql/mysql.go
3AceShowHand: Yes, `strings.Join` can be used in this case.

PR #10037, opened 2023-11-08
kafka(ticdc): claim check do not use filepath join to build the full file path.

### LINKED ISSUE #10036: claim check file full path is not correct since duplicate slash is clean up
`filepath.join` is used to build the file full path, it makes the `s3://` to `s3:/`, this is not expected.

3AceShowHand:

Issue Number: close #10036

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```

### DIFF
--- a/pkg/sink/kafka/claimcheck/claim_check.go
+++ b/pkg/sink/kafka/claimcheck/claim_check.go
@@ -16,6 +16,7 @@ package claimcheck
 import (
 	"context"
 	"encoding/json"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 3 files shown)


sdojjy (approved):

pkg/sink/kafka/claimcheck/claim_check.go
```diff
@@ -102,7 +101,7 @@ func (c *ClaimCheck) WriteMessage(ctx context.Context, key, value []byte, fileNa
 
 // FileNameWithPrefix returns the file name with prefix, the full path.
 func (c *ClaimCheck) FileNameWithPrefix(fileName string) string {
-	return filepath.Join(c.storage.URI(), fileName)
+	return c.storage.URI() + "/" + fileName
```
CharlesCheung96, line 104: Please add unit test to cover this function.

nongfushanquan: if the configure end with "/", whether it is still work?

hicqu (approved):

CharlesCheung96 (approved):

Issue #9419, opened 2023-07-21
Always upload schema after creating changefeed


Lloyd-Pottiger:
### Is your feature request related to a problem?

Now, when we execute a DML after changefeed was created, cdc storage sink will upload current schema to external storage.

But, when we execute a DDL at once after changefeed was created, cdc storage sink only upload the schema after ddl. So we can not get the original schema from external storage.

### Describe the feature you'd like

CDC storage sink always upload schema after creating changefeed.

PR #10045, opened 2023-11-09
ddlManager (ticdc): add bootstrap sending function

### LINKED ISSUE #10107: DDL manager sends bootstrap events

as title.

as title.

_No response_

_No response_

asddongmen:

Issue Number: ref #9898 close #10107

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```


cdc/owner/ddl_sink.go
```diff
@@ -54,6 +54,7 @@ type DDLSink interface {
 	// the DDL event will be sent to another goroutine and execute to downstream
 	// the caller of this function can call again and again until a true returned
 	emitDDLEvent(ctx context.Context, ddl *model.DDLEvent) (bool, error)
+	emitBootstrapEvent(ctx context.Context, ddl *model.DDLEvent) error
```
3AceShowHand, line 57: can we use the `emitDDLEvent` directly ?

asddongmen: I have tried, but it will bring some complexity into the `emitDDLEvent` interface.

cdc/owner/ddl_manager.go
```diff
@@ -184,6 +198,18 @@ func (m *ddlManager) tick(
 	checkpointTs model.Ts,
 	tableCheckpoint map[model.TableName]model.Ts,
 ) ([]model.TableID, *schedulepb.BarrierWithMinTs, error) {
+	// needBootstrap is true when the the downstream is kafka
+	// and the protocol is simple protocol.
+	if m.needBootstrap {
```
CharlesCheung96, line 203: it is more appropriate to make it configurable, since other scenarios require this feature as well. Ref: https://github.com/pingcap/tiflow/issues/9419.

cdc/owner/ddl_sink.go
```diff
@@ -384,6 +385,21 @@ func (s *ddlSinkImpl) emitDDLEvent(ctx context.Context, ddl *model.DDLEvent) (bo
...
+func (s *ddlSinkImpl) emitBootstrapEvent(ctx context.Context, ddl *model.DDLEvent) error {
+	if !ddl.IsBootstrap {
+		return nil
+	}
+	err := s.sink.WriteDDLEvent(ctx, ddl)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	// TODO: change this log to debug level after testing complete.
+	log.Info("emit bootstrap event", zap.Any("bootstrapEvent", ddl))
```
CharlesCheung96, line 399: Please output changefeed and namaspace.

cdc/owner/ddl_sink.go
CharlesCheung96: It seems that `emitDDLEvent` will not be called until `emitBootstrapEvent` completes, and the only difference between them is whether or not the caller is blocked, so perhaps this behavior could be controlled by adding an argument to `emitDDLEvent`?

cdc/owner/ddl_sink.go
```diff
@@ -384,6 +385,21 @@ func (s *ddlSinkImpl) emitDDLEvent(ctx context.Context, ddl *model.DDLEvent) (bo
 	return false, nil
 }
 
+// emitBootstrapEvent sent bootstrap event to downstream.
+// It is a synchronous operation.
+func (s *ddlSinkImpl) emitBootstrapEvent(ctx context.Context, ddl *model.DDLEvent) error {
+	if !ddl.IsBootstrap {
+		return nil
+	}
+	err := s.sink.WriteDDLEvent(ctx, ddl)
```
CharlesCheung96, line 394: Does addSpecialComment need to be called before writing ddl?

cdc/owner/ddl_manager.go
```diff
@@ -184,6 +198,18 @@ func (m *ddlManager) tick(
 	checkpointTs model.Ts,
 	tableCheckpoint map[model.TableName]model.Ts,
 ) ([]model.TableID, *schedulepb.BarrierWithMinTs, error) {
+	// needBootstrap is true when the the downstream is kafka
```
3AceShowHand, line 201:
```suggestion
	// needBootstrap is true when the downstream is kafka
```

cdc/owner/ddl_manager.go
```diff
@@ -184,6 +198,18 @@ func (m *ddlManager) tick(
...
+	// and the protocol is simple protocol.
+	if m.needBootstrap {
+		ok, err := m.checkAndBootstrap(ctx)
```
3AceShowHand, line 204: It looks that the `checkAndBootstrap` is called on every tick if the `needBootstrap` is true.

can we simplify this?

cdc/owner/ddl_manager.go
```diff
@@ -598,6 +624,53 @@ func (m *ddlManager) cleanCache() {
...
+	if atomic.LoadInt32(&m.bootstrapState) == int32(bootstrapStateRunning) {
+		return false, nil
+	}
+	// begin bootstrap
+	atomic.StoreInt32(&m.bootstrapState, int32(bootstrapStateRunning))
+	tables, err := m.allTables(ctx)
+	if err != nil {
+		return false, err
+	}
+	var bootstrapEvents []*model.DDLEvent
```
3AceShowHand, line 647: the length of the `bootstrapEvents` is known, initialized the slice with length to avoid reallocation here.

cdc/owner/ddl_manager.go
```diff
@@ -598,6 +624,53 @@ func (m *ddlManager) cleanCache() {
...
+		for _, event := range bootstrapEvents {
+			err := m.ddlSink.emitBootstrapEvent(ctx, event)
+			if err != nil {
+				log.Error("emit bootstrap event failed",
+					zap.Any("bootstrapEvent", event), zap.Error(err))
+				atomic.StoreInt32(&m.bootstrapState, int32(bootstrapStateNone))
+				m.errCh <- err
+				return
+			}
+			atomic.StoreInt32(&m.bootstrapState, int32(bootstrapStateCompleted))
```
3AceShowHand, line 668: When the bootstrap state should be completed? 

For one bootstrap event is sent or all the bootstrap events are sent ?

3AceShowHand: Shall we move L668 outside of the loop ?

cdc/owner/ddl_manager.go
```diff
@@ -598,6 +624,53 @@ func (m *ddlManager) cleanCache() {
...
+		ddlEvent := &model.DDLEvent{
+			StartTs:     m.startTs,
+			CommitTs:    m.startTs,
+			TableInfo:   table,
+			IsBootstrap: true,
+		}
+		bootstrapEvents = append(bootstrapEvents, ddlEvent)
+	}
+	// send bootstrap events
+	go func() {
```
3AceShowHand, line 658: Why do we have to send bootstrap events in another goroutine?

cdc/owner/ddl_manager.go
asddongmen: It is configured by a parameters when the ddl_manager was created.

cdc/owner/ddl_sink.go
asddongmen: It shouldn't, because it is not a ddl actually.

cdc/owner/ddl_manager.go
asddongmen: After bootstrap complete, the `checkAndBootstrap ` will return immediately when it is called.

cdc/owner/ddl_manager.go
asddongmen: Otherwise it will block the whole `Owner.Tick()` loop.

cdc/owner/ddl_manager.go
3AceShowHand: It looks it's set after the first one bootstrap event is sent at the moment.

cdc/owner/ddl_manager.go
3AceShowHand: It looks cannot guarantee that all bootstrap events are sent before other event ?

asddongmen: No, it can.
If `!bootstrapStateCompleted`, the ddlBarrierTs would not advance by ddlManager, so other event will not be sent.

cdc/owner/ddl_sink.go
asddongmen: I also think it is better to separate this two function.

cdc/owner/ddl_manager.go
```diff
@@ -184,6 +198,18 @@ func (m *ddlManager) tick(
 	checkpointTs model.Ts,
 	tableCheckpoint map[model.TableName]model.Ts,
 ) ([]model.TableID, *schedulepb.BarrierWithMinTs, error) {
+	// needBootstrap is true when the downstream is kafka
+	// and the protocol is simple protocol.
```
3AceShowHand, line 202: I think this comment is redundant.

You can move it to the declaration of the `needBootstrap` variable

cdc/owner/ddl_manager.go
3AceShowHand: > Otherwise it will block the whole `Owner.Tick()` loop.

Will the send DDL event block the owner?

IMHO, sending Bootstrap event is just like sending DDL event.

3AceShowHand (approved):

cdc/owner/ddl_manager.go
asddongmen: After next PR merged, it will be remove.

CharlesCheung96 (approved):

PR #10049, opened 2023-11-09
config(ticdc): change case-sensitive default value

### LINKED ISSUE #10047: change case-sensitive parameter default value to false
now the default value of changefeed case-sensitive parameter is true, which means it will filter tables in the sensitive model

but cdc always get a lowercase db name from tikv, so in some case, some DDL will be discarded by mistake.
for example
```
case-sensitive=true
[filter]
rules=["DB1.*"]
```
ddls
```sql
USE DB1;
create table t1 (id int primary key);
alter table t1 rename to t2;
alter table t2 rename to t1;
```
the last ddl  `alter table t2 rename to t1;` will be ignored

sdojjy:

Issue Number: close #10047

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
 `None`.
```


nongfushanquan (approved):

hicqu (approved):

zhangjinpeng1987 (approved):
LGTM

Issue #10061, opened 2023-11-09
DM optimistic locks didn't show correct ddl locks info


lichunzhu:
### What did you do?

Use DM to replicate some DDLs in optimistic shard ddl mode.

task.yaml:
```yaml
---
name: test
task-mode: all
is-sharding: true
shard-mode: "optimistic"
meta-schema: "dm_meta"

mysql-instances:
  - source-id: "mysql-replica-01"
    block-allow-list:  "instance"
    route-rules: ["sharding-table-rules","sharding-schema-rules"]
    mydumper-config-name: "global"
    loader-config-name: "global"
    syncer-config-name: "global"
  - source-id: "mysql-replica-02"
    block-allow-list:  "instance"
    route-rules: ["sharding-table-rules","sharding-schema-rules"]
    mydumper-config-name: "global"
    loader-config-name: "global"
    syncer-config-name: "global"

block-allow-list:
  instance:
    do-dbs: ["shardddl1","shardddl2"]

routes:
  sharding-table-rules:
    schema-pattern: "shardddl*"
    target-schema: "shardddl"
    table-pattern: "tb*"
    target-table: "tb"
  sharding-schema-rules:
    schema-pattern: "shardddl*"
    target-schema: "shardddl"
```

ddl executed on "mysql-replica-01":
```sql
create table shardddl1.tb1 (a int primary key, id datetime);
```

ddl executed on "mysql-replica-02":
```sql
create table shardddl1.tb1 (a int primary key, id datetime);
create table shardddl1.tb2 (a int primary key, id datetime);
alter table shardddl1.tb1 modify column id binary;
create table shardddl1.test (a int primary key, col1 int);
create table shardddl2.test (a int primary key, col1 int);
alter table shardddl1.test modify column col1 varchar(11);
alter table shardddl2.test modify column col1 binary;
```

### What did you expect to see?

DM can show conflicts of table `shardddl`.`test`. For `shardddl`.`tb`, all tables should be unsynced.

### What did you see instead?

» shard-ddl-lock
{
    "result": true,
    "msg": "",
    "locks": [
        {
            "ID": "test-`shardddl`.`test`",
            "task": "test",
            "mode": "optimistic",
            "owner": "mysql-replica-02-`shardddl2`.`test`",
            "DDLs": [
                "ALTER TABLE `shardddl`.`test` MODIFY COLUMN `col1` BINARY"
            ],
            "synced": [
            ],
            "unsynced": [
[...]

PR #10062, opened 2023-11-09
DM: fix wrong shard-ddl-lock info

lichunzhu:

Issue Number: close https://github.com/pingcap/tiflow/issues/10061

add lock id check before show ddl locks.

#### Tests 

 - Integration test

No
No
### Release note 

```release-note
None
```


GMHDBJD (approved):
LGTM

Issue #10167, opened 2023-11-28
[sink to cloud storage] output old values to CSV files


zhangjinpeng1987:
### Is your feature request related to a problem?

In the repairing upstream data scenario (described in https://github.com/pingcap/tiflow/issues/10100 ), if the mistake operation is update, the sink-to-cloudstorage CSV protocol can not generate the undo DML because for update operations the CSV protocol just record the after-update value.

### Describe the feature you'd like

Record the old value for update operations(delete recorded old value already) when `oldvalue` parameter is true `sink-uri="s3://xxx/xxx?protocol=csv&oldvalue=true"`. For example, there are following update operations:
```
1) update row1 = b
2) update row1 = c
```
Current CSV protocol will generate CSV file looks like:
```
U, db, table, [commit-ts1], rowid=1, value=b
U, db, table, [commit-ts2], rowid=1, value=c
```
There is no way to generate undo DML for these update because they lack of old value. My proposal is add an option to support using delete and insert to replace update operation for the CSV protocol, and the CSV file looks like:
```
D, db, table, [commit-ts1], rowid=1, value=a
I, db, table, [commit-ts1], rowid=1, value=b
D, db, table, [commit-ts2], rowid=1, value=b
I, db, table, [commit-ts2], rowid=1, value=c
```
The advantage of this proposal is we can record the old value for Insert(I), Update(U) and Delete(D) operations. The disadvantage of this proposal is the file size grows if there are many update operations.
And when generate the undo DML, it can generate following DMLs:
```
Delete * from db.table where rowid=1 and value=c;
Insert into db.table values(1, b);
Delete * from db.table where rowid=1 and value=b;
Insert into db.table values(1, a);
```

### Describe alternatives you've considered

Using two update rows to record old value and new value, and the generated CSV file for above example should looks like:
```
U, db, table, [commit-ts1], rowid=1, value=a
U, db, table, [commit-ts1], rowid=1, value=b
U, db, table, [commit-ts2], rowid=1, value=b
U, db, table, [commit-ts2], rowid=1, value=c
```
But the semantic is not clear when generating undo DML.

### Teachability, Documentation, Adoption, Migration Strategy

todo

flowbehappy:
Another option is using "U-" or "U+", instead of "I" and "D", to represent UPDATE. cc @Benjamin2037

benmeadowcroft:
[...]

PR #10174, opened 2023-11-29
sink to csv: output old value to CSV files

zhangjinpeng1987:

Issue Number: close https://github.com/pingcap/tiflow/issues/10167

When changefeed configuration set:
```
[sink.csv]
output-old-value = true
```
As https://github.com/pingcap/tiflow/issues/10167 described, sink to csv replace update with a pair of delete and insert, in this way to record the old value for UPDATE statement, so it is possible to generate undo DML for some data repairing cases.

Before this change, the CSV output statement of `update schema.table set name="def" where id=1 and name="abc"` is
```
U, table, schema, [commit-ts], 1, def
```
After this change, the CSV output looks like:
```
D, table, schema, [commit-ts], 1, abc
I, table, schema, [commit-ts], 1, def
```

#### Tests 

 - Unit test

After set `[sink.csv] output-old-value = true` configuration for the sink to storage changefeed, for each UPDATE statement there will be 2 rows of data in the output CSV file. The output CSV files size is larger than before if there are many UPDATE statements.

### Release note 

```release-note
TiCDC sink to CSV files: support output old values for csv files, this is helpful for data repairing scenario.
```

### DIFF
--- a/cdc/api/v2/model.go
+++ b/cdc/api/v2/model.go
@@ -919,7 +919,7 @@ type CSVConfig struct {
 	NullString           string `json:"null"`
 	IncludeCommitTs      bool   `json:"include_commit_ts"`
 	BinaryEncodingMethod string `json:"binary_encoding_method"`
-	EnableOldValue       bool   `json:"enable_old_value"`
+	OutputOldValue       bool   `json:"output_old_value"`
 }
 
 // LargeMessageHandleConfig denotes the large message handling config
--- a/cdc/api/v2/model.go
+++ b/cdc/api/v2/model.go
@@ -919,6 +919,7 @@ type CSVConfig struct {
 	NullString           string `json:"null"`
 	IncludeCommitTs      bool   `json:"include_commit_ts"`
 	BinaryEncodingMethod string `json:"binary_encoding_method"`
+	EnableOldValue       bool   `json:"enable_old_value"`
 }
 
 // LargeMessageHandleConfig denotes the large message handling config
--- a/pkg/cmd/util/changefeed_storage_sink.toml
+++ b/pkg/cmd/util/changefeed_storage_sink.toml
@@ -15,3 +15,6 @@ quote = '"'
 null = '\N'
 # Include commit-ts in the row data. The default value is false.
 include-commit-ts = false
+# Enable Old Value, update statement will be replaced by a pair of delete and insert
+# in the CSV files.
+enable-old-value = false
--- a/pkg/cmd/util/changefeed_storage_sink.toml
+++ b/pkg/cmd/util/changefeed_storage_sink.toml
@@ -15,6 +15,6 @@ quote = '"'
 null = '\N'
 # Include commit-ts in the row data. The default value is false.
 include-commit-ts = false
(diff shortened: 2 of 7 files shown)


cdc/api/v2/model.go
```diff
@@ -919,6 +919,7 @@ type CSVConfig struct {
 	NullString           string `json:"null"`
 	IncludeCommitTs      bool   `json:"include_commit_ts"`
 	BinaryEncodingMethod string `json:"binary_encoding_method"`
+	EnableOldValue       bool   `json:"enable_old_value"`
```
CharlesCheung96, line 922: `enable_old_value` has different meanings on different versions of ticdc and has been confusing for users, so a more explicit parameter name such as `output_old_value` is recommended here.

3AceShowHand: `force-split-update` is recommended.

3AceShowHand:
There is no need to update the CSV implementation.

It's recommended to split the update event https://github.com/pingcap/tiflow/blob/master/cdc/model/sink.go#L833

Add one new variable `forceSplitUpdate` to the `shouldSplitUpdateEvent` method should solve this problem.

cdc/api/v2/model.go
zhangjinpeng1987: output-old-value is more obvious from the semantic perspective, because split-update is just one specific implementation case of output old value in the context of sink to csv files.

3AceShowHand:
> output-old-value is more obvious from the semantic perspective, because split-update is just one specific implementation case of output old value in the context of sink to csv files.

IMHO, TiCDC always fetch old value from the TiKV, and also output old value if the output format allows to do so. In this aspect, It's ok to name it as `output-old-value` and only bound to the CSV format.

It's ok to support the `output-old-value` by modifying CSV encoder. IMHO, it would be easy to implement the `output-old-value` in the https://github.com/pingcap/tiflow/blob/master/cdc/model/sink.go#L833 by splitting the update into a delete and insert, this keeps the implementation of CSV format simple and clean.

zhangjinpeng1987:
> > output-old-value is more obvious from the semantic perspective, because split-update is just one specific implementation case of output old value in the context of sink to csv files.
> 
> IMHO, TiCDC always fetch old value from the TiKV, and also output old value if the output format allows to do so. In this aspect, It's ok to name it as `output-old-value` and only bound to the CSV format.
> 
> It's ok to support the `output-old-value` by modifying CSV encoder. IMHO, it would be easy to implement the `output-old-value` in the https://github.com/pingcap/tiflow/blob/master/cdc/model/sink.go#L833 by splitting the update into a delete and insert, this keeps the implementation of CSV format simple and clean.

@3AceShowHand 
IMO, we are not going to choose an implementation just because it is easy, more consideration is focus on where most other CSV changefeed configuration take affects, and split update is not equal to the semantic of `output old value` for a protocol. In this way, we can make the semantic more clear and easy to change/maintain in the future.

@3AceShowHand @CharlesCheung96 Thanks for your review, please let me know if you have other concerns about this PR.

3AceShowHand (approved):

CharlesCheung96 (approved):

PR #10203, opened 2023-11-30
Support Sequences

### LINKED ISSUE #4559: Sequence support is incomplete

Sequence support is incomplete in TiCDC.

The current capability of TiCDC:  

> The [DDL operation CREATE SEQUENCE](https://docs.pingcap.com/tidb/stable/sql-statement-create-sequence) and the [SEQUENCE function](https://docs.pingcap.com/tidb/stable/sql-statement-create-sequence#sequence-function) in TiDB. When the upstream TiDB uses SEQUENCE, TiCDC ignores SEQUENCE DDL operations/functions performed upstream. However, DML operations using SEQUENCE functions can be correctly replicated.
> 
> 2022/02/10, https://docs.pingcap.com/tidb/stable/ticdc-overview#unsupported-scenarios

Besides, there are two flaws we found,
1. TiCDC have no idea about how to deal with DDLs below:
    ```SQL
    CREATE SEQUENCE `SEQ1` start with 1 minvalue 1 maxvalue 999999999999999 increment by 1 nocache cycle;
    CREATE TABLE t(k int not null, a int default next value for SEQ1, PRIMARY KEY (k));
    ```
2. TiCDC filters all SEQUENCE DDL, it means we lost SEQUENCE even if downstream is TiDB. We lose data in TiDB disaster recovery scenario.

Support sequence feature completely when downstream is TiDB, and disable sequence completely when downstream is MySQL.

_No response_

_No response_

dveeden:

Issue Number: close #4559 

As MariaDB [supports Sequences](https://mariadb.com/kb/en/sequences/) it would be good if DM allows MariaDB→DM→TiDB for things like `CREATE SEQUENCE`.

#### Tests 

### Release note 

```release-note
Support for `CREATE SEQUENCE`, `ALTER SEQUENCE` and `DROP SEQUENCE` was added. 
```


dveeden:
Things that might need work:
- What would be a good way to add unittests or integration tests for this?
- There is a differences for when calling `DROP TABLE` on a sequence. Should we re-write that in DM?

asddongmen (approved):

lyzx2001 (approved):

lance6716:
Strange that I didn't find ALTER SEQUENCE support in TiDB document https://docs.pingcap.com/tidb/stable/sql-statement-create-sequence . So I think at least we should not allow replicating this type of DDL.

And it will cause a behaviour change, so let PM start a review @frank945946

dveeden:
> Strange that I didn't find ALTER SEQUENCE support in TiDB document https://docs.pingcap.com/tidb/stable/sql-statement-create-sequence . So I think at least we should not allow replicating this type of DDL.
> 
> And it will cause a behaviour change, so let PM start a review @Frank945946

This is supported:
```
sql> CREATE SEQUENCE s1 MINVALUE=2;
Query OK, 0 rows affected (0.1489 sec)

sql> ALTER SEQUENCE s1 MINVALUE=3;
ERROR: 4136 (HY000): Sequence 'test.s1' values are conflicting

sql> ALTER SEQUENCE s1 MINVALUE=1;
Query OK, 0 rows affected (0.1619 sec)

sql> SELECT TIDB_VERSION()\G
TIDB_VERSION(): Release Version: v7.6.0
Edition: Community
Git Commit Hash: 52794d985ba6325d75a714d4eaa0838d59425eb6
Git Branch: heads/refs/tags/v7.6.0
UTC Build Time: 2024-01-22 14:20:42
GoVersion: go1.21.5
Race Enabled: false
Store: tikv
1 row in set (0.0007 sec)
```

AilinKid (approved):

Frank945946:
As create, alter, and drop operations on sequences have already been supported in TiDB, LGTM supports migrating the sequence event from MariaDB to TiDB by DM. @dveeden @lance6716

lance6716 (approved):

PR #10205, opened 2023-11-30
syncer(dm): Improve logging of ignored TableMap/FormatDesc

### LINKED ISSUE #10204: Logging of FormatDesc and TableMap doesn't need to be a warning

DM logs a `unhandled event` for events it doesn't recognize. This includes TableMap and FormatDescriptionEvent. 

Lower log level for these expected events.

_No response_

_No response_

dveeden:

Issue Number: close #10204 

This makes the logging on the warning level less noisy and adds detail to the TableMap event logs.

#### Tests 

### Release note 

```release-note
Logging of `TableMapEvent` and `FormatDescriptionEvent` was changed from the warning level to the info level.
```

### DIFF
--- a/dm/syncer/syncer.go
+++ b/dm/syncer/syncer.go
@@ -2442,6 +2442,13 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 			if needContinue {
 				continue
 			}
+		case *replication.TableMapEvent:
+			s.tctx.L().Info("unhandled TableMap event",
+				zap.String("schema", string(ev.Schema)),
+				zap.String("table", string(ev.Table)),
+			)
+		case *replication.FormatDescriptionEvent:
+			s.tctx.L().Info("unhandled FormatDescription event")
 		default:
 			s.tctx.L().Warn("unhandled event", zap.String("type", fmt.Sprintf("%T", ev)))
 		}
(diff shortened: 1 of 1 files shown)


dm/syncer/syncer.go
```diff
@@ -2442,6 +2442,13 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 			if needContinue {
 				continue
 			}
+		case *replication.TableMapEvent:
+			s.tctx.L().Info("unhandled TableMap event",
+				zap.String("schema", string(ev.Schema)),
+				zap.String("table", string(ev.Table)),
+			)
```
lance6716, line 2449:
```suggestion
```

The "unhandled event" is used to detect some new events of future MySQL/MariaDB versions (I guess). So for these old type of events we don't need to log them, they are handled as no-op.

BTW we may also need to change line 2438

dveeden: @lance6716 The suggestion in your comment doesn't seem to change anything. Did you intent to suggest a code change?

I made changes based on your suggestion in ecc9985a47d63ed82d4bd059af4908ca9ba122ca

lance6716: I mean we don't need to print log for these two types of events. Just move them out of `default` case

dveeden: Done in 5773c89d79b2ee8598414018ce1a11f975012bb0

okJiang (approved):

PR #10351, opened 2023-12-23
sink(ticdc): use pd clock in storage sink

### LINKED ISSUE #10352: remove the local clock dependency from storage sink
Currently, the cloud storage sink uses local clock to generate files. If A table is scheduled from cdc node A to cdc node B, and ntp is not enabled on the cluster, or the time of the two machines is different, the generated file names may be backward or not incremental.

CharlesCheung96:

Issue Number: close #10352, ref #10374

- Use pdClock instead of local clock to generate date seperator.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
`Fix the issue that date seperator generated by the storage sink may roll back in table scheduling scenarios`.
```

### DIFF
--- a/cdc/api/v1/validator.go
+++ b/cdc/api/v1/validator.go
@@ -173,7 +173,8 @@ func verifyCreateChangefeedConfig(
 	}
 	if err := validator.Validate(ctx,
 		model.ChangeFeedID{Namespace: changefeedConfig.Namespace, ID: changefeedConfig.ID},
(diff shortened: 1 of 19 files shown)


zhangjinpeng1987:
It is ok for me if this is a quick and short term fix. But for long term we eventually will decouple the dependency with PD, we should have a better comprehensive solution later.

sdojjy (approved):

asddongmen (approved):

Issue #9969, opened 2023-10-30
[feature-request] sink to cloud storage: customized delimiter with 2 characters


zhangjinpeng1987:
### Is your feature request related to a problem?

TiCDC has a one character limitation for CSV delimiter, this will limit users using special characters like hex(FF, FE) as delimiter.

### Describe the feature you'd like

When user specify the delimiter, supports at most 2 characters delimiter.

### Describe alternatives you've considered

There is a hard limitation that the delimiter is 1 character, so there is no alternative to workaround this problem.

### Teachability, Documentation, Adoption, Migration Strategy

This is an extension for existing sink to CSV capability, this will give users better capability to customize their CSV format.

lidezhu:
Some customers use separators like `|@|` in their env. So we need to relief the restriction more to support 3 characters.

PR #10357, opened 2023-12-25
csv(cdc): delimiter supports at most 3 characters

lidezhu:

Issue Number: ref #9969 

Some customers use separator like `|@|` in their env. So we need to relief the delimiter restriction further to support at most 3 characters.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/config/sink.go
+++ b/pkg/config/sink.go
@@ -221,14 +221,14 @@ func (c *CSVConfig) validateAndAdjust() error {
 	case 0:
 		return cerror.WrapError(cerror.ErrSinkInvalidConfig,
 			errors.New("csv config delimiter cannot be empty"))
-	case 1, 2:
+	case 1, 2, 3:
 		if strings.ContainsRune(c.Delimiter, CR) || strings.ContainsRune(c.Delimiter, LF) {
 			return cerror.WrapError(cerror.ErrSinkInvalidConfig,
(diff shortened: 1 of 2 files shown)


lidezhu:
PTAL @CharlesCheung96 @nongfushanquan

pkg/config/sink.go
```diff
@@ -221,14 +221,14 @@ func (c *CSVConfig) validateAndAdjust() error {
 	case 0:
 		return cerror.WrapError(cerror.ErrSinkInvalidConfig,
 			errors.New("csv config delimiter cannot be empty"))
-	case 1, 2:
+	case 1, 2, 3:
```
CharlesCheung96, line 224: Perhaps we should add a delimiter size limit, which only allows a maximum of two 2-character delimiter by default, but users can turn on this configuration to use more delimiter

lidezhu: I guess it's not a good action to add a new configuration for this small behavior. Because it will increase unnecessary mental burden on the user.

flowbehappy: I don't suggest adding another config at this moment. Maybe we can consider adding a new config like "max-delimiter-length" if there is really necessary in the future.

CharlesCheung96 (approved):

asddongmen (approved):

PR #10366, opened 2023-12-27
encoder (ticdc): simple protocol send bootstrap event periodically

asddongmen:

Issue Number: ref #9898 

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


cdc/sink/dmlsink/mq/worker.go
```diff
@@ -297,6 +297,7 @@ func (w *worker) sendMessages(ctx context.Context) error {
 			if err = future.Ready(ctx); err != nil {
 				return errors.Trace(err)
 			}
+
```
3AceShowHand, line 300: remove this blank changes.

pkg/config/sink.go
```diff
@@ -81,6 +81,11 @@ const (
 
 	// DefaultEncoderGroupConcurrency is the default concurrency of encoder group.
 	DefaultEncoderGroupConcurrency = 32
+
+	// DefaultSendBootstrapIntervalInSec is the default interval to send bootstrap message.
```
3AceShowHand, line 85: Please mention that it's only works for the simple protocol at the moment.

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
+// Copyright 2022 PingCAP, Inc.
```
3AceShowHand, line 1:
```suggestion
// Copyright 2023 PingCAP, Inc.
```

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+	activeTables            sync.Map
+	builder                 RowEventEncoderBuilder
+	sendBootstrapInterval   time.Duration
+	sendBootstrapInMsgCount int32
+	// maxInactiveDuration is the max duration that a table can be inactive
+	maxInactiveDuration time.Duration
+	outCh               chan<- *future
+}
+
+// newBootstrapWorker creates a new bootstrapGenerator instance
```
3AceShowHand, line 46: what is `bootstrapGenerator`

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+		activeTables:            sync.Map{},
+		sendBootstrapInterval:   sendBootstrapInterval,
+		sendBootstrapInMsgCount: sendBootstrapInMsgCount,
+		maxInactiveDuration:     maxInactiveDuration,
+	}
+}
+
+func (b *bootstrapWorker) run(ctx context.Context) error {
+	sendTicker := time.NewTicker(bootstrapWorkerTickerInterval)
+	defer sendTicker.Stop()
```
3AceShowHand, line 68: this 2 defer calls can be unified together

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+				zap.String("topic", table.topic),
+				zap.Stringer("changefeed", b.changefeedID))
+			b.activeTables.Delete(key)
+		}
+		return true
+	})
+}
+
+// tableStatus is used to record the status of a table
+type tableStatus struct {
```
3AceShowHand, line 193: how about rename this to `tableStatistics` ?

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+		maxInactiveDuration:     maxInactiveDuration,
+	}
+}
+
+func (b *bootstrapWorker) run(ctx context.Context) error {
+	sendTicker := time.NewTicker(bootstrapWorkerTickerInterval)
+	defer sendTicker.Stop()
+	gcTicker := time.NewTicker(bootstrapWorkerGCInterval)
+	defer gcTicker.Stop()
+	errCh := make(chan error, 1)
```
3AceShowHand, line 71: it looks this `errCh` is redundant, the error can be returned immediately.

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	return nil
+}
+
+// sendBootstrapMsg sends a bootstrap message if the table meets the condition
+// 1. The time since last bootstrap message sent is larger than sendBootstrapInterval
+// 2. The received row event count since last bootstrap message sent is larger than sendBootstrapInMsgCount
+// Note: It is a blocking method, it will block if the outCh is full.
+func (b *bootstrapWorker) sendBootstrapMsg(ctx context.Context, table *tableStatus) error {
+	if !table.shouldSendBootstrapMsg(
```
3AceShowHand, line 121: I would prefer `shouldSendBootstrapMsg` become a method of `bootstrapWorker`, which takes `tableStatus` as the argument.

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+
+func (b *bootstrapWorker) generateEvents(
+	topic string,
+	totalPartition int32,
+	tableInfo *model.TableInfo,
+) ([]*future, error) {
+	res := make([]*future, 0, int(totalPartition))
+	// Bootstrap messages of a table should be sent to all partitions.
+	for i := 0; i < int(totalPartition); i++ {
+		encoder := b.builder.Build()
```
3AceShowHand, line 150: it looks that one `encoder` can be used multiple times, no need to create the encoder for each message, this allocates too many small memory

3AceShowHand: I think the bootstrap message should be sent to the first partition.

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+		msg, err := encoder.EncodeDDLEvent(&model.DDLEvent{
+			StartTs:     0,
+			CommitTs:    0,
```
3AceShowHand, line 153: `StartTs` and `CommitTs` should not be `0`, but the real value.

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+			TableInfo:   tableInfo,
+			IsBootstrap: true,
+		})
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+
+		key := TopicPartitionKey{
```
3AceShowHand, line 161: I still do not think it's a good idea to make `TopicPartitionKey` as a part of the `codec`

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+		t.totalPartition.Store(key.TotalPartition)
+	}
+
+	if t.version.Load() != row.TableInfo.UpdateTS {
+		t.version.Store(row.TableInfo.UpdateTS)
+		t.tableInfo.Store(row.TableInfo)
+	}
+}
+
+func (t *tableStatus) isActive(maxInactiveDuration time.Duration) bool {
```
3AceShowHand, line 255: I would prefer a method named `isInactive`

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+// Note: It is a blocking method, it will block if the outCh is full.
+func (b *bootstrapWorker) sendBootstrapMsg(ctx context.Context, table *tableStatus) error {
+	if !table.shouldSendBootstrapMsg(
+		b.sendBootstrapInterval,
+		b.sendBootstrapInMsgCount) {
+		return nil
+	}
+	table.reset()
+	tableInfo := table.tableInfo.Load().(*model.TableInfo)
+	events, err := b.generateEvents(table.topic, table.totalPartition.Load(), tableInfo)
```
3AceShowHand, line 128: It looks that the `error` only return not nil when `EncodeDDL` failed, this cause the changefeed failed.

pkg/sink/codec/encoder_group.go
```diff
@@ -56,11 +57,16 @@ type encoderGroup struct {
...
 	outputCh chan *future
+
+	bootstrapWorker *bootstrapWorker
 }
 
 // NewEncoderGroup creates a new EncoderGroup instance
-func NewEncoderGroup(builder RowEventEncoderBuilder,
-	concurrency int, changefeedID model.ChangeFeedID,
+func NewEncoderGroup(
+	cfg *config.SinkConfig,
```
3AceShowHand, line 66: `protocol config.Protocol` should ba add into the variables.

pkg/sink/codec/encoder_group.go
```diff
@@ -70,15 +76,40 @@ func NewEncoderGroup(builder RowEventEncoderBuilder,
...
 		inputCh[i] = make(chan *future, defaultInputChanSize)
 	}
+	outCh := make(chan *future, defaultInputChanSize*concurrency)
+
+	var bootstrapWorker *bootstrapWorker
+	protocol := util.GetOrZero(cfg.Protocol)
+	if protocol == config.ProtocolSimple.String() {
+		log.Info("Sending bootstrap event is enable for simple protocol, creating bootstrap worker...",
+			zap.Stringer("changefeed", changefeedID))
+		sendBootstrapIntervalInSec := util.GetOrZero(cfg.SendBootstrapIntervalInSec)
```
3AceShowHand, line 86: `sendBootstrapIntervalInSec` should already be correctly initialized when create the changefeed, there is no need to adjust the value.

pkg/sink/codec/encoder_group.go
```diff
@@ -70,15 +76,40 @@ func NewEncoderGroup(builder RowEventEncoderBuilder,
...
+		if sendBootstrapIntervalInSec <= 0 {
+			sendBootstrapIntervalInSec = config.DefaultSendBootstrapIntervalInSec
+		}
+		msgCount := util.GetOrZero(cfg.SendBootstrapInMsgCount)
+		if msgCount <= 0 {
+			msgCount = config.DefaultSendBootstrapInMsgCount
+		}
+		interval := time.Duration(sendBootstrapIntervalInSec) * time.Second
+		bootstrapWorker = newBootstrapWorker(
```
3AceShowHand, line 95: when creates the `bootstrapWorker`, `sendBootstrapIntervalInSec` and  should be passed as variables.

3AceShowHand: The caller should not know too much about the `bootstrapWorker`.

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+	// It is used to generate bootstrap message
+	tableInfo atomic.Value
+}
+
+func newTableStatus(key TopicPartitionKey, row *model.RowChangedEvent) *tableStatus {
+	res := &tableStatus{
+		id:    row.Table.TableID,
+		topic: key.Topic,
+	}
+	res.counter.Add(1)
```
3AceShowHand, line 223: Why the count is set `1`

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,263 @@
...
+	ctx context.Context,
+	key TopicPartitionKey,
+	row *model.RowChangedEvent,
+) error {
+	table, ok := b.activeTables.Load(row.Table.TableID)
+	if !ok {
+		tb := newTableStatus(key, row)
+		b.activeTables.Store(tb.id, tb)
+		// Send bootstrap message immediately when a new table is added
+		err := b.sendBootstrapMsg(ctx, tb)
```
3AceShowHand, line 105: Will this sent the first bootstrap event ?

pkg/sink/codec/bootstraper.go
asddongmen: typo

pkg/sink/codec/bootstraper.go
asddongmen: It can't return out of the Range method of sync.Map.

pkg/sink/codec/bootstraper.go
asddongmen: Since it is generated by CDC rather than upstream, it does not have a commitTs and startTs.

pkg/sink/codec/bootstraper.go
asddongmen: A `tableStatistic` will only be created when a `rowChangeEvent` goes through the `bootstrapWorker`, so the counter should add `1`.

pkg/sink/codec/bootstraper.go
asddongmen: Yes.

cdc/sink/dmlsink/mq/kafka_dml_sink.go
```diff
@@ -122,7 +122,7 @@ func NewKafkaDMLSink(
 	metricsCollector := factory.MetricsCollector(tiflowutil.RoleProcessor, adminClient)
 	dmlProducer := producerCreator(ctx, changefeedID, asyncProducer, metricsCollector, errCh, failpointCh)
 	concurrency := tiflowutil.GetOrZero(replicaConfig.Sink.EncoderConcurrency)
```
3AceShowHand, line 124: since `SinkConfig` is passed to the encoder group, `concurrency` can be moved inside `NewEncoderGroup`

cdc/sink/dmlsink/mq/mq_dml_sink.go
```diff
@@ -196,8 +195,11 @@ func (s *dmlSink) WriteEvents(txns ...*dmlsink.CallbackableEvent[*model.SingleTa
 			// We already limit the memory usage by MemoryQuota at SinkManager level.
 			// So it is safe to send the event to a unbounded channel here.
 			s.alive.worker.msgChan.In() <- mqEvent{
-				key: codec.TopicPartitionKey{
-					Topic: topic, Partition: index, PartitionKey: key,
+				key: model.TopicPartitionKey{
+					Topic:          topic,
+					Partition:      index,
+					PartitionKey:   key,
+					TotalPartition: partitionNum,
```
3AceShowHand, line 202: Is it necessary to pass in `TotalPartition`? 

Bootstrap is sent to the first partition.

3AceShowHand (approved):

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,250 @@
...
+// newBootstrapWorker creates a new bootstrapWorker instance
+func newBootstrapWorker(
+	changefeedID model.ChangeFeedID,
+	outCh chan<- *future,
+	encoder RowEventEncoder,
+	sendBootstrapInterval int64,
+	sendBootstrapInMsgCount int32,
+	maxInactiveDuration time.Duration,
+) *bootstrapWorker {
+	log.Info("Sending bootstrap event is enable for simple protocol,"+
```
sdojjy, line 59:
```suggestion
	log.Info("Sending bootstrap event is enabled for simple protocol,"+
```

pkg/sink/codec/bootstraper.go
```diff
@@ -0,0 +1,250 @@
...
+	lastSendTime := t.lastSendTime.Load().(time.Time)
+	return time.Since(lastSendTime) >= sendBootstrapInterval ||
+		t.counter.Load() >= sendBootstrapMsgCountInterval
+}
+
+func (t *tableStatistic) update(row *model.RowChangedEvent) {
+	t.counter.Add(1)
+	t.lastMsgReceivedTime.Store(time.Now())
+
+	if t.version.Load() != row.TableInfo.UpdateTS {
```
sdojjy, line 236: send a  bootstrap message if the table version is changed.

sdojjy (approved):

pkg/sink/codec/bootstraper.go
asddongmen: This is unneeded because we've already sent DDL event.

Issue #10341, opened 2023-12-21
sink to mysql (cdc) workload skew issue


zhangjinpeng1987:
## Workload Skew Problem

Sink to MySQL use the `conflict detector` to detect potential conflicts between transactions, and build the dependency graph. When a transaction depend on other transactions, this transaction must execute after these dependent transactions. The logic is simple and straight forward. But the `conflict detector` introduced a fast dependencies resolving optimization: for example transaction B only depend on transaction A, and if transaction A has sent to a special worker, the `conflict detector` will resolve the dependency and then send B to the same worker where transaction A is, even before transaction A actually executed. This is because each worker execute received transaction one by one in order. So this optimization can guarantee the transaction B must execute after transaction A. This optimization can resolve dependency early and send transaction to worker early when a transaction just depend on another single transaction.

But this early resolve dependency optimization also can result in workload skew issue (one worker handle most transactions) in some cases, and result in the sink to MySQL has a big lag because of its throughput can't catch up with upstream QPS. For example we have A, B, C, D 4 transactions, they come into the conflict detector in order. Transaction A has no dependency on other transactions, and it has been sent to worker 1. Transaction B is a big transaction contains multiple row changes, it has dependency on transaction A because of key 1, because transaction B just depend on transaction A, so according to the above early dependency resolving optimization, transaction B should be send to worker 1 ASAP. And then transaction C contains one Key 555 and has a dependency on transaction B because of key 555, transaction D contains 2 keys [1000, 1001] and has a dependency on transaction B because of key 1000. Both transaction C and transaction D can apply the early dependency resolving optimization. The consequence is transaction A, B, C, D, all of them will be routed to the same worker 1. From the above example we can see that, the early dependency resolving optimization has a problem that will "attract" more and more transactions to a single worker. Actually after Transaction B executed, transaction C and transaction D can be run parallel in different workers.

```
Txn C(row555)                                         Txn D(row1000, row1001)
         |                                                       |
         |                                                       |
         ------------> Txn B(row1, row2, ..., row1000) <-------
                                          |
                                          |
                                          -------------> Txn A(row1)    worker1
```

## New Proposal

The motivation of early dependency resolving optimization is resolve the transaction dependency as early as possible, but resolve the dependency and send the transaction to a random worker after depended transactions executed is a better choice in terms of stability and predictable throughput.

asddongmen:
Even though the early conflict resolving mechanism has been removed, the workload skew issue may still exist in some rare cases.
[...]

PR #10376, opened 2023-12-27
sink-to-mysql(cdc) simplify conflict detector, prevent workload skew issue

zhangjinpeng1987:

Issue Number: close https://github.com/pingcap/tiflow/issues/10341

- Remove early conflict resolving optimization for the conflict detector, prevent workload skew issue described in https://github.com/pingcap/tiflow/issues/10341
- Simplify the whole code logic, make it more easy to maintain.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```

### DIFF
--- a/pkg/causality/internal/node.go
+++ b/pkg/causality/internal/node.go
@@ -14,6 +14,7 @@
 package internal
 
 import (
+	"fmt"
 	"sync"
 	stdatomic "sync/atomic"
 
@@ -209,13 +210,11 @@ func (n *Node) checkReadiness() bool {
 	}
 
 	removedDependencies := stdatomic.LoadInt32(&n.removedDependencies)
-
-	// All dependcies are removed, so assign the node to any worker is fine.
-	if removedDependencies == n.totalDependencies {
-		return true
+	if removedDependencies > n.totalDependencies {
+		panic(fmt.Sprintf("removedDependencies %d > totalDependencies %d which is not expected",
+			removedDependencies, n.totalDependencies))
 	}
-
-	return false
+	return removedDependencies == n.totalDependencies
 }
 
 func (n *Node) getOrCreateDependers() *btree.BTreeG[*Node] {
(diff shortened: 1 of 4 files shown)


pkg/causality/internal/node.go
```diff
@@ -212,96 +180,41 @@ func (n *Node) assignTo(workerID int64) bool {
 	defer n.mu.Unlock()
 
 	if n.assignedTo != unassigned {
-		// Already resolved by some other guys.
+		// Already handled by some other guys.
```
asddongmen, line 183: // Already assign to a worker by another goroutrine.

asddongmen (approved):

pkg/causality/internal/node.go
```diff
@@ -212,96 +180,41 @@ func (n *Node) assignTo(workerID int64) bool {
...
-// tryResolve try to find a worker to assign the node to.
-// Returns (_, false) if there is a conflict,
-// returns (rand, true) if there is no conflict,
-// returns (N, true) if only worker N can be used.
-func (n *Node) tryResolve(resolvedDependencies, removedDependencies int32) (int64, bool) {
-	assignedTo, resolved := n.doResolve(resolvedDependencies, removedDependencies)
-	if resolved && assignedTo == assignedToAny {
-		assignedTo = n.RandWorkerID()
+func (n *Node) maybeReadyToRun() {
+	if ok := n.checkReadiness(); ok {
```
CharlesCheung96, line 197: When a large number of small transactions are in conflict, does performance degrade significantly?

zhangjinpeng1987: It is suppose not fast in the case of there are a large number of small transactions have conflict, no matter what you do.

pkg/causality/internal/node.go
zhangjinpeng1987: `handled` is a generic word here, not equal to `executed`.

hicqu (approved):

PR #10415, opened 2024-01-03
dm: support MySQL 8.0 GA

### LINKED ISSUE #10405: DM supports MySQL 8.0 GA
## Enhancement
Currently descring of DM as below:
Database version requirements
MySQL version 5.5 ~ 5.7
MySQL version 8.0 (experimental features)
As EOL of Aurora2 (MySQL5.7) is approaching, customer is considering if there is any way to migrate from Aurora3 (MySQL8.0) to TiDB for the future.

lyzx2001:

Issue Number: close #10405 

DM supports MySQL 8.0 GA.

#### Tests 

 - Unit test

No.
https://github.com/pingcap/docs/pull/15959
https://github.com/pingcap/docs-cn/pull/15985
### Release note 

```release-note
DM supports MySQL 8.0 GA.
```

### DIFF
--- a/dm/pkg/checker/mysql_server.go
+++ b/dm/pkg/checker/mysql_server.go
@@ -35,14 +35,14 @@ func NewMySQLVersionChecker(db *sql.DB, dbinfo *dbutil.DBConfig) RealChecker {
 }
 
 // SupportedVersion defines the MySQL/MariaDB version that DM/syncer supports
-// * 5.6.0 <= MySQL Version < 8.0.36.
+// * 5.6.0 <= MySQL Version < 8.1.0.
 var SupportedVersion = map[string]struct {
 	Min MySQLVersion
 	Max MySQLVersion
 }{
 	"mysql": {
 		MySQLVersion{5, 6, 0},
-		MySQLVersion{8, 0, 36},
+		MySQLVersion{8, 1, 0},
 	},
 }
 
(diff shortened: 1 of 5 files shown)


dm/pkg/checker/mysql_server.go
```diff
@@ -35,14 +35,14 @@ func NewMySQLVersionChecker(db *sql.DB, dbinfo *dbutil.DBConfig) RealChecker {
...
-// * 5.6.0 <= MySQL Version < 8.0.0.
+// * 5.6.0 <= MySQL Version < 8.0.34.
 var SupportedVersion = map[string]struct {
 	Min MySQLVersion
 	Max MySQLVersion
 }{
 	"mysql": {
 		MySQLVersion{5, 6, 0},
-		MySQLVersion{8, 0, 0},
+		MySQLVersion{8, 0, 34},
```
lance6716, line 45: why do we choose this version?

https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-35.html

lyzx2001: right, supported version should be < `8.0.36`

lance6716: can you explain the reason why we choose that version? currently latest MySQL's release version?

lyzx2001: yes, I initially tend to support `8.0.x`, which means we support all the versions starting with `8.0`, but in the code we need to be specific about the version, so I use the latest MySQL's release version

lance6716: You can use `MySQLVersion{8, 1, 0}` then

lyzx2001: modified

lance6716 (approved):

D3Hunter (approved):

PR #10422, opened 2024-01-05
api(ticdc): only update upstreamInfo that has changed

### LINKED ISSUE #10430: TiCDC return ErrChangeFeedAlreadyExists when changefeeds are created concurrently
### What did you do?

Create 32 changefeeds concurrently.

_No response_

### What did you see instead?

img_v3_026q_a6e36f04-19a8-4a4f-ae1f-63185e36874g

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

CharlesCheung96:

Issue Number: close #10430

there are two apis that update or create upstream information:

1. create changfeed and upstreamPUT /api/v2/changefeeds/{changefeed_id}
2. update changefeed and upstream POST /api/v2/changefeeds

And, the expected usage of these apis is:
1. Avoid modifying the same changefeed concurrently .
2. Avoid modifying the same upstream concurrently.

Before this PR, when a changefeed is created, cdc always updates the uptream information, regardless of whether the uptream changes.  As a result, when different ChangeFeeds are created concurrently, cdc conflicts with concurrent updates to the same upstream key.

To solve this issue, we should only update upstreamInfo that has changed when creating changefeed.

------

Besides, cdc did not perform any checks when updating changefeed info:
https://github.com/pingcap/tiflow/blob/c6b3fec5c91ed1204daa9b649bf1d529a7dd6956/pkg/etcd/etcd.go#L513

To prevent concurrent updates to the same changefeed, we should compare the versions of the info and status keys. 

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
`Fix a issue that TiCDC return ErrChangeFeedAlreadyExists when changefeeds are created concurrently`.
```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -14,17 +14,17 @@
 package etcd
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"net/url"
 	"strings"
 	"time"
 
-	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
 	"github.com/pingcap/tiflow/cdc/model"
 	"github.com/pingcap/tiflow/pkg/config"
-	cerror "github.com/pingcap/tiflow/pkg/errors"
+	"github.com/pingcap/tiflow/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/tikv/pd/pkg/utils/tempurl"
 	"go.etcd.io/etcd/api/v3/mvccpb"
@@ -150,7 +150,6 @@ type CDCEtcdClient interface {
 	UpdateChangefeedAndUpstream(ctx context.Context,
 		upstreamInfo *model.UpstreamInfo,
 		changeFeedInfo *model.ChangeFeedInfo,
-		changeFeedID model.ChangeFeedID,
 	) error
 
 	PutCaptureInfo(context.Context, *model.CaptureInfo, clientv3.LeaseID) error
@@ -201,7 +200,7 @@ func (c *CDCEtcdClientImpl) Close() error {
 // ClearAllCDCInfo delete all keys created by CDC
 func (c *CDCEtcdClientImpl) ClearAllCDCInfo(ctx context.Context) error {
 	_, err := c.Client.Delete(ctx, BaseKey(c.ClusterID), clientv3.WithPrefix())
-	return cerror.WrapError(cerror.ErrPDEtcdAPIError, err)
+	return errors.WrapError(errors.ErrPDEtcdAPIError, err)
 }
 
 // GetClusterID gets CDC cluster ID.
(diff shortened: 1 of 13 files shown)


pkg/etcd/etcd.go
```diff
@@ -458,16 +459,21 @@ func (c *CDCEtcdClientImpl) CreateChangefeedInfo(ctx context.Context,
...
-			clientv3.Compare(clientv3.ModRevision(upstreamInfoKey.String()),
+			clientv3.Compare(clientv3.ModRevision(upstreamEtcdKeyStr),
 				"=", upstreamResp.Kvs[0].ModRevision))
+		if !bytes.Equal(upstreamResp.Kvs[0].Value, upstreamData) {
+			opsThen = append(opsThen,
+				clientv3.OpPut(upstreamEtcdKeyStr, string(upstreamData)))
+		}
 	}
 
 	resp, err := c.Client.Txn(ctx, cmps, opsThen, TxnEmptyOpsElse)
```
sdojjy, line 479: if save changefeed or upstream info to etcd failed, the error message should not be "changefeed already exists"， I was confused when I see this error message.
 Something like "save changefeed to meta store failed" would be better.

pkg/etcd/etcd.go
```diff
@@ -458,27 +458,33 @@ func (c *CDCEtcdClientImpl) CreateChangefeedInfo(ctx context.Context,
...
 	opsThen := []clientv3.Op{
 		clientv3.OpPut(infoKey, value),
-		clientv3.OpPut(upstreamEtcdKeyStr, string(upstreamData)),
 	}
 	if len(upstreamResp.Kvs) == 0 {
 		cmps = append(cmps,
 			clientv3.Compare(clientv3.ModRevision(upstreamEtcdKeyStr),
 				"=", 0))
+		opsThen = append(opsThen,
+			clientv3.OpPut(upstreamEtcdKeyStr, string(upstreamData)))
```
AndreMouche, line 467: Should we consider the case that this function(`CreateChangefeedInfo`) was called concurrently, is it still possible for it to start as `upstreamResp.Kvs` is empty and then become non-empty?
for example:
goroutine1 at t1: upstreamResp.kvs is empty 
goroutine2 at t2:upstreamResp.kv is empty
goroutine1 comes to L466 and finished L478 successfully
goroutine2 comes to L466 and append this PutOp and goes to L478 failed.
Meanwhile, could you please add a failpoint for it and add a unit test for the above case?

CharlesCheung96: The multi-upstream feature has not been released, so this issue does not exist in customer scenarios. 

Note that there are two apis that update or create upstream information:
1. create changfeed and upstream`PUT /api/v2/changefeeds/{changefeed_id}`
2. update changefeed and upstream `POST /api/v2/changefeeds`

And, the expected usage of these apis is:
1. Avoid modifying the same changefeed concurrently (This is obvious).
2. Avoid modifying the same upstream concurrently (However, this is not obvious).

The second scenario here is described in the comments. 

In the future, perhaps we should update the upstream information using a separate api.

asddongmen (approved):

flowbehappy:
@CharlesCheung96 Please describe the detail cause of this issue.

sdojjy (approved):

Issue #9230, opened 2023-06-14
engine: support new dm task mode


lichunzhu:
### Is your feature request related to a problem?

dm cloud import needs to speed up. We need engine to support a separate dumpling and load&sync job.

### Describe the feature you'd like

engine supports a separate dumpling and load&sync job.

PR #10444, opened 2024-01-10
dm: merge dm on cloud accelerate chunk related code into master

lichunzhu:

Issue Number: ref https://github.com/pingcap/tiflow/issues/9230

1. ignore gcp not exist error when we use external storage to delete all files
2. adjust checkpoint to mysql when we use s3/gcs storage
3. skip load table structure for S3 loader

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```


D3Hunter (approved):

dm/syncer/syncer.go
```diff
@@ -2994,6 +2994,10 @@ func (s *Syncer) genRouter() error {
 
 func (s *Syncer) loadTableStructureFromDump(ctx context.Context) error {
 	logger := s.tctx.L()
+	if !storage.IsLocalDiskPath(s.cfg.LoaderConfig.Dir) {
+		logger.Info("skip load table structure from dump files for non-local-dir loader because it may be slow", zap.String("loaderDir", s.cfg.LoaderConfig.Dir))
```
lance6716, line 2998:
```suggestion
		logger.Warn("skip load table structure from dump files for non-local-dir loader because it may be slow", zap.String("loaderDir", s.cfg.LoaderConfig.Dir))
```

I prefer log a warning because this affects correctness when downstream table structure is different with upstream. And leave a TODO that we parallel open the files to improve speed

lance6716 (commented):
rest lgtm

lance6716 (approved):

Issue #10457, opened 2024-01-11
Refine code 


hongyunyan:
# Enhancement 
just a stub issue, welcome to ref this in your pr when you think it needn't open a new issue.

Don't abuse it!

PR #10466, opened 2024-01-15
mysql (ticdc): Improve the performance of the mysql sink by refining the transaction event batching logic

hongyunyan:

Issue Number: ref #10457

1. When there are a lot of txns in worker's txnCh, we try to do flush until save batches to maxTxnRows, instead of doing flush with tiny txns because of ticker.C.
2. Adjust the calculation of worker-busy-ratio to make it more precise and more clear.
     The original calculation of worker-busy-ratio may have certain deviation when the flush time is long (such as larger than hundreds of milliseconds). And the longer the flush time, the larger the deviation. The reason for the deviation is that when the flsuh time is long, the interval between each time we increase the worker busy ratio value clearly exceeds 1s, resulting in the growth rate of the worker-busy-ratio per second being less than actual rate.

#### Performance Test Result
We take a large workload on upstream to keep worker-busy-ratio as 100%, and compare the improvement of sink performance before and after optimization.
1. Simulate network latency between the upstream and downstream is 2ms.  
    After Optimization -- 33339 rows/s (**+14%**). Before Optimization -- 29075 rows/s
2. Simulate network latency between the upstream and downstream is 5ms.  
     After Optimization -- 20528 rows/s(**+13%**). Before Optimization -- 18129 rows/s
3. Simulate network latency between the upstream and downstream is 20ms.  
    After Optimization -- 7273 rows/s(**+20%**). Before Optimization -- 6048 rows/s
image

#### New panel of worker busy ratio 
img_v3_0274_02f665a8-9887-4a35-a4f7-9dceaa8d497g

#### Tests 

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/sink/dmlsink/txn/worker.go
+++ b/cdc/sink/dmlsink/txn/worker.go
@@ -132,8 +132,12 @@ func (w *worker) run() error {
 							needFlush = true
 						}
 					}
+					//Release resources promptly
 					if !delay.Stop() {
-						<-delay.C
+						select {
+						case <-delay.C:
+						default:
+						}
 					}
 				}
 				// needFlush must be true here, so we can do flush.
--- a/cdc/sink/dmlsink/txn/worker.go
+++ b/cdc/sink/dmlsink/txn/worker.go
@@ -123,15 +123,12 @@ func (w *worker) runLoop() error {
 			// we get the data from txnCh.out until no more data here or reached the state that can be flushed.
 			if txn.txnEvent != nil {
 				needFlush = w.onEvent(txn)
-				if !needFlush {
-					for txn = range w.txnCh.Out() {
-						if txn.txnEvent == nil {
-							break
-						}
+				for !needFlush {
+					select {
+					case txn := <-w.txnCh.Out():
 						needFlush = w.onEvent(txn)
-						if needFlush {
-							break
-						}
+					default:
+						break
 					}
 				}
 			}
--- a/cdc/sink/dmlsink/txn/worker.go
+++ b/cdc/sink/dmlsink/txn/worker.go
@@ -123,16 +123,18 @@ func (w *worker) run() error {
 			if txn.txnEvent != nil {
 				needFlush = w.onEvent(txn)
 				if !needFlush {
-					delayTicker := time.NewTicker(w.flushInterval)
+					delay := time.NewTimer(w.flushInterval)
 					for !needFlush {
 						select {
 						case txn := <-w.txnCh.Out():
 							needFlush = w.onEvent(txn)
-						case <-delayTicker.C:
+						case <-delay.C:
 							needFlush = true
 						}
 					}
-					delayTicker.Stop()
+					if !delay.Stop() {
+						<-delay.C
+					}
 				}
 				// needFlush must be true here, so we can do flush.
 				if err := w.doFlush(); err != nil {
--- a/cdc/sink/dmlsink/txn/worker.go
+++ b/cdc/sink/dmlsink/txn/worker.go
@@ -124,13 +124,12 @@ func (w *worker) runLoop() error {
 				needFlush = w.onEvent(txn)
 				if !needFlush {
 					delay := time.After(w.flushInterval)
-				loop:
 					for !needFlush {
 						select {
 						case txn := <-w.txnCh.Out():
 							needFlush = w.onEvent(txn)
 						case <-delay:
-							break loop
+							needFlush = true
 						}
 					}
 				}
--- a/cdc/sink/dmlsink/txn/worker.go
+++ b/cdc/sink/dmlsink/txn/worker.go
@@ -92,7 +92,7 @@ func (w *worker) close() {
 	w.txnCh.CloseAndDrain()
 }
 
-// Run a loop.
+// Continuously get events from txnCh and call backend flush based on conditions.
 func (w *worker) runLoop() error {
 	defer func() {
 		if err := w.backend.Close(); err != nil {
@@ -157,7 +157,7 @@ func (w *worker) runLoop() error {
 }
 
 // onEvent is called when a new event is received.
-// It returns true if the event is sent to backend.
+// It returns true if it needs flush immediately.
 func (w *worker) onEvent(txn txnWithNotifier) bool {
 	w.hasPending = true
 
@@ -178,7 +178,6 @@ func (w *worker) onEvent(txn txnWithNotifier) bool {
 }
 
 // doFlush flushes the backend.
-// It returns true only if it can no longer be flushed.
 func (w *worker) doFlush() error {
 	if w.hasPending {
 		start := time.Now()
--- a/cdc/sink/dmlsink/txn/worker.go
+++ b/cdc/sink/dmlsink/txn/worker.go
@@ -107,7 +107,7 @@ func (w *worker) runLoop() error {
 		zap.Int("workerID", w.ID))
 
 	needFlush := false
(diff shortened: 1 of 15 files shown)


cdc/sink/dmlsink/txn/worker.go
```diff
@@ -122,27 +117,39 @@ func (w *worker) runLoop() error {
 				zap.Int("workerID", w.ID))
 			return nil
 		case txn := <-w.txnCh.Out():
+			// we get the data from txnCh.out until no more data here or reached the state that can be flushed.
+			// If no more daa in txnCh.out, and also not reached the state that can be flushed,
```
asddongmen, line 121: A typo in this sentence.

cdc/sink/dmlsink/txn/worker.go
```diff
@@ -122,27 +117,39 @@ func (w *worker) runLoop() error {
...
+			// we will wait for 10ms and then do flush to avoid too much flush wish small amount of txns.
 			if txn.txnEvent != nil {
 				needFlush = w.onEvent(txn)
+				if !needFlush {
```
CharlesCheung96, line 125: It seems that the core idea here is to ensure that the flush interval is greater than 10ms.  Maybe we could record lastFlushTime at the end of each flush and check it in each ticker?

hongyunyan: I believe both of these two methods are feasible, and I'm wondering if the current code might be a bit more straightforward and easier to understand?

cdc/sink/dmlsink/txn/worker.go
```diff
@@ -122,27 +117,39 @@ func (w *worker) runLoop() error {
...
+					delay := time.After(w.flushInterval)
```
asddongmen, line 126: IMO, it is better to new a `ticker` outside the `for loop` and reuse it by ticker.Reset() instead of to new a timer every time.

cdc/sink/dmlsink/txn/worker.go
CharlesCheung96: In general, the nested logic is more complex.

cdc/sink/dmlsink/txn/worker.go
hongyunyan: ticker.Reset() can't clear the value in channel, if there is a value in ticker.C. So I think ticker.Reset() here can not be useful.

cdc/sink/dmlsink/txn/worker.go
```diff
@@ -122,27 +117,39 @@ func (w *worker) runLoop() error {
...
+				loop:
+					for !needFlush {
+						select {
+						case txn := <-w.txnCh.Out():
+							needFlush = w.onEvent(txn)
+						case <-delay:
+							break loop
```
asddongmen, line 133:
```suggestion
							needFlush = true
							break loop
```

asddongmen (approved):

cdc/sink/dmlsink/txn/worker.go
hongyunyan: I think it's ok here orz

cdc/sink/dmlsink/txn/worker.go
```diff
@@ -122,27 +117,38 @@ func (w *worker) runLoop() error {
...
+				if !needFlush {
+					delay := time.After(w.flushInterval)
+					for !needFlush {
+						select {
+						case txn := <-w.txnCh.Out():
+							needFlush = w.onEvent(txn)
+						case <-delay:
+							needFlush = true
+						}
+					}
```
hicqu, line 134: If it ends the `if` block starts at line `125`, I think we can flush here directly.

cdc/sink/dmlsink/txn/worker.go
```diff
@@ -122,27 +117,38 @@ func (w *worker) runLoop() error {
 				zap.Int("workerID", w.ID))
 			return nil
 		case txn := <-w.txnCh.Out():
+			// we get the data from txnCh.out until no more data here or reach the state that can be flushed.
```
hicqu, line 120: `startToBatch` is not a good name.
There are 3 stages: waiting for a new batch, collecting a new batch and then flush it. Busy ratio is based on stage 2 and 3.
So I think we can record `startToCollectAndEmit` here, and then observe it after flushed.

cdc/sink/dmlsink/txn/worker.go
hongyunyan: Do you mean that we can move L139-L151 to L135 ？

cdc/sink/dmlsink/txn/worker.go
```diff
@@ -122,27 +117,38 @@ func (w *worker) runLoop() error {
...
+			// we will wait for 10ms and then do flush to avoid too much flush with small amount of txns.
 			if txn.txnEvent != nil {
 				needFlush = w.onEvent(txn)
+				if !needFlush {
+					delay := time.After(w.flushInterval)
+					for !needFlush {
+						select {
+						case txn := <-w.txnCh.Out():
+							needFlush = w.onEvent(txn)
+						case <-delay:
```
hicqu, line 131: The doc of `time.After` shows that `time.NewTimer` is better here. And we should call `delay.Stop` when dropping an not-triggered timer.

cdc/sink/dmlsink/txn/worker.go
hicqu: Yes.

cdc/sink/dmlsink/txn/worker.go
hongyunyan: I think startToCollectAndEmit maybe not appropriate here, we just want to use startToBatching to calculate the total time spent in the worker.

cdc/sink/dmlsink/txn/worker.go
```diff
@@ -122,27 +117,38 @@ func (w *worker) runLoop() error {
...
+					delayTicker := time.NewTicker(w.flushInterval)
+					for !needFlush {
+						select {
+						case txn := <-w.txnCh.Out():
+							needFlush = w.onEvent(txn)
+						case <-delayTicker.C:
+							needFlush = true
+						}
+					}
+					delayTicker.Stop()
```
hicqu, line 135: `Stop` return value should be handled.

cdc/sink/dmlsink/txn/worker.go
```diff
@@ -122,27 +117,38 @@ func (w *worker) runLoop() error {
...
+				}
+				// needFlush must be true here, so we can do flush.
+				if err := w.doFlush(); err != nil {
```
hicqu, line 138: I think `start = time.Now()` should be moved to here.

cdc/sink/dmlsink/txn/worker.go
```diff
@@ -122,27 +117,38 @@ func (w *worker) runLoop() error {
...
+						zap.String("changefeedID", w.changefeed),
+						zap.Int("workerID", w.ID),
+						zap.Error(err))
+					return err
+				}
+				needFlush = false
+				// we record total time to calcuate the worker busy ratio.
+				// so we record the total time after flushing, to unified statistics on
+				// flush time and total time
+				w.metricTxnWorkerTotalDuration.Observe(time.Since(start).Seconds())
```
hicqu, line 149: `metricTxnWorkerFlushDuration` and `metricTxnWorkerTotalDuration` are very similar. We can unify them.

hicqu (approved):

PR #10475, opened 2024-01-15
codec(ticdc): simple protocol add table id as message field

### LINKED ISSUE #10474: simple protocol add table id as message field
Encode `TableID` into simple protocol DML and DDL message content

3AceShowHand:

Issue Number: close #10474 

* Encode `TableID` into simple protocol DML and DDL message content

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```

### DIFF
--- a/pkg/sink/codec/simple/message.go
+++ b/pkg/sink/codec/simple/message.go
@@ -358,8 +358,9 @@ func buildRowChangedEvent(
 	result := &model.RowChangedEvent{
 		CommitTs: msg.CommitTs,
 		Table: &model.TableName{
-			Schema: msg.Schema,
-			Table:  msg.Table,
+			Schema:  msg.Schema,
+			Table:   msg.Table,
+			TableID: msg.TableID,
 		},
 		TableInfo: tableInfo,
 	}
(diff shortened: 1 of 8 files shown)


pkg/sink/codec/simple/message.go
```diff
@@ -543,16 +550,19 @@ func newDDLMessage(ddl *model.DDLEvent) (*message, error) {
...
-		Schema:        event.Table.Schema,
-		Table:         event.Table.Table,
-		CommitTs:      event.CommitTs,
-		BuildTs:       time.Now().UnixMilli(),
-		SchemaVersion: event.TableInfo.UpdateTS,
-		HandleKeyOnly: onlyHandleKey,
+		Version:            defaultVersion,
+		Schema:             event.Table.Schema,
+		Table:              event.Table.Table,
+		TableID:            event.Table.TableID,
```
asddongmen, line 560: Should use the `ID` from `TableInfo` instead of the `TableID` in `TableName`, because the latter one would be the partition's physical ID.

asddongmen (approved):

CharlesCheung96 (approved):

PR #10483, opened 2024-01-16
*(ticdc): optimize memory usage of RowChangedEvent

### LINKED ISSUE #10386: Optimize memory usage of TiCDC

lidezhu:

Issue Number: close #10386 

1. Add a more memory efficient struct `ColumnData` to represent column data, and the schema info of column can only be fetched from `TableInfo`;
2. Replace `[]*Column` in `RowChangedEvent` with `[]*ColumnData`;
3. Add method `RowChangedEvent.GetColumns` and `RowChangedEvent.GetPreColumns` to get the `Column` representation of columns;
4. There should be no `nil` in `RowChangedEvent.Columns` and `RowChangedEvent.PreColumns`, so this pr remove some code in sink module to handle `nil` value.
5. Maintain a correct `TableInfo` in all kinds of tests;

#### Tests 

 - Unit test
1. disable tidb gc;
6. load 10M rows of a wide table;
7. create changefeed with `memory-quota` set 8GB;
8. compare the peak memory usage;
Before optimization, peak memory usage: 26.5GB,
image
After optimization, peak memory usage: 10.2GB,
image
The peak memory usage decrease about 60%.

### Release note 

```release-note
None.
```


cdc/entry/mounter_test.go
```diff
@@ -1641,7 +1641,12 @@ func TestBuildTableInfo(t *testing.T) {
 		originTI, err := ddl.BuildTableInfoFromAST(stmt.(*ast.CreateTableStmt))
 		require.NoError(t, err)
 		cdcTableInfo := model.WrapTableInfo(0, "test", 0, originTI)
-		cols, _, _, _, err := datum2Column(cdcTableInfo, map[int64]types.Datum{})
```
asddongmen, line 1644: This err should also be checked.

asddongmen (approved):

asddongmen:
Excellent work! Respect!

cdc/entry/mounter_test.go
lidezhu: Fixed.

sdojjy (approved):

PR #10487, opened 2024-01-17
disable region detail logs in kv-client by default

### LINKED ISSUE #10486: Region logs are too noisy in module kv-client

- [X] Googled your question

### What is your question?

There is a known problem that there are too many region detail logs in module kv-client. Generally they are not helpful in most cases.
So we can disable them by default. If we meet issues about resolved timestamp, then we can re-enable them with a configuration.

hicqu:

Issue Number: close #10486 

* Disable region detail logs in kv-client by default;
* Add a new debug option to enable those logs.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


sdojjy (approved):

pkg/config/debug.go
```diff
@@ -60,4 +60,6 @@ type PullerConfig struct {
 	EnableResolvedTsStuckDetection bool `toml:"enable-resolved-ts-stuck-detection" json:"enable-resolved-ts-stuck-detection"`
 	// ResolvedTsStuckInterval is the interval of checking resolved ts stuck.
 	ResolvedTsStuckInterval TomlDuration `toml:"resolved-ts-stuck-interval" json:"resolved-ts-stuck-interval"`
+	// LogRegionDetails determines whether logs Region details or not in puller and kv-client.
```
CharlesCheung96, line 63: Perhaps it would be more appropriate to use changefeed level parameters and support online updates for LogRegionDetails?

hicqu: Just a server level config is fine enough.

asddongmen: @hicqu @CharlesCheung96 TiCDC has already supported online log level changing. So we can change log level to debug level to print these logs when needed.

CharlesCheung96 (approved):

Issue #10157, opened 2023-11-24
TiCDC changefeed's resolved-ts is stuck


sdojjy:
<img width="1441" alt="image" src="https://github.com/pingcap/tiflow/assets/1779704/15daecf6-b958-48e1-bf36-315a293bc02e">

the workaround is pause/resume changefeed

asddongmen:
This issue has occurred a couple of times recently. 
It always happens when a cross-table region merging operation finishes in the upstream TiKV. Logs like the ones below were spotted in the environment where this issue occurred:
```
cdc log:
[2023/12/14 07:44:47.156 +08:00] [INFO] [region_cache.go:1070] ["switch region peer to next due to send request fail"] [current="region ID: 41899, meta: id:41899 start_key:\"t\\200\\000\\000\\000\\000\\000\\002\\377K_i\\200\\000\\000\\000\\000\\377\\000\\000\\003\\003\\200\\000\\001\\214\\377]\\r\\035\\036\\0014f2\\377bc965\\377-8\\37797d-4e\\377c\\3773-86b4-\\377\\377da33ab63\\377\\3778c94\\000\\000\\000\\377\\000\\373\\000\\000\\000\\000\\000\\000\\371\" end_key:\"t\\200\\000\\000\\000\\000\\000\\002\\377K_r\\001debb\\3777402\\377-87\\37769-40\\37744\\377-b0c6-\\3778\\3778a6e01d\\377\\377aea6\\000\\000\\000\\000\\377\\373\\000\\000\\000\\000\\000\\000\\000\\370\" region_epoch:<conf_ver:5 version:2150 > peers:<id:41900 store_id:1 > peers:<id:41901 store_id:3 > peers:<id:41902 store_id:2 > , peer: id:41900 store_id:1 , addr: 0.0.0.17:20160, idx: 0, reqStoreType: TiKvOnly, runStoreType: tikv"] [needReload=false] [error="[CDC:ErrEventFeedAborted]single event feed aborted"]
[2023/12/14 07:45:16.448 +08:00] [INFO] [replication_manager.go:583] ["schedulerv3: slow table"] [namespace=default] [changefeed=x] [tableID=587] [tableStatus=] [checkpointTs=446303066237960202] [resolvedTs=446303066237960202] [checkpointLag=30.005267224s]

pd log:
[...]

PR #10488, opened 2024-01-17
puller(ticdc):  fix resolvedTs get stuck when region split and merge

sdojjy:

Issue Number: ref #10157

Always delete the the mapping in `cachedRegions` when a span's regionID changes. 

This can prevent an old span node in the frontier from becoming an orphan node, which would cause the resolvedTs of the corresponding span to not be updated.

#### Tests 

 - Unit test

### Release note 

```release-note
Fix a bug where region split and merge cause the changefeed resolvedTs to get stuck.
```

### DIFF
--- a/cdc/puller/frontier/frontier.go
+++ b/cdc/puller/frontier/frontier.go
@@ -15,6 +15,7 @@ package frontier
 
 import (
 	"bytes"
+	"encoding/hex"
 	"fmt"
 	"math"
 	"strings"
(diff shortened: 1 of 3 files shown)


asddongmen (approved):

cdc/puller/frontier/frontier.go
```diff
@@ -102,7 +103,7 @@ func (s *spanFrontier) insert(regionID uint64, span tablepb.Span, ts uint64) {
 		if bytes.Equal(seekRes.Node().Key(), span.StartKey) &&
 			bytes.Equal(next.Key(), span.EndKey) {
 			s.minTsHeap.UpdateKey(seekRes.Node().Value(), ts)
-
+			delete(s.cachedRegions, seekRes.Node().regionID)
```
flowbehappy, line 106: Please add some comment to explain why do we delete this node.

PR #10492, opened 2024-01-17
cdc: fix build table info

lidezhu:

Issue Number: ref #10386 

Fix and improve `BuildTiDBTableInfo`;
1. Add mock column id and mock index id for the generated TableInfo;
2. Add checks to verify the legality of the arguments;
3. Fix some small incorrect behaviour of the previous implementation including set charset and collation correctly;

Fix the wrong behavior that a primary column is always set an extra unique key flag in `WrapTableInfo`;

Add some helper method on `TableInfo` to query related information for columns;

#### Tests 

 - Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/model/sink.go
+++ b/cdc/model/sink.go
@@ -530,16 +530,23 @@ type RedoColumn struct {
 }
 
 // BuildTiDBTableInfo builds a TiDB TableInfo from given information.
-func BuildTiDBTableInfo(columns []*Column, indexColumns [][]int) *model.TableInfo {
+// Note the result TableInfo may not be same as the original TableInfo in tidb.
+// The only guarantee is that you can restore the `Name`, `Type`, `Charset`, `Collation`
+// and `Flag` field of `Column` using the result TableInfo.
+func BuildTiDBTableInfo(tableName string, columns []*Column, indexColumns [][]int) *model.TableInfo {
 	ret := &model.TableInfo{}
-	// nowhere will use this field, so we set a debug message
-	ret.Name = model.NewCIStr("BuildTiDBTableInfo")
+	ret.Name = model.NewCIStr(tableName)
 
+	hasPrimaryKeyColumn := false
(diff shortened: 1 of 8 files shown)


cdc/model/sink.go
```diff
@@ -589,7 +626,20 @@ func BuildTiDBTableInfo(columns []*Column, indexColumns [][]int) *model.TableInf
...
+	if !hasHandleKeyColumn {
+		log.Panic("Handle key not found",
+			zap.String("table", tableName),
+			zap.Any("columns", columns),
+			zap.Any("indexColumns", indexColumns))
+	}
 
+	hasPrimaryKeyIndex := false
+	hasHandleIndex := false
+	// When there is no primary key, cdc will select the not null unique index with least columns.
```
asddongmen, line 638:
```suggestion
	// TiCDC handles columns according to the following rules:
	// 1. If a primary key (PK) exists, it is chosen.
	// 2. If there is no PK, TiCDC looks for a not null unique key (UK) with the least number of columns and the smallest index ID.
```

asddongmen (approved):

cdc/model/sink.go
```diff
@@ -589,7 +626,22 @@ func BuildTiDBTableInfo(columns []*Column, indexColumns [][]int) *model.TableInf
...
+			zap.Any("columns", columns),
+			zap.Any("indexColumns", indexColumns))
+	}
 
+	hasPrimaryKeyIndex := false
+	hasHandleIndex := false
+	// TiCDC handles columns according to the following rules:
+	// 1. If a primary key (PK) exists, it is chosen.
+	// 2. If there is no PK, TiCDC looks for a not null unique key (UK) with the least number of columns and the smallest index ID.
+	// And when the columns num is the same, it will choose the one with smallest index id.
```
asddongmen, line 641:
```suggestion
```

CharlesCheung96 (approved):

PR #10506, opened 2024-01-18
kv-client(cdc): introduce a more simple frontier

hicqu:

Issue Number: ref #10574 

* Add a new server configuration `kv-client.advance-interval-in-ms`:
  a. by default it's `300`, which means calculate ResolvedTs every `300ms`;
  b. if it's set to 0, `cdc/puller.Frontier` will still be used to advance ResolvedTs.

This PR can reduce CPU and memory usage and make TiCDC more happy with the large single table case.

We can remove `cdc/puller.Frontier` and the new configuration after the change is stable.

#### Tests 

 - Unit test
 - Manual test with a heavy sysbench insert load
    With the patch:
    图片

    Without the patch:
    图片

    The result shows the patch could increase lag about `100ms`. I think it's acceptable.

### Release note 

```release-note
None
```


cdc/kv/shared_region_worker.go
```diff
@@ -205,27 +192,24 @@ func (w *sharedRegionWorker) handleResolvedTs(ctx context.Context, batch resolve
...
-			zap.Int("spanCount", len(spansAndChan.spans)))
-		if len(spansAndChan.spans) > 0 {
-			revent := model.RegionFeedEvent{Resolved: &model.ResolvedSpans{
-				Spans: spansAndChan.spans, ResolvedTs: batch.ts,
-			}}
-			x := spansAndChan.requestedTable.associateSubscriptionID(revent)
+	rt := batch.regions[0].sri.requestedTable
+	now := time.Now().UnixMilli()
+	lastAdvance := rt.lastAdvanceTime.Load()
+	if now-lastAdvance > 300 && rt.lastAdvanceTime.CompareAndSwap(lastAdvance, now) {
```
flowbehappy, line 200: Please don't use `300`, use a config setting or constant var instead.

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -218,7 +324,7 @@ func (l *RegionRangeLock) tryLockRange(startKey, endKey []byte, regionID, versio
 	overlappingEntries := l.getOverlappedEntries(startKey, endKey, regionID)
 
 	if len(overlappingEntries) == 0 {
-		checkpointTs := l.rangeCheckpointTs.GetMin(startKey, endKey)
+		checkpointTs := l.rangeCheckpointTs.getMinTsInRange(startKey, endKey)
```
flowbehappy, line 327: I wonder why it is called `rangeCheckpointTs` instead of `rangeResolvedTs`.

hicqu: It should be `ResolvedTs`. `CheckpointTs` is a mistake and we follow it in a long time.

cdc/kv/client.go
```diff
@@ -1467,16 +1466,12 @@ func (s *eventFeedSession) logSlowRegions(ctx context.Context) error {
...
-			for _, hole := range attr.Holes {
-				holes = append(holes, fmt.Sprintf("[%s,%s)", hole.StartKey, hole.EndKey))
-			}
 			log.Info("event feed holes exist",
 				zap.String("namespace", s.changefeed.Namespace),
 				zap.String("changefeed", s.changefeed.ID),
 				zap.Int64("tableID", s.tableID),
 				zap.String("tableName", s.tableName),
-				zap.String("holes", strings.Join(holes, ", ")))
+				zap.Any("holes", attr.Holes))
```
asddongmen, line 1474: What does it mean and what will happen when there is a hole? Can you add some explanations?

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -31,84 +31,190 @@ import (
...
 }
 
-// Set sets the corresponding ts of the given range to the specified value.
-func (m *rangeTsMap) Set(startKey, endKey []byte, ts uint64) {
-	if _, ok := m.m.Get(rangeTsEntryWithKey(endKey)); !ok {
-		// To calculate the minimal ts, the default value is math.MaxUint64
-		tailTs := uint64(math.MaxUint64)
-		m.m.DescendLessOrEqual(rangeTsEntryWithKey(endKey), func(i *rangeTsEntry) bool {
-			tailTs = i.ts
+func (m *rangeTsMap) clone() (target *rangeTsMap) {
```
asddongmen, line 68:
```suggestion
func (m *rangeTsMap) clone() (res *rangeTsMap) {
```

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -31,84 +31,190 @@ import (
 )
 
 type rangeTsEntry struct {
-	// Only startKey is necessary. End key can be inferred by the next item since the map always keeps a continuous
-	// range.
+	// Only startKey is necessary. End key can be inferred by the next item,
+	// since the map always keeps a continuous range.
 	startKey []byte
 	ts       uint64
+	isUnset  bool
```
asddongmen, line 38: Can you add a comment about what does `isUnset` mean?

cdc/kv/client.go
hicqu: Commentted at the defination of `Holes`.

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -31,84 +31,194 @@ import (
...
+			found = true
+			endKeyOverlapped.ts = i.ts
+			endKeyOverlapped.isUnset = i.isUnset
 			return false
 		})
-		m.m.ReplaceOrInsert(&rangeTsEntry{
-			startKey: endKey,
-			ts:       tailTs,
-		})
+		if found {
```
CharlesCheung96, line 95: Is it possible that found is false?

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -31,84 +31,194 @@ import (
...
+	}
+	m.m.Ascend(func(i rangeTsEntry) bool {
+		res.m.ReplaceOrInsert(i)
+		return true
+	})
+	return
+}
+
+func (m *rangeTsMap) set(startKey, endKey []byte, ts uint64) {
+	if !bytes.Equal(m.end, endKey) && !m.m.Has(rangeTsEntryWithKey(endKey)) {
```
CharlesCheung96, line 86: Will `endkey` be inserted into the btree when initialization?

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -31,84 +31,194 @@ import (
...
 
-	entriesToDelete := make([]*rangeTsEntry, 0)
-	m.m.AscendRange(rangeTsEntryWithKey(startKey), rangeTsEntryWithKey(endKey),
-		func(i *rangeTsEntry) bool {
-			entriesToDelete = append(entriesToDelete, i)
-			return true
+	if !m.m.Has(rangeTsEntryWithKey(startKey)) {
+		var found bool
+		var startKeyOverlapped rangeTsEntry
+		m.m.DescendLessOrEqual(rangeTsEntryWithKey(startKey), func(i rangeTsEntry) bool {
```
CharlesCheung96, line 107: Is the core idea here to find the node before the target node?

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -31,84 +31,194 @@ import (
...
-	// range.
+	// Only startKey is necessary. End key can be inferred by the next item,
+	// since the map always keeps a continuous range.
 	startKey []byte
 	ts       uint64
+
+	// rangeTsEntry is used in rangeTsMap. rangeTsMap.set will associate a timestamp to a given key range,
+	// and rangeTsMap.unset will revoke the relationship. `isUnset` indicates whether a key range is
+	// generated from a set or unset operation.
+	isUnset bool
```
asddongmen, line 42: May be name it isSet and let isSet = !isUnset is easier to understand.

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -31,84 +31,194 @@ import (
...
+		return true
+	})
+	return
+}
+
+func (m *rangeTsMap) set(startKey, endKey []byte, ts uint64) {
+	if !bytes.Equal(m.end, endKey) && !m.m.Has(rangeTsEntryWithKey(endKey)) {
+		var found bool
+		var endKeyOverlapped rangeTsEntry
+		m.m.DescendLessOrEqual(rangeTsEntryWithKey(endKey), func(i rangeTsEntry) bool {
```
asddongmen, line 89: IMOP, it is better to declare:
```
startEntry := rangeTsEntryWithKey(startKey)
endEntry := rangeTsEntryWithKey(endKey)
```
at the begining of the function rather than new them multiple times by `rangeTsEntryWithKey(x)`.

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -31,84 +31,194 @@ import (
...
+		end:   m.end,
+	}
+	m.m.Ascend(func(i rangeTsEntry) bool {
+		res.m.ReplaceOrInsert(i)
+		return true
+	})
+	return
+}
+
+func (m *rangeTsMap) set(startKey, endKey []byte, ts uint64) {
```
asddongmen, line 85:
```suggestion
// set the timestamp of range [startKey, endKey) to ts. 
// Note: It leaves the timestamp of range [endKey, +∞) intact.  
func (m *rangeTsMap) set(startKey, endKey []byte, ts uint64) {
```

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -31,84 +31,194 @@ import (
...
-		startKey: startKey,
-		ts:       ts,
+	m.m.ReplaceOrInsert(rangeTsEntry{startKey: startKey, ts: ts})
+}
+
+func (m *rangeTsMap) unset(startKey, endKey []byte) {
+	var neighbor rangeTsEntry
+	var exist bool
+
+	if !bytes.Equal(m.end, endKey) {
```
asddongmen, line 136: IMOP, it is better to check `if bytes.Equal(m.end, endKey)`  first to reduce the nested if blocks.

asddongmen, line 136: It would be highly appreciated if you could provide a brief explanation before each major step in the set and unset functions.

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -218,31 +328,33 @@ func (l *RegionRangeLock) tryLockRange(startKey, endKey []byte, regionID, versio
 	overlappingEntries := l.getOverlappedEntries(startKey, endKey, regionID)
 
 	if len(overlappingEntries) == 0 {
-		checkpointTs := l.rangeCheckpointTs.GetMin(startKey, endKey)
+		checkpointTs := l.rangeResolvedTs.getMinTsInRange(startKey, endKey)
```
asddongmen, line 331:
```suggestion
		resolvedTs := l.rangeResolvedTs.getMinTsInRange(startKey, endKey)
```

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -218,31 +328,33 @@ func (l *RegionRangeLock) tryLockRange(startKey, endKey []byte, regionID, versio
...
 		l.regionIDLock[regionID] = newEntry
 
+		l.rangeResolvedTs.unset(startKey, endKey)
 		log.Debug("range locked",
 			zap.String("changefeed", l.changefeedLogInfo),
-			zap.Uint64("lockID", l.id), zap.Uint64("regionID", regionID),
+			zap.Uint64("lockID", l.id),
+			zap.Uint64("regionID", regionID),
 			zap.Uint64("version", version),
 			zap.Uint64("checkpointTs", checkpointTs),
```
asddongmen, line 349:
```suggestion
			zap.Uint64("resolvedTs", resolvedTs),
```

asddongmen: Or we could rename them to watermark

pkg/config/kvclient.go
```diff
@@ -22,6 +22,8 @@ type KVClientConfig struct {
 	WorkerConcurrent uint `toml:"worker-concurrent" json:"worker-concurrent"`
 	// how many grpc streams will be established to every TiKV node
 	GrpcStreamConcurrent uint `toml:"grpc-stream-concurrent" json:"grpc-stream-concurrent"`
+	// Advance table ResolvedTs interval.
+	AdvanceIntervalInMs uint `toml:"advance-interval-in-ms" json:"advance-interval-in-ms"`
```
asddongmen, line 26: Would you think `WatermarkAdvanceIntervalInMs` is a better name?

cdc/kv/regionlock/region_range_lock.go
hicqu: No. With the PR the btree only contains start keys.

cdc/kv/regionlock/region_range_lock.go
hicqu: There must be a node overlapped with the target. We need to find and update it.

cdc/kv/regionlock/region_range_lock.go
hicqu: Yes, it's possible.

pkg/config/kvclient.go
hicqu: I prefer to a shorter name. Besides `ResolvedTs`, there is no other things can be advanced in puller. So I think the short name is OK.

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -150,12 +279,12 @@ type RegionRangeLock struct {
...
 
-	mu                sync.Mutex
-	rangeCheckpointTs *rangeTsMap
-	rangeLock         *btree.BTreeG[*rangeLockEntry]
-	regionIDLock      map[uint64]*rangeLockEntry
-	stopped           bool
-	refCount          uint64
+	mu              sync.RWMutex
+	rangeResolvedTs *rangeTsMap
+	rangeLock       *btree.BTreeG[*rangeLockEntry]
```
CharlesCheung96, line 284: The rangeResolvedTs only maintain information about regions that are not locked.  Also, it appears that there is no overlap between the data maintained by rangeResolvedTs and rangeLock.

Is it possible to use single btree data to maintain locked and unlocked regions?

asddongmen (approved):

cdc/kv/regionlock/region_range_lock.go
hicqu: Good idea, we can take a try later, in another PR.

cdc/kv/regionlock/region_range_lock.go
```diff
@@ -150,12 +279,12 @@ type RegionRangeLock struct {
...
+	regionIDLock    map[uint64]*rangeLockEntry
+	stopped         bool
+	refCount        uint64
```
CharlesCheung96, line 287: Is refCount alwys euqls to `rangeLock.Len()`?

CharlesCheung96:
Please add basic integration test for the new frontier.

asddongmen:
I think it is worth mention this will also improve TiCDC ability of replicating big table. 
If I am right, after this PR, a single TiCDC node(8c, 16gb) can replicate a table that has more than 1.5 million regions.

CharlesCheung96 (approved):

CharlesCheung96 (approved):

PR #10518, opened 2024-01-22
pkg/filter(ticdc): ignore ddl in ddl_manager

### LINKED ISSUE #10524: ddl events should not be filtered in puller
The dataflow of ddl in TiCDC is:
```
TiKV --> kvClient --> ddlPuller --> owner.schemaStorage  --> owner.ddlManager  --> ddlSink
```	
There are two methods to filter ddl event:
1. Discard: filter ddl event in ddlPuller.
2. Ignore: filter ddl event in ddlManager.

In the current implementation, TiCDC discrd ddl events, which should be filtered, in [ddlPuller](https://github.com/pingcap/tiflow/blob/master/cdc/puller/ddl_puller.go#L521-L530) and the subsequent modules of puller do not handle such ddl events. This causes cdc to behave unreasonably when using  [`event-filter rules`](https://docs.pingcap.com/tidb/stable/ticdc-filter#:~:text=create%20schema,drop%20view%20event). 
For example, if we create a changefeed-test with the following config:
```toml
[filter]
rules = ['test.*']

[[filter.event-filters]]
matcher = ["test.t1"] # matcher is an allow list, which means this rule only applies to the worker table in the test database.
ignore-event = ["create table"] # Ignore `create table` ddl events.
```
Then, 
1. After executing `create table test.t0`, cdc will capture table t0.
3. After executing `create table test.t1`, cdc will discard the ddl since it match event-filter rules. Therefore, cdc will not capture table t1.
4. However, after restarting changefeed-test, cdc will get a schema snapshot from tikv, and capture both t0 and t1. This is confusing to the user.

-------
In conclusion, the behavior of TiCDC in ignoring create/delete/truncate/alter DDL in puller module is as follows:

1. After a table or partition is created, TiCDC does not synchronize the new physical table. However, cdc will synchronize the new physical table after the changefeed restarts.
5. After a table or partition is dropped, TiCDC still synchronizes the deleted physical table.
6. After a table or partition is truncated, TiCDC does not synchronize the new physical table, but still synchronizes the deleted physical table.
7. After the Alter table or partition ddl, CDC still uses the schma of the old table to parse data, wich leads to unpredictable data inconsistencies.

CharlesCheung96:

Issue Number: close #10524

ignore ddl in ddl_manager, and the behavior of TiCDC in ignoring create/delete/truncate/alter DDL scenarios is as follows:

1. After a table or partition is created, TiCDC synchronizes the new physical table. 
2. After a table or partition is dropped, TiCDC does not synchronize the deleted physical table.
3. After a table or partition is truncated, TiCDC does not synchronize the deleted physical table, and synchronizes the new physical table.
4. After the Alter table or partition ddl, CDC still uses **the schma of the new table** to parse data.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`Fix the issue of cdc does not synchronize related tables after ddl was filtered`.
```


pkg/filter/filter.go
```diff
@@ -150,26 +154,20 @@ func (f *filter) ShouldIgnoreDMLEvent(
 // If a ddl is discarded, it will not be applied to cdc's schema storage
 // and sent to downstream.
 func (f *filter) ShouldDiscardDDL(startTs uint64, ddlType timodel.ActionType, schema, table, query string) (discard bool, err error) {
```
asddongmen, line 156: Please remove the useless parameters from this method.

cdc/owner/ddl_manager.go
```diff
@@ -327,29 +336,34 @@ func (m *ddlManager) shouldExecDDL(nextDDL *model.DDLEvent) bool {
...
-			zap.String("bdrRole", m.executingDDL.BDRRole))
-		tableName := m.executingDDL.TableInfo.TableName
-		// Set it to nil first to accelerate GC.
-		m.pendingDDLs[tableName][0] = nil
-		m.pendingDDLs[tableName] = m.pendingDDLs[tableName][1:]
-		m.schema.DoGC(m.executingDDL.CommitTs - 1)
-		m.justSentDDL = m.executingDDL
-		m.executingDDL = nil
-		m.cleanCache()
+	skip, cleanMsg, err := m.shouldSkipDDL(m.executingDDL)
```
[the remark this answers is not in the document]
CharlesCheung96: Because all ddls must be cached in pengingDDLs until [`m.checkpointTs == nextDDL.CommitTs`](https://github.com/pingcap/tiflow/blob/master/cdc/owner/ddl_manager.go#L298-L328). Otherwise, the changefeed's checkpoint cannot be guaranteed to be correct.

tests/integration_tests/event_filter/data/test_truncate.sql
```diff
@@ -0,0 +1,15 @@
...
+VALUES (1, 'guagua');
+
+INSERT INTO t_truncate
+VALUES (2, 'huahua');
+
+INSERT INTO t_truncate
+VALUES (3, 'xigua');
+
+INSERT INTO t_truncate
+VALUES (4, 'yuko');
```
CharlesCheung96, line 15:
```suggestion
VALUES (4, 'yuko');

```

tests/integration_tests/event_filter/data/test_alter.sql
```diff
@@ -0,0 +1,49 @@
...
+    "key2": "value2"
+  }');
+
+UPDATE t_alter
+SET t_bigint = 888,
+    t_json   = '{
+      "key0": "value0",
+      "key2": "value2"
+    }'
+WHERE id = 2;
```
CharlesCheung96, line 49:
```suggestion
WHERE id = 2;

```

tests/integration_tests/event_filter/data/test.sql
```diff
@@ -48,16 +48,44 @@ CREATE TABLE t2 (
...
+    t_varchar   VARCHAR(10),
+    t_blob      BLOB,
+    t_text      TEXT,
+    t_enum      ENUM ('enum1', 'enum2', 'enum3'),
+    t_set       SET ('a', 'b', 'c'),
+    t_json      JSON,
+    PRIMARY KEY (id)
+) ENGINE = InnoDB
+  DEFAULT CHARSET = utf8
+  COLLATE = utf8_bin;
```
CharlesCheung96, line 91:
```suggestion
  COLLATE = utf8_bin;
  
```

cdc/owner/changefeed.go
```diff
@@ -597,13 +597,13 @@ LOOP2:
 	}
 	c.barriers.Update(finishBarrier, c.latestInfo.GetTargetTs())
 
-	f, err := filter.NewFilter(c.latestInfo.Config, "")
+	filter, err := filter.NewFilter(c.latestInfo.Config, "")
```
3AceShowHand, line 600: `filter` is also the package name, so I renamed it to `f`.  `filter` looks good to me, but report warning by the IDE

cdc/owner/ddl_manager.go
```diff
@@ -327,29 +336,34 @@ func (m *ddlManager) shouldExecDDL(nextDDL *model.DDLEvent) bool {
...
+		return false, "", errors.Trace(err)
+	}
+	if ignored {
+		return true, "ddl is ignored by event filter rule, skip it", nil
+	}
+
+	// In a BDR mode cluster, TiCDC can receive DDLs from all roles of TiDB.
+	// However, CDC only executes the DDLs from the TiDB that has BDRRolePrimary role.
+	if m.BDRMode && ddl.BDRRole != string(ast.BDRRolePrimary) {
+		return true, "changefeed is in BDRMode and the DDL is not executed by Primary Cluster, skip it", nil
```
3AceShowHand, line 351: the second returned value looks used to print log, can we just print the log here, instead of return it ?

3AceShowHand (approved):

hongyunyan (approved):
LGTM

cdc/owner/changefeed.go
CharlesCheung96: Use the package alias `pfilter`, PTAL

cdc/owner/ddl_manager.go
CharlesCheung96: If redo is enabled, shouldSkipDDL will be called twice, so we use the return value to avoid printing duplicate logs.

sdojjy (approved):

Issue #10525, opened 2024-01-23
large transactions block watermark and result in replication lag increasing


zhangjinpeng87:
When there are large transactions running in upstream TiDB cluster, the watermark(resolve ts) will be blocked (explained in PR https://github.com/pingcap/tiflow/pull/10526), this will cause a large replication lag for TiCDC. We should have a good mechanism to smoothly advancing the watermark even there are some large transaction running in upstream TiDB cluster, to let TiCDC can replicate data in time and achieve better RPO.

PR #10526, opened 2024-01-23
[RFC] large transactions don't block watermark advancing

zhangjinpeng87:

Issue Number: close https://github.com/pingcap/tiflow/issues/10525

Propose a better watermark mechanism to advancing watermark smoothly when there are large transactions ongoing in upstream TiDB cluster.

#### Tests 

 - No code

### Release note 

```release-note
`None`.
```


hicqu (approved):

docs/design/2024-01-22-ticdc-large-txn-not-block-wm.md
```diff
@@ -0,0 +1,205 @@
...
+
+![image](../media/large-txn-no-block-wm-1.png)
+
+## Mechanism of updating primary key TTL periodically for large transaction
+
+In TiDB's 2PC transaction protocol, TiDB prewrites all kv pairs concurrently at the prewrite stage, and then commit the primary key first and then asynchrously commit all secondary keys. If a read request encountered a prewritten but uncommitted data, it will try to check its primary key's status(committed or rollbacked, or still ongoing) to determine wait the data to be committed or rollback it. The read request uses the TTL information enveloped in the primary key to determine if this is an aborted transaction caused by crashed TiDB instances or if this is an ongoing transaction if the TTL is still alive.
+For a normal transaction, the default TTL is 3s ahead of the transaction's start-ts. It means if a read request encounters a lock whose primary key's status is undertimed, the read request will check the TTL of this primary key. If the TTL is less than the read request's current ts, the read request can rollback this undertermined lock to make this read request unblocked.
```
CharlesCheung96, line 56:
```suggestion
For a normal transaction, the default TTL is 3s ahead of the transaction's start-ts. It means if a read request encounters a lock whose primary key's status is undetermined, the read request will check the TTL of this primary key. If the TTL is less than the read request's current ts, the read request can rollback this undetermined lock to make this read request unblocked.
```

docs/design/2024-01-22-ticdc-large-txn-not-block-wm.md
```diff
@@ -0,0 +1,205 @@
...
+In order to prevent the read request and other lock resolving mechanisms disturbing or aborting a long running large transaction, the large transaction will periodically update and advance the TTL of its primary key. The following diagram demostrate how large transaction update TTL of its primary key:
+
+![image](../media/large-txn-no-block-wm-2.png)
+
+## Solution: large transactions periodically update its min-commit-ts
+
+From "How large transactions block the advancing of Resolve TS" section we can see the root cause of large transactions block resolve ts is:
+
+- Current Resolve TS(water mark) calculation depends on the start ts of all ongoing transactions, and the start ts of a large transaction is relevant "old" and not advanced when the large transaction is running.
+  More reasonable choice is using min-commit-ts of large transactions to calculate the water mark, and same as the TTL, the large transaction periodically update its primary key's min-commit-tsinformation to let the water mark mechanism can advance the water mark smoothly. The following diagram demostrates the basic idea of this new proposal:
```
CharlesCheung96, line 66:
```suggestion
  More reasonable choice is using min-commit-ts of large transactions to calculate the water mark, and same as the TTL, the large transaction periodically update its primary key's min-commit-ts information to let the water mark mechanism can advance the water mark smoothly. The following diagram demostrates the basic idea of this new proposal:
```

docs/design/2024-01-22-ticdc-large-txn-not-block-wm.md
```diff
@@ -0,0 +1,205 @@
...
+    // Ongoing large transactions related info
+    // start_ts -> (min-commit-ts, primary key) for ongoing large transactions
+    large_txn_map: HashMap<TimeStamp, (TimeStamp, primary key)>
+    // (min-commit-ts, start_ts) set, in order to distinguish different large txns with
+    // the same min-commit-ts value
+    large_txn_min_commit_ts_heap: BTreeSet<(TimeStamp,TimeStamp)>
+}
+```
+
+We separate normal transactions and large transactions in the TS Resolver module. Because each lock contains the transaction-size (or we expliciltely add large-transaction flag in the lock) information, when the Lock Observer receives a lock operation, we can dispatch the lock to related region's Resolver and according to the large transaction flag. We just store a little information like start-ts, primary-key, update-to-date min-commit-ts in the Resolver for a large transaction even though there are 1M rows changed in a region. This can significantly reduce the total memory usage of Resolver for a large transaction which can reduce the OOM issue of TiKV (we previously met some OOM issue because the Resolver occupies a lot of memory for large transactions).
```
CharlesCheung96, line 99:
```suggestion
We separate normal transactions and large transactions in the TS Resolver module. Because each lock contains the transaction-size (or we explicitly add large-transaction flag in the lock) information, when the Lock Observer receives a lock operation, we can dispatch the lock to related region's Resolver and according to the large transaction flag. We just store a little information like start-ts, primary-key, update-to-date min-commit-ts in the Resolver for a large transaction even though there are 1M rows changed in a region. This can significantly reduce the total memory usage of Resolver for a large transaction which can reduce the OOM issue of TiKV (we previously met some OOM issue because the Resolver occupies a lot of memory for large transactions).
```

docs/design/2024-01-22-ticdc-large-txn-not-block-wm.md
```diff
@@ -0,0 +1,205 @@
...
+
+![image](../media/large-txn-no-block-wm-4.png)
+
+### WatermarkAdvancer handles locks
+
+When there is a new added lock for a large transaction, this must be a prewrite request,
+
+- Write lock
+  - if the start-ts of the new added lock is already in current large_txn_map we can skip it directly since we already tracked this large transaction.
+  - If the start-ts of the new added lock is not in current large_txn_map , this must be the first prewrite key in this region(data range), we add it to the large_txn_map
```
CharlesCheung96, line 109:
```suggestion
  - If the start-ts of the new added lock is not in current large_txn_map , this must be the first prewrite key in this region(data range), we add it to the large_txn_map.
```

docs/design/2024-01-22-ticdc-large-txn-not-block-wm.md
```diff
@@ -0,0 +1,205 @@
...
+
+![image](../media/large-txn-no-block-wm-5.png)
+
+The above diagram demonstrates a large transaction with start-ts = 100 write multiple data ranges. Each data range's WatermarkAdvancer will periodically query the min-commit-ts in its primary key for this large transaction.
+
+#### Performance Assessment (range-level request VS store-level request)
+
+You may have noticed that, letting each data range's WatermarkAdvancer proactively query the primary key of this large transaction may introduce a performance issue. For example, if the large transaction involves 100,000 or more data ranges, each WatermarkAdvancer queries the primary key once per second (in order to fetch fresh enough min-commit-ts), the primary key data range needs to serve 100,000 queries.
+In order to decline the query QPS by many magnititude, we use store level requests instead of data range level requests. We can reuse TiKV's 'store level check leader request' to achieve this goal. There are at most N requests per second for the primary key in the above example, where N is the number of TiKV nodes, let's say 200, which represents a relatively large cluster.
```
CharlesCheung96, line 126:
```suggestion
In order to decline the query QPS by many magnitude, we use store level requests instead of data range level requests. We can reuse TiKV's 'store level check leader request' to achieve this goal. There are at most N requests per second for the primary key in the above example, where N is the number of TiKV nodes, let's say 200, which represents a relatively large cluster.
```

docs/design/2024-01-22-ticdc-large-txn-not-block-wm.md
```diff
@@ -0,0 +1,205 @@
...
+Inside CDC, there is an uncommitted data buffer named Matcher for each data range, the matcher caches all prewrite kvs until these kvs receive corresponding commit message and then move them to Sorter one by one.
+The Matcher is a kv-level map, when there is a transaction modified multiple rows, these rows will be moved from the Matcher to the Sorter one by one. Because each row will stay in the Matcher until it received its own commit message. The following diagram left part demostrates a transaction involves key0 and key1, when the Machter receive key1's commit message, it will only move key1 the Sorter (bold part). But at that point of time, the state of this transaction is determined and the commit ts of other keys of this transaction must be 101. key0 will be moved to the Sorter after it received "key0 commit, commit-ts = 101" message.
```
CharlesCheung96, line 136:
```suggestion
The Matcher is a kv-level map, when there is a transaction modified multiple rows, these rows will be moved from the Matcher to the Sorter one by one. Because each row will stay in the Matcher until it received its own commit message. The following diagram left part demostrates a transaction involves key0 and key1, when the Machter receive key1's commit message, it will only move key1 to the Sorter (bold part). But at that point of time, the state of this transaction is determined and the commit ts of other keys of this transaction must be 101. key0 will be moved to the Sorter after it received "key0 commit, commit-ts = 101" message.
```

CharlesCheung96 (approved):

PR #10551, opened 2024-01-26
cli(ticdc): verify start-ts to make sure it's not in the future

### LINKED ISSUE #10499: start-ts is not verified when creating the changefeed
At the moment, if user specifies the `start-ts` when creating the changefeed, it's not checked whether it's valid.

A valid start-ts should be:
1. not smaller than the gc-safepoint
2. not larger than the current tso

If the `start-ts` is in the future, which may caused by mis-operations, the changefeed can be created but cannot make any progress, since the replication is not started yet, this makes users confused.

## How to solve it
There are 2 ways to solve the issue:
1. When creating the changefeed, verify the `start-ts`, make sure it's legal, otherwise report the error and failed to create the changefeed.
2. If the `start-ts` is smaller than the `gc-safepoint`, just report the error, since the changefeed cannot replicate data already gc. If the `start-ts` is in the future, warning the user and need user's confirmation to create the changefeed.

3AceShowHand:

Issue Number: close #10499

When creating and resuming the changefeed, make sure the `start-ts` is not in the future, this should be helpful to prevent the changefeed not make progressing after initialized, since the `start-ts` is not coming yet.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
validate the start-ts when creating and resuming changefeed
```

### DIFF
--- a/cdc/api/v1/validator.go
+++ b/cdc/api/v1/validator.go
@@ -63,13 +63,17 @@ func verifyCreateChangefeedConfig(
 		return nil, cerror.ErrChangeFeedAlreadyExists.GenWithStackByArgs(changefeedConfig.ID)
 	}
 
+	ts, logical, err := up.PDClient.GetTS(ctx)
+	if err != nil {
+		return nil, cerror.ErrPDEtcdAPIError.GenWithStackByArgs("fail to get ts from pd client")
+	}
+	currentTSO := oracle.ComposeTS(ts, logical)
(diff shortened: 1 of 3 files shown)


flowbehappy:
Please explain what does this PR do to the PR description. And I think it needs a release note.

CharlesCheung96 (approved):

hicqu (approved):

PR #10559, opened 2024-01-29
kv (ticdc): fix kvClient reconnection downhill loop

### LINKED ISSUE #10239: TiCDC may incorrectly cancel gRPC streams to TiKV, causing the resolvedTs to become stuck
### What did you do?

We have found there could be many "stream to store closed" errors.
We think it's because of this logic in `cdc/kv/region_worker.go`:
```go
// checkShouldExit checks whether the region worker should exit, if exit return an error
func (w *regionWorker) checkShouldExit() error {
    empty := w.checkRegionStateEmpty()
    // If there is no region maintained by this region worker, exit it and
    // cancel the gRPC stream.
    if empty { /* It should be wrong because pending regions are not considered correctly. */
        w.cancelStream(time.Duration(0))
        return cerror.ErrRegionWorkerExit.GenWithStackByArgs()
    }
    return nil
}
```

_No response_

### What did you see instead?

Pending regions should be considered correctly in `checkShouldExit`.

### Versions of the cluster

6.5.x, 7.1.x.

asddongmen:

Issue Number: close #10239 

1. Add an ID to the `eventFeedStream` struct to identify a stream and prevent it from being deleted unexpectedly.
2. Bind the cancel function of a gRPC stream to its `eventFeedStream` to prevent the stream from being canceled unexpectedly.
3. Reduce the number of calls to `s.deleteStream` to only once to prevent the stream from being deleted unexpectedly.

#### Tests 

1. Add hard code `time.sleep(5 * time.second)` before calling `s.deleteStream` in both unfixed and fixed version of code. 
2. Deploy a TiDB cluster with 3 TiKV nodes and 1 TiCDC server, create a changefeed. 
3. Patch the changed cdc binary, and restart a TiKV node randomly and observe the changefeed's lag. 

unfixed cdc:
img_v3_027f_9b5e20b7-2aab-44b7-8884-fc29d2797beg

fixed cdc:
image
 
From the above graphs, it is evident that in the unfixed CDC, the lag of resolvedTs can exceed 12 minutes when a TiKV node is restarted. However, in the fixed CDC, the increase in resolvedTs is limited to a maximum of 35 seconds. This demonstrates the effectiveness of the fix.

Moreover, when the hard-coded `time.sleep(5 * time.second)` is removed and the fixed version of CDC is tested again, the lag becomes even smaller:
image

### Release note 

```release-note
Fix a bug in kv client that could cause an increase in changefeed lag when TiKV is restarted.
```

### DIFF
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -1444,15 +1444,17 @@ func (s *eventFeedSession) deleteStream(streamToDelete *eventFeedStream) {
 				zap.Uint64("streamIDInMap", streamInMap.id))
 			return
 		}
+
 		if lastTime, ok := s.storeStreamsCache.lastDeleteTime[streamToDelete.addr]; ok {
 			if time.Since(lastTime) < streamDeleteInterval {
-				log.Warn("delete stream failed, delete too frequently, wait 1 second",
+				log.Warn(
+					"delete a stream of a same store too frequently, wait a while and try again",
 					zap.String("namespace", s.changefeed.Namespace),
 					zap.String("changefeed", s.changefeed.ID),
 					zap.Int64("tableID", s.tableID),
 					zap.String("tableName", s.tableName),
 					zap.Uint64("streamID", streamToDelete.id),
-					zap.Duration("duration", time.Since(lastTime)))
+					zap.Duration("sinceLastTime", time.Since(lastTime)))
 				time.Sleep(streamDeleteInterval - time.Since(lastTime))
 			}
 		}
@@ -1460,6 +1462,7 @@ func (s *eventFeedSession) deleteStream(streamToDelete *eventFeedStream) {
 		delete(s.storeStreamsCache.m, streamToDelete.addr)
 		s.storeStreamsCache.lastDeleteTime[streamToDelete.addr] = time.Now()
 	}
+
 	streamToDelete.cancel()
 	log.Info("A stream to store has been removed",
 		zap.String("namespace", s.changefeed.Namespace),
--- a/cdc/kv/client.go
+++ b/cdc/kv/client.go
@@ -1438,6 +1438,7 @@ func (s *eventFeedSession) addStream(stream *eventFeedStream) {
 		}
 	}
 	s.storeStreamsCache.m[stream.addr] = stream
+	s.storeStreamsCache.lastAlterTime[stream.addr] = time.Now()
 }
 
 // deleteStream deletes a stream from the session.streams.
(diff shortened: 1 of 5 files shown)


cdc/kv/client.go
```diff
@@ -752,14 +769,16 @@ func (s *eventFeedSession) requestRegionToStore(
 					zap.String("store", storeAddr),
 					zap.Error(err))
 			}
-			// Delete the stream from the map so that the next time the store is accessed, the stream will be
-			// re-established.
-			s.deleteStream(storeAddr)
+			// TODO(dongmen): remove this line after testing.
+			time.Sleep(time.Second * 5)
```
hicqu, line 773: Do tests still need this? We can use a failpoint.

cdc/kv/region_worker.go
```diff
@@ -595,21 +600,18 @@ func (w *regionWorker) checkErrorReconnect(err error) error {
...
-			zap.String("namespace", w.session.client.changefeed.Namespace),
-			zap.String("changefeed", w.session.client.changefeed.ID))
-	}
+	// cancel the stream to make strem.Recv returns a context cancel error
+	// This will make the receiveFromStream goroutine exit and the stream can
+	// be re-established by the caller.
+	// Note: use context cancel is the only way to terminate a gRPC stream.
+
+	// TODO(dongmen): remove this line after testing.
+	time.Sleep(time.Second * 5)
```
hicqu, line 609: Needs updates.

asddongmen:
I found some others problem in kvClient when I testing this PR. Further investigation is needed.

cdc/kv/client.go
```diff
@@ -232,7 +242,12 @@ func NewCDCClient(
 	return c
 }
 
-func (c *CDCClient) newStream(ctx context.Context, addr string, storeID uint64) (stream *eventFeedStream, newStreamErr error) {
+func (c *CDCClient) newStream(
+	ctx context.Context,
+	cancel context.CancelFunc,
```
3AceShowHand, line 247: pass the `cancel` as a function argument looks so weired.

asddongmen: I have removed it.

asddongmen:
> I found some other problems in kvClient when I testing this PR. Further investigation is needed.

I add a `streamAlterInterval` to prevent a stream of a same store being created/deleted frequently.

hicqu (approved):

3AceShowHand (approved):

PR #10587, opened 2024-02-06
filter (ticdc): fix synpoint table is replicated in BDR mode

### LINKED ISSUE #10576: TiCDC: The sync-point table got replicated (esp. in bdr-mode)
### What did you do?

1. Create two clusters

    ```sh
    tiup playground v7.6.0 --host 10.0.0.2 --db 1 --kv 1 --pd 1 --ticdc 1 --tiflash 0 --without-monitor
    rm /tmp/tidb-4000.sock
    tiup playground v7.6.0 --host 10.0.0.48 --db 1 --kv 1 --pd 1 --ticdc 1 --tiflash 0 --without-monitor
    ```

2. Setup two changefeeds between them

    ```toml
    # 2to48.toml
    bdr-mode = true
    enable-sync-point = true
    sync-point-interval = "31s"
    ```

    ```toml
    # 48to2.toml
    bdr-mode = true
    enable-sync-point = true
    sync-point-interval = "37s"
    ```

    ```sh
    tiup cdc:v7.6.0 cli changefeed create --pd http://10.0.0.2:2379 --sink-uri mysql://root@10.0.0.48 -c 2to48 --config 2to48.toml
    tiup cdc:v7.6.0 cli changefeed create --pd http://10.0.0.48:2379 --sink-uri mysql://root@10.0.0.2 -c 48to2 --config 48to2.toml
    ```

3. Wait for several minutes (because `minSyncPointInterval` is 30s 🤷)

4. Check the syncpoint tables

    ```sh
    mysql --host 10.0.0.2 --port 4000 -u root -e 'select * from tidb_cdc.syncpoint_v1;'
    mysql --host 10.0.0.2 --port 4000 -u root -e 'select count(*) from tidb_cdc.syncpoint_v1;'

    mysql --host 10.0.0.48 --port 4000 -u root -e 'select * from tidb_cdc.syncpoint_v1;'
    mysql --host 10.0.0.48 --port 4000 -u root -e 'select count(*) from tidb_cdc.syncpoint_v1;'
    ```

The syncpoint table in 10.0.0.2 only contained entries of the 48to2 changefeed, 
and similarly 10.0.0.48 only contained entries of the 2to48 changefeed.

### What did you see instead?

Both clusters contained sync-point entries from both changefeeds.

### Versions of the cluster

v7.6.0

asddongmen:

Issue Number: close #10576 

Make isSysSchema() also treat `tidb_cdc` as a system schema. 
This also helps the case of non-BDR "A→B→C" scenario, where the syncpoint table of A→B polluted B→C.
When the schema `tidb_cdc` is considered as a system schema, it will not be listened to by CDC at all.

#### Tests 

 - Unit test

### Release note 

```release-note
Fix a bug that causes syncpoint table being replicated by cdc.
```

### DIFF
--- a/pkg/filter/utils.go
+++ b/pkg/filter/utils.go
@@ -27,7 +27,14 @@ import (
 
 // isSysSchema returns true if the given schema is a system schema
 func isSysSchema(db string) bool {
-	return tifilter.IsSystemSchema(db)
+	switch db {
+	// schema `tidb_cdc` is used in syncpoint function.
+	// Tables in `tidb_cdc` should not be replicated by cdc.
+	case "tidb_cdc":
+		return true
+	default:
+		return tifilter.IsSystemSchema(db)
+	}
 }
 
 // VerifyTableRules checks the table filter rules in the configuration
--- a/pkg/filter/utils.go
+++ b/pkg/filter/utils.go
@@ -28,9 +28,9 @@ import (
 // isSysSchema returns true if the given schema is a system schema
 func isSysSchema(db string) bool {
 	switch db {
-	// schema `tidb_cdc` is used in syncpoint function.
-	// Tables in `tidb_cdc` should not be replicated by cdc.
-	case "tidb_cdc":
+	// TiCDCSystemSchema is used by TiCDC only.
+	// Tables in TiCDCSystemSchema should not be replicated by cdc.
+	case TiCDCSystemSchema:
 		return true
 	default:
 		return tifilter.IsSystemSchema(db)
--- a/pkg/filter/utils_test.go
+++ b/pkg/filter/utils_test.go
@@ -42,6 +42,7 @@ func TestIsSchema(t *testing.T) {
 		{tifilter.InspectionSchemaName, true},
 		{tifilter.PerformanceSchemaName, true},
 		{tifilter.MetricSchemaName, true},
+		{"tidb_cdc", true},
 	}
 	for _, c := range cases {
 		require.Equal(t, c.result, isSysSchema(c.schema))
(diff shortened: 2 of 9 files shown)


CharlesCheung96 (approved):

sdojjy (approved):

pkg/filter/utils.go
```diff
@@ -27,7 +27,14 @@ import (
 
 // isSysSchema returns true if the given schema is a system schema
 func isSysSchema(db string) bool {
-	return tifilter.IsSystemSchema(db)
+	switch db {
+	// schema `tidb_cdc` is used in syncpoint function.
+	// Tables in `tidb_cdc` should not be replicated by cdc.
+	case "tidb_cdc":
```
flowbehappy, line 33: Use a constant var please. And also update the code in syncpoint module.

flowbehappy:
I think we need a BDR test with synpoint enable.

pkg/filter/utils.go
asddongmen: done

asddongmen:
> I think we need a BDR test with synpoint enable.

Added in integration test.

kennytm (approved):

pkg/filter/utils_test.go
```diff
@@ -42,6 +42,7 @@ func TestIsSchema(t *testing.T) {
 		{tifilter.InspectionSchemaName, true},
 		{tifilter.PerformanceSchemaName, true},
 		{tifilter.MetricSchemaName, true},
+		{"tidb_cdc", true},
```
kennytm, line 45: maybe use `TiCDCSystemSchema` instead of `"tidb_cdc"` for this one

asddongmen: Good catch, thank you.

PR #10591, opened 2024-02-06
dm: remove dependency of secret key from dmctl & remove decrypt command

D3Hunter:

Issue Number: ref #9492

- after spec is finalized, we decide to remove dependency of secret key from dmctl, `config export`/`encrypt` sub-commands call api of dm-master
- remove `decrypt` command to avoid user without access to secret key see plaintext password
- some renaming

#### Tests 

 - [x] Unit test
 - [x] Integration test
check password in exported config, it's encrypted if secret-key is provided, otherwise plaintext
 - No code

### Release note 

```release-note

```


dm/common/config/config.go
D3Hunter, line 1: now only dm-master use it, so remove it

dm/config/source_config.go
```diff
@@ -158,20 +158,28 @@ func (c *SourceConfig) Yaml() (string, error) {
...
 }
 
-// Parse parses flag definitions from the argument list.
+// FromToml parses flag definitions from the argument list.
 // accept toml content for legacy use (mainly used by etcd).
-func (c *SourceConfig) Parse(content string) error {
+func (c *SourceConfig) FromToml(content string) error {
 	// Parse first to get config file.
 	metaData, err := toml.Decode(content, c)
-	err2 := c.check(&metaData, err)
```
D3Hunter, line 166: `c.check` is inlined

dm/config/source_config.go
```diff
@@ -241,8 +249,6 @@ func (c *SourceConfig) Verify() error {
 		}
 	}
 
-	c.DecryptPassword()
```
D3Hunter, line 244: this call does nothing useful, so remove it, below calls too

dm/config/task_converters.go
```diff
@@ -664,8 +664,6 @@ func TaskConfigToOpenAPITask(c *TaskConfig, sourceCfgMap map[string]*SourceConfi
 	cfgs := make(map[string]dbconfig.DBConfig)
 	for _, source := range c.MySQLInstances {
 		if cfg, ok := sourceCfgMap[source.SourceID]; ok {
-			// check the password
-			cfg.DecryptPassword()
```
D3Hunter, line 668: here too, no check is done

dm/config/source_config.go
```diff
@@ -158,20 +158,28 @@ func (c *SourceConfig) Yaml() (string, error) {
...
 	// Parse first to get config file.
 	metaData, err := toml.Decode(content, c)
-	err2 := c.check(&metaData, err)
-	if err2 != nil {
-		return err2
+	if err != nil {
+		return terror.ErrWorkerDecodeConfigFromFile.Delegate(err)
+	}
+	undecoded := metaData.Undecoded()
+	if len(undecoded) > 0 && err == nil {
```
lance6716, line 170:
```suggestion
	if len(undecoded) > 0 {
```

seems always err is always nil

lance6716 (approved):

GMHDBJD (approved):
LGTM

PR #10652, opened 2024-02-27
sorter(cdc): correct stale table range cleaning and metric it

### LINKED ISSUE #10641: CDC on disk data size keep increasing during normal replicating and results in CDC disk full
### What did you do?
1. TiDB Cluster with Titan On
```
 config: |
            [cdc]                      
              min-ts-interval = "200ms"
            [raftstore]
              store-io-pool-size = 16
              region-compact-tombstones-percent = 10
              max-leader-missing-duration = "20m"
            [resolved-ts]
              advance-ts-interval = "2s"
            [storage]
              reserve-space = "300GB"
              scheduler-worker-pool-size = 16
            [storage.block-cache]
              num-shard-bits = 4
              capacity = "12GB"
            [quota]
              foreground-write-bandwidth = "40MB"
              max-delay-duration = "1s"
            [rocksdb]
              max-open-files = 10240
              [rocksdb.titan]
                enabled = true
              [rocksdb.defaultcf.titan]
                blob-cache-size = "32GB"
                min-blob-size = "4KB"
                discardable-ratio = 0.4
            [coprocessor]
              region-split-size = "512MiB"
              region-split-keys = 5120000
              region-max-size = "640MiB"
              region-max-keys = 6400000
            [raftdb]
              max-open-files = 10240
            [readpool.unified]
              max-thread-count = 80
              max-tasks-per-worker = 20000
            [server]
              grpc-concurrent-stream = 65535
            [raft-engine]
              enable = true
            [storage.flow-control]
              l0-files-threshold = 45
            [rocksdb.defaultcf]
              level0-stop-writes-trigger = 50
            [gc]
              enable-compaction-filter = false
            [import]
              num-threads = 10
            [backup]
              num-threads = 4
```
1. kafka simple protocol changefeed running to sync a workload whose row width is 10kb, throughput is ~16MB.
```
[sink]
send-bootstrap-interval-in-sec = 120
send-bootstrap-in-msg-count = 10000
send-bootstrap-to-all-partition = false

[sink.kafka-config.codec-config]
encoding-format="avro"

[sink.kafka-config.large-message-handle]
large-message-handle-option = "claim-check"
claim-check-storage-uri = "s3://tmp/redolog/cf-xxx?access-key=minioadmin&secret-access-key=minioadmin&endpoint=http://minio-peer:9000&force-path-style=true"
large-message-handle-compression = "snappy"

[integrity]
integrity-check-level = "correctness"
corruption-handle-level = "error"

[scheduler]
enable-table-across-nodes = true
region-threshold = 10000
write-key-threshold = 3000

```

CDC on disk size should be not too much when CDC changefeed is replicating normally.

### What did you see instead?

We can see cdc-1 on disk data size keep increasing, when changefeed lag is normal, which eventually results in CDC disk full and changefeed error.
image
image
image

### Versions of the cluster

CDC version:
```
[root@tc-ticdc-0 /]# /cdc version
Release Version: v8.0.0-alpha
Git Commit Hash: 833943f79c051d0ae7a0fd157e65f1f07b779348
Git Branch: heads/refs/tags/v8.0.0-alpha
UTC Build Time: 2024-02-23 11:38:20
Go Version: go version go1.21.6 linux/amd64
Failpoint Build: false                                                                                     
```

hicqu:

Issue Number: close #10641 

#10467 introduces a bug that `SortEngine` disk space can't be recycled if redo isn't enabled. This PR corrects it.

And, add a metric to count stale table range cleaning.

#### Tests 

   here is a test for **WITHOUT the patch** and **WITH the patch**:
   图片

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/processor/sinkmanager/redo_log_worker.go
+++ b/cdc/processor/sinkmanager/redo_log_worker.go
@@ -74,13 +74,10 @@ func (w *redoWorker) handleTask(ctx context.Context, task *redoTask) (finalErr e
 	)
 	advancer.lastPos = lowerBound.Prev()
 
-	iter := w.sourceManager.FetchByTable(task.span, lowerBound, upperBound, w.memQuota)
 	allEventCount := 0
 
+	iter := w.sourceManager.FetchByTable(task.span, lowerBound, upperBound, w.memQuota)
 	defer func() {
-		eventCount := newRangeEventCount(advancer.lastPos, allEventCount)
-		task.tableSink.updateRangeEventCounts(eventCount)
-
 		if err := iter.Close(); err != nil {
 			log.Error("redo worker fails to close iterator",
 				zap.String("namespace", w.changefeedID.Namespace),
@@ -88,6 +85,9 @@ func (w *redoWorker) handleTask(ctx context.Context, task *redoTask) (finalErr e
 				zap.Stringer("span", &task.span),
 				zap.Error(err))
 		}
+	}()
+
+	defer func() {
 		log.Debug("redo task finished",
 			zap.String("namespace", w.changefeedID.Namespace),
 			zap.String("changefeed", w.changefeedID.ID),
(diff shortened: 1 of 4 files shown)


flowbehappy:
@hicqu Can you post an example screenshot of the newly added metric in the PR comments?

3AceShowHand (approved):

flowbehappy:
And it is possible to add some unit tests @hicqu ?

asddongmen (approved):

hicqu:
> And it is possible to add some unit tests @hicqu ?

@flowbehappy we have unit tests for the core function. However, the core function caller is in a complex context so it's hard to add unit tests for the caller.

PR #10680, opened 2024-02-29
tiflow 7.6: Bump TiDB dependency to v7.6.0

### LINKED ISSUE #10679: tiflow 7.6: Bump TiDB dependency to v7.6.0
This is to ensure pingcap/tidb#50537 is brought into TiCDC's release-7.6 branch.

kennytm:

Issue Number: close #10679

Upgrade the go.mod version of tidb from pingcap/tidb@54d8a1416ab0 to pingcap/tidb@52794d985ba6 which is same as the tag v7.6.0.

This is required to bring in pingcap/tidb#50537, so that using a populated GCS sink won't cause DeadlineExceeded.

#### Tests 

 - No code

### Release note 

```release-note
None
```


dm/pkg/schema/tracker.go
```diff
@@ -141,7 +167,12 @@ func (tr *Tracker) Init(
 		tableInfos:     make(map[string]*DownstreamTableInfo),
 	}
 	// TODO: need to use upstream timezone to correctly check literal is in [1970, 2038]
-	se := executorContext{Context: mock.NewContext()}
+	seContext := mock.NewContext()
+	seContext.Store = &trackerMockStore{}
```
lance6716, line 171: /cc @okJiang 

can you check this change is expected? Can we move it into mock.NewContext()

kennytm: i think this should be moved into `mock.NewContext()` or make `GetBDRMode()` not requiring a valid `txn` but that would involve updating TiDB first 🥲

okJiang: I think it is unnecessary?
image

https://github.com/pingcap/tidb/pull/50939/files

okJiang: But for v7.6, the pr is not merged.

okJiang: > can you check this change is expected? Can we move it into mock.NewContext()

Yes

lance6716: I just see it targets to release-7.6. We don't need to maintain non LTS versions 😂  If there's an user request I prefer we cherry-pick https://github.com/pingcap/tidb/pull/50939 to TiDB's release-7.6 branch

kennytm: if pingcap/tidb#50939  can be cherry-picked to 7.6 I'd prefer that ;)

PR #10735, opened 2024-03-07
test: add a case of Mariadb for dm_upstream_switch

### LINKED ISSUE #10733: test: dm_upstream_switch doesn't work
https://github.com/pingcap/tiflow/actions/workflows/dm_upstream_switch.yaml

Due to this update https://github.com/pingcap/tiflow/pull/9494

https://github.com/pingcap/tiflow/blob/e5edb1906cc6c1c3addf9e82120ebefd96b13e81/dm/tests/_utils/run_dm_master#L27

### LINKED ISSUE #10734: Mariadb need the test case dm_upstream_switch too

okJiang:

Issue Number: close https://github.com/pingcap/tiflow/issues/10734, close https://github.com/pingcap/tiflow/issues/10733

- Add DM_MASTER_EXTRA_ARG to dm_upstream_switch case
- Add Mariadb dm_upstream_switch case(copy from MySQL&&fix some compatible problems)
- FYI, Mariadb vs MySQL in gtid replication
    - need set `gtid_domain_id`
    - gtid mode is `gtid_strict_mode`
    - change master to need set `master_use_gtid=slave_pos`

#### Tests 

 - Integration test, https://github.com/okJiang/tiflow/actions/runs/8182319412

### Release note 

```release-note
`None`.
```


dm/tests/upstream_switch_mariadb/docker-compose.yml
```diff
@@ -0,0 +1,80 @@
...
+
+# re-usable extension fields, https://docs.docker.com/compose/compose-file/#extension-fields.
+x-node: &default-node
+  privileged: true
+  restart: always
+  environment:
+    MYSQL_ROOT_PASSWORD: "123456"
+
+services:
+  mysql57_master:
```
okJiang, line 17: Can add a comment to explain this naming.
```suggestion
  # In Mariadb case, mysql57 represents mariadb binlog test; mysql8 represents mariadb gtid test. It has nothing to do with the version
  mysql57_master:
```

.github/workflows/dm_upstream_switch.yaml
```diff
@@ -82,3 +74,69 @@ jobs:
...
+          name: upstream-switch-mariadb-logs
+          path: |
+            ./logs
+
+      # send Slack notify if failed.
+      # NOTE: With the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository.
+      - name: Slack notification
+        if: ${{ failure() }}
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_NOTIFY }}
```
GMHDBJD, line 139: no one use slack now, could we use feishu instead

dm/tests/upstream_switch_mariadb/mysql57_master/keepalive.conf
```diff
@@ -0,0 +1,25 @@
+vrrp_script chk_mysql_port {
```
GMHDBJD, line 1: the folder name need change

GMHDBJD (commented):
can we combine it with upstream_swith? e.g. we only need add mariadb pod in original docker file, and add new case in original case.sh

.github/workflows/dm_upstream_switch.yaml
okJiang: lgtm

okJiang:
> can we combine it with upstream_swith? e.g. we only need add mariadb pod in original docker file, and add new case in original case.sh

Solving this problem https://github.com/pingcap/tiflow/issues/10741 later may require modifying the case. If there is no need to modify it, I will merge it together.

.github/workflows/dm_upstream_switch.yaml
okJiang: I'll delete them first
image

Issue #10741, opened 2024-03-08
DM: Mariadb master-slave switch twice, sync failed


okJiang:
### What did you do?

"Mariadb binlog gtid format: {domain_id}{}{server_id}{}{seq_no}

gtid_strict_mode = off (gtid_strict_mode behavior validation: http://youdidwhatwithtsql.com/behavior-gtidstrictmode-mariadb/2089/)

Assumed Example: Data Sync Chain: A -> B -> DM -> TiDB

At time t0:

A's current latest binlog gtid: 0_101_100
B syncs A's latest binlog gtid: 0_101_100
At time t1:

B's slave writes binlog with gtid changes: 0_102_101, 0_102_102
At time t2:

A writes with gtid change: 0_101_101
At time t2:

B backs up gtid sequence:
0_101_100
0_102_101
0_102_102
0_101_101
DM synchronization reports an error: 'less than global checkpoint position.'

At time t3 (Mariadb behavior):

Master-slave switch
B's gtid (A consumption scenario):

0_101_100
0_102_101 (A consumes B)
0_102_102 (A consumes B)
0_101_101 (A Skip)
From verification tests, in mariadb gtid_strict_mode = off mode, when there is a master-slave switch between A and B, B promotes to master, and Mariadb pulls binlog from the master based on its own gtid seq, ignoring seq_no comparisons. In the above example, it consumes data from t1, skipping data already present in 0_101_101."

### What did you expect to see?

sync normally

### What did you see instead?

'less than global checkpoint position.'

### Versions of the cluster

DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):
```console
6.5.3
```
Upstream MySQL/MariaDB server version:
```console
10.1.9
```
Downstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):
```console
(paste TiDB cluster version here)
```
How did you deploy DM: tiup or manually?
```console
(leave TiUP or manually here)
```
Other interesting information (system version, hardware config, etc):
```console
>
>
```

lance6716:
Is this the case of multiple master(writer/primary) node? We should setup separate domain ID for them.

https://mariadb.com/kb/en/gtid/#use-with-multi-source-replication-and-other-multi-primary-setups

However if the master node is caused by failover and multiple master nodes will not occur simultaneously, DM may need to configurate a `gtid_strict_mode = off` behaviour :thinking: . I don't know if that behaviour will cause more trouble like data loss.

[...]

PR #10753, opened 2024-03-11
DM/mariadb: sync the gtid executed in slave during master-slave replication

okJiang:

Issue Number: close https://github.com/pingcap/tiflow/issues/10741

- add test case
- update go-mysql version https://github.com/go-mysql-org/go-mysql/pull/852

#### Tests 

 - Unit test
 - Integration test 
 
~~https://github.com/okJiang/tiflow/actions/runs/8243030893~~
https://github.com/okJiang/tiflow/actions/runs/8290808978

### Release note 

```release-note
`None`
```


dm/tests/mariadb_master_down_and_up/docker-compose.yml
```diff
@@ -0,0 +1,41 @@
...
+        - subnet: "172.28.128.0/24"
+
+# re-usable extension fields, https://docs.docker.com/compose/compose-file/#extension-fields.
+x-node: &default-node
+  privileged: true
+  restart: always
+  environment:
+    MYSQL_ROOT_PASSWORD: "123456"
+
+services:
```
okJiang, line 16:
```suggestion
# In this case, gtid_domain_id should be the same, and gtid_strict_mode should be OFF.
services:
```

GMHDBJD (commented):
can you enable relay？rest LGTM

GMHDBJD (approved):
LGTM

dm/pkg/binlog/event/common.go
```diff
@@ -207,7 +211,9 @@ func verifySingleGTID(flavor string, gSet gmysql.GTIDSet) (interface{}, error) {
 			return nil, terror.ErrBinlogOnlyOneGTIDSupport.Generate(len(mariaGTIDs.Sets), gSet)
 		}
 		var mariaGTID *gmysql.MariadbGTID
-		for _, mariaGTID = range mariaGTIDs.Sets {
+		for _, set := range mariaGTIDs.Sets {
+			for _, mariaGTID = range set {
```
lance6716, line 215: check `len` is 1

dm/pkg/binlog/event/event.go
```diff
@@ -736,21 +736,23 @@ func GenMariaDBGTIDListEvent(header *replication.EventHeader, latestPos uint32,
...
-		// Server_ID, 4 bytes
-		err = binary.Write(payload, binary.LittleEndian, mGTID.ServerID)
-		if err != nil {
-			return nil, terror.ErrBinlogWriteBinaryData.AnnotateDelegate(err, "write Server_ID %d", mGTID.ServerID)
-		}
-		// GTID sequence, 8 bytes
-		err = binary.Write(payload, binary.LittleEndian, mGTID.SequenceNumber)
-		if err != nil {
-			return nil, terror.ErrBinlogWriteBinaryData.AnnotateDelegate(err, "write GTID sequence %d", mGTID.SequenceNumber)
+	for _, set := range mariaDBGSet.Sets {
```
lance6716, line 739: in line 733 there's `numOfGTIDs` which should also be updated. Before fix it, find out why the test didn't reveal the problem.

okJiang: image
This function is used in test only 😅

okJiang: Let me take a closer look at unit test

okJiang: I found the reason why the unit test succeeds. All executions of GenMariaDBGTIDListEvent will execute
https://github.com/pingcap/tiflow/blob/6453fef89e71d3a76dd0b47f784c1e370dfb37e0/dm/pkg/binlog/event/util.go#L162 

This function will decode the event into the "correct" form to compare with other results. There is a high probability that other results will also be decoded through it.

I will add an unit test case for it.

okJiang: https://github.com/pingcap/tiflow/pull/10753/commits/3b094a0358b60403809b17c1c814d50f29887819

dm/pkg/binlog/event/common.go
lance6716: we can check len(set) before the for loop

lance6716 (commented):
rest lgtm

dm/pkg/binlog/event/common.go
okJiang: That can't get the right gtid numbers in the error message

lance6716 (approved):

lance6716 (approved):

PR #10804, opened 2024-03-18
mounter(ticdc): default value convert to the correct data type

### LINKED ISSUE #10803: TiCDC fill the default value without consider the real data type
### What did you do?

```
create table t (a int primary key, b varchar(100));

insert into t values (1, "nitto");
insert into t values (2, "hotd");

alter table t add column c int default 17;

update t set b = "earth" where a = 1;

delete from t where a = 4;
```

The old value part for the column `c`, should have type `int`, instead of `string`

### What did you see instead?

For both update and delete event, the old value is encoded as `string`, instead of real data type `int`. Since TiDB and TiCDC stores the default value as `string`, and TiCDC just use that value without considering the column data type.

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
master
```

3AceShowHand:

Issue Number: close #10803

* convert the datum to the specific column type

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```


hongyunyan (approved):
LGTM

CharlesCheung96 (approved):

PR #10838, opened 2024-03-25
codec(ticdc): simple protocol decoder ignore generated column if the value cannot be found

### LINKED ISSUE #10837: simple protocol decoder should ignore generated column
For the generated column, it won't be replicated, so there is not value for it in the encoded message, just ignore it.

3AceShowHand:

Issue Number: close #10837 

* ignore the column if the value cannot be found from the encoded message.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```

### DIFF
--- a/pkg/sink/codec/simple/message.go
+++ b/pkg/sink/codec/simple/message.go
@@ -501,7 +501,7 @@ func decodeColumns(
 		value, ok := rawData[info.Name.O]
 		if !ok {
 			log.Warn("cannot found the value for the column, "+
-				"it may be a generated column",
+				"it must be a generated column and TiCDC does not replicate generated column value",
 				zap.String("column", info.Name.O))
 			continue
 		}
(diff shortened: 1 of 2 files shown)


hongyunyan (approved):

pkg/sink/codec/simple/message.go
```diff
@@ -500,8 +500,10 @@ func decodeColumns(
 	for _, info := range tableInfo.Columns {
 		value, ok := rawData[info.Name.O]
 		if !ok {
-			log.Panic("cannot found the value for the column",
+			log.Warn("cannot found the value for the column, "+
+				"it may be a generated column",
```
asddongmen, line 504:
```suggestion
				"it must be a generated column and TiCDC does not replicate generated column value",
```

asddongmen (approved):

sdojjy (approved):

PR #10855, opened 2024-03-27
kvClient (ticdc): resolve stale lock more aggressively to reduce resolvedTs lag

asddongmen:

Issue Number: ref #10880 

Reduce the delay of the resolvedTs required to trigger the sending of resolve stale lock requests, adjusting from 20s to 4s. This can more actively request the upstream TiKV to advance the resolvedTs as soon as possible.

In the test environment, after using this PR, the changefeed CheckpointTs lag caused by the rolling restart of TiKV and TiDB is stable between 6-15s.

#### Tests 

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/kv/shared_client.go
+++ b/cdc/kv/shared_client.go
@@ -680,7 +680,9 @@ func (s *SharedClient) resolveLock(ctx context.Context) error {
 			}
 		}
 		start := time.Now()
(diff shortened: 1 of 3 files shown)


asddongmen:
Need to make resolve lock interval configurable.

hicqu (approved):

CharlesCheung96 (approved):

Issue #10856, opened 2024-03-27
TiCDC Puller Resolved Ts calculation uncorrectly when some region's incremental scan is stuck


hongyunyan:
### What did you do?

run a workload and some region stuck in incremental scan

### What did you expect to see?

cdc puller resolved ts is stuck

### What did you see instead?

cdc puller resolved ts increase normally

### Versions of the cluster

Upstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):

```console
(paste TiDB cluster version here)
```

Upstream TiKV version (execute `tikv-server --version`):

```console
(paste TiKV version here)
```

TiCDC version (execute `cdc version`):

```console
(paste TiCDC version here)
```

asddongmen:
It is a metrics issue.

PR #10858, opened 2024-03-28
ticdc: correctly initialize replicationSet to make puller resolved ts calculation correctly

hongyunyan:

Issue Number: close https://github.com/pingcap/tiflow/issues/10856

Initialize ReplicationSet stats with checkpointTs. Only when a table into ReplicationSetStateReplicating state, then owner will update stats based on handleMessage. 

Thus, if we don't initialize stats at the beginning, before the table into ReplicationSetStateReplicating state, it can't join to calculation the minimal puller resolve ts.  It will lead some misleading when some table is stuck in incremental scan.

#### Tests 

I add sleep for a while in incremental scan logic. Then I exam the minimal puller resolved ts before the table created and after the table created. We can see before the table created, the minimal puller resolved ts will stuck, and back to increase continuely after the table is created.
image

 - No code

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/scheduler/internal/v3/replication/replication_set.go
+++ b/cdc/scheduler/internal/v3/replication/replication_set.go
@@ -153,6 +153,36 @@ func NewReplicationSet(
 			CheckpointTs: checkpoint,
 			ResolvedTs:   checkpoint,
 		},
+		// We need to initialize the stats with the checkpoint ts.
+		// Only when the table into ReplicationSetStateReplicating state, we can update the stats
+		// In advanceCheckpoint, it will first check table.Stats.StageCheckpoints["puller-egress"] whether it is nil,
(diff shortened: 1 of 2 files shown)


CharlesCheung96 (approved):

CharlesCheung96:
Great fix !

cdc/scheduler/internal/v3/replication/replication_set.go
```diff
@@ -153,6 +153,36 @@ func NewReplicationSet(
 			CheckpointTs: checkpoint,
 			ResolvedTs:   checkpoint,
 		},
+		// We need to initialize the stats with the checkpoint ts.
```
3AceShowHand, line 156: How about extract this part as a method, to make the current method not too long, hard to read, since the screen is small.

3AceShowHand (approved):

Issue #10862, opened 2024-03-29
Enhance TiCDC Observability


hongyunyan:
Observability is a very important and needed capability of ticdc. More timely and comprehensive monitoring and logs can better help us discover and locate problems in a timely manner.

PR #10863, opened 2024-03-29
ticdc: Add metrics to observe slow initialize region count

hongyunyan:

Issue Number: ref #10862

Add metrics to observe slow initialize region count
image

### Release note 

```release-note
None
```


CharlesCheung96 (approved):

CharlesCheung96:
Please add an example of the new monitoring.

3AceShowHand (approved):

PR #10878, opened 2024-04-03
kv, puller (ticdc):  eliminating unnecessary goroutines, streamlining the kvClient's operation, and reducing code complexity.

asddongmen:

Issue Number: ref #10879

1. Renaming and Adjusting Data Structures: The singleRegionInfo structure has been renamed to regionInfo, and some of its fields have been adjusted. For example, requestedTable has been replaced with subscribedTable, which helps to more accurately reflect the purpose and content of the structure.

2. Function Signature and Logic Updates: Multiple functions have had their signatures modified to use the new regionInfo structure instead of the old singleRegionInfo. The internal logic of the functions has also been updated to accommodate the new data structures and naming.

3. Code Refactoring and Optimization: Redundant goroutines have been removed, and the code logic of kvClient has been simplified to make the code more clear and efficient.

#### Tests 

 - No code

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/kv/region_state.go
+++ b/cdc/kv/region_state.go
@@ -56,6 +56,18 @@ func (s singleRegionInfo) resolvedTs() uint64 {
 	return s.lockedRange.ResolvedTs.Load()
 }
 
+type regionErrorInfo struct {
+	singleRegionInfo
+	err error
+}
+
+func newRegionErrorInfo(info singleRegionInfo, err error) regionErrorInfo {
+	return regionErrorInfo{
+		singleRegionInfo: info,
+		err:              err,
+	}
+}
+
 type regionFeedState struct {
 	sri       singleRegionInfo
 	requestID uint64
--- a/cdc/kv/region_state.go
+++ b/cdc/kv/region_state.go
@@ -75,7 +75,7 @@ func newRegionErrorInfo(info regionInfo, err error) regionErrorInfo {
 }
 
 type regionFeedState struct {
-	sri       regionInfo
+	region    regionInfo
 	requestID uint64
 	matcher   *matcher
 
@@ -93,9 +93,9 @@ type regionFeedState struct {
 	}
 }
 
-func newRegionFeedState(sri regionInfo, requestID uint64) *regionFeedState {
+func newRegionFeedState(region regionInfo, requestID uint64) *regionFeedState {
 	return &regionFeedState{
-		sri:       sri,
+		region:    region,
 		requestID: requestID,
 	}
 }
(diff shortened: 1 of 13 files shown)


cdc/kv/region_state.go
```diff
@@ -31,33 +31,56 @@ const (
 	stateRemoved uint32 = 2
 )
 
-type singleRegionInfo struct {
-	verID  tikv.RegionVerID
+type regionInfo struct {
+	verID tikv.RegionVerID
+	// The span of the region.
+	// Note: The span is not always the real span of the region, it is the span of the
```
3AceShowHand, line 37: this comment is not easy to understand.

cdc/kv/shared_client.go
```diff
@@ -476,57 +491,50 @@ func (s *SharedClient) requestStore(
...
-func (s *SharedClient) broadcastRequest(r *requestedStore, sri singleRegionInfo) {
+func (s *SharedClient) broadcastRequest(r *requestedStore, region regionInfo) {
 	for _, stream := range r.streams {
-		stream.requests.In() <- sri
+		stream.requests.In() <- region
 	}
 }
 
-func (s *SharedClient) handleRequestRanges(ctx context.Context) error {
+func (s *SharedClient) handleRangeTask(ctx context.Context) error {
```
hicqu, line 514: Maybe `handleRangeTasks` is better.

cdc/puller/multiplexing_puller.go
```diff
@@ -254,30 +322,38 @@ func (p *MultiplexingPuller) run(ctx context.Context, includeClient bool) error
 			zap.String("changefeed", p.changefeed.ID))
 	}()
 
-	g, ctx := errgroup.WithContext(ctx)
```
3AceShowHand, line 257: I prefer `g`.

cdc/kv/shared_client.go
```diff
@@ -540,119 +548,129 @@ func (s *SharedClient) divideAndScheduleRegions(
...
 ) {
 	select {
-	case s.requestRangeCh.In() <- rangeTask{span: span, requestedTable: requestedTable}:
+	case s.rangeTaskCh.In() <- rangeTask{span: span, subscribedTable: subscribedTable}:
 	case <-ctx.Done():
 	}
 }
 
-func (s *SharedClient) handleErrors(ctx context.Context) error {
+func (s *SharedClient) handleError(ctx context.Context) error {
```
hicqu, line 656: Maybe `handleErrors` is better. It's a goroutine, not a single call for one region.

cdc/kv/shared_stream.go
```diff
@@ -75,10 +75,10 @@ func newStream(ctx context.Context, c *SharedClient, g *errgroup.Group, r *reque
 			select {
 			case <-ctx.Done():
 				return ctx.Err()
-			case sri := <-stream.requests.Out():
-				if sri.lockedRange != nil {
-					stream.preFetchForConnecting = new(singleRegionInfo)
-					*stream.preFetchForConnecting = sri
+			case region := <-stream.requests.Out():
+				if region.lockedRange != nil {
```
hicqu, line 79: Should use `isStopped`. Please also grep and replace other places.

hicqu:
Rest LGTM.

cdc/kv/region_state.go
CharlesCheung96: Please add comments to explain that this situation is possible only when a single region contains data from multiple tables.

cdc/puller/multiplexing_puller.go
asddongmen: The term "eg" is used elsewhere in the codebase, so I chose to maintain a consistent style here.

3AceShowHand: `g` is also used elsewhere, it's not a big matter, just ignore it. looks good to me

CharlesCheung96 (approved):

hicqu (approved):

PR #10892, opened 2024-04-09
sink(ticdc): limit the number of transactions cached in a mysql worker

### LINKED ISSUE #10896: limit the maximum number of cached txns in mysql worker

ref [sink to mysql (cdc) workload skew issue](https://github.com/pingcap/tiflow/issues/10341)

PR #10379 tries to fix the skew problem by sending a transaction to a random worker after the depended transactions are executed. For conflicting transactions, only one transaction can be executed among all workers at a time, which can also be called serial execution or **one by one**. During synchronous real-time streaming, conflicting transactions are executed serially in the upstream cluster, so it is a reasonable choice for TiCDC to execute these transactions serially. 
```
Txn N(row1)... ------> Txn C(row1)------> Txn B(row1)------> Txn A(row1) 
                                                                    |
                                                                    |-----> worker 1
```

However, for other common scenarios, this approach can be problematic:
1. Incremental Scan: during the synchronization of historical data, if conflicting transactions are consistently executed in the upstream cluster. Then the MySQL sink needs **at least twice the serial throughput** to catch up with the latency. However, the serial execution never satisfies this condition.
```bash
Incremental Scan: Txn N(row1)... ------> Txn C(row1)------> Txn B(row1)------> Txn A(row1) 
                                                                                       |
                                                                                       |-----> worker 1
                                                                                       | 
Real-time streaming: Txn N(row1)... ------> Txn C(row1)------> Txn B(row1)------> Txn A(row1) 
```
2. Cross regional replication: in this scenario, the throughput of a single MySQL worker is limited by network latency, so the throughput of executing transactions one by one may be much smaller than that of the upstream cluster.

### New Proposal
It is better to use a compromise optimization that replaces `one by one` with `batch by batch`:

1. The batch mechanism can effectively improve the throughput of a single worker, so it is necessary to preserve [the fast dependencies resolving optimization](https://github.com/pingcap/tiflow/issues/10341#:~:text=a%20fast%20dependencies,another%20single%20transaction.) in `conflict detector`.  

2. At the same time, to avoid workload skew problems, we could limit the maximum number of cached txns in single worker. When the limit is exceeded, the conflict detector should wait for all transactions cached in the worker to complete before sending a new event to it.

CharlesCheung96:

Issue Number: close #10896

1. Remove the Add method of [`Txn.Worker`](https://github.com/pingcap/tiflow/pull/10892/files#diff-8a11c6b2875e3842bbe7368cde56dd7f4847a3faab72c0b147a879dab0baab05L87) to reduce interdependencies between modules.
2. Use the method of conflictDetector to receive event in [`Txn.Worker`](https://github.com/pingcap/tiflow/pull/10892/files#diff-8a11c6b2875e3842bbe7368cde56dd7f4847a3faab72c0b147a879dab0baab05R91
).
3. Implement [`causality.worker`](https://github.com/pingcap/tiflow/pull/10892/files#diff-3df22c7d15f23c5a539bddeba69f7f295fd02a69366183be498f390772d9f34eR60-R120) that support different batch control policies.

#### Tests 

In the workload skew case, the maximum performance improvement after optimization is about 300% ~ 1000%.

 - Workload: sysbench random update
   1. Before #10376 
image
   2. #10376 
image
   3. With this pr
image

### Release note 

```release-note
`Fix the issue that workload could not be evenly distributed among different mysql sink workers in some scenarios`.
```

### DIFF
--- a/cdc/sink/dmlsink/txn/worker.go
+++ b/cdc/sink/dmlsink/txn/worker.go
@@ -97,8 +97,8 @@ func (w *worker) runLoop(conflictDetector *causality.ConflictDetector[*txnEvent]
 				zap.Int("workerID", w.ID))
 			return nil
 		case txn := <-txnCh:
-			// we get the data from txnCh.out until no more data here or reach the state that can be flushed.
-			// If no more data in txnCh.out, and also not reach the state that can be flushed,
+			// we get the data from txnCh until no more data here or reach the state that can be flushed.
+			// If no more data in txnCh, and also not reach the state that can be flushed,
 			// we will wait for 10ms and then do flush to avoid too much flush with small amount of txns.
 			if txn.TxnEvent != nil {
 				needFlush := w.onEvent(txn.TxnEvent, txn.PostTxnExecuted)
--- a/cdc/sink/dmlsink/txn/worker.go
+++ b/cdc/sink/dmlsink/txn/worker.go
@@ -74,7 +74,7 @@ func newWorker(ctx context.Context, changefeedID model.ChangeFeedID,
 }
 
 // Run a loop.
-func (w *worker) runLoop(conflictDetector *causality.ConflictDetector[*txnEvent]) error {
+func (w *worker) runLoop(txnCh <-chan causality.TxnWithNotifier[*txnEvent]) error {
 	defer func() {
 		if err := w.backend.Close(); err != nil {
 			log.Info("Transaction dmlSink backend close fail",
@@ -88,7 +88,6 @@ func (w *worker) runLoop(conflictDetector *causality.ConflictDetector[*txnEvent]
 		zap.Int("workerID", w.ID))
 
 	start := time.Now()
-	txnCh := conflictDetector.GetOutChByWorkerID(int64(w.ID))
 	for {
 		select {
 		case <-w.ctx.Done():
--- a/cdc/sink/dmlsink/txn/worker.go
+++ b/cdc/sink/dmlsink/txn/worker.go
@@ -22,15 +22,10 @@ import (
 	"github.com/pingcap/tiflow/cdc/model"
 	"github.com/pingcap/tiflow/cdc/sink/metrics/txn"
 	"github.com/pingcap/tiflow/cdc/sink/tablesink/state"
-	"github.com/pingcap/tiflow/pkg/chann"
+	"github.com/pingcap/tiflow/pkg/causality"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 )
-
-type txnWithNotifier struct {
-	*txnEvent
-	postTxnExecuted func()
-}
 
 type worker struct {
 	ctx         context.Context
@@ -38,7 +33,6 @@ type worker struct {
 	workerCount int
 
 	ID      int
-	txnCh   *chann.DrainableChann[txnWithNotifier]
 	backend backend
 
 	// Metrics.
@@ -58,13 +52,13 @@ func newWorker(ctx context.Context, changefeedID model.ChangeFeedID,
 	ID int, backend backend, workerCount int,
 ) *worker {
 	wid := fmt.Sprintf("%d", ID)
+
 	return &worker{
 		ctx:         ctx,
 		changefeed:  fmt.Sprintf("%s.%s", changefeedID.Namespace, changefeedID.ID),
 		workerCount: workerCount,
 
 		ID:      ID,
-		txnCh:   chann.NewAutoDrainChann[txnWithNotifier](chann.Cap(-1 /*unbounded*/)),
 		backend: backend,
 
 		metricConflictDetectDuration: txn.ConflictDetectDuration.WithLabelValues(changefeedID.Namespace, changefeedID.ID),
@@ -79,21 +73,8 @@ func newWorker(ctx context.Context, changefeedID model.ChangeFeedID,
 	}
 }
 
-// Add adds a txnEvent to the worker.
-// The worker will call postTxnExecuted() after the txn executed.
-// The postTxnExecuted will remove the txn related Node in the conflict detector's
-// dependency graph and resolve related dependencies for these transacitons
-// which depend on this executed txn.
-func (w *worker) Add(txn *txnEvent, postTxnExecuted func()) {
-	w.txnCh.In() <- txnWithNotifier{txn, postTxnExecuted}
-}
-
-func (w *worker) close() {
-	w.txnCh.CloseAndDrain()
-}
-
-// Continuously get events from txnCh and call backend flush based on conditions.
-func (w *worker) run() error {
+// Run a loop.
+func (w *worker) runLoop(conflictDetector *causality.ConflictDetector[*txnEvent]) error {
 	defer func() {
 		if err := w.backend.Close(); err != nil {
 			log.Info("Transaction dmlSink backend close fail",
@@ -107,6 +88,7 @@ func (w *worker) run() error {
 		zap.Int("workerID", w.ID))
 
 	start := time.Now()
+	txnCh := conflictDetector.GetOutChByWorkerID(int64(w.ID))
 	for {
 		select {
 		case <-w.ctx.Done():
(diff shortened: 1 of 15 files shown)


pkg/causality/worker.go
```diff
@@ -24,6 +31,90 @@ type txnEvent interface {
...
+		return false
+	}
+}
+
+func (w *boundedWorker[Txn]) out() <-chan TxnWithNotifier[Txn] {
+	return w.ch
+}
+
+// boundedWorkerWithBlock is a special boundedWorker. Once the worker
+// is full, it will block until all cacehed txns are consumed.
```
lidezhu, line 94:
```suggestion
// is full, it will block until all cached txns are consumed.
```

cdc/sink/dmlsink/txn/worker.go
```diff
@@ -107,25 +88,26 @@ func (w *worker) run() error {
...
 	for {
 		select {
 		case <-w.ctx.Done():
 			log.Info("Transaction dmlSink worker exits as canceled",
 				zap.String("changefeedID", w.changefeed),
 				zap.Int("workerID", w.ID))
 			return nil
-		case txn := <-w.txnCh.Out():
+		case txn := <-txnCh:
 			// we get the data from txnCh.out until no more data here or reach the state that can be flushed.
```
asddongmen, line 100: Please correct the comments.

pkg/causality/internal/node.go
```diff
@@ -46,10 +48,11 @@ var (
 // in conflict detection.
 type Node struct {
 	// Immutable fields.
-	id int64
+	id                  int64
+	sortedDedupKeysHash []uint64
 
 	// Called when all dependencies are resolved.
-	OnResolved func(id workerID)
+	TrySendToWorker func(id workerID) bool
```
asddongmen, line 55: IMHO, `OnResolved` is a better name than `TrySendToWorker` because the node doesn't care about the outside logic.

pkg/causality/worker.go
```diff
@@ -24,6 +31,99 @@ type txnEvent interface {
...
+// In current implementation, the conflict detector will push txn to the workerCache.
+type workerCache[Txn txnEvent] interface {
+	// add adds a event to the workerCache.
+	add(txn TxnWithNotifier[Txn]) bool
+	// out returns a channel to receive events which are ready to be executed.
+	out() <-chan TxnWithNotifier[Txn]
+}
+
+func newWorker[Txn txnEvent](opt WorkerOption) workerCache[Txn] {
+	log.Info("create new worker cache in conflict detector",
```
asddongmen, line 63:
```suggestion
	log.Info("create new worker in conflict detector",
```

pkg/causality/worker.go
```diff
@@ -24,6 +31,99 @@ type txnEvent interface {
...
+		zap.Int("cacheSize", opt.CacheSize), zap.Bool("isBlock", opt.IsBlock))
+	if opt.CacheSize <= 0 {
+		log.Panic("WorkerOption.CacheSize should be greater than 0, please report a bug")
+	}
+
+	if opt.IsBlock {
+		return &boundedWorkerCacheWithBlock[Txn]{ch: make(chan TxnWithNotifier[Txn], opt.CacheSize)}
+	}
+	return nil
+	// return &boundedWorkerCache[Txn]{ch: make(chan TxnWithNotifier[Txn], opt.CacheSize)}
```
asddongmen, line 74: Remove it

pkg/causality/worker.go
```diff
@@ -24,6 +31,99 @@ type txnEvent interface {
...
+
+func newWorker[Txn txnEvent](opt WorkerOption) workerCache[Txn] {
+	log.Info("create new worker cache in conflict detector",
+		zap.Int("workerCount", opt.WorkerCount),
+		zap.Int("cacheSize", opt.CacheSize), zap.Bool("isBlock", opt.IsBlock))
+	if opt.CacheSize <= 0 {
+		log.Panic("WorkerOption.CacheSize should be greater than 0, please report a bug")
+	}
+
+	if opt.IsBlock {
```
asddongmen, line 70: What to do if `!opt.IsBlock`?

pkg/causality/worker.go
```diff
@@ -24,6 +31,99 @@ type txnEvent interface {
...
+	// out returns a channel to receive events which are ready to be executed.
+	out() <-chan TxnWithNotifier[Txn]
+}
+
+func newWorker[Txn txnEvent](opt WorkerOption) workerCache[Txn] {
+	log.Info("create new worker cache in conflict detector",
+		zap.Int("workerCount", opt.WorkerCount),
+		zap.Int("cacheSize", opt.CacheSize), zap.Bool("isBlock", opt.IsBlock))
+	if opt.CacheSize <= 0 {
+		log.Panic("WorkerOption.CacheSize should be greater than 0, please report a bug")
```
asddongmen, line 67: It's preferable to use Error instead of Panic, and it's advisable to perform the check earlier.

pkg/causality/worker.go
```diff
@@ -24,6 +31,99 @@ type txnEvent interface {
...
+func (w *boundedWorkerCache[Txn]) add(txn TxnWithNotifier[Txn]) bool {
+	select {
+	case w.ch <- txn:
+		return true
+	default:
+		return false
+	}
+}
+
+//nolint:unused
```
asddongmen, line 94: Why should we retain an unused struct?

pkg/causality/conflict_detector.go
```diff
@@ -16,17 +16,19 @@ package causality
...
 )
 
 // ConflictDetector implements a logic that dispatches transaction
 // to different workers in a way that transactions modifying the same
 // keys are never executed concurrently and have their original orders
 // preserved.
-type ConflictDetector[Worker worker[Txn], Txn txnEvent] struct {
-	workers []Worker
+type ConflictDetector[Txn txnEvent] struct {
+	workers []workerCache[Txn]
```
asddongmen, line 31: In my opinion, it would be more appropriate to rename it as resolvedTxnCaches and include comments to clarify its usage.

pkg/causality/conflict_detector.go
```diff
@@ -16,17 +16,19 @@ package causality
...
 	"sync"
 
+	"github.com/pingcap/log"
 	"github.com/pingcap/tiflow/pkg/causality/internal"
 	"github.com/pingcap/tiflow/pkg/chann"
 	"go.uber.org/atomic"
+	"go.uber.org/zap"
 )
 
 // ConflictDetector implements a logic that dispatches transaction
```
asddongmen, line 26: // ConflictDetector takes in a series of transactions.
// It checks if these transactions depend on each other,
// and sends them to different workerCaches channels.
// Transactions in different channels can be executed concurrently.
// It ensures that transactions changing the same
// rows don't executed concurrently and keep their original order.

pkg/causality/worker.go
```diff
@@ -24,6 +31,99 @@ type txnEvent interface {
...
+	// The PostTxnExecuted will remove the txn related Node in the conflict detector's
+	// dependency graph and resolve related dependencies for these transacitons
+	// which depend on this executed txn.
+	//
+	// NOTE: the PostTxnExecuted() must be called after the txn executed.
+	PostTxnExecuted func()
+}
+
+// WorkerOption is the option for creating a worker.
+type WorkerOption struct {
```
asddongmen, line 46:
```suggestion
type CacheOption struct {
```

pkg/causality/worker.go
```diff
@@ -24,6 +31,99 @@ type txnEvent interface {
...
+	TxnEvent Txn
+	// The PostTxnExecuted will remove the txn related Node in the conflict detector's
+	// dependency graph and resolve related dependencies for these transacitons
+	// which depend on this executed txn.
+	//
+	// NOTE: the PostTxnExecuted() must be called after the txn executed.
+	PostTxnExecuted func()
+}
+
+// WorkerOption is the option for creating a worker.
```
asddongmen, line 45: I think it is better to rename this file to workerCache.

pkg/causality/internal/node.go
CharlesCheung96: The callback function is named differently to avoid confusion, since There are several functions in node that are named resolved.

pkg/causality/worker.go
CharlesCheung96: The golang ling tool has false positives in generic scenarios and I tried to use a new version of the tool and couldn't fix the problem. Therefore, it can only be bypassed through comments.

pkg/causality/worker.go
CharlesCheung96: This panic is prepared to prevent developers from using it incorrectly. Any version that can be successfully built will not have this problem.

asddongmen (approved):

pkg/causality/conflict_detector.go
```diff
@@ -16,17 +16,21 @@ package causality
...
-// preserved.
-type ConflictDetector[Worker worker[Txn], Txn txnEvent] struct {
-	workers []Worker
+// to different worker cache channels in a way that transactions
+// modifying the same keys are never executed concurrently and
+// have their original orders preserved. Transactions in different
+// channels can be executed concurrently.
+type ConflictDetector[Txn txnEvent] struct {
+	// resovedTxnCaches are used to cache resolved transactions.
+	resovedTxnCaches []txnCache[Txn]
```
lidezhu, line 33:
```suggestion
	resolvedTxnCaches []txnCache[Txn]
```

pkg/causality/conflict_detector.go
```diff
@@ -16,17 +16,21 @@ package causality
...
-// keys are never executed concurrently and have their original orders
-// preserved.
-type ConflictDetector[Worker worker[Txn], Txn txnEvent] struct {
-	workers []Worker
+// to different worker cache channels in a way that transactions
+// modifying the same keys are never executed concurrently and
+// have their original orders preserved. Transactions in different
+// channels can be executed concurrently.
+type ConflictDetector[Txn txnEvent] struct {
+	// resovedTxnCaches are used to cache resolved transactions.
```
lidezhu, line 32:
```suggestion
	// resolvedTxnCaches are used to cache resolved transactions.
```

lidezhu (approved):
LGTM

lidezhu:
Need to merge the latest master branch to include the commit ts`c8ed99f6396968d53642d19de22e54f1c8d1d285` to pass the `pull-dm-compatibility-test`. @CharlesCheung96

PR #10907, opened 2024-04-16
ticdc: Support `create table` ddl only appear in tidb_ddl_history instead of tidb_ddl_job table

### LINKED ISSUE #10908: TiCDC support create table accelerate in TiDB
In https://github.com/pingcap/tidb/issues/50052, TiDB do the optimization for creating tables. In new implementations, `create table` ddls will not be inserted into table tidb_ddl_job, but only be inserted into table tidb_ddl_history when `create table` is finished.

Currently TiCDC only listen to tidb_ddl_job table to fetch the ddl changes. So in order to adapt to the new implementations in ddl, we need to also listen to tidb_ddl_history table to fetch the `create table` ddl. 

Fortunately, `create table` ddl still has a eariler CommitTS and StartTs than other ddls in the same table. So we can still use CommitTS and StartTs to sort all ddls in tidb_ddl_job and tidb_ddl_history.

hongyunyan:

Issue Number: close #10908 

1. Support to listen to tidb_ddl_history table in ddlPuller
2. filter other ddls in tidb_ddl_history, and only choose `create table` ddl as the real ddl job
3. remove useless implementations, like listen to DDLJobList etc.

#### Tests 

1. test mannually with `set global tidb_enable_fast_create_table =on` and `set global tidb_enable_fast_create_table =off`. To make sure TiCDC can work well in both old ddl version and new ddl versions.
2. I compared the time it took for TiCDC to process 100 `create table` and `add columns` before and after the modification. Both took 5 mins. Thus, the modification will not affect TiCDC performance.

### Release note 

```release-note
None
```


CharlesCheung96:
Please add some integration tests to cover the `tidb_enable_fast_create_table=on` case.

cdc/puller/ddl_puller_test.go
```diff
@@ -660,15 +657,6 @@ func TestDDLPuller(t *testing.T) {
 	require.Equal(t, resolvedTs, uint64(30))
 	require.Nil(t, ddl)
 
-	inputDDL(t, ddlJobPullerImpl, &timodel.Job{
```
hicqu, line 663: Why we can remove this?

pkg/version/check.go
```diff
@@ -37,13 +37,13 @@ import (
 
 var (
 	// minPDVersion is the version of the minimal compatible PD.
-	minPDVersion = semver.New("5.1.0-alpha")
+	minPDVersion = semver.New("7.1.0-alpha")
```
hicqu, line 40: Please comment why 7.1 is required.

cdc/puller/ddl_puller_test.go
hongyunyan: In the new implementation, it is impossible to get the `create table` with status JobStateCancelled from tidb_ddl_history. Only the `create table` is done, the ddl can be inserted into tidb_ddl_history.

hicqu:
Rest LGTM.

hicqu (approved):

cdc/entry/mounter.go
```diff
@@ -299,38 +312,77 @@ func IsLegacyFormatJob(rawKV *model.RawKVEntry) bool {
...
+		return job, err
+	}
+
+	// DDL job comes from `tidb_ddl_job` table after we support concurrent DDL. We should decode the job from the column.
+	recordID, err := tablecodec.DecodeRowKey(rawKV.Key)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	// first parse it with tidb_ddl_job, if failed, parse it with tidb_ddl_history
```
CharlesCheung96, line 335: Maybe you can decode the tableID from the key?

hongyunyan: Good Suggestion! Fixed now.

CharlesCheung96 (approved):

PR #10919, opened 2024-04-17
*(ticdc): split old update kv entry after restarting changefeed

### LINKED ISSUE #10918: TiCDC meet error: Duplicate entry xxx for key xxx
### What did you do?

If the upstream cluster runs a transaction with multiple updates to the pk/uk like the following:
```
begin;
update t set pk=4 where pk=3;
update t set pk=3 where pk=2;
update t set pk=2 where pk =1;
commit;
```
This transaction may be split into two transactions inside ticdc:
txn1
```
update t set pk=4 where pk=3;
update t set pk=3 where pk=2;
```

txn2
```
update t set pk=2 where pk =1;
```

And these two transactions have no order between them, so they may be executed in any order.
If txn2 executes first, it will meet error like `Duplicate entry xxx for key xxx`;

The transaction can be successfully synced to downstream without error

### What did you see instead?

ticdc report error: `Duplicate entry xxx for key xxx`

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

lidezhu:

Issue Number: close #10918 

1. Split all kv entry of update event which commit ts is older than the ts when changefeed start;
2. Do not split update event inside sink module when the downstream is mysql;
3. Restart changefeed when meet duplicate entry error;
4. When apply redo log, split update events which update handle key to delete events and insert events, and cache the insert events until all delete events in the same transaction are emitted. If the insert events is too many(larger than 50), events will be written to a temp local file;

#### Tests 

 - Integration test
 - Unit test

### Release note 

```release-note
Fix potential risk of data inconsistency when there are dependencies between update statements in the same transaction.
```


sdojjy (approved):

cdc/processor/processor.go
```diff
@@ -486,6 +488,18 @@ func isProcessorIgnorableError(err error) bool {
...
+// needPullerSafeModeAtStart returns true if the scheme is mysql compatible.
+// pullerSafeMode means to split all update kv entries whose commitTS
+// is older then the start time of this changefeed.
+func needPullerSafeModeAtStart(sinkURIStr string) (bool, error) {
+	sinkURI, err := url.Parse(sinkURIStr)
+	if err != nil {
+		return false, cerror.WrapError(cerror.ErrSinkURIInvalid, err)
+	}
+	scheme := sink.GetScheme(sinkURI)
+	return sink.IsMySQLCompatibleScheme(scheme), nil
```
CharlesCheung96, line 500: Other sinks have similar problems, so it's best to have consistent behavior.

lidezhu: 1. Other sinks can't detect error like "duplicate entry", so I think support them here can't prevent the problem, and even doesn't alleviate the problem. So we should find a more systematic way for other sinks if needed;
2. The action to split all update events to delete and insert events at restart time is visible to other sinks, so users of these kinds of sinks may be confused for this kind of changes;

flowbehappy: I suggest we open an new issue to track the overall solution, including for other sink types. And this PR focus on resolving the mysql issue.

asddongmen (approved):

cdc/processor/sourcemanager/manager.go
```diff
@@ -85,13 +92,31 @@ func NewForTest(
 	}
 }
 
+func isOldUpdateKVEntry(raw *model.RawKVEntry, thresholdTs model.Ts) bool {
+	return raw != nil && raw.OpType == model.OpTypePut && raw.OldValue != nil && raw.Value != nil && raw.CRTs < thresholdTs
```
CharlesCheung96, line 96: Maybe we can extract the `RawKVEntry.IsUpdate` method to simplify this.

lidezhu: Fixed.

PR #10932, opened 2024-04-19
mounter(ticdc): timezone fill default value should also consider tz.

### LINKED ISSUE #10931: timestamp fill default value should also consider timezone
#10803, we fix the default value that does not consider the data type.

Since TiDB stores the timestamp value in the UTC timezone, when filling the default value for the timezone type value, we should also convert the default value to the TiCDC specified timezone.

3AceShowHand:

Issue Number: close #10931 

* Ensure to correctly set the default value for the timestamp. It's in UTC and needs to be changed to the timezone set by TiCDC.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```


asddongmen (approved):

hongyunyan (approved):
LGTM

CharlesCheung96:
Is there any change in compatibility?

PR #10942, opened 2024-04-22
simple (ticdc): reduce map access and memory allocation to save some cpu 

asddongmen:
Signed-off-by: dongmen <414110582@qq.com>

Issue Number: ref #10944 

#### MarshalRowChangedEvent Benchmark Test
The benchmark test results below compare performance before and after implementing this PR. The data shows a reduction of approximately `46.4%` in memory allocations, from 13520 B/op to 7242 B/op. This significant decrease can help mitigate the GC overhead and reduce CPU usage of this function.

**After:**
```
goos: linux
goarch: amd64
pkg: github.com/pingcap/tiflow/pkg/sink/codec/simple
cpu: Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz
BenchmarkMarshalRowChangedEvent-16    	   45183	     29715 ns/op	    7242 B/op	     127 allocs/op
```

**Before:**
```
goos: linux
goarch: amd64
pkg: github.com/pingcap/tiflow/pkg/sink/codec/simple
cpu: Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz
BenchmarkMarshalRowChangedEvent-16    	   37898	     35701 ns/op	   13520 B/op	     187 allocs/op
```

#### E2E Performance test 

I conducted an end-to-end (e2e) test, replicating 30 million rows of incremental data from 4000 jitu tables on a TiCDC with a 32-core CPU.

**After**:
The maximum output rate of the sink was 62k rows/s, with an average rate consistently above 60k rows/s.
image

**Before**:
The maximum output rate of the sink was 59k rows/s, with an average rate about 58.5k rows/s.
image

From this, we can conclude that this PR offers a performance improvement of approximately `5%`.
This is expected as this PR solely optimizes a function of the encoder.

 

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


pkg/sink/codec/simple/decoder.go
```diff
@@ -168,7 +168,7 @@ func (d *Decoder) NextRowChangedEvent() (*model.RowChangedEvent, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	log.Debug("row changed event assembled", zap.Any("event", event))
```
3AceShowHand, line 171: this log is useless

pkg/sink/codec/simple/avro.go
```diff
@@ -265,98 +276,104 @@ func (a *avroMarshaller) newDMLMessageMap(
...
 			continue
 		}
+		colInfo := tableInfo.ForceGetColumnInfo(col.ColumnID)
 		value, avroType := a.encodeValue4Avro(col.Value, &colInfo.FieldType)
 		holder := genericMapPool.Get().(map[string]interface{})
 		holder[avroType] = value
-		result[colName] = holder
+		result[colInfo.Name.O] = holder
 	}
-
```
3AceShowHand, line 359: remove empty change

pkg/sink/codec/simple/avro.go
```diff
@@ -265,98 +276,104 @@ func (a *avroMarshaller) newDMLMessageMap(
...
 		holder["com.pingcap.simple.avro.Checksum"] = cc
-		m["checksum"] = holder
+		dmlMessagePayload["checksum"] = holder
 	}
 
 	if event.IsInsert() {
 		data := a.collectColumns(event.Columns, event.TableInfo, onlyHandleKey)
-		m["data"] = data
-		m["type"] = string(DMLTypeInsert)
+		dmlMessagePayload["data"] = data
```
3AceShowHand, line 284: I would prefer use `m`, since it's usage range is quite small, no need to make the variable name too verbose.

pkg/sink/codec/simple/decoder.go
asddongmen: It is helpful in debug. I think it is worth to remain it.

pkg/sink/codec/simple/avro.go
asddongmen: This var is frequently used in this function, and `m` make it hard to understand what it is.

3AceShowHand: You can name it as `payload`, the `dmlMessagePayload` is not good name IMHO, it's too mouthful

3AceShowHand:
Which code block do you think is the most critical part that improves the performance ?

asddongmen:
> Which code block do you think is the most critical part that improves the performance ?

If you are referring to the changes in this pull request, it could be the usage of dmlMessagePool, which reduces memory allocations. However, I didn't test the changes separately, so it's just my guess.

3AceShowHand:
I suggest that you test this PR by using some large table workload which has a lot of columns. I did some similar things before, the result did not meet my expectations, since some map related operations such as `clear` is also not cheap.

asddongmen:
> I suggest that you test this PR by using some large table workload which has a lot of columns. I did some similar things before, the result did not meet my expectations, since some map related operations such as `clear` is also not cheap.

done

3AceShowHand (approved):

CharlesCheung96 (approved):

hongyunyan (approved):
LGTM

PR #11048, opened 2024-05-08
ticdc: fix detecting kafka version 

### LINKED ISSUE #11047: ticdc: detection Kafka version is not the expected version
### What did you do?

create a changefeed with kafka protocol
`bin/cdc cli changefeed create -c test --sink-uri "kafka://127.0.0.1:9092/test?topic=test&protocol=open-protocol&max-message-bytes=41943040&compressionType=ZLIB&kafka-version=0.11.0.0"`

```log
[WARN] [sarama.go:139] ["The Kafka version you assigned may not be correct. Please assign a version equal to or less than the specified version"] [assignedVersion=0.11.0.0] [desiredVersion=2.8.0]
```

### What did you see instead?

```log
[WARN] [sarama.go:226] ["Kafka fail to get ApiVersions"] [addr=127.0.0.1:9092]
```

### Versions of the cluster

```console
v8.0.0
```

```console
v8.0.0
```

```console
v8.2.0-master-dirty
```

Kafka version

```console
kafka_2.13-3.7.0
```

wk989898:

Issue Number: close #11047 

Expose `requestVersion` param and use the low version by default.
After getting version information, set kafka version as high as possible.

#### Tests 

 - Unit test
 - Integration test
 - Manual test
 1. start cluster and Kafka (version=3.7.0)
 2. create a changefeed with kafka protocol
```bash
bin/cdc cli changefeed create -c test --sink-uri "kafka://127.0.0.1:9092/test?topic=test&protocol=open-protocol&max-message-bytes=41943040&compressionType=ZLIB&kafka-version=0.11.0.0"
```
3. check CDC log
```log
[WARN] [sarama.go:139] ["The Kafka version you assigned may not be correct. Please assign a version equal to or less than the specified version"] [assignedVersion=0.11.0.0] [desiredVersion=2.8.0]
```
note: The max Kafka version which sarama detects is `v2.8.0`.

| kafka version |  status |
|   --- | ---     |
| 3.7.0 | success |
| 3.0.1 | success |
| 2.8.0 | success |
| 2.7.0 | success |
| 2.5.0 | success |
| 2.4.0 | success |
| 2.3.1 | success |
| 2.1.1 | success |
| 2.1.0 | success |
| 2.0.1 | success |
| 2.0.0 | success |
| 1.0.1 | success |
| 0.11.0 | success |

### Release note 

```release-note
None
```


3AceShowHand (approved):

asddongmen:
Please update this section in the PR description:
```
Check List
Tests
Unit test
Integration test
Manual test (add detailed scripts or steps below)
No code
```

pkg/sink/kafka/sarama.go
```diff
@@ -202,54 +201,53 @@ func getKafkaVersion(config *sarama.Config, o *Options) (sarama.KafkaVersion, er
...
 	defer func() {
 		broker.Close()
 	}()
 	if err != nil {
-		log.Warn("Kafka fail to open broker", zap.String("addr", addr))
+		log.Warn("Kafka fail to open broker", zap.String("addr", addr), zap.Error(err))
 		return KafkaVersion, err
 	}
-	apiResponse, err := broker.ApiVersions(&sarama.ApiVersionsRequest{Version: 3})
+	apiResponse, err := broker.ApiVersions(&sarama.ApiVersionsRequest{Version: requestVersion})
```
CharlesCheung96, line 223: Please explain the detailed behavior with different requestVersion.

pkg/sink/kafka/options.go
```diff
@@ -151,6 +151,7 @@ type Options struct {
 	ReplicationFactor int16
 	Version           string
 	IsAssignedVersion bool
+	RequestVersion    int16
```
CharlesCheung96, line 154: Do we need to verify the value range of the RequestVersion?

CharlesCheung96:
Please add the test result in pr body.

pkg/sink/kafka/options.go
wk989898: No, it only use in unit test

pkg/sink/kafka/sarama.go
wk989898: mockbroker only works when version  is equal 3, this version is not supported for
low version kafka, so this change always uses the lowest version 0 to connect

pkg/sink/kafka/options.go
CharlesCheung96: Then, maybe we should use a const variable and set it in only in tests.

wk989898:
We need to consider default-kafka-version. It should be set to the lowest version supported by TiCDC. @3AceShowHand @flowbehappy  https://github.com/pingcap/tiflow/blob/1c7c9aec24454baa8624193e0499666fe629a5f3/pkg/sink/kafka/sarama.go#L32

CharlesCheung96 (approved):

pkg/sink/kafka/options.go
wk989898: it will be 0 as default, and the unit test needs to be 3.

PR #11089, opened 2024-05-14
codec(ticdc): simple decoder set the table id on handle key only event and make consumer more strict on the ts order

### LINKED ISSUE #11088: flaky test: simple protocol handle key only may failed
### Which jobs are flaking?

kafka integration test

simple protocol avro handle key only

```log
(paste logs here)
```

3AceShowHand:

Issue Number: close #11088

* set the table id when decode the handle key only event
* consumer panics if commit-ts less than the resolved ts
* update resolved ts after receive resolved ts event and ddl event.

#### Tests 

 - Integration test

### Release note 

```release-note
`None`
```

### DIFF
--- a/cmd/kafka-consumer/main.go
+++ b/cmd/kafka-consumer/main.go
@@ -636,7 +636,7 @@ func (c *Consumer) ConsumeClaim(session sarama.ConsumerGroupSession, claim saram
 				}
 
 				partitionResolvedTs := atomic.LoadUint64(&sink.resolvedTs)
-				if ddl.CommitTs <= partitionResolvedTs {
+				if ddl.CommitTs < partitionResolvedTs {
 					log.Panic("DDL event commit-ts less than the resolved ts",
 						zap.Int32("partition", partition),
 						zap.Uint64("partitionResolvedTs", partitionResolvedTs),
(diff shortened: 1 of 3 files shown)


CharlesCheung96 (approved):

cmd/kafka-consumer/main.go
```diff
@@ -669,43 +674,42 @@ func (c *Consumer) ConsumeClaim(session sarama.ConsumerGroupSession, claim saram
...
+				tableID := row.PhysicalTableID
+				// simple protocol decoder should have set the table id already.
+				if c.option.protocol != config.ProtocolSimple {
+					var partitionID int64
+					if row.TableInfo.IsPartitionTable() {
+						partitionID = row.PhysicalTableID
+					}
+					tableID = c.fakeTableIDGenerator.
+						generateFakeTableID(row.TableInfo.GetSchemaName(), row.TableInfo.GetTableName(), partitionID)
+					row.TableInfo.TableName.TableID = tableID
```
asddongmen, line 704: Why is this needed?

3AceShowHand: to make other protocol, such as canal-json works, they does not encode the table id into the message

asddongmen: Would it still overwrite the actual value of `TableInfo.TableName.TableID` as we saw in the log of failed tests?

3AceShowHand: simple protocol won't hit this code snippets

asddongmen (approved):

cmd/kafka-consumer/main.go
asddongmen: Got it.

PR #11112, opened 2024-05-16
dep(dm): move column-mapping from tidb-tools to tiflow repo

### LINKED ISSUE #11114: move column-mapping from tidb-tools to tiflow repo

lance6716:

Issue Number: close #11114

column-mapping is a dead package, we won't develop its feature in future, so the duplication of code with tidb-tools is fine because the code will not change.

#### Tests 

 - No code

### Release note 

```release-note
`None`.
```


3AceShowHand:
please create new issue for each PR, this makes the cherry-pick process easier.

3AceShowHand (approved):

okJiang (approved):

pkg/column-mapping/column.go
D3Hunter, line 1: are all code copied from tidb repo?

lance6716: tidb-tools. And the test files are changed to `require` package

Issue #11096, opened 2024-05-14
redo apply panic: "insert events should be emitted after other kinds of events in the same transaction"


fubinzh:
### What did you do?

1. create mysql sink changefeed with redo log on
2. run gotpc workload 
4. After 30m, pause changefeed and run redo apply

### What did you expect to see?

redo apply should succeed.

### What did you see instead?

redo apply failed with panic:
```
[...]

PR #11117, opened 2024-05-16
(redo)ticdc: fix the event orderliness in redo log

hongyunyan:

Issue Number: close https://github.com/pingcap/tiflow/issues/11096

1. Fix the less function of logHeap to make dml event of redo log in the same txn sorted as delete/update/insert.
2. use startTs instead of commitTs in redo apply to distinguish different txns.

#### Tests 

 - Unit test
 
    create mysql sink changefeed with redo log on
    run gotpc workload
    After 30m, pause changefeed and run redo apply

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/redo/reader/reader.go
+++ b/cdc/redo/reader/reader.go
@@ -365,10 +365,8 @@ func (h logHeap) Less(i, j int) bool {
 	}
 
 	if h[i].data.RedoRow.Row.CommitTs == h[j].data.RedoRow.Row.CommitTs {
-		if h[i].data.RedoRow.Row.StartTs < h[j].data.RedoRow.Row.StartTs {
-			return true
-		} else if h[i].data.RedoRow.Row.StartTs > h[j].data.RedoRow.Row.StartTs {
-			return false
+		if h[i].data.RedoRow.Row.StartTs != h[j].data.RedoRow.Row.StartTs {
+			return h[i].data.RedoRow.Row.StartTs < h[j].data.RedoRow.Row.StartTs
 		} else {
 			// in the same txn, we need to sort by delete/update/insert order
 			if h[i].data.RedoRow.Row.ToRowChangedEvent().IsDelete() {
--- a/cdc/redo/reader/reader.go
+++ b/cdc/redo/reader/reader.go
@@ -374,14 +374,9 @@ func (h logHeap) Less(i, j int) bool {
 			if h[i].data.RedoRow.Row.ToRowChangedEvent().IsDelete() {
 				return true
 			} else if h[i].data.RedoRow.Row.ToRowChangedEvent().IsUpdate() {
-				if h[j].data.RedoRow.Row.ToRowChangedEvent().IsDelete() {
-					return false
-				} else {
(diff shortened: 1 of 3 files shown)


lidezhu (approved):
LGTM

cdc/redo/reader/reader.go
```diff
@@ -363,10 +364,22 @@ func (h logHeap) Less(i, j int) bool {
 		return false
 	}
 
-	if h[i].data.RedoRow.Row.CommitTs == h[j].data.RedoRow.Row.CommitTs &&
-		h[i].data.RedoRow.Row.StartTs < h[j].data.RedoRow.Row.StartTs {
-		return true
+	if h[i].data.RedoRow.Row.CommitTs == h[j].data.RedoRow.Row.CommitTs {
+		if h[i].data.RedoRow.Row.StartTs < h[j].data.RedoRow.Row.StartTs {
```
CharlesCheung96, line 368: remove useless else branch:
```suggestion
		if h[i].data.RedoRow.Row.StartTs != h[j].data.RedoRow.Row.StartTs {
		return h[i].data.RedoRow.Row.StartTs < h[j].data.RedoRow.Row.StartTs
		}
		...
```

hongyunyan: Good Point!

CharlesCheung96 (approved):

cdc/redo/reader/reader.go
```diff
@@ -363,10 +364,19 @@ func (h logHeap) Less(i, j int) bool {
...
 
-	if h[i].data.RedoRow.Row.CommitTs == h[j].data.RedoRow.Row.CommitTs &&
-		h[i].data.RedoRow.Row.StartTs < h[j].data.RedoRow.Row.StartTs {
-		return true
+	if h[i].data.RedoRow.Row.CommitTs == h[j].data.RedoRow.Row.CommitTs {
+		if h[i].data.RedoRow.Row.StartTs != h[j].data.RedoRow.Row.StartTs {
+			return h[i].data.RedoRow.Row.StartTs < h[j].data.RedoRow.Row.StartTs
+		}
+		// in the same txn, we need to sort by delete/update/insert order
+		if h[i].data.RedoRow.Row.ToRowChangedEvent().IsDelete() {
```
3AceShowHand, line 372: `ToRowChangedEvent` should only be called once, if it may cost some resource.

`else` is redundant I think.

cdc/redo/reader/reader_test.go
```diff
@@ -252,3 +252,515 @@ func genMetaFile(t *testing.T, dir string, meta *common.LogMeta) {
...
+func TestLogHeapLess(t *testing.T) {
+	tests := []struct {
+		name   string
+		h      logHeap
+		i      int
+		j      int
+		expect bool
+	}{
+		{
+			name: "Delete before Update",
```
3AceShowHand, line 265: These data is hard to read, use schema test helper to new events is preferred.

I would suggest that test for each case independently, instead of create a `tests` and loop over each.

PR #11186, opened 2024-05-28
kvclient(cdc): handle out-of-order prewrite events for pipelined DMLs

### LINKED ISSUE #11185: TiCDC may fail if pipelined DML transactions occur during incremental scans
### What did you do?

After Pipelined DML is introduced, we can know there is such a case described in https://github.com/tikv/tikv/pull/16864/files#diff-ad3993a53cda1f90261e58fbea7a450c2752714b16d21c5ede54a9f4bd39a95cR600.

_No response_

### What did you see instead?

Currently it can not be handled by TiCDC.

### Versions of the cluster

```console
```

```console
8.0, 8.1 and later.
```

```console
8.0, 8.1 and later.
```

hicqu:

Issue Number: close #11185 

Handles out-of-order prewrite events, which come from Pipelined DML transactions during incremental scans.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


asddongmen (approved):

cdc/kv/matcher.go
```diff
@@ -62,12 +69,15 @@ func (m *matcher) putPrewriteRow(row *cdcpb.Event_Row) {
...
 	if value, exist := m.unmatchedValue[newMatchKey(row)]; exist {
 		// TiKV may send a fake prewrite event with empty value caused by txn heartbeat.
-		//
 		// We need to skip match if the region is not initialized,
 		// as prewrite events may be sent out of order.
 		if !initialized && len(value.GetValue()) == 0 {
 			return false
 		}
+		// Pipelined-DML transactions can only be matched after initialized.
+		if !initialized && value.Generation > 0 {
```
3AceShowHand, line 78: Is it possible that the `Generation` is 0 ?

3AceShowHand (approved):

cdc/kv/matcher.go
hicqu: Yes. Normal transactions it will be 0. Only in pipelined-dml transacitons it will be greater than 0.

PR #11202, opened 2024-05-30
model(ticdc): remove table id from the table name struct to reduce code complexity

### LINKED ISSUE #11203: simplify the table id related logic
* remove the table id from the table name struct

3AceShowHand:

Issue Number: close #11203 

* remove the `TableID` field from the `TableName` struct， since it is set by the `physicalTableID`, which is duplicate and make the caller confused

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```


cdc/model/schema_storage.go
```diff
@@ -104,7 +104,6 @@ func WrapTableInfo(schemaID int64, schemaName string, version uint64, info *mode
 		TableName: TableName{
 			Schema:      schemaName,
 			Table:       info.Name.O,
-			TableID:     info.ID,
```
CharlesCheung96, line 107: Please remove redunant comments in [TableInfo.TableName](https://github.com/pingcap/tiflow/blob/master/cdc/model/schema_storage.go#L40-L49).

PR #11214, opened 2024-05-31
mysql(ticdc): remove error filter when check isTiDB in backend init

### LINKED ISSUE #11213: ticdc: Removed error type detection in `CheckIsTiDB`
Different downstreams will have different error messages for "select tidb_version()". We cannot provide a complete error whitelist in `CheckIsTiDB`. 
Considering that we just want to know whether the downstream is tidb, the specific error cause is not very important. Therefore, we can directly output error information for all errors instead of returning errors.

hongyunyan:

Issue Number: close #11213

remove error detection in `checkIsTiDB` to avoid unexcept error when the downstream is different mysql-class server

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```

### DIFF
--- a/pkg/sink/mysql/db_helper.go
+++ b/pkg/sink/mysql/db_helper.go
@@ -303,7 +303,7 @@ func CheckIsTiDB(ctx context.Context, db *sql.DB) (bool, error) {
 	err := row.Scan(&tidbVer)
 	if err != nil {
 		log.Error("check tidb version error", zap.Error(err))
-		return false, errors.Trace(err)
+		return false, nil
 	}
 	return true, nil
 }
--- a/pkg/sink/mysql/db_helper.go
+++ b/pkg/sink/mysql/db_helper.go
@@ -303,11 +303,6 @@ func CheckIsTiDB(ctx context.Context, db *sql.DB) (bool, error) {
 	err := row.Scan(&tidbVer)
 	if err != nil {
 		log.Error("check tidb version error", zap.Error(err))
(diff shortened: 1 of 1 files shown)


asddongmen (approved):

pkg/sink/mysql/db_helper.go
```diff
@@ -303,11 +303,6 @@ func CheckIsTiDB(ctx context.Context, db *sql.DB) (bool, error) {
 	err := row.Scan(&tidbVer)
 	if err != nil {
 		log.Error("check tidb version error", zap.Error(err))
-		// downstream is not TiDB, do nothing
-		if mysqlErr, ok := errors.Cause(err).(*dmysql.MySQLError); ok && (mysqlErr.Number == tmysql.ErrNoDB ||
-			mysqlErr.Number == tmysql.ErrSpDoesNotExist || mysqlErr.Number == tmysql.ErrDBaccessDenied) {
-			return false, nil
-		}
 		return false, errors.Trace(err)
```
asddongmen, line 306:
```suggestion
		return false, nil
```

asddongmen:
Please refine this PR description and release note.

CharlesCheung96 (approved):

lidezhu:
If `select tidb_version()` return a network error when the downstream is tidb, it may think the downstream is mysql which is a wrong assumption. And a wrong assumption may cause some potential confusing problem later.
But the chances of this happening are very small, so it is ok to fix it in this way.

PR #11238, opened 2024-06-04
sink(cdc): conflict detector hashes transactions into different table spaces

### LINKED ISSUE #11246: Conflict detector doesn't seperate Txns into different table space
### What did you do?

Starts a changefeed with many tables.

_No response_

### What did you see instead?

Conflict detector doesn't seperate Txns into different table space, which may cause wrong dependencies between Txns from different tables. This may reduce sink throughput.

Here is a result. We can see conflict detect duration is very high:
图片

### Versions of the cluster

TiCDC 6.5.9 and all later versions.

hicqu:

Issue Number: close #11246 

* Make conflict detector hash transactions into different table spaces, to reduce mistake conflicts;
* Log slow conflict detect tables every 60s;
* Conflict detector doesn't need conflict keys to be sorted and deduped.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/sink/dmlsink/txn/event.go
+++ b/cdc/sink/dmlsink/txn/event.go
@@ -52,7 +52,7 @@ func genTxnKeys(txn *model.SingleTableTxn) []uint64 {
 		return nil
 	}
 
-	keys := make([]uint64, 0, 128)
+	keys := make([]uint64, 0, len(txn.Rows))
 	hasher := fnv.New32a()
 	for _, row := range txn.Rows {
 		if uint64(row.GetTableID()) > uint64(math.MaxUint32) {
--- a/cdc/sink/dmlsink/txn/event.go
+++ b/cdc/sink/dmlsink/txn/event.go
@@ -16,6 +16,7 @@ package txn
 import (
 	"encoding/binary"
 	"hash/fnv"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


cdc/sink/dmlsink/txn/event.go
```diff
@@ -51,21 +52,23 @@ func genTxnKeys(txn *model.SingleTableTxn) []uint64 {
...
+		if uint64(row.GetTableID()) > uint64(math.MaxUint32) {
+			log.Panic("TableID shouldn't be larger than math.MaxUint32")
+		}
+		prefix := uint64(row.GetTableID()) << 32
+
 		for _, key := range genRowKeys(row) {
 			if n, err := hasher.Write(key); n != len(key) || err != nil {
 				log.Panic("transaction key hash fail")
 			}
-			hashRes[uint64(hasher.Sum32())] = struct{}{}
```
CharlesCheung96, line 61: Why is the deduplication logic removed here?

pkg/causality/internal/slots.go
```diff
@@ -82,29 +74,19 @@ func (s *Slots) Add(elem *Node) {
 		// Add this node to the slot, make sure new coming nodes with the same hash should
 		// depend on this node.
 		s.slots[slotIdx].nodes[hash] = elem
+
+		s.slots[slotIdx].mu.Unlock()
 	}
 
 	// Construct the dependency graph based on collected `dependencyNodes` and with corresponding
 	// slots locked.
 	elem.dependOn(dependencyNodes)
```
lidezhu, line 83: Calling `dependOn` without locks on the slots is dangerous. Because the `Node` it depends may be removed at the same time. The same dependency may be removed both in `elem.dependOn` and `Slots.Remove`.

cdc/sink/dmlsink/txn/event.go
hicqu: They are not necessary to be deduped now. So we can save this step to reduce CPU and memory usage.

pkg/causality/internal/slots.go
hicqu: The slot lock just protects queue leaders in the slot. This has been already changed at line 76.

lidezhu: ok

flowbehappy:
We need strong evidence for the hash conflict issue before discussion about this PR.

PR #11269, opened 2024-06-06
puller(ticdc): fix wrong update splitting behavior after table scheduling

### LINKED ISSUE #11219: data inconsistency after inject kill all cdc chaos
### What did you do?

1、create BDR cluster
task1：tc->cdc-downstream，run workload for sysbench_64_7000w
task2：cdc-downstream->tc，run workload for test

2、inject some fault to upstream and downstream
img_v3_02bd_60e1f6aa-10d5-4f8b-babe-313923f701fg

3、stop workload

4、compare data between upstream and downstream

the data is consistent

### What did you see instead?

data inconsistency after inject some fault to BDR cluster

img_v3_02bd_a1458a25-bcf8-4be9-81c8-6c7a32256b5g

### Versions of the cluster

./cdc version
 Release Version: v7.5.2
Git Commit Hash: 3625884d98f471feab6b74f72bd2526ba160a50b
Git Branch: HEAD
UTC Build Time: 2024-05-30 09:00:27
Go Version: go version go1.21.10 linux/amd64
Failpoint Build: false
2024-05-31T13:10:18.390+0800

_No response_

lidezhu:

Issue Number: close #11219

After https://github.com/pingcap/tiflow/pull/11030, we introduce a mechanism to get the current timestamp `thresholdTS` from pd when changefeed starts, and split all update kv entries which `commitTS` is smaller than the `thresholdTS`. 

This mechanism has the following problem:
1. There are two cdc nodes `A` and `B`, and `B` start before `A`, that is `thresholdTSB` < `thresholdTSA`;
2. The sync task of table `t` is first on `A`;
3. Table `t` has an update event which `commitTS` is smaller than `thresholdTSA` and larger than `thresholdTSB`. So the update event is split to a delete event and an insert event on node `A`;
4. But the delete event and insert event cannot be send to the downstream in an atomic way. So if after the delete event is send to downstream and before the insert event being send, the table sync task is scheduling to node `B`, the update event are received by node `B` again;
5. The update event is not split by node `B` because its `commitTS` is larger than the `thresholdTSB`, and node `B` just send an update sql to downstream which cause data inconsistency;

And there is also another thing to notice that after scheduling, node `B` will send some events to downstream which are already send by node `A`; So node `B` must send these events in an idempotent way;
Previously, this is handled by getting a `replicateTS` in sink module when sink starts and split these events which `commitTS` is smaller than `replicateTS`. But this mechanism is also removed in https://github.com/pingcap/tiflow/pull/11030. So we need to handle this case in puller too.

In this pr, instead of maintaining a separate `thresholdTS` in sourcemanager, we try to get the `replicateTS` from sink when puller need to check whether to split the update event.
And since puller module starts working before sink module, so we give `replicateTS` a default value `MAXUInt64` which means to split all update events. After sink starts working, `replicateTS` will be set to the correct value.

The last thing to notice, when sink restarts due to some error, after restart, the sink may send some events downstream which are already send before restart. These events also need be send in an idempotent way. But these events are already in sorter, so just restart sink cannot accomplish this goal. So we forbid restarting sink in this pr and just restart the changefeed when meet error.

#### Tests 

 1. deploy a cluster with three cdc nodes;
 2. kill all nodes occasionally while running workload and check whether the data is consistent;

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/processor/sinkmanager/table_sink_worker.go
+++ b/cdc/processor/sinkmanager/table_sink_worker.go
@@ -237,7 +237,7 @@ func (w *sinkWorker) handleTask(ctx context.Context, task *sinkTask) (finalErr e
 		// NOTICE: The event can be filtered by the event filter.
 		if e.Row != nil {
 			// For all rows, we add table replicate ts, so mysql sink can determine safe-mode.
-			e.Row.ReplicatingTs = task.tableSink.replicateTs.Load()
+			e.Row.ReplicatingTs = task.tableSink.GetReplicaTs()
 			x, size := handleRowChangedEvents(w.changefeedID, task.span, e)
 			advancer.appendEvents(x, size)
 			allEventSize += size
--- a/cdc/processor/sinkmanager/table_sink_worker.go
+++ b/cdc/processor/sinkmanager/table_sink_worker.go
@@ -237,7 +237,7 @@ func (w *sinkWorker) handleTask(ctx context.Context, task *sinkTask) (finalErr e
 		// NOTICE: The event can be filtered by the event filter.
 		if e.Row != nil {
 			// For all rows, we add table replicate ts, so mysql sink can determine safe-mode.
-			e.Row.ReplicatingTs = task.tableSink.replicateTs
+			e.Row.ReplicatingTs = task.tableSink.replicateTs.Load()
 			x, size := handleRowChangedEvents(w.changefeedID, task.span, e)
 			advancer.appendEvents(x, size)
 			allEventSize += size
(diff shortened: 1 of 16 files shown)


cdc/processor/sinkmanager/table_sink_worker.go
```diff
@@ -237,7 +237,7 @@ func (w *sinkWorker) handleTask(ctx context.Context, task *sinkTask) (finalErr e
 		// NOTICE: The event can be filtered by the event filter.
 		if e.Row != nil {
 			// For all rows, we add table replicate ts, so mysql sink can determine safe-mode.
-			e.Row.ReplicatingTs = task.tableSink.replicateTs
+			e.Row.ReplicatingTs = task.tableSink.replicateTs.Load()
```
CharlesCheung96, line 240:
```suggestion
			e.Row.ReplicatingTs = task.tableSink.GetReplicaTs()
```

CharlesCheung96 (approved):

asddongmen (approved):

PR #11296, opened 2024-06-11
puller(ticdc): fix wrong update splitting behavior after table scheduling

lidezhu:

Issue Number: close pingcap/tiflow#11219

1. There are two cdc nodes `A` and `B`, and `B` start before `A`, that is `thresholdTSB` < `thresholdTSA`;
2. The sync task of table `t` is first on `A`;
3. Table `t` has an update event which `commitTS` is smaller than `thresholdTSA` and larger than `thresholdTSB`. So the update event is split to a delete event and an insert event on node `A`;
4. But the delete event and insert event cannot be send to the downstream in an atomic way. So if after the delete event is send to downstream and before the insert event being send, the table sync task is scheduling to node `B`, the update event are received by node `B` again;
5. The update event is not split by node `B` because its `commitTS` is larger than the `thresholdTSB`, and node `B` just send an update sql to downstream which cause data inconsistency;

And there is also another thing to notice that after scheduling, node `B` will send some events to downstream which are already send by node `A`; So node `B` must send these events in an idempotent way;
Previously, this is handled by getting a `replicateTS` in sink module when sink starts and split these events which `commitTS` is smaller than `replicateTS`. But this mechanism is also removed in https://github.com/pingcap/tiflow/pull/11030. So we need to handle this case in puller too.

In this pr, instead of maintaining a separate `thresholdTS` in sourcemanager, we try to get the `replicateTS` from sink when puller need to check whether to split the update event.
And since puller module starts working before sink module, so we give `replicateTS` a default value `MAXUInt64` which means to split all update events. After sink starts working, `replicateTS` will be set to the correct value.

The last thing to notice, when sink restarts due to some error, after restart, the sink may send some events downstream which are already send before restart. These events also need be send in an idempotent way. But these events are already in sorter, so just restart sink cannot accomplish this goal. So we forbid restarting sink in this pr and just restart the changefeed when meet error.

#### Tests 

 1. deploy a cluster with three cdc nodes;
 2. kill all nodes occasionally while running workload and check whether the data is consistent;

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/processor/sinkmanager/manager.go
+++ b/cdc/processor/sinkmanager/manager.go
@@ -304,7 +304,7 @@ func (m *SinkManager) Run(ctx context.Context, warnings ...chan<- error) (err er
 			}
 
 			if m.isMysqlBackend {
-				// For MySQL backend, we should restart sink. Let owner to handle the error.
+				// For MySQL backend, we should restart changefeed. Let owner to handle the error.
 				return errors.Trace(err)
 			}
 		}
--- a/cdc/processor/sinkmanager/manager.go
+++ b/cdc/processor/sinkmanager/manager.go
@@ -128,6 +128,9 @@ type SinkManager struct {
 	// wg is used to wait for all workers to exit.
 	wg sync.WaitGroup
 
+	// isMysqlBackend indicates whether the backend is MySQL compatible.
+	isMysqlBackend bool
+
 	// Metric for table sink.
 	metricsTableSinkTotalRows prometheus.Counter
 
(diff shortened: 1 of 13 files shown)


cdc/processor/sinkmanager/manager.go
```diff
@@ -298,6 +302,11 @@ func (m *SinkManager) Run(ctx context.Context, warnings ...chan<- error) (err er
 			if cerror.IsDupEntryError(err) {
 				return errors.Trace(err)
 			}
+
+			if m.isMysqlBackend {
+				// For MySQL backend, we should restart sink. Let owner to handle the error.
```
CharlesCheung96, line 307:
```suggestion
				// For MySQL backend, we should restart changefeed. Let owner to handle the error.
```

CharlesCheung96 (approved):

asddongmen (approved):

PR #11339, opened 2024-06-24
processor: fix a bug that will cause processor Tick get stuck when downstream is Kafka

### LINKED ISSUE #11340: etcd worker tick stuck up to 2 minutes periodically when cdc can't connect to Kafka server
### What did you do?

1. Creata a changefeed sink to kafka.
2. Inject a network partition between cdc and kafka.

changefeed stucks but processor will not be stuck. 

### What did you see instead?

```
[2024/06/21 19:03:54.617 +08:00] [WARN] [etcd_worker.go:293] ["EtcdWorker reactor tick took too long"] [duration=50.822188739s] [role=processor]
[2024/06/21 20:04:52.975 +08:00] [WARN] [processor.go:540] ["processor tick took too long"] [namespace=default] [changefeed=kafka-task] [capture=4556c672-6e93-4054-a705-c598fcf64b84] [duration=2m1.979526934s]


```

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
master
```

asddongmen:

Issue Number: close #11340 

#### Tests 
 - Unit test

Deploy a cdc cluster, create two changefeeds and synchronize the same table to both Kafka and TiDB separately.

Every 20 minutes, inject CDC to network isolation:

- First, isolate cdc to Kafka network for 10 minutes.
- Then, 20 minutes later, isolate cdc to downstream TiDB network for 10 minutes.

Repeat this process six times.

Before this PR, the lag in the MySQL sink changefeed was influenced by the Kafka sink changefeed when network isolation was introduced between CDC and Kafka:
nHHvw2tHOE

After this PR, only Kafka sink changefeed lag increasing during network isolation injection:
image

### Release note 

```release-note
Fix a bug that causes the processor tick to get stuck when the downstream is Kafka and becomes unreachable.```

### CODE REVIEW

File: pkg/sink/kafka/sarama_factory.go [hunk 1]
```diff
@@ -43,17 +46,32 @@ func NewSaramaFactory(
 }
 
 func (f *saramaFactory) AdminClient(ctx context.Context) (ClusterAdminClient, error) {
+	start := time.Now()
 	config, err := NewSaramaConfig(ctx, f.option)
+	duration := time.Since(start).Seconds()
+	if duration > 2 {
+		log.Warn("new sarama config cost too much time", zap.Any("duration", duration), zap.Stringer("changefeedID", f.changefeedID))
```
CharlesCheung96, line 53: please use debug log here

asddongmen: I believe this should be a warning or info level log. A two-second threshold is quite large and this log could help identify potential issues.

cdc/processor/sinkmanager/manager.go
```diff
@@ -93,7 +94,7 @@ type SinkManager struct {
 		// When every time we want to create a new factory, version will be increased and
 		// errors will be replaced by a new channel. version is used to distinct different
 		// sink factories in table sinks.
-		version uint64
+		version atomic.Uint64
```
hicqu, line 97: version is protected by a lock. Seems it's unnecessary to be an atomic.

asddongmen: I added comments, pointing out that this lock is only used to protect the f field. I think this can reduce the mental burden of using this lock and prevent similar blocking problems in the future.

cdc/processor/sinkmanager/manager.go
```diff
@@ -88,6 +87,7 @@ type SinkManager struct {
 
 	// sinkFactory used to create table sink.
 	sinkFactory struct {
+		// This lock is only used to protect the f field.
```
hicqu, line 90: The line can be removed.

hicqu (approved):

pkg/config/sink.go
```diff
@@ -178,6 +178,7 @@ type SinkConfig struct {
 
 	// AdvanceTimeoutInSec is a duration in second. If a table sink progress hasn't been
 	// advanced for this given duration, the sink will be canceled and re-established.
+	// Deprecated from v8.1.1.
```
3AceShowHand, line 181:
```suggestion
	// Deprecated since v8.1.1.
```

pkg/sink/kafka/sarama_factory.go
3AceShowHand: which condition may cost too much time on new the sarama config?

asddongmen: Network partition.

3AceShowHand (approved):

PR #11370, opened 2024-07-02
mounter(ticdc): fix decoding when upstream doesn't enable new collation

### LINKED ISSUE #11371: TiCDC meet duplicate entry error when run ycsb insert workload
### What did you do?

1. set tidb config `new_collations_enabled_on_first_bootstrap` to false;
2. create changefeed;
3. run bin/ycsb load mysql ...

TiCDC can sync data correctly.

### What did you see instead?

TiCDC report error like duplicate entry ''

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```
nightly

lidezhu:

Issue Number: close #11371 

Explicitly call `collate.SetNewCollationEnabledForTest(false)` before run cdc.

Previously, this function was called in a tidb package. But it was removed as described in https://github.com/pingcap/tidb/pull/52191#issuecomment-2024836481.

If the upstream doesn't enable new collation, the primary key is not encoded in the value part when the table is a cluster index table. And we rely on the following line to decode the primary key. https://github.com/pingcap/tiflow/blob/2db092244d18c043f7f8ecc85a598680b33cc324/cdc/entry/mounter.go#L301

But this line only decode the primary key when the global variable `newCollationEnabled` in tidb package is false. (It means `collate.SetNewCollationEnabledForTest(false)` must be called.)

#### Tests 

 - Integration test

### Release note 

```release-note
Fix the issue that TiCDC cannot correctly decode primary key of table with cluster index when the upstream does not enable new collation.
```


cmd/cdc/main.go
```diff
@@ -15,9 +15,12 @@ package main
 
 import (
 	_ "github.com/pingcap/tidb/pkg/types/parser_driver"
+	"github.com/pingcap/tidb/pkg/util/collate"
 	"github.com/pingcap/tiflow/pkg/cmd"
 )
 
 func main() {
+	// NOTE: the line is removed from TiDB repo in https://github.com/pingcap/tidb/pull/52191#issuecomment-2024836481.
```
flowbehappy, line 23: Add more comment to explain why we need to call this function.

lidezhu: Added.

CharlesCheung96 (approved):

asddongmen (approved):

PR #11401, opened 2024-07-11
sink(ticdc): split update event also set the checksum field to 0

### LINKED ISSUE #11402: update event splitted without considering checksum field
### What did you do?

```
create table t (
   a varchar(5),
   b int,
   c text,

   primary key (a, b)
);

insert into t values ("a", 1, "a1");
update t set a = 'b' where b = 1;
```

`update t set a = 'b' where b = 1;` generate one update event, but splitted into delete event and insert event, since the primary key column `a` changed.

In the delete event, the checksum.Current should be 0. In the insert event, the checksum.Previous should be 0

### What did you see instead?

the checksum fields is not changed, this makes the consumer confusing.

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
master
```

3AceShowHand:

Issue Number: close #11402 

* set the checksum field to 0 after split the update event

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
fix the checksum field not set to 0 after split update event
```

### DIFF
--- a/cdc/model/sink.go
+++ b/cdc/model/sink.go
@@ -1252,10 +1252,16 @@ func SplitUpdateEvent(
 	// so it won't have an impact and no more full deep copy wastes memory.
 	deleteEvent := *updateEvent
 	deleteEvent.Columns = nil
+	if deleteEvent.Checksum != nil {
+		deleteEvent.Checksum.Current = 0
+	}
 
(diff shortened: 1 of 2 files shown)


3AceShowHand:
close this

PR #11443, opened 2024-07-30
ticdc: Fix potential SQL injection due to string-formatted SQL query

### LINKED ISSUE #11433: ticdc: Potential SQL Injection Due to String-Formatted SQL Query
risk description:
The impact of SQL injection can be severe, compromising an application's security. Attackers can exploit this vulnerability to perform unauthorized actions on a database, such as accessing sensitive data, modifying or deleting data, and executing administrative operations on the database. This can lead to data breaches, data integrity loss, and service denial. In some cases, it may also be used to gain unauthorized access to host systems or perform other attacks.

suggestion: Use parameterized queries or prepared statements. Prepared statements can be used with the 'Prepare' and 'PrepareContext' calls.

wk989898:

Issue Number: close #11433

Use prepared statements to process SQL query.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
+++ b/cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
@@ -199,8 +199,12 @@ func (m *DDLSink) execDDL(pctx context.Context, ddl *model.DDLEvent) error {
 	}
 
 	if shouldSwitchDB {
-		_, err = tx.ExecContext(ctx, "USE "+quotes.QuoteName(ddl.TableInfo.TableName.Schema)+";")
+		stmt, err := tx.PrepareContext(ctx, "USE "+quotes.QuoteName(ddl.TableInfo.TableName.Schema)+";")
 		if err != nil {
+			log.Error("Failed to creates a prepared statement", zap.String("query", "USE "+quotes.QuoteName(ddl.TableInfo.TableName.Schema)+";"))
+			return err
+		}
+		if _, err = stmt.ExecContext(ctx); err != nil {
 			if rbErr := tx.Rollback(); rbErr != nil {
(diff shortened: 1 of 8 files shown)


cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -199,8 +199,12 @@ func (m *DDLSink) execDDL(pctx context.Context, ddl *model.DDLEvent) error {
 	}
 
 	if shouldSwitchDB {
-		_, err = tx.ExecContext(ctx, "USE "+quotes.QuoteName(ddl.TableInfo.TableName.Schema)+";")
+		stmt, err := tx.PrepareContext(ctx, "USE "+quotes.QuoteName(ddl.TableInfo.TableName.Schema)+";")
 		if err != nil {
+			log.Error("Failed to creates a prepared statement", zap.String("query", "USE "+quotes.QuoteName(ddl.TableInfo.TableName.Schema)+";"))
+			return err
+		}
+		if _, err = stmt.ExecContext(ctx); err != nil {
```
kennytm, line 207: 1. it is obvious that this statement has no SQL injection opportunity
2. even if it is possible to inject through `QuoteName()`, I don't see how changing `Exec` to `Prepare` the entire statement unchanged and then `Exec` it is going to prevent anything.

`PrepareContext` can prevent SQL injection only when you replace e.g. `Exec("select x where y = " + val)` to `Prepare("select x where y = ?").Exec(val)`[^1]. Simply changing it to `Prepare("select x where y = " + val).Exec()` is meaningless.

[^1]: in Go you could even short-hand it as `Exec("select x where y = ?", val)`.

kennytm (changes requested):

PR #11469, opened 2024-08-05
mounter(ticdc): do not report error is verify old value checksum v1 failed

### LINKED ISSUE #11468: do not report error is verify old value checksum v1 failed
The old value checksum v1 cannot correctly handle old value checksum verification after `Add Column / Drop Column` executed.

Let's take the following case as an example, the TiDB / TiCDC cluster is deployed in v7.5.0, then upgrade to v8.4.0

```
create table t (a int primary key, b int);

insert into t values (1, 2);   -> v7.5 write in, checksum v1

alter table t add column c int default 3;

-- upgrade cluster -> v8.4.0

update t set b = 1 where a = 1;  -> v8.4 write in

old value  -> v1

value  -> v2

```

The old value is written in by checksum v1, cannot be handled correctly, and makes the changefeed error.

The changefeed do not report error and do not stop replication, just skip old value checksum v1 verification.

3AceShowHand:

Issue Number: close #11468

* if verification on the old value checksum v1 failed, just skip it, print some logs to warn it.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```

### DIFF
--- a/cdc/entry/mounter.go
+++ b/cdc/entry/mounter.go
@@ -100,6 +100,8 @@ type mounter struct {
 	// they should not be nil after decode at least one event in the row format v2.
 	decoder    *rowcodec.DatumMapDecoder
 	preDecoder *rowcodec.DatumMapDecoder
+
+	lastSkipOldValueTime time.Time
 }
 
 // NewMounter creates a mounter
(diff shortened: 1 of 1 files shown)


hongyunyan (approved):
LGTM

asddongmen (approved):

flowbehappy (commented):
LGTM

PR #11476, opened 2024-08-08
sink, ddl(ticdc): support add index ddl in downstream

### LINKED ISSUE #10682: drop index after add index finishing on primary cluster, drop index was retried repeatedly by ticdc due to add index was running on secondary and after a time, changefeed status was failed
### What did you do?

1、restore data for primary and secondary
2、create changefeed and set bdr role for primary and secondary
3、run sysbench on primary and secondary
4、add index and then drop index when add index finished on primary
5、inject network partition between one of tikv and all other pods, which last for 3mins and recover
 chaos recover time :  2024-02-29 17:29:24

ticdc logs:
[endless-ha-test-ticdc-tps-7080582-1-976-tc-ticdc-0.tar.gz](https://github.com/pingcap/tiflow/files/14447186/endless-ha-test-ticdc-tps-7080582-1-976-tc-ticdc-0.tar.gz)

changefeed status are normal
after recover fault, ddl can sync success and changefeed lag can recover to normal

### What did you see instead?

1、drop index was was retried repeatedly even if drop index was queueing on secondary due to add index was running
primary：
img_v3_028h_268d8fab-fb75-4038-89cb-01f630ea404g

secondary：
img_v3_028h_3e24f88d-28af-4d65-980c-b274339c741g

2、after a time，changefeed status was failed
changefeed status：
img_v3_028h_ec4db81a-3ac8-4491-9f75-c1b8dd8e638g

### Versions of the cluster

./cdc version
 Release Version: v8.0.0-alpha
Git Commit Hash: 25ce29c2a1802bbb4cd26008f322728959a91f7a
Git Branch: heads/refs/tags/v8.0.0-alpha
UTC Build Time: 2024-02-27 11:37:29
Go Version: go version go1.21.6 linux/amd64
Failpoint Build: false

_No response_

CharlesCheung96:

Issue Number: close #10682

- [Always wait for async ddl done before executing the next normal ddl](https://github.com/pingcap/tiflow/pull/11476/files#diff-67001e77c3d88c6c47cfed312c5b1f5f36d2a404163aefb5da2939fce5051674R96)
- Note: [wait indefinitely](https://github.com/pingcap/tiflow/pull/11476/files#diff-67001e77c3d88c6c47cfed312c5b1f5f36d2a404163aefb5da2939fce5051674R111-R121) here to avoid pointless retries
- Any errors encountered will be [ignored](https://github.com/pingcap/tiflow/pull/11476/files#diff-67001e77c3d88c6c47cfed312c5b1f5f36d2a404163aefb5da2939fce5051674R127-R133) in the check, and if the add index operation in downstream is not complete at this point, TiCDC will fall back to the current master's behavior.

#### Tests 

 - Unit test
 - Manual test 

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/sink/ddlsink/mysql/async_ddl.go
+++ b/cdc/sink/ddlsink/mysql/async_ddl.go
@@ -1,4 +1,4 @@
-// Copyright 2022 PingCAP, Inc.
+// Copyright 2024 PingCAP, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cdc/sink/ddlsink/mysql/async_ddl.go
+++ b/cdc/sink/ddlsink/mysql/async_ddl.go
@@ -1,0 +1,185 @@
+// Copyright 2022 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mysql
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/pingcap/log"
+	timodel "github.com/pingcap/tidb/pkg/parser/model"
+	"github.com/pingcap/tiflow/cdc/model"
+	"github.com/pingcap/tiflow/pkg/errors"
+	"go.uber.org/zap"
+)
+
+var checkRunningAddIndexSQL = `
+SELECT JOB_ID, JOB_TYPE, SCHEMA_STATE, SCHEMA_ID, TABLE_ID, STATE, QUERY
+FROM information_schema.ddl_jobs
+WHERE DB_NAME = "%s" 
+    AND TABLE_NAME = "%s"
+    AND JOB_TYPE LIKE "add index%%"
+    AND (STATE = "running" OR STATE = "queueing")
+LIMIT 1;
+`
+
+func (m *DDLSink) shouldAsyncExecDDL(ddl *model.DDLEvent) bool {
+	return m.cfg.IsTiDB && ddl.Type == timodel.ActionAddIndex
+}
+
+// asyncExecDDL executes ddl in async mode.
+// this function only works in TiDB, because TiDB will save ddl jobs
+// and execute them asynchronously even if ticdc crashed.
+func (m *DDLSink) asyncExecDDL(ctx context.Context, ddl *model.DDLEvent) error {
+	done := make(chan error, 1)
+	// Use a longer timeout to ensure the add index ddl is sent to tidb before executing the next ddl.
+	tick := time.NewTimer(10 * time.Second)
+	defer tick.Stop()
+	log.Info("async exec add index ddl start",
+		zap.String("changefeedID", m.id.String()),
+		zap.Uint64("commitTs", ddl.CommitTs),
+		zap.String("ddl", ddl.Query))
+	go func() {
+		if err := m.execDDLWithMaxRetries(ctx, ddl); err != nil {
+			log.Error("async exec add index ddl failed",
+				zap.String("changefeedID", m.id.String()),
+				zap.Uint64("commitTs", ddl.CommitTs),
+				zap.String("ddl", ddl.Query))
+			done <- err
+			return
+		}
+		log.Info("async exec add index ddl done",
+			zap.String("changefeedID", m.id.String()),
+			zap.Uint64("commitTs", ddl.CommitTs),
+			zap.String("ddl", ddl.Query))
+		done <- nil
+	}()
+
+	select {
+	case <-ctx.Done():
+		// if the ddl is canceled, we just return nil, if the ddl is not received by tidb,
+		// the downstream ddl is lost, because the checkpoint ts is forwarded.
+		log.Info("async add index ddl exits as canceled",
+			zap.String("changefeedID", m.id.String()),
+			zap.Uint64("commitTs", ddl.CommitTs),
+			zap.String("ddl", ddl.Query))
+		return nil
+	case err := <-done:
+		// if the ddl is executed within 2 seconds, we just return the result to the caller.
+		return err
+	case <-tick.C:
+		// if the ddl is still running, we just return nil,
+		// then if the ddl is failed, the downstream ddl is lost.
+		// because the checkpoint ts is forwarded.
+		log.Info("async add index ddl is still running",
+			zap.String("changefeedID", m.id.String()),
+			zap.Uint64("commitTs", ddl.CommitTs),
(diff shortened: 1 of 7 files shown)


3AceShowHand:
I am curious that, if the previous asynchronous DDL execution is not finished, wait for it, then the next DDL is blocked, this looks regressed back to synchronous DDL execution. Why not just remove asynchronous DDL execution?

cdc/sink/ddlsink/mysql/async_ddl.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/pingcap/tiflow/pkg/errors"
+	"go.uber.org/zap"
+)
+
+var checkRunningAddIndexSQL = `
+SELECT JOB_ID, JOB_TYPE, SCHEMA_STATE, SCHEMA_ID, TABLE_ID, STATE, QUERY
+FROM information_schema.ddl_jobs
+WHERE DB_NAME = "%s" 
+    AND TABLE_NAME = "%s"
+    AND JOB_TYPE LIKE "add index%%"
```
3AceShowHand, line 34: what if it's an `drop index`, will it also takes a long time ?

3AceShowHand (approved):

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -112,9 +111,10 @@ func NewDDLSink(
 
 // WriteDDLEvent writes a DDL event to the mysql database.
 func (m *DDLSink) WriteDDLEvent(ctx context.Context, ddl *model.DDLEvent) error {
-	if ddl.Type == timodel.ActionAddIndex && m.cfg.IsTiDB {
-		return m.asyncExecAddIndexDDLIfTimeout(ctx, ddl)
+	if m.shouldAsyncExecDDL(ddl) {
+		return m.asyncExecDDL(ctx, ddl)
 	}
+	m.waitAsynExecDone(ctx, ddl)
```
asddongmen, line 117: It seems all DDLs need to go through this logic before they can be executed. Would this increase the time needed for a DDL to be replicated?

cdc/sink/ddlsink/mysql/async_ddl.go
CharlesCheung96: No, drop index is not a reorg ddl, [ref the doc](https://docs.pingcap.com/tidb/dev/ddl-introduction#:~:text=Currently%2C%20physical%20DDL%20statements%20only%20include%20ADD%20INDEX%20and%20lossy%20column%20type%20changes%20(such%20as%20changing%20from%20an%20INT%20type%20to%20a%20CHAR%20type).)

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -78,6 +76,7 @@ func NewDDLSink(
 	}
 
 	dsnStr, err := pmysql.GenerateDSN(ctx, sinkURI, cfg, GetDBConnImpl)
+	log.Warn("dsn", zap.String("dsn", dsnStr))
```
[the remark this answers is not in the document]
CharlesCheung96: It's dummy log only used to debug. I wiil remove this later.

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
CharlesCheung96: Yes, it takes at least one additional request in TiDB scenario.

3AceShowHand: sounds like we can add a variable to indicate that there might be one asynchronous running ddl ?

asddongmen: It would be unacceptable if there is a huge amount of records in the downstream DDL table, let alone the time spent waiting here: https://github.com/pingcap/tiflow/pull/11476/files#diff-67001e77c3d88c6c47cfed312c5b1f5f36d2a404163aefb5da2939fce5051674R109

wk989898: Can multiple async DDLs be executed without waiting？

CharlesCheung96: We only [query one record](https://github.com/pingcap/tiflow/pull/11476/files#diff-67001e77c3d88c6c47cfed312c5b1f5f36d2a404163aefb5da2939fce5051674R36), so I understand that the overhead is fixed and does not depend on the size of the table.

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -62,6 +61,8 @@ type DDLSink struct {
 	// statistics is the statistics of this sink.
 	// We use it to record the DDL count.
 	statistics *metrics.Statistics
+	// map: model.TableName -> struct{}
+	runningAsyncDDLs sync.Map
```
CharlesCheung96, line 65: @3AceShowHand  @asddongmen @wk989898 PTAL, A cache is added here to quickly determine if a asyncCheck is needed.

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
CharlesCheung96: > It would be unacceptable if there is a huge amount of records in the downstream DDL table, let alone the time spent waiting here: https://github.com/pingcap/tiflow/pull/11476/files#diff-67001e77c3d88c6c47cfed312c5b1f5f36d2a404163aefb5da2939fce5051674R109

Got it, there is indeed a bug here that we should not be waited for 5s before the first check.

cdc/sink/ddlsink/mysql/async_ddl.go
```diff
@@ -0,0 +1,185 @@
...
+		done := m.doCheck(ctx, table)
+		if !done {
+			return false
+		}
+	}
+	return true
+}
+
+func (m *DDLSink) doCheck(ctx context.Context, table model.TableName) (done bool) {
+	if _, ok := m.runningAsyncDDLs.Load(table); !ok {
```
asddongmen, line 145: If there is a running async DDL after the CDC restarts, this check will fail to detect it here. 
A mark needs to be added to determine if the CDC has just restarted. The justRestarted state can transition to normal after one DDL is executed.

CharlesCheung96: we can not determine the state by single mark, suppose there are two tables t1 and t2, then:
1.  Before restart: t1->running ddl1
2.  After restart: if there is a ddl event of t2 executed before t1, then the justRestarted is useless.

CharlesCheung96: The purpose of adding caching here is to reduce the frequency of database reads. A fixed-size LRU seems like a reasonable choice.

asddongmen (approved):

asddongmen:
> I am curious that, if the previous asynchronous DDL execution is not finished, wait for it, then the next DDL is blocked, this looks regressed back to synchronous DDL execution. Why not just remove asynchronous DDL execution?

@3AceShowHand If the user executes an add index and no other DDL operations comes for the same table before the add index is finished, the benefit of async mode add index is significant.

PR #11491, opened 2024-08-16
mysql sink(ticdc): Refactor DB Connection Creation to Facilitate Unit Testing

### LINKED ISSUE #11490: ticdc: Difficulty in Mocking Database Connections in Unit Tests

- [X] Googled your question

### What is your question?

**Description:**

During the process of creating a connection to a downstream database using a Sink URI, two distinct connections are established:

1. **Temporary Connection:** Initially, a temporary connection is created to query important information from the downstream database, such as version, charset, and other details. Once this information is retrieved, the temporary connection is closed.
   
2. **Standard Connection:** The information gathered from the temporary connection is used to populate additional parameters into the Sink URI. This refined Sink URI is then used to create the standard connection that the Sink will use for subsequent operations.

While this process of creating two connections works perfectly fine in normal system operations, where the same method can be used to establish both connections, it presents a significant challenge in unit testing. In unit tests, we cannot start an actual downstream database; instead, we need to mock the database connections. Since both connections will execute SQL queries, the unit tests require mocking two different connections to handle each phase of the connection creation process.

The current approach, where both connections are created using the same method, falls short in unit testing scenarios. This limitation has led to the use of highly tricky and complex workarounds in unit tests, as shown in the example below:

```go
dbIndex := 0
mockGetDBConn := func(ctx context.Context, dsnStr string) (*sql.DB, error) {
	defer func() { dbIndex++ }()

	if dbIndex == 0 {
		// test db
		db, err := pmysql.MockTestDB()
		require.Nil(t, err)
		return db, nil
	}

	// normal db
	db, mock := newTestMockDB(t)
	mock.ExpectClose()
	return db, nil
}
```

**Proposed Solution:**

To address this issue, we can leverage an interface to manage the creation of these connections more effectively. By defining an interface with two methods—`CreateTemporaryConnection` and `CreateStandardConnection`—we can implement different strategies for creating the temporary and standard connections. This approach allows us to mock and test the connections separately, reducing the complexity and improving the maintainability of our unit tests.

This issue aims to implement such an interface to simplify unit testing by providing distinct methods for creating the necessary database connections.

wlwilliamx:

Issue Number: close #11490

**Overview:**

This PR addresses the challenges associated with creating multiple database connections during the establishment of downstream connections using a Sink URI, particularly in the context of unit testing. The changes introduce a flexible interface and supporting implementations that allow for distinct handling of temporary and standard connections, simplifying the testing process.

**Key Changes:**

1. **Introduction of `IDBConnectionFactory` Interface:**
   - Defined an interface `IDBConnectionFactory` with two methods:
     - `CreateTemporaryConnection`: For creating a temporary connection to the downstream database.
     - `CreateStandardConnection`: For creating the standard connection used in normal operations.
   - This separation of concerns allows different connection creation strategies to be used during testing and in production.

2. **Implementation of `DBConnectionFactory`:**
   - Created a `DBConnectionFactory` struct that implements `IDBConnectionFactory`.
   - Both `CreateTemporaryConnection` and `CreateStandardConnection` methods use the same connection creation logic (`CreateMySQLDBConn`), making this implementation suitable for normal system operations where there’s no need to distinguish between connection types.

3. **Creation of `DBConnectionFactoryForTest`:**
   - Added a `DBConnectionFactoryForTest` struct to facilitate unit testing.
   - This struct allows separate connection factories to be set for temporary and standard connections via `SetTemporaryConnectionFactory` and `SetStandardConnectionFactory`.
   - During testing, `CreateTemporaryConnection` and `CreateStandardConnection` will use the specified mock connection factories, allowing distinct behaviors for each connection.

**Problem Solved:**

Prior to this refactor, unit tests were forced to use complicated and error-prone methods to mock the behavior of two distinct database connections. The introduction of the `IDBConnectionFactory` interface and its implementations significantly simplifies the mocking process, providing a clean and maintainable way to test different connection behaviors without resorting to complex or tricky code.

**Conclusion:**

This PR provides a robust solution to the issues faced during unit testing of connection creation logic. By introducing a clear interface and appropriate implementations, it allows for both production and testing environments to handle connection creation in a straightforward and maintainable manner.

#### Tests 

 - Unit test

No

No

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/sink/ddlsink/mysql/async_ddl_test.go
+++ b/cdc/sink/ddlsink/mysql/async_ddl_test.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
-	"sync/atomic"
 	"testing"
 	"time"
 
@@ -33,18 +32,13 @@ import (
 )
 
 func TestWaitAsynExecDone(t *testing.T) {
-	var dbIndex int32 = 0
-	GetDBConnImpl = func(ctx context.Context, dsnStr string) (*sql.DB, error) {
-		defer func() {
-			atomic.AddInt32(&dbIndex, 1)
-		}()
-		if atomic.LoadInt32(&dbIndex) == 0 {
-			// test db
-			db, err := pmysql.MockTestDB()
-			require.Nil(t, err)
-			return db, nil
-		}
-		// normal db
+	dbConnFactory := pmysql.DBConnectionFactoryForTest{}
+	dbConnFactory.SetTemporaryConnectionFactory(func(ctx context.Context, dsnStr string) (*sql.DB, error) {
+		db, err := pmysql.MockTestDB()
+		require.Nil(t, err)
+		return db, nil
+	})
+	dbConnFactory.SetStandardConnectionFactory(func(ctx context.Context, dsnStr string) (*sql.DB, error) {
 		db, mock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
 		require.Nil(t, err)
 		mock.ExpectQuery("select tidb_version()").
@@ -71,7 +65,8 @@ func TestWaitAsynExecDone(t *testing.T) {
 
 		mock.ExpectClose()
 		return db, nil
-	}
+	})
+	GetDBConnImpl = &dbConnFactory
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
@@ -116,18 +111,13 @@ func TestWaitAsynExecDone(t *testing.T) {
 
 func TestAsyncExecAddIndex(t *testing.T) {
 	ddlExecutionTime := time.Second * 15
-	var dbIndex int32 = 0
-	GetDBConnImpl = func(ctx context.Context, dsnStr string) (*sql.DB, error) {
-		defer func() {
-			atomic.AddInt32(&dbIndex, 1)
-		}()
-		if atomic.LoadInt32(&dbIndex) == 0 {
-			// test db
-			db, err := pmysql.MockTestDB()
-			require.Nil(t, err)
-			return db, nil
-		}
-		// normal db
+	dbConnFactory := pmysql.DBConnectionFactoryForTest{}
+	dbConnFactory.SetTemporaryConnectionFactory(func(ctx context.Context, dsnStr string) (*sql.DB, error) {
+		db, err := pmysql.MockTestDB()
+		require.Nil(t, err)
+		return db, nil
+	})
+	dbConnFactory.SetStandardConnectionFactory(func(ctx context.Context, dsnStr string) (*sql.DB, error) {
 		db, mock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
 		require.Nil(t, err)
 		mock.ExpectQuery("select tidb_version()").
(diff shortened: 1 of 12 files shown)


cdc/sink/dmlsink/txn/txn_dml_sink.go
```diff
@@ -58,11 +58,9 @@ type dmlSink struct {
 	scheme string
 }
 
-// GetDBConnImpl is the implementation of pmysql.Factory.
+// GetDBConnImpl is the implementation of pmysql.IDBConnectionFactory.
```
asddongmen, line 61:
```suggestion
// GetDBConnImpl is the implementation of pmysql.DBConnectionFactory.
```

cdc/sink/dmlsink/txn/txn_dml_sink.go
```diff
@@ -58,11 +58,9 @@ type dmlSink struct {
...
 // Exported for testing.
-// Maybe we can use a better way to do this. Because this is not thread-safe.
-// You can use `SetupSuite` and `TearDownSuite` to do this to get a better way.
-var GetDBConnImpl pmysql.Factory = pmysql.CreateMySQLDBConn
+var GetDBConnImpl pmysql.IDBConnectionFactory = &pmysql.DBConnectionFactory{}
```
asddongmen, line 63: Why call it `IDBConnectionFactory` ?

wlwilliamx: Just code naming style. Adding an "I" in front of the interface name makes it clear that it's an interface.

asddongmen (approved):

pkg/sink/mysql/factory.go
```diff
@@ -16,7 +16,126 @@ package mysql
...
+	if d.standard == nil {
+		return nil, cerror.ErrCodeNilFunction.GenWithStackByArgs()
+	}
+	return d.standard(ctx, dsnStr)
+}
+
+// ConnectionFactory is a function type that takes a context and a DSN (Data Source Name) string
+// as input parameters, and returns a pointer to an sql.DB object and an error. This function type
+// is typically used to create and configure database connections.
+type ConnectionFactory func(ctx context.Context, dsnStr string) (*sql.DB, error)
```
CharlesCheung96, line 141: Please move the definition of CreateMySQLDBConn here.

cdc/sink/ddlsink/mysql/async_ddl_test.go
```diff
@@ -33,18 +32,13 @@ import (
...
-		}()
-		if atomic.LoadInt32(&dbIndex) == 0 {
-			// test db
-			db, err := pmysql.MockTestDB()
-			require.Nil(t, err)
-			return db, nil
-		}
-		// normal db
+	dbConnFactory := pmysql.DBConnectionFactoryForTest{}
+	dbConnFactory.SetTemporaryConnectionFactory(func(ctx context.Context, dsnStr string) (*sql.DB, error) {
```
CharlesCheung96, line 36: The implementation of SetTemporaryConnectionFactory is the same at all calls, maybe it should be handled during construction to reduce redundant code.

wlwilliamx: Thank you for your suggestion! That's a good idea; moving the implementation of SetTemporaryConnectionFactory to a new constructor is a good way to reduce this redundant code. I’ll make the necessary changes and update the PR shortly.

pkg/sink/mysql/factory.go
wlwilliamx: Good point! I’ll move the definition of CreateMySQLDBConn as suggested. Thanks for the feedback!

CharlesCheung96 (approved):

PR #11506, opened 2024-08-21
mysql(ticdc): remove returned error and modify the log in `CheckIsTiDB `

### LINKED ISSUE #11505: ticdc: Refactor `CheckIsTiDB` function to remove the returned error

- [X] Googled your question

### What is your question?

**Description**:
In the `CheckIsTiDB` function, the current implementation returns an `error` which is no need for the external callers. In earlier versions, this function returned an `error` along with a boolean value, allowing callers to differentiate between network-related issues and the absence of TiDB. However, since the specific error content wasn't critical to the logic—external callers only needed to know whether the downstream was TiDB—the decision was made to simplify the function. External callers should not handle the error returned here because even if the downstream is not TiDB, TiCDC should still function properly. For more details, refer to https://github.com/pingcap/tiflow/pull/11214.

The primary goal of this issue is to refactor the function to remove the error return type. Instead of returning an error, we will log a warning if an issue occurs while querying the TiDB version, as the error content itself is not critical to the logic. This change will simplify the function while maintaining awareness of potential network issues through logging.

**Action Items**:
1. Remove the error return value from `CheckIsTiDB`.
2. Log a warning instead of an error when `row.Scan()` fails.
3. Add comments to clarify the logic and reason for not returning errors.

wlwilliamx:

Issue Number: close #11505

**Description**:
This PR refactors the `CheckIsTiDB` function in response to internal discussions. Previously, the function returned an `error` which was not necessary, as the external callers should not care about the error of the `CheckIsTiDB` function due to https://github.com/pingcap/tiflow/issues/11213.

**Changes**:
1. Removed the `error` return type from `CheckIsTiDB`.
2. Replaced the error log with a warning to handle cases where querying the TiDB version fails.
3. Updated comments to explain the background and rationale for the change.

#### Tests 

 - Unit test

No

No

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
+++ b/cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
@@ -91,10 +91,7 @@ func NewDDLSink(
 		return nil, err
 	}
 
-	cfg.IsTiDB, err = pmysql.CheckIsTiDB(ctx, db)
-	if err != nil {
-		return nil, err
-	}
+	cfg.IsTiDB = pmysql.CheckIsTiDB(ctx, db)
 
 	cfg.IsWriteSourceExisted, err = pmysql.CheckIfBDRModeIsSupported(ctx, db)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


hongyunyan (approved):
LGTM

asddongmen (approved):

PR #11520, opened 2024-08-23
mysql(ticdc): do not check downstream ddl job status when create table

### LINKED ISSUE #11521: create table ddl executed too slow when using tidb as the downstream sink
### What did you do?

run `create table` ddl on the upstream TiDB.

TiCDC receive and execute ddl should be very quick

### What did you see instead?

img_v3_02e1_71b504c8-8e37-4040-a340-2ac72edaa77g

The second log is printed around 23 seconds later after the first one printed.

The root cause is that following query too slow, it may takes more than 20 seconds.

```
SELECT JOB_ID, JOB_TYPE, SCHEMA_STATE, SCHEMA_ID, TABLE_ID, STATE, QUERY
FROM information_schema.ddl_jobs
WHERE DB_NAME = "%s" 
    AND TABLE_NAME = "%s"
    AND JOB_TYPE LIKE "add index%%"
    AND (STATE = "running" OR STATE = "queueing")
LIMIT 1;
```

```
[2024/08/23 18:19:31.924 +08:00] [INFO] [async_ddl.go:165] ["async ddl query success"] [table=test.spmi_transfer_bill206] [duration=23.31010039s]
```

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

3AceShowHand:

Issue Number: close #11521 

* do not check async ddl status for the create table ddl

image

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None`
```

### DIFF
--- a/cdc/sink/ddlsink/mysql/async_ddl.go
+++ b/cdc/sink/ddlsink/mysql/async_ddl.go
@@ -98,6 +98,12 @@ func (m *DDLSink) waitAsynExecDone(ctx context.Context, ddl *model.DDLEvent) {
 		return
 	}
 
+	switch ddl.Type {
+	case timodel.ActionCreateTable, timodel.ActionCreateTables:
+		return
+	default:
+	}
+
 	tables := make(map[model.TableName]struct{})
 	if ddl.TableInfo != nil {
 		tables[ddl.TableInfo.TableName] = struct{}{}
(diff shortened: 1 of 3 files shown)


cdc/sink/ddlsink/mysql/async_ddl.go
```diff
@@ -98,6 +98,12 @@ func (m *DDLSink) waitAsynExecDone(ctx context.Context, ddl *model.DDLEvent) {
 		return
 	}
 
+	switch ddl.Type {
+	case timodel.ActionCreateTable, timodel.ActionCreateTables:
```
sdojjy, line 102: how about other DDLs?

cdc/sink/ddlsink/mysql/async_ddl.go
```diff
@@ -29,7 +29,7 @@ import (
 var checkRunningAddIndexSQL = `
 SELECT JOB_ID, JOB_TYPE, SCHEMA_STATE, SCHEMA_ID, TABLE_ID, STATE, QUERY
 FROM information_schema.ddl_jobs
```
CharlesCheung96, line 31: Maybe use `ADMIN SHOW DDL JOBS` here is a better choice?

PR #11523, opened 2024-08-27
mounter(ticdc): ignore checksum v2

### LINKED ISSUE #11524: ignore checksum if the version is not v1
For release-8.1 and release-7.5, only verify checksum v1.

In the upgrade scenario, the upstream TiDB is upgraded first, then may output checksum v2, in the such case, downstream TiCDC cannot verify it, and makes the changefeed error, so just ignore the error

3AceShowHand:

Issue Number: close #11524

* ignore the checksum if not v1 (version != 0)

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/entry/mounter.go
+++ b/cdc/entry/mounter.go
@@ -435,8 +435,9 @@ func (m *mounter) verifyChecksum(
 	}
 
 	version := decoder.ChecksumVersion()
-	if version == 1 {
-		log.Info("checksum v2 found, this happens since upstream TiDB is in a higher version, ignore it")
+	if version != 0 {
+		log.Info("checksum is not v1, this happens since upstream TiDB is in a higher version, ignore it",
+			zap.Int("version", version))
 		return 0, true, nil
 	}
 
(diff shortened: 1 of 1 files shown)


cdc/entry/mounter.go
```diff
@@ -434,6 +434,13 @@ func (m *mounter) verifyChecksum(
 		decoder = m.decoder
 	}
 
+	version := decoder.ChecksumVersion()
+	if version != 0 {
+		log.Info("checksum is not v1, this happens since upstream TiDB is in a higher version, ignore it",
```
asddongmen, line 439: Would this log flooding?

3AceShowHand: this only happens in the upgrade scenario, it should not take too long, so not too match logs.

sdojjy (approved):

cdc/entry/mounter.go
asddongmen: Rolling upgrade would take more than 10 minutes if there are a lot of tables.

asddongmen (approved):

PR #11535, opened 2024-08-28
sink(ticdc): use admin statement to query async ddl status

### LINKED ISSUE #11521: create table ddl executed too slow when using tidb as the downstream sink

CharlesCheung96:

Issue Number: close #11521

- Use admin statement to query async ddl status since the original querying operation has performance issue.

#### Tests 

 - Unit test
 - Integration test
 - No code

- After this PR, the user need grant admin permission to TiCDC in downstream database, otherwise TiCDC will disable the optimization of #11476.

### Release note 

```release-note

```


cdc/sink/ddlsink/mysql/async_ddl.go
```diff
@@ -16,24 +16,27 @@ package mysql
...
 
+// TODO: Use the flollowing SQL to check the ddl job status after tidb optimize
+// the information_schema.ddl_jobs table. Ref: https://github.com/pingcap/tidb/issues/55725
+//
+// SELECT JOB_ID, JOB_TYPE, SCHEMA_STATE, SCHEMA_ID, TABLE_ID, STATE, QUERY
+// FROM information_schema.ddl_jobs
 var checkRunningAddIndexSQL = `
-SELECT JOB_ID, JOB_TYPE, SCHEMA_STATE, SCHEMA_ID, TABLE_ID, STATE, QUERY
-FROM information_schema.ddl_jobs
+ADMIN SHOW DDL JOBS 1
```
lance6716, line 35: I'm not sure if this query will see an already finished DDL on the same table.

BTW DM uses `getDDLStatusFromTiDB` please see if we can unify it for our two projects.

CharlesCheung96: Here, TiCDC does not care about the finished ddl event, in fact, it only needs to ensure that there are no running add index ddl downstream of the same table before executing the next ddl event. In addition, the `getDDLStatusFromTiDB` function iterates through the job history to query for a particular ddl event, it may take more time than here.

cdc/sink/ddlsink/mysql/async_ddl.go
```diff
@@ -16,24 +16,27 @@ package mysql
...
 	"time"
 
 	"github.com/pingcap/log"
 	timodel "github.com/pingcap/tidb/pkg/parser/model"
 	"github.com/pingcap/tiflow/cdc/model"
-	"github.com/pingcap/tiflow/pkg/errors"
 	"go.uber.org/zap"
 )
 
+// TODO: Use the flollowing SQL to check the ddl job status after tidb optimize
```
lance6716, line 29: Did you check the performance of this new query? I take a glimpse into ADMIN SHOW DDL JOBS and it still loads many ddl job before filtering

CharlesCheung96: We have tested scenarios with a large number of historical DDLs:
image

Also, note that the tidb implementation here [queries all records](https://github.com/pingcap/tidb/blob/master/pkg/executor/executor.go#L417) in the `tidb_ddl_job` table, and query performance depends on the size of the `tidb_ddl_job`. Seems the results of this query have to be stored in memory. Maybe it is more appropriate to push down the filter criteria in tidb?

In this PR, we add a timeout to control the max query cost.

cdc/sink/ddlsink/mysql/async_ddl.go
```diff
@@ -92,9 +96,23 @@ func (m *DDLSink) asyncExecDDL(ctx context.Context, ddl *model.DDLEvent) error {
...
 
+func (m *DDLSink) needWaitAsyncExecDone(ddl *model.DDLEvent) bool {
+	if !m.cfg.IsTiDB {
+		return false
+	}
+	switch ddl.Type {
+	case timodel.ActionCreateTable, timodel.ActionCreateTables:
+		return false
+	case timodel.ActionCreateSchema:
+		return false
```
3AceShowHand, line 107: how about add a `default` branch here ?

cdc/sink/ddlsink/mysql/async_ddl.go
```diff
@@ -92,9 +96,23 @@ func (m *DDLSink) asyncExecDDL(ctx context.Context, ddl *model.DDLEvent) error {
 	}
 }
 
+func (m *DDLSink) needWaitAsyncExecDone(ddl *model.DDLEvent) bool {
```
3AceShowHand, line 99: It looks this method only use `ddl.Type`, so just pass the ddl type as the parameter, no need to pass the whole DDL, this makes write unit test easier.

3AceShowHand: can you add a unit test to cover this method ?

3AceShowHand (approved):

hongyunyan (approved):

asddongmen (approved):

PR #11579, opened 2024-09-09
kv(ticdc): kv client should handle region congested error explicitly

### LINKED ISSUE #11578: make TiCDC knows region fuse happend on the TiKV-CDC side
TiKV-CDC may fuse the region if tries to send event but meet TiKV-CDC sink memory quota full. At the moment, `region_not_found` returned to the ticdc.

* TiKV-CDC return `Congested` error to the TiCDC, so that TiCDC knows the real problem happend.

3AceShowHand:

Issue Number: close #11578 

TiKV-CDC return `Congested` error if region fuse happens, after the PR https://github.com/tikv/tikv/pull/17507/files gets merged.

* kv client handle `Congested` error explicitly, record it to the metrics, handle it like `region_not_found`

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```


asddongmen (approved):

cdc/kv/shared_client.go
```diff
@@ -693,6 +694,11 @@ func (s *SharedClient) doHandleError(ctx context.Context, errInfo regionErrorInf
 			s.scheduleRangeRequest(ctx, errInfo.span, errInfo.subscribedTable)
 			return nil
 		}
+		if innerErr.GetCongested() != nil {
+			metricKvCongestedCounter.Inc()
+			s.scheduleRangeRequest(ctx, errInfo.span, errInfo.subscribedTable)
```
hicqu, line 699: It should be `scheduleRegionRequest` I guess.

3AceShowHand: previous, `region_not_found` is handled by `scheduleRangeRequest`

hicqu: They are different. If the region is not found, we need to ask PD whether the region is changed or not. But for Congested, it's not necessary.

hicqu (approved):

PR #11595, opened 2024-09-13
sink: avoid memory allocations for transforming ColumnData to Column

### LINKED ISSUE #11590: columnData2Column utilizes too much CPU in a single point component
### What did you do?

Replicate a table with 2 unique keys, which are both union index.

_No response_

### What did you see instead?

Here is a result of CPU profile in `30s`.

[profiling_2024-09-11_20-43-08.zip](https://github.com/user-attachments/files/16964110/profiling_2024-09-11_20-43-08.zip)

图片

The stack is very clear. `model.columnData2Column` utilizes almost one full core. And, GC also works in a heavy load.

The function `model.columnData2Column` is introduced in `v8.1.0`, but we don't know whether it exists in earlier versions or not. 

### Versions of the cluster

TiCDCs after `v8.1.0`.

hicqu:

Issue Number: close #11590 

Avoid to alloc temp objects for ColumnData.

#### Tests 

 - Unit test

### Release note 

```release-note
None
```


asddongmen (approved):

lidezhu (approved):
LGTM

PR #11624, opened 2024-09-24
cordinator(ticdc): Fix Puller Resolved TS Lag Calculation and Deprecate current_ts Field in Stats

### LINKED ISSUE #11561: resolved ts slowest table puller lag become negative year when executing DDL
image

wlwilliamx:

Issue Number: close #11561

#### Summary:
This PR addresses a critical issue in the Puller Resolved TS Lag metric, where the lag was incorrectly displayed as -55 years. The root cause of the problem was the improper initialization of the `CurrentTs` field in the `Stats` structure within `table.pb.go`. This led to a situation where `CurrentTs` remained zero while `ResolvedTs` was correctly populated, resulting in a negative lag calculation that erroneously reflected a time difference of 55 years (from 1970 to the current year).

#### Detailed Explanation:
- **Issue Background:**
  - The Puller Resolved TS Lag was calculated as the difference between `CurrentTs` and `ResolvedTs` from the `Stats` structure.
  - However, the `NewReplicationSet` function did not correctly initialize `CurrentTs`, leading to incorrect lag values when metrics were collected.

- **Design Intent of `current_ts`:**
  - The `current_ts` field was initially designed to record the time at the processor side, which would then be used by the coordinator for lag calculation. This design was meant to avoid including the transmission delay and processing time on the coordinator side, as these could introduce inaccuracies in the lag calculation.
  - However, this potential inaccuracy is minimal since the communication between the processor and coordinator happens within the TiCDC cluster, where network delays are typically in the millisecond range. Given that the lag metric is reported in seconds, any millisecond-level discrepancy is considered acceptable.
  - Deprecating the `current_ts` field may introduce a slight inaccuracy in the lag calculation, as the "current time" will now be fetched directly by the coordinator rather than from the processor. This could result in a minor error due to transmission delays and processing time. However, this inaccuracy is negligible for practical purposes, as the delay is typically within milliseconds and does not significantly impact the metric, which is measured in seconds.
  - Additionally, since this field was part of the protobuf (`pb`) structure, it added extra overhead to every network transmission by carrying an additional value. It also introduced more complexity into the code, requiring maintenance of this field, which was part of the issue that led to the incorrect metric calculations in this case.

- **Root Cause:**
  - The use of `CurrentTs` from the `Stats` structure was flawed because this timestamp was not reliably up-to-date.
  - The metric calculation requires the "current time," which should always be fetched directly from PD to ensure accuracy, rather than relying on a potentially outdated `CurrentTs` stored in the `Stats` structure.

- **Solution Implemented:**
  - This PR updates the metric calculation to fetch the latest "current time" directly from PD whenever calculating the lag.
  - All references to the `CurrentTs` field within the codebase, particularly in metric calculations, have been updated to reflect this change.
  
- **Deprecation of `current_ts` Field:**
  - The `current_ts` field in the `Stats` structure, defined in `cdc/processor/tablepb/table.proto`, has been marked as deprecated.
  - The `Stats` structure is used for data transmission between the Coordinator and Processor, where `current_ts` was unnecessarily included in the state for each Table Span.
  - This PR removes all dependencies on `current_ts` within the project, ensuring that the Coordinator and Processor no longer rely on this outdated field.

#### Impact:
This PR ensures accurate lag calculation in the Puller Resolved TS Lag metric and removes the outdated `current_ts` field, improving both performance and maintainability by reducing unnecessary transmission overhead and field maintenance. While this change may introduce a slight inaccuracy in the lag calculation due to the use of coordinator-side timestamps, this impact is minimal and well within acceptable thresholds.

#### Tests 

 - Unit test
 - Integration test

No. The deprecation of the current_ts field in the Stats structure is backward-compatible because the field is not actively used elsewhere in the project beyond the affected metrics. The existing field remains in the protocol buffer definition but is marked as deprecated, so it won’t break compatibility with older versions that still use the field.

No.

### Release note 

```release-note
Fix the issue that the puller resolved ts lag metric shows incorrect values. #11561
```


cdc/processor/tablepb/table.proto
```diff
@@ -65,7 +65,7 @@ message Stats {
     // Number of captured regions.
     uint64 region_count = 1;
     // The current timestamp from the table's point of view.
-    uint64 current_ts = 2 [(gogoproto.casttype) = "Ts"];
```
3AceShowHand, line 68: can we remove this field?

3AceShowHand (approved):

cdc/scheduler/internal/v3/coordinator.go
```diff
@@ -449,7 +449,13 @@ func (c *coordinator) maybeCollectMetrics() {
 	}
 	c.lastCollectTime = now
 
+	pdTime := time.Now()
```
wk989898, line 452:
```suggestion
	pdTime := now
```

wlwilliamx: Good suggestion!

cdc/processor/tablepb/table.proto
wlwilliamx: Currently, we cannot directly remove the current_ts field from the protobuf (pb) definition. Although this message is used solely for communication within the TiCDC cluster, there is a risk of incompatibility during rolling upgrades. In such scenarios, different versions of TiCDC (coordinator and processor) may coexist, and removing the current_ts field could lead to communication errors between nodes running different versions.

However, the field is being deprecated and is no longer used for any critical functionality. Once we are confident that it is not in use across any versions, we can plan to safely remove it in a future release without breaking compatibility.

asddongmen (approved):

Issue #11512, opened 2024-08-23
ticdc: tar Extraction Failure in scripts/download-integration-test-binaries.sh Due to Wildcard Handling


wlwilliamx:
**Description:**

I encountered an issue with the `scripts/download-integration-test-binaries.sh` script while running integration tests for TiCDC. This script is responsible for downloading the necessary binaries for the integration tests. In particular, it attempts to download and extract the PD binaries using the following commands:

```bash
download "$pd_download_url" "pd-server.tar.gz" "tmp/pd-server.tar.gz"
tar -xz -C third_bin 'bin/*' -f tmp/pd-server.tar.gz && mv third_bin/bin/* third_bin/
```

The intent here is to extract all the executable files located under the `bin` directory in the `pd-server.tar.gz` package. However, the script may fail when running on systems with certain versions of `tar`. The executable files located under the `bin` directory in the `pd-server.tar.gz` package:

![image](https://github.com/user-attachments/assets/ef4dd1ad-cd78-4576-b5e5-c49542a2936e)

### Problem:

![img_v3_02e0_8f2d7564-014a-4ff5-84c9-540fe16b91bg](https://github.com/user-attachments/assets/c030bb85-16dc-4ce5-81fd-f986b3d5720d)

The issue arises because not all versions of `tar` support wildcard extraction by default. Specifically, the command:

```bash
tar -xz -C third_bin 'bin/*' -f tmp/pd-server.tar.gz
```

uses the wildcard `'bin/*'` to match all files under the `bin` directory in the package. While this works on some systems, other versions of `tar` (typically older or non-GNU implementations) do not enable wildcard matching by default. To ensure that `tar` interprets the wildcard patterns correctly, the `--wildcards` option needs to be explicitly included. Without this option, the extraction fails, leaving no PD executables in the `third_bin` directory.

### Impact:

As a result, when the script fails to extract the PD binaries correctly, it leaves the `third_bin` directory empty of PD-related executables, causing the integration tests to fail since the required binaries are missing.

### Suggested Fix:

To address this issue, we should modify the `tar` extraction command to include the `--wildcards` option, ensuring that the wildcard pattern is correctly interpreted across all versions of `tar`. The updated command would look like this:

```bash
tar -xz --wildcards -C third_bin 'bin/*' -f tmp/pd-server.tar.gz
```

This ensures that the script works consistently across different environments and `tar` implementations.

### Versions of the cluster

TiCDC version (execute `cdc version`):

```console
8.3.0
```

PR #11632, opened 2024-09-26
integration test(ticdc): Reintroduce `--wildcards` Option in tar Command for download-integration-test-binaries.sh

### LINKED ISSUE #11631: ticdc: Recurrence of tar Wildcard Handling Issue in download-integration-test-binaries.sh After Refactor
### What did you do?

The issue previously described in #11512 (where tar wildcard handling caused extraction failures in the scripts/download-integration-test-binaries.sh script) has resurfaced after the script was refactored.

During the refactoring, the --wildcards option was not reintroduced, which is required for tar to correctly handle wildcard patterns in some versions. As a result, the script is once again failing to extract the necessary binaries for PD, particularly on systems where tar does not support wildcard patterns by default.

I expected the scripts/download-integration-test-binaries.sh script to successfully extract the PD binaries from the pd-server.tar.gz archive. This would allow the integration tests to run without any errors or missing binaries.

The solution is to reintroduce the --wildcards option in the tar command, as was done in #11513. This ensures compatibility across different environments and versions of tar.

### What did you see instead?

The script failed to extract the PD binaries from the pd-server.tar.gz archive due to the missing --wildcards option in the tar command. As a result, the third_bin directory was empty, and the integration tests could not run because the necessary PD executables were missing.
CleanShot 2024-09-26 at 16 13 53@2x

### Versions of the cluster

```console
No TiDB.
```

```console
No TiKV.
```

```console
v8.3.0
```

wlwilliamx:

Issue Number: close #11631

**Description:**

This PR addresses the reoccurrence of the `tar` extraction issue in `scripts/download-integration-test-binaries.sh` following a refactor, where the `--wildcards` option was omitted. This option is necessary for wildcard pattern matching in certain versions of `tar` to correctly extract the required binaries for PD.

#### **Context:**

In a previous issue:#11512 and pr:#11513, a similar problem was identified and resolved by adding the `--wildcards` option to the `tar` command. However, during the refactoring of the script, this option was unintentionally removed, causing the issue to resurface.

The current command:

https://github.com/pingcap/tiflow/blob/4e177d94ca41d5f49a44c5545adfca1c72820f87/scripts/download-integration-test-binaries.sh#L156

https://github.com/pingcap/tiflow/blob/4e177d94ca41d5f49a44c5545adfca1c72820f87/scripts/download-integration-test-binaries.sh#L176-L180

fails on some systems due to lack of support for wildcard matching by default. Without the `--wildcards` option, the necessary binaries are not extracted, leading to test failures.

#### **Solution:**

To fix this issue, I have reintroduced the `--wildcards` option in the `tar` command, ensuring compatibility across different systems. 

This guarantees that `tar` will correctly interpret wildcard patterns, even in environments where it isn’t enabled by default.

#### **Impact:**

- Ensures consistent extraction of binaries across different environments and `tar` implementations.
- Prevents failures in the integration tests by ensuring the correct binaries are extracted and available.

This PR is related to the previous #11513 that addressed the same issue before the script refactor.

#### Tests 

```bash
make prepare_test_binaries
```

No.

No.

### Release note 

```release-note
None
```

### DIFF
--- a/scripts/download-integration-test-binaries.sh
+++ b/scripts/download-integration-test-binaries.sh
@@ -174,9 +174,9 @@ download_and_extract() {
 
 	download_file "$url" "$file_name" "${TMP_DIR}/$file_name"
 	if [ -n "$extract_path" ]; then
-		tar -xz -C ${THIRD_BIN_DIR} $extract_path -f ${TMP_DIR}/$file_name
+		tar -xz --wildcards -C ${THIRD_BIN_DIR} $extract_path -f ${TMP_DIR}/$file_name
 	else
-		tar -xz -C ${THIRD_BIN_DIR} -f ${TMP_DIR}/$file_name
+		tar -xz --wildcards -C ${THIRD_BIN_DIR} -f ${TMP_DIR}/$file_name
 	fi
 
 	# Move extracted files if necessary
@@ -206,6 +206,7 @@ cleanup() {
 
 setup() {
 	cleanup
+	rm -rf ${BIN_DIR}
 	mkdir -p ${THIRD_BIN_DIR} ${TMP_DIR} ${BIN_DIR}
 }
 


scripts/download-integration-test-binaries.sh
```diff
@@ -206,6 +206,7 @@ cleanup() {
 
 setup() {
 	cleanup
+	rm -rf ${BIN_DIR}
```
lance6716, line 209: What will it delete?

wlwilliamx: The `BIN_DIR` is the final location where the binaries are stored. The reason I delete this folder before creating it is that if the folder is not empty, running make prepare_test_binaries will result in some strange tar command errors, as shown in the image below. This causes make prepare_test_binaries to fail unless BIN_DIR is manually deleted beforehand.

CleanShot 2024-09-26 at 17 08 55@2x

lance6716: I think to avoid repeatly downloading and reduce waiting time, we should not delete the whole folder. It's better to ignore existing files one-by-one, and use `mv` to atomically move the files.

wlwilliamx: This problem is not caused by the `mv` command but by the `tar` command. The error occurs when extracting `pd-server.tar.gz`. The reason is that the script needs to extract all the binary files under the `bin` directory from `pd-server.tar.gz`, so a wildcard `"bin/*"` is used to extract the contents of the `bin/*` directory. However, if there is already a `bin` folder in the current working directory and it contains files that do not exist in the `bin` directory of `pd-server.tar.gz`, the error will occur. This happens because before `tar` is executed, the wildcard is expanded, so it reads all the contents of the current working directory's `bin` folder and passes them as arguments to `tar`, which then tries to extract these files that are not present in the archive, leading to the error.

Additionally, I have also tried not using the wildcard and directly extracting the `bin` directory from `pd-server.tar.gz`. However, if the `BIN_DIR` already contains a folder with the same name, the `mv` command will fail:
CleanShot 2024-09-29 at 11 57 07@2x

Therefore, the simplest and most effective solution is to directly clean the `BIN_DIR` before proceeding. This way, the script can be executed repeatedly without causing any issues. It will not have any negative impact since `BIN_DIR` does not exist by default and is `gitignored` in the project. It is created specifically by the script for this purpose.

lance6716: Oh, I see that `prepare_test_binaries` is manually executed. I wrongly think it will be executed automatically in integration test so I don't want the cached binaries to be deleted.

BTW, to achieve "atomic" extracting, we can extract them to a temporary folder in the same file system mounting entry and `mv` the folder.

lance6716 (approved):

lidezhu (approved):

PR #11641, opened 2024-10-10
api(ticdc): Not show the error info during query the changefeed when state is stopped, finished and removed

hongyunyan:

Issue Number: close https://github.com/pingcap/tiflow/issues/11642

Make the previous error message not return when changefeed state is stopped / finished / removed, to avoid confused.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/api/v2/changefeed.go
+++ b/cdc/api/v2/changefeed.go
@@ -284,7 +284,7 @@ func (h *OpenAPIV2) listChangeFeeds(c *gin.Context) {
 
 		// if the state is normal, we shall not return the error info
 		// because changefeed will is retrying. errors will confuse the users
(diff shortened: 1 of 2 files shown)


cdc/api/v2/changefeed.go
```diff
@@ -284,7 +284,7 @@ func (h *OpenAPIV2) listChangeFeeds(c *gin.Context) {
 
 		// if the state is normal, we shall not return the error info
 		// because changefeed will is retrying. errors will confuse the users
-		if commonInfo.FeedState == model.StateNormal {
+		if commonInfo.FeedState == model.StateNormal || commonInfo.FeedState == model.StateStopped || commonInfo.FeedState == model.StateFinished || commonInfo.FeedState == model.StateRemoved {
```
sdojjy, line 287: too complex here, exact it to a function, then use it in line 1039

asddongmen (approved):

sdojjy (approved):

benmeadowcroft (approved):

PR #11649, opened 2024-10-11
sink(ticdc): add event key output for debezium

### LINKED ISSUE #11652: The Key field is not filled in the Kafka event of Debezium protocol
### What did you do?

According to the definition of Debezium, the Key should be the primary key or unique key of an event. https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-events

image

_No response_

### What did you see instead?

Currently, TiCDC always fill it with nill
https://github.com/pingcap/tiflow/blob/master/pkg/sink/codec/debezium/encoder.go#L64
image

### Versions of the cluster

master

wk989898:

Issue Number: close #11652

Debezium send event key and event value to kafka. I added event key output to this PR.
Also modify function `SanitizeName` for keeping same with official code. 

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
Fix the issue that the Key of Kafka messages has't been filled in Debezium protocol.
```


tests/integration_tests/debezium/sql/debezium/real_test.sql
```diff
@@ -1,3 +1,5 @@
+-- set sql_mode REAL_AS_FLOAT is necessary
```
hongyunyan, line 1: Why set REAL_AS_FLOAT is neccesary?

pkg/sink/codec/debezium/codec.go
```diff
@@ -529,7 +714,45 @@ func (c *dbzCodec) writeBinaryField(writer *util.JSONWriter, fieldName string, v
...
+	// see https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-events
+	cols, colInfos := e.HandleKeyColInfos()
+	jWriter := util.BorrowJSONWriter(dest)
+	defer util.ReturnJSONWriter(jWriter)
+
+	var err error
+	jWriter.WriteObject(func() {
+		jWriter.WriteObjectField("payload", func() {
+			for i, col := range cols {
+				colx := model.Column2ColumnDataX(col)
```
hongyunyan, line 732: Could we use GetColumnDataX instead of  Column2ColumnDataX as writeDebeziumFieldValues.

3AceShowHand (approved):

tests/integration_tests/debezium/sql/debezium/real_test.sql
wk989898: Because Debezium treats REAL as a synonym for FLOAT. https://github.com/pingcap/tiflow/blob/45b439b210688a28afc46bcb47fb24c81b02d6c7/pkg/sink/codec/debezium/codec.go#L385-L386

pkg/sink/codec/debezium/codec.go
wk989898: GetColumnDataX has existed. https://github.com/pingcap/tiflow/blob/45b439b210688a28afc46bcb47fb24c81b02d6c7/cdc/model/sink.go#L1309

hongyunyan (approved):
LGTM

PR #11669, opened 2024-10-17
metric(cdc): fix uncorrect metric WorkerBusyRatio

### LINKED ISSUE #11664: Optimize the metrics for redo module for TiCDC
Currently the metrics for redo module is not clear and correct enough to analyze the working situations for redo. 
Such as
1. It is impossible to achieve 100% worker busy ratio under the current calculation way.
2. The duration for the flush shows strange and uncorrectly when the duration is quite long.
...

To make the metrics more useful and convenient, we need to reorganize the calculation and display of redo metrics.

wk989898:

Issue Number: close #11664

1. Update worker_busy_ratio calculation. 
2. Split redo metrics by log type.

#### Tests 

 - Unit test
 - Integration test
 
Screenshot 2024-10-17 at 19 21 17

Screenshot 2024-10-17 at 19 22 08

### Release note 

```release-note
None
```

### DIFF
--- a/cdc/redo/common/metric.go
+++ b/cdc/redo/common/metric.go
@@ -29,7 +29,7 @@ var (
 		Subsystem: subsystem,
 		Name:      "write_bytes_total",
 		Help:      "Total number of bytes redo log written",
-	}, []string{"namespace", "changefeed"})
+	}, []string{"namespace", "changefeed", "type"})
 
 	// RedoFsyncDurationHistogram records the latency distributions of fsync called by redo writer.
 	RedoFsyncDurationHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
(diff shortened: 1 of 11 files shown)


cdc/redo/common/metric.go
```diff
@@ -38,25 +38,25 @@ var (
 		Name:      "fsync_duration_seconds",
 		Help:      "The latency distributions of fsync called by redo writer",
 		Buckets:   prometheus.ExponentialBuckets(0.001, 2.0, 13),
```
hongyunyan, line 40: We should adjust the bucket setting. In `prometheus.ExponentialBuckets(0.001, 2.0, 13)`, the largest bucket is 4s-8s, which leading when flush time > 4s even 40s， they all just into one bucket. It makes us can't distinguish them clearly.

hongyunyan:
Could you do some tests to check whether the redo metrics now is correctly? Such as 
1. when there are too many events need to write and flush --> to exam whether the busy ratio is 100%?
2. To make a situation where flush is particularly slow --> to see whether the variour flush duration is clear enough
....

cdc/redo/common/metric.go
wk989898: Any suggestions about bucket settings? In the old setting, the flush time > 4s seems as an intolerated situation.

wk989898:
> 1. when there are too many events need to write and flush --> to exam whether the busy ratio is 100%?

this [metric](https://github.com/pingcap/tiflow/pull/11669/files#diff-fe2e80732b903e0dbaf66663c74cf59bc9e03864b418e9b2f526fcdd2a47ea46R19406) statistics work time in past one minute. It will be 100% if work time equals the ticker's duration, which means all time is working time.

hongyunyan (approved):

cdc/redo/common/metric.go
wk989898: 截屏2024-10-29 16 53 36

metrics/grafana/TiCDC-Monitor-Summary.json
```diff
@@ -3165,4 +3165,4 @@
   "title": "Test-Cluster-TiCDC-Summary",
   "uid": "McUpY954z",
   "version": 2
```
3AceShowHand, line 3167: add 1 to this, set version to 3.

metrics/grafana/ticdc.json
```diff
@@ -21652,4 +21652,4 @@
   "title": "${DS_TEST-CLUSTER}-TiCDC",
   "uid": "YiGL8hBZ1",
   "version": 61
```
3AceShowHand, line 21654: set to 62

3AceShowHand (approved):

PR #11674, opened 2024-10-21
sink(ticdc): DDL event support for the Debezium protocol

### LINKED ISSUE #11566: Support ddl event of debezium protocol

#10197

The complete Debezium output format for Kafka sink

_No response_

_No response_

wk989898:

Issue Number: close #11566

Debezium provides the SQL CREATE, ALTER, or DROP statement that results in the schema change. https://debezium.io/documentation/reference/2.7/connectors/mysql.html#mysql-schema-change-topic
This PR implements encoding for the DDL event and checkpoint event.

Difference between TiCDC and Debezium:
- Default length display
   - Debezium output is different when setting the default length. e.g. `int` length is nil but `int(11)` length is 11
- Unsupported data type
  - TiCDC converts DECIMAL to NUMERIC
  - TiCDC converts INTEGER to INT
  - TiCDC converts REAL to FLOAT when setting sql_mode='REAL_AS_FLOAT'
  - TiCDC converts BOOLEAN to TINYINT(1)
- TimeZone
needs to set time_zone='UTC' to compare time. https://debezium.io/documentation/reference/3.0/connectors/mysql.html#mysql-temporal-types
- Generated column
TiCDC does't output the virtual generated column. https://github.com/pingcap/tiflow/issues/11704
- TiCDC does not support some DDLs
  - Alter table add/drop foreign key
  - Adding a new column and setting it to the primary key
  - Dropping primary key columns or columns covered by the composite index
- Debezium does not support some DDLs
  - Alter/Remove TTL
  - Recover table 
- Default value
  - In TiCDC, BLOB, TEXT, GEOMETRY, or JSON column 'binaryRepresentation' can't have a default value
  - BIT `defaultValueExpression` may not same when the length of the default value is not equal to the column length
- Incorrect FLOAT data output
  - Debezium value is incorrect. e.g. The value is "5.61", but Debezium output value is "5.610000133514404"
  - TiCDC incorrect length output. https://github.com/pingcap/tidb/issues/57060
- Incorrect charset name
  - Debezium converts charsetName to "utf8mb4" when column `COLLATE` is "utf8_unicode_ci" and `CHARACTER` is null, but TiCDC doesn't
- Incorrect enum elements
  - Debezium doesn't escape character, but TiCDC does. e.g. ENUM('c,\'d', 'g,''h')
- Incorrect default about TIME
  - TiCDC converts TIME default '1000-00-00 01:00:00.000' to "1000-00-00", but Debezium doesn't
  - Debezium produces the wrong result when the insert value is illegal like "-838:59:59.000000"

#### Tests 

 - Unit test
 - Integration test

yes
### Release note 

```release-note
DDL event support for the Debezium protocol
```


pkg/sink/codec/common/helper.go
```diff
@@ -384,3 +377,42 @@ func SanitizeTopicName(name string) string {
...
+	return *(*[]byte)(unsafe.Pointer(
+		&struct {
+			string
+			Cap int
+		}{s, len(s)},
+	))
+}
+
+// UnsafeStringToBinary create binary string from string without copying
+func UnsafeStringToBinary(s string) string {
```
hongyunyan, line 397: Is this function focus on performance and may be called frequently?
If so, we may be better not to use fmt.Sprintf and += to concat string.

pkg/sink/codec/debezium/codec.go
```diff
@@ -49,10 +50,17 @@ func (c *dbzCodec) writeDebeziumFieldValues(
 	var err error
 	colInfos := tableInfo.GetColInfosForRowChangedEvent()
 	writer.WriteObjectField(fieldName, func() {
+		rows := make([]any, len(tableInfo.Columns))
```
hongyunyan, line 53: what is rows for ?

pkg/sink/codec/debezium/codec.go
```diff
@@ -68,13 +76,31 @@ func (c *dbzCodec) writeDebeziumFieldSchema(
 	switch col.GetType() {
 	case mysql.TypeBit:
 		n := ft.GetFlen()
+		var v uint64
+		var err error
+		if col.GetDefaultValue() != nil {
+			val, ok := col.GetDefaultValue().(string)
+			if !ok {
+				log.Error(
```
hongyunyan, line 84: May be we don't need to print error info here? we return the error info to outside, and maybe outside have print the error already. I'm not sure, you can check again

pkg/sink/codec/debezium/codec.go
```diff
@@ -998,9 +1190,745 @@ func (c *dbzCodec) EncodeValue(
...
+									jWriter.WriteIntField("position", pos+1)
+									jWriter.WriteBoolField("optional", !mysql.HasNotNullFlag(flag))
+
+									autoIncrementFlag := mysql.HasAutoIncrementFlag(flag)
+									jWriter.WriteBoolField("autoIncremented", autoIncrementFlag)
+									jWriter.WriteBoolField("generated", autoIncrementFlag)
+								})
+							}
+						})
+						// if e.TableInfo.Comment != "" {
```
hongyunyan, line 1433: Delete these codes if it's not important.

pkg/sink/codec/debezium/codec.go
```diff
@@ -998,9 +1190,745 @@ func (c *dbzCodec) EncodeValue(
...
+						// 	jWriter.WriteObjectElement(func() {
+						// 		jWriter.WriteStringField("collation", e.Collate)
+						// 	})
+						// })
+					})
+				})
+			})
+		})
+
+		if !c.config.DebeziumDisableSchema {
```
hongyunyan, line 1450: Here seems a lot same code with L1052 and L1787, please extract the same part to a function.

pkg/sink/codec/common/helper.go
```diff
@@ -384,3 +377,42 @@ func SanitizeTopicName(name string) string {
...
+	var result string
+	bs := UnsafeStringToBytes(escapeBackslash(s))
```
CharlesCheung96, line 399: This function is obscure, since:
1. There is no need to use UnsafeStringToBytes, the escape logic can be handled directly when iterating the `s`.

2. The correctness of this function depends on the specific encoding rules. Please explain that there is no problem dealing with utf-8 encoded strings, since the high-bit pattern of UTF-8 bytes is as follows:
```log
Single-byte characters: 0xxxxxxx
Double-byte characters: 110xxxxx 10xxxxxx
Triple-byte characters: 1110xxxx 10xxxxxx 10xxxxxx
Quadruple-byte characters: 11110xxx 10xxxxxx 10xxxxxx 10xxxxxx
```

pkg/sink/codec/debezium/codec.go
wk989898: for the generated column. Remove it since the generated column doesn't output

pkg/sink/codec/common/helper.go
```diff
@@ -384,3 +377,18 @@ func SanitizeTopicName(name string) string {
...
 }
+
+// UnsafeBytesToString create string from byte slice without copying
+func UnsafeBytesToString(b []byte) string {
+	return *(*string)(unsafe.Pointer(&b))
+}
+
+// UnsafeStringToBytes create byte slice from string without copying
+func UnsafeStringToBytes(s string) []byte {
+	return *(*[]byte)(unsafe.Pointer(
```
CharlesCheung96, line 388:
```suggestion
	return *(*[]byte)(unsafe.Pointer(&s))
```

pkg/sink/codec/debezium/codec.go
```diff
@@ -856,6 +1126,542 @@ func (c *dbzCodec) EncodeValue(
...
+			})
+
+			// ts_ms: displays the time at which the connector processed the event
+			// https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-create-events
+			jWriter.WriteInt64Field("ts_ms", c.nowFunc().UnixMilli())
+			jWriter.WriteNullField("transaction")
+			jWriter.WriteStringField("op", "m")
+		})
+
+		if !c.config.DebeziumDisableSchema {
```
hongyunyan, line 1657: Can we also use writeSourceSchema function here?

hongyunyan (approved):
LGTM

PR #11709, opened 2024-11-01
*(dm): add redact log config for dm worker

### LINKED ISSUE #11489: Support query argument redaction in DM logging

When DM logs queries, either with debug logging for all queries, or when encountering errors with warning/error logging, it logs the entire text of the query with arguments. This risks logging sensitive data.

For example, many of the log lines in https://github.com/pingcap/tiflow/blob/master/dm/pkg/conn/baseconn.go will log (truncated) full queries and arguments

Configuration, similar to [tidb log redaction](https://docs.pingcap.com/tidb/stable/log-redaction), that will redact sensitive arguments from queries when DM logs them.

Since they are logged at all possible log levels, no alternative but turning off logging entirely.

N/A

db-will:

Issue Number: close #11489

### Check List 

Add redact log config for dm-worker to control whether to log DM query arguments details in dm worker logs

#### Tests 

 - Unit test
 - Integration test
 - No code

No

Yes, add extra config guides in:
- https://docs.pingcap.com/tidb/stable/log-redaction
- dm config

### Release note 

```release-note
Add redact info log config for dm-worker
```


dm/pkg/log/log.go
```diff
@@ -194,3 +195,34 @@ func WrapStringerField(message string, object fmt.Stringer) zap.Field {
 func WithCtx(ctx context.Context) Logger {
 	return Logger{appLogger.With(getZapFieldsFromCtx(ctx)...)}
 }
+
+var (
+	enabledRedactLog atomic.Value
+)
```
dveeden, line 201: From `make fmt; git diff`:

```suggestion
var enabledRedactLog atomic.Value
```

dveeden:
Can you give some example output/logs? What about adding some tests?

cmd/dm-worker/main.go
```diff
@@ -52,6 +52,7 @@ func main() {
 		Format: cfg.LogFormat,
 		Level:  strings.ToLower(cfg.LogLevel),
 	})
+	log.SetRedactLog(cfg.RedactInfoLog)
```
dveeden, line 55: Should we maybe add a message so it easy to confirm that it is indeed enabled?

I assume it isn't supported to change this on a running instance? otherwise we should probably log in `SetRedactLog()`.

db-will: yes! let me add a log item here

db-will:
@dveeden here are a few examples:

```
[2024/10/31 21:12:04.446 -04:00] [ERROR] [db.go:204] ["execute statements failed after retry"] [task=testdm] [unit="binlog replication"] [queries="[INSERT INTO `testdm`.`t3` (`id`,`uid`,`name`,`info`) VALUES (?,?,?,?)]"] [arguments=?] [error="[code=10006:class=database:scope=not-set:level=high], Message: execute statement failed: begin, RawCause: invalid connection"]
```

db-will:
A quick notes here: when we enable redact info log, since it will redact query arguments from basecon.go/basedb.go etc, it will also redact system generated sql statements query arguments, eg:

```
[2024/11/01 13:23:48.663 -04:00] [DEBUG] [baseconn.go:184] ["execute statement"] [task=testdm] [unit="binlog replication"] [query="INSERT INTO `dm_meta`.`testdm_syncer_checkpoint`\n\t\t(id, cp_schema, cp_table, binlog_name, binlog_pos, binlog_gtid, exit_safe_binlog_name, exit_safe_binlog_pos, exit_safe_binlog_gtid, table_info, is_global) VALUES\n\t\t(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)\n\t\tON DUPLICATE KEY UPDATE\n\t\t\tbinlog_name = VALUES(binlog_name),\n\t\t\tbinlog_pos = VALUES(binlog_pos),\n\t\t\tbinlog_gtid = VALUES(binlog_gtid),\n\t\t\texit_safe_binlog_name = VALUES(exit_safe_binlog_name),\n\t\t\texit_safe_binlog_pos = VALUES(exit_safe_binlog_pos),\n\t\t\texit_safe_binlog_gtid = VALUES(exit_safe_binlog_gtid),\n\t\t\ttable_info = VALUES(table_info),\n\t\t\tis_global = VALUES(is_global);\n\t"] [argument=?]
```

dveeden:
> A quick notes here: when we enable redact info log, since it will redact query arguments from basecon.go/basedb.go etc, it will also redact system generated sql statements query arguments, eg:


Could we change/fix this? And if so, should we? Could there be passwords or other secrets in there?

If any important info like GTID positions are in there, we might have to log them separately

db-will:
Yes, i agree! the current design limiting us to distinguish between system generated queries and workload queries at connection level. and i also notice most of them are Debug level or Error level related.

dveeden (approved):

dm/pkg/log/log.go
```diff
@@ -194,3 +195,32 @@ func WrapStringerField(message string, object fmt.Stringer) zap.Field {
 func WithCtx(ctx context.Context) Logger {
 	return Logger{appLogger.With(getZapFieldsFromCtx(ctx)...)}
 }
+
+var enabledRedactLog atomic.Value
```
lance6716, line 199: `"go.uber.org/atomic".Bool` so don't need type cast

lance6716 (commented):
rest lgtm

lance6716 (approved):

GMHDBJD (approved):
LGTM

yudongusa (approved):

PR #11733, opened 2024-11-08
ddl_puller.go(ticdc): fix DDLs are ignored when schema versions are out of order

### LINKED ISSUE #11714: DDL ignored  by ddl job finishedTs less than puller resolvedTs
assume there are two ddl jobs
Job 60 : `ALTER TABLE a ADD COLUMN x, y, z`， it's commit ts is 400
Job 62:  `ALTER TABLE b ADD COLUMN y` it's commit ts is 300

TiCDC sorts these jobs and run  Job 62 first, then  Job 60
DDL Puller  processed Job 62 and set p.schemaVersion to 62, https://github.com/pingcap/tiflow/blob/d728d0271e804b6759fcca35de80564a062f766a/cdc/puller/ddl_puller.go#L483
 Job 60 will be ignored due to `b.BinlogInfo.SchemaVersion <= p.schemaVersion `  
```
ddl job finishedTs less than puller resolvedTs,discard the ddl job
```

https://github.com/pingcap/tiflow/blob/d728d0271e804b6759fcca35de80564a062f766a/cdc/puller/ddl_puller.go#L321

wlwilliamx:

Issue Number: close #11714

**Description:**
This pull request addresses a bug in TiCDC where a DDL job could be inadvertently dropped if a subsequent DDL job with a higher `SchemaVersion` but lower `CommitTs` is processed first. This behavior occurs because the `ddlJobPuller` processes DDL jobs based on their `CommitTs` in ascending order and uses a check against `SchemaVersion` that can prevent older DDL jobs from being processed, leading to data inconsistencies.

**Background:**
The issue manifests when two DDL jobs are processed:
- Job 60: `ALTER TABLE a ADD COLUMN x, y, z` with `CommitTs` 400
- Job 62: `ALTER TABLE b ADD COLUMN y` with `CommitTs` 300

In this scenario, Job 62 is processed first due to its lower `CommitTs`. The `ddlJobPuller` then updates its `SchemaVersion` to that of Job 62. When Job 60 is subsequently processed, the current logic discards it because its `SchemaVersion` is deemed older, even though its `CommitTs` is higher.

This issue occurs because the `SchemaVersion` increment and job metadata write to TiKV are separate transactions. During a TiDB owner change, different instances might write these transactions without synchronization, leading to potential out-of-order `CommitTs` relative to `SchemaVersion`.

**Solution:**
To resolve this, the check on `SchemaVersion` is removed, ensuring that only the `CommitTs` is verified:

**Updated Code:**
```go
if job.BinlogInfo.FinishedTS <= p.getResolvedTs() {
    log.Info("ddl job finishedTs less than puller resolvedTs," +
        "discard the ddl job",
        zap.String("changefeed", p.changefeedID.ID),
        zap.String("schema", job.SchemaName),
        zap.String("table", job.TableName),
        zap.Uint64("startTs", job.StartTS),
        zap.Uint64("finishedTs", job.BinlogInfo.FinishedTS),
        zap.String("query", job.Query),
        zap.Uint64("pullerResolvedTs", p.getResolvedTs()))
    return true, nil
}
```

**Reasoning:**
The `ResolvedTs` check ensures that only DDL jobs with `FinishedTS` greater than the current `ResolvedTs` are processed. Since the `ddlJobPuller` receives DDLs sorted by `CommitTs`, any new DDL received with a `FinishedTS` greater than `ResolvedTs` must be handled, making the `SchemaVersion` check redundant.

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
Fix the issue that DDL jobs could be incorrectly discarded in TiCDC when their schema versions were not in a strictly linear order with commit timestamps due to TiDB owner changes.
```


3AceShowHand (commented):
can you add some unit test to cover this ?

lidezhu (approved):

PR #11738, opened 2024-11-12
Add a stand-alone load mode to DM. (WIP)

### LINKED ISSUE #9230: engine: support new dm task mode

OliverS929:

Issue Number: close #9230

Primary work to add a stand-alone load mode to DM, alongside with dump mode. Load&Sync mode might not be needed anymore and could be depreciated in the future.

UT/IT added are under testing and will be pushed soon.

#### Tests 

 - [ ] Unit test
 - [ ] Integration test
 - [ ] No code

N/A

Yes. Description of those standalone modes would need to be added to docs.

### Release note 

```release-note

```


dm/config/subtask.go
```diff
@@ -330,6 +331,7 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 	}
 
 	// adjust dir, no need to do for load&sync mode because it needs its own s3 repository
+	// still use dir for standalone load mode (different from the behavior of load&sync mode)
```
lance6716, line 334:
```suggestion
	// adjust dir for modes with load unit except load&sync mode because it needs its own s3 repository
```

BTW because I don't know how does the new "load" mode is used, I'm not sure should DM adjust the dir or not. Maybe the new mode should be similar with "load&sync"?

dm/pkg/terror/error_list.go
```diff
@@ -935,7 +935,7 @@ var (
 	ErrConfigSyncerCfgConflict      = New(codeConfigSyncerCfgConflict, ClassConfig, ScopeInternal, LevelMedium, "syncer-config-name and syncer should only specify one", "Please check the `syncer-config-name` and `syncer` config in task configuration file.")
 	ErrConfigReadCfgFromFile        = New(codeConfigReadCfgFromFile, ClassConfig, ScopeInternal, LevelMedium, "read config file %v", "")
 	ErrConfigNeedUniqueTaskName     = New(codeConfigNeedUniqueTaskName, ClassConfig, ScopeInternal, LevelMedium, "must specify a unique task name", "Please check the `name` config in task configuration file.")
-	ErrConfigInvalidTaskMode        = New(codeConfigInvalidTaskMode, ClassConfig, ScopeInternal, LevelMedium, "please specify right task-mode, support `full`, `incremental`, `all`", "Please check the `task-mode` config in task configuration file.")
+	ErrConfigInvalidTaskMode        = New(codeConfigInvalidTaskMode, ClassConfig, ScopeInternal, LevelMedium, "please specify right task-mode, support `full`, `incremental`, `all`, `dump`, `load`", "Please check the `task-mode` config in task configuration file.")
```
lance6716, line 938: if we don't want to expose the two modes in `dmctl` use case, we shouldn't change this message

engine/jobmaster/dm/task_manager.go
```diff
@@ -287,7 +287,8 @@ func (tm *TaskManager) allFinished(ctx context.Context) bool {
 			if runningTask.Unit != frameModel.WorkerDMLoad {
 				return false
 			}
-		case dmconfig.ModeDump:
+		case dmconfig.ModeDump, dmconfig.ModeLoad, dmconfig.ModeLoadSync:
```
lance6716, line 290:
```suggestion
		case dmconfig.ModeDump, dmconfig.ModeLoad:
```

sync unit will never finish

OliverS929: Thank you for pointing this out. Will change this back.

dm/pkg/terror/error_list.go
OliverS929: Got it. Will change it back.

OliverS929: On a second thought, terror is shared between openapi and dmctl. I think we might want to change this first to make sure our behavior with openapi is correct. Support of these two modes on dmctl could come a bit later. What do you think?

dm/config/subtask.go
OliverS929: I think a comment in "[engine/executor/dm/worker.go](https://github.com/pingcap/tiflow/pull/9228/files#diff-f77f6466a71f2ea3496d26f3540bf784f145e530c8316cb574f35fa5d223dd2d)" mentioned that load&sync mode has its own external storage. From my understanding,load mode usage work with LoaderConfig.Dir, so calling storage.AdjustPath is necessary in this case. also /cc @D3Hunter

OliverS929: Change this further so that load and loadsync both skip the adjust part of the code and use "dir" directly.

PR #11745, opened 2024-11-14
dm: fix upstream switch job test

GMHDBJD:

Issue Number: close #4287 

- add dependency
- fix env variable
- add feishu notification

#### Tests 

 - Integration test

### Release note 

```release-note
None
```


dm/tests/upstream_switch/case.sh
```diff
@@ -2,6 +2,7 @@
 
 set -eu
 
+export DM_MASTER_EXTRA_ARG=""
```
lance6716, line 5: Will it overwrite the DM_MASTER_EXTRA_ARG set by GitHub secret?

lance6716 (approved):

dm/tests/upstream_switch/case.sh
GMHDBJD: just same with other test https://github.com/pingcap/tiflow/blob/68c22f32b2090e83e5ae64ae3b1f25f1a9575ff8/dm/tests/run.sh#L6

D3Hunter (approved):

PR #11794, opened 2024-11-25
ddl puller(ticdc): Fix memory retention issue in ddlPullerImpl's PopFrontDDL method

### LINKED ISSUE #11793: Potential Memory Retention in `PopFrontDDL` Method of `ddlPullerImpl`
### What did you do?

**Problem**

In the `PopFrontDDL` method of the `ddlPullerImpl` struct, the operation `h.pendingDDLJobs = h.pendingDDLJobs[1:]` is used to remove the first element from the `pendingDDLJobs` slice. However, this can lead to memory retention issues due to the behavior of Go slices:
* Slices in Go are references to the underlying array. When an element is removed by slicing, the reference to the element still exists in the underlying array, even though it is no longer part of the slice.
* As a result, the removed element (in this case, the first element of pendingDDLJobs) cannot be garbage collected, leading to unnecessary memory usage if the element is no longer needed and holds significant memory.

The problematic section of the code is as follows:

```go
func (h *ddlPullerImpl) PopFrontDDL() (uint64, *timodel.Job) {
    h.mu.Lock()
    defer h.mu.Unlock()
    if len(h.pendingDDLJobs) == 0 {
        return atomic.LoadUint64(&h.resolvedTS), nil
    }
    job := h.pendingDDLJobs[0]
    h.pendingDDLJobs = h.pendingDDLJobs[1:] // Memory retention issue here
    return job.BinlogInfo.FinishedTS, job
}
```

**Proposed Solution**

To resolve this issue, the reference to the removed element should be explicitly set to nil before updating the slice. This will break the reference to the element, allowing the garbage collector to reclaim the memory.

Updated code:

```go
func (h *ddlPullerImpl) PopFrontDDL() (uint64, *timodel.Job) {
    h.mu.Lock()
    defer h.mu.Unlock()
    if len(h.pendingDDLJobs) == 0 {
        return atomic.LoadUint64(&h.resolvedTS), nil
    }
    job := h.pendingDDLJobs[0]
    h.pendingDDLJobs[0] = nil // Clear reference to allow GC
    h.pendingDDLJobs = h.pendingDDLJobs[1:]
    return job.BinlogInfo.FinishedTS, job
}
```

This issue is not a memory leak but a memory retention problem due to Go’s slice mechanics. The fix should improve the memory efficiency of the system.

No memory retention problem.

### What did you see instead?

Memory retention problem in `PopFrontDDL` method.

### Versions of the cluster

```console
(paste TiCDC version here)
```

wlwilliamx:

Issue Number: close #11793

#### Problem
In the `PopFrontDDL` method of `ddlPullerImpl`, the operation `h.pendingDDLJobs = h.pendingDDLJobs[1:]` creates a new slice by slicing the existing one. However, this leaves the first element (`pendingDDLJobs[0]`) still referenced in the underlying array, potentially causing memory retention issues. 

If the removed element (a `Job` object) occupies significant memory and is no longer needed, this could lead to increased memory usage since it cannot be garbage collected.

#### Solution
The fix ensures that the reference to the removed element is explicitly set to `nil` before updating the slice. This helps the garbage collector to reclaim memory associated with the removed element.

##### Code Changes
- Explicitly set `h.pendingDDLJobs[0] = nil` before updating the slice in `PopFrontDDL`.

Updated code snippet:
```go
func (h *ddlPullerImpl) PopFrontDDL() (uint64, *timodel.Job) {
    h.mu.Lock()
    defer h.mu.Unlock()
    if len(h.pendingDDLJobs) == 0 {
        return atomic.LoadUint64(&h.resolvedTS), nil
    }
    job := h.pendingDDLJobs[0]
    h.pendingDDLJobs[0] = nil // Explicitly clear reference to allow GC
    h.pendingDDLJobs = h.pendingDDLJobs[1:]
    return job.BinlogInfo.FinishedTS, job
}
```

#### Tests 

 - Unit test
 - Integration test

No.

No.

### Release note 

```release-note
Fix a memory retention issue in TiCDC’s DDL job processing to enhance memory efficiency in long-running tasks.
```

### DIFF
--- a/cdc/puller/ddl_puller.go
+++ b/cdc/puller/ddl_puller.go
@@ -734,6 +734,7 @@ func (h *ddlPullerImpl) PopFrontDDL() (uint64, *timodel.Job) {
 		return atomic.LoadUint64(&h.resolvedTS), nil
 	}
 	job := h.pendingDDLJobs[0]
+	h.pendingDDLJobs[0] = nil
 	h.pendingDDLJobs = h.pendingDDLJobs[1:]
 	return job.BinlogInfo.FinishedTS, job
 }


wlwilliamx:
CC @CharlesCheung96 @asddongmen

PR #11815, opened 2024-12-02
debug(ticdc): add more logs to the kafka sink and enable sarama log

### LINKED ISSUE #11823: kafka consumer meet message for the same row with large offset but smaller commit-ts
The kafka consumer receive message for the same row, which at larger offset but the commit-ts is smaller than the previous messages, this may caused by the Kafka sink send message out of order.

At the moment, we think this may caused by the issue:
* https://github.com/IBM/sarama/issues/2619

This cannot be reproduced by the internal tests now, we are adding more test cases.

3AceShowHand:

Issue Number: close #11823 

* add logs to the kafka sink before send out the message
* replace sarama by a custom sarama branch, add more log if retry happens
* add logs to the kafka sink when receive response and detect if commit-ts regressed

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
`None`
```

### DIFF
--- a/cdc/sink/dmlsink/mq/dmlproducer/kafka_dml_producer.go
+++ b/cdc/sink/dmlsink/mq/dmlproducer/kafka_dml_producer.go
@@ -119,8 +119,7 @@ func (k *kafkaDMLProducer) AsyncSendMessage(
 		k.failpointCh <- errors.New("kafka sink injected error")
 		failpoint.Return(nil)
 	})
-	return k.asyncProducer.AsyncSend(ctx, topic, partition,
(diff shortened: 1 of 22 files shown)


flowbehappy (approved):

wk989898 (approved):

CharlesCheung96 (approved):

purelind:
The failure of the dm-compatibility-test is due to Qiniu Cloud's block, which prevents downloads; it has been temporarily bypassed through a PR https://github.com/PingCAP-QE/ci/pull/3269

lance6716 (approved):

Issue #11831, opened 2024-12-05
dm support multi security config for connect TiDB cluster components and tidb-server


River2000i:
### Is your feature request related to a problem?

If the TLS certificates used by the TiDB cluster's internal components and TiDB-server are not issued by the same CA, DM cannot use the physical import mode when enable TLS.

### Describe the feature you'd like

dm support multi security config for connect TiDB cluster components and tidb-server

PR #11844, opened 2024-12-09
dm: openapi support multi TLS security config for downstream db and cluster

River2000i:

Issue Number: close https://github.com/pingcap/tiflow/issues/11831 https://github.com/pingcap/tiflow/issues/11945

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note

```


dm/config/task.go
```diff
@@ -301,6 +302,9 @@ type LoaderConfig struct {
 	RangeConcurrency    int                          `yaml:"range-concurrency" toml:"range-concurrency" json:"range-concurrency"`
 	CompressKVPairs     string                       `yaml:"compress-kv-pairs" toml:"compress-kv-pairs" json:"compress-kv-pairs"`
 	PDAddr              string                       `yaml:"pd-addr" toml:"pd-addr" json:"pd-addr"`
+	// now only creating task by OpenAPI will use the `Security` field to connect PD.
+	// TODO: support setting `Security` by dmctl
+	Security *security.Security `yaml:"-" toml:"security" json:"security"`
```
River2000i, line 307: not support yaml format since it will block few engine test, also we not support set the filed by yaml.

dm/checker/checker.go
```diff
@@ -550,16 +547,6 @@ func (c *Checker) Init(ctx context.Context) (err error) {
 	return nil
 }
 
-func lightningCheckGroupOnlyTableEmpty(checkingItems map[string]string) bool {
```
GMHDBJD, line 553: why remove this check

GMHDBJD (commented):
rest lgtm

dm/checker/checker.go
River2000i: To fix this issue https://github.com/pingcap/tiflow/issues/11945. Root cause is this check will set pd addr to `noop:2379`. In fact, when we run checker by `RunCheckOnConfigs()` https://github.com/pingcap/tiflow/blob/master/dm/checker/cmd.go#L113, it will initial all checker and remove few ignored checkers https://github.com/pingcap/tiflow/blob/master/dm/checker/cmd.go#L49. But the rest checker do not ignored, will use pdClient to get some info.

River2000i:
@GMHDBJD @D3Hunter PTAL~

GMHDBJD (approved):
LGTM

dm/config/security/security.go
```diff
@@ -95,5 +97,23 @@ func (s *Security) Clone() *Security {
...
 	clone.SSLKeyBytes = append([]byte(nil), s.SSLKeyBytes...)
 	clone.SSLCertBytes = append([]byte(nil), s.SSLCertBytes...)
+	clone.SSLCA = s.SSLCA
+	clone.SSLCert = s.SSLCert
+	clone.SSLKey = s.SSLKey
 	return &clone
 }
+
+// WriteTLSContentToFiles write tls content to temp file and update tls path fields.
+func (s *Security) WriteTLSContentToFiles(fileName string) error {
```
D3Hunter, line 107: when and why write to temp file?

dm/config/security_test.go
```diff
@@ -171,3 +171,21 @@ func (c *testTLSConfig) TestClone() {
 	clone.CertAllowedCN[0] = "g"
 	c.Require().NotEqual(s, clone)
 }
+
+func (c *testTLSConfig) TestWriteTLSContentToFiles() {
```
D3Hunter, line 175: can you write new test with `"github.com/stretchr/testify/require"`, we ever planned to replace all those old ones with this, but only half done

dm/loader/lightning.go
```diff
@@ -329,6 +329,41 @@ func GetLightningConfig(globalCfg *lcfg.GlobalConfig, subtaskCfg *config.SubTask
 	if err := cfg.LoadFromGlobal(globalCfg); err != nil {
 		return nil, err
 	}
+	cfg.TiDB.Security = &globalCfg.Security
+	// TODO: Using TLS content cannot verify certificates correctly when lightning access PD server.
```
D3Hunter, line 333:
```suggestion
	// TODO: avoid writing to local file. right now we don't know how to verify certificates correctly using TLS content in a short time, but we have a time schedule to keep.
```
so the reason we write to file is we hasn't found the way to fix this in short time?

D3Hunter (approved):

dm/config/security/security.go
River2000i: when write to temp file need meet few condition:
1.  when we call `GetLightningConfig()`
2. subtask config set `LoaderConfig.Security` (now only openapi can set the field)
3. `LoaderConfig.Security` only set certificates content and do not set certificates path.

> why write to temp file
Since only set certificates content cannot verify certificates correctly. Yes, just like your comment below.

dm/loader/lightning.go
River2000i: yes

dm/config/security_test.go
River2000i: fixed

yudongusa (approved):

PR #11875, opened 2024-12-13
codec(ticdc): canal-json encode table id

### LINKED ISSUE #11874: canal-json protocol also output the table id and table partition id

At the moment, canal-json message does not output the table id and table partition id.

In the partition table scenario, since all table partitions have the same table name, so the consumer does not know which table partition the row belongs to, the consumer may want to know that information to do some specific works.

In the `tidb_extension` field, add 2 new fields, `table_id` and `table_partition_id`.

The `table_id` is the logic table id of the partition table, the `table_partition_id` is the physical table id of the table partition.

For example:
```
create table t (a int primary key, bint) partition by range(a)
(
  partition p0 values less than (10),
  partition p1 values less than (20),
);
```

For the table t, the logical table id should be 110, but for partition p0 and p1, each table_partition_id is `111` and `112`, so the consumer can distinguish rows belongs to different table partitions.

_No response_

* add documentation to mention that also output the `table_id` and `partition_table_id`.

3AceShowHand:

Issue Number: close #11874

* https://github.com/pingcap/tiflow/pull/11875/files#diff-55f12a539f579d4f81c66eb444ea14b9c017be1e1826392fc351aaff0e2229bcR258-R269

encode logical table id for all DML events, also encode the partition table id for the partition table

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
canal-json output table id and partition id in the tidb_extension field
```


pkg/sink/codec/canal/canal_json_row_event_encoder_test.go
```diff
@@ -735,7 +753,9 @@ func TestE2EPartitionTable(t *testing.T) {
 		decodedEvent, err := decoder.NextRowChangedEvent()
 		require.NoError(t, err)
 		// canal-json does not support encode the table id, so it's 0
```
wk989898, line 755: This comment should be removed

asddongmen (approved):

wk989898 (approved):

Issue #11879, opened 2024-12-16
Tracking integration test failed case


wk989898:
There are some failed integration tests because of unstable cases. We need to track these issues.

PR #11880, opened 2024-12-16
test(ticdc): fix batch_add_table test case

wk989898:

Issue Number: ref #11879

set more times to check if the table exists

#### Tests 

 - Integration test

### Release note 

```release-note
`None`.
```


wk989898:
close with https://github.com/pingcap/tiflow/pull/11886

PR #11895, opened 2024-12-17
consumer(ticdc): canal-json decoder cache table info for each message to avoid allocate memory on each new message

### LINKED ISSUE #11894: kafka consumer takes too much memory when using canal-json protocol
Image

A large portion of the memory allocated when build table information for each individual events, this should can be reduced

### LINKED ISSUE #11907: canal-json encoder panic if set only output updated column to true when check comparison on the vector type
### What did you do?

1. create canal-json encoder, set the `only-output-updated-columns` to true

the encoder panic

### What did you see instead?

```
--- FAIL: TestCanalJSONContentCompatibleE2E (155.03s)
panic: runtime error: comparing uncomparable type types.VectorFloat32 [recovered]
	panic: runtime error: comparing uncomparable type types.VectorFloat32
```

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
master
```

3AceShowHand:
…sage

Issue Number: close #11894, close #11907

* canal-json decoder reuse table info, instead create new one for each message
* fix the vector type failed on the comparison by using `==`

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```


pkg/sink/codec/canal/canal_json_decoder.go
```diff
@@ -263,6 +274,52 @@ func (b *batchDecoder) assembleHandleKeyOnlyRowChangedEvent(
...
+		name := col.Name.O
+		if _, ok := pkNames[name]; ok {
+			indexColumns = append(indexColumns, &timodel.IndexColumn{
+				Name:   pmodel.NewCIStr(name),
+				Offset: idx,
+			})
+		}
+	}
+	indexInfo := &timodel.IndexInfo{
+		ID:      1,
```
wk989898, line 314: Please follow the previous behavior. https://github.com/pingcap/tiflow/blob/master/cdc/model/sink.go#L917-L920

3AceShowHand: there is only primary key, and should 1, it's the previous behavior.

wk989898 (approved):

pkg/sink/codec/canal/canal_json_decoder.go
```diff
@@ -48,6 +56,8 @@ type batchDecoder struct {
 
 	upstreamTiDB *sql.DB
 	bytesDecoder *encoding.Decoder
+
+	tableInfoCache map[tableKey]*model.TableInfo
```
lidezhu, line 60: after drop table and create table, table with same schema name and table name may have different TableInfo?

3AceShowHand: cached item is deleted if receive table level DDL, and recreate a new table info.

lidezhu (approved):

PR #11903, opened 2024-12-17
puller: fix retry logic when check store version failed

### LINKED ISSUE #11766: cdc: fix usage of tikv go-client
### What did you do?

scale out a tikv node

changefeed resolved ts can advance normally.

### What did you see instead?

changefeed resolved ts stuck, and cdc cannot connect to the new tikv store because `GetTiKVRPCContext` return wrong info for the store.

### Versions of the cluster

nightly

lidezhu:

Issue Number: close #11766 

Change the retry logic to reload region when `client.GetStore` failed.

#### Tests 

 - Unit test

### Release note 

```release-note
Fix the problem that changefeed may get stuck after scaling out new tikv nodes.
```

### DIFF
--- a/cdc/kv/shared_stream.go
+++ b/cdc/kv/shared_stream.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pingcap/log"
 	"github.com/pingcap/tiflow/cdc/kv/sharedconn"
 	"github.com/pingcap/tiflow/pkg/chann"
+	cerrors "github.com/pingcap/tiflow/pkg/errors"
 	"github.com/pingcap/tiflow/pkg/util"
 	"github.com/pingcap/tiflow/pkg/version"
 	"go.uber.org/zap"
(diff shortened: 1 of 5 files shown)


cdc/kv/shared_stream.go
```diff
@@ -125,16 +140,7 @@ func newRequestedStream(streamID uint64) *requestedStream {
...
-		select {
-		case <-ctx.Done():
-			return true
-		default:
-			return false
-		}
-	}
-
+func (s *requestedStream) run(ctx context.Context, c *SharedClient, rs *requestedStore) error {
 	if err := version.CheckStoreVersion(ctx, c.pd, rs.storeID); err != nil {
```
hicqu, line 144: Seems we just need to move `CheckStoreVersion` out from `run`.

lidezhu: Change `requestedStream.run` from return `bool` to return `error` enable us to handle more kinds of error in the future.
And what's the benefits of move `CheckStoreVersion` out from run?

asddongmen:
Below is my understanding of how this PR works. Please correct me if I am wrong.

**Initial Problem:**
1. When the region fails to connect to the current store, the system attempts to switch to the next store address.

2. After switching, the system enters the newStream function and calls stream.run.

3. Since stream.run fails, the system retrieves the next store address from the region cache.

4. However, due to **incorrect information in the region cache**, the retrieved store is always unreachable, causing the system to fall into an infinite loop, repeatedly performing the above 1,2,3 steps.

**Fix:**
1. When the region fails to connect to the current store, log the error and attempt to switch to the next store address.

2. After switching to the next store address, proceed to the `newStream` function.

3. Within the newStream function, call `stream.run`. If `stream.run` fails, handle the error based on its type.

4. If the error type indicates that the information in the region cache might be incorrect, **reset the region cache** to ensure that the next retrieved store address is accurate.

asddongmen (approved):

3AceShowHand (approved):

cdc/kv/shared_stream.go
```diff
@@ -90,12 +90,28 @@ func newStream(ctx context.Context, c *SharedClient, g *errgroup.Group, r *reque
...
+					zap.String("namespace", c.changefeed.Namespace),
+					zap.String("changefeed", c.changefeed.ID),
+					zap.Uint64("streamID", stream.streamID),
+					zap.Uint64("storeID", r.storeID),
+					zap.String("addr", r.storeAddr),
+					zap.Error(err))
+				if errors.Cause(err) == context.Canceled {
+					return nil
+				}
+				regionErr = &getStoreErr{}
```
hicqu, line 105: It's better to move `getStoreErr` into `CheckStoreVersion`, because `CheckStoreVersion` may fails really without the `getStore` bug.

lidezhu: OK, change it to check the error type return by `CheckStoreVersion`.

hicqu (approved):

PR #11911, opened 2024-12-19
codec(ticdc): reduce canal-json decode memory consumption by using buffered json decoder

3AceShowHand:

Issue Number: ref #11894

This is the second part to reduce the memory cost by the canal-json decoder when works with the consumer.

* use json decoder to decode events, this looks takes relative little memory than the `Unmarshal` method, but still takes a lot.
* add `CanalJSONTxnDecoder`, split related logic, to simplify the normal canal-json decoder.

image

* After this PR, JSON decoder phase takes around 40% memory. Previous takes more than 60%. 

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```


wk989898:
How much memory will be reduced after applying this?

3AceShowHand:
> How much memory will be reduced after applying this?

previous, takes around 62%. now takes around 40%. 

I tried several json library, all takes a lot of memory in the consumer scenario.

wk989898 (approved):

PR #11934, opened 2024-12-25
consumer(ticdc): do not sort to detect out of order events

### LINKED ISSUE #11947: kafka consumer should not sort events
after https://github.com/pingcap/tiflow/issues/11874 gets merged, canal-json encoded message can be distinguished which table partition the row comes from, so there is no need to sort events.

If any commit-ts regression panic issue found, we can think it as unknown problems comes from the producer side.

3AceShowHand:

Issue Number: close #11947

* do not sort events when resolve events by the resolve ts

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`
```


asddongmen (approved):

cmd/kafka-consumer/writer.go
```diff
@@ -379,31 +386,24 @@ func (w *writer) WriteMessage(ctx context.Context, message *kafka.Message) bool
 			w.checkPartition(row, partition, message.TopicPartition.Offset)
 
 			tableID := row.GetTableID()
-			if w.option.protocol != config.ProtocolSimple {
+			switch w.option.protocol {
+			case config.ProtocolSimple, config.ProtocolCanalJSON:
```
wk989898, line 390: Why exclude open protocol?

3AceShowHand: Simple and Canal-JSON, the table id is already encoded into the message.

Open-Protocol, table id is only set for the partition table, so we have to generate fake table id for all open protocol encoded messages.

wk989898 (approved):

PR #11956, opened 2024-12-31
filter(ticdc): fix incorrect event filter with "rename" DDLs

### LINKED ISSUE #11946: In [[filter.event-filters]], with "rename" DDLs the `matcher` is filtered based on the new name rather than old name
### What did you do?

1. Create 3 tables:

    ```sql
    use test;
    create table a1 (pk int primary key, val int);
    create table a2 like a1;
    create table b3 like a1;
    ```

2. Create a changefeed with configuration:

    ```toml
    [filter]
    rules = ['test.a*', 'test.b*']
    [[filter.event-filters]]
    matcher = ['test.a*']
    ignore-event = ['all ddl']
    ```

    ```sh
    cdc cli changefeed create --sink-uri='file:///tmp/test-cdc-storage?protocol=canal-json' --config=cfg.toml -c xxx
    ```

3. Perform the following renames:

    ```sql
    rename table a1 to b1;
    rename table a2 to c2;
    rename table b3 to a3;
    ```

As explained in https://docs.pingcap.com/tidb/stable/ticdc-ddl#ddl-replication-considerations-for-renaming-tables, since `filter.rules` is looking at the *old name*, we expect `filter.event-filters[].matcher` is also looking at the *old name* too. That means the downstream should *ignore* the first two renames and only replicate the last.

```sql
-- rename table a1 to b1;  -- ignored
-- rename table a2 to c2;  -- ignored
rename table b3 to a3;  -- replicated
```

### What did you see instead?

The `filter.event-filters[].matcher` rule actually looks at the *new name*, so instead the first two DDLs are replicated while the last one is ignored.

```sql
rename table a1 to b1;  -- incorrectly replicated
rename table a2 to c2;  -- incorrectly replicated
-- rename table b3 to a3;  -- incorrectly ignored
```

### Versions of the cluster

```console
v8.5.0
```

```console
v8.5.0
```

```console
v8.5.0
```

wk989898:

Issue Number: close #11946

Event Filter wrongly uses the current table info with "rename" DDLs and It should use the previous table info to filter events.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
Fix incorrect event filter with "rename table" DDLs
```

### DIFF
--- a/pkg/filter/sql_event_filter.go
+++ b/pkg/filter/sql_event_filter.go
@@ -144,7 +144,7 @@ func (f *sqlEventFilter) shouldSkipDDL(ddl *model.DDLEvent) (skip bool, err erro
 			log.Warn("sql event filter doesn't find old table info when the event type is `rename table`",
 				zap.String("schema", schema),
 				zap.String("table", table),
-				zap.String("type", ddl.Type.String()),
+				zap.Stringer("type", ddl.Type),
 				zap.String("query", ddl.Query))
 			return true, cerror.ErrTableIneligible.GenWithStackByArgs(ddl.TableInfo.TableName)
 		}
--- a/pkg/filter/sql_event_filter.go
+++ b/pkg/filter/sql_event_filter.go
@@ -136,9 +136,21 @@ func (f *sqlEventFilter) shouldSkipDDL(ddl *model.DDLEvent) (skip bool, err erro
 	if len(f.rules) == 0 {
 		return false, nil
 	}
+	evenType := ddlToEventType(ddl.Type)
 	schema := ddl.TableInfo.TableName.Schema
 	table := ddl.TableInfo.TableName.Table
-	evenType := ddlToEventType(ddl.Type)
+	if evenType == bf.RenameTable {
+		if ddl.PreTableInfo == nil {
+			log.Warn("sql event filter doesn't find old table info when the event type is `rename table`",
+				zap.String("schema", schema),
+				zap.String("table", table),
+				zap.String("type", ddl.Type.String()),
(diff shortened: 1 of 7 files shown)


kennytm:
does it work if the DDL renamed multiple tables (e.g. `rename a1 to b1, a2 to b2, b3 to a3;`)?

wk989898:
> does it work if the DDL renamed multiple tables (e.g. `rename a1 to b1, a2 to b2, b3 to a3;`)?

Yes, you can check the integration test.

pkg/filter/sql_event_filter.go
```diff
@@ -136,9 +136,21 @@ func (f *sqlEventFilter) shouldSkipDDL(ddl *model.DDLEvent) (skip bool, err erro
...
+	evenType := ddlToEventType(ddl.Type)
 	schema := ddl.TableInfo.TableName.Schema
 	table := ddl.TableInfo.TableName.Table
-	evenType := ddlToEventType(ddl.Type)
+	if evenType == bf.RenameTable {
+		if ddl.PreTableInfo == nil {
+			log.Warn("sql event filter doesn't find old table info when the event type is `rename table`",
+				zap.String("schema", schema),
+				zap.String("table", table),
+				zap.String("type", ddl.Type.String()),
```
kennytm, line 147:
```suggestion
				zap.Stringer("type", ddl.Type),
```

kennytm (approved):
rest lgtm

asddongmen (approved):

kennytm:
i haven't tested but i think the issue affected 6.5 and 7.1 too

PR #12027, opened 2025-01-20
*: add warning to deprecate config sink.enable-partition-separator in the future

### LINKED ISSUE #11979: data of partition table may lost when sink type is storage services and enable-partition-separator is false
### What did you do?

1. create a changefeed with sink type set to storage services and enable-partition-separator set to false;
2. create a partition table with multiple partitions and write data;

All rows are written to storage services;

### What did you see instead?

Some rows are lost

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
(paste TiCDC version here)
```

lidezhu:

Issue Number: close #11979

Report a warning to the user if config sink.enable-partition-separator is set explicitly in the config file.

#### Tests 

1. create a changefeed with sink.enable-partition-separator set in config file.
image

2. update the changefeed with sink.enable-partition-separator set in config file.
image

### Release note 

```release-note
`None`.
```

### DIFF
--- a/pkg/cmd/cli/cli_changefeed_create.go
+++ b/pkg/cmd/cli/cli_changefeed_create.go
@@ -19,6 +19,7 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/BurntSushi/toml"
 	"github.com/fatih/color"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/log"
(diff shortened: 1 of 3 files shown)


hongyunyan (approved):

3AceShowHand (approved):

pkg/cmd/cli/cli_changefeed_create.go
```diff
@@ -187,12 +188,27 @@ func (o *createChangefeedOptions) completeReplicaCfg() error {
...
+	}
+
+	if len(o.commonChangefeedOptions.configFile) > 0 {
+		cfg := config.GetDefaultReplicaConfig()
+		metaData, err := toml.DecodeFile(o.commonChangefeedOptions.configFile, cfg)
+		if err != nil {
+			return errors.Trace(err)
+		}
+
+		if metaData.IsDefined("sink", "enable-partition-separator") {
```
asddongmen, line 201: Can we ensure it has no effect when set to false?  Since this could potentially lead to data inconsistency.

lidezhu: Yes, but we decide to change the behavior in later versions. You can check the comments in the issue.

PR #12038, opened 2025-01-24
sink(ticdc): fix incorrect `default` field

### LINKED ISSUE #12037: incorrect default value after altering an ADDed column from NOT NULL to NULL
### What did you do?

run sql
```sql
create table t1(
    id int primary key
);
alter table t1 add column col json not null;
alter table t1 modify column col json default null;
insert into t1(id) values(1);
```

the default value of `col` should be null

### What did you see instead?

the default value of `col` is string literal "null"

### Versions of the cluster

```console
master
```

```console
master
```

wk989898:

Issue Number: close #12037

`OriginDefaultValue` is used as the column value when the default value is invisible. 
It should not be used as the `default` field.

**Affected protocol**
- Cloud Storage
- Simple
- Avro
- Debezium

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
fix incorrect `default` field.
```

### DIFF
--- a/cdc/model/schema_storage.go
+++ b/cdc/model/schema_storage.go
@@ -501,9 +501,6 @@ func (ti *TableInfo) GetPrimaryKeyColumnNames() []string {
 // GetColumnDefaultValue returns the default definition of a column.
 func GetColumnDefaultValue(col *model.ColumnInfo) interface{} {
 	defaultValue := col.GetDefaultValue()
-	if defaultValue == nil {
-		defaultValue = col.GetOriginDefaultValue()
-	}
 	defaultDatum := datumTypes.NewDatum(defaultValue)
 	return defaultDatum.GetValue()
 }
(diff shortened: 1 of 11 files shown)


flowbehappy (approved):

cdc/model/schema_storage.go
```diff
@@ -501,9 +501,6 @@ func (ti *TableInfo) GetPrimaryKeyColumnNames() []string {
 // GetColumnDefaultValue returns the default definition of a column.
 func GetColumnDefaultValue(col *model.ColumnInfo) interface{} {
 	defaultValue := col.GetDefaultValue()
```
3AceShowHand, line 503: It looks can return the `col.GetDefaultValue()` directly.

3AceShowHand:
Is the method `GetColumnDefaultValue` only used by encoding protocols ?

Will this modification also affect other components?

wk989898:
> Is the method `GetColumnDefaultValue` only used by encoding protocols ?
> 
> Will this modification also affect other components?

I tracked this method and found it was only used in the schema encoding.

3AceShowHand:
`GetColumnDefaultValue` can be removed, just call `col.GetDefaultValue()` should works.

lidezhu (approved):
LGTM

asddongmen (approved):

PR #12040, opened 2025-01-29
syncer(dm): option for explicit auto-id-cache handling in create table DDLs

### LINKED ISSUE #12039: Allow explicit control of `AUTO_ID_CACHE` in DM incremental sync of new tables

We cannot update the `AUTO_ID_CACHE` setting of a table after its creation.

When we replicate new tables via incremental DM, the `CREATE TABLE` statements are applied with default auto-id-cache size, which is problematic if we want to control it directly, or if we want to use MySQL compatibility mode w/ `AUTO_ID_CACHE=1`. 

We'd like to have control to explicitly set it in create table DDLs that are incrementally replicated in DM. When DM syncs a new table w/ an AUTO_INCREMENT column, the syncer can modify the DDL to set an explicit auto-id-cache size based on the syncer configuration.

We propose a new `auto-id-cache-size` Syncer config. When set, DM will update the create statement w/ the corresponding auto-id-cache size. When unset (0 or default), the statement will be left as is.

_No response_

_No response_

michaelmdeng:

Issue Number: close #12039

Adds a new `create-table-auto-id-cache-size` Syncer config. `create-table` refers to the fact that this config affects `CREATE TABLE` statements, and when set, DM incremental replication of new tables will update the `create table` statement w/ the corresponding auto-id-cache size. When unset (0 or default), the statement will be left as is.

```
syncers:
  <name>:
    auto-id-cache-size: 1 # This sets MySQL-compatiblity mode for new tables
    ...

...
```

I believe this change should not affect other DM components, including import for "full" or "all" mode, but would appreciate guidance here as well.

#### Tests 

 - [x] Unit test
 - Integration test
 - No code

No

Yes, needs documentation on the new config setting and its purpose/usage.

### Release note 

```release-note
Allow control of `AUTO_ID_CACHE` size when DM in incremental mode creates new tables
```

### DIFF
--- a/dm/openapi/fixtures/task.go
+++ b/dm/openapi/fixtures/task.go
@@ -41,7 +41,7 @@ var (
 			  "cert_allowed_cn": ["PD1", "PD2"]
 			}
 		  },
-		  "incr_migrate_conf": { "repl_batch": 200, "repl_threads": 32 },
+		  "incr_migrate_conf": { "auto_id_cache_size": 0, "repl_batch": 200, "repl_threads": 32 },
 		  "source_conf": [{ "source_name": "mysql-replica-01" }]
 		},
 		"table_migrate_rule": [
@@ -83,7 +83,7 @@ var (
 			"export_threads": 4,
 			"import_threads": 16
 		  },
-		  "incr_migrate_conf": { "repl_batch": 200, "repl_threads": 32 },
+		  "incr_migrate_conf": { "auto_id_cache_size": 0, "repl_batch": 200, "repl_threads": 32 },
 		  "source_conf": [{ "source_name": "mysql-replica-01" }]
 		},
 		"table_migrate_rule": [
@@ -126,7 +126,7 @@ var (
 			"import_mode": "logical",
 			"import_threads": 16
 		  },
-		  "incr_migrate_conf": { "repl_batch": 200, "repl_threads": 32 },
+		  "incr_migrate_conf": { "auto_id_cache_size": 0, "repl_batch": 200, "repl_threads": 32 },
 		  "source_conf": [
 			{
 			  "binlog_gtid": "",


michaelmdeng:
@lance6716 sorry about the delay, just managed to fix the relevant integration tests, can you take a look and also check if the release notes and any docs/education look ok?

OliverS929:
Since a new configuration is added in here, could you also add this option to OpenAPI so we can keep it aligned on both ends?

OliverS929:
@michaelmdeng Just a quick reminder — one of the two Git email addresses you used for commits in this PR hasn’t signed the CLA yet. Please make sure to sign it so we can proceed with the review. Thanks!

dm/config/task.go
```diff
@@ -1175,6 +1177,7 @@ type SyncerConfigForDowngrade struct {
 	SafeModeDuration string `yaml:"safe-mode-duration,omitempty"`
 	Compact          bool   `yaml:"compact,omitempty"`
 	MultipleRows     bool   `yaml:"multipleRows,omitempty"`
+	AutoIDCacheSize  uint64 `yaml:"auto-id-cache-size,omitempty"`
```
lance6716, line 1180: It's better to skip this field to `SyncerConfigForDowngrade`. This structure is used to export the config and import it to an old version of DM cluster. The unknown field will cause problems

dm/config/task_test.go
```diff
@@ -1196,3 +1196,30 @@ func TestTaskYamlForDowngrade(t *testing.T) {
...
+			syncerConfigName: {
+				AutoIDCacheSize: 100,
+			},
+		},
+	}
+	content, err := originCfg.YamlForDowngrade()
+	require.NoError(t, err)
+	newCfg := &TaskConfig{}
+	require.NoError(t, newCfg.FromYaml(content))
+	require.Equal(t, originCfg.Syncers[syncerConfigName].AutoIDCacheSize, newCfg.Syncers[syncerConfigName].AutoIDCacheSize)
```
lance6716, line 1224: After above change, this check should expected to have a NotEqual result.

dm/syncer/ddl_test.go
```diff
@@ -789,6 +789,66 @@ func TestAdjustCollation(t *testing.T) {
...
+			"CREATE TABLE `test`.`t1` (`id` INT AUTO_INCREMENT)",
+			"CREATE TABLE `test`.`t1` (`id` INT AUTO_INCREMENT) /*T![auto_id_cache] AUTO_ID_CACHE = 1 */",
+			1,
+		},
+		{
+			"CREATE TABLE `test`.`t1` (`id` INT AUTO_INCREMENT)",
+			"CREATE TABLE `test`.`t1` (`id` INT AUTO_INCREMENT) /*T![auto_id_cache] AUTO_ID_CACHE = 30 */",
+			30,
+		},
+	}
```
lance6716, line 818: seems you didn't cover cases that original CREATE TABLE already set auto_id_cache

lance6716 (commented):
rest lgtm

dm/config/task.go
```diff
@@ -404,6 +404,8 @@ type SyncerConfig struct {
 	Compact      bool `yaml:"compact" toml:"compact" json:"compact"`
 	MultipleRows bool `yaml:"multiple-rows" toml:"multiple-rows" json:"multiple-rows"`
 
+	AutoIDCacheSize uint64 `yaml:"auto-id-cache-size" toml:"auto-id-cache-size" json:"auto-id-cache-size"`
```
OliverS929: It seems that the new config "auto-id-cache-size" does not have a default value attached to it when it is not included in the config. I think it would be great if we could have a explicit default value here that aligns with behavior from previous versions, for clarity purpose.

dm/syncer/ddl_test.go
michaelmdeng: thanks, added a relevant test case

michaelmdeng:
@lance6716 addressed your feedback and made a couple of naming changes based on discussion in #12039

lance6716 (approved):

michaelmdeng:
@lance6716 can I get help w/ approval for the dm/config changes?

lance6716:
ptal @D3Hunter @GMHDBJD @OliverS929

michaelmdeng:
@OliverS929 thanks for taking a look! wanted to tag you because we were specifically discussing this in the upstream issue

OliverS929:
Hi @michaelmdeng, thanks for tagging me again. I noticed that my earlier comment about default value initialization in the syncer config hasn't received a response yet.

I believe it's important to ensure default values are properly set when creating a new sync subtask. Could you take a look and consider making the suggested changes?

For reference, please see 
https://github.com/pingcap/tiflow/blob/master/cmd/dm-syncer/config.go#L63 
and
https://github.com/pingcap/tiflow/blob/master/cmd/dm-syncer/config.go#L146

michaelmdeng:
> Hi @michaelmdeng, thanks for tagging me again. I noticed that my earlier comment about default value initialization in the syncer config hasn't received a response yet.
> 
> I believe it's important to ensure default values are properly set when creating a new sync subtask. Could you take a look and consider making the suggested changes?

sorry @OliverS929 , not sure I understand the request here. those are configs for the `dm-syncer` cmd/binary, which appears to me to be unused?

and when/if `dm-syncer` is run, it should initialize configs either [here](https://github.com/pingcap/tiflow/blob/master/cmd/dm-syncer/config.go#L147) or from parsing the config toml file, which uses the default `SubTaskConfig` and `SyncerConfig` struct initialization which will set `AutoIDCacheSize` to 0, which is the desired default behavior.

OliverS929:
> > Hi @michaelmdeng, thanks for tagging me again. I noticed that my earlier comment about default value initialization in the syncer config hasn't received a response yet.
> > I believe it's important to ensure default values are properly set when creating a new sync subtask. Could you take a look and consider making the suggested changes?
> 
> sorry @OliverS929 , not sure I understand the request here. those are configs for the `dm-syncer` cmd/binary, which appears to me to be unused?
> 
> and when/if `dm-syncer` is run, it should initialize configs either [here](https://github.com/pingcap/tiflow/blob/master/cmd/dm-syncer/config.go#L147) or from parsing the config toml file, which uses the default `SubTaskConfig` and `SyncerConfig` struct initialization which will set `AutoIDCacheSize` to 0, which is the desired default behavior.

I understand that this is handled by Go’s struct defaulting mechanism. I was just hoping that AutoIDCacheSize (and similar fields) could be explicitly initialized with comments, so it's more obvious to future readers and maintainers. That said, I agree it doesn't affect the functionality of this PR.

OliverS929:
rest LGTM

alastori:
> /cc @alastori. Because this PR involves modification to DM config, I think it would need approval from you and Ben.

@OliverS929, thanks for asking. The longer key name `create-table-auto-id-cache-size` looks great and matches what we agreed on in the original issue discussion ([link](https://github.com/pingcap/tiflow/issues/12039#issuecomment-2704884778)).

@michaelmdeng Tiny nit: the sample YAML in the PR description still shows the old
`auto-id-cache-size`. Could you update that snippet (and any other docs/tests) to use the new name so everything stays consistent before we merge?

Thanks!

Issue #12068, opened 2025-02-21
MQ fields like DispatchRules not getting honor in pulsar+http protocol for pulsar changefeeds.


SandeepPadhi:
### What did you do?

Steps:
1. Create a pulsar changefeed with pulsar+http protocol.
2. In changefeed payload, add dispatchRules.
3. create the changefeed. 

### What did you expect to see?

DispatcheRules as stated in the replication Config

### What did you see instead?

DispatchRules did not get picked up.

### Versions of the cluster

TiCDC version (execute `cdc version`):

```console
v8.5.1
```

PR #11338, opened 2024-06-24
ticdc : http & https support in pulsar schema for pulsar downstream type changefeed creation.

SandeepPadhi:
### What problem does this PR solve?
Added support for HTTP and HTTPS schema for pulsar downstream. Currently, only pulsar and pulsar+ssl support is there.

Issue Number: close #11336 

### What is changed and how it works?
-> In addition to pulsar and pulsar+ssl schema types in pulsar changefeed creation , users will now will able to create changefeeds with HTTP and HTTPS  type as broker schema for pulsar client.   

Clients have to use: 
1. pulsar+http -> for creating pulsar changefeed with HTTP endpoint
2. pulsar+https -> for creating pulsar chnagefeed with HTTPS endpoint.

#### Tests 
1. Unit test
2. Feature test by deploying the image on  k8s environment and creating multiple changefeeds.

##### Will it cause performance regression or break compatibility?
-> It should not cause any performance regression or break compatibility. It only addition schema addition for pulsar client creation.

##### Do you need to update user documentation, design documentation or monitoring documentation?

Documentation for pulsar sink should be update: https://docs.pingcap.com/tidb/dev/ticdc-sink-to-pulsar#sink-uri

```release-note
Support http and https schema in pulsar sink uri. 
```

PR #12069, opened 2025-02-21
ticdc:bug fix - added pulsar+http and pulsar+https protocols in IsMQScheme to honor MQ fields in changefeed

SandeepPadhi:
MQ fields like DispatchRules not getting honor in pulsar+http protocol for pulsar changefeeds because pulsar+http and pulsar+https protocols were not added iin IsMQScheme function. It is bug fix for the same. The bug was introduced in below PR:https://github.com/pingcap/tiflow/pull/11338

Issue Number: ref #12068 

-> pulsar+http and pulsar+https protocols are added in IsMQScheme function so that MQ related fields will get picked up for these protocols

#### Tests 
1. Unit test
2. Feature test by deploying the image on  k8s environment and creating multiple changefeeds.

-> It should not cause any performance regression or break compatibility. It only addition schema addition for pulsar client creation.

```release-note
Bug fix for  pulsar+http and pulsar+https schema  pulsar protocols to honor MQ fields. 
```

### DIFF
--- a/pkg/sink/sink_type.go
+++ b/pkg/sink/sink_type.go
@@ -81,7 +81,7 @@ const (
 // IsMQScheme returns true if the scheme belong to mq scheme.
 func IsMQScheme(scheme string) bool {
 	return scheme == KafkaScheme || scheme == KafkaSSLScheme ||
-		scheme == PulsarScheme || scheme == PulsarSSLScheme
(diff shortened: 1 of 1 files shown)


wk989898 (approved):

wk989898:
Please run `make fmt` to format codes.

3AceShowHand (approved):

SandeepPadhi:
@wk989898 fixed formatting

PR #12078, opened 2025-02-28
*(dm): retry statement execution on invalid connections in some scenarios

### LINKED ISSUE #11805: TiDB rolling restart causes connection errors that trigger tasks to pause and resume
### What did you do?

Triggered a rolling restart of TiDB, which causes connections from the DM sql driver to TiDB to be half closed.

The client would get connection errors that would be automatically retried after a backoff without causing the DM task to exit.

### What did you see instead?

The sql execution fails with the following un-retryable error
```
	[2024/09/27 21:16:13.544 +00:00] [ERROR] [db.go:206] ["execute statements failed after retry"] [task=dm-task] [unit="binlog replication"] [queries="[INSERT INTO ...]"] [arguments="[...]]"] [error="[code=10006:class=database:scope=not-set:level=high], Message: execute statement failed: begin, RawCause: invalid connection"]
```
This triggers the task to get paused. It will eventually resume, but this causes a backoff and it takes a few minutes for lag to catch up.

### Proposed fix
We believe this is related to the condition described in this [blog post](https://github.blog/engineering/infrastructure/three-bugs-in-the-go-mysql-driver/#fixing-the-issue). An upstream [fix](https://github.com/go-sql-driver/mysql/pull/934) for the go-mysql-driver was merged to health check connections before checking them out of the queue. However, since the DM only checks out connections at the start of worker's execution and does not return them to the pool, this fix does not help us.

Instead, we have been working on a proposed fix to add a retry for this case when an "invalid connection" error occurs on a begin transaction call for a task in safe mode. We believe that this should be safe and not run into issues with partially applied transactions since no DML has been executed if the task fails on `begin` and safe mode helps ensure that DML queries are idempotent.

Additionally, we have to add retries to the resetConn call since that could also happen to check out a stale connection from the pool.

Here is a gist of the proposed fix:  https://gist.github.com/cgtz/f1ead42ae585a3219cc9c381e86c9e50

### Versions of the cluster

```console
# ./dmctl -V
Release Version: v8.1.0-master
Git Commit Hash:
Git Branch:
UTC Build Time: 2024-11-06 18:51:15
Failpoint Build: false
```
```console
```
```console
8.1.1
```
```console
(leave TiUP or manually here)
Using tidb-operator
```
```console
>
>
```

```console
```

cgtz:
This PR tries to address a case where, when a TiDB machine is restarted, stale connections will live on in the connection pool. When DM attempts to use these connections, it will receive an "Invalid Connection" error, which will cause the task to pause and increase replication lag while waiting to reschedule it.

Issue Number: close #11805 

When we get an invalid connection error on a begin statement, we can safely retry after resetting the connection. The reset connection call can also fail due to similar reasons or due to a load balancer layer refusing the connection, so we can also do a retry with backoff here.

#### Tests 
Tested on our organization's deployment of DM.
 - Unit test
 - Integration test
 - No code

### Release note 

```release-note
Retry when the DM syncer receives an invalid connection error while at the beginning of a transaction to avoid full task pauses that increase replication lag
```


dm/pkg/retry/errors.go
```diff
@@ -100,3 +101,24 @@ func IsUnretryableConnectionError(err error) bool {
 	// If the last write isn't idempotent, retry it may cause problems.
 	return errors.Cause(err) == dmysql.ErrInvalidConn
 }
+
+// IsRetryableConnectionErrorOnBegin checks whether an invalid connection error occured on a
+// transaction begin statement, when there are no operations that may need to be rolled back.
+func IsRetryableConnectionErrorOnBegin(err error) bool {
+	if !strings.Contains(strings.ToLower(err.Error()), "execute statement failed: begin") {
```
lance6716, line 108: Seems it's generated from here

https://github.com/pingcap/tiflow/blob/e23fa78655fa5122430dc92fed1be6f0a827c725/dm/pkg/conn/baseconn.go#L168

Maybe you can create a special error variable called `ErrDBExecuteFailedBegin` and use `terror.ErrDBExecuteFailedBegin.Equal(err)` to do a more structured check. The current way of check substring in error message will fail after another developer accidentally change the message.

dm/syncer/dbconn/db.go
```diff
@@ -202,8 +203,11 @@ func (conn *DBConn) ExecuteSQLWithIgnore(
 		})
 	if err != nil {
 		tctx.L().ErrorFilterContextCanceled("execute statements failed after retry",
+			zap.Int("numRetries", numRetries),
+			zap.Float64("costSeconds", time.Since(startTime).Seconds()),
 			zap.String("queries", utils.TruncateInterface(queries, -1)),
 			log.ZapRedactString("arguments", utils.TruncateInterface(args, -1)),
+			zap.Error(err),
 			log.ShortError(err))
```
lance6716, line 211: `zap.Error(err)` and `log.ShortError(err)` is a bit duplicated. Do you want to remove `log.ShortError(err)`?

dm/syncer/dbconn/db.go
```diff
@@ -241,16 +245,41 @@ func (conn *DBConn) ExecuteSQLAutoSplit(
 
 func (conn *DBConn) retryableFn(tctx *tcontext.Context, queries, args any) func(int, error) bool {
 	return func(retryTime int, err error) bool {
-		if retry.IsConnectionError(err) {
-			err = conn.ResetConn(tctx)
+		if retry.IsConnectionError(err) || (conn.cfg.SafeMode && retry.IsRetryableConnectionErrorOnBegin(err)) {
```
lance6716, line 248: As discuss in the linked issue https://github.com/pingcap/tiflow/issues/11805#issuecomment-2505470457, I think it's

```suggestion
		if retry.IsConnectionError(err) || (conn.cfg.SafeMode || retry.IsRetryableConnectionErrorOnBegin(err)) {
```

dm/pkg/retry/errors.go
cgtz: Thanks, this is a good idea!

dm/syncer/dbconn/db.go
cgtz: Good catch, I changed this to be a `(safemode || begin) && invalidconnection` condition.

lance6716:
CI says an integration test failed after the change, the script of that integration test is https://github.com/pingcap/tiflow/blob/5c3ed3247992e827ddcf80fcfcb4b4076c348303/dm/tests/many_tables/run.sh#L156-L159

Can you find the root cause and fix it? If not, we can help you.

cgtz:
Hi @lance6716 I think this is related to changing the retry behavior and making the DM less likely to restart. 
```
	# now worker will process some binlog events, save table checkpoint and meet downstream error
	echo "start incremental_data_2"
	incremental_data_2
	echo "finish incremental_data_2"
	sleep 30

	resume_num=$(grep 'unit process error' $WORK_DIR/worker1/log/dm-worker.log | wc -l)
	echo "resume_num: $resume_num"
	# because we check auto resume every 5 seconds...
	[ $resume_num -ge 4 ]
```
resume_num is 2 instead of >= 4.

Do you have a command for running a single integration test locally? It seems like there are environment variables/dependencies that have to be set up.

lance6716:
You can refer to https://github.com/pingcap/tiflow/blob/master/dm/step_by_step_contribute.md#step-8-add-integration-test to setup 2 MySQL server and run `make dm_integration_test many_tables`. (I haven't run the integration tests of this project for a long time 😂 if you meet some error, you can ask @GMHDBJD )

cgtz:
@lance6716 I was able to debug using the log artifacts from the ci build. The integration tests should pass now after adjusting the wait times to account for 30 seconds of retrying on connection refused.

lance6716 (approved):

cgtz:
Hi @lance6716, thank you for reviewing and approving the PR. I requested a cherry pick to 8.5 but it has merge conflicts. Could you help resolve them? I don't think I can write to the branch in https://github.com/pingcap/tiflow/pull/12160

PR #12106, opened 2025-03-18
*(ticdc): support ssl when fetching downstream cluster id

### LINKED ISSUE #12100: cdc: tidb sink with tls cannot connect with scheme 'mysql://'
### What did you do?

open changefeed to tls tidb sink with schema  'mysql://'

no err

### What did you see instead?

cannot create changefeed
Image
use 'tidb://' can work

Image

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
master 
or
v9.0-beta.1
```

wlwilliamx:

Issue Number: close #12100

This PR fixes an issue in TiCDC where fetching the downstream TiDB cluster_id fails when SSL is enabled.

#### Background

TiCDC introduces a feature to prevent configuring the upstream and downstream as the same TiDB cluster. This is achieved by executing an SQL query to retrieve the cluster_id from the downstream database and comparing it with the upstream.

However, the implementation did not consider SSL-enabled connections, which caused connection failures when querying the cluster_id if the downstream TiDB required SSL. And added a integration test about this scenario.

#### Fix
* Updated the SQL query logic to correctly support SSL when retrieving cluster_id from the downstream TiDB.
* Ensured that SSL configurations are properly handled during the connection setup.

This fix ensures that TiCDC works correctly in environments where SSL is enforced for secure database connections.

#### Tests 

 - Unit test
 - Integration test

Verified that TiCDC can successfully fetch cluster_id when SSL is enabled and `require_secure_transport=ON`.
Confirmed that non-SSL connections remain unaffected.

No.

No.

### Release note 

```release-note
None
```


pkg/check/cluster.go
```diff
@@ -85,6 +103,8 @@ func getClusterIDBySinkURI(ctx context.Context, sinkURI string) (uint64, bool, e
 	err = row.Scan(&clusterIDStr)
 	if err != nil {
 		// If the cluster ID is not set, it is a legacy TiDB cluster, these should be compatible with it.
+		// So we return 0 and false.
+		log.Info("the downstream is TiDB, but the cluster ID is not set, it is a legacy TiDB cluster")
```
lidezhu, line 107:
```suggestion
		log.Info("the downstream is TiDB, but the cluster ID is not set, it is a legacy TiDB cluster", zap.Error(err))
```

lidezhu (approved):
LGTM with minor comments

PR #12132, opened 2025-04-11
sink(ticdc): Treat column and index names as case insensitive

### LINKED ISSUE #12103: make the case-sensitive works on the whole dispatchers configuration

At the moment, there is one configuration named as  [case-sensitive](https://docs.pingcap.com/tidb/stable/ticdc-changefeed-config/#case-sensitive), which controls the database name and table name specified in the dispatcher part.

Let's take the following as an example, the `columns` specified the columns used to calculate the partition. 

It may be hard for the user to specify columns with case if there are a lot of tables. For example, table's may have different columns name to indicate the same thing, such as `ID`, `Id`, `id`, etc.
```
[sink]
dispatchers = [
    {matcher = ['test2.*'], partition = "columns", columns = ["id", "a"]},
]
```

This should also takes effect on the column-selector configurations
```
[sink]
column-selectors = [
    {matcher = ['test.t1'], columns = ['a', 'b']},
    {matcher = ['test.*'], columns = ["*", "!b"]},
    {matcher = ['test1.t1'], columns = ['column*', '!column1']},
    {matcher = ['test3.t'], columns = ["column?", "!column1"]},
]
```

the `case-sensitive` field, should also take affect on the columns configurations.

_No response_

After this feature is implemented, should also update the document to mention this.

wk989898:

Issue Number: close #12103

According to [MySQL doc](https://dev.mysql.com/doc/refman/5.7/en/identifier-case-sensitivity.html), column and index names are not case-sensitive on any platform. 

Make cdc compatible with it.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
Treat column and index names as case-insensitive in dispatchers configuration.
```

### DIFF
--- a/cdc/model/schema_storage.go
+++ b/cdc/model/schema_storage.go
@@ -15,6 +15,7 @@ package model
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/pingcap/log"
 	"github.com/pingcap/tidb/pkg/meta/model"
@@ -435,7 +436,7 @@ func (ti *TableInfo) Clone() *TableInfo {
 // GetIndex return the corresponding index by the given name.
 func (ti *TableInfo) GetIndex(name string) *model.IndexInfo {
 	for _, index := range ti.Indices {
-		if index != nil && index.Name.O == name {
+		if index != nil && index.Name.L == strings.ToLower(name) {
 			return index
 		}
 	}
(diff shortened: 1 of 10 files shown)


3AceShowHand (approved):

3AceShowHand:
please add more unit test to cover user scenario.

asddongmen (approved):

PR #12143, opened 2025-04-17
[DNM] kafka: support GSSAPI auth with custom SPN

kennytm:

Issue Number: close #xxx

Allow setting a custom SPN (service principal name) of the Kafka servers when using GSSAPI authentication, in case the Kerboros server only recognizes `kafka/<spn>` but not `kafka/<ip>`.

#### Tests 

 - Unit test
 - Integration test
 - No code

No

Yes

### Release note 

```release-note
Support customizing the SPN for GSSAPI
```


kennytm:
Confirmed this works; will submit a feature request later.

Issue #12141, opened 2025-04-16
A flaky test in sync_diff_inspector


wk989898:
### Which jobs are flaking?

Unit test in sync_diff_inspector

### Which test(s) are flaking?

TestMysqlRouter

### Jenkins logs or GitHub Actions link

- Unexpected error https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflow%2Fghpr_verify/detail/ghpr_verify/3749/pipeline

- DATA RACE https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflow%2Fghpr_verify/detail/ghpr_verify/3754/pipeline
```log
=== FAIL: sync_diff_inspector/source TestMysqlRouter (1.01s)

[2025/04/16 16:41:56.031 +08:00] [INFO] [mysql_shard.go:392] ["will increase connection configurations for DB of instance"] ["connection limit"=5]

[2025/04/16 16:41:56.032 +08:00] [INFO] [source.go:425] ["table match check finished"]

[2025/04/16 16:41:56.033 +08:00] [INFO] [random.go:119] ["get chunk size for table"] ["chunk size"=50000] [db=source_test] [table=test2]

[2025/04/16 16:41:56.033 +08:00] [INFO] [random.go:125] ["split range by random"] ["row count"=0] ["split chunk num"=0]

[2025/04/16 16:41:56.033 +08:00] [INFO] [chunks_iter.go:97] ["Stop do produce chunks by context done"]

    source_test.go:532: 

        	Error Trace:	/home/jenkins/agent/workspace/pingcap/tiflow/ghpr_verify/tiflow/sync_diff_inspector/source/source_test.go:532

        	Error:      	Received unexpected error:

        	            	Query: could not match actual sql: "SELECT /*!40001 SQL_NO_CACHE */ `a`, `b`, round(`c`, 5-floor(log10(abs(`c`)))) as `c` FROM `source_test_t`.`test_t` WHERE ((((`a` > ?) OR (`a` = ? AND `b` > ?)) AND ((`a` < ?) OR (`a` = ? AND `b` <= ?))) AND ()) ORDER BY `a`,`b`" with expected regexp "SELECT COUNT.*"

        	            	github.com/pingcap/errors.AddStack

        	            		/go/pkg/mod/github.com/pingcap/errors@v0.11.5-0.20240318064555-6bd07397691f/errors.go:178

        	            	github.com/pingcap/errors.Trace

[...]

PR #12159, opened 2025-04-24
sync_diff_inspector: fix flaky test TestMysqlRouter

joechenrh:

Issue Number: close https://github.com/pingcap/tiflow/issues/12141

There are two problems solved in this PR

The first is unexpected error(see https://github.com/pingcap/tiflow/issues/12141#issuecomment-2826624187). To fix this, some of the mock queries are replaced with failpoint to make tests more stable.

The second is data race problem(see https://github.com/pingcap/tiflow/issues/12141#issuecomment-2826666728). To fix this, `WaitGroup.Add` and `WaitGroup.Wait` operations are moved into one goroutine.

#### Tests 

 - Unit test
 - Integration test
   - Test should be more stable after this PR.
 - No code

### Release note 

```release-note

```

### DIFF
--- a/sync_diff_inspector/splitter/random.go
+++ b/sync_diff_inspector/splitter/random.go
@@ -41,7 +41,7 @@ type RandomIterator struct {
 	dbConn *sql.DB
 }
 
-// Exported for test
+// MockRowCount is the mock row count used in unit tests
 var MockRowCount int64
 
 // a wrapper for get row count to integrate failpoint.
--- a/sync_diff_inspector/splitter/random.go
+++ b/sync_diff_inspector/splitter/random.go
@@ -41,13 +41,12 @@ type RandomIterator struct {
 	dbConn *sql.DB
 }
 
-// MockRowCount is the mock row count used in unit tests
-var MockRowCount int64
-
 // a wrapper for get row count to integrate failpoint.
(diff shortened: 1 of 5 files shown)


GMHDBJD (approved):

sync_diff_inspector/splitter/random.go
```diff
@@ -41,6 +41,17 @@ type RandomIterator struct {
 	dbConn *sql.DB
 }
 
+// MockRowCount is the mock row count used in unit tests
+var MockRowCount int64
+
+// a wrapper for get row count to integrate failpoint.
+func getRowCount(ctx context.Context, db dbutil.QueryExecutor, schemaName string, tableName string, where string, args []any) (int64, error) {
+	failpoint.Inject("getRowCount", func() {
```
lance6716, line 49: you can pass the injected value as the argument of failpoint. No need to use `MockRowCount`. You can find this usage in many failpoint inject in tidb repo.

sync_diff_inspector/splitter/random.go
lance6716, line 1: the file permission is changed. I remember `failpoint-enable` then `git restore` will cause this problem. No need to use 755 for code files

lance6716 (commented):
rest lgtm

PR #12170, opened 2025-05-15
*(dm): retry statement execution on invalid connections in some scenarios (#12078)

### LINKED ISSUE #11805: TiDB rolling restart causes connection errors that trigger tasks to pause and resume

cgtz:
This is a manual cherry-pick of #12078
 
This PR tries to address a case where, when a TiDB machine is restarted, stale connections will live on in the connection pool. When DM attempts to use these connections, it will receive an "Invalid Connection" error, which will cause the task to pause and increase replication lag while waiting to reschedule it.

Issue Number: close #11805

When we get an invalid connection error on a begin statement, we can safely retry after resetting the connection. The reset connection call can also fail due to similar reasons or due to a load balancer layer refusing the connection, so we can also do a retry with backoff here.

#### Tests
Tested on our organization's deployment of DM.

* Unit test
* Integration test
* Manual test (add detailed scripts or steps below)
* No code

### Release note
```
Retry when the DM syncer receives an invalid connection error while at the beginning of a transaction to avoid full task pauses that increase replication lag
```


cgtz:
@lance6716 @GMHDBJD , here is the new cherry pick PR for DM invalid connection retries. Please take a look.

PR #12173, opened 2025-05-16
sarama: add keep-alive mechanism for sarama connections

### LINKED ISSUE #12171: cdc changefeed meet error: broken pipe
### What did you do?

The CDC normally runs and synchronizes data to the downstream Kafka. Sometimes, there may be no data to synchronize from the upstream for more than 10 minutes. Then, during the next synchronization, the CDC receives an error from Sarama:

`kafka async send message failed: kafka: Failed to produce message to topic xxx: write tcp IP:PORT->IP:PORT: write: broken pipe.`

Then the changefeed lag will increase.

The changefeed lag should be normal.

### What did you see instead?

CDC receives broken pipe error, and changefeed lag increases.

### Versions of the cluster

```console
v7.5.6
```

```console
v7.5.6
```

```console
v7.5.6
```

wlwilliamx:

Issue Number: close #12171

This PR addresses the issue of frequent broken pipe errors in CDC when connecting to Kafka. The user reported that multiple clusters frequently triggered CDC alerts (cdc changefeed meet error) with the error message `write tcp ...: write: broken pipe`, occurring on average once every dozens of minutes. Although CDC appeared to function normally — with business operations unaffected and checkpoint lag remaining within acceptable limits — the alerts persisted.

The reason is that CDC had recently removed retries in the Sarama Kafka client (used by CDC) — reducing the retry count from 3 to 0 in #11870. This change was originally made to avoid potential message reordering issues. However, disabling retries caused CDC to surface connection-related errors directly, such as broken pipe, leading to changefeed failures instead of silent retries.

This PR improves the reliability of Kafka connections in TiCDC by introducing a **keep-alive mechanism** for sarama that periodically sends heartbeats to all known Kafka brokers. This is a workaround for a known issue where Kafka broker silently closes idle connections, especially when sarama retry is disabled. The keep-alive interval will be a third of(for fault tolerance) `connections.max.idle.ms` in Kafka broker configs. In CDC, three types of Kafka connections are established, which are used for admin requests, DDL and watermark, and DML respectively.

* For the Kafka connection used for admin requests, I added periodic heartbeat sending to all Kafka brokers in the `kafkaTopicManager.backgroundRefreshMeta` goroutine to refresh the connections with all brokers.

* For the Kafka connection used for DDL and watermark, I added the heartbeat sending operation in the `DDLSink.WriteCheckpointTs` function, which will be invoked periodically.

* For the Kafka connection used for DML, I added periodic heartbeat sending to all Kafka brokers in the `saramaAsyncProducer.AsyncRunCallback`.

#### UT
The units tests cover the entire lifecycle of the keep-alive logic. We verify that the keep-alive interval is correctly calculated from the broker's settings by testing the AdjustOptions function. For the DML sink, we ensure the kafkaTopicManager's background routine sends heartbeats periodically. For the DDL sink, we confirm that a call to WriteCheckpointTs reliably triggers a heartbeat.

At the producer level, the tests validate the specific logic within both the synchronous and asynchronous Sarama producers. This includes testing the throttling mechanism in the sync producer to avoid sending heartbeats too frequently, and the background ticker in the async producer that drives the periodic calls.

To achieve this test coverage, a couple of specific strategies were used. To test the DDL sink's heartbeat trigger in isolation, a mock encoder is introduced that returns an error immediately after the heartbeat logic, simplifying the test by avoiding the need to set up a full message production pipeline. Furthermore, to test the producer-level heartbeats, we mock the sarama.Client interface itself. This approach allows us to verify that the heartbeat logic is being invoked, working around a limitation in the Sarama library that prevents direct mocking of the *sarama.Broker struct's methods.

With these tests in place, the keep-alive feature is now well-covered.

#### Tests 

 - Unit test
 - Integration test
 - Manual test
 - Test case PR: https://github.com/PingCAP-QE/test-infra/pull/4094
 - Test plan PR: https://github.com/PingCAP-QE/test-plan/pull/3765, will be run in daily and future patch release.
 - Kafka compatibility test: 0.11.0-0-r0, 0.11.0-1-r0, 1.0.0-r0, 1.0.1-r0, 1.1.0, 1.1.1, 2.0.0, 2.0.1, 2.1.0, 2.2.0, 2.3.0, 2.4.0, 2.5.0, 2.6.0, 2.7.0, 2.8.0, 3.0.0, 3.1.0, 3.2.0. 3.4.0, 3.5.0, 3.6.0, 3.7.0.

A cluster was launched for testing, synchronizing upstream data to the downstream Kafka. The `connections.max.idle.ms` setting on the Kafka broker was configured to 10 seconds. Then, a load was added on the upstream side, randomly writing some data and pausing for a while. This load was run continuously, and it was confirmed that there were no issues.

No.

No.

### Release note 

```release-note
Improve Kafka connection reliability by introducing a keep-alive mechanism for sarama. This helps prevent unexpected timeouts caused by idle connections being silently closed by Kafka.
```

### DIFF
--- a/cdc/sink/ddlsink/mq/mq_ddl_sink.go
+++ b/cdc/sink/ddlsink/mq/mq_ddl_sink.go
@@ -72,6 +72,8 @@ type DDLSink struct {
 	statistics *metrics.Statistics
 	// admin is used to query kafka cluster information.
 	admin kafka.ClusterAdminClient
+	// connRefresherForDDL is used to refresh the connection for DDL events.
+	connRefresherForDDL kafka.SyncProducer
 }
 
 func newDDLSink(
@@ -82,6 +84,7 @@ func newDDLSink(
 	eventRouter *dispatcher.EventRouter,
 	encoder codec.RowEventEncoder,
 	protocol config.Protocol,
+	connRefresherForDDL kafka.SyncProducer,
 ) *DDLSink {
 	return &DDLSink{
 		id:           changefeedID,
@@ -137,9 +140,19 @@ func (k *DDLSink) WriteDDLEvent(ctx context.Context, ddl *model.DDLEvent) error 
 }
 
 // WriteCheckpointTs sends the checkpoint ts to the MQ system.
+// This function will be called at least once per second.
 func (k *DDLSink) WriteCheckpointTs(ctx context.Context,
 	ts uint64, tables []*model.TableInfo,
 ) error {
+	// This operation is used to keep the kafka connection alive.
+	// For more details, see https://github.com/pingcap/tiflow/pull/12173
+	if k.connRefresherForDDL != nil {
+		// The implementation is saramaSyncProducer.HeartbeatBrokers. And
+		// there is a keepConnAliveInterval in the saramaSyncProducer, so
+		// we don't need to worry about the heartbeat is too frequent.
+		k.connRefresherForDDL.HeartbeatBrokers()
+	}
+
 	var (
 		err          error
 		partitionNum int32
--- a/cdc/sink/ddlsink/mq/mq_ddl_sink.go
+++ b/cdc/sink/ddlsink/mq/mq_ddl_sink.go
@@ -87,14 +87,15 @@ func newDDLSink(
 	connRefresherForDDL kafka.SyncProducer,
 ) *DDLSink {
 	return &DDLSink{
-		id:           changefeedID,
-		protocol:     protocol,
-		eventRouter:  eventRouter,
-		topicManager: topicManager,
-		encoder:      encoder,
-		producer:     producer,
-		statistics:   metrics.NewStatistics(changefeedID, sink.RowSink),
-		admin:        adminClient,
+		id:                  changefeedID,
+		protocol:            protocol,
+		eventRouter:         eventRouter,
+		topicManager:        topicManager,
+		encoder:             encoder,
+		producer:            producer,
+		statistics:          metrics.NewStatistics(changefeedID, sink.RowSink),
+		admin:               adminClient,
+		connRefresherForDDL: connRefresherForDDL,
 	}
 }
 
--- a/cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
+++ b/cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
@@ -25,6 +25,12 @@ import (
 	"github.com/pingcap/tiflow/pkg/config"
 	"github.com/pingcap/tiflow/pkg/sink/kafka"
 	"github.com/stretchr/testify/require"
+
+	"sync"
+
+	"github.com/pingcap/tiflow/pkg/errors"
+	"github.com/pingcap/tiflow/pkg/sink/codec"
+	"github.com/pingcap/tiflow/pkg/sink/codec/common"
 )
 
 func TestNewKafkaDDLSinkFailed(t *testing.T) {
@@ -291,4 +297,95 @@ func TestGetDLLDispatchRuleByProtocol(t *testing.T) {
 	require.Equal(t, PartitionAll, getDDLDispatchRule(config.ProtocolCraft))
 	require.Equal(t, PartitionAll, getDDLDispatchRule(config.ProtocolSimple))
 	require.Equal(t, PartitionAll, getDDLDispatchRule(config.ProtocolDebezium))
+}
+
+// mockSyncProducer is used to count the calls to HeartbeatBrokers.
+type mockSyncProducer struct {
+	kafka.MockSaramaSyncProducer
+	heartbeatCount int
+	mu             sync.Mutex
+}
+
+func (m *mockSyncProducer) HeartbeatBrokers() {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.heartbeatCount++
+}
+
+func (m *mockSyncProducer) GetHeartbeatCount() int {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return m.heartbeatCount
+}
+
+// mockEncoder is a mock implementation of codec.RowEventEncoder.
+// It is used to prevent the test from needing to set up a real, complex encoder.
+type mockEncoder struct{}
+
+// This line ensures at compile time that mockEncoder correctly implements the interface.
+var _ codec.RowEventEncoder = (*mockEncoder)(nil)
+
+// A predefined error that our mock encoder will return.
+var errMockEncoder = errors.New("mock encoder error")
+
+// EncodeCheckpointEvent returns a specific error to halt the execution
+// of the function under test right after the heartbeat logic.
+func (m *mockEncoder) EncodeCheckpointEvent(ts uint64) (*common.Message, error) {
+	return nil, errMockEncoder
+}
+
+// The following methods are part of the RowEventEncoder interface.
+// They can be left with a minimal implementation as they are not called
+// in the tested code path.
+func (m *mockEncoder) AppendRowChangedEvent(
+	_ context.Context, _ string, _ *model.RowChangedEvent, _ func(),
+) error {
+	return nil
+}
+func (m *mockEncoder) Build() []*common.Message {
+	return nil
+}
+func (m *mockEncoder) EncodeDDLEvent(event *model.DDLEvent) (*common.Message, error) {
+	return nil, nil
+}
+func (m *mockEncoder) SetMaxMessageBytes(bytes int) {}
+
+func TestDDLSinkHeartbeat(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	changefeedID := model.DefaultChangeFeedID("test-ddl-sink")
+	// Protocol is needed by newDDLSink but not used in the tested path.
+	proto := config.ProtocolOpen
+	encoder := &mockEncoder{}
+
+	// Case 1: DDL Sink with a connection refresher.
+	t.Run("DDLSinkWithConnectionRefresher", func(t *testing.T) {
+		t.Parallel()
+		producer := &mockSyncProducer{}
+		// Other dependencies for newDDLSink can be nil as they are not used
+		// before our mock encoder returns an error.
+		ddlSink := newDDLSink(changefeedID, nil, nil, nil, nil, encoder, proto, producer)
+
+		require.Equal(t, 0, producer.GetHeartbeatCount())
+
+		// WriteCheckpointTs should first call HeartbeatBrokers, then fail at the encoder.
+		err := ddlSink.WriteCheckpointTs(ctx, 12345, nil)
+
+		// Assert that the heartbeat was called exactly once.
+		require.Equal(t, 1, producer.GetHeartbeatCount())
+		// Assert that the function failed with the mock encoder's specific error.
+		require.ErrorIs(t, err, errMockEncoder)
+	})
+
+	// Case 2: DDLSink with a nil connection refresher (e.g., for Pulsar).
+	t.Run("DDLSinkWithNilConnectionRefresher", func(t *testing.T) {
+		t.Parallel()
+		// Create the sink with a nil refresher.
+		ddlSinkNilRefresher := newDDLSink(changefeedID, nil, nil, nil, nil, encoder, proto, nil)
+
+		// The call should not panic and should fail with the mock encoder's error.
+		err := ddlSinkNilRefresher.WriteCheckpointTs(ctx, 12347, nil)
+		require.ErrorIs(t, err, errMockEncoder)
+	})
 }
--- a/cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
+++ b/cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
@@ -17,20 +17,18 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"sync"
 	"testing"
 
 	mm "github.com/pingcap/tidb/pkg/meta/model"
 	"github.com/pingcap/tiflow/cdc/model"
 	"github.com/pingcap/tiflow/cdc/sink/ddlsink/mq/ddlproducer"
 	"github.com/pingcap/tiflow/pkg/config"
-	"github.com/pingcap/tiflow/pkg/sink/kafka"
-	"github.com/stretchr/testify/require"
-
-	"sync"
-
 	"github.com/pingcap/tiflow/pkg/errors"
 	"github.com/pingcap/tiflow/pkg/sink/codec"
 	"github.com/pingcap/tiflow/pkg/sink/codec/common"
+	"github.com/pingcap/tiflow/pkg/sink/kafka"
+	"github.com/stretchr/testify/require"
 )
 
 func TestNewKafkaDDLSinkFailed(t *testing.T) {
--- a/cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
+++ b/cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
@@ -340,9 +340,11 @@ func (m *mockEncoder) AppendRowChangedEvent(
 ) error {
 	return nil
 }
+
 func (m *mockEncoder) Build() []*common.Message {
 	return nil
 }
+
 func (m *mockEncoder) EncodeDDLEvent(event *model.DDLEvent) (*common.Message, error) {
 	return nil, nil
 }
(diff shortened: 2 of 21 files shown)


pkg/sink/kafka/sarama.go
```diff
@@ -29,8 +29,9 @@ import (
 )
 
 var (
-	defaultKafkaVersion = sarama.V2_0_0_0
-	maxKafkaVersion     = sarama.V2_8_0_0
+	defaultKafkaVersion   = sarama.V2_0_0_0
+	maxKafkaVersion       = sarama.V2_8_0_0
+	keepConnAliveInterval = 5 * time.Minute
```
wk989898: It is better to use it locally.

wlwilliamx: Yes, but this is the better way to share a interval to all AdminClient, SyncProducer and AsyncProducer.

wk989898: There is no need to share the variable, and it should be updated when the user alters `connections.max.idle.ms` in Kafka?

pkg/sink/kafka/admin.go
```diff
@@ -31,6 +31,7 @@ type saramaAdminClient struct {
 
 	client sarama.Client
 	admin  sarama.ClusterAdmin
+	done   chan struct{}
```
3AceShowHand: This looks useless.

pkg/sink/kafka/factory.go
```diff
@@ -212,6 +216,8 @@ func (p *saramaAsyncProducer) AsyncRunCallback(
 					callback()
 				}
 			}
+		case <-ticker.C:
+			hearbeatBroker(p.client)
```
3AceShowHand: typo, heartbeat

pkg/sink/kafka/sarama.go
```diff
@@ -255,3 +256,40 @@ func getKafkaVersionFromBroker(config *sarama.Config, requestVersion int16, addr
 	}
 	return KafkaVersion, nil
 }
+
+// keepConnAlive is used to keep the connection alive.
+func keepConnAlive(client sarama.Client, done chan struct{}) {
```
3AceShowHand: This method looks good to me, but should be controlled by the context, instead of the done channel.

pkg/sink/kafka/sarama_factory.go
```diff
@@ -75,11 +76,25 @@ func (f *saramaFactory) AdminClient(ctx context.Context) (ClusterAdminClient, er
...
+		return nil, errors.Trace(err)
+	}
+	idleMsInt, err := strconv.Atoi(idleMs)
+	if err != nil || idleMsInt <= 0 {
+		log.Warn("invalid broker config",
+			zap.String("configName", BrokerConnectionsMaxIdleMsConfigName), zap.String("configValue", idleMs))
+		return nil, errors.Trace(err)
+	}
+	keepConnAliveInterval = time.Duration(idleMsInt/2) * time.Millisecond
+	go keepConnAlive(client, a.done)
```
3AceShowHand: Do not spawn goroutine here.

AdminClient is called by the ddl sink and dml sink.

it's used to initialize the topicManager, this method refresh metadata each 10 minutes, add one branch to the backgroundRefreshMeta method to send heartbeat if necessary.

3AceShowHand: On the DML sink side, add one branch to the metrics collector is also a choice.

pkg/sink/kafka/sarama.go
3AceShowHand: This should be set to the `option`, and adjust in the AdjustOption method.

pkg/sink/kafka/sarama_factory.go
```diff
@@ -90,14 +105,22 @@ func (f *saramaFactory) SyncProducer(ctx context.Context) (SyncProducer, error)
...
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	sp := saramaSyncProducer{
 		id:       f.changefeedID,
 		producer: p,
-	}, nil
+		done:     make(chan struct{}),
+	}
+	go keepConnAlive(client, sp.done)
```
3AceShowHand: ditto, do not spawn goroutine here.

DDLSink.WriteCheckpointTs method, is called periodically, you can call the heartbeat method here

3AceShowHand:
If the encoding protocol would send checkpoint to the kafka server, there is no need to send heartbeat.

cdc/sink/ddlsink/mq/mq_ddl_sink.go
```diff
@@ -72,6 +72,8 @@ type DDLSink struct {
 	statistics *metrics.Statistics
 	// admin is used to query kafka cluster information.
 	admin kafka.ClusterAdminClient
+	// connRefresherForDDL is used to refresh the connection for DDL events.
+	connRefresherForDDL kafka.SyncProducer
```
wk989898, line 76: I think it is a redundant parameter, you can modify the `DDLProducer` interface to add `Heartbeat` method.

pkg/sink/kafka/factory.go
```diff
@@ -115,6 +121,27 @@ func (p *saramaSyncProducer) SendMessages(_ context.Context, topic string, parti
...
+
+	// We don't care about the response and error here, even the connection
+	// is unestablished, we just need to keep the connection alive WHEN it's established.
+	// The connection will be established when a producer send messages.
+	// This is a workaround for the issue that sarama doesn't keep the connection alive
+	// when the connection is idle for a long time and we have disabled the retry in sarama.
+	brokers := p.client.Brokers()
+	for _, b := range brokers {
+		_, _ = b.Heartbeat(&sarama.HeartbeatRequest{})
+	}
```
wk989898: It is better to use a function to replace it.

cdc/sink/ddlsink/mq/mq_ddl_sink.go
3AceShowHand: k.producer.Heartbeat()

cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
```diff
@@ -292,3 +296,96 @@ func TestGetDLLDispatchRuleByProtocol(t *testing.T) {
...
+func (m *mockEncoder) SetMaxMessageBytes(bytes int) {}
+
+func TestDDLSinkHeartbeat(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	changefeedID := model.DefaultChangeFeedID("test-ddl-sink")
+	// Protocol is needed by newDDLSink but not used in the tested path.
+	proto := config.ProtocolOpen
+	encoder := &mockEncoder{}
```
3AceShowHand, line 360: You set the protocol as the ProtocolOpen but not used, this mockEncoder is redundant, just new a open protocol encoder is much better.

cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
```diff
@@ -292,3 +296,96 @@ func TestGetDLLDispatchRuleByProtocol(t *testing.T) {
...
+
+		require.Equal(t, 0, producer.GetHeartbeatCount())
+
+		// WriteCheckpointTs should first call HeartbeatBrokers, then fail at the encoder.
+		err := ddlSink.WriteCheckpointTs(ctx, 12345, nil)
+
+		// Assert that the heartbeat was called exactly once.
+		require.Equal(t, 1, producer.GetHeartbeatCount())
+		// Assert that the function failed with the mock encoder's specific error.
+		require.ErrorIs(t, err, errMockEncoder)
```
3AceShowHand, line 378: Why test this, just make sure the heartbeat is called should be fine.

cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
```diff
@@ -292,3 +296,96 @@ func TestGetDLLDispatchRuleByProtocol(t *testing.T) {
...
+	})
+
+	// Case 2: DDLSink with a nil connection refresher (e.g., for Pulsar).
+	t.Run("DDLSinkWithNilConnectionRefresher", func(t *testing.T) {
+		t.Parallel()
+		// Create the sink with a nil refresher.
+		ddlSinkNilRefresher := newDDLSink(changefeedID, nil, nil, nil, nil, encoder, proto, nil)
+
+		// The call should not panic and should fail with the mock encoder's error.
```
3AceShowHand, line 387: Just make sure it won't panic should be fine.

cdc/sink/dmlsink/mq/manager/kafka_manager.go
```diff
@@ -61,13 +62,15 @@ func NewKafkaTopicManager(
...
-		changefeedID:      changefeedID,
-		admin:             admin,
-		cfg:               cfg,
-		metaRefreshTicker: time.NewTicker(metaRefreshInterval),
+		defaultTopic:        defaultTopic,
+		changefeedID:        changefeedID,
+		admin:               admin,
+		cfg:                 cfg,
+		metaRefreshTicker:   time.NewTicker(metaRefreshInterval),
+		keepConnAliveTicker: time.NewTicker(keepConnAliveInterval),
```
3AceShowHand, line 73: close the ticker when when exits the loop.

cdc/sink/dmlsink/mq/manager/kafka_manager_test.go
```diff
@@ -15,13 +15,19 @@ package manager
...
 	"testing"
+	"time"
 
 	"github.com/pingcap/tiflow/cdc/model"
 	"github.com/pingcap/tiflow/pkg/sink/kafka"
 	"github.com/stretchr/testify/require"
 )
 
+// A fake interval to keep the connection alive in kafka topic manager.
+const FOREVER = time.Duration(math.MaxInt64)
```
3AceShowHand, line 29: What is the test purpose here.

cdc/sink/util/helper_test.go
```diff
@@ -15,14 +15,19 @@ package util
...
 	"testing"
+	"time"
 
 	"github.com/pingcap/tiflow/cdc/model"
 	"github.com/pingcap/tiflow/pkg/sink/kafka"
 	"github.com/stretchr/testify/require"
 )
 
+// A fake interval to keep the connection alive in kafka topic manager.
+const FOREVER = time.Duration(math.MaxInt64)
```
3AceShowHand, line 29: Ditto.

pkg/sink/kafka/options.go
```diff
@@ -588,6 +594,20 @@ func AdjustOptions(
 		}
 	}
 
+	// adjust keepConnAliveInterval by `connections.max.idle.ms` broker config.
+	idleMs, err := admin.GetBrokerConfig(ctx, BrokerConnectionsMaxIdleMsConfigName)
+	if err != nil {
```
3AceShowHand, line 599: If cannot fetch the configuration, simply use the default configuration, print a log to indicate the error.

pkg/sink/kafka/sarama_factory.go
```diff
@@ -90,14 +92,23 @@ func (f *saramaFactory) SyncProducer(ctx context.Context) (SyncProducer, error)
 		return nil, err
 	}
 	config.MetricRegistry = f.registry
-	p, err := sarama.NewSyncProducer(f.option.BrokerEndpoints, config)
+
+	client, err := sarama.NewClient(f.option.BrokerEndpoints, config)
```
3AceShowHand, line 95: this client should also be closed.

pkg/sink/kafka/sarama_factory.go
```diff
@@ -90,14 +92,23 @@ func (f *saramaFactory) SyncProducer(ctx context.Context) (SyncProducer, error)
...
+	p, err := sarama.NewSyncProducerFromClient(client)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	sp := saramaSyncProducer{
+		id:                    f.changefeedID,
+		producer:              p,
+		client:                client,
+		keepConnAliveInterval: f.option.KeepConnAliveInterval,
+		lastHeartbeatTime:     time.Now().Add(-f.option.KeepConnAliveInterval),
```
3AceShowHand, line 109: simply set this as time.Now() should be fine.

pkg/sink/kafka/sarama_factory.go
```diff
@@ -75,11 +75,13 @@ func (f *saramaFactory) AdminClient(ctx context.Context) (ClusterAdminClient, er
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	return &saramaAdminClient{
+	a := saramaAdminClient{
```
3AceShowHand: unnecessary modification.

cdc/sink/ddlsink/mq/mq_ddl_sink.go
wlwilliamx: The DDLProducer interface has multiple implementations. Adding a method like Heartbeat solely for this specific case, which is not needed by other implementations, is unnecessary and could cause confusion for users of those other implementations.

pkg/sink/kafka/factory.go
wlwilliamx: It's already a function: `func (p *saramaSyncProducer) HeartbeatBrokers()`.

cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
wlwilliamx: Creating a new encoder is quite complex and involves a lot of unnecessary setup. In this case, mocking one is the simplest and most convenient approach, especially since we don’t need any real encoder functionality at all.

cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
wlwilliamx: This is the expected logic coverage — since WriteCheckpointTs is only called once, the heartbeat should also be received only once.

cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
wlwilliamx: This is intentionally designed to return an error. The WriteCheckpointTs function calls EncodeCheckpointEvent, and if EncodeCheckpointEvent does not return an error, WriteCheckpointTs will continue executing. However, the subsequent logic is not relevant to this unit test.

cdc/sink/dmlsink/mq/manager/kafka_manager.go
wlwilliamx: Good suggestion

cdc/sink/util/helper_test.go
wlwilliamx: This change is only because an interval parameter was added to the function. Since this unit test doesn’t need it, a fake FOREVER interval is used.

cdc/sink/dmlsink/mq/manager/kafka_manager_test.go
wlwilliamx: Same as below.

pkg/sink/kafka/options.go
wlwilliamx: You want to minimize the impact, right? Good suggestion.

pkg/sink/kafka/sarama_factory.go
wlwilliamx: Just make sure the first heartbeat can be sent successfully.

pkg/sink/kafka/sarama_factory.go
wlwilliamx: done

cdc/sink/ddlsink/mq/mq_ddl_sink.go
wk989898: There are two implementations: Kafka and Pulsar. I think it is more elegant to add a heartbeat method. The user doesn't need to understand the implementations.

pkg/sink/kafka/factory.go
wk989898: This code snippet repeats three times in your PR. If the code needs to be modified, we have to update it three times.

pkg/sink/kafka/sarama_factory.go
wk989898: It is not necessary to send a heartbeat when the Sarama client is created.

cdc/sink/ddlsink/mq/mq_ddl_sink_test.go
3AceShowHand: but should not report error.

wk989898 (approved):

pkg/sink/kafka/sarama_factory.go
wlwilliamx: It will be closed in sarama producer.

PR #12183, opened 2025-05-23
codec(ticdc): support header line for CSV protocol

### LINKED ISSUE #12179: Add header for CSV protocol

No

Add Field name output to TiCDC CSV Output Protocol as an option. When the option is true, adding a header for CSV output.
Define header column name according to the [doc](https://docs.pingcap.com/tidb/stable/ticdc-csv/#definition-of-the-data-format).
> In the CSV file, each column is defined as follows:  
>
> Column 1: The operation-type indicator, including I, U, and D. I means INSERT, U means UPDATE, and D means DELETE.
Column 2: Table name.
Column 3: Schema name.
Column 4: The commit-ts of the source transaction. This column is optional.
Column 5: The is-update column only exists when the value of output-old-value is true, which is used to identify whether the row data change comes from the UPDATE event (the value of the column is true) or the INSERT/DELETE event (the value is false).
Column 6 to the last column: One or more columns with data changes.

_No response_

When creating the changefeed configuration file, the user may specify the field `output-field-header` as a boolean configuration setting when specifying the sink configuration for the changefeed, e.g.
```toml
[sink.csv]
output-field-header = true
```
The default value will be false (for off) and match the current behavior of CSV output.

wk989898:

Issue Number: close #12179

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
Support header line for CSV protocol.
```


pkg/sink/codec/common/message.go
```diff
@@ -32,6 +32,7 @@ const MaxRecordOverhead = 5*binary.MaxVarintLen32 + binary.MaxVarintLen64 + 1
 
 // Message represents an message to the sink
 type Message struct {
+	Header    []byte // only for csv header
 	Key       []byte
```
3AceShowHand: can we just reuse the `key` field?

3AceShowHand (approved):

wk989898:
@benmeadowcroft PTAL.

PR #12236, opened 2025-07-09
sink(ticdc): calculate partition by the orignal column name

### LINKED ISSUE #12103: make the case-sensitive works on the whole dispatchers configuration

wk989898:

Issue Number: close #12103

This pr https://github.com/pingcap/tiflow/pull/12132 makes the column and index names case-insensitive in dispatcher. But it also miscalculates the partition with the lowercase column name.
The `ColumnsDispatcher` and `IndexValueDispatcher` now correctly use the original column names for hashing, ensuring consistent and accurate data partitioning.

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note
`None`.
```

### DIFF
--- a/cdc/sink/dmlsink/mq/dispatcher/partition/index_value_test.go
+++ b/cdc/sink/dmlsink/mq/dispatcher/partition/index_value_test.go
@@ -151,7 +151,9 @@ func TestIndexValueDispatcherWithIndexName(t *testing.T) {
 		Name:       pmodel.NewCIStr("t1"),
 		PKIsHandle: true,
 		Columns: []*timodel.ColumnInfo{
-			{ID: 1, Name: pmodel.NewCIStr("A"), FieldType: *types.NewFieldType(mysql.TypeLong)},
+			{ID: 1, Name: pmodel.NewCIStr("col2"), Offset: 1, FieldType: *types.NewFieldType(mysql.TypeLong)},
+			{ID: 2, Name: pmodel.NewCIStr("col1"), Offset: 0, FieldType: *types.NewFieldType(mysql.TypeLong)},
+			{ID: 3, Name: pmodel.NewCIStr("col3"), Offset: 2, FieldType: *types.NewFieldType(mysql.TypeLong)},
 		},
 		Indices: []*timodel.IndexInfo{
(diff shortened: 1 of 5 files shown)


3AceShowHand (approved):

cdc/sink/dmlsink/mq/dispatcher/partition/index_value_test.go
```diff
@@ -183,13 +183,14 @@ func TestIndexValueDispatcherWithIndexName(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, int32(2), index)
 
+	idx := index
```
flowbehappy, line 193: Just a confirm: can the test case cover the case that not using the original column name can not pass the test?

asddongmen (approved):

cdc/sink/dmlsink/mq/dispatcher/partition/index_value_test.go
wk989898: If the column names are not the original column names, the partition should match the original column names.

PR #12268, opened 2025-07-29
sync_diff_inspector: Use TLS by default

### LINKED ISSUE #12269: sync-diff-inspector: doesn't use TLS if available
## Feature Request

Currently sync-diff-inspector only uses TLS when configured with a CA certificate.

The result is that it doesn't use TLS if there is no security section in the config.

This leads to failures if `require_secure_transport=ON` is used.

dveeden:

Issue Number: close #12269 

This PR makes it:
- Use TLS if available. With fallback to non-TLS if it is not available.
- This will not do CA validation as there is no CA configured.
- This results that without TLS config `sync-diff-inspector` can connect to non-TLS servers and servers that use TLS, including `require_secure_transport=ON`. 

It is still recommended to configure a CA certificate and if needed a client certificate and key as the CA certificate can prevent against MitM attacks.

Alternatives
- We could add an option like `security.tlsmode` with values like `disabled`/`required`/`preferred`/`validate_ca`/`validate_identity` that follows the similarly named options in the MySQL Client.
- We could make an option that turns on TLS without fallback to non-TLS, but without the CA validation.

For the unittest
- Fixed the order of the actual and expected output for the lines that I touched
- Using predictive uuids
#### Tests 

### Release note 

```release-note
`sync_diff_inspector` now uses TLS when it is available
```


dveeden:
From the `pull-syncdiff-integration-test`:
```
An error occurred while initializing diff: tls: server selected unsupported protocol version 302, please check log info in /tmp/sync_diff_inspector_test/sync_diff_inspector/output/sync_diff.log for full details
```

Note that TLS version 302 is also known asl TLSv1.1.

See also https://go.dev/src/crypto/tls/common.go
```
	VersionTLS10 = 0x0301
	VersionTLS11 = 0x0302
	VersionTLS12 = 0x0303
	VersionTLS13 = 0x0304
```

And also in crypto/tls:
```
        // MinVersion contains the minimum TLS version that is acceptable.
	//
	// By default, TLS 1.2 is currently used as the minimum. TLS 1.0 is the
	// minimum supported by this package.
	//
	// The server-side default can be reverted to TLS 1.0 by including the value
	// "tls10server=1" in the GODEBUG environment variable.
	MinVersion uint16
```

So I think this test might be using a old and EOL version of MySQL and/or be misconfigured.

However failures against servers with TLSv1.1 and older is something we should at least mention in the release notes.

Leavrth (approved):

sync_diff_inspector/config/config.go
```diff
@@ -155,17 +156,26 @@ func (d *DataSource) ToDBConfig() *dbutil.DBConfig {
 
 // RegisterTLS register TLS config for driver
 func (d *DataSource) RegisterTLS() error {
+	var err error
+	var sec *Security
+	var tlsConfig *tls.Config
+
 	if d.Security == nil {
-		return nil
+		d.Security = &Security{}
```
lance6716, line 171: Do we need to modify `d`? Seems in this function we can use `sec = &Security{}` and don't need to change it.

lance6716 (commented):
/cc @joechenrh

sync_diff_inspector/config/config.go
joechenrh: Seems like we need the tls name in creating connection, maybe we should change it here.

dveeden: In `ToDriverConfig()` there is a `if d.Security != nil {`

dveeden: Maybe this would work:
```patch
diff --git a/sync_diff_inspector/config/config.go b/sync_diff_inspector/config/config.go
index 6677cafb6..494d3e9b1 100644
--- a/sync_diff_inspector/config/config.go
+++ b/sync_diff_inspector/config/config.go
@@ -161,8 +161,7 @@ func (d *DataSource) RegisterTLS() error {
        var tlsConfig *tls.Config
 
        if d.Security == nil {
-               d.Security = &Security{}
-               sec = d.Security
+               sec = &Security{}
                tlsConfig = &tls.Config{
                        InsecureSkipVerify: true, // #nosec G402
                }
@@ -210,9 +209,8 @@ func (d *DataSource) ToDriverConfig() *mysql.Config {
                if len(d.Security.TLSName) > 0 {
                        cfg.TLSConfig = d.Security.TLSName
                }
-               if d.Security.CAPath == "" {
-                       cfg.AllowFallbackToPlaintext = true
-               }
+       } else {
+               cfg.AllowFallbackToPlaintext = true
        }
 
        for param, value := range d.SessionConfig {
```

dveeden: That doesn't work. `d.Security.TLSName` is used in `ToDriverConfig()`

lance6716:
PTAL @joechenrh

joechenrh:
I can't approve 😂 /cc @kennytm

lance6716:
> I can't approve 😂 /cc @kennytm

Sorry just remind you to review the code because I remember you have more recent experience than us in sync-diff. Seem CI is broken due to TLS version @dveeden

dveeden:
> > I can't approve 😂 /cc @kennytm
> 
> Sorry just remind you to review the code because I remember you have more recent experience than us in sync-diff. Seem CI is broken due to TLS version @dveeden

See also https://github.com/pingcap/tiflow/pull/12268#issuecomment-3133949130

I think CI was upgraded from a vintage 5.7 MySQL to MySQL 8.0.4x. That should solve this issue.

joechenrh (approved):

kennytm:
4 tests failed, seems legit.

```
An error occurred while initializing diff: tls: server selected unsupported protocol version 302, please check log info in /tmp/sync_diff_inspector_test/sync_diff_inspector/output/sync_diff.log for full details
```

sync_diff_inspector/config/config.go
```diff
@@ -155,17 +156,26 @@ func (d *DataSource) ToDBConfig() *dbutil.DBConfig {
...
+		sec = d.Security
+		tlsConfig = &tls.Config{
+			InsecureSkipVerify: true, // #nosec G402
```
lance6716, line 174: dumping can export min TLS version

https://github.com/pingcap/tidb/blob/55cf58bf1681fe1d584e95cede4fa49c7197682a/dumpling/export/config.go#L282-L288

and DM sets it to VersionTLS10

https://github.com/pingcap/tiflow/blob/b949fa6674958aca43dc55c711e0fc41a6600588/dm/dumpling/dumpling.go#L433

I remember the decision is that user may still need to migaret from old MySQL instances. So we should allow old TLS version. It's also friendly for sync-diff to do this.

dveeden: Setting the minversion isn't sufficient for that anymore iirc.

Note that more most old MySQL 5.7 versions support TLSv1.2 or newer. Only very old versions don't.

But having said that, I'll try to see what's needed to make this work

PR #12335, opened 2025-09-25
sync-diff-inspector: check foreign key when comparing table struct

### LINKED ISSUE #12333: sync diff does not detect table strcture mismatch on the foreign key
Image

The upstream and downstream table structure is not the same, but can pass the sync diff check. 

TiDB officially announce the Foreign Key functionality GA since v8.5

joechenrh:

Issue Number: close #12333

Check foreign key when comparing table structure, and different fk won't skip the data comparison.

Because MySQL and TiDB behave different for the name of foreign key, we only check the foreign keys one by one from both tables.

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note

```

### DIFF
--- a/sync_diff_inspector/utils/utils.go
+++ b/sync_diff_inspector/utils/utils.go
@@ -380,12 +380,38 @@ func sameProperties(c1, c2 *model.ColumnInfo) bool {
 	}
 }
 
-// CompareStruct compare tables' columns and indices from upstream and downstream.
+// equalFK checks whether 2 foreign keys are equal.
+func equalFK(fk1, fk2 *model.FKInfo) bool {
+	if fk1.RefSchema.L != fk2.RefSchema.L ||
+		fk1.RefTable.L != fk2.RefTable.L ||
+		fk1.OnDelete != fk2.OnDelete ||
+		fk1.OnUpdate != fk2.OnUpdate ||
+		len(fk1.Cols) != len(fk2.Cols) ||
+		len(fk1.RefCols) != len(fk2.RefCols) {
+		return false
+	}
+
+	for i := range fk1.Cols {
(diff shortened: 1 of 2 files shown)


sync_diff_inspector/utils/utils.go
```diff
@@ -439,6 +469,34 @@ func CompareStruct(upstreamTableInfos []*model.TableInfo, downstreamTableInfo *m
...
 	}
 
+	// compare foreign key
+	fkEqual := true
+	for _, upstreamTableInfo := range upstreamTableInfos {
+		upstreamFKs := upstreamTableInfo.ForeignKeys
+		downstreamFKs := downstreamTableInfo.ForeignKeys
+		if len(upstreamFKs) != len(downstreamFKs) {
+			log.Warn("Upstream and downstream foreign key count not equal",
+				zap.Int("upstream count", len(upstreamFKs)), zap.Int("downstream count", len(downstreamFKs)))
```
kennytm: also which which `upstreamTableInfo` it is working on in the warning.

sync_diff_inspector/utils/utils.go
```diff
@@ -439,6 +469,34 @@ func CompareStruct(upstreamTableInfos []*model.TableInfo, downstreamTableInfo *m
...
+			fkEqual = false
+			break
+		}
+
+		for i, upFK := range upstreamFKs {
+			downFK := downstreamFKs[i]
+			if !equalFK(upFK, downFK) {
+				log.Warn("Foreign key not equal",
+					zap.String("upstream FK", upFK.Name.O),
+					zap.String("downstream FK", downFK.Name.O))
```
kennytm: ditto. include the db+table name too.

Issue #12350, opened 2025-09-30
DM: Foreign Key Support in DM (master issue)


OliverS929:
### Context
DM does not natively support foreign keys or cascading actions, since MySQL executes cascades inside InnoDB and they do not appear in binlogs. This causes replication errors and inconsistencies in workloads that depend on Foreign keys.

### Problem Areas
- Safe mode: current UPDATE → DELETE+REPLACE breaks cascades, leading to child row loss.
- Schema tracker: no FK awareness; cannot distinguish updates touching FK-referenced columns.
- Causality & concurrency: out-of-order parent/child execution due to lack of FK relations in conflict detection.
- Partial inclusion: filtering/merging may include some tables in a FK chain but not others.

### Planned Work
- Minimal fix: skip unnecessary DELETE when PK/UK unchanged; REPLACE without triggering cascades. Intended for scenarios where FK-referenced PK/UK are never updated and schemas are pre-created manually.
- Schema tracker + DDL: collect FK info at init or from DDL, distinguish safe vs unsafe updates, enforce consistent table inclusion.
- Prechecks: validate FK usage is standard and compatible (e.g. no non-standard keys).
- Causality extension: incorporate FK relations into conflict detection to ensure parent/child order.

PR #12351, opened 2025-09-30
DM: remove redundant DELETE and force non-cascade REPLACE in safe mode to support Foreign Key constraint with non-key updates

OliverS929:

Issue Number: ref #12350

1. Remove redundant Delete statement from safe mode rewrite of Update when row identity(pk and uks) are not changed during the Update.
2. Force replace into from safe mode to be executed without triggering cascade behaviors.
3. Add integration test to verify safe mode under insert and non-key updates with FK and on cascade delete enabled.

#### Tests 

 - [x] Unit test
 - [x] Integration test
 - No code

### Release note 

```release-note

```
Remove redundant DELETE and force non-cascade REPLACE in safe mode to support Foreign Key constraint with non-key updates.

### DIFF
--- a/dm/syncer/dml_worker.go
+++ b/dm/syncer/dml_worker.go
@@ -194,19 +194,19 @@ func (w *DMLWorker) executeJobs(queueID int, jobCh chan *job) {
 	disableForeignKeyChecksForBatch := false
 
 	flushJobs := func() {
-		if len(jobs) == 0 {
-			return
-		}
-
 		failpoint.Inject("syncDMLBatchNotFull", func() {
 			if len(jobCh) == 0 && len(jobs) < w.batch {
 				w.logger.Info("execute not full job queue")
 			}
 		})
 
-		w.executeBatchJobs(queueID, jobs, disableForeignKeyChecksForBatch)
-		jobs = jobs[0:0]
-		disableForeignKeyChecksForBatch = false
+		// Skip the actual execution if there’s nothing to flush
+		if len(jobs) > 0 {
+			w.executeBatchJobs(queueID, jobs, disableForeignKeyChecksForBatch)
+			jobs = jobs[0:0]
+			disableForeignKeyChecksForBatch = false
+		}
+
 		if len(jobCh) == 0 {
 			failpoint.Inject("noJobInQueueLog", func() {
 				w.logger.Debug("no job in queue, update lag to zero", zap.Int(


OliverS929:
Manual tests:

Operation and results in upstream MySQL 8.0:
```sql
mysql> use fk_demo;
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
mysql> SELECT * FROM parent ORDER BY parent_id;
+-----------+---------+
| parent_id | payload |
+-----------+---------+
|         1 | p1      |
|         2 | p2      |
|         3 | p3      |
+-----------+---------+
3 rows in set (0.00 sec)

mysql> SELECT parent_id, COUNT(*) AS cnt FROM child GROUP BY parent_id ORDER BY parent_id;
+-----------+-----+
| parent_id | cnt |
+-----------+-----+
|         1 |   2 |
|         2 |   1 |
|         3 |   1 |
+-----------+-----+
3 rows in set (0.00 sec)

mysql> UPDATE parent SET payload='p1_v2' WHERE parent_id=1;
Query OK, 1 row affected (0.01 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> SELECT payload FROM parent WHERE parent_id=1;          -- expect 'p1_v2'
+---------+
| payload |
+---------+
| p1_v2   |
+---------+
1 row in set (0.00 sec)

mysql> SELECT COUNT(*) FROM child WHERE parent_id=1;          -- expect 2 (no loss)
+----------+
| COUNT(*) |
+----------+
|        2 |
+----------+
1 row in set (0.00 sec)

mysql> UPDATE parent SET payload='p1_v2' WHERE parent_id=1;   -- same value OK
Query OK, 0 rows affected (0.01 sec)
Rows matched: 1  Changed: 0  Warnings: 0

mysql> UPDATE parent SET payload='p1_v3' WHERE parent_id=1;
Query OK, 1 row affected (0.01 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> SELECT payload FROM parent WHERE parent_id=1;          -- 'p1_v2' or 'p1_v3' accordingly
+---------+
| payload |
+---------+
| p1_v3   |
+---------+
1 row in set (0.01 sec)

mysql> SELECT COUNT(*) FROM child WHERE parent_id=1;          -- still 2
+----------+
| COUNT(*) |
+----------+
|        2 |
+----------+
1 row in set (0.00 sec)

mysql> UPDATE child SET child_data='c20_v2' WHERE child_id=20;
Query OK, 1 row affected (0.00 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> SELECT child_data FROM child WHERE child_id=20;        -- 'c20_v2'
+------------+
| child_data |
+------------+
| c20_v2     |
+------------+
1 row in set (0.00 sec)

mysql> SELECT COUNT(*) FROM parent WHERE parent_id=2;         -- 1 (no parent impact)
+----------+
| COUNT(*) |
+----------+
|        1 |
+----------+
1 row in set (0.00 sec)

mysql> DELETE FROM parent WHERE parent_id=3;
Query OK, 1 row affected (0.00 sec)

mysql> SELECT COUNT(*) FROM parent WHERE parent_id=3;         -- 0
+----------+
| COUNT(*) |
+----------+
|        0 |
+----------+
1 row in set (0.00 sec)

mysql> SELECT COUNT(*) FROM child  WHERE parent_id=3;         -- 0 (cascaded)
+----------+
| COUNT(*) |
+----------+
|        0 |
+----------+
1 row in set (0.01 sec)

mysql> -- insert a new child, then re-insert to force a duplicate
mysql> INSERT INTO child(child_id, parent_id, child_data) VALUES (21,2,'c21');
Query OK, 1 row affected (0.00 sec)

mysql> INSERT INTO child(child_id, parent_id, child_data) VALUES (21,2,'c21'); -- duplicate/replay
ERROR 1062 (23000): Duplicate entry '21' for key 'child.PRIMARY'
mysql>
mysql> SELECT COUNT(*) FROM child WHERE child_id=21;          -- 1
+----------+
| COUNT(*) |
+----------+
|        1 |
+----------+
1 row in set (0.01 sec)
```

Query Results from downstream TiDB with a DM sync task running:
```sql
mysql> use fk_demo;
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed

mysql> SELECT * FROM parent ORDER BY parent_id;
+-----------+---------+
| parent_id | payload |
+-----------+---------+
|         1 | p1      |
|         2 | p2      |
|         3 | p3      |
+-----------+---------+
3 rows in set (0.00 sec)

mysql> SELECT parent_id, COUNT(*) AS cnt FROM child GROUP BY parent_id ORDER BY parent_id;
+-----------+-----+
| parent_id | cnt |
+-----------+-----+
|         1 |   2 |
|         2 |   1 |
|         3 |   1 |
+-----------+-----+
3 rows in set (0.00 sec)

mysql> SELECT payload FROM parent WHERE parent_id=1;          -- expect 'p1_v2'
+---------+
| payload |
+---------+
| p1_v2   |
+---------+
1 row in set (0.00 sec)

mysql> SELECT COUNT(*) FROM child WHERE parent_id=1;          -- expect 2 (no loss)
+----------+
| COUNT(*) |
+----------+
|        2 |
+----------+
1 row in set (0.00 sec)

mysql> SELECT payload FROM parent WHERE parent_id=1;          -- 'p1_v2' or 'p1_v3' accordingly
+---------+
| payload |
+---------+
| p1_v3   |
+---------+
1 row in set (0.00 sec)

mysql> SELECT COUNT(*) FROM child WHERE parent_id=1;          -- still 2
+----------+
| COUNT(*) |
+----------+
|        2 |
+----------+
1 row in set (0.00 sec)

mysql> SELECT child_data FROM child WHERE child_id=20;        -- 'c20_v2'
+------------+
| child_data |
+------------+
| c20_v2     |
+------------+
1 row in set (0.00 sec)

mysql> SELECT COUNT(*) FROM parent WHERE parent_id=2;         -- 1 (no parent impact)
+----------+
| COUNT(*) |
+----------+
|        1 |
+----------+
1 row in set (0.00 sec)

mysql> SELECT COUNT(*) FROM parent WHERE parent_id=3;         -- 0
+----------+
| COUNT(*) |
+----------+
|        0 |
+----------+
1 row in set (0.00 sec)

mysql> SELECT COUNT(*) FROM child  WHERE parent_id=3;         -- 0 (cascaded)
+----------+
| COUNT(*) |
+----------+
|        0 |
+----------+
1 row in set (0.00 sec)

mysql> SELECT COUNT(*) FROM child WHERE child_id=21;          -- 1
+----------+
| COUNT(*) |
+----------+
|        1 |
+----------+
1 row in set (0.00 sec)
```

OliverS929:
cc @GMHDBJD @lance6716 @D3Hunter @Benjamin2037 PTAL.

dm/syncer/dml_worker.go
```diff
@@ -58,31 +62,66 @@ type DMLWorker struct {
...
+		worker:                  syncer.cfg.WorkerName,
+		logger:                  syncer.tctx.Logger.WithFields(zap.String("component", "dml_worker")),
+		successFunc:             syncer.successFunc,
+		fatalFunc:               syncer.fatalFunc,
+		lagFunc:                 syncer.updateReplicationJobTS,
+		updateJobMetricsFunc:    syncer.updateJobMetrics,
+		syncCtx:                 syncer.syncCtx, // this ctx can be used to cancel all the workers
+		metricProxies:           syncer.metricsProxies,
+		toDBConns:               syncer.toDBConns,
+		foreignKeyChecksEnabled: isForeignKeyChecksEnabled(syncer.cfg.To.Session),
```
GMHDBJD, line 114: Should we check it via `show variables` from downsteam? What if user manually set `ForeignKeyChecks` in downsteam in GLOBAL scope?

OliverS929: I believe DM intentionally sets foreign_key_check to off unless the user explicitly enables it in the session variables (see [code reference](https://github.com/pingcap/tiflow/blob/master/dm/pkg/conn/basedb.go#L140-L153)). This ensures that DM forces the foreign key behavior to be off regardless of the downstream configuration, since DM does not currently support foreign keys.

I think it’s sufficient to simply check the session variables in the settings, since the session-level configuration will override the global setting anyway.

OliverS929:
> Should we report error if user update the PK/UK in this pr?

Yes, we should, but not in this PR. This PR only addresses the issue of avoiding "unnecessary" incorrect cascade behavior triggered by safe mode. PK/UK updates in safe mode will still cause problems regardless, and I will raise a separate PR to add the necessary check and error message.

> Should we report error if syncer.compact is true but foreign key check is enabled?

Yes, we should handle this for both compact and merger. Adding another precheck item would be a better solution here, and this will also be included in a future PR.

cc @GMHDBJD

dm/syncer/dml.go
```diff
@@ -419,10 +419,12 @@ func genDMLsWithSameTable(op sqlmodel.DMLType, jobs []*job) ([]string, [][]inter
 	if op == sqlmodel.DMLUpdate {
 		for i, j := range jobs {
 			if j.safeMode {
-				query, arg := j.dml.GenSQL(sqlmodel.DMLDelete)
-				queries = append(queries, query)
-				args = append(args, arg)
-				query, arg = j.dml.GenSQL(sqlmodel.DMLReplace)
+				if j.dml.IsPrimaryOrUniqueKeyUpdated() {
```
lance6716, line 422: in a rare case, if user manually change downstream structure and didn't notify DM, DM may wrongly calculate `IsPrimaryOrUniqueKeyUpdated` 🤔 

I forget if we have forbidden ti in user doc

dm/tests/shardddl1/run.sh
```diff
@@ -801,7 +801,7 @@ function DM_CAUSALITY_USE_DOWNSTREAM_SCHEMA() {
 
 function DM_DML_EXECUTE_ERROR_CASE() {
 	run_sql_source1 "insert into ${shardddl1}.${tb1}(a,b) values(1,1)"
-	run_sql_source1 "update ${shardddl1}.${tb1} set b=b+1 where a=1"
+	run_sql_source1 "update ${shardddl1}.${tb1} set a=a+1, b=b+1 where a=1"
```
lance6716: why we need to change it?

dm/syncer/dml_worker.go
```diff
@@ -58,31 +62,66 @@ type DMLWorker struct {
 	flushCh chan *job
 }
 
+func isForeignKeyChecksEnabled(session map[string]string) bool {
+	if session == nil {
```
lance6716: seems no need to check `nil`. `for` over a `nil` map is OK in go.

dm/syncer/dml_worker.go
```diff
@@ -58,31 +62,66 @@ type DMLWorker struct {
...
+		return false
+	}
+	for key, value := range session {
+		if strings.EqualFold(key, "foreign_key_checks") {
+			trimmed := strings.TrimSpace(value)
+			trimmed = strings.Trim(trimmed, "'\"")
+			switch strings.ToLower(trimmed) {
+			case "1", "on", "true":
```
lance6716: maybe ref the usage of https://github.com/pingcap/tidb/blob/50a1e5a631d6ea8bb4e3aa7950e9458351976449/pkg/sessionctx/variable/varsutil.go#L184

dm/syncer/dml_worker.go
```diff
@@ -58,31 +62,66 @@ type DMLWorker struct {
...
+			default:
+				return false
+			}
+		}
+	}
+	return false
+}
+
+func (w *DMLWorker) shouldDisableForeignKeyChecksForJob(j *job) bool {
+	if !w.foreignKeyChecksEnabled {
```
lance6716, line 77: note this line means if check is not enabled, so it's disabled, so return `true` to `shouldDisableForeignKeyChecksForJob`?

dm/syncer/dml.go
OliverS929: Yeah, I think this is a more general issue that applies to all use cases involving the schema tracker, not just `IsPrimaryOrUniqueKeyUpdated`. We don’t really keep close track of downstream schemas, and the current implementation works on an on-demand basis. We more or less rely on users to know what they are doing and not make changes to the downstream schemas. In any case, `IsPrimaryOrUniqueKeyUpdated` follows the same assumption, and addressing this would require broader changes beyond the scope of this PR. It should probably be tracked as a feature request or improvement.

dm/syncer/dml_worker.go
OliverS929: The actual purpose of the `shouldDisableForeignKeyChecksForJob` function is to determine whether the caller needs to execute a new batch with FKC turned off. If FKC is not enabled in the task configuration, we don’t need to handle it at all, since the session will have FKC turned off by default.

dm/syncer/dml_worker.go
OliverS929: Thanks. Changed in 506211ee30fb2777ef2666aa6fb92081daa30c03.

dm/syncer/dml_worker.go
OliverS929: Thanks. Changed in 506211e.

dm/tests/shardddl1/run.sh
OliverS929: I tried adjusting the test setup so that the failpoint `ErrorOnLastDML` in `DM_DML_EXECUTE_ERROR_CASE` could still be triggered as expected. However, it no longer works with the current setup, even after the adjustment. I commented it out in 25b77b4451a1dde7d5e36b69c7d8922e1f575cb9 after discussing with @GMHDBJD.

PR #12357, opened 2025-10-11
dm: add primary key checking item for aliyun rds

### LINKED ISSUE #12338: DM: add primary key precheck item

aliyun oss need primary key for upstream table, otherwise the sync task will report column mismatch error

add a primary key precheck item

_No response_

_No response_

GMHDBJD:

Issue Number: close #12338
#### Tests 

 - Unit test

### Release note 

```release-note
None
```


dm/pkg/checker/primary_key.go
```diff
@@ -0,0 +1,180 @@
...
+		ret = append(ret, opt)
+		return ret, nil
+	}
+
+	// check primary key existence
+	pkuk := getPKAndUK(upstreamStmt)
+	if _, ok := pkuk["PRIMARY"]; !ok {
+		ret = append(ret, &incompatibilityOption{
+			state:       StateFailure,
+			instruction: "You need to set primary key for the upstream table. Otherwise replication efficiency might become very low and exactly-once replication cannot be guaranteed.",
```
OliverS929: I don't think we are guaranteeing exactly-once replication with or without this check. This part of information in the error messsage here is potentially stonger than intended and missleading. May be something like “…replication efficiency may be low, and correctness (e.g., deduplication and idempotency) may be affected if no primary key exists.”

dm/config/checking_item.go
```diff
@@ -45,6 +45,9 @@ const (
 	LightningFreeSpaceChecking          = "free_space"
 	LightningMutexFeatureChecking       = "downstream_mutex_features"
 	LightningTableEmptyChecking         = "downstream_table_empty"
+	// Aliyun RDS prechecks.
```
OliverS929: Maybe consider make the comment more generic, with less emphasize on Aliyun RDS, since this check would be in general useful for any MySQL upstream.
A potential comment could be 
"
// primary_key: checks that each upstream table has a defined PRIMARY KEY.
// This check is mainly required for Aliyun RDS migration scenarios,
// where missing primary keys can cause replication inefficiency or correctness issues.
// Not enabled by default; runs only when explicitly included in AdditionalCheckingItems.
"
in addition to what you have already wrote.

dm/pkg/checker/primary_key.go
```diff
@@ -0,0 +1,180 @@
...
+		var err error
+		w.upstreamParser, err = dbutil.GetParserForDB(ctx, w.c.upstreamDBs[w.lastSourceID].DB)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	db := w.c.upstreamDBs[item.sourceID].DB
+	upstreamSQL, err := dbutil.GetCreateTableSQL(ctx, db, table.Schema, table.Name)
+	if err != nil {
```
OliverS929, line 78: It might be worth adding a small unit test to cover this branch, where SHOW CREATE TABLE returns a mysql.ErrNoSuchTable error.

dm/pkg/checker/primary_key.go
```diff
@@ -0,0 +1,180 @@
...
+		// continue if table was deleted when checking
+		if isMySQLError(err, mysql.ErrNoSuchTable) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	upstreamStmt, err := getCreateTableStmt(w.upstreamParser, upstreamSQL)
+	if err != nil {
```
OliverS929, line 87: It might be worth adding an unit test to cover this branch.

dm/pkg/checker/primary_key.go
GMHDBJD: [a325392](https://github.com/pingcap/tiflow/pull/12357/commits/a325392f55dab688a60ee40b80d75c1e1e6fbfa2)

dm/pkg/checker/primary_key.go
GMHDBJD: [a325392](https://github.com/pingcap/tiflow/pull/12357/commits/a325392f55dab688a60ee40b80d75c1e1e6fbfa2)

dm/pkg/checker/primary_key.go
```diff
@@ -0,0 +1,180 @@
...
+					resultInstructions[opt.instruction] = struct{}{}
+				}
+				r.Errors = append(r.Errors, e)
+			case StateFailure:
+				r.State = StateFailure
+				e := NewError(tableMsg + opt.errMessage)
+				if _, ok := resultInstructions[opt.instruction]; !ok && opt.instruction != "" {
+					resultInstructions[opt.instruction] = struct{}{}
+				}
+				r.Errors = append(r.Errors, e)
```
JQWong7, line 147: Maybe L137-L140 and L144-147 can uniformly move to the outer layer of switch

dm/config/checking_item.go
```diff
@@ -45,6 +45,12 @@ const (
 	LightningFreeSpaceChecking          = "free_space"
 	LightningMutexFeatureChecking       = "downstream_mutex_features"
 	LightningTableEmptyChecking         = "downstream_table_empty"
+	// primary_key: checks that each upstream table has a defined PRIMARY KEY.
```
lance6716, line 48: Sorry I'm not sure about the background of this PR. The doc says if table has no PK in Aliyun RDS DM can't work due to the hidden column. And here the comment says DM can work but it's less efficiency. I think if DM can make use of the hidden column (please check https://github.com/pingcap/tiflow/issues/12338#issuecomment-3343406585) it will be as fast as normal tables.

GMHDBJD: Although this is to solve the hidden primary key problem of Aliyun RDS, this checkitem is universal, so I think the universal explanation is better?

dm/config/checking_item.go
```diff
@@ -84,6 +90,13 @@ var LightningPrechecks = []string{
 	LightningTableEmptyChecking,
 }
 
+// AdditionalCheckingItems contains checking items that are not enabled by default
+// via AllCheckingItems but can be exposed to callers (for example, cloud APIs)
+// so they can request these checks explicitly.
+var AdditionalCheckingItems = map[string]string{
+	PrimaryKeyChecking: "primary key checking item (Aliyun RDS precheck)",
```
lance6716, line 97: Do we need to mention Aliyun RDS here? 
If it's mainly designed for Aliyun RDS use case, I think use `show_ipk_info ` is more friendly.

dm/pkg/checker/primary_key.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/pingcap/tidb/pkg/util/dbutil"
+	"github.com/pingcap/tidb/pkg/util/filter"
+	"github.com/pingcap/tiflow/dm/pkg/conn"
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"go.uber.org/zap"
+	"golang.org/x/exp/maps"
+)
+
+// PrimaryKeyChecker checks whether upstream tables contain a primary key.
+type PrimaryKeyChecker struct {
```
lance6716, line 32: (didn't open my IDE. Does it need to be public?)

dm/tests/all_mode/run.sh
```diff
@@ -266,8 +266,6 @@ function test_regexpr_router() {
 	run_sql_tidb 'create database dtest2;'
 	run_sql_tidb 'drop database if exists dtest4;'
 	run_sql_tidb 'create database dtest4;'
-	run_sql_tidb 'create table if not exists dtest2.dtable2(a int, b int);'
```
lance6716: can you explain why delete them?

lance6716 (approved):

PR #12385, opened 2025-11-05
Update dependencies to fix GO-2024-2978

### LINKED ISSUE #12386: Dependencies need updating because of GO-2024-2978
https://pkg.go.dev/vuln/GO-2024-2978

dveeden:
https://pkg.go.dev/vuln/GO-2024-2978

Issue Number: close #12386 

#### Tests 

 - No code

### Release note 

```release-note
Dependencies were updated to fix a potential security issue.
```


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/pingcap/tiflow
 
-go 1.23.12
+go 1.24.0
```
D3Hunter: the latest is go1.24.10

dveeden: I think that's fine as this is only setting the minimum version.

As documented on https://go.dev/doc/modules/gomod-ref#go for go 1.21 or higher: 
"The go line must be greater than or equal to the go line of all dependencies."

The reason for the increase is this:
```
$ go mod tidy -go=1.23.12
go: golang.org/x/crypto@v0.43.0 requires go@1.24.0, but 1.23.12 is requested
```

lance6716: IMO we also need to upgrade the golang version in CD images. And I remember we have chosen to skip upgrading go 1.24 and wait for go 1.25. Maybe all repo's 1.25 upgraing will come soon.

dveeden: Go 1.25 is already out for quite a while (latest: 1.25.4). Note that we can keep this line as is if we upgrade the CD images to 1.25.x.

dveeden: Who is in charge of the CD images? Is there a Dockerfile for that somewhere?

lance6716: in another PR @purelind cares about MySQL images, I think he is also responsible of CD images.

dveeden: @lance6716 I think this now moved to Go 1.25

PR #12396, opened 2025-11-11
DM: Support for MySQL 8.4

### LINKED ISSUE #11020: DM: Support MySQL 8.4

[MySQL 8.4](https://dev.mysql.com/doc/refman/8.4/en/mysql-nutshell.html) removes things like `SHOW MASTER STATUS`. These have been replaced with `SHOW BINARY LOG STATUS`.

So things like this needs an update:

https://github.com/pingcap/tiflow/blob/be1553484fe4c03594eabb8d7435c694e5fd7224/dm/pkg/conn/utils.go#L112

https://github.com/pingcap/tiflow/blob/be1553484fe4c03594eabb8d7435c694e5fd7224/dm/pkg/conn/db.go#L108

This should be done without breaking compatibility with MySQL 8.0, 5.7 or MariaDB

.

_No response_

_No response_

dveeden:

Issue Number: close #11020

~~The version is now a string and checking is done with `strings.HasPrefix(...,"8.4")` which obviously isn't very robust but it does work for now. We could change this into a version range check. The result is that for now this only fixes compatibility with 8.4 (which is an LTS release) and not with 9.x or perhaps with 8.3 or earlier (but after 8.0). But for now only 8.0 and 8.4 are LTS releases.~~

~~It is also possible to store the version as 3 ints, but then we can't use this to detect MariaDB, TiDB, etc in the future.~~

~~This might need more unittests and other kinds of testing.~~

#### Tests 

It improves compatibility for new versions without breaking compatibility with older versions or with MariaDB.

Yes, the version matrix etc.

### Release note 

```release-note
MySQL 8.4 is now supported.
```

### DIFF
--- a/dm/pkg/checker/mysql_server.go
+++ b/dm/pkg/checker/mysql_server.go
@@ -35,14 +35,14 @@ func NewMySQLVersionChecker(db *sql.DB, dbinfo *dbutil.DBConfig) RealChecker {
 }
 
 // SupportedVersion defines the MySQL/MariaDB version that DM/syncer supports
-// * 5.6.0 <= MySQL Version < 8.1.0.
+// * 5.6.0 <= MySQL Version < 8.5.0.
 var SupportedVersion = map[string]struct {
 	Min MySQLVersion
 	Max MySQLVersion
 }{
 	"mysql": {
 		MySQLVersion{5, 6, 0},
-		MySQLVersion{8, 1, 0},
+		MySQLVersion{8, 5, 0},
 	},
 }
 
--- a/dm/pkg/checker/mysql_server_test.go
+++ b/dm/pkg/checker/mysql_server_test.go
@@ -15,6 +15,7 @@ package checker
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/DATA-DOG/go-sqlmock"
@@ -36,7 +37,10 @@ func TestMysqlVersion(t *testing.T) {
 		{"8.0.1-log", true},
 		{"8.0.20", true},
 		{"8.0.35", true},
-		{"8.1.0", false},
+		{"8.1.0", true}, // Innovation release
+		{"8.4.7", true}, // LTS
+		{"8.5.0", false},
+		{"9.5.0", false}, // Innovation release
 		{"5.5.50-MariaDB-1~wheezy", false},
 		{"10.1.1-MariaDB-1~wheezy", false},
 		{"10.1.2-MariaDB-1~wheezy", false},
@@ -48,7 +52,9 @@ func TestMysqlVersion(t *testing.T) {
 			State: StateWarning,
 		}
 		versionChecker.checkVersion(cs.rawVersion, result)
-		require.Equal(t, result.State == StateSuccess, cs.pass)
+		require.Equal(t, result.State == StateSuccess, cs.pass,
+			fmt.Sprintf("version %s, expected pass=%v, but got pass=%v", cs.rawVersion, cs.pass, result.State == StateSuccess),
+		)
 	}
 }
 
@@ -59,8 +65,14 @@ func TestVersionInstruction(t *testing.T) {
 		db:     db,
 		dbinfo: &dbutil.DBConfig{},
 	}
-	mock.ExpectQuery("SHOW GLOBAL VARIABLES LIKE 'version';").WillReturnRows(sqlmock.NewRows([]string{"Variable_name", "Value"}).AddRow("version", "8.1.0"))
+	mock.ExpectQuery("SHOW GLOBAL VARIABLES LIKE 'version';").WillReturnRows(
+		sqlmock.NewRows(
+			[]string{"Variable_name", "Value"},
+		).AddRow(
+			"version", "9.5.0"), // This should be an unsupported version
+	)
 	result := versionChecker.Check(context.Background())
 	require.Equal(t, result.State, StateWarning)
-	require.Equal(t, result.Instruction, "It is recommended that you select a database version that meets the requirements before performing data migration. Otherwise data inconsistency or task exceptions might occur.")
+	require.Equal(t, result.Instruction,
+		"It is recommended that you select a database version that meets the requirements before performing data migration. Otherwise data inconsistency or task exceptions might occur.")
 }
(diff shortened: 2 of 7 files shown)


dm/pkg/conn/db.go
```diff
@@ -104,8 +104,14 @@ func GetRandomServerID(ctx *tcontext.Context, db *BaseDB) (uint32, error) {
 
 // GetSlaveServerID gets all slave server id.
 func GetSlaveServerID(ctx *tcontext.Context, db *BaseDB) (map[uint32]struct{}, error) {
+	replicaQuery := `SHOW SLAVE HOSTS`
+
+	if strings.HasPrefix(db.Version, "8.4") {
```
mjonss: Could we have this as a function like:
```golang
// ReplacedMasterWithPrimary() returns true if commands should use new non-Master/Slave syntax.
func (db *BaseDB) ReplacedMasterWithPrimary() bool {
    // to be extended to properly handle MySQL 8.4+ version, but not MariaDB/TiDB
    return strings.HasPrefix(db.Version, "8.4")
}
```

dm/pkg/checker/mysql_server.go
mjonss: Please update the comment as well.

dm/pkg/checker/mysql_server_test.go
```diff
@@ -36,7 +37,10 @@ func TestMysqlVersion(t *testing.T) {
 		{"8.0.1-log", true},
 		{"8.0.20", true},
 		{"8.0.35", true},
-		{"8.1.0", false},
+		{"8.1.0", true}, // Innovation release
```
mjonss, line 40: Just a stupid, curious question, will we support older Innovation releases as well? I think it is fine...

dm/pkg/conn/basedb.go
```diff
@@ -357,3 +370,37 @@ func (d *BaseDB) Close() error {
...
+	// - https://mariadb.com/docs/server/reference/sql-statements/administrative-sql-statements/show/show-binlog-status
+	// - https://mariadb.com/docs/server/reference/sql-statements/administrative-sql-statements/show/show-replica-hosts
+	//
+	// Old syntax is still accepted.
+	if strings.Contains(d.Version, "MariaDB") {
+		return false
+	}
+
+	// https://dev.mysql.com/doc/relnotes/mysql/8.4/en/news-8-4-0.html#mysqld-8-4-0-deprecation-removal
+	// MySQL 8.4 removed `SHOW MASTER STATUS``.
```
mjonss:
```suggestion
	// MySQL 8.4 removed `SHOW MASTER STATUS`.
```

mjonss (commented):
LGTM, this should fix the DM replication command compatibility with MySQL 8.4

Only some minor questions/suggestions.

dm/pkg/checker/mysql_server_test.go
dveeden: In `SupportedVersion` we set a `Min` and `Max` version.

So if we want to include 5.7, 8.0, 8.4 then that means we also include 8.1 etc. I think this is fine.

On https://docs.pingcap.com/tidb/stable/dm-compatibility-catalog/#data-sources we have more details on this.

D3Hunter (commented):
lgtm

D3Hunter (approved):

dm/pkg/conn/basedb.go
```diff
@@ -190,17 +199,21 @@ type BaseDB struct {
 
 	// only use in unit test
 	doNotClose bool
+
+	// SELECT VERSION()
+	Version string
```
lance6716: no need to be an exported field?

dm/pkg/conn/basedb.go
```diff
@@ -357,3 +370,37 @@ func (d *BaseDB) Close() error {
...
+// NeedsModernTerminology is for checking whether the database version requires modern terminolog.
+//
+// MySQL:
+//   - `SHOW MASTER STATUS` -> `SHOW BINARY LOG STATUS`
+//   - `SHOW SLAVE HOSTS`   -> `SHOW REPLICAS`
+//
+// MariaDB:
+//   - `SHOW MASTER STATUS` -> `SHOW BINLOG STATUS`
+//   - `SHOW SLAVE HOSTS`   -> `SHOW REPLICA HOSTS`
+func (d *BaseDB) NeedsModernTerminology() bool {
```
lance6716: also this one. (I didn't open IDE to check if other package are referring these symbols)

dm/pkg/conn/basedb.go
dveeden: Yes, you're right.

dveeden:
> I have a question: does the MySQL image version in the corresponding pipeline also need to be updated?
> 
>     * pull-dm-compatibility-test https://github.com/PingCAP-QE/ci/blob/main/pipelines/pingcap/tiflow/latest/pod-pull_dm_compatibility_test.yaml
> 
>     * pull-dm-integration-test https://github.com/PingCAP-QE/ci/blob/main/pipelines/pingcap/tiflow/latest/pod-pull_dm_integration_test.yaml

I think we should use a test matrix containing MySQL 5.7, 8.0 and 8.4. Testing compatibility with 8.0 for now is more important than 8.4 compatibility. It would also be good to test against the latest MySQL (9.5 at the moment, this is a innovation release, not LTS) to already track changes in MySQL before they end up in an LTS release.

Besides the addition of MySQL 8.4 in testing we might also need to check what features in 8.0 and 8.4 we should try that might impact compatibility.

- `binlog_checksum=NONE/CRC32`
- `binlog_format=row/statement/mixed`
- `binlog_row_image=minimal`
- `binlog_row_value_options=PARTIAL_JSON`
- `binlog_transaction_compression=ON`
- `binlog_rows_query_log_events=ON`
- `binlog_encryption=ON` (probably doesn't affect us)
- [Tagged GTIDs](https://dev.mysql.com/doc/refman/8.4/en/replication-gtids-concepts.html#:~:text=MySQL%208%2E4%20also%20supports%20tagged%20GTIDs)

See also https://dev.mysql.com/doc/refman/8.4/en/replication-options-binary-log.html

Some of these we might not support, but then we should check if we get a clear and useful error.

However this might be something we can do after this PR is merged.

PR #12405, opened 2025-11-21
dm/syncer: advance checkpoint on rotate events

### LINKED ISSUE #12339: DM's global checkpoint does not follow RotateEvent, causing the checkpoint position to be stuck as long as no real event happened
### What did you do?

1. Create a task. Note the global checkpoint position from the DM log. Let's call it `(000001, 123)`
2. Force binlog rotation several times (e.g. use `FLUSH BINARY LOGS`), so the master binlog file becomes e.g. `000003`+.
3. See that the global checkpoint position is still `(000001, 123)` despite many `meet heartbeat event and then flush jobs`.
4. Perform some DDL/DML. Check that the checkpoint is quickly forwarded to the new position.

<sup><sub>(ref: NAID-11659)</sub></sup>

When all the binlog are exhausted, the syncer should follow binlog rotation events and use the latest binlog file as checkpoint `(000003, 4)`.

### What did you see instead?

Syncer does not follow binlog rotation events.

### Versions of the cluster

DM v8.5.2

_No response_

OliverS929:

Issue Number: close #12339

Advance global checkpoint position based on end location of rotate events.

#### Tests 

 - [x] Unit test
 - Integration test
 - No code

### Release note 

```release-note

```

### DIFF
--- a/dm/syncer/syncer.go
+++ b/dm/syncer/syncer.go
@@ -2557,11 +2557,15 @@ func (s *Syncer) handleRotateEvent(ev *replication.RotateEvent, ec eventContext)
 		} else {
 			ec.tctx.L().Debug("re-replicate shard group", zap.String("event", "rotate"), log.WrapStringerField("location", ec.endLocation), zap.Reflect("re-shard", ec.shardingReSync))
 		}
-		s.saveGlobalPoint(ec.endLocation)
+		if !utils.IsFakeRotateEvent(ec.header) {
+			s.saveGlobalPoint(ec.endLocation)
+		}
 		return nil
 	}
 
-	s.saveGlobalPoint(ec.endLocation)
+	if !utils.IsFakeRotateEvent(ec.header) {
+		s.saveGlobalPoint(ec.endLocation)
+	}
 	ec.tctx.L().Info("", zap.String("event", "rotate"), log.WrapStringerField("location", ec.endLocation))
 	return nil
 }


dm/syncer/syncer.go
```diff
@@ -2557,9 +2557,11 @@ func (s *Syncer) handleRotateEvent(ev *replication.RotateEvent, ec eventContext)
 		} else {
 			ec.tctx.L().Debug("re-replicate shard group", zap.String("event", "rotate"), log.WrapStringerField("location", ec.endLocation), zap.Reflect("re-shard", ec.shardingReSync))
 		}
+		s.saveGlobalPoint(ec.endLocation)
 		return nil
 	}
 
+	s.saveGlobalPoint(ec.endLocation)
```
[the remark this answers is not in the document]
OliverS929: This is intentional: the shardingReSync path returns early, so we must save the global point before returning. The two call sites are mutually exclusive at runtime.

dm/syncer/syncer.go
```diff
@@ -2557,9 +2557,15 @@ func (s *Syncer) handleRotateEvent(ev *replication.RotateEvent, ec eventContext)
 		} else {
 			ec.tctx.L().Debug("re-replicate shard group", zap.String("event", "rotate"), log.WrapStringerField("location", ec.endLocation), zap.Reflect("re-shard", ec.shardingReSync))
 		}
+		if !utils.IsFakeRotateEvent(ec.header) {
```
D3Hunter, line 2560: from the title, seems we expect to save on valid rotate event, not `none fake rotate event` which include normal events

OliverS929: Just to clarify the intent here: this code is inside Syncer.handleRotateEvent, so we’re already handling only ROTATE_EVENT. In this context, I think “non-fake rotate event” is exactly the valid/real rotate event we want.

Issue #12410, opened 2025-11-25
Make EventCacheCount in go-mysql configurable


luyomo:
### Is your feature request related to a problem?

The current DM-Worker memory consumption profile shows that a significant portion of the Go Heap (∼20 GiB in our tests) is consumed by large transaction buffers, primarily within the go-mysql replication loop (specifically, the (*RowsEvent).decodeImage function).

While the Go runtime's allocation is involved, the primary constraint is the size and retention of the event objects created during binlog decoding. The EventCacheCount variable, currently hardcoded, controls the size of the event object pool (EventCache).

In high-throughput environments dealing with very large transactions (e.g., 50 million row INSERTs or large schema migrations), setting a high hardcoded EventCacheCount means the DM-Worker pre-allocates and retains a substantial amount of memory for event objects, which unnecessarily contributes to the Resident Set Size (RSS) and can trigger Out-of-Memory (OOM) issues on resource-constrained hosts.

### Describe the feature you'd like

I propose exposing the EventCacheCount field (defined in go-mysql/replication/binlogsyncer.go#L129) as a configurable parameter that DM users can set in their task configuration.
https://github.com/go-mysql-org/go-mysql/blob/master/replication/binlogsyncer.go#L129

This will allow users to:
1. Reduce Memory Consumption: Lower the EventCacheCount value in environments where memory is critical, thus reducing the size of the pre-allocated event object pool.
2. Optimize Performance: Increase the value if memory is plentiful and the application frequently processes large, complex events, reducing the overhead of repeated memory allocation.

PR #12411, opened 2025-11-25
dm: make binlog event cache configurable

OliverS929:

Issue Number: ref https://github.com/pingcap/tiflow/issues/12410

Introduces a new configuration option, event-cache-count, for both the Syncer and Validator components. It allows users to control the number of events cached during binlog sync and decoding.

By lowering event-cache-count, users can reduce memory usage in scenarios where binlog processing consumes a significant amount of memory, helping prevent potential OOM issues on the worker.

Default value for both behave the same as before.

#### Tests 

 - [x] Unit test
 - Integration test
 - No code

### Release note 

```release-note

```

### DIFF
--- a/dm/syncer/data_validator.go
+++ b/dm/syncer/data_validator.go
@@ -314,7 +314,7 @@ func (v *DataValidator) initialize() error {
 		return err
 	}
 
-	v.syncCfg, err = subtaskCfg2BinlogSyncerCfg(v.cfg, v.timezone, v.syncer.baList)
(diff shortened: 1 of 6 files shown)


dm/syncer/data_validator.go
```diff
@@ -314,7 +314,7 @@ func (v *DataValidator) initialize() error {
 		return err
 	}
 
-	v.syncCfg, err = subtaskCfg2BinlogSyncerCfg(v.cfg, v.timezone, v.syncer.baList)
+	v.syncCfg, err = subtaskCfg2BinlogSyncerCfg(v.cfg, v.timezone, v.syncer.baList, v.cfg.ValidatorCfg.EventCacheCount)
```
D3Hunter, line 317: maybe use the same config as syncer, validator shares many config as syncer

PR #12414, opened 2025-11-25
DM: Add foreign key causality for DM syncer.

OliverS929:

Issue Number: ref https://github.com/pingcap/tiflow/issues/12350

This PR introduces foreign key-causality support to the DM syncer.
- In the schema tracker, DownstreamTableInfo now captures a set of ForeignKeyCausalityRelation entries that represent parent→child table dependencies via foreign key constraints.
- The syncer’s RowChange objects now include these relations so that change processing can honour FK dependencies.
- A new integration test suite foreign_key_multi_worker validates correct multi-worker behaviour when foreign key chains exist, helping ensure correctness under concurrent execution.

#### Tests 

 - [x] Unit test
 - [x] Integration test
 - No code

### Release note 

```release-note

```

### DIFF
--- a/dm/tests/ha_cases_1/run.sh
+++ b/dm/tests/ha_cases_1/run.sh
@@ -118,33 +118,33 @@ function test_kill_and_isolate_worker() {
 	isolate_worker 4 "isolate"
 	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT1" \
 		"query-status test" \
-		"\"stage\": \"Running\"" 2
+		"\"unit\": \"Sync\"" 2
 
 	echo "isolate dm-worker3"
 	isolate_worker 3 "isolate"
 	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT1" \
 		"query-status test" \
-		"\"stage\": \"Running\"" 1 \
+		"\"unit\": \"Sync\"" 1 \
 		"\"result\": false" 1
 
 	echo "disable isolate dm-worker4"
 	isolate_worker 4 "disable_isolate"
 	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT1" \
 		"query-status test" \
-		"\"stage\": \"Running\"" 2
+		"\"unit\": \"Sync\"" 2
 
 	echo "query-status from all dm-master"
 	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT1" \
 		"query-status test" \
-		"\"stage\": \"Running\"" 2
+		"\"unit\": \"Sync\"" 2
 
 	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT2" \
 		"query-status test" \
-		"\"stage\": \"Running\"" 2
+		"\"unit\": \"Sync\"" 2
 
 	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT3" \
 		"query-status test" \
-		"\"stage\": \"Running\"" 2
+		"\"unit\": \"Sync\"" 2
 
 	run_dm_ctl $WORK_DIR "127.0.0.1:$MASTER_PORT" \
 		"pause-task test" \


dm/tests/foreign_key_multi_worker/run.sh
```diff
@@ -0,0 +1,58 @@
...
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
+		"query-status test" \
+		"\"result\": true" 2 \
+		"\"stage\": \"Running\"" 2
+
+	run_sql_tidb_with_retry "SELECT COUNT(*) FROM fk_chain.grandparent;" "COUNT(*): 3"
+	run_sql_tidb_with_retry "SELECT COUNT(*) FROM fk_chain.parent;" "COUNT(*): 2"
+	run_sql_tidb_with_retry "SELECT COUNT(*) FROM fk_chain.child;" "COUNT(*): 4"
+	run_sql_tidb_with_retry "SELECT data FROM fk_chain.child WHERE child_id=100;" "data: c100_updated"
+	run_sql_tidb_with_retry "SELECT parent_id FROM fk_chain.child WHERE child_id=201;" "parent_id: 10"
```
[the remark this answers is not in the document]
OliverS929: In the current dm/tests/foreign_key_multi_worker/run.sh, the table name is consistently fk_chain, not fk.chain. I do not see the typo Gemini flagged.

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +523,139 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+	cache map[string][]sqlmodel.ForeignKeyCausalityRelation,
+	visiting map[string]struct{},
+) ([]sqlmodel.ForeignKeyCausalityRelation, error) {
+	if relations, ok := cache[tableID]; ok {
+		return relations, nil
+	}
+
+	if _, ok := visiting[tableID]; ok {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
OliverS929: The cycle case is already handled deliberately (returns nil and proceeds), and it can occur in valid schemas. Logging at warning level may be too noisy (especially if cycles are part of a workload or appear repeatedly)

pkg/sqlmodel/causality.go
```diff
@@ -153,6 +155,46 @@ func truncateIndexValues(
...
+			continue
+		}
+
+		relationValues := make([]interface{}, len(relation.ChildColumnIdx))
+		skip := false
+		for i, idx := range relation.ChildColumnIdx {
+			if idx >= len(values) {
+				skip = true
+				break
+			}
```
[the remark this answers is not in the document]
OliverS929: The FK causality string builder skips relations when idx >= len(values) or when any child value is nil. This is defensive behavior to avoid panics and to avoid generating invalid keys. Logging could help diagnose schema/data mismatches, but it can also be noisy because this path is inside per-row processing. I would not elevate this to warning-level logging.

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +523,143 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+			if err != nil {
+				return nil, err
+			}
+			relations = append(relations, sqlmodel.ForeignKeyCausalityRelation{
+				ParentTable:    parentTableName,
+				ParentColumns:  parentColumns,
+				ChildColumnIdx: childIdxs,
+			})
+			continue
+		}
```
GMHDBJD: If the parent has unrelated FKs and none map via fk.RefCols, could we end up missing the direct child→parent causality key and hit FK constraint errors under multi-worker with foreign_key_checks=1?

OliverS929: Great catch! Thank you for pointing it out. I have added fallback to direct FK causality when transitive mapping misses in https://github.com/pingcap/tiflow/pull/12414/commits/df0d94668a28574e8ad5abcf9aeecca5537b13de.

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +523,153 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
 	return ti, nil
 }
 
+func (tr *Tracker) buildForeignKeyRelations(
+	tctx *tcontext.Context,
+	tableID string,
+	downstreamTI *model.TableInfo,
+	originTI *model.TableInfo,
+	cache map[string][]sqlmodel.ForeignKeyCausalityRelation,
```
lance6716, line 577: Can `cache` become a field of `Tracker`? In the case that GetDownStreamTableInfo(t1) t1 -> t0 and GetDownStreamTableInfo(t2) t2 -> t0, t0 can be cached for the second call.

But when to GC them is a problem 🤔

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +523,153 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+		childIdxs := make([]int, 0, len(fk.Cols))
+		for _, col := range fk.Cols {
+			idx, ok := childNameToIdx[col.L]
+			if !ok {
+				return nil, dmterror.ErrSchemaTrackerCannotFetchDownstreamCreateTableStmt.Generatef("column %s not found in table %s", col.L, tableID)
+			}
+			childIdxs = append(childIdxs, idx)
+		}
+
+		parentSchema := fk.RefSchema.O
```
GMHDBJD: Should we handle schema name if user use route feature?

pkg/sqlmodel/foreign_key.go
```diff
@@ -0,0 +1,27 @@
...
+	timodel "github.com/pingcap/tidb/pkg/meta/model"
+)
+
+// ForeignKeyCausalityRelation describes how a row's values relate to a root parent table for causality checks.
+// ChildColumnIdx indexes refer to the row values of the current table (excluding hidden columns),
+// while ParentColumns and ParentTable describe the ultimate parent key the row should collide with.
+type ForeignKeyCausalityRelation struct {
+	ParentTable    string
+	ParentColumns  []*timodel.ColumnInfo
+	ChildColumnIdx []int
```
lance6716, line 28: I see that `Hidden` columns are skipped during counting index. Can there be a `Hidden` in the middle of column list and the index is still legal? Please add a test for this case.

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +523,153 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+	}
+	visiting[tableID] = struct{}{}
+	defer delete(visiting, tableID)
+
+	childNameToIdx := buildColumnIndexMap(originTI)
+
+	relations := make([]sqlmodel.ForeignKeyCausalityRelation, 0, len(downstreamTI.ForeignKeys))
+	for _, fk := range downstreamTI.ForeignKeys {
+		if len(fk.Cols) == 0 || len(fk.Cols) != len(fk.RefCols) {
+			continue
```
lance6716, line 617: we can use DPanic log level to reveal unexpected behaviours in debug level (which mostly in our integration tests)

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +523,153 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+		}
+
+		childIdxs := make([]int, 0, len(fk.Cols))
+		for _, col := range fk.Cols {
+			idx, ok := childNameToIdx[col.L]
+			if !ok {
+				return nil, dmterror.ErrSchemaTrackerCannotFetchDownstreamCreateTableStmt.Generatef("column %s not found in table %s", col.L, tableID)
```
lance6716: Can you check how to use dmctl to fix it? Somewhat user meet this problem and should have a way to continue replication

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +523,153 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+		appendDirectRelation := func() {
+			relations = append(relations, sqlmodel.ForeignKeyCausalityRelation{
+				ParentTable:    parentTableName,
+				ParentColumns:  parentColumns,
+				ChildColumnIdx: childIdxs,
+			})
+		}
+
+		parentNameToIdx := buildColumnIndexMap(parentOriginTI)
+		parentRelations, err := tr.buildForeignKeyRelations(tctx, parentTableID, parentDTI.TableInfo, parentOriginTI, cache, visiting)
```
lance6716: please add comments why we need this recursion.

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +523,153 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+			return nil, err
+		}
+
+		if len(parentRelations) == 0 {
+			appendDirectRelation()
+			continue
+		}
+
+		parentIndexToChild := make(map[int]int, len(fk.RefCols))
+		for i, refCol := range fk.RefCols {
```
lance6716, line 714: Hard to understand 😂 Please add more comments

dm/pkg/schema/tracker_test.go
```diff
@@ -620,6 +621,112 @@ func TestGetDownStreamIndexInfo(t *testing.T) {
...
+			AddRow("a", "create table a(id int primary key)"))
+
+	originTI, err := tracker.GetTableInfo(&filter.Table{Schema: "db", Name: "c"})
+	require.NoError(t, err)
+	dti, err := tracker.GetDownStreamTableInfo(tcontext.Background(), tableIDC, originTI)
+	require.NoError(t, err)
+	require.Len(t, dti.ForeignKeyRelations, 1)
+
+	relation := dti.ForeignKeyRelations[0]
+	require.Equal(t, (&cdcmodel.TableName{Schema: "db", Table: "a"}).String(), relation.ParentTable)
```
lance6716, line 669: Oh seems the `relation.ParentTable` can mean grand parent table 🤔 maybe a better field name for it.

dm/syncer/ddl.go
```diff
@@ -1295,6 +1297,9 @@ func (ddl *DDLWorker) genDDLInfo(qec *queryEventContext, sql string) (*ddlInfo,
 	if err != nil {
 		return nil, terror.Annotatef(terror.ErrSyncerUnitParseStmt.New(err.Error()), "ddl %s", sql)
 	}
+	if err := ddl.rejectForeignKeyDDL(stmt); err != nil {
```
lance6716, line 1300: if `foreign_key_checks` is on, why we reject all related DDLs?

dm/pkg/schema/tracker.go
OliverS929: I agree there is some potential benefit here: a tracker-scoped cache could avoid recomputing shared ancestor FK relations across separate GetDownStreamTableInfo calls.

That said, the gain is limited to cases where different top-level tables share the same FK subgraph, and FK relation building happens once per table initialization rather than on the per-row hot path, so the overall benefit is relatively limited.

Given that, I would prefer to keep the cache local for now. Moving it to Tracker would require us to solve cache invalidation, dependency propagation, and concurrency protection. At the moment we only remove downstream table infos, and we do not have a mechanism to transitively invalidate descendant relation caches yet. This feels more like a follow-up optimization that we can evaluate separately later.

dm/pkg/schema/tracker.go
OliverS929: Good point.

The current scope of this PR is narrower: it builds FK causality from downstream table metadata and targets the existing use case where downstream schema is already prepared and upstream/downstream table identities are effectively aligned. Bridging upstream and downstream identities, especially under route, is not handled in this PR. This is also one of the reasons why we still reject creating new FK constraints here: full support requires a more consistent model across schema tracking, DDL handling, and causality key generation.

We plan to address that in the follow-up work for FK-related DDL support. The expectation there is to connect the upstream and downstream tracker views more consistently, and make FK causality and the existing PK/UK causality follow the same identity model instead of relying on the current mixed source/target assumptions.

That said, we should not let route fail silently. I’m finishing a follow-up fix for this.

The plan there is:
1. Add a minimal fail-fast check: when `foreign_key_checks=1`, `worker_count>1`, and route changes the schema/table identity for tables involved in FK causality, we will return an explicit unsupported error instead of proceeding with the current mixed source/target assumptions.
2. Add a precheck so this unsupported combination can be detected earlier, rather than surfacing only when FK relations are initialized at runtime.
3. Preserve `worker_count=1` as an escape hatch. Before FK causality support, users could still run with `foreign_key_checks=1` by serializing replication with a single worker, even with route. We should keep that path available instead of regressing it. Concretely, the single-worker path will bypass FK causality relation initialization/key generation and rely on the original binlog-order execution semantics.

pkg/sqlmodel/foreign_key.go
OliverS929: Added a test for the hidden-column-in-the-middle case in 9c25db6762.

dm/pkg/schema/tracker.go
OliverS929: Added a debug log for this case in 5a63128ef1. I do not think DPanic is a good fit here: for FK-related usage we may ask users to enable debug logging in advance because the normal info log often is not enough, so DPanic would be too aggressive. This keeps the current tolerant behavior while still surfacing unexpected FK metadata.

dm/pkg/schema/tracker.go
OliverS929: Please refer to [08ffaf5](https://github.com/pingcap/tiflow/pull/12414/commits/08ffaf5c8fe407518ee2e1f4d71331d6db59ec61).

Also cc @alastori

dm/pkg/schema/tracker.go
OliverS929: I agree users should have a way to continue replication, but I do not think there is a single dmctl fix for all cases here.

If the issue is a stale schema tracker / binlog schema, dmctl binlog-schema update (or the old operate-schema set) can help. But for cases like route + FK causality being unsupported, or parent tables being filtered out, dmctl cannot really repair the semantics. In those cases it is better to return a clearer unsupported error or point users to the actual workaround, instead of implying there is a generic one-command fix.

dm/pkg/schema/tracker_test.go
OliverS929: Good point. I did not rename the field in this PR, but I clarified in the comments that ParentTable here can represent the lifted ancestor/root parent domain rather than only the direct parent. See d383222177.

dm/pkg/schema/tracker.go
OliverS929: Added comments around the index-mapping part to explain how the parent relation is projected back to the current child row. See d383222177.

dm/pkg/schema/tracker.go
OliverS929: Added a function-level comment to explain why this walks parent relations recursively and why the result is lifted to the root parent domain. See d383222177.

dm/syncer/ddl.go
OliverS929: This reject is intentional.

The scope of this PR is to handle DML causality for tables whose foreign key metadata is already present on the downstream side before replication starts. It does not support introducing new foreign key constraints on the fly during incremental replication.

So for this PR I would prefer to keep the reject and make that design boundary explicit. Support for FK-related DDL should be handled in the follow-up work for that area.

OliverS929 (commented):
> Do we need to add precheck about block-allow list for FK referenced tables?

@lance6716 Yes, I think we should guard this as well.

I added a precheck for the current supported scope: when oreign_key_checks=1, if a replicated table depends on a parent/ancestor table filtered out by block-allow-list, we now return an explicit unsupported error instead of failing later during relation initialization or surfacing a downstream FK error.  See 07bb19fa29.

dm/pkg/schema/tracker.go
lance6716: There can be complicate actions to fix this error, but I want to double check that it reallly works.

1. The error message should be clear ( maybe "column not found" does not guide the user how to fix it)
2. Use some integration test to check it really works

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +564,272 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+// causality key domain as the root/ultimate parent row, not only its direct parent.
+func (tr *Tracker) buildForeignKeyRelations(
+	tctx *tcontext.Context,
+	tableID string,
+	sourceTable *filter.Table,
+	targetTable *filter.Table,
+	downstreamTI *model.TableInfo,
+	originTI *model.TableInfo,
+	cache map[string][]sqlmodel.ForeignKeyCausalityRelation,
+	visiting map[string]struct{},
```
lance6716, line 578: I'm not sure if we don't need `visiting`. Using `cache` is enough?

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +564,272 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+		}
+
+		parentTableID := utils.GenTableID(targetParentTable)
+		parentTableName := (&cdcmodel.TableName{Schema: targetParentSchema, Table: fk.RefTable.O}).String()
+
+		parentOriginTI, err := tr.GetTableInfo(sourceParentTable)
+		if err != nil {
+			return nil, err
+		}
+		parentDTI, err := tr.downstreamTracker.getOrInit(tctx, parentTableID, parentOriginTI)
```
lance6716, line 675: This function may fail because of some transient error (network problems?). But it can only run once inside `foreignKeyInitOnce`. Maybe better to let it retry on transient error.

Or this is acceptable because the task start/stop or pause/auto-resume acts like an upper layer retry? I'm not sure the `Tracker` will be reset on auto-resume. Maybe add an integration test for the case that network error happens inside `foreignKeyInitOnce`.

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +564,272 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+		if mappedCount == 0 {
+			appendDirectRelation()
+		}
+	}
+
+	cache[tableID] = relations
+	return relations, nil
+}
+
+func findMatchingForeignKey(originTI *model.TableInfo, downstreamFK *model.FKInfo, idx int) *model.FKInfo {
```
lance6716, line 761: better to add comments to give a summary of this function: find the FK name, then assuming the index is not changed and double check FK columns, then find by FK columns

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +564,272 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+
+		targetParentSchema := fk.RefSchema.O
+		if targetParentSchema == "" {
+			targetParentSchema = targetTable.Schema
+		}
+		targetParentTable := &filter.Table{Schema: targetParentSchema, Name: fk.RefTable.O}
+		if !sameTableIdentity(sourceParentTable, targetParentTable) {
+			return nil, newForeignKeyRouteUnsupportedError(
+				fmt.Sprintf(
+					"foreign key causality with route under foreign_key_checks=1 and worker_count>1 is not supported yet; upstream parent table %s and downstream parent table %s do not align for FK %s on table %s; please use worker_count=1",
```
lance6716, line 659: ".. with route ..."

Maybe it's not caused by table routing, just upstream and downstream have different table structure

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +564,272 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+		}
+		parentColumns, err := getColumnsByNames(parentDTI.TableInfo, fk.RefCols)
+		if err != nil {
+			return nil, err
+		}
+		appendDirectRelation := func() {
+			relations = append(relations, sqlmodel.ForeignKeyCausalityRelation{
+				ParentTable:    parentTableName,
+				ParentColumns:  parentColumns,
+				ChildColumnIdx: childIdxs,
```
lance6716, line 687: maybe move the creation of `childIdxs` closer to its usage. During my human review I have forgotten what's the meaning of this variable 😂

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +564,272 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+		if err != nil {
+			return nil, err
+		}
+
+		if len(parentRelations) == 0 {
+			appendDirectRelation()
+			continue
+		}
+
+		// Map the direct parent column positions referenced by the current FK back to
```
lance6716, line 711: (please let AI review this part line 706~739, I don't have a clear mind today 😂 )

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +564,272 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+		}
+		nameToIdx[col.Name.L] = valueIdx
+		valueIdx++
+	}
+	return nameToIdx
+}
+
+func getColumnsByNames(ti *model.TableInfo, names []ast.CIStr) ([]*model.ColumnInfo, error) {
+	columns := make([]*model.ColumnInfo, 0, len(names))
+	for _, name := range names {
```
lance6716, line 834: nit: if `names` are many, building a map of `ti.Columns.Name.L` can be better.

pkg/sqlmodel/causality.go
```diff
@@ -153,6 +155,46 @@ func truncateIndexValues(
...
+
+			relationValues[i] = values[idx]
+			if relationValues[i] == nil {
+				skip = true
+				break
+			}
+		}
+
+		if skip {
+			continue
```
lance6716, line 191: also add debug log on this case?

lance6716 (commented):
rest lgtm

dm/pkg/schema/tracker.go
OliverS929: I updated the error message to make it clearer that this is a FK-causality schema mismatch issue, and to point users to schema repair (`dmctl binlog-schema update` / `operate-schema set`) instead of only saying `column not found`.

I also added coverage for this repair flow in `TestForeignKeyRelationSchemaAlignmentErrorGuidesRepair` and the new DM integration case `foreign_key_schema_repair`, which breaks the tracked child schema, repairs it with `binlog-schema update --from-target`, and verifies the task can resume and continue syncing. See dccc87d78f.

dm/pkg/schema/tracker.go
OliverS929: We still need `visiting` here. `cache` only memoizes fully-built results, while `visiting` protects the current DFS path before a table is written into `cache`, so it is what breaks FK cycles like `A -> B -> A`.

dm/pkg/schema/tracker.go
OliverS929: `foreignKeyInitOnce` itself does not retry. But pause/resume and auto-resume go through `schemaTracker.Init` again and rebuild the downstream tracker state, so the current retry point is at the task level rather than inside `foreignKeyInitOnce`.

Given that, I would prefer to keep the current boundary in this PR instead of adding a more complex retry path here. If transient failures around this path turn out to be a real issue later, we can evaluate a more fine-grained retry design separately.

dm/pkg/schema/tracker.go
OliverS929: Added a short function-level summary to explain the matching order. See ada77cb48a.

dm/pkg/schema/tracker.go
OliverS929: When I added this error originally I was mainly handling the unsupported route case. I’ve adjusted the wording so it no longer implies route is the only cause here. See ada77cb48a.

dm/pkg/schema/tracker.go
OliverS929: Added a local comment to clarify what `childIdxs` means in the rest of this block, while keeping the logic unchanged. See 6f440a50ba.

dm/pkg/schema/tracker.go
OliverS929: I rechecked this part and I do not see a new correctness issue here. The logic is to project the already-lifted parent relation back onto the current child row, and fall back to the direct child->parent relation when that mapping is not possible.

I also added comments around this part to make the flow easier to follow. See 6f440a50ba.

dm/pkg/schema/tracker.go
OliverS929: That makes sense as a micro-optimization, but I’d prefer to keep the simpler implementation here for now. The FK column list is usually very small and this path only runs during relation initialization, not on the per-row hot path. If this later shows up as a real hotspot or bottleneck, we can optimize it more directly then.

pkg/sqlmodel/causality.go
OliverS929: I added a debug log for the out-of-range case only. NULL FK values are still skipped silently, since that path is expected and would be too noisy otherwise. See d36648512d.

dm/pkg/schema/tracker.go
```diff
@@ -497,6 +564,293 @@ func (dt *downstreamTracker) getTableInfoByCreateStmt(tctx *tcontext.Context, ta
...
+}
+
+func newForeignKeyRouteUnsupportedError(msg string) error {
+	return dmterror.ErrSyncerUnitNotSupportedOperate.Generatef(msg)
+}
+
+func newForeignKeySchemaAlignmentError(tableID string, detail string) error {
+	return dmterror.ErrSchemaTrackerCannotFetchDownstreamCreateTableStmt.Generatef(
+		"foreign key causality initialization failed for table %s: %s; this usually means the schema metadata used for FK causality are out of sync (for example, the tracked schema and downstream table schema differ, or the schema was repaired only partially). Please align the schema first; if the tracked schema is stale, use `dmctl binlog-schema update` (or the old `operate-schema set`) and then resume the task",
+		tableID,
```
GMHDBJD: How about use table name instead of table id, which is easy to investigate

OliverS929: Here `tableID` is actually already the schema-qualified table name (`utils.GenTableID` returns `table.String()`), not the numeric ID. I kept the existing tracker naming convention, but renamed this local helper parameter to `tableName` to make the error path a bit clearer. See a1993516ef.

PR #12415, opened 2025-11-25
DM: Check if binlog_legacy_event_pos is set to ON on MariaDB

### LINKED ISSUE #12416: `binlog_legacy_event_pos` (new in 11.4)
This variable should be checked for the right value

- https://mariadb.com/docs/server/ha-and-performance/standard-replication/replication-and-binary-log-system-variables#binlog_legacy_event_pos
- https://github.com/pingcap/docs/pull/22087
- https://github.com/go-mysql-org/go-mysql/pull/1052

dveeden:

Issue Number: close #12416 

#### Tests 

### Release note 

```release-note
A check for the correct value of `binlog_legacy_event_pos` on MariaDB was added.
```


dveeden:
With:
```
podman run --rm --name mariadb -e MARIADB_ALLOW_EMPTY_ROOT_PASSWORD=1 -p 3306:3306 mariadb:11.8 --log-bin=mariadb-log --binlog-format=ROW
```

```
» start-task /home/dvaneeden/task-mariadb.yaml
{
    "result": false,
    "msg": "",
    "sources": [
    ],
    "checkResult": "[code=26005:class=dm-master:scope=internal:level=medium], Message: fail to check synchronization configuration with type: check was failed, please see detail
    	detail: {
		"results": [
			{
				"id": 3,
				"name": "mysql_version",
				"desc": "check whether mysql version is satisfied",
				"state": "warn",
				"errors": [
					{
						"severity": "warn",
						"short_error": "Migrating from MariaDB is still experimental."
					}
				],
				"instruction": "It is recommended that you upgrade MariaDB to 10.1.2 or a later version.",
				"extra": "address of db instance - 127.0.0.1:3306"
			},
			{
				"id": 12,
				"name": "mariadb_binlog_legacy_event_pos",
				"desc": "check whether mariadb binlog_legacy_event_pos is ON",
				"state": "fail",
				"errors": [
					{
						"severity": "fail",
						"short_error": "binlog_legacy_event_pos is OFF, and should be ON"
					}
				],
				"instruction": "MariaDB 11.4 and newer as source: please execute 'SET GLOBAL binlog_legacy_event_pos=ON;' or update the configuration and apply the configuration",
				"extra": "address of db instance - 127.0.0.1:3306"
			}
		],
		"summary": {
			"passed": false,
			"total": 14,
			"successful": 12,
			"failed": 1,
			"warning": 1
		}
	}"
}
»  
```

lance6716:
Maybe we can upgrade go-mysql to use https://github.com/go-mysql-org/go-mysql/pull/1053, instead of adding a user side requirement?

dveeden:
> Maybe we can upgrade go-mysql to use [go-mysql-org/go-mysql#1053](https://github.com/go-mysql-org/go-mysql/pull/1053), instead of adding a user side requirement?

I mentioned 1052 on https://github.com/pingcap/tiflow/issues/12416 which is more or less the same as the 1053 that you mentioned. If we can make that work that would be good. However until that's released and backported to old versions setting `binlog_legacy_event_pos` is needed. I don't see any big drawbacks of this setting, so it is probably ok in almost all cases.

PR #12435, opened 2025-12-03
ticdc: MySQL sink dispatcher routing

### LINKED ISSUE #12434: Schema/table routing support for TiCDC

No

Support routing to a specified destination schema/table in TiCDC. 

_No response_

[\[EXTERNAL\] RFC - Sink routing in TiCDC.pdf](https://github.com/user-attachments/files/23911799/EXTERNAL.RFC.-.Sink.routing.in.TiCDC.pdf)

a-cong:

Issue Number: close #12434

- Extended the `dispatchers` field (currently used for mq sinks https://docs.pingcap.com/tidb/stable/ticdc-sink-to-kafka/#topic-dispatchers) to support schema/table routing for mysql sinks. The config looks like this:
```
[[sink.dispatchers]]
matcher = ["source_db1.*"]
schema = "dest_db1"
table = "{table}"

[[sink.dispatchers]]
matcher = ["source_db2.*"]
schema = "source_db2"
table = "{table}"
```
- The `Dispatcher` interface routes the given `sourceSchema, sourceTable` to `targetSchema, targetTable`. `SinkRouter` wraps the dispatchers
- Added `TargetSchema` and `TargetTable` fields to `TableName` struct in cdc/model/sink.go, which are populated when the sink router is defined. This is handled in 2 ways - for tables that already exist when the changefeed is created/started, the fields are added to the existing snapshot of the schema. For tables that are created later, the fields are added to the new table before the new snapshot is saved
- Use `TargetSchema` and `TargetTable` for the USE statement in `mysql_ddl_sink.go`, and in new method `QuoteSinkString()` which is called to construct mysql DMLs in `pkg/sqlmodel/multirow.go` and `pkg/sqlmodel/row_change.go`
- Use `FetchDDLTables` and `RenameDDLTable` from the dm parser pkg to rewrite DDLs after they are constructed if the schema router is specified. `FetchDDLTables` returns all of the tables referenced in the DDL in the order that they appear in the DDL. We then use the schema router to map each source schema to its target schema and pass this list of tables to `RenameDDLTable`

#### Tests 

 - Unit test
 - Manual test - tested in our local development environment

No, the additional latency for DDL rewrites should be minimal compared to barrier ts, and there should not be compatibility issues since this is a new feature

Yes, this is a new feature for mysql sinks, will need to update https://docs.pingcap.com/tidb/stable/ticdc-sink-to-mysql/

### Release note 
```
Support schema and table routing for mysql-compatible sinks by extending the `dispatchers` config
```


cdc/model/sink.go
```diff
@@ -1116,6 +1150,76 @@ func (d *DDLEvent) FromJobWithArgs(
...
+	if d.TableInfo != nil {
+		defaultSchema = d.TableInfo.TableName.Schema
+	} else if d.PreTableInfo != nil {
+		defaultSchema = d.PreTableInfo.TableName.Schema
+	}
+
+	sourceTables, err := dmparser.FetchDDLTables(defaultSchema, stmts[0], conn.LCTableNamesSensitive)
+	if err != nil {
+		log.Warn("failed to fetch source tables for sink routing, keeping original query",
+			zap.String("query", d.Query), zap.Error(err))
```
a-cong: do we want to fall back here with the warning or explicitly return an error?

cdc/model/sink.go
```diff
@@ -1116,6 +1150,76 @@ func (d *DDLEvent) FromJobWithArgs(
...
+		d.PreTableInfo.TableName.TargetSchema = routedSchema
+		d.PreTableInfo.TableName.TargetTable = routedTable
+	}
+
+	// Use RenameDDLTable to rewrite the query
+	routedQuery, err := dmparser.RenameDDLTable(stmts[0], targetTables)
+	if err != nil {
+		// If rewriting fails, keep the original query
+		log.Warn("failed to rewrite DDL for sink routing, keeping original query",
+			zap.String("query", d.Query), zap.Error(err))
```
a-cong: same question here

wk989898:
Could you provide a case to demonstrate this feature？ It's better to add an integration test to cover it.

cdc/model/sink.go
wk989898: Please use an explicit error to avoid data inconsistency.

a-cong:
> Could you provide a case to demonstrate this feature？ It's better to add an integration test to cover it.

I added an integration test in `sink_routing`, please take a look. I fixed the rest of the tests but pull-dm-integration-test is still failing with a mysql connection error and I'm not sure why - is it related to my changes?

cdc/api/v2/model.go
```diff
@@ -994,14 +1000,18 @@ type LargeMessageHandleConfig struct {
 	ClaimCheckRawValue            bool   `json:"claim_check_raw_value"`
 }
 
-// DispatchRule represents partition rule for a table
+// DispatchRule represents dispatch and routing rules for a table.
+// For MQ sinks (Kafka, Pulsar): controls topic and partition assignment.
+// For MySQL/TiDB sinks: controls schema and table name routing.
```
3AceShowHand, line 1005: Will `schema routing` functionality also works for the MQ Sinks ?

cdc/model/sink.go
```diff
@@ -210,18 +215,42 @@ type TableName struct {
 	Table       string `toml:"tbl-name" msg:"tbl-name"`
 	TableID     int64  `toml:"tbl-id" msg:"tbl-id"`
 	IsPartition bool   `toml:"is-partition" msg:"is-partition"`
+	// TargetSchema is the routed schema name for sinks.
```
3AceShowHand, line 218: I am wondering whether this is good place to put this 2 fields.

TableName is not a good data structure, since it's ambiguity, and exposed to the open api by the query operation.

I would suggest add a new data structure to represent the target information, and set it as a pointer field into the tableInfo directly.

cdc/model/sink.go
```diff
@@ -210,18 +215,42 @@ type TableName struct {
...
+// QuoteString returns quoted full table name using the source schema.
+// This is used for logging and debugging to show the original table name.
 func (t TableName) QuoteString() string {
 	return quotes.QuoteSchema(t.Schema, t.Table)
 }
 
+// QuoteSinkString returns quoted full table name for SQL generation.
+// Uses TargetSchema if set (for sink routing), otherwise uses Schema.
+// This should be used when generating SQL statements for the downstream sink.
+func (t TableName) QuoteSinkString() string {
```
3AceShowHand, line 242: This method looks unnecessary, just modify the QuoteString method should be OK.

pkg/config/replica_config.go
```diff
@@ -323,6 +323,26 @@ func (c *ReplicaConfig) ValidateAndAdjust(sinkURI *url.URL) error { // check sin
...
+	// Validate sink routing configuration (dispatcher approach)
+	// Check if any dispatch rules have schema/table routing configured
+	for _, rule := range c.Sink.DispatchRules {
+		if rule.SchemaRule == "" && rule.TableRule == "" {
+			continue
+		}
+		// Schema routing is only supported for MySQL/TiDB sinks
+		if !sink.IsMySQLCompatibleScheme(sinkURI.Scheme) {
+			return cerror.ErrInvalidReplicaConfig.GenWithStackByArgs(
+				"sink routing (via dispatch rules) is only supported for MySQL/TiDB sinks")
```
3AceShowHand, line 335: got it, LGTM.

3AceShowHand (approved):

cdc/model/sink.go
a-cong: I kept QuoteString separate to use it in a few places for logging/debugging - it's used in the processor to log the "source" table, I found this helpful while testing

cdc/model/sink.go
a-cong: Got it, just to clarify, I saw this comment was marked as resolved, was that intentional?

cdc/model/sink.go
```diff
@@ -1120,6 +1154,72 @@ func (d *DDLEvent) FromJobWithArgs(
...
+	var defaultSchema string
+	if d.TableInfo != nil {
+		defaultSchema = d.TableInfo.TableName.Schema
+	} else if d.PreTableInfo != nil {
+		defaultSchema = d.PreTableInfo.TableName.Schema
+	}
+
+	sourceTables, err := dmparser.FetchDDLTables(defaultSchema, stmts[0], conn.LCTableNamesSensitive)
+	if err != nil {
+		return errors.Annotate(err, "failed to fetch source tables for sink routing")
```
3AceShowHand: `errors.Annotate` is not a standard way to handle error in the tiflow repository.

It's recommend to use the cerror.WrapError to wrap the thirdpart error, and print necessary logs here. Also for other error handling cases.

tests/integration_tests/sink_routing/data/test.sql
```diff
@@ -0,0 +1,27 @@
+-- Test DML operations
```
3AceShowHand: It suggested add more DDLs and DMLs mixed queries here, to cover more scenario.

wk989898 (approved):

wk989898:
@benmeadowcroft PTAL

3AceShowHand:
I think this PR should be merged into the master branch first, and then cherry-pick to the release-8.5 branch.

At the moment, it tries to merge into the release-8.5 directly, this is not standard operation.

a-cong:
> I think this PR should be merged into the master branch first, and then cherry-pick to the release-8.5 branch.
> 
> At the moment, it tries to merge into the release-8.5 directly, this is not standard operation.

Oh, I thought we were not planning to merge this since we aren't upstreaming any more features in this repo - we will merge this change internally, but for upstreaming I will just merge this one when it's ready https://github.com/pingcap/ticdc/pull/3704

3AceShowHand:
> > I think this PR should be merged into the master branch first, and then cherry-pick to the release-8.5 branch.
> > At the moment, it tries to merge into the release-8.5 directly, this is not standard operation.
> 
> Oh, I thought we were not planning to merge this since we aren't upstreaming any more features in this repo - we will merge this change internally, but for upstreaming I will just merge this one when it's ready [pingcap/ticdc#3704](https://github.com/pingcap/ticdc/pull/3704)

got it, so let's only review the PR in the ticdc repository.

PR #12450, opened 2025-12-07
cdc(ddl): ensure strict ordering for multi-table DDLs after split

### LINKED ISSUE #12449: Fix incorrect execution order for multi-table RENAME TABLE DDLs
### What did you do?

1. create a changefeed
2. create table t3 (id int primary key, val int)
3. create table t1 (id int primary key, val int)
4. create table t2 (id int primary key, val int)
5. rename table t1 to t4, t2 to t1, t3 to t2;

3 rename table DDL was executed in the downstream.

### What did you see instead?

Just rename table t1 to t4 was executed in the downstream.

Currently, when TiCDC encounters a multi-table DDL statement (e.g., RENAME TABLE t1 TO t4, t2 TO t1, t3 TO t2;), it splits this single DDL job into multiple individual DDL events (one for each table rename).

Since these split events originate from the same DDL job, they share the exact same CommitTs. In the current implementation, the ddl_manager sorts DDLs primarily by CommitTs. Because the timestamps are identical, the execution order of these split events becomes non-deterministic or unstable.

### Versions of the cluster

```console
v6.5.12
```

```console
v6.5.12
```

```console
v6.5.12
```

wlwilliamx:

Issue Number: close #12449

This PR addresses an issue where split DDLs from a multi-table `RENAME` statement could be executed out of order downstream because they share the same `CommitTs` and the order of ranging map is non-deterministic.

#### Tests 

 - Unit test
 - Integration test
 - Manual test

None
None
### Release note 

```release-note
Fix the incorrect execution order of split DDLs generated from a multi-table DDL statement (e.g., RENAME TABLE).
```


asddongmen (approved):

tests/integration_tests/common_1/data/test.sql
```diff
@@ -175,3 +175,10 @@ VALUES (1),
 UPDATE `column_is_null`
 SET t = NULL
 WHERE id = 1;
+
+-- rename tables DDL should keep the order of events
+
+create table rename_t3 (id int primary key, val int);
+create table rename_t1 (id int primary key, val int);
+create table rename_t2 (id int primary key, val int);
+rename table rename_t1 to rename_t4, rename_t2 to rename_t1, rename_t3 to rename_t2;
```
lidezhu: If rename table doesn't behave as expected, it seems this test case still doesn't report any error. May be we need create different table structures and insert some data after the rename?

wlwilliamx: yes, thanks for reminding

wlwilliamx: done

lidezhu (approved):

PR #12478, opened 2025-12-30
dm: add import into for full load

### LINKED ISSUE #12530: DM: support import-into mode for full load

DM full load currently supports logical and physical loader modes, but does not expose the Lightning `IMPORT INTO` backend as a first-class option.

For large initial loads on TiDB clusters, users may want to use `IMPORT INTO` for better import throughput and smoother resource usage. Without explicit support in DM, users cannot configure this mode through task config/OpenAPI, and unsupported combinations are not rejected early with clear errors.

Support `loader.import-mode: "import-into"` in DM full-load workflow.

- Add `import-into` as a valid `import-mode` in DM config and OpenAPI.
- Run Lightning with import-into backend when this mode is selected.
- Add explicit config checks for unsupported cases:
  - reject sharding (multi-source) tasks in import-into mode
  - require shared storage URI for loader dir (reject local disk path)
- Keep pause/resume behavior consistent with existing DM task lifecycle for this mode.

Use existing `logical` or `physical` modes only. These modes work for many scenarios, but they do not cover the import-into path and do not provide dedicated guardrails for import-into-specific constraints.

GMHDBJD:

Issue Number: close #12530

#### Tests 

 - Unit test
 - Integration test

### Release note 

```release-note

```


joechenrh (approved):

dm/openapi/spec/dm.yaml
```diff
@@ -1715,6 +1715,7 @@ components:
           enum:
             - "logical"
             - "physical"
+            - "import-into"
```
D3Hunter, line 1732: #### ⚠️ **[Major]** OpenAPI import_mode contract omits import-into preconditions and error intent

**Why**
The change adds behavior-level validation for `import-into` that rejects sharding and local storage paths, but the public API schema only adds a new enum value without documenting those constraints.

**Scope**
dm/openapi/spec/dm.yaml:1718

**Risk if unchanged**
API consumers can configure `import_mode=import-into` with unsupported combinations and only discover failures at runtime, increasing integration churn and operator confusion.

**Evidence**
`dm/config/subtask.go:317` and `dm/config/subtask.go:323` enforce `import-into` restrictions, while `dm/openapi/spec/dm.yaml:1714` only describes generic import-mode control and does not state sharding/shared-storage requirements.

**Change request**
Please add explicit `import-into` contract notes in the OpenAPI schema (and related examples) covering: sharding unsupported, shared storage URI required for loader/data dir, and expected validation failure conditions.

dm/config/subtask.go
```diff
@@ -314,6 +314,16 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 		return terror.ErrConfigStrictOptimisticShardMode.Generate()
 	}
 
+	// import-into mode does not support sharding (multi-source) scenario
+	if (c.ShardMode != "" || c.IsSharding) && c.LoaderConfig.ImportMode == LoadModeImportInto {
```
D3Hunter: #### ⚠️ **[Major]** Import-into safety validation can be bypassed by mixed-case import-mode values

**Why**
The new import-into safety checks run before `LoaderConfig.ImportMode` is normalized to lowercase. A mixed-case value such as `Import-Into` or `IMPORT-INTO` skips these exact-string checks, then later gets lowercased and accepted as `import-into`.

**Scope**
dm/config/subtask.go:318, dm/config/subtask.go:323, dm/config/subtask.go:445, dm/config/task.go:343

**Risk if unchanged**
Unsupported configurations (sharding or local loader dir) can pass validation depending on input letter case, then fail later at runtime, creating nondeterministic operator outcomes during rollout and restart flows.

**Evidence**
`SubTaskConfig.Adjust` checks `c.LoaderConfig.ImportMode == LoadModeImportInto` at `dm/config/subtask.go:318` and `dm/config/subtask.go:323` before calling `c.LoaderConfig.adjust()`. Normalization happens later via `strings.ToLower` (`dm/config/subtask.go:445`), so mixed-case inputs bypass the guard and become valid import-into afterward. Existing new tests cover only lowercase `LoadModeImportInto` paths.

**Change request**
Normalize import mode before the import-into guards (or use case-insensitive comparison for those guards), and add mixed-case `import-mode` rejection tests for both sharding and local-dir paths.

dm/worker/source_worker.go
```diff
@@ -230,9 +231,9 @@ func (w *SourceWorker) Stop(graceful bool) {
 
 	// close or kill all subtasks
 	if graceful {
-		w.subTaskHolder.closeAllSubTasks()
+		w.subTaskHolder.closeAllSubTasksWithCause(importinto.ErrFailoverCancel)
```
D3Hunter: #### ⚠️ **[Major]** Failover cancel cause can be surfaced as a runtime error during graceful worker stop

**Why**
Worker shutdown now cancels subtasks with a failover-specific cause, but loader cancellation filtering only treats `context.Canceled` as canceled. If the loader returns the failover cause, shutdown is recorded as an error instead of cancellation.

**Scope**
dm/worker/source_worker.go:234

**Risk if unchanged**
Graceful failover/worker-stop paths can intermittently end in paused-error semantics instead of clean stop semantics, reducing recovery determinism and increasing manual intervention risk.

**Evidence**
`Stop()` calls `closeAllSubTasksWithCause(importinto.ErrFailoverCancel)` and `killAllSubTasksWithCause(importinto.ErrFailoverCancel)` (`dm/worker/source_worker.go:234`, `dm/worker/source_worker.go:236`). Loader-side cancellation filtering still requires root cause `context.Canceled` (`dm/pkg/utils/util.go:143`, `dm/loader/lightning.go:605`). Stage resolution sets `Paused` whenever `result.Errors` is non-empty, even when current stage is `Stopping` (`dm/worker/subtask.go:327`).

**Change request**
Make failover-stop cancellation classify as cancellation end-to-end (for example by recognizing the failover sentinel alongside `context.Canceled`) so `Stopping` paths do not degrade into paused-error outcomes.

dm/worker/source_worker.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/failpoint"
+	"github.com/pingcap/tidb/lightning/pkg/importinto"
```
D3Hunter: #### ⚠️ **[Major]** Source worker stop path now hard-codes lightning import-into failover semantics

**Why**
The generic worker shutdown path now imports lightning's import-into package and always uses a failover-specific cancel cause, which mixes layer responsibilities and conflates normal stop with failover handling.

**Scope**
dm/worker/source_worker.go:26, dm/worker/source_worker.go:234, dm/worker/source_worker.go:236, dm/worker/server.go:489, dm/worker/server.go:588

**Risk if unchanged**
Worker lifecycle behavior becomes coupled to an external loader-internal sentinel error, and non-failover shutdowns can carry misleading cancel semantics that are harder to reason about, validate, and evolve safely.

**Evidence**
`SourceWorker.Stop(graceful bool)` calls `closeAllSubTasksWithCause(importinto.ErrFailoverCancel)` and `killAllSubTasksWithCause(importinto.ErrFailoverCancel)` in both branches, and the file adds a direct `github.com/pingcap/tidb/lightning/pkg/importinto` import solely for that cause. The same stop path is used from general shutdown flows in `Server.doClose` and `stopSourceWorker`, not only failover-specific flows.

**Change request**
Consider a DM-local cancel-cause abstraction: keep default stop paths on `context.Canceled`, and pass a failover-specific cause only from explicit failover call sites so worker lifecycle logic stays decoupled from lightning internals.

dm/worker/subtask.go
```diff
@@ -547,12 +556,17 @@ func (st *SubTask) Result() *pb.ProcessResult {
 
 // Close stops the sub task.
 func (st *SubTask) Close() {
+	st.CloseWithCause(context.Canceled)
+}
+
+// CloseWithCause stops the sub task with a cancel cause.
+func (st *SubTask) CloseWithCause(cause error) {
```
D3Hunter: #### 🟡 **[Minor]** Cause-aware SubTask APIs are exported without external consumers

**Why**
The new methods increase the visible method surface on `SubTask`, but usage in this change remains package-local, so export is not required for current behavior.

**Scope**
dm/worker/subtask.go:563

**Risk if unchanged**
Unnecessary exported symbols harden naming and semantics prematurely and raise long-term maintenance cost for the `worker` package API.

**Evidence**
`CloseWithCause` and `KillWithCause` are introduced at `dm/worker/subtask.go:563` and `dm/worker/subtask.go:583`; call sites in this diff are package-internal (`dm/worker/subtask_holder.go:76`, `dm/worker/subtask_holder.go:91`, plus wrappers/tests).

**Change request**
If no external consumer needs these yet, keep them unexported (`closeWithCause` and `killWithCause`) and preserve `Close` and `Kill` as the external-facing names.

dm/unit/unit.go
```diff
@@ -145,9 +145,17 @@ func IsResumableError(err *pb.ProcessError) bool {
...
-		// treat it as resumable error.
-		return strings.Contains(err.RawCause, "doesn't exist")
+		// for lightning parallel import, some runtime errors are resumable:
+		// - table may be dropped by other lightning: treat "doesn't exist" as resumable
+		// - table is temporarily in Import mode (e.g. "Table t1 is in mode Import") after cancel,
+		//   downstream may switch it back to normal later, so treat it as resumable too.
+		if strings.Contains(err.RawCause, "doesn't exist") {
+			return true
+		}
+		if strings.Contains(err.RawCause, "is in mode Import") {
```
D3Hunter: #### 🟡 **[Minor]** Import-mode resumable error matching is brittle due to case-sensitive raw string check

**Why**
The new resumable path depends on an exact case-sensitive substring in a free-form error message, which is fragile across upstream wording or casing changes.

**Scope**
dm/unit/unit.go:155

**Risk if unchanged**
Auto-resume behavior for failover-related import-mode errors can silently regress across dependency upgrades, causing avoidable paused tasks.

**Evidence**
The new logic checks `strings.Contains(err.RawCause, "is in mode Import")` at `dm/unit/unit.go:155`, while the same function otherwise normalizes text with `strings.ToLower(...)` for other message-based classifications (`dm/unit/unit.go:122`, `dm/unit/unit.go:127`, `dm/unit/unit.go:132`, `dm/unit/unit.go:138`).

**Change request**
Please use a case-insensitive or stable classifier for this condition (preferably code/class based if available) and add unit tests for variant message forms.

D3Hunter (commented):
**AI-generated review based on @D3Hunter's standards; manual review will be performed after all comments are resolved.**

### Summary

- Total findings: 6
- Inline comments: 6

<details>

#### ⚠️ **[Major]** (4)

1. OpenAPI import_mode contract omits import-into preconditions and error intent (`dm/openapi/spec/dm.yaml:1718`)
2. Import-into safety validation can be bypassed by mixed-case import-mode values (`dm/config/subtask.go:318, dm/config/subtask.go:323, dm/config/subtask.go:445, dm/config/task.go:343`)
3. Failover cancel cause can be surfaced as a runtime error during graceful worker stop (`dm/worker/source_worker.go:234`)
4. Source worker stop path now hard-codes lightning import-into failover semantics (`dm/worker/source_worker.go:26, dm/worker/source_worker.go:234, dm/worker/source_worker.go:236, dm/worker/server.go:489, dm/worker/server.go:588`)

#### 🟡 **[Minor]** (2)

1. Cause-aware SubTask APIs are exported without external consumers (`dm/worker/subtask.go:563`)
2. Import-mode resumable error matching is brittle due to case-sensitive raw string check (`dm/unit/unit.go:155`)

</details>

GMHDBJD:
Addressed the review feedback by documenting import-into OpenAPI constraints, making import-mode validation case-insensitive, and refining stop vs failover cancel-cause handling.

dm/worker/server.go
```diff
@@ -585,7 +585,7 @@ func (s *Server) stopSourceWorker(sourceID string, needLock, graceful bool) erro
 	s.UpdateKeepAliveTTL(s.cfg.KeepAliveTTL)
 	s.setWorker(nil, false)
 	s.setSourceStatus("", nil, false)
-	w.Stop(graceful)
+	w.StopForFailover(graceful)
```
D3Hunter, line 588: #### ⚠️ **[Major]** Failover-only cancellation is used on generic worker-stop paths

**Why**
`StopForFailover` is now called from shared stop flows that are also used by explicit unbind/disable operations, not only takeover/rebalance failover.

**Scope**
dm/worker/server.go:588

**Risk if unchanged**
IMPORT INTO jobs may continue when an operator intentionally stops or unbinds a source, creating detached job ownership and confusing stop/cleanup behavior.

**Evidence**
`stopSourceWorker` always calls `w.StopForFailover(graceful)`, and this path is reached from unbind/disable flows (`disableHandleSubtasks`, `disableRelay`) in addition to server close. `WorkerFailoverCause` is documented as takeover-specific.

**Change request**
Please route stop reason explicitly: use normal `Stop` for intentional stop/unbind paths, and keep `StopForFailover` only for verified failover/rebalance takeover paths.

dm/worker/subtask.go
```diff
@@ -562,13 +576,22 @@ func (st *SubTask) Close() {
...
 	st.l.Info("killing")
 	if !st.setStageIfNotIn([]pb.Stage{pb.Stage_Stopped, pb.Stage_Stopping, pb.Stage_Finished}, pb.Stage_Stopping) {
 		st.l.Info("subTask is already closed, no need to close")
 		return
 	}
+	if cause == nil {
+		cause = context.Canceled
+	}
+	// Kill should be ungraceful: kill the running unit first, then cancel the subtask context.
 	st.killCurrentUnit()
```
D3Hunter: #### ⚠️ **[Major]** Ungraceful failover path can overwrite failover cause with `context.Canceled`

**Why**
In the keepalive-loss/ungraceful path, unit kill runs before cause-aware close and loader pause hard-cancels with `context.Canceled`, so the failover sentinel can be lost.

**Scope**
dm/worker/subtask.go:593

**Risk if unchanged**
During failover under load, IMPORT INTO jobs can be canceled instead of preserved for takeover, reducing HA correctness and making recovery behavior inconsistent.

**Evidence**
`KeepAlive` uses `stopSourceWorker(..., graceful=false)`. `killWithCause` executes `killCurrentUnit()` before `closeUnitsWithCause(cause)`. `LightningLoader.Kill()` delegates to `Close()->Pause()` and calls `l.cancel(context.Canceled)`, where first cancel cause wins.

**Change request**
Please preserve the provided failover cause in ungraceful loader kill paths as well, and add an integration test for keepalive-loss failover during import-into load.

dm/config/subtask.go
```diff
@@ -314,6 +314,18 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 		return terror.ErrConfigStrictOptimisticShardMode.Generate()
 	}
 
+	isImportInto := strings.EqualFold(string(c.LoaderConfig.ImportMode), string(LoadModeImportInto))
+
+	// import-into mode does not support sharding (multi-source) scenario
+	if (c.ShardMode != "" || c.IsSharding) && isImportInto {
```
D3Hunter, line 320: #### ⚠️ **[Major]** `import-into` multi-source restriction is documented but not validated

**Why**
The API contract says import-into does not support sharding/multi-source tasks, but current validation only checks sharding flags and can miss multi-source configurations.

**Scope**
dm/config/subtask.go:320

**Risk if unchanged**
Unsupported multi-source import-into tasks may pass admission and fail later at runtime with less actionable errors.

**Evidence**
`SubTaskConfig.Adjust` only checks `(ShardMode != "" || IsSharding)` for import-into. OpenAPI docs describe both sharding and multi-source restrictions, but there is no explicit task-level `len(MySQLInstances) > 1` rejection in this path.

**Change request**
Please add explicit multi-source validation for import-into (for example reject when source instance count > 1) and cover it with config/openapi tests.

dm/config/subtask.go
```diff
@@ -314,6 +314,18 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
...
+		return terror.ErrConfigImportIntoShardingNotSupport.Generate()
+	}
+
+	// import-into mode requires shared storage (s3, gcs, azure, etc.)
+	if isImportInto && storage.IsLocalDiskPath(c.LoaderConfig.Dir) {
```
D3Hunter, line 328: #### 🟡 **[Minor]** Empty `loader.dir` bypasses import-into shared-storage validation

**Why**
The new import-into validation checks local disk path but does not reject empty directory values.

**Scope**
dm/config/subtask.go:325

**Risk if unchanged**
Invalid import-into config can pass adjust-time checks and fail later with weaker diagnostics.

**Evidence**
Guard is `isImportInto && storage.IsLocalDiskPath(c.LoaderConfig.Dir)`. `IsLocalDiskPath` returns false for empty input, so empty `loader.dir` is not rejected here.

**Change request**
Please add an explicit non-empty shared-URI check for import-into mode and cover empty-dir in tests.

dm/unit/unit.go
```diff
@@ -145,9 +145,18 @@ func IsResumableError(err *pb.ProcessError) bool {
...
-		return strings.Contains(err.RawCause, "doesn't exist")
+		// for lightning parallel import, some runtime errors are resumable:
+		// - table may be dropped by other lightning: treat "doesn't exist" as resumable
+		// - table is temporarily in Import mode (e.g. "Table t1 is in mode Import") after cancel,
+		//   downstream may switch it back to normal later, so treat it as resumable too.
+		rawCauseLower := strings.ToLower(err.RawCause)
+		if strings.Contains(rawCauseLower, "doesn't exist") {
+			return true
+		}
+		if strings.Contains(rawCauseLower, "is in mode import") {
```
D3Hunter, line 156: #### 🟡 **[Minor]** Resumable classification for 'is in mode import' is overly broad

**Why**
Any runtime error containing that substring is treated as resumable without a stronger transient/failover signal.

**Scope**
dm/unit/unit.go:156

**Risk if unchanged**
Persistent import-mode issues can enter long auto-resume loops and mask terminal operator-actionable failures.

**Evidence**
`IsResumableError` now returns true on substring match `"is in mode import"`, while task checker auto-resume continues with backoff and no attempt cap.

**Change request**
Please gate this rule with a stronger condition (for example failover marker/checkpoint state) and add a bounded retry-stop policy for persistent failures.

dm/tests/import_into_mode/run.sh
```diff
@@ -0,0 +1,282 @@
...
+	echo "prepare source data"
+	run_sql_file $cur/data/db1.prepare.sql $MYSQL_HOST1 $MYSQL_PORT1 $MYSQL_PASSWORD1
+
+	echo "start task with import-into mode"
+	cp $cur/conf/dm-task.yaml $WORK_DIR/dm-task.yaml
+	sed -i "s#dir: placeholder#dir: $S3_DIR#g" $WORK_DIR/dm-task.yaml
+	dmctl_start_task_standalone $WORK_DIR/dm-task.yaml "--remove-meta"
+
+	# wait until the task enters the Load phase, then kill worker1 to simulate failover
+	run_dm_ctl_with_retry $WORK_DIR "127.0.0.1:$MASTER_PORT" \
```
D3Hunter, line 162: #### 🟡 **[Minor]** HA failover integration test depends on transient `Load` stage observation

**Why**
The test kills worker1 after polling for `unit=Load`, but with small fixtures that stage can be too brief to observe reliably.

**Scope**
dm/tests/import_into_mode/run.sh:152

**Risk if unchanged**
The new coverage can be flaky in CI and fail nondeterministically without a product regression.

**Evidence**
`run_ha_failover_test` waits for `"unit": "Load"` then kills worker1 immediately; the same test data set is very small and can transition quickly.

**Change request**
Please use a deterministic hook/failpoint to hold load state before kill, instead of racing on a transient status field.

D3Hunter (commented):
**AI-generated review based on @D3Hunter's standards; manual review will be performed after all comments are resolved.**

### Summary

- Total findings: 6
- Inline comments: 6

<details>

#### ⚠️ **[Major]** (3)

1. Failover-only cancellation is used on generic worker-stop paths (`dm/worker/server.go:588`)
2. Ungraceful failover path can overwrite failover cause with `context.Canceled` (`dm/worker/subtask.go:593`)
3. `import-into` multi-source restriction is documented but not validated (`dm/config/subtask.go:320`)

#### 🟡 **[Minor]** (3)

1. Empty `loader.dir` bypasses import-into shared-storage validation (`dm/config/subtask.go:325`)
2. Resumable classification for 'is in mode import' is overly broad (`dm/unit/unit.go:156`)
3. HA failover integration test depends on transient `Load` stage observation (`dm/tests/import_into_mode/run.sh:152`)

</details>

dm/worker/server.go
GMHDBJD: Stop-source/remove-source already implies related tasks are stopped (scheduler won’t let you stop a source that still has subtasks/relay). So when we get here there’s no running IMPORT INTO job to “keep alive”.

dm/unit/unit.go
GMHDBJD: This only affects Lightning runtime errors. The “is in mode Import” one is usually temporary after cancel/failover, so treating it as resumable is intentional; if it keeps happening, `stop-task` can break the loop.

dm/worker/subtask.go
GMHDBJD: fixed in 0d3a8c0

dm/config/subtask.go
GMHDBJD: fixed in 0d3a8c0

dm/tests/import_into_mode/run.sh
GMHDBJD: fixed in 0d3a8c0

dm/config/subtask.go
GMHDBJD: fixed in 0d3a8c0

D3Hunter (approved):

Benjamin2037 (approved):
LGTM

yudongusa (approved):

PR #12481, opened 2026-01-07
master(dm): support extra SHOW FULL TABLES columns from PolarDB-X

### LINKED ISSUE #12569: DM: support PolarDB-X SHOW FULL TABLES result with extra columns
### What did you do?

Use DM with a PolarDB-X upstream. DM reaches the table listing path that uses `dbutil.GetTables`.

A reproducible upstream query is:

```sql
SHOW FULL TABLES IN `test` WHERE Table_Type != 'VIEW';
```

On PolarDB-X this returns 4 columns, for example:

```text
Tables_in_test | Table_type | Auto_partition | Table_group
```

Current DM code calls `dbutil.GetTables`, which scans the result as exactly 2 columns, so table discovery fails on this upstream.

DM should keep the existing behavior for normal MySQL/TiDB two-column results, and also accept PolarDB-X results by reading the first two columns and ignoring extra columns.

### What did you see instead?

DM cannot list tables from the upstream because `SHOW FULL TABLES` returns more columns than `dbutil.GetTables` expects.

### Versions of the cluster

```console
master
```
```console
PolarDB-X
```
```console
N/A
```
```console
N/A
```
```console
The incompatibility is in the table listing SQL result shape.
```

```console
N/A
```

GMHDBJD:

Issue Number: close #12569

- add a DM-local `conn.GetTables` helper for table discovery
- keep the original `SHOW FULL TABLES ... WHERE Table_Type != 'VIEW'` behavior for normal MySQL/TiDB results
- ignore extra columns after the first two scan targets so PolarDB-X 4-column results are accepted
- add regression tests for both standard 2-column results and PolarDB-X-style 4-column results

#### Tests

- Unit test: `go test ./dm/pkg/conn -run 'TestGetTables|TestFetchAllDoTables'`
- Build/compile check: `go test ./dm/pkg/checker ./dm/master -run '^$'`

No. Existing behavior for standard upstreams is kept unchanged. The parsing only becomes more permissive for extra result columns.

No.

### Release note

```release-note
Support PolarDB-X SHOW FULL TABLES results with extra columns in DM table discovery.
```


dm/pkg/conn/table.go
```diff
@@ -0,0 +1,74 @@
...
+			continue
+		}
+
+		tables = append(tables, table.String)
+	}
+
+	return tables, errors.Trace(rows.Err())
+}
+
+func escapeName(name string) string {
```
joechenrh, line 72: Not related to this PR, but I found that there are three places in TiDB with this same (unexported) function 😂

joechenrh (approved):

dm/pkg/conn/table.go
```diff
@@ -0,0 +1,74 @@
...
+	"strings"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/tidb/pkg/util/dbutil"
+)
+
+// GetTables returns names of all non-view tables in the specified schema.
+// It is compatible with upstreams like PolarDB-X, whose SHOW FULL TABLES
+// may return extra columns after the standard table name and table type.
+func GetTables(ctx context.Context, db dbutil.QueryExecutor, schemaName string) ([]string, error) {
```
OliverS929, line 29: This looks like a generic `SHOW FULL TABLES` parsing issue rather than a DM-only one. If fixing the shared helper upstream is out of scope here, do we need to check the remaining `dbutil.GetTables` callers (for example in sync-diff) and confirm they don't need the same compatibility fix?

dm/master/openapi_view.go
```diff
@@ -391,7 +391,7 @@ func (s *Server) DMAPIGetSourceTableList(c *gin.Context, sourceName string, sche
 		return
 	}
 	defer baseDB.Close()
-	tableList, err := dbutil.GetTables(c.Request.Context(), baseDB.DB, schemaName)
+	tableList, err := conn.GetTables(c.Request.Context(), baseDB.DB, schemaName)
```
OliverS929, line 394: Could we add one 4-column regression for `OpenAPIViewSuite.TestSourceAPI` and one for the online-DDL checker path too? Right now the new coverage is mostly helper-level.

dm/pkg/conn/db_test.go
```diff
@@ -80,6 +80,34 @@ func TestGetRandomServerID(t *testing.T) {
 	require.NotEqual(t, 101, serverID)
 }
 
+func TestGetTables(t *testing.T) {
```
OliverS929, line 83: Could we add a small negative test or two for `conn.GetTables` as well?

dm/pkg/conn/table.go
GMHDBJD: 6155c52

dm/master/openapi_view.go
GMHDBJD: 6155c52

dm/pkg/conn/db_test.go
GMHDBJD: 6155c52

PR #12490, opened 2026-01-13
sink/mysql: align DDL time defaults with origin_default

### LINKED ISSUE #11368: Product Behavior: Data will inconsistent after TiCDC replicates the DDL as `alter table xx add column xx datetime default current_timestamp`
### What did you do?

Reproduction Steps

1. Create a table t1
    
    ```markdown
    create table t1 (id int primary key);
    
    ```
    
2. Create a changefeed to replicate this table
3. Insert some data into t1
4. Pause changefeed (the purpose is to manually create synchronization delay), assuming the current time is time1.
5. Add a new column to t1
    
    ```markdown
    alter table t1 add column `updatetime` datetime default current_timestamp;
    
    ```
    
6. Resume the changefeed and wait for the changefeed to replicate the above DDL downstream. Assume the current time is time2. 
    
    When comparing the upstream and downstream data, it will be found that the values of the newly added column updatetime upstream and downstream inconsistent. The difference in this column upstream and downstream is `time2 - time1`.
    

Reason: This is because cdc only synchronizes the DDL itself when synchronizing DDLs that will cause destructive changes like add column, and does not synchronize the row changes caused by the DDL. The reason for doing this is to reduce synchronization latency.

Currently, these inconsistent data can only be manually fixed.

_No response_

### What did you see instead?

As described above.

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
All version.
```

haiboumich:
DDL statements that reference CURRENT_TIMESTAMP/NOW/LOCALTIME/LOCALTIMESTAMP can evaluate to different values downstream when executed later or in a different time zone. This can lead to inconsistent default values for newly added columns.

Issue Number: close #11368

- Set session `TIMESTAMP` for each DDL execution using the DDL StartTs (fallback to CommitTs).
- Parse CREATE/ALTER statements to detect CURRENT_TIMESTAMP-style defaults and use the column `origin_default` from TableInfo to set the session timestamp so defaults match upstream evaluation.
- Reset session `TIMESTAMP` after DDL execution and on failures to avoid leaking session state.
- Add unit tests for origin_default-driven timestamps and StartTs fallback; update existing DDL sink tests to expect the session timestamp set/reset.

#### Tests
- Unit test
- Manual test
  ```sql
  -- create changefeed
  CREATE DATABASE IF NOT EXISTS test_time_funcs;
  USE test_time_funcs;
  -- Timestamp & Datetime Functions
  -- Test: NOW(), CURRENT_TIMESTAMP(), CURRENT_TIMESTAMP, LOCALTIME(), LOCALTIME, LOCALTIMESTAMP, LOCALTIMESTAMP(), UTC_TIMESTAMP()
  CREATE TABLE t_timestamp (id INT PRIMARY KEY, info VARCHAR(20));
  INSERT INTO t_timestamp VALUES (1, 'initial_row');
  -- pause the changefeed
  ALTER TABLE t_timestamp ADD COLUMN ts_now DATETIME DEFAULT NOW();
  ALTER TABLE t_timestamp ADD COLUMN ts_current_timestamp_func TIMESTAMP DEFAULT CURRENT_TIMESTAMP();
  ALTER TABLE t_timestamp ADD COLUMN ts_current_timestamp_const TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
  ALTER TABLE t_timestamp ADD COLUMN ts_localtime_func TIMESTAMP DEFAULT LOCALTIME();
  ALTER TABLE t_timestamp ADD COLUMN ts_localtime_const TIMESTAMP DEFAULT LOCALTIME;
  ALTER TABLE t_timestamp ADD COLUMN ts_localtimestamp_func TIMESTAMP DEFAULT LOCALTIMESTAMP();
  ALTER TABLE t_timestamp ADD COLUMN ts_localtimestamp_const TIMESTAMP DEFAULT LOCALTIMESTAMP;
  SELECT * FROM test_time_funcs.t_timestamp\G
  -- wait for seconds & resume the changefeed
  SELECT * FROM test_time_funcs.t_timestamp\G
  ```

No. The change only affects DDL execution paths and sets a session variable around the DDL, then resets it.

No.

### Release note
```release-note
Fix MySQL DDL execution to use upstream-evaluated timestamps for CURRENT_TIMESTAMP-style defaults, ensuring consistent default values downstream.
```

### DIFF
--- a/cdc/sink/ddlsink/mysql/helper.go
+++ b/cdc/sink/ddlsink/mysql/helper.go
@@ -49,7 +49,7 @@ func formatUnixTimestamp(unixTimestamp float64) string {
 }
 
 func ddlSessionTimestampFromOriginDefault(ddl *model.DDLEvent, timezone string) (float64, bool) {
-	if ddl == nil || ddl.TableInfo == nil {
+	if ddl == nil || ddl.TableInfo == nil || ddl.TableInfo.TableInfo == nil {
 		return 0, false
 	}
 	targetColumns, err := extractCurrentTimestampDefaultColumns(ddl.Query)
@@ -58,6 +58,9 @@ func ddlSessionTimestampFromOriginDefault(ddl *model.DDLEvent, timezone string) 
 	}
 
 	for _, col := range ddl.TableInfo.Columns {
+		if col == nil {
+			continue
+		}
 		if _, ok := targetColumns[col.Name.L]; !ok {
 			continue
 		}
--- a/cdc/sink/ddlsink/mysql/mysql_ddl_timestamp_test.go
+++ b/cdc/sink/ddlsink/mysql/mysql_ddl_timestamp_test.go
@@ -122,3 +122,16 @@ func TestExecDDL_DoesNotSetTimestampWhenNoCurrentTimestampDefault(t *testing.T) 
 	require.NoError(t, err)
 	require.NoError(t, mock.ExpectationsWereMet())
 }
+
+func TestDDLSessionTimestampFromOriginDefaultWithNilTableInfo(t *testing.T) {
+	ddlEvent := &model.DDLEvent{
+		Query:     "create table t (id int primary key auto_increment, t datetime default current_timestamp)",
+		TableInfo: &model.TableInfo{},
+	}
+
+	require.NotPanics(t, func() {
+		ts, ok := ddlSessionTimestampFromOriginDefault(ddlEvent, "UTC")
+		require.False(t, ok)
+		require.Equal(t, float64(0), ts)
+	})
+}


wk989898:
Please add an integration test to cover it.

tests/integration_tests/ddl_default_current_timestamp/run.sh
```diff
@@ -0,0 +1,38 @@
...
+	run_cdc_server --workdir $WORK_DIR --binary $CDC_BINARY
+	run_cdc_cli changefeed create --start-ts=$start_ts --sink-uri="mysql://normal:123456@127.0.0.1:3306/"
+
+	run_sql "DROP DATABASE IF EXISTS ${DB_NAME};" ${UP_TIDB_HOST} ${UP_TIDB_PORT}
+	run_sql "CREATE DATABASE ${DB_NAME};" ${UP_TIDB_HOST} ${UP_TIDB_PORT}
+	run_sql "CREATE TABLE ${DB_NAME}.t (id int primary key, v int);" ${UP_TIDB_HOST} ${UP_TIDB_PORT}
+	run_sql "INSERT INTO ${DB_NAME}.t VALUES (1, 10), (2, 20), (3, 30);" ${UP_TIDB_HOST} ${UP_TIDB_PORT}
+
+	run_sql "SET @@timestamp = 1000000000.123456; ALTER TABLE ${DB_NAME}.t ADD COLUMN c DATETIME(6) DEFAULT CURRENT_TIMESTAMP(6);" ${UP_TIDB_HOST} ${UP_TIDB_PORT}
+
```
wk989898, line 37: Could you add more dml before and after adding the column?

haiboumich: thx, i just added some more dml, PTAL

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -211,7 +211,34 @@ func (m *DDLSink) execDDL(pctx context.Context, ddl *model.DDLEvent) error {
...
+				log.Error("Failed to rollback", zap.String("changefeed", m.id.ID), zap.Error(rbErr))
+			}
+			return err
+		}
+	}
+
 	if _, err = tx.ExecContext(ctx, ddl.Query); err != nil {
+		log.Error("Failed to ExecContext", zap.Any("err", err), zap.Any("query", ddl.Query))
+		if useSessionTimestamp {
+			if tsErr := resetSessionTimestamp(ctx, tx); tsErr != nil {
```
wk989898: If resetSessionTimestamp failed, does it affect subsequent ddl?

haiboumich: yes, i think resetSessionTimestamp fails, the session timestamp can leak and affect later DDLs. I [added a pre-DDL SET TIMESTAMP = DEFAULT so every DDL starts from a clean session timestam](https://github.com/pingcap/tiflow/pull/12490/commits/29d6e56c4410cb24ebaf157d779d00faea25494e)

wk989898: So the resetSessionTimestamp is duplicated after ddl is executed.

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
```diff
@@ -211,7 +211,46 @@ func (m *DDLSink) execDDL(pctx context.Context, ddl *model.DDLEvent) error {
 		return err
 	}
 
+	// Reset session timestamp before DDL to avoid leaking from pooled connections.
+	if err := resetSessionTimestamp(ctx, tx); err != nil {
```
wk989898, line 220: Please add useSessionTimestamp check.

haiboumich: I think it might not be necessary?
The pre-DDL SET TIMESTAMP = DEFAULT is a defensive reset for pooled connections and protects against leaked session state from prior failures.
And useSessionTimestamp only reflects the current DDL’s needs, coupling them could leave leaks unaddressed.

wk989898: Ok, I think we can remove resetSessionTimestamp after the current DDL executes. Because every DDL will reset the sessionTimestamp.

wk989898: It looks like set the time zone at the session level, so the resetSessionTimestamp is not necessary?

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
haiboumich: yes, i think it is just a safeguard with low cost

cdc/sink/ddlsink/mysql/mysql_ddl_sink.go
haiboumich: I think the resetSessionTimestamp might still be necessary and the pre-ddl one could be a safety net for any case where the previous DDL didn’t reset, and I think the cost is cheap as DDLs are low‑frequency and extra round‑trip is cost is negligible compared to the general metadata locks, schema change.

I did a failpiont test, if there is no reset timestamp, it may still leak to the following ddl  added a debug failpoint test PR: https://github.com/pingcap/tiflow/pull/12496, which I did the following thing:
1. force the same session to be reused: MySQLSinkForceSingleConnection sets MaxOpenConns=1 / MaxIdleConns=1, so DDL1 and DDL2 run on the same session.
2. DDL1 (c2) leaves the session timestamp stale.
      - The sink normally does: SET TIMESTAMP = <ddl1_ts> → execute DDL → SET TIMESTAMP = DEFAULT.
      - The failpoint MySQLSinkSkipResetSessionTimestampAfterDDL=return("c2") skips the post‑DDL reset for the DDL that contains c2.
      - Result: after DDL1, the session still has the old timestamp (t1).
3. DDL2 (d2) deliberately skips setting a new timestamp.
      - The failpoint MySQLSinkSkipSetSessionTimestamp=return("d2") skips SET TIMESTAMP for DDL2.
      - So DDL2 does not overwrite the session timestamp, and if nothing clears it, it will use whatever is still in the session.
      
      
For the test scenario above, I run twice for wihout and with pre-ddl reset timestamp, and I got:
1. If there is no pre-DDL reset timestamp, I found that the following ddl will use the former leak timestamp, we can see from the job: https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflow%2Fpull_cdc_integration_test/detail/pull_cdc_integration_test/4337/pipeline/523: "downstream d2 should not reuse stale timestamp: expected != 2001-09-09 01:46:40.123456, got 2001-09-09 01:46:40.123456"
2. If there is the pre-DDL reset timestamp, the test pass: https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflow%2Fpull_cdc_integration_test/detail/pull_cdc_integration_test/4339/pipeline/477

PR #12559, opened 2026-03-13
test: stabilize sync_diff_inspector checkpoint test

### LINKED ISSUE #12553: sync_diff_inspector checkpoint integration test flakes on resumed chunk selection
### Which jobs are flaking?

- `pingcap/tiflow/pull_syncdiff_integration_test`

- `sync_diff_inspector/tests/sync_diff_inspector/checkpoint/run.sh`
- The flaky assertion is in the bucket-checkpoint resume path, where the script picks the first resumed chunk by sorting `lowerBounds` only.

```log
https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflow%2Fpull_syncdiff_integration_test/detail/pull_syncdiff_integration_test/815/pipeline

Observed failure from build 815:
- resumed candidate logged as `39 upperBounds= indexCode=0:1-250:0:1`
- old assertion expected a fixed resumed index pattern and failed on `first_chunk_index`
```

  - Likely yes, if the same checkpoint test script is present.
  - The same symptom has been seen in other PRs.
- Related parent tracking issue:
  - Subtask of #2246
- Root cause summary:
  - The checkpoint test assumes the first resumed chunk selected from logs is stable after sorting by `lowerBounds` only. In practice, multiple resumed chunks can share the same lower bound, so the script may pick a different valid resumed chunk and fail nondeterministically.

joechenrh:

Issue Number: close #12553

The checkpoint integration test (`checkpoint/run.sh`) was flaky because it pre-computed a `bucket_index_right` (or `bucket_index_left-bucket_index_right`) from the *last* run and then used `check_contains_regex` to match a full `indexCode` pattern in the *resumed* run's first chunk. When chunks are created in parallel, multiple resumed chunks can share the same lower bound, so the selected first chunk may carry a different—but equally valid—`indexCode`, causing the regex match to fail nondeterministically.

**Changes (bucket splitter, cases 1 & 2):**
- Removed the pre-computed `bucket_index_right`/`bucket_index_left` variables that were derived from the previous run's index.
- Instead of matching the entire `indexCode` string with a regex, we now parse the resumed first chunk's `indexCode` into an array and assert only the structurally stable fields:
  - **Case 1 (last chunk in bucket):** after resuming, the next chunk must start a new bucket, so we assert `chunkIndex == 0`.
  - **Case 2 (middle of bucket):** the next chunk must be the immediate successor within the same bucket, so we assert `chunkIndex == lastChunkIndex + 1` and `chunkCount == lastChunkCount`.

**Changes (random splitter, case 1):**
- Removed the `check_contains_regex` assertion on `indexCode` entirely, because the random splitter may re-split the remaining range after resuming, making the `indexCode` unstable across runs.
- We still assert that the resumed chunk's *bound* matches the last chunk's upper bound, which is the meaningful invariant.

#### Tests 

 - Integration test

No. This is a test-only change.

No.

### Release note 

```release-note
None
```

### DIFF
--- a/sync_diff_inspector/tests/sync_diff_inspector/checkpoint/run.sh
+++ b/sync_diff_inspector/tests/sync_diff_inspector/checkpoint/run.sh
@@ -36,10 +36,6 @@ for s in ${last_chunk_index_array[@]}; do
 done
 # chunkIndex should be the last Index
 [[ $((${last_chunk_index_array[2]} + 1)) -eq ${last_chunk_index_array[3]} ]] || exit 1
-# Save bucketIndexRight, which should be equal to bucketIndexLeft of the chunk first created in the next running.
-bucket_index_right=$(($(echo ${last_chunk_index_array[1]} | awk -F '-' '{print $2}') + 1))
-echo $bucket_index_right
-
 rm -f $OUT_DIR/sync_diff.log
 export GO_FAILPOINTS="github.com/pingcap/tiflow/sync_diff_inspector/splitter/print-chunk-info=return()"
 sync_diff_inspector --config=./config.toml >$OUT_DIR/checkpoint_diff.output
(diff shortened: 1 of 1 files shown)


joechenrh:
Makes sense, simplified both blocks in https://github.com/joechenrh/tiflow/commit/9b0ed66d38e422b055f715a1ee3a074e99fb47bc . Replaced with a single `IFS=: read -r -a` and dropped the debug loops.

PR #12563, opened 2026-03-16
master(dm): avoid slow OpenAPI delete on unavailable downstream

### LINKED ISSUE #12562: DM/OpenAPI: delete task is slow when downstream is unreachable
### What did you do?

Reproduced with DM OpenAPI:

1. Create a task successfully.
2. Make the downstream unavailable while keeping the target port reachable.
3. Call `DELETE /api/v1/tasks/{task-name}`.

`DELETE /api/v1/tasks/{task-name}` should return quickly. Downstream metadata cleanup should not block task deletion for a long time when the downstream is unavailable.

### What did you see instead?

The request blocks for about 30 seconds and then returns a downstream connection timeout error such as `context deadline exceeded`.

### Versions of the cluster

```console
Reproduced on current master branch.
```

```console
MySQL 5.7.32-log / 8.0.23
```

```console
v9.0.0-beta.2.pre-1098-g7607607
```

```console
manually, with dm/tests/openapi reproduction
```

```console
The reproduced slow path is the downstream connection timeout path in OpenAPI delete.
```

GMHDBJD:

Issue Number: close #12562

- use a 10s downstream connection timeout for OpenAPI delete
- let task deletion continue when downstream metadata cleanup fails
- log skipped downstream metadata cleanup instead of failing the delete request
- simplify the related OpenAPI integration case to use failpoint

#### Tests

- Unit test: `make dm_unit_test_pkg PKG=github.com/pingcap/tiflow/dm/master`
- Integration test: `tests/run.sh openapi` (currently blocked by an unrelated existing failure in `test_multi_tasks`: `ERROR 1236 (HY000): Client requested master to start replication from position > file size`)

No. It only changes the OpenAPI delete path when downstream is unavailable.

No.

### Release note

```release-note
Fix slow DM OpenAPI task deletion when the downstream is unavailable.
```

### DIFF
--- a/dm/master/openapi_controller.go
+++ b/dm/master/openapi_controller.go
@@ -40,7 +40,7 @@ import (
 	"go.uber.org/zap"
 )
 
-const openAPIDeleteTaskDownstreamTimeout = 10 * time.Second
+var openAPIDeleteTaskDownstreamTimeout = 10 * time.Second
 
 // nolint:unparam
 func (s *Server) getClusterInfo(ctx context.Context) (*openapi.GetClusterInfoResponse, error) {
@@ -475,9 +475,12 @@ func (s *Server) deleteTask(ctx context.Context, taskName string, force bool) er
 	}
 
 	toDBCfg := config.GetTargetDBCfgFromOpenAPITask(task)
-	if adjustErr := AdjustTargetDBSessionCfgWithTimeout(ctx, toDBCfg, openAPIDeleteTaskDownstreamTimeout); adjustErr != nil {
+	// Bound the entire downstream cleanup flow, not just the initial connection setup.
+	cleanupCtx, cleanupCancel := context.WithTimeout(ctx, openAPIDeleteTaskDownstreamTimeout)
+	defer cleanupCancel()
+	if adjustErr := AdjustTargetDBSessionCfgWithTimeout(cleanupCtx, toDBCfg, openAPIDeleteTaskDownstreamTimeout); adjustErr != nil {
 		log.L().Warn("skip downstream metadata cleanup when deleting task", zap.String("task", taskName), zap.Error(adjustErr))
-	} else if err = s.removeDownstreamMetaData(ctx, taskName, metaSchema, toDBCfg, openAPIDeleteTaskDownstreamTimeout); err != nil {
+	} else if err = s.removeDownstreamMetaData(cleanupCtx, taskName, metaSchema, toDBCfg, openAPIDeleteTaskDownstreamTimeout); err != nil {
 		log.L().Warn("failed to remove downstream metadata when deleting task", zap.String("task", taskName), zap.Error(err))
 	}
 	release()
--- a/dm/master/server.go
+++ b/dm/master/server.go
@@ -1831,6 +1831,17 @@ func (s *Server) removeDownstreamMetaData(
 			terror.ScopeDownstream,
 		))
 	})
+	failpoint.Inject("MockBlockOnDownstreamMetaDataCleanup", func(val failpoint.Value) {
+		wait := time.Duration(val.(int)) * time.Millisecond
+		timer := time.NewTimer(wait)
+		defer timer.Stop()
+		select {
+		case <-ctx.Done():
+			failpoint.Return(ctx.Err())
+		case <-timer.C:
+			failpoint.Return(nil)
+		}
+	})
 
 	// set up db and clear meta data in downstream db
 	baseDB, err := conn.GetDownstreamDBWithTimeout(toDBCfg, dbTimeout)
--- a/dm/master/openapi_controller_test.go
+++ b/dm/master/openapi_controller_test.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/golang/mock/gomock"
 	"github.com/pingcap/failpoint"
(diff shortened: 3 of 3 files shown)


dm/master/server.go
```diff
@@ -1792,9 +1815,25 @@ func (s *Server) removeMetaData(ctx context.Context, taskName, metaSchema string
 	if err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
GMHDBJD: removeInternalMetaData() already returns internal cleanup errors directly, and the final return nil is only reached after all three steps succeed.

dm/master/openapi_controller.go
```diff
@@ -459,36 +462,32 @@ func (s *Server) deleteTask(ctx context.Context, taskName string, force bool) er
...
-	err = s.removeMetaData(ctx, taskName, metaSchema, toDBCfg)
-	if err != nil {
-		if !ignoreCannotConnectError(err) {
-			return terror.Annotate(err, "while removing metadata")
-		}
+	if adjustErr := AdjustTargetDBSessionCfgWithTimeout(ctx, toDBCfg, openAPIDeleteTaskDownstreamTimeout); adjustErr != nil {
+		log.L().Warn("skip downstream metadata cleanup when deleting task", zap.String("task", taskName), zap.Error(adjustErr))
+	} else if err = s.removeDownstreamMetaData(ctx, taskName, metaSchema, toDBCfg, openAPIDeleteTaskDownstreamTimeout); err != nil {
+		log.L().Warn("failed to remove downstream metadata when deleting task", zap.String("task", taskName), zap.Error(err))
 	}
```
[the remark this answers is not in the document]
GMHDBJD: The else is intentional because downstream cleanup must be skipped entirely when AdjustTargetDBSessionCfgWithTimeout() fails.

dm/master/server.go
```diff
@@ -1774,11 +1778,30 @@ func withHost(addr string) string {
...
+) error {
 	failpoint.Inject("MockSkipRemoveMetaData", func() {
 		failpoint.Return(nil)
 	})
 	toDBCfg.Adjust()
 
+	err := s.removeInternalMetaData(taskName)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
GMHDBJD: This PR only makes downstream cleanup best-effort; internal metadata cleanup must stay fatal to avoid leaving DM state inconsistent.

dm/master/openapi_controller.go
```diff
@@ -459,36 +462,32 @@ func (s *Server) deleteTask(ctx context.Context, taskName string, force bool) er
...
 	release()
+	released = true
 	sourceNameList := s.getTaskSourceNameList(taskName)
 	// delete subtask on worker
-	return s.scheduler.RemoveSubTasks(taskName, sourceNameList...)
+	if err := s.scheduler.RemoveSubTasks(taskName, sourceNameList...); err != nil {
+		return err
+	}
+	return nil
```
joechenrh: Why we change this

GMHDBJD: revert

dm/master/openapi_controller.go
```diff
@@ -459,33 +462,26 @@ func (s *Server) deleteTask(ctx context.Context, taskName string, force bool) er
...
-			return adjustErr
-		}
-	}
-	metaSchema := *task.MetaSchema
-	err = s.removeMetaData(ctx, taskName, metaSchema, toDBCfg)
-	if err != nil {
-		if !ignoreCannotConnectError(err) {
-			return terror.Annotate(err, "while removing metadata")
-		}
+	if adjustErr := AdjustTargetDBSessionCfgWithTimeout(ctx, toDBCfg, openAPIDeleteTaskDownstreamTimeout); adjustErr != nil {
```
OliverS929: One concern here: we now return success even if the downstream metadata cleanup fails, which means the downstream checkpoint / validator metadata may not actually be removed.

If a task is deleted while downstream is unavailable, and later recreated with the same name after downstream recovers, it looks like syncer prefers an existing DB checkpoint, so this may cause the new task to continue from a stale position or mix in old validator state.

Am I missing anything, or is this case not handled by the current PR?

dm/master/openapi_controller.go
```diff
@@ -39,6 +40,8 @@ import (
 	"go.uber.org/zap"
 )
 
+const openAPIDeleteTaskDownstreamTimeout = 10 * time.Second
```
OliverS929: The new `10s` timeout does not look end-to-end.

It seems to cover the initial connect/ping path, but later steps in the downstream cleanup flow still use the original context / default timeout.

If the goal is to keep OpenAPI delete from hanging on an unavailable downstream, is it expected that parts of the cleanup can still block beyond `10s`? The current behavior otherwise looks incomplete.

dm/master/openapi_controller.go
OliverS929: My concern is about the API semantics here.

If `delete task` returns success, users would naturally expect the task state to be fully removed, and recreating a task with the same name should not inherit any old state. In this PR, downstream cleanup becomes best-effort, but I do not see a corresponding mechanism that prevents residual metadata from being picked up again by a later task with the same name. That makes the `success` result a bit risky semantically.

GMHDBJD: In cloud, each time create a new task, it use a new task name. For op, we use start-task --remove-meta for dmctl. So I think it's reasonable here and we also log the timeout here.

dm/master/openapi_controller.go
GMHDBJD: 6a5e199ca

dm/master/openapi_controller.go
OliverS929: Yeah, I understand. I agree that for the happy path you described, this should be fine.

The reason I raised this point was mainly to highlight that there is still a loophole in this PR. The OpenAPI is also available to on-prem users, not just cloud interfaces, so an on-prem user could still run into the scenario I mentioned above.

It would be great if we could at least give users a clearer signal here, so they have the right expectation of what a successful delete task actually means in this case.

Other than that, the rest looks good to me. I just wanted to call out this gap.

Issue #12542, opened 2026-03-09
[release-8.5] Update the GitHub action workflow files


wuhuizuo:
like we did in pr: https://github.com/pingcap/tiflow/pull/12362, but now we need to cherry-pick it to release-8.5 branch

PR #12576, opened 2026-03-26
CI: resurect workflows

dveeden:
Issue Number: ref #12542

- Upgrade from Ubuntu 20.04 to Ubuntu Latest as 20.04 is no longer available.
- Upgrade actions
- Use Go version from go.mod instead of hardcoding it.

#### Tests 

 - No code

### Release note 

```release-note
None
```

### Tests

img_v3_02105_52a1b2d4-fef1-4edf-9bed-aee22151fcix

https://github.com/dveeden/tiflow/actions/runs/23583701374/job/68671780340

This is only for one of the updated workflows.


lance6716 (approved):
please paste the green github actions link to PR description

.github/workflows/dataflow_engine_chaos.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
   # This workflow contains a single job called "base"
   base:
     # The type of runner that the job will run on
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
wuhuizuo: Better to use ubuntu-22.04 or ubuntu-24.04 rather than the `latest` items.

dveeden: Changed this in all files, could you check again?

.github/workflows/upgrade_dm_via_tiup.yaml
```diff
@@ -84,18 +84,18 @@ jobs:
 
     steps:
       - name: Set up Go env
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v6
         with:
-          go-version: '1.23'
+          go-version-file: 'go.mod'
```
OliverS929: I noticed this job runs setup-go before checkout, but `go-version-file: go.mod` requires the repository files to already exist. Could this fail because `go.mod` is not available yet? The same seems to apply to `upgrade_tidb`.

OliverS929 (approved):
Rest LGTM, just one question about the setup-go order/path.

.github/workflows/upgrade_dm_via_tiup.yaml
dveeden: Yes this is a real issue. Both the order and the path of the `go.mod` file.

This is fixed in fa7740b5a842842a42e19b7297f2f974d8ae24a8

PR #12577, opened 2026-03-27
CI: Add basic integration test

### LINKED ISSUE #12614: DM: Extend testing for the latest LTS releases of MySQL and MariaDB
This should include recent LTS releases including MariaDB 11.8 and MySQL 8.4 and also the latest release for both even if they are not LTS releases

dveeden:

Issue Number: close #12614 

#### Tests 

 - Integration test

### Release note 

```release-note
None
```


.github/workflows/dm_integration_basic.yaml
```diff
@@ -0,0 +1,255 @@
+name: Basic Integration Tests
+
+on:
+  schedule:
+    - cron: '0 3 * * *'
+  workflow_dispatch:
+  pull_request:
```
OliverS929, line 7: Could we avoid running this full matrix for every pull request? This builds DM and starts external MySQL/MariaDB/TiDB services, so unrelated PRs would also pay this CI cost and flake surface. If the goal is nightly compatibility coverage, schedule/workflow_dispatch should be enough; otherwise please add path filters for DM-related changes.

.github/workflows/dm_integration_basic.yaml
```diff
@@ -0,0 +1,255 @@
...
+        run: |
+          ./bin/dm-master -V
+          ./bin/dm-worker -V
+          ./bin/dmctl -V
+          mysql -h 127.0.0.1 -u root -BNe "SELECT VERSION()"
+          mysql -h 127.0.0.1 -P 4000 -u root -BNe "SELECT VERSION()"
+      - name: Run DM
+        run: |
+          ./bin/dm-master -master-addr ${DM_MASTER_ADDR} -log-file dm_master.log -log-format json &
+          ./bin/dm-worker -join ${DM_MASTER_ADDR} -worker-addr ${DM_WORKER_ADDR} -log-file dm_worker.log -log-format json &
```
OliverS929, line 70: Please add a readiness gate after starting dm-master and dm-worker. The next step calls dmctl immediately, so a slower runner can fail before the processes are listening or the worker has joined. Existing DM tests usually use retry-based online checks for this layer.

dm/tests/integration_basic/t/basic.test
```diff
@@ -0,0 +1 @@
+select * from test_basic.t1;
```
OliverS929: Please make this validation query deterministic, for example `select * from test_basic.t1 order by id;`. The result file depends on row order, but SQL does not guarantee that order without an explicit ORDER BY.

.github/workflows/dm_integration_basic.yaml
```diff
@@ -0,0 +1,255 @@
...
+      - name: Run input
+        run: mysql -h 127.0.0.1 -u root -v < dm/tests/integration_basic/setup.sql
+      - name: Check status
+        run: |
+          ./bin/dmctl operate-source show
+          ./bin/dmctl query-status
+          ./bin/dmctl query-status mysql-to-tidb
+      - name: install mysql-tester
+        run: |
+          go install github.com/pingcap/mysql-tester/src@latest
```
OliverS929, line 122: Can we pin mysql-tester instead of installing `@latest` at runtime? Otherwise an unrelated upstream mysql-tester change can break this workflow without any TiFlow change. We can do manual updates if needed.

.github/workflows/dm_integration_basic.yaml
dveeden: Restricting this to DM changes only makes sense. I'll try to do that.

.github/workflows/dm_integration_basic.yaml
dveeden: Yes, I'll do that. However as there are no releases nor tags for mysql-tester this is a bit more painful than it should.

.github/workflows/dm_integration_basic.yaml
```diff
@@ -0,0 +1,296 @@
+name: Basic Integration Tests
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'dm/**'
+      - 'cmd/dm*/*'
```
wuhuizuo, line 8: maybe it should be triggered when go.mod file is changed.

.github/workflows/dm_integration_basic.yaml
```diff
@@ -0,0 +1,296 @@
...
+      - name: Run input
+        run: mysql -h 127.0.0.1 -u root -v < dm/tests/integration_basic/setup.sql
+      - name: Check status
+        run: |
+          ./bin/dmctl operate-source show
+          ./bin/dmctl query-status
+          ./bin/dmctl query-status mariadb-to-tidb
+      - name: install mysql-tester
+        run: |
+          go install github.com/pingcap/mysql-tester/src@f2d90ea9522d30c9a8e8d70cc31c7f016ca2801f
```
wuhuizuo, line 263: can we manager the version in go.mod file?

dveeden: Yes, we could do this by using the tool functionality, see also: https://go.dev/doc/go1.24#tools

Do you think this is better?

```patch
diff --git a/go.mod b/go.mod
index aceb438e4..bba975342 100644
--- a/go.mod
+++ b/go.mod
@@ -180,6 +180,7 @@ require (
        github.com/cncf/xds/go v0.0.0-20251022180443-0feb69152e9f // indirect
        github.com/cockroachdb/fifo v0.0.0-20240606204812-0bbfbd93a7ce // indirect
        github.com/cockroachdb/tokenbucket v0.0.0-20230807174530-cc333fc44b06 // indirect
+       github.com/defined2014/mysql v0.0.0-20231121061906-fcfacaa39f49 // indirect
        github.com/envoyproxy/go-control-plane/envoy v1.35.0 // indirect
        github.com/envoyproxy/protoc-gen-validate v1.2.1 // indirect
        github.com/felixge/httpsnoop v1.0.4 // indirect
@@ -216,13 +217,14 @@ require (
        github.com/oasdiff/yaml3 v0.0.0-20250309153720-d2182401db90 // indirect
        github.com/perimeterx/marshmallow v1.1.5 // indirect
        github.com/pingcap/metering_sdk v0.0.0-20251110022152-dac449ac5389 // indirect
+       github.com/pingcap/mysql-tester v0.0.0-20260121034350-f2d90ea9522d // indirect
        github.com/planetscale/vtprotobuf v0.6.1-0.20240319094008-0393e58bdf10 // indirect
        github.com/qri-io/jsonpointer v0.1.1 // indirect
        github.com/qri-io/jsonschema v0.2.1 // indirect
        github.com/robfig/cron/v3 v3.0.1 // indirect
        github.com/segmentio/asm v1.2.0 // indirect
        github.com/segmentio/fasthash v1.0.3 // indirect
-       github.com/sergi/go-diff v1.3.1 // indirect
+       github.com/sergi/go-diff v1.4.0 // indirect
        github.com/spf13/afero v1.15.0 // indirect
        github.com/spiffe/go-spiffe/v2 v2.6.0 // indirect
        github.com/tidwall/btree v1.7.0 // indirect
@@ -462,3 +464,5 @@ replace golang.org/x/text => golang.org/x/text v0.28.0
 
 // tls10server=1
 godebug tlsrsakex=1
+
+tool github.com/pingcap/mysql-tester/src
```

And then:
```sh
cd dm/tests/integration_basic
go tool github.com/pingcap/mysql-tester/src
```

.github/workflows/dm_integration_basic.yaml
OliverS929, line 8: Could we widen the `pull_request.paths` filter a bit? Changes in shared DM-related code or files like `pkg/**`, `go.mod`, `go.sum`, or this workflow itself can still affect `make dm`, but this workflow would not run for them.

.github/workflows/dm_integration_basic.yaml
OliverS929: Got it, thanks. I think my main concern was just avoiding `@latest`; pinning to a commit SHA is totally fine as well. It looks like the MariaDB job is pinned now, while the MySQL job is still floating. Non-blocking from my side.

OliverS929 (approved):
Other than these two small issues, the rest looks good to me.

.github/workflows/dm_integration_basic.yaml
dveeden: I thought `dm` only used `dm/pkg/` and not `pkg/`, but that's not the case. Maybe at some point the two `pkg` directories should be merged...

I've update the paths.

GMHDBJD (approved):

Issue #12580, opened 2026-04-01
secondsBehindMaster / replication lag spuriously drops to 0 between DML batches


mengxian-li:
### What did you do?

  Run a DM task replicating from MySQL to TiDB with a significant replication lag (e.g. DM is        
  processing binlogs from several minutes ago). Then repeatedly query query-status <task-name> at a  
  high frequency (every ~100ms or faster).
                                                                                                     
  Minimal reproduction steps:                                                                        
  1. Start a DM task with a significant upstream lag (e.g. insert a large batch of rows ahead of DM's
   checkpoint).                                                                                      
  2. While DM is actively catching up (DML batches flowing through syncer workers), run query-status 
  <task-name> in a loop.                                                                             
  3. Observe the secondsBehindMaster field in the output. 

### What did you expect to see?

 `secondsBehindMaster` should remain at the actual lag value (e.g. ~300s if 5 minutes behind). It     
  should only drop to 0 when DM has genuinely caught up to the upstream.  

### What did you see instead?

`secondsBehindMaster` intermittently and spuriously drops to 0, then jumps back to the real lag value
   a moment later. This happens repeatedly during active replication, even though DM is significantly
   behind the upstream. The false 0 reading disappears on the next metric update cycle (~100ms).
                                                                                                     
  Root cause: `updateReplicationLagMetric()` runs every 100ms and computes lag from the minimum        
  non-zero entry in `workerJobTSArray`. After each DML batch completes, `successFunc` calls
  `updateReplicationJobTS(nil, idx)` which resets that worker's slot to 0. During the brief window     
  between one batch finishing and the next starting, all worker TSes can be 0 simultaneously. If the
  100ms cron fires in that window, `minTS == 0`, `lag` is computed as 0, and `secondsBehindMaster` is
  stored as 0 — a false reading.

### Versions of the cluster

(affects all versions with workerJobTSArray-based lag computation)
DM version (run `dmctl -V` or `dm-worker -V` or `dm-master -V`):
```
Release Version: v8.5.2-18-g27e2ad16b
```
Downstream TiDB cluster version (execute `SELECT tidb_version();` in a MySQL client):
```
Release Version: v8.5.2-57-gad8d7f1c03
```
>
>
```

### current status of DM cluster (execute `query-status <task-name>` in dmctl)

(task running normally, but secondsBehindMaster oscillates between 0 and the real lag value)

mengxian-li:
Proposed fix:                                                                                      
                  
[...]

PR #12581, opened 2026-04-01
[dm/syncer]: Fix secondsBehindMaster spurious drop to 0 between DML batches

mengxian-li:
## Problem
Ref: https://github.com/pingcap/tiflow/issues/12580

`secondsBehindMaster` (replication lag) spuriously drops to 0 between DML batches in the syncer.

Each DM worker resets its job timestamp to 0 after completing a batch. `updateReplicationLagMetric` computes the minimum non-zero timestamp across all workers — but when all workers finish a batch simultaneously, every worker TS is 0, causing `minTS` to be 0 and reporting a false lag of 0 before the next batch starts.

## Fix

Introduce `lastNonZeroMinTS` atomic field that tracks the last known minimum event timestamp when at least one worker was active. In `updateReplicationLagMetric`, when all worker TSes are 0 (i.e. between batches), fall back to `lastNonZeroMinTS` instead of emitting 0, so the metric holds its last valid value until a new batch begins.

## Tests

Added unit tests to `syncer_test.go` covering:
- Normal lag reporting when workers are active
- Fallback to last non-zero TS when all workers are between batches
- Reset behavior on `reset()`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dm/syncer/syncer_test.go
+++ b/dm/syncer/syncer_test.go
@@ -965,9 +965,11 @@ func (s *testSyncerSuite) TestRun(c *check.C) {
 	resultCh = make(chan pb.ProcessResult)
 	// simulate `syncer.Resume` here, but doesn't reset database conns
 	syncer.secondsBehindMaster.Store(100)
+	syncer.lastNonZeroMinTS.Store(100)
 	syncer.workerJobTSArray[ddlJobIdx].Store(100)
 	syncer.reset()
 	c.Assert(syncer.secondsBehindMaster.Load(), check.Equals, int64(0))
+	c.Assert(syncer.lastNonZeroMinTS.Load(), check.Equals, int64(0))
 	c.Assert(syncer.workerJobTSArray[ddlJobIdx].Load(), check.Equals, int64(0))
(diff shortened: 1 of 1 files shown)


dm/syncer/syncer.go
```diff
@@ -965,7 +970,13 @@ func (s *Syncer) updateReplicationLagMetric() {
 		}
 	}
 	if minTS != int64(0) {
+		s.lastNonZeroMinTS.Store(minTS)
 		lag = s.calcReplicationLag(minTS)
+	} else if lastTS := s.lastNonZeroMinTS.Load(); lastTS != int64(0) {
+		// All workers are between batches (their TSes were reset to 0 after the
+		// last batch completed). Use the most recent known event timestamp so we
+		// don't emit a spurious drop to 0 while no worker is actively processing.
+		lag = s.calcReplicationLag(lastTS)
```
GMHDBJD, line 979: When DM is actually caught up and the upstream is simply idle. In that state calcReplicationLag(lastTS) keeps growing with wall-clock time, so secondsBehindMaster starts reporting artificial lag even though there is no backlog.

Issue #12499, opened 2026-01-22
DM: Reduce "unhandled event" log level


kennytm:
### Is your feature request related to a problem?

When upstream MySQL enabled [binlog_rows_query_log_events](https://dev.mysql.com/doc/refman/8.4/en/replication-options-binary-log.html#sysvar_binlog_rows_query_log_events), the binlog file will contain load of `ROWS_QUERY_LOG_EVENT`. This leads to many useless logs like 

```
[2026/01/21 08:10:28.073 +00:00] [WARN] [syncer.go:2456] ["unhandled event"] [task=t-sync] [unit="binlog replication"] [type=*replication.RowsQueryEvent]
```

and fill up the disk space, rotating useful logs away.

This issue is a generalization of #10204.

### Describe the feature you'd like

Reduce the noise by `"unhandled event"`. It can be done either by
1. Reduce the log level from Warn to Debug. To compensate for the loss of information, we should at the same time create a new CounterVec metric, with the `e.Event.(type)` used as a key; *or*
2. Debounce similar unhandled events, e.g. we maintain a `map[string /*evtype*/]int /*counter*/`, and flush the content to the log every 5 minutes if the sum is nonzero

PR #12597, opened 2026-04-08
log(dm): sample repeated retry logs in retry loops

GMHDBJD:

Issue Number: ref #12499 

This change is a follow-up to #12579 to reduce DM log noise from repeated retry loops outside the unhandled-event path.

- add `log.NewRetrySampleLogger` in `dm/pkg/log` to sample repeated logs with the same level and message during a 5 minute window
- apply the sampled logger to retry-heavy paths in the scheduler, shard DDL optimist/pessimist, worker server/controller, and syncer binlog fetching
- keep component-specific context fields on the logger so the emitted logs still identify the caller
- add unit test coverage for repeated message sampling across level and message combinations

Verification:
- `go test ./dm/pkg/log -run TestNewRetrySampleLogger`
- `go test ./dm/master/scheduler ./dm/master/shardddl ./dm/syncer ./dm/worker -run TestDoesNotExist`

#### Tests

 - Unit test

No material compatibility change. This reduces repeated retry-loop log volume and keeps the first log in each sampling window.

No.

### Release note

```release-note
Reduce repeated DM retry-loop error logs by using sampled logging.
```


dm/pkg/log/log.go
```diff
@@ -33,6 +35,9 @@ const (
 	defaultLogLevel   = "info"
 	defaultLogMaxDays = 7
 	defaultLogMaxSize = 512 // MB
+
+	retryLogSampleInterval = 5 * time.Minute
+	retryLogSampleFirst    = 1
```
D3Hunter: 1 log per 5 minutes? 

maybe 10 per minutes?

GMHDBJD: Updated. The retry sampler now caps identical retry logs at 10 entries per minute instead of 1 entry per 5 minutes, and I also updated the unit test to cover dropping the 11th repeated error in the same window.

D3Hunter (approved):

joechenrh (approved):

Benjamin2037 (approved):
LGTM

PR #12612, opened 2026-04-17
dm: support timezone-aware start-time values

### LINKED ISSUE #12611: DM start_time can be misinterpreted across timezones
### What did you do?

Consider a DM source whose upstream timezone is `UTC` and an operator in `UTC+8` starting incremental replication with a wall-clock time such as `2026-04-17 08:00:00`.

Current behavior across the start-task paths is:

1. `dmctl` / OpenAPI only accept bare `start_time` values without timezone information.
2. The backend parses a bare `start_time` in the upstream timezone when resolving the binlog position.
3. The DM UI currently converts the selected local time to UTC and then sends a bare timestamp string, which drops the original timezone offset.

That means a user intent like “my local `2026-04-17 08:00:00` in `UTC+8`” can end up being interpreted as upstream-local `2026-04-17 08:00:00 UTC`, shifting the effective start point by hours.

`start_time` should support timezone-aware values, for example RFC3339 timestamps with offsets such as `2026-04-17T08:00:00+08:00`, so DM can resolve the intended absolute time correctly.

For backward compatibility, legacy bare timestamps can continue to be interpreted in the upstream timezone.

### What did you see instead?

`start_time` loses the caller's timezone context. In cross-timezone scenarios, DM can resolve the wrong binlog start position.

### Versions of the cluster

```console
Observed on current master branch.
```

```console
Any timezone-aware upstream where the source timezone differs from the caller timezone.
```

```console
N/A for the parsing issue itself.
```

```console
Not deployment-specific.
```

```console
The affected code paths include dmctl/OpenAPI validation, syncer start_time parsing, and the DM UI start-task request payload formatting.
```

GMHDBJD:

Issue Number: close #12611

This change removes the timezone ambiguity in DM `start_time` handling while keeping backward compatibility for existing callers.

- support timezone-aware `start_time` values such as `2026-04-17T08:00:00+08:00` and `2026-04-17 08:00:00+08:00`
- keep legacy bare timestamps interpreted in the upstream timezone
- update `start-task`, `check-task`, validation, and OpenAPI validation messages to advertise the supported formats
- change the DM UI to preserve the selected timezone offset instead of converting to UTC and sending a bare timestamp string
- extend parser and config validation tests for offset-aware timestamps

Verification:

- `go test ./dm/pkg/utils -run 'TestParseStartTime|TestParseStartTimeInLoc' -count=1`
- `go test ./dm/config -check.f TaskCliArgsVerify -count=1`
- `go test ./dm/ctl/master -run TestDoesNotExist -count=1`
- `go test ./dm/master -run TestDoesNotExist -count=1`

#### Tests 

 - [x] Unit test
 - [ ] Integration test
 - [ ] No code

No material performance impact. This is backward compatible because legacy bare timestamps still use the upstream timezone, while offset-aware timestamps are parsed as absolute times.

No.

### Release note 

```release-note
DM now accepts timezone-aware start-time values, and the DM UI preserves timezone offsets when starting tasks.
```


GMHDBJD:
Quick status update after reruns:

- `pull-unit-test-cdc` rerun passed
- `codecov/project` also recovered and is now passing
- the remaining failures are `pull-dm-integration-test` (required) and `pull-dm-integration-test-next-gen` (optional)

I checked both rerun logs and they still look like infra/timeout-style failures rather than failures caused by this DM `start_time` change:

- many matrix groups fail together in the same run
- the common pattern is `make: *** [Makefile:388: sync-diff-inspector] Terminated` or `make: *** [Makefile:528: dm_integration_test_in_group] Terminated`
- the terminated steps exit with code `143`

Examples from the reruns:
- `pull-dm-integration-test-next-gen` job 109
- `pull-dm-integration-test` job 219

So at this point the remaining blockers look more like flaky/infra interruption than a deterministic regression from this PR.

D3Hunter (approved):

dm/pkg/utils/time.go
```diff
@@ -68,16 +86,28 @@ func ParseTimeZone(s string) (*time.Location, error) {
...
 
 // ParseStartTimeInLoc parses start-time of task-start and validation-start.
+// If the time string contains a timezone offset, it is parsed as an absolute time.
+// Otherwise it is interpreted in the specified location.
 func ParseStartTimeInLoc(timeStr string, loc *time.Location) (time.Time, error) {
-	t, err := time.ParseInLocation(StartTimeFormat, timeStr, loc)
-	if err != nil {
-		return time.ParseInLocation(StartTimeFormat2, timeStr, loc)
+	for _, layout := range startTimeLayoutsWithTZ {
+		t, parseErr := time.Parse(layout, timeStr)
```
joechenrh: Is it possible to use `types.ParseTime` here?

joechenrh (commented):
Rest LGTM

dm/pkg/utils/time.go
GMHDBJD: Yes — updated to use `types.ParseTime` for the actual parsing. I kept a narrow regex gate in front of it because `types.ParseTime` accepts a much wider set of datetime literals (for example `20060102 150405`, `2026/04/17 08:00:00`, date-only values, fractional seconds, etc.), while the DM CLI here still intends to support only the documented start-time shapes.

joechenrh (approved):

Benjamin2037 (approved):
LGTM

Issue #11935, opened 2024-12-26
sarama produces messages in out-of-order when retries happen


wk989898:
we found sarama sends out-of-order messages when retries happen, and we think the issue causes this:  
- https://github.com/IBM/sarama/issues/2619

This could be reproduced by hacking codes. https://github.com/wk989898/sarama/commit/d06ee56ae5986f77f5bc88b8a63290e068a9c16c

3AceShowHand:
This affect release after v6.4.0

PR #12618, opened 2026-04-22
kafka: bump sarama version and enable the retry to fix the broken pipe and out of order

### LINKED ISSUE #12655: kafka: Consider reintroducing retry (or at least make it configurable)

Currently CDC will not retry when meeting a failure because of pingcap/tiflow#11935 = IBM/sarama#2619, because retrying would cause messages to be sent in wrong order. <sub><sup>(ONCALL-7528 = TICKET-5339)</sup></sub> There is also a data-loss bug related to retrying. <sub><sup>(TCOC-3167 = ST-1473)</sup></sub>

Retrying has been disabled since v9.0.0-beta (pingcap/tiflow#11870), v8.5.1, v8.1.3, v7.5.6, v7.1.7 and v6.5.12.

However, the current behavior (failing the sink immediately on error) leads to many bad side-effect that also affects user experience:

1. Sink errors are recorded in metrics and leads to alerts. 
2. Hard-reset via sink-error is known to cause another sarama bug IBM/sarama#2681. 

In order of preference:

1. Actually fix IBM/sarama#2619 and the related data-loss bug, and re-enable retrying
2. Add a sink-URI option to unsafely enable retry.

_No response_

_No response_

3AceShowHand:

TiCDC disabled Kafka producer retries to avoid the Sarama out-of-order retry issue tracked by #11935. With retries disabled, transient Kafka send errors such as broken pipe fail the sink immediately, create noisy sink-error metrics and alerts, and can trigger costly sink rebuilds.

Issue Number: close #12655, ref #11935

- Update the Sarama replacement to `github.com/pingcap/sarama v1.41.2-pingcap-20260508`, which includes the producer retry ordering fix.
- Re-enable bounded Kafka producer retry by setting Sarama `Producer.Retry.Max` to `5` by default while keeping `Net.MaxOpenRequests = 1` as an additional ordering guard.
- Add Kafka sink URI parameter `max-retry` to configure Sarama `Producer.Retry.Max`; `max-retry=0` disables producer retry and negative values are ignored.
- Remove the Kafka heartbeat keep-alive workaround and related interfaces/tests now that bounded producer retry handles transient stale-connection failures.
- Simplify Kafka producer/admin factory code after removing the heartbeat path.

#### Tests

- Unit test: `go test ./pkg/sink/kafka`

No breaking compatibility is expected. This intentionally changes Kafka sink behavior by enabling bounded producer retries for transient send failures. Users can set `max-retry=0` in the sink URI to disable producer retry if needed.

User documentation should mention the new Kafka sink URI parameter `max-retry`. No design or monitoring documentation update is required.

### Release note

```release-note
Kafka sink now retries transient producer send failures by default.
```


wk989898 (approved):

flowbehappy (approved):

hongyunyan (approved):

wk989898:
Manual Test:
1. Start a kafka server
2. Hack the sarama code like https://github.com/wk989898/sarama/tree/hack
3. Add more logs for the producer
```Go
// Add HandleKey field for message
HandleKey    *string
```
5. Send message with follower code
```Go
package main

import (
    "context"
    "fmt"
    "math/rand"
    "sync/atomic"
    "time"

    "github.com/IBM/sarama"
    "github.com/pingcap/log"
    "github.com/pingcap/tiflow/pkg/config"
    "github.com/pingcap/tiflow/pkg/sink/codec/common"
    "github.com/pingcap/tiflow/pkg/sink/kafka"
    "go.uber.org/zap"
    "go.uber.org/zap/zapcore"
)

var (
    PartitionNum      int32 = 3
    ReplicationFactor int16 = 1
    partition         int32 = 0
    schema                  = ""
    table                   = ""
)

func initlog() {
    logger, err := zap.NewStdLogAt(log.L().With(zap.String("component", "sarama")), zapcore.InfoLevel)
    if err != nil {
        fmt.Println("err", err)
    }
    sarama.Logger = logger
}
func main() {
    initlog()
    topic := fmt.Sprintf("test-%d", rand.Int())
    ctx := context.Background()
    o := kafka.NewOptions()
    o.BrokerEndpoints = []string{"127.0.0.1:9092"}
    o.ClientID = "sarama-test"

    f, err := kafka.NewSaramaFactory(o, model.DefaultChangeFeedID("sarama-test"))
    if err != nil {
        fmt.Println("err", err)
    }
    admin, err := f.AdminClient(ctx)
    if err != nil {
        fmt.Println("err", err)
    }
    err = admin.CreateTopic(ctx, &kafka.TopicDetail{Name: topic,
        NumPartitions:     PartitionNum,
        ReplicationFactor: ReplicationFactor,
    }, false)
    if err != nil {
        fmt.Println("err", err)
    }
    async, err := f.AsyncProducer(ctx, make(chan error, 1))
    if err != nil {
        fmt.Println("err", err)
    }
    go func() {
        err := async.AsyncRunCallback(ctx)
        if err != nil {
            fmt.Println("err", err)
        }
    }()
    var count int32
    for i := 0; i < 200; i++ { 
        val := fmt.Sprintf("val: %d", i)
        msg := common.NewMsg(config.ProtocolCanalJSON, nil, []byte(val), 0, model.MessageTypeRow, &schema, &table)
        msg.Callback = func() {
            atomic.AddInt32(&count, 1)
        }
        msg.HandleKey = []string{val}
        err := async.AsyncSend(ctx, topic, partition, msg)
        if err != nil {
            fmt.Println("AsyncSend err", err)
        }
        time.Sleep(time.Microsecond * 1000) 
    }
    for atomic.LoadInt32(&count) < 200 {

    }
    fmt.Println("topic", topic)
}

```
6. Consume the message from kafka server
```
kafka-console-consumer.sh --bootstrap-server localhost:9092 --from-beginning --topic test-xxx
```
7. Check the out-of-order from the logs

PR #12622, opened 2026-04-24
sync-diff: replace hardcoded use of MD5 with configurable option to support FIPS-compliant environments.

maxz-db:

Issue Number: close #12627

Added a new checksum-algorithm configuration flag to sync-diff-inspector:
Supported options: md5" and "sha256" hash functions for checksumming
Default: md5 for backwards compatibility.

#### Tests 

- [x] Unit test
- [ ] Integration test
Manual test on dev environment.
- [ ] No code

No. This only reuses the existing binlog value normalization path in validator.

No.

### Release note 

```release-note
sync-diff: replace hardcoded use of MD5 with configurable option to support FIPS-compliant environments.
```

### DIFF
--- a/sync_diff_inspector/config/config.go
+++ b/sync_diff_inspector/config/config.go
@@ -65,6 +65,16 @@ const (
 	UnifiedTimeZone string = "+0:00"
 )
 
+// ChecksumAlgorithm specifies the hash function to use for chunk checksumming.
+type ChecksumAlgorithm string
+
+const (
+	// MD5 uses MD5 hash function (default for backwards compatibility)
+	MD5 ChecksumAlgorithm = "md5"
+	// SHA256 uses SHA256 hash function (for FIPS-compliant environments)
+	SHA256 ChecksumAlgorithm = "sha256"
+)
+
 // TableConfig is the config of table.
 type TableConfig struct {
 	// table's filter to tell us which table should adapt to this config.
@@ -135,6 +145,8 @@ type DataSource struct {
 
 	Conn          *sql.DB
 	SessionConfig SessionConfig `toml:"session" json:"session"`
+
+	ChecksumAlgorithm ChecksumAlgorithm `toml:"checksum-algorithm" json:"checksum-algorithm"`
 }
 
 // IsAutoSnapshot returns true if the tidb_snapshot is expected to automatically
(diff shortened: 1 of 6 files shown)


sync_diff_inspector/source/mysql_shard.go
```diff
@@ -408,9 +409,14 @@ func NewMySQLSources(ctx context.Context, tableDiffs []*common.TableDiff, ds []*
 		return nil, errors.Annotatef(err, "please make sure the filter is correct.")
 	}
 
+	checksumAlgorithm := config.MD5
+	if len(ds) > 0 {
+		checksumAlgorithm = ds[0].ChecksumAlgorithm
+	}
```
joechenrh, line 415: `len(ds)` is validated to be positive before entering this function.

sync_diff_inspector/config/config.go
```diff
@@ -135,6 +145,8 @@ type DataSource struct {
 
 	Conn          *sql.DB
 	SessionConfig SessionConfig `toml:"session" json:"session"`
+
+	ChecksumAlgorithm ChecksumAlgorithm `toml:"checksum-algorithm" json:"checksum-algorithm"`
```
joechenrh, line 149: How about passing `Config` to `buildSourceFromCfg`, so we don't need store `ChecksumAlgorithm` in the `DataSource`.

joechenrh (commented):
And please update the PR description.

sync_diff_inspector/utils/utils.go
```diff
@@ -857,21 +857,25 @@ func GetTableSize(ctx context.Context, db *sql.DB, schemaName, tableName string)
...
 
-// GetCountAndMD5Checksum returns checksum code and count of some data by given condition
-func GetCountAndMD5Checksum(
+// GetCountAndChecksum returns checksum code and count of some data by given condition
+func GetCountAndChecksum(
 	ctx context.Context,
 	db *sql.DB, schemaName, tableName string,
-	tbInfo *model.TableInfo, limitRange string, indexHint string, args []any,
+	tbInfo *model.TableInfo, limitRange string, indexHint string,
+	args []any, checksumAlgorithm string,
```
D3Hunter, line 865: can we pass the param as ChecksumAlgorithm instead of string

PR #12654, opened 2026-05-27
dm: make FK relation builder route-aware

OliverS929:

Issue Number: ref #12350

This PR makes DM schema tracker build foreign-key causality relations from source-side FK metadata when source and downstream table identities differ.

The existing syncer path still uses the identity route, so this change does not enable routed multi-worker FK causality by itself.

#### Tests 

 - [x] Unit test
 - [ ] Integration test
 - [ ] No code

### Release note 

```release-note
None
```

### DIFF
--- a/dm/pkg/schema/tracker.go
+++ b/dm/pkg/schema/tracker.go
@@ -627,9 +627,14 @@ func (tr *Tracker) buildForeignKeyRelations(
 	visiting[sourceTableID] = struct{}{}
 	defer delete(visiting, sourceTableID)
 
-	routedSourceTable := resolveTableRoute(routeResolver, sourceTable)
+	routedSourceTable := sourceTable
+	if routeResolver != nil {
+		routedSourceTable = routeResolver(sourceTable)
+	}
 	if len(downstreamTI.ForeignKeys) > 0 && !sameTableIdentity(routedSourceTable, targetTable) {
 		if routeResolver == nil {
+			// Without a route resolver, source and target tables must match. Routed FK
+			// causality is enabled only after the syncer passes a validated resolver.
 			return nil, newForeignKeyRouteUnsupportedError(
 				fmt.Sprintf(
 					"foreign key causality with route under foreign_key_checks=1 and worker_count>1 is not supported yet; child table %s routes to %s; please use worker_count=1",
@@ -638,6 +643,8 @@ func (tr *Tracker) buildForeignKeyRelations(
 				),
 			)
 		}
+		// With a route resolver, this builder checks only the current source table's
+		// routed target. Syncer must reject many-to-one topology before this call.
 		return nil, newForeignKeyRouteUnsupportedError(
 			fmt.Sprintf(
 				"foreign key causality with route under foreign_key_checks=1 and worker_count>1 requires 1:1 route alignment; source child table %s routes to %s, but downstream child table is %s",
@@ -704,7 +711,11 @@ func (tr *Tracker) buildForeignKeyRelations(
 			targetParentSchema = targetTable.Schema
 		}
 		targetParentTable := &filter.Table{Schema: targetParentSchema, Name: fk.RefTable.O}
-		routedSourceParent := resolveTableRoute(routeResolver, sourceParentTable)
+		routedSourceParent := sourceParentTable
+		if routeResolver != nil {
+			routedSourceParent = routeResolver(sourceParentTable)
+		}
+		// The source parent table must route to the downstream parent table.
 		if !sameTableIdentity(routedSourceParent, targetParentTable) {
 			if routeResolver == nil {
 				return nil, newForeignKeyRouteUnsupportedError(
@@ -877,17 +888,6 @@ func sameTableIdentity(a *filter.Table, b *filter.Table) bool {
 		return a == b
 	}
 	return a.Schema == b.Schema && a.Name == b.Name
-}
-
-func resolveTableRoute(routeResolver TableRouteResolver, table *filter.Table) *filter.Table {
-	if routeResolver == nil || table == nil {
-		return table
-	}
-	routed := routeResolver(table)
-	if routed == nil {
-		return table
-	}
-	return routed
 }
 
 func newForeignKeyRouteUnsupportedError(msg string) error {
--- a/dm/pkg/schema/tracker.go
+++ b/dm/pkg/schema/tracker.go
@@ -609,22 +609,23 @@ func (tr *Tracker) buildForeignKeyRelations(
 	cache map[string][]sqlmodel.ForeignKeyCausalityRelation,
 	visiting map[string]struct{},
 ) ([]sqlmodel.ForeignKeyCausalityRelation, error) {
-	if relations, ok := cache[tableID]; ok {
-		return relations, nil
-	}
-
-	if _, ok := visiting[tableID]; ok {
-		return nil, nil
-	}
-	visiting[tableID] = struct{}{}
-	defer delete(visiting, tableID)
-
 	if sourceTable == nil {
 		sourceTable = utils.UnpackTableID(tableID)
 	}
 	if targetTable == nil {
 		targetTable = utils.UnpackTableID(tableID)
 	}
+
+	sourceTableID := utils.GenTableID(sourceTable)
+	if relations, ok := cache[sourceTableID]; ok {
+		return relations, nil
+	}
+
+	if _, ok := visiting[sourceTableID]; ok {
+		return nil, nil
+	}
+	visiting[sourceTableID] = struct{}{}
+	defer delete(visiting, sourceTableID)
 
 	routedSourceTable := resolveTableRoute(routeResolver, sourceTable)
 	if len(downstreamTI.ForeignKeys) > 0 && !sameTableIdentity(routedSourceTable, targetTable) {
@@ -814,7 +815,7 @@ func (tr *Tracker) buildForeignKeyRelations(
 		}
 	}
 
-	cache[tableID] = relations
+	cache[sourceTableID] = relations
 	return relations, nil
 }
 
--- a/dm/pkg/schema/tracker.go
+++ b/dm/pkg/schema/tracker.go
@@ -668,7 +668,7 @@ func (tr *Tracker) buildForeignKeyRelations(
 			continue
 		}
 
-		sourceFK := findMatchingForeignKey(originTI, fk, i)
+		sourceFK := findMatchingForeignKey(originTI, sourceTable, targetTable, fk, i, routeResolver)
 		if sourceFK == nil {
 			return nil, newForeignKeySchemaAlignmentError(
 				sourceTable.String(),
@@ -832,9 +832,17 @@ func (tr *Tracker) buildForeignKeyRelations(
 }
 
 // findMatchingForeignKey maps a downstream FK back to source-side metadata.
-// It matches by FK name first, then by the same ordinal position if the child/ref
-// columns still align, and finally falls back to a linear scan by FK columns.
-func findMatchingForeignKey(originTI *model.TableInfo, downstreamFK *model.FKInfo, idx int) *model.FKInfo {
+// A same-name source FK must match the columns and never falls back to another FK.
+// Without a same-name FK, candidates must match both columns and the referenced
+// table when a route resolver is available.
+func findMatchingForeignKey(
+	originTI *model.TableInfo,
+	sourceTable *filter.Table,
+	targetTable *filter.Table,
+	downstreamFK *model.FKInfo,
+	idx int,
+	routeResolver TableRouteResolver,
+) *model.FKInfo {
 	if originTI == nil || len(originTI.ForeignKeys) == 0 || downstreamFK == nil {
 		return nil
 	}
@@ -850,13 +858,13 @@ func findMatchingForeignKey(originTI *model.TableInfo, downstreamFK *model.FKInf
 
 	if idx < len(originTI.ForeignKeys) {
 		candidate := originTI.ForeignKeys[idx]
-		if sameForeignKeyColumns(candidate, downstreamFK) {
+		if sameForeignKeyMetadata(candidate, sourceTable, downstreamFK, targetTable, routeResolver) {
 			return candidate
 		}
 	}
 
 	for _, fk := range originTI.ForeignKeys {
-		if sameForeignKeyColumns(fk, downstreamFK) {
+		if sameForeignKeyMetadata(fk, sourceTable, downstreamFK, targetTable, routeResolver) {
 			return fk
 		}
 	}
@@ -864,11 +872,41 @@ func findMatchingForeignKey(originTI *model.TableInfo, downstreamFK *model.FKInf
 	return nil
 }
 
+func sameForeignKeyMetadata(
+	sourceFK *model.FKInfo,
+	sourceTable *filter.Table,
+	downstreamFK *model.FKInfo,
+	targetTable *filter.Table,
+	routeResolver TableRouteResolver,
+) bool {
+	if !sameForeignKeyColumns(sourceFK, downstreamFK) {
+		return false
+	}
+	if routeResolver == nil {
+		return true
+	}
+
+	sourceParentTable := foreignKeyRefTable(sourceTable, sourceFK)
+	sourceParentTable = routeResolver(sourceParentTable)
+	return sameTableIdentity(sourceParentTable, foreignKeyRefTable(targetTable, downstreamFK))
+}
+
 func sameForeignKeyColumns(sourceFK *model.FKInfo, downstreamFK *model.FKInfo) bool {
 	if sourceFK == nil || downstreamFK == nil {
 		return false
 	}
 	return sameColumns(sourceFK.Cols, downstreamFK.Cols) && sameColumns(sourceFK.RefCols, downstreamFK.RefCols)
+}
+
+func foreignKeyRefTable(childTable *filter.Table, fk *model.FKInfo) *filter.Table {
+	if fk == nil {
+		return nil
+	}
+	schema := fk.RefSchema.O
+	if schema == "" && childTable != nil {
+		schema = childTable.Schema
+	}
+	return &filter.Table{Schema: schema, Name: fk.RefTable.O}
 }
 
 func sameColumns(a []ast.CIStr, b []ast.CIStr) bool {
--- a/dm/pkg/schema/tracker.go
+++ b/dm/pkg/schema/tracker.go
@@ -91,6 +91,10 @@ type DownstreamTableInfo struct {
 	foreignKeyInitErr   error
 }
 
+// TableRouteResolver resolves a source table to its downstream routed table.
+// A nil resolver means identity route.
+type TableRouteResolver func(*filter.Table) *filter.Table
+
 type executorContext struct {
 	sessionctx.Context
 }
@@ -406,7 +410,7 @@ func (tr *Tracker) GetDownStreamTableInfo(tctx *tcontext.Context, tableID string
 		return nil, err
 	}
 
-	if err := dti.initForeignKeyRelations(tr, tctx, tableID, targetTable, targetTable, originTI); err != nil {
+	if err := dti.initForeignKeyRelations(tr, tctx, tableID, targetTable, targetTable, originTI, nil); err != nil {
 		return nil, err
 	}
 
@@ -424,6 +428,7 @@ func (tr *Tracker) InitDownStreamForeignKeyRelations(
 	sourceTable *filter.Table,
 	targetTable *filter.Table,
 	originTI *model.TableInfo,
+	routeResolver TableRouteResolver,
 ) (*DownstreamTableInfo, error) {
 	tableID := utils.GenTableID(targetTable)
 	dti, err := tr.GetDownStreamTableInfoWithoutForeignKey(tctx, tableID, originTI)
@@ -431,7 +436,31 @@ func (tr *Tracker) InitDownStreamForeignKeyRelations(
 		return nil, err
 	}
 
-	if err := dti.initForeignKeyRelations(tr, tctx, tableID, sourceTable, targetTable, originTI); err != nil {
+	// Routed FK relations depend on the source table and resolver, so keep them
+	// out of the downstream table cache that is keyed only by target table ID.
+	if routeResolver != nil {
+		relations, err := tr.buildForeignKeyRelations(
+			tctx,
+			tableID,
+			sourceTable,
+			targetTable,
+			dti.TableInfo,
+			originTI,
+			routeResolver,
+			make(map[string][]sqlmodel.ForeignKeyCausalityRelation),
+			make(map[string]struct{}),
+		)
+		if err != nil {
+			return nil, err
+		}
+		return &DownstreamTableInfo{
+			TableInfo:           dti.TableInfo,
+			WhereHandle:         dti.WhereHandle,
+			ForeignKeyRelations: relations,
+		}, nil
+	}
+
+	if err := dti.initForeignKeyRelations(tr, tctx, tableID, sourceTable, targetTable, originTI, routeResolver); err != nil {
 		return nil, err
 	}
 	return dti, nil
@@ -484,6 +513,7 @@ func (dti *DownstreamTableInfo) initForeignKeyRelations(
 	sourceTable *filter.Table,
 	targetTable *filter.Table,
 	originTI *model.TableInfo,
+	routeResolver TableRouteResolver,
 ) error {
 	dti.foreignKeyInitOnce.Do(func() {
 		relations, err := tr.buildForeignKeyRelations(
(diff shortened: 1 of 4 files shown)


dm/pkg/schema/tracker.go
```diff
@@ -574,6 +605,7 @@ func (tr *Tracker) buildForeignKeyRelations(
 	targetTable *filter.Table,
 	downstreamTI *model.TableInfo,
 	originTI *model.TableInfo,
+	routeResolver TableRouteResolver,
```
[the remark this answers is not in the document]
OliverS929: Good catch. Yes the route-aware builder works in the source-table domain, so its recursive cache and visiting keys should use source table IDs instead of routed downstream table IDs. Updated that and added a focused unit test. Shared-target routing remains unsupported and must still be rejected before production use though.

OliverS929:
@lance6716 @GMHDBJD @D3Hunter @joechenrh PTAL

dm/pkg/schema/tracker.go
```diff
@@ -484,6 +513,7 @@ func (dti *DownstreamTableInfo) initForeignKeyRelations(
 	sourceTable *filter.Table,
 	targetTable *filter.Table,
 	originTI *model.TableInfo,
+	routeResolver TableRouteResolver,
```
joechenrh, line 516: It looks like `initForeignKeyRelations` should only be called with a nil `routeResolver`.

joechenrh:
/retest

😂 Quite strange...

dm/pkg/schema/tracker.go
OliverS929: Yes, this cached init path should only be used with a nil resolver. Routed callers will pass a real resolver to `InitDownStreamForeignKeyRelations`, but that path bypasses this helper and builds relations per call.

joechenrh: How about adding something like (not sure if it's overdesigned😂)

```go
cachedRelations map[string]fkEntry  // sourceTableID -> cached entry

type fkEntry struct {
    once      sync.Once
    relations []sqlmodel.ForeignKeyCausalityRelation
    err       error
}
```

OliverS929: I’d avoid adding this cache in this preparatory PR. The important part here is that routed relations with a non-nil resolver must not reuse the target-table sync.Once cache, since those relations depend on the source table and route resolver. The existing sync.Once path still covers the nil-resolver behavior. 
If this becomes hot after the routed caller is wired in, we can consider this path together with clear route/topology invalidation rules.

joechenrh (approved):

dm/pkg/schema/tracker.go
```diff
@@ -617,30 +661,37 @@ func (tr *Tracker) buildForeignKeyRelations(
...
-					fmt.Sprintf("downstream FK column %s was not found in the tracked schema", col.L),
-				)
-			}
-			childIdxs = append(childIdxs, idx)
-		}
-
 		sourceFK := findMatchingForeignKey(originTI, fk, i)
 		if sourceFK == nil {
-			return nil, newForeignKeyRouteUnsupportedError(
+			return nil, newForeignKeySchemaAlignmentError(
```
D3Hunter, line 673: #### ⚠️ **[Major]** FK metadata mismatch uses a downstream-fetch error code

**Why**
Prefer a more specific error name here; the changed path reports source/downstream FK metadata alignment failures through `ErrSchemaTrackerCannotFetchDownstreamCreateTableStmt`, whose name and base message describe fetching `SHOW CREATE TABLE` from downstream rather than metadata mismatch or stale alignment.

**Scope**
dm/pkg/schema/tracker.go:666, dm/pkg/terror/error_list.go:1384

**Risk if unchanged**
Operators and any error-code based handling can classify a stale or mismatched FK metadata problem as a downstream fetch failure, which points repair toward connectivity or `SHOW CREATE TABLE` instead of schema alignment.

**Evidence**
The diff changes the `sourceFK == nil` branch to `newForeignKeySchemaAlignmentError(...)`, and that helper generates `ErrSchemaTrackerCannotFetchDownstreamCreateTableStmt`; the underlying error definition at `dm/pkg/terror/error_list.go:1384` is still named and worded as a downstream create-table fetch error.

**Change request**
Introduce or use a dedicated schema-tracker FK alignment error, for example `ErrSchemaTrackerForeignKeyMetadataMismatch`, or rename/update the existing error contract so the code/name/message all describe schema alignment rather than downstream fetch.

dm/pkg/schema/tracker.go
```diff
@@ -424,14 +428,39 @@ func (tr *Tracker) InitDownStreamForeignKeyRelations(
...
 	tableID := utils.GenTableID(targetTable)
 	dti, err := tr.GetDownStreamTableInfoWithoutForeignKey(tctx, tableID, originTI)
 	if err != nil {
 		return nil, err
 	}
 
-	if err := dti.initForeignKeyRelations(tr, tctx, tableID, sourceTable, targetTable, originTI); err != nil {
+	// Routed FK relations depend on the source table and resolver, so keep them
+	// out of the downstream table cache that is keyed only by target table ID.
+	if routeResolver != nil {
```
D3Hunter, line 441: #### ⚠️ **[Major]** Route-aware FK causality accepts unsupported many-to-one routes

**Why**
The new non-nil `routeResolver` path builds FK causality relations after checking only that the current source child maps to the current downstream child and that the current source parent maps to the current downstream parent. That local check does not prove the task-wide 1:1 route alignment contract because another source table can still map to the same downstream table.

**Scope**
dm/pkg/schema/tracker.go:441, dm/pkg/schema/tracker.go:630, dm/pkg/schema/tracker.go:707, dm/pkg/schema/tracker_test.go:1002, dm/pkg/schema/tracker_test.go:1020, pkg/sqlmodel/causality.go:30

**Risk if unchanged**
If this route-aware API is wired into multi-worker FK causality, shard routes such as `p1`/`p2` -> `parent_r` and `gp1`/`gp2` -> `grandparent_r` can produce separate source-domain causality keys for rows that collide on the same downstream parent key. Those rows can then be dispatched independently even though downstream FK checks require serialized ordering, leading to intermittent foreign-key failures or out-of-order writes under `foreign_key_checks=1` and `worker_count>1`.

**Evidence**
`InitDownStreamForeignKeyRelations` bypasses the cached init whenever `routeResolver != nil` and calls `buildForeignKeyRelations` directly at `dm/pkg/schema/tracker.go:441`. The builder's child check compares only `resolveTableRoute(routeResolver, sourceTable)` with `targetTable` at `dm/pkg/schema/tracker.go:630`, and its parent check does the same local comparison at `dm/pkg/schema/tracker.go:707`. The added test at `dm/pkg/schema/tracker_test.go:1002` documents this shared-target route as unsupported, but the same test calls the exported route-aware API at `dm/pkg/schema/tracker_test.go:1020` and expects success with generated relations. `pkg/sqlmodel/causality.go:30` defines these keys as the serialization boundary for parallel replication.

**Change request**
Please enforce the one-to-one route precondition before returning route-aware FK relations, or change the API to accept only prevalidated route metadata that proves injective child, parent, and ancestor mappings. Add a negative case for the `p1`/`p2` to `parent_r` and `gp1`/`gp2` to `grandparent_r` topology so this path fails closed instead of relying on a future caller-side precheck.

dm/syncer/syncer.go
```diff
@@ -3830,7 +3830,7 @@ func (s *Syncer) prepareDownStreamTableInfo(
 	if err := s.precheckForeignKeyCausality(sourceTable, targetTable, dti); err != nil {
 		return nil, err
 	}
-	return s.schemaTracker.InitDownStreamForeignKeyRelations(tctx, sourceTable, targetTable, originTI)
+	return s.schemaTracker.InitDownStreamForeignKeyRelations(tctx, sourceTable, targetTable, originTI, nil)
```
D3Hunter, line 3833: #### ⚠️ **[Major]** Route-aware FK relations are unreachable from the syncer

**Why**
The new resolver-based FK relation construction only runs when `InitDownStreamForeignKeyRelations` receives a non-nil resolver, but the syncer's DML/validator preparation path still rejects routed child tables before that call and then passes `nil` as the resolver.

**Scope**
dm/syncer/syncer.go:3830, dm/syncer/syncer.go:3833, dm/pkg/schema/tracker.go:441

**Risk if unchanged**
A 1:1 aligned route such as source child -> routed child and source parent -> routed parent still stops with the old `worker_count=1` route error under `foreign_key_checks=1` and `worker_count>1`. The route-aware source-domain relation logic added in the schema tracker is therefore test-only for the production DML and validator path.

**Evidence**
`precheckForeignKeyCausality` returns an unsupported-route error whenever the downstream child has FKs and `sourceTable` differs from `targetTable`; only after that does `prepareDownStreamTableInfo` call `InitDownStreamForeignKeyRelations`, and it passes `nil`, so the branch guarded by `routeResolver != nil` cannot execute.

**Change request**
Please pass a resolver such as `s.route` into `InitDownStreamForeignKeyRelations` and make the precheck validate the same 1:1 child/parent alignment, or remove the route-aware production-facing API until the syncer is wired to use it.

dm/pkg/schema/tracker.go
```diff
@@ -91,6 +91,10 @@ type DownstreamTableInfo struct {
 	foreignKeyInitErr   error
 }
 
+// TableRouteResolver resolves a source table to its downstream routed table.
```
D3Hunter, line 94: #### ⚠️ **[Major]** Route-aware FK API is exposed without a production owner

**Why**
This logic is better colocated with the component that owns routing. The schema package now exports `TableRouteResolver` and a non-nil route-aware branch on `InitDownStreamForeignKeyRelations`, but the only production caller still passes `nil` after syncer-side prechecks. That exposes a public route contract before production code actually consumes or owns it.

**Scope**
dm/pkg/schema/tracker.go:94, dm/pkg/schema/tracker.go:425, dm/pkg/schema/tracker.go:441, dm/syncer/syncer.go:3833, dm/pkg/schema/tracker_test.go:1020

**Risk if unchanged**
The schema package gains a public route-aware contract whose semantics are still implied by tests and comments. Future callers can depend on the partially wired API, bypass syncer-owned route checks, or make the non-nil branch the de facto route policy boundary before the 1:1 route precondition is documented and enforced.

**Evidence**
`TableRouteResolver` is newly exported and added to the exported `InitDownStreamForeignKeyRelations` signature, but the only non-test caller is `dm/syncer/syncer.go:3833`, which passes `nil`; non-nil resolver values appear only in `dm/pkg/schema/tracker_test.go`. The route-aware path constructs a detached `DownstreamTableInfo` directly instead of using the cached initialization path, and `TestForeignKeyRelationInternalCacheUsesSourceIdentityForUnsupportedSharedTarget` documents a routed shared-target shape as unsupported for production while still calling the exported route-aware API successfully.

**Change request**
Keep the resolver type and route-aware relation builder private/test-scoped until syncer wires `s.route` through with enforced route prechecks, or move the resolver contract to the syncer boundary and document/enforce the 1:1 route precondition before schema builds relations.

dm/pkg/schema/tracker.go
```diff
@@ -666,7 +730,7 @@ func (tr *Tracker) buildForeignKeyRelations(
 		}
 
 		parentTableID := utils.GenTableID(targetParentTable)
-		parentTableName := (&cdcmodel.TableName{Schema: targetParentSchema, Table: fk.RefTable.O}).String()
+		parentTableName := (&cdcmodel.TableName{Schema: sourceParentSchema, Table: sourceFK.RefTable.O}).String()
```
D3Hunter, line 744: #### 🟡 **[Minor]** ParentTable docs do not state the source-domain route behavior

**Why**
Please add a short contract note for this branch because the route-aware change now stores source-side parent identity in `ForeignKeyCausalityRelation.ParentTable`, not the downstream routed table identity.

**Scope**
pkg/sqlmodel/foreign_key.go:20, dm/pkg/schema/tracker.go:733

**Risk if unchanged**
Consumers reading the public `ForeignKeyCausalityRelation` contract can reasonably assume `ParentTable` names the downstream parent table, then produce or validate causality keys against the wrong table domain under routed FK replication.

**Evidence**
The diff changes `parentTableName` in `dm/pkg/schema/tracker.go:733` to use `sourceParentSchema` and `sourceFK.RefTable`, and the new route tests assert source names such as ``db`.`parent`` while the downstream tables are ``db_r`.`parent_r``. The public struct comment only says `ParentTable` may be a direct parent or ancestor/root parent, without stating source/upstream versus downstream routed identity.

**Change request**
Update the `ForeignKeyCausalityRelation` comment to state that `ParentTable` and `ParentColumns` are in the source/upstream causality-key domain after recursive lifting and may differ from the downstream routed table.

dm/pkg/schema/tracker.go
```diff
@@ -424,14 +428,39 @@ func (tr *Tracker) InitDownStreamForeignKeyRelations(
 	sourceTable *filter.Table,
 	targetTable *filter.Table,
 	originTI *model.TableInfo,
+	routeResolver TableRouteResolver,
```
D3Hunter, line 431: #### 🟡 **[Minor]** Exported FK initializer signature changes without a compatibility shim

**Why**
The route-aware change adds a required parameter to the exported `Tracker.InitDownStreamForeignKeyRelations` method. Existing consumers that import `github.com/pingcap/tiflow/dm/pkg/schema` and call this method will fail to compile after upgrade unless they update source at the same time.

**Scope**
dm/pkg/schema/tracker.go:426

**Risk if unchanged**
Out-of-repo DM tools, tests, or downstream integrations using this exported package API can hit an avoidable source-compatibility break during dependency upgrades.

**Evidence**
The diff changes `InitDownStreamForeignKeyRelations` from four arguments to five by adding `routeResolver TableRouteResolver`; the in-repo syncer call was updated at `dm/syncer/syncer.go:3833`, but there is no old-signature wrapper for external callers.

**Change request**
Can we keep the existing exported method as a wrapper that passes `nil`, and add a separate route-aware helper/method for callers that need routing, or explicitly confirm this package has no external compatibility promise?

D3Hunter (commented):
### Summary

- Total findings: 6
- Inline comments: 6

<details>

#### ⚠️ **[Major]** (4)

1. FK metadata mismatch uses a downstream-fetch error code (`dm/pkg/schema/tracker.go:666, dm/pkg/terror/error_list.go:1384`)
2. Route-aware FK causality accepts unsupported many-to-one routes (`dm/pkg/schema/tracker.go:441, dm/pkg/schema/tracker.go:630, dm/pkg/schema/tracker.go:707, dm/pkg/schema/tracker_test.go:1002, dm/pkg/schema/tracker_test.go:1020, pkg/sqlmodel/causality.go:30`)
3. Route-aware FK relations are unreachable from the syncer (`dm/syncer/syncer.go:3830, dm/syncer/syncer.go:3833, dm/pkg/schema/tracker.go:441`)
4. Route-aware FK API is exposed without a production owner (`dm/pkg/schema/tracker.go:94, dm/pkg/schema/tracker.go:425, dm/pkg/schema/tracker.go:441, dm/syncer/syncer.go:3833, dm/pkg/schema/tracker_test.go:1020`)

#### 🟡 **[Minor]** (2)

1. ParentTable docs do not state the source-domain route behavior (`pkg/sqlmodel/foreign_key.go:20, dm/pkg/schema/tracker.go:733`)
2. Exported FK initializer signature changes without a compatibility shim (`dm/pkg/schema/tracker.go:426`)

</details>

dm/pkg/schema/tracker.go
OliverS929: The user-facing message is intentionally schema-alignment specific here, and I kept the existing schema-tracker error code to preserve the current repair guidance. When routed FK causality is wired in in the next PR, we can align this error contract together with the route/topology validation errors.

dm/pkg/schema/tracker.go
OliverS929: The task-wide many-to-one/shared-target check belongs to the syncer topology guard. The next PR will reject that FK routing topology before calling the route-aware builder.

dm/syncer/syncer.go
OliverS929: Yes, that is intentional for this PR. This change only prepares the schema-tracker relation builder; the syncer route/topology precheck and non-nil resolver wiring are handled in the next PR.

dm/pkg/schema/tracker.go
OliverS929: Yes, passing nil here is intentional because this PR does not change production behavior. In the next PR, after the syncer route/topology precheck passes, the syncer will pass a real route resolver.

dm/pkg/schema/tracker.go
OliverS929: Clarified in 0304d3659.

dm/pkg/schema/tracker.go
OliverS929: I checked the in-repo call sites and public GitHub code search, and only tiflow itself shows up. I think this is a DM internal package API rather than a compatibility surface, so I’d keep the current signature unless there is a known external caller we need to preserve.

dm/pkg/schema/tracker.go
```diff
@@ -574,31 +605,44 @@ func (tr *Tracker) buildForeignKeyRelations(
...
+					"foreign key causality with route under foreign_key_checks=1 and worker_count>1 is not supported yet; child table %s routes to %s; please use worker_count=1",
+					sourceTable,
+					targetTable,
+				),
+			)
+		}
 		return nil, newForeignKeyRouteUnsupportedError(
 			fmt.Sprintf(
-				"foreign key causality with route under foreign_key_checks=1 and worker_count>1 is not supported yet; child table %s routes to %s; please use worker_count=1",
+				"foreign key causality with route under foreign_key_checks=1 and worker_count>1 requires 1:1 route alignment; source child table %s routes to %s, but downstream child table is %s",
```
D3Hunter, line 650: above error states `foreign key causality with route under foreign_key_checks=1 and worker_count>1` **is not supported yet;**, but here states it `requires 1:1 route alignment`

that's confusing, maybe clearify more about what's not supported

D3Hunter, line 650: can you write a comment which contains examples to explain which cases are supported , which are not

dm/pkg/schema/tracker.go
```diff
@@ -653,11 +704,24 @@ func (tr *Tracker) buildForeignKeyRelations(
 			targetParentSchema = targetTable.Schema
 		}
 		targetParentTable := &filter.Table{Schema: targetParentSchema, Name: fk.RefTable.O}
-		if !sameTableIdentity(sourceParentTable, targetParentTable) {
+		routedSourceParent := resolveTableRoute(routeResolver, sourceParentTable)
+		if !sameTableIdentity(routedSourceParent, targetParentTable) {
+			if routeResolver == nil {
+				return nil, newForeignKeyRouteUnsupportedError(
+					fmt.Sprintf(
+						"foreign key causality with route under foreign_key_checks=1 and worker_count>1 is not supported yet; upstream parent table %s and downstream parent table %s do not align for FK %s on table %s; please use worker_count=1",
```
D3Hunter, line 723: doitto

D3Hunter (commented):
rest lgtm

dm/pkg/schema/tracker.go
```diff
@@ -804,6 +879,17 @@ func sameTableIdentity(a *filter.Table, b *filter.Table) bool {
 	return a.Schema == b.Schema && a.Name == b.Name
 }
 
+func resolveTableRoute(routeResolver TableRouteResolver, table *filter.Table) *filter.Table {
+	if routeResolver == nil || table == nil {
```
lance6716: AI is too defensive? It will not be nil. We can add the prerequisite in function comments.

dm/pkg/schema/tracker.go
```diff
@@ -765,26 +830,36 @@ func findMatchingForeignKey(originTI *model.TableInfo, downstreamFK *model.FKInf
 
 	if downstreamFK.Name.L != "" {
 		if fk := model.FindFKInfoByName(originTI.ForeignKeys, downstreamFK.Name.L); fk != nil {
-			return fk
+			if sameForeignKeyColumns(fk, downstreamFK) {
+				return fk
+			}
+			return nil
```
lance6716, line 855: should we continue to match other FK?

dm/pkg/schema/tracker.go
```diff
@@ -765,26 +830,36 @@ func findMatchingForeignKey(originTI *model.TableInfo, downstreamFK *model.FKInf
...
 	}
 
 	return nil
 }
 
+func sameForeignKeyColumns(sourceFK *model.FKInfo, downstreamFK *model.FKInfo) bool {
+	if sourceFK == nil || downstreamFK == nil {
+		return false
+	}
+	return sameColumns(sourceFK.Cols, downstreamFK.Cols) && sameColumns(sourceFK.RefCols, downstreamFK.RefCols)
```
GMHDBJD, line 898: Could findMatchingForeignKey include the referenced table/schema (after route resolution) in the match, so FK name/order changes do not select the wrong source FK or false-reject a valid candidate?

dm/pkg/schema/tracker.go
OliverS929: I prefer not to continue when a same-name FK exists. This PR allows route to rename schema/table, but it does not allow FK columns to change. If a same-name FK uses different child or referenced columns, the source and downstream FK metadata are not aligned, and matching another FK could hide that mismatch. We still try column-based matching when there is no same-name source FK, so different FK names with the same FK columns can still work.

dm/pkg/schema/tracker.go
OliverS929: The nil-resolver check is intentional because the current caller still uses the no-route path. I inlined the helper and kept that check at the two call sites in 0491da0 because the table argument is not nil in the current call path. If we later have a nil-table caller, we can add it back then.

dm/pkg/schema/tracker.go
OliverS929: Comments added in 0491da0 for clarification.

OliverS929: Sure. Comments added in 0491da0.

dm/pkg/schema/tracker.go
OliverS929: Added comments in 0491da0.

dm/pkg/schema/tracker.go
OliverS929: Good point. I updated the fallback matching to include the referenced table when a route resolver is available, so it won't pick another source FK just because the child/ref columns match. Same-name FK still does not fall back to another FK. Added tests for the reordered/name-changed case, same-name no-fallback, and the nil-resolver error classification.

Fixed in 2fa94b747.

dm/pkg/schema/tracker.go
OliverS929: cc @alastori for one behavior choice here.

If source and downstream have a same-name FK but the referenced table does not align, I think we should keep that FK strict and not fall back to another source FK. 

For example:

source:
  fk_a(parent_id) -> p1(id)
  fk_b(parent_id) -> p2(id)

downstream:
  fk_a(parent_id) -> p2_r(id)

route:
  p2 -> p2_r

Even though source `fk_b` would match the downstream parent after routing, using it would silently match a different FK under the downstream `fk_a` name.


This affects user-visible schema/route alignment behavior, so I’d like to hear your view.

alastori: Let me confirm the route-aware builder behavior with a concrete example.

Entity relationship:
- In the source schema, `orders.sales_rep_id` is intended to reference `sales_rep(id)`.
- The source schema also has another possible FK relationship from `orders.sales_rep_id` to `employee(id)`.
- The route config maps source table `employee` to downstream table `employee_r`.

Constraints:
- Source has `fk_orders_sales_rep(sales_rep_id)` referencing `sales_rep(id)`.
- Source also has `fk_orders_employee(sales_rep_id)` referencing `employee(id)`.
- Downstream has `fk_orders_sales_rep(sales_rep_id)` referencing `employee_r(id)`. This keeps the FK name `fk_orders_sales_rep`, but changes the parent table from `sales_rep` to `employee_r`.
- The route maps `employee` to `employee_r`.

In this example, `fk_orders_employee` is a tempting fallback because it uses the same child column `sales_rep_id`, the same referenced column `id`, and its parent table `employee` routes to the downstream parent table `employee_r`. But I do not think DM should use that fallback here. Because the downstream FK keeps the name `fk_orders_sales_rep`, users will expect it to correspond to source `fk_orders_sales_rep`, not source `fk_orders_employee`.

Counterexample that should still work:
- Source has `fk_orders_employee(sales_rep_id)` referencing `employee(id)`.
- Downstream has a regenerated name like `fk_orders_emp_2026(sales_rep_id)` referencing `employee_r(id)`.
- The route maps `employee` to `employee_r`.
- There is no same-name FK in the source.

In that case, matching by child columns, referenced columns, and routed parent table makes sense. The FK relationship is the same, only the FK name changed.

If this interpretation is correct, I agree with the current behavior. When the same FK name exists on both sides, DM should require that specific FK relationship to match. If it does not match, returning a clear error is better than silently using another source FK. That makes schema or route mistakes visible to the user.

Suggested error wording:

same-name foreign key does not match after routing: table %s has foreign key %s, source parent table %s routes to %s, but downstream parent table is %s. DM with parallel workers will not match a different source foreign key when the source and downstream foreign keys have the same name. Please align the foreign key definition or route rule.

D3Hunter (approved):

PR #12656, opened 2026-05-28
[DNM]: test

joechenrh:
Test-fix PR based on #12654.

Purpose: validate whether the failing DM integration CI is related to the route-aware FK relation path not being wired into syncer.

Change: pass syncer's route resolver into schema tracker FK relation initialization, so the new route-aware builder can validate route alignment instead of always using identity route.

Issue Number: ref #12654

Tests:
- git diff --check
- Local go test was attempted but stopped because first-time dependency compilation did not finish quickly; leaving full validation to CI.

Release note: None


joechenrh:
Closing this test PR because it was based on #12654. I will create a separate master-based CI baseline PR instead.

PR #12674, opened 2026-06-04
*: update tidb and parser

### LINKED ISSUE #12672: update tidb depedencies

wk989898:

Issue Number: close #12672

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note

```


go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/pingcap/tiflow
 
-go 1.25.8
```
3AceShowHand, line 3: why bump this ?

3AceShowHand (approved):

OliverS929 (commented):
pingcap/tidb#67778 changes schemacmp so charset/collation comparison is no longer strict singleton matching: utf8/utf8mb3 -> utf8mb4 and latin1 -> utf8mb4 are treated as compatible upgrades, while unrelated charset/collation families remain incompatible.

My current understanding is that this is explainable for DM optimistic shard DDL, since DM also relies on schemacmp semantics here. The direction looks reasonable to me, but could @D3hunter and @gmhdbjd help confirm whether this matches the intended DM behavior?

A follow-up could also consider covering the classic DM optimistic shard DDL path with a small charset/collation case, so this behavior is pinned on the DM side as well.

OliverS929:
Since this PR also changes engine-related files, I think @D3hunter's approval is needed before merging.

OliverS929:
> [pingcap/tidb#67778](https://github.com/pingcap/tidb/pull/67778) changes schemacmp so charset/collation comparison is no longer strict singleton matching: utf8/utf8mb3 -> utf8mb4 and latin1 -> utf8mb4 are treated as compatible upgrades, while unrelated charset/collation families remain incompatible.
> 
> My current understanding is that this is explainable for DM optimistic shard DDL, since DM also relies on schemacmp semantics here. The direction looks reasonable to me, but could @D3Hunter and @GMHDBJD help confirm whether this matches the intended DM behavior?
> 
> A follow-up could also consider covering the classic DM optimistic shard DDL path with a small charset/collation case, so this behavior is pinned on the DM side as well.

@alastori PTAL.

GMHDBJD (approved):

alastori:
@OliverS929 Thanks for the context. Non-blocking PM note from the customer-impact angle.

The direction makes sense for rolling charset upgrades in sharded DM tasks. A common customer scenario is one shard moving from utf8/utf8mb3 to utf8mb4 before the others; treating that as a compatible widening avoids a false optimistic DDL conflict.

The main caveat I want to make explicit is latin1. I think we should frame latin1 -> utf8mb4 as schema compatibility, not as a data conversion guarantee. For example, a latin1 value like Café may be stored as bytes 43 61 66 E9, while the same value in utf8mb4 is 43 61 66 C3 A9. If DM treats the schema change as compatible, can we confirm how full-load and incremental paths handle existing latin1 bytes after the downstream schema is widened? I want to make sure we are not accidentally changing equality, uniqueness, sorting, or query behavior for users with non-ASCII latin1 data.

For test coverage, I would prioritize the customer-facing boundaries that should remain incompatible:

- Collation changes that alter comparison semantics. Example: utf8mb4_bin vs utf8mb4_general_ci, where values like A and a may compare differently.
- Unrelated charset families. Example: utf8mb4 vs gbk should still conflict because this is not a simple widening path.

A lower-priority follow-up is table-default-only changes, for example ALTER TABLE ... DEFAULT CHARACTER SET utf8mb4, because changing the table default does not necessarily convert existing explicit columns.

PR #12675, opened 2026-06-05
dm: support FK causality for one-to-one routes

OliverS929:

Issue Number: ref #12350

Support DM FK causality when `foreign_key_checks=1` and `syncer.worker-count > 1` with static one-to-one route rules.

This PR:
- builds FK causality relations with source table identity after route validation
- makes causality key generation follow the task case-sensitive setting, including routed FK causality keys
- respects the task case-sensitive setting when matching FK metadata and checking route topology
- rejects route rules that map multiple source tables in the task to the same target table
- rejects compact and multiple-rows DML boundary options when `foreign_key_checks=1`
- caches successful route topology checks and invalidates the cache after schema/config state changes
- adds unit and integration test coverage for routed multi-worker FK causality

A follow-up PR will explicitly reject hot updates that can change route/filter/case-sensitive/FK causality mode.

#### Tests 

 - [x] Unit test
 - [x] Integration test
 - [ ] No code

### Release note 

```release-note
Support foreign key causality for static one-to-one routing in DM multi-worker sync.
```


dm/pkg/schema/tracker.go
```diff
@@ -928,6 +935,28 @@ func sameTableIdentity(a *filter.Table, b *filter.Table) bool {
 	return a.Schema == b.Schema && a.Name == b.Name
 }
 
+func sameForeignKeyTableIdentity(a *filter.Table, b *filter.Table, caseSensitive bool) bool {
+	a = normalizeForeignKeyTable(a, caseSensitive)
+	b = normalizeForeignKeyTable(b, caseSensitive)
+	return sameTableIdentity(a, b)
+}
```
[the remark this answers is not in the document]
OliverS929: `sameTableIdentity` already handles nil inputs, so `normalizeForeignKeyTable` returning nil here will not panic. I don't think an extra nil check changes the behavior.

dm/syncer/dml.go
```diff
@@ -305,13 +308,24 @@ RowLoop:
...
 
+func (s *Syncer) causalityKeySourceTableNameForRowChange(sourceTable *filter.Table) *cdcmodel.TableName {
+	if !s.needForeignKeyCausality() || s.cfg.CaseSensitive {
+		return nil
+	}
+	return &cdcmodel.TableName{
+		Schema: strings.ToLower(sourceTable.Schema),
+		Table:  strings.ToLower(sourceTable.Name),
+	}
+}
```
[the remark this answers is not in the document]
OliverS929: `sourceTable` is built from the rows event before this helper is called, and the same value is already used to build the `RowChange`. Nil is not a valid input for this path.

dm/syncer/syncer.go
```diff
@@ -3727,26 +3747,95 @@ func (s *Syncer) needForeignKeyCausality() bool {
...
+// foreignKeyRouteTopologyTableID returns the table ID used for one routed target table.
+func (s *Syncer) foreignKeyRouteTopologyTableID(table *filter.Table) string {
+	if table == nil || s.cfg.CaseSensitive {
+		return utils.GenTableID(table)
+	}
+	return utils.GenTableID(&filter.Table{
+		Schema: strings.ToLower(table.Schema),
+		Name:   strings.ToLower(table.Name),
+	})
+}
```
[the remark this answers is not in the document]
OliverS929: The current caller does not pass nil here.  I removed that branch instead in f58f57d.

OliverS929:
cc @joechenrh PTAL.

dm/syncer/dml_worker.go
```diff
@@ -64,13 +64,7 @@ type DMLWorker struct {
...
 func isForeignKeyChecksEnabled(session map[string]string) bool {
-	for key, value := range session {
-		if strings.EqualFold(key, "foreign_key_checks") {
-			trimmed := strings.Trim(value, " '\"")
-			return variable.TiDBOptOn(trimmed)
-		}
-	}
-	return false
+	return config.IsForeignKeyChecksEnabled(session)
 }
```
joechenrh: Why not removing this function

dm/config/subtask.go
```diff
@@ -414,6 +414,9 @@ func (c *SubTaskConfig) Adjust(verifyDecryptPassword bool) error {
 	} else if c.SyncerConfig.SafeMode && duration == 0 {
 		return terror.ErrConfigConfictSafeModeDurationAndSafeMode.Generate()
 	}
+	if err := CheckUnsupportedForeignKeyChecksSyncerOptions(c.To.Session, c.SyncerConfig); err != nil {
```
D3Hunter:
```suggestion
	if err := CheckFKChecksOptions(c.To.Session, c.SyncerConfig); err != nil {
```

dm/pkg/schema/tracker.go
```diff
@@ -410,7 +410,7 @@ func (tr *Tracker) GetDownStreamTableInfo(tctx *tcontext.Context, tableID string
 		return nil, err
 	}
 
-	if err := dti.initForeignKeyRelations(tr, tctx, tableID, targetTable, targetTable, originTI, nil); err != nil {
+	if err := dti.initForeignKeyRelations(tr, tctx, tableID, targetTable, targetTable, originTI, nil, true); err != nil {
```
D3Hunter, line 413: always true? ~is this related to the feature supported in this PR~

tidb table name is not case sensitive

dm/pkg/schema/tracker.go
```diff
@@ -766,12 +772,13 @@ func (tr *Tracker) buildForeignKeyRelations(
 		parentNameToIdx := buildColumnIndexMap(parentOriginTI)
 		parentRelations, err := tr.buildForeignKeyRelations(
 			tctx,
-			parentTableID,
+			parentDownstreamTableID,
```
D3Hunter, line 775: `parentDownstreamTableID` `targetParentTable` `parentDTI`, seems the same thing

dm/pkg/schema/tracker_test.go
```diff
@@ -1156,7 +1223,117 @@ func TestForeignKeyRelationMatchesReferencedTableWhenNameAndOrderDiffer(t *testi
...
+
+	relation := dti.ForeignKeyRelations[0]
+	require.Equal(t, (&cdcmodel.TableName{Schema: "db", Table: "p2"}).String(), relation.ParentTable)
+	require.Equal(t, []int{0}, relation.ChildColumnIdx)
+	require.Len(t, relation.ParentColumns, 1)
+	require.Equal(t, "id", relation.ParentColumns[0].Name.L)
+	require.NoError(t, mock.ExpectationsWereMet())
+}
+
+func TestForeignKeyRelationMatchesReferencedSchemaWhenNameAndOrderDiffer(t *testing.T) {
```
D3Hunter: too long to be read, maybe use shorter names for top test name, and use subtests to use more describtive names
```
TestForeignKeyRelationMatchesReferencedSchemaWhenNameAndOrderDiffer
TestForeignKeyRelationMatchesReferencedTableWithoutRouteWhenNameAndOrderDiffer
TestForeignKeyRelationRejectsReferencedTableMismatchWithoutRoute
```

dm/syncer/dml_worker.go
```diff
@@ -64,13 +64,7 @@ type DMLWorker struct {
...
 
 func isForeignKeyChecksEnabled(session map[string]string) bool {
```
D3Hunter: can be removed now

dm/tests/foreign_key_route_multi_worker/data/source1.prepare.sql
D3Hunter, line 1: make the file name the same as the inner used db name `fk_route_src`

dm/tests/foreign_key_route_multi_worker/run.sh
D3Hunter, line 1: ~seems only one worker is started, what the name `foreign_key_route_multi_worker` mean~

dm/syncer/dml.go
```diff
@@ -305,13 +308,24 @@ RowLoop:
...
+		rowChange.SetCausalityKeySourceTable(s.causalityKeySourceTableNameForRowChange(param.sourceTable))
 		rowChange.SetForeignKeyRelations(downstreamTableInfo.ForeignKeyRelations)
 		dmls = append(dmls, rowChange)
 	}
 
 	return dmls, nil
 }
 
+func (s *Syncer) causalityKeySourceTableNameForRowChange(sourceTable *filter.Table) *cdcmodel.TableName {
+	if !s.needForeignKeyCausality() || s.cfg.CaseSensitive {
```
D3Hunter, line 323: seems no need to compute `needForeignKeyCausality` for each row, we can store it somewhere

dm/syncer/dml.go
```diff
@@ -305,13 +308,24 @@ RowLoop:
...
 		rowChange.SetWhereHandle(downstreamTableInfo.WhereHandle)
+		rowChange.SetCausalityKeySourceTable(s.causalityKeySourceTableNameForRowChange(param.sourceTable))
 		rowChange.SetForeignKeyRelations(downstreamTableInfo.ForeignKeyRelations)
 		dmls = append(dmls, rowChange)
 	}
 
 	return dmls, nil
 }
 
+func (s *Syncer) causalityKeySourceTableNameForRowChange(sourceTable *filter.Table) *cdcmodel.TableName {
```
D3Hunter, line 322:
```suggestion
func (s *Syncer) getCausalityKeySourceTableName(sourceTable *filter.Table) *cdcmodel.TableName {
```

dm/tests/foreign_key_route_multi_worker/run.sh
D3Hunter, line 1: maybe name as `fk_route_multi_dml_worker`

dm/syncer/syncer.go
```diff
@@ -170,6 +171,9 @@ type Syncer struct {
 	exprFilterGroup *ExprFilterGroup
 	sessCtx         sessionctx.Context
 
+	foreignKeyRouteTopologyMu      sync.Mutex
```
D3Hunter:
```suggestion
	fkRouteShapeMu      sync.Mutex
```
or `Mapping` as AI suggested, `Topology` is a hard to know what it represents

and those names are too long, maybe abbr `foreignKey` as `fk`, it's quite common abbr

and please comment that we expect 1:1 source-> target route shape

dm/syncer/foreign_key_causality_route_test.go
```diff
@@ -133,14 +185,400 @@ func TestPrepareDownStreamTableInfoRejectsChildRouteForMultiWorkerFKCausality(t
...
+
+func TestPrepareDownStreamTableInfoBuildsFKRelationsWithChildAndParentRoute(t *testing.T) {
+	syncer, mock := newForeignKeyRouteTestSyncer(t, 2)
+	setForeignKeyRouteTestRoutes(t, syncer, []*router.TableRule{
+		{SchemaPattern: "db", TablePattern: "parent", TargetSchema: "db_r", TargetTable: "parent_r"},
+		{SchemaPattern: "db", TablePattern: "child", TargetSchema: "db_r", TargetTable: "child_r"},
+	})
+	p := parser.New()
+
+	execTrackedDDL(t, syncer.schemaTracker, p, "", "create database db")
```
D3Hunter: maybe abstract those common pattern a bit, such as `prepareSimpleParentChildSchema`

pkg/sqlmodel/row_change.go
```diff
@@ -58,6 +58,8 @@ func (t RowChangeType) String() string {
 type RowChange struct {
 	sourceTable *cdcmodel.TableName
 	targetTable *cdcmodel.TableName
+	// Optional source table name used only by CausalityKeys.
+	causalityKeySourceTable *cdcmodel.TableName
```
D3Hunter, line 62: can we set `sourceTable` to the right table directly, so no need to add a separate field

dm/syncer/syncer.go
```diff
@@ -3727,26 +3747,95 @@ func (s *Syncer) needForeignKeyCausality() bool {
...
+}
+
+func (s *Syncer) isForeignKeyRouteTopologyChecked(cacheKey string) bool {
+	s.foreignKeyRouteTopologyMu.Lock()
+	defer s.foreignKeyRouteTopologyMu.Unlock()
+	_, ok := s.foreignKeyRouteTopologyChecked[cacheKey]
+	return ok
+}
+
+func (s *Syncer) markForeignKeyRouteTopologyChecked(cacheKey string) {
```
D3Hunter: the key is always the same, what's the purpose to make a map?
add comment to explain this, if for future extension, we can do it later

dm/syncer/foreign_key_causality_route_test.go
```diff
@@ -88,6 +114,31 @@ func expectDownstreamSQLModeInit(mock sqlmock.Sqlmock) {
...
+}
+
+func requireNextCausalityJob(t *testing.T, causalityCh <-chan *job) *job {
+	t.Helper()
+
+	select {
+	case job, ok := <-causalityCh:
+		require.True(t, ok)
+		return job
+	case <-time.After(3 * time.Second):
```
joechenrh, line 144: Will this make the test flaky?

dm/syncer/foreign_key_causality_route_test.go
```diff
@@ -133,14 +185,400 @@ func TestPrepareDownStreamTableInfoRejectsChildRouteForMultiWorkerFKCausality(t
...
+	mock.ExpectQuery("SHOW CREATE TABLE " + utils.GenTableID(parentTarget)).WillReturnRows(
+		sqlmock.NewRows([]string{"Table", "Create Table"}).
+			AddRow("parent_r", "create table parent_r(id int primary key)"),
+	)
+	expectFetchAllDoTables(mock, "DB", "Parent", "Child")
+
+	parentDMLs, err := syncer.genAndFilterInsertDMLs(tcontext.Background(), &genDMLParam{
+		sourceTable:     parentSource,
+		targetTable:     parentTarget,
+		originalData:    [][]interface{}{{10}},
```
joechenrh: nit: replace `interface{}` with `any`

dm/syncer/syncer.go
```diff
@@ -3727,26 +3747,95 @@ func (s *Syncer) needForeignKeyCausality() bool {
 	return isForeignKeyChecksEnabled(s.cfg.To.Session) && s.cfg.WorkerCount > 1
 }
 
-func (s *Syncer) precheckForeignKeyCausality(
-	sourceTable *filter.Table,
-	targetTable *filter.Table,
-	downstreamTableInfo *schema.DownstreamTableInfo,
-) error {
-	if downstreamTableInfo == nil || downstreamTableInfo.TableInfo == nil {
+const foreignKeyRouteTopologyTaskCacheKey = "task"
```
joechenrh: Will we have other task keys in the future?

dm/pkg/schema/tracker.go
OliverS929: This `true` is only for the legacy direct tracker API. Its current callers are tracker tests / no-route helper paths with exact-case or lowercase fixtures. The syncer production path in this PR goes through `InitDownStreamForeignKeyRelations(..., s.cfg.CaseSensitive)`, so the real task config is still used there. TiDB table lookup is generally case-insensitive, but this flag controls DM's FK table identity matching, not the physical downstream lookup itself.

dm/pkg/schema/tracker.go
OliverS929: They are related but not the same thing: `targetParentTable` is the downstream parent table from FK metadata, `parentDownstreamTableID` is its schema-cache key, and `parentDTI` is the cached downstream table info. The comment above this block is meant to clarify this source/downstream split.

dm/pkg/schema/tracker_test.go
OliverS929: I suggest we keep them as top-level tests because they cover separate FK metadata matching cases with different setup/route paths, and the explicit names show the exact regression scenario being tested. This feels more direct for these cases.

dm/tests/foreign_key_route_multi_worker/data/source1.prepare.sql
OliverS929: I used `db1.prepare.sql` to follow the existing DM integration naming convention for source-1 SQL files. The schema name itself is still `fk_route_src` in the SQL and task config.

dm/tests/foreign_key_route_multi_worker/run.sh
OliverS929: Here `multi_worker` refers to the syncer `worker-count` / DML workers, not the number of DM worker processes. This case uses `worker-count: 4`; the existing `foreign_key_route_single_worker` case covers the `worker-count: 1` path.

OliverS929: I used the current name to keep it consistent with the existing `foreign_key_*` integration cases. The `multi_worker` part means syncer `worker-count > 1` here. If we want to rename this style, I’d prefer to handle it later together with the existing FK integration case names.

dm/syncer/syncer.go
OliverS929: The 1:1 expectation is documented on `foreignKeyRouteTopologyChecked` and above the `targetToSources` block. 

For the names, `Topology` is meant to describe the task-level route graph rather than one route rule mapping, and the full `foreignKey` wording follows the surrounding FK-related helper names. If you still prefer the shorter `fk` form here, I can adjust it.

dm/syncer/foreign_key_causality_route_test.go
OliverS929: My concern with a higher-level helper here is that the nearby cases vary in schema case, route rules, target table names, and downstream FK definitions. The shared lower-level pieces are already covered by `execTrackedDDL` and `expectFetchAllDoTables`.

pkg/sqlmodel/row_change.go
OliverS929: I think we should keep these two identities separate here. `sourceTable` is still the real row-change source table and is used by `GetSourceTable()`, DML job metadata, validator/checkpoint paths, and logs. This field only overrides the table name used by `CausalityKeys()`, so the FK causality key can use the normalized source-domain identity without changing the row change's source table semantics.

dm/syncer/foreign_key_causality_route_test.go
OliverS929: I don't think this adds a flaky timing dependency. The timeout is meant as a fail-fast guard if the causality goroutine does not produce output. The test has already put both jobs into a buffered channel and closed it, so it is expected that the receive is waiting for local goroutine processing rather than an external timing condition. Without the timeout, a bug in the causality path could leave the test hanging.

dm/syncer/syncer.go
OliverS929: No, not in the current design. The map/key came from an earlier version that cached checks per routed target set for the current FK graph, instead of one task-level check. We moved back to task-level checking because scoped checks need more support to catch parent-only DML cases early enough. With the current task-level check, there is only one checked state, so I simplified it to a single checked flag in d8c444975e. If we further extend this to scoped FK-graph checks later, we can bring back keyed cache behavior together with the needed parent-side detection.

dm/syncer/foreign_key_causality_route_test.go
OliverS929: Changed in d8c444975e.

dm/syncer/dml_worker.go
OliverS929: I kept the wrapper initially to avoid moving the existing syncer-side coverage while moving the parsing helper into `config`. I’ve moved that coverage to the config package and removed the wrapper in d8c444975e.

dm/syncer/syncer.go
OliverS929: The map/key came from an earlier version that cached checks per routed target set for the current FK graph, instead of one task-level check. We moved back to task-level checking because scoped checks need more support to catch parent-only DML cases early enough. With the current task-level check, there is only one checked state, so I simplified it to a single checked flag in d8c444975e. 

If we further extend this to scoped FK-graph checks later, the keyed cache can come back with the needed parent-side detection.

dm/syncer/dml.go
OliverS929: The current name is meant to tie this helper to the `RowChange.SetCausalityKeySourceTable` hook. It only changes the source table used by `CausalityKeys()`, not the row change source table itself. If the shorter name is not a must-change for you, I'd prefer to keep this one.

dm/syncer/dml.go
OliverS929: I moved this to a local value computed once before the row loop in d8c444975e.

dm/syncer/dml_worker.go
OliverS929: Removed in d8c444975e. The old wrapper coverage was moved to the config package.

dm/config/subtask.go
OliverS929: Renamed in d8c444975e.

dm/tests/foreign_key_route_multi_worker/data/source1.prepare.sql
D3Hunter: i guess existing are wrong, please remove the `db1` prefix or rename using real schema name or `source1`

joechenrh (approved):

dm/syncer/foreign_key_causality_route_test.go
D3Hunter: just let AI try to do it without hurting readibility

dm/tests/foreign_key_route_multi_worker/data/source1.prepare.sql
OliverS929: Changed in 996dcf1994.

D3Hunter (commented):
rest lgtm

dm/syncer/foreign_key_causality_route_test.go
OliverS929: Changed in ccd9638780.

D3Hunter (approved):

Benjamin2037 (approved):
LGTM

yudongusa (approved):

PR #12697, opened 2026-06-10
syncer, sqlmodel(dm): materialize generated columns for unique expression indexes

### LINKED ISSUE #12696: DM: UNIQUE expression index crashes dm-worker
### What did you do?

1. Start playground

    ```sh
    tiup playground-ng:nightly v8.5.6 --db 1 --pd 1 --kv 1 --dm-master 1 --dm-worker 1 --tiflash 0
    ```

2. Create a source

    ```yaml
    # m1.yaml
    source-id: m1
    from:
      host: '192.168.64.3'  # <- note: should be a MySQL 8, as expression index is involved.
      port: 3306
      user: 'root'
      password: '111'
    ```
    ```sh
    tiup dmctl:v8.5.6 --master-addr=127.0.0.1:8261 operate-source create ./m1.yaml
    ```

3. Create a task

    ```yaml
    # t1.yaml
    name: t1
    task-mode: all
    target-database:
      host: '127.0.0.1'
      port: 4000
      user: 'root'
      password: ''
    block-allow-list:
      ba1:
        do-dbs: ["test"]
    mysql-instances:
    - source-id: m1
      block-allow-list: ba1
    ```
    ```sh
    tiup dmctl:v8.5.6 --master-addr=127.0.0.1:8261 start-task ./t1.yaml
    ```

4. Create a table with unique expression index.

    ```sql
    create schema if not exists test;
    create table if not exists test.t1 (
      id bigint primary key,
      name varchar(255),
      unique key only_one_alice ((case name when 'Alice' then 1 else null end))
    );
    ```

5. Write some DMLs.

    ```sql
    insert into test.t1 values (1, 'Alice'), (2, 'Bob'), (3, 'Bob'), (4, 'Bob'), (5, 'Bob');
    ```

Step 5 replicates the DMLs to downstream.

### What did you see instead?

dm-worker simply crashed (Exit code 2) without any logs.

### Versions of the cluster

v8.5.6

_No response_

joechenrh:

Issue Number: close #12696

A `dm-worker` panics with `index out of range` and crash-loops when replicating a table that has an **expression-based (functional) UNIQUE index**. A single affected row takes down the worker process; after restart, the task replays the same binlog row and panics again, so it does not self-recover.

Skipping UNIQUE expression indexes would avoid the panic, but it is not correctness-preserving. For example, two `REPLACE` statements that both hit `UNIQUE KEY ((lower(name)))` must keep their upstream order. If `(20, 'ALICE')` is replicated before `(10, 'Alice')`, the final downstream row can become `(10, 'Alice')` instead of the upstream final row `(20, 'ALICE')`.

This PR keeps expression UNIQUE indexes in DM causality while keeping hidden columns out of WHERE generation.

- `UniqueIdxs` remains the visible-column index set used for WHERE clauses.
- A separate causality index set includes all UNIQUE indexes, including expression indexes.
- DM materializes hidden virtual generated columns by evaluating cached TiDB expressions with downstream SQL mode and timezone.
- Row images are mapped by visible-column order before generated values are written back by `ColumnInfo.Offset`, so hidden columns interleaved by DDL are handled.
- Visible stored generated columns stay on the normal visible-column path and are covered by tests.
- If materialization fails, DM skips the hidden-column causality key and update detection falls back to the visible unique-index path.

#### Tests

 - Unit test: `go test ./pkg/sqlmodel`
 - Unit test: `go test ./pkg/sqlmodel -run 'TestCausalityKeysInterleavedHiddenColumn|TestPrimaryOrUniqueKeyUpdatedInterleavedHiddenColumn' -count=1`
 - Unit test: `go test ./dm/syncer -run '^$'`
 - Unit test with race detector: `go test -race ./pkg/sqlmodel -run 'TestCausalityKeysExpressionIndexNoRace|TestPrimaryOrUniqueKeyUpdatedWithExpressionIndex|TestPrimaryOrUniqueKeyUpdatedExpressionIndexMaterializeFailure|TestCausalityKeysExpressionIndexMaterializeFailure'`
 - Integration test case added: `dm/tests/all_mode`, covering stored generated unique indexes, `ALTER TABLE ... ADD UNIQUE KEY ((expr))`, and inline functional unique indexes in `CREATE TABLE` (`make dm_integration_test CASE=all_mode` not run locally)
 - Manual real-environment verification with MySQL 8 source, TiDB v8.5.5 target, and PR-built `dm-master`/`dm-worker`/`dmctl`:
   - Replicated a table with `UNIQUE KEY ((lower(name)))` in incremental mode.
   - Ran mixed DML (`REPLACE`, `ON DUPLICATE KEY UPDATE`, `UPDATE`, `DELETE`) plus 2,000 repeated `REPLACE` statements over colliding functional unique keys.
   - DM stayed `Running` and reached `synced=true`; source and target row counts and sampled grouped results matched.
   - Worker logs had no `panic`, `index out of range`, or generated-column materialization errors.
   - Also checked a stale-target-schema `admin_groups` scenario: DM reported a task-level column-count mismatch error instead of crashing the worker.

No compatibility break. The WHERE-index selection path still excludes hidden columns. Generated expression evaluation is only attempted for tables that have a hidden-column UNIQUE index; expressions are parsed once per table `WhereHandle` and evaluated per row only when needed.

No.

### Release note

```release-note
fix a dm-worker panic (index out of range) when replicating a table with an expression-based (functional) UNIQUE index
```


pkg/sqlmodel/causality.go
```diff
@@ -239,3 +248,64 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
+	// cached on the (per-table) WhereHandle; here we just evaluate them per row.
+	exprs, ok := r.whereHandle.generatedColumns.getOrBuildExprs(exprCtx)
+	if !ok {
+		return values, false
+	}
+
+	// DM row values are compacted by dropping hidden columns. TiDB keeps hidden
+	// expression-index columns at the end of TableInfo.Columns, so the row image
+	// corresponds to the visible prefix.
+	datums, err := utils.AdjustBinaryProtocolForDatum(r.tiSessionCtx, values, cols[:len(values)])
```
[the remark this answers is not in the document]
joechenrh: I am leaving this one unchanged for now. The eval row needs to preserve TiDB column offsets, and generated columns can depend on other generated columns defined before them, so narrowing the datum cast set would need extra dependency analysis and more bookkeeping. Since this path is only enabled for tables with unique expression indexes, I prefer keeping the conservative full visible-row cast unless we see a measured hot-path issue.

pkg/sqlmodel/causality.go
```diff
@@ -239,3 +248,64 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
+			return values, false
+		}
+		d, err := expr.Eval(exprCtx.GetEvalCtx(), chunk.MutRowFromDatums(fullDatums).ToRow())
+		if err != nil {
+			log.L().Warn("cannot evaluate generated column expression",
+				zap.String("table", r.sourceTable.String()),
+				zap.String("column", col.Name.O), zap.Error(err))
+			return values, false
+		}
+		full[col.Offset] = d.GetValue()
```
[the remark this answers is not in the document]
joechenrh: Done in https://github.com/pingcap/tiflow/commit/678602423f19f1743dca040f10e3aecf454a83e7 . Normalized materialized `[]byte` datum values to `string`, and made the update-key comparison safe for non-comparable values. Also added a VARBINARY expression-index update case to cover this path.

pkg/sqlmodel/causality.go
```diff
@@ -206,20 +208,27 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
+	values, fullValuesOK := r.fillVirtualGeneratedValues(values)
+
 	for _, indexCols := range pkAndUks {
 		// TODO: should not support multi value index and generate the value
 		// TODO: also fix https://github.com/pingcap/tiflow/issues/3286#issuecomment-971264282
 		if indexCols.MVIndex {
 			continue
 		}
+
+		if indexHasHiddenColumn(indexCols, r.sourceTableInfo) && !fullValuesOK {
```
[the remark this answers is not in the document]
joechenrh: Done in https://github.com/pingcap/tiflow/commit/678602423f19f1743dca040f10e3aecf454a83e7 . Added failure-path coverage for both causality key generation and `IsPrimaryOrUniqueKeyUpdated`: if generated-column materialization fails, we do not panic, the expression index is dropped, and ordinary keys still drive the fallback behavior.

pkg/sqlmodel/where_handle.go
```diff
@@ -27,8 +31,44 @@ type WhereHandle struct {
...
 	// NULL, we can still use it.
 	// every index that is UNIQUE should be added to UniqueIdxs, even for
-	// PK and NOT NULL.
+	// PK and NOT NULL. Indexes backed by a hidden column are excluded here.
 	UniqueIdxs []*model.IndexInfo
+	// CausalityIdxs is the superset of UniqueIdxs that also keeps indexes backed
+	// by a hidden column to detect conflicts.
+	CausalityIdxs []*model.IndexInfo
+
+	generatedColumns *generatedColumnCache
```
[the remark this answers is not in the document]
joechenrh: Done in https://github.com/pingcap/tiflow/commit/678602423f19f1743dca040f10e3aecf454a83e7 . Renamed the field to `hiddenGeneratedColumnExprs`, so the nil check and `getOrBuildExprs` call read as an expression cache instead of a column slice.

pkg/sqlmodel/causality.go
```diff
@@ -239,3 +248,64 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
+	full := make([]any, len(cols))
+	copy(full, values)
+	fullDatums := make([]types.Datum, len(cols))
+	copy(fullDatums, datums)
+
+	// A generated column may depend on generated columns defined before it, so
+	// evaluate generated columns in column order after visible values are in
+	// their full TableInfo offsets.
+	for _, col := range cols {
+		if !col.Hidden || !col.IsGenerated() {
```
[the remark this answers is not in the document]
joechenrh: Done in https://github.com/pingcap/tiflow/commit/678602423f19f1743dca040f10e3aecf454a83e7 . The generated-column cache now owns the ordered hidden generated-column list, and both expression build and per-row eval iterate that list instead of duplicating the predicate.

pkg/sqlmodel/where_handle.go
```diff
@@ -27,8 +31,44 @@ type WhereHandle struct {
 	// If the index and columns have no NOT NULL constraint, but all data is NOT
 	// NULL, we can still use it.
 	// every index that is UNIQUE should be added to UniqueIdxs, even for
-	// PK and NOT NULL.
+	// PK and NOT NULL. Indexes backed by a hidden column are excluded here.
 	UniqueIdxs []*model.IndexInfo
+	// CausalityIdxs is the superset of UniqueIdxs that also keeps indexes backed
+	// by a hidden column to detect conflicts.
+	CausalityIdxs []*model.IndexInfo
```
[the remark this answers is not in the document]
joechenrh: Done in https://github.com/pingcap/tiflow/commit/678602423f19f1743dca040f10e3aecf454a83e7 . Updated the `WhereHandle` doc for the causality/expression-eval role, and made `CausalityIdxs` package-private since it is only consumed inside `pkg/sqlmodel`.

pkg/sqlmodel/causality.go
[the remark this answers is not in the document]
joechenrh: Follow-up in https://github.com/pingcap/tiflow/commit/b76d09bfb26251bffd46d2411d30d3e75ad2ee70 . I narrowed this back to the actual new issue: materialized generated-column values normalize `[]byte` to `string`. I dropped the broader `reflect.DeepEqual` comparison change so existing `IsPrimaryOrUniqueKeyUpdated` value-comparison semantics stay unchanged.

dm/pkg/utils/common.go
```diff
@@ -201,6 +201,7 @@ var ZeroSessionCtx sessionctx.Context
 // NewSessionCtx return a session context with specified session variables.
 func NewSessionCtx(vars map[string]string) sessionctx.Context {
 	variables := variable.NewSessionVars(nil)
+	variables.StmtCtx.SetTimeZone(variables.TimeZone)
```
kennytm: Looks like this caused the unit tests to fail because `variables.TimeZone` is nil.

https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/pingcap_tiflow/12697/pull-unit-test-cdc/2064998364511997952

dm/pkg/utils/common.go
```diff
@@ -201,7 +201,6 @@ var ZeroSessionCtx sessionctx.Context
 // NewSessionCtx return a session context with specified session variables.
 func NewSessionCtx(vars map[string]string) sessionctx.Context {
 	variables := variable.NewSessionVars(nil)
-	variables.StmtCtx.SetTimeZone(variables.TimeZone)
```
kennytm: reverting this breaks the expression index test 🤣 

```
=== FAIL: pkg/sqlmodel TestDpanicSuite/TestExpressionIndex (0.00s)
    assert_common.go:58: test panicked: assert failed, location mismatch, ctxLoc: UTC, varsLoc: Asia/Shanghai, stmtLoc: UTC
        goroutine 96 [running]:
```

joechenrh: What a dilemma 😸

pkg/sqlmodel/causality.go
```diff
@@ -239,3 +248,60 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
+		full[col.Offset] = datumValue(d)
+		fullDatums[col.Offset] = d
+		mutRow.SetDatum(col.Offset, d)
+	}
+	return full, true
+}
+
+func datumValue(d types.Datum) any {
+	value := d.GetValue()
+	if bs, ok := value.([]byte); ok {
```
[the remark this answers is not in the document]
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/5ad4143081ba495cf5999de6c48904a8c223a18d . `datumValue` now canonicalizes decimal, bit/binary, temporal, enum/set, JSON, and vector datums before they flow into update-key comparison. I also added unchanged decimal, binary, and bit expression-index UPDATE cases so this does not regress.

pkg/sqlmodel/causality.go
```diff
@@ -239,3 +248,60 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
+
+	mutRow := chunk.MutRowFromDatums(fullDatums)
+	for _, col := range r.whereHandle.hiddenGeneratedColumnExprCache.columns {
+		expr, ok := exprs[col.Offset]
+		if !ok {
+			return values, false
+		}
+		d, err := expr.Eval(exprCtx.GetEvalCtx(), mutRow.ToRow())
+		if err != nil {
+			log.L().Warn("cannot evaluate generated column expression",
```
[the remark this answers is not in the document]
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/5ad4143081ba495cf5999de6c48904a8c223a18d . These materialization failures now log at Debug instead of Warn, so a systematic failure will not flood warn logs on the row path.

pkg/sqlmodel/causality.go
```diff
@@ -206,20 +208,27 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
 		// Only causality keys use this table name; r.sourceTable keeps the original source table.
 		sourceTable = r.causalityKeySourceTable
 	}
-	pkAndUks := r.whereHandle.UniqueIdxs
+	pkAndUks := r.whereHandle.causalityIdxs
```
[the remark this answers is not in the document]
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/5ad4143081ba495cf5999de6c48904a8c223a18d . Renamed the local `pkAndUks` variable to `causalityIndexes` and updated the no-index comment to describe causality indexes instead of only PK/UK.

pkg/sqlmodel/causality.go
```diff
@@ -239,3 +248,60 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
+			return values, false
+		}
+		full[col.Offset] = datumValue(d)
+		fullDatums[col.Offset] = d
+		mutRow.SetDatum(col.Offset, d)
+	}
+	return full, true
+}
+
+func datumValue(d types.Datum) any {
```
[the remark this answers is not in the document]
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/5ad4143081ba495cf5999de6c48904a8c223a18d . Added the comment around byte-slice normalization, and broadened the normalization to other non-value-stable datum kinds as well.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,94 @@
...
+		c.exprs = exprs
+		c.ok = true
+	})
+	return c.exprs, c.exprCtx, c.ok
+}
+
+func generatedColumnExprContext(tiSessionCtx sessionctx.Context) *exprstatic.ExprContext {
+	vars := tiSessionCtx.GetSessionVars()
+	charset, collation := vars.GetCharsetInfo()
+	// TODO(joechenrh): Carry downstream charset/collation when needed.
```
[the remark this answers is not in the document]
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/5ad4143081ba495cf5999de6c48904a8c223a18d . Made the TODO removal condition explicit: carry downstream charset/collation once expression-index causality needs charset/collation-sensitive generated expressions.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,94 @@
...
+// WhereHandle is used to generate a WHERE clause in SQL and select causality
+// keys. It also caches generated-column expressions needed to materialize
+// hidden columns backing unique expression indexes.
 type WhereHandle struct {
 	UniqueNotNullIdx *model.IndexInfo
 	// If the index and columns have no NOT NULL constraint, but all data is NOT
 	// NULL, we can still use it.
 	// every index that is UNIQUE should be added to UniqueIdxs, even for
-	// PK and NOT NULL.
+	// PK and NOT NULL. Indexes backed by a hidden column are excluded here.
```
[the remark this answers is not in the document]
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/5ad4143081ba495cf5999de6c48904a8c223a18d . Reworded the `UniqueIdxs` comment so it describes the actual WHERE-eligible subset and points expression indexes to `causalityIdxs`.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,94 @@
...
+		}
+	}
+	return &generatedColumnExprCache{
+		sourceTableInfo: source,
+		columns:         cols,
+	}
+}
+
+func (c *generatedColumnExprCache) getOrBuildExprs(
+	tiSessionCtx sessionctx.Context,
```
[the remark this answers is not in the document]
joechenrh: Documented in https://github.com/pingcap/tiflow/commit/5ad4143081ba495cf5999de6c48904a8c223a18d . I kept the API shape because `WhereHandle` is constructed without a session context, while DM passes a stable syncer apply session context on the row path. The comment now states that the context is used only for the first cache build.

pkg/sqlmodel/reduce.go
```diff
@@ -130,11 +130,25 @@ func (r *RowChange) IsPrimaryOrUniqueKeyUpdated() bool {
...
+		postValues    = r.postValues
+		checkIndexes  = r.whereHandle.UniqueIdxs
+		preOK, postOK bool
+	)
+
+	if r.whereHandle.hiddenGeneratedColumnExprCache != nil {
+		preValues, preOK = r.fillVirtualGeneratedValues(r.preValues)
+		postValues, postOK = r.fillVirtualGeneratedValues(r.postValues)
+		if preOK && postOK {
+			checkIndexes = r.whereHandle.causalityIdxs
```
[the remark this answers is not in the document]
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/5ad4143081ba495cf5999de6c48904a8c223a18d . Added comments at both call sites to make the difference explicit: causality key generation can skip one failed hidden-column index, while update detection needs both pre/post values and therefore falls back to the visible-column index set.

pkg/sqlmodel/causality.go
```diff
@@ -206,20 +208,27 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
-	pkAndUks := r.whereHandle.UniqueIdxs
+	pkAndUks := r.whereHandle.causalityIdxs
 	if len(pkAndUks) == 0 {
 		// the table has no PK/UK, all values of the row consists the causality key
 		return []string{genKeyString(sourceTable.String(), r.sourceTableInfo.Columns, values)}
 	}
 
 	ret := make([]string, 0, len(pkAndUks))
 
+	values, fullValuesOK := r.fillVirtualGeneratedValues(values)
```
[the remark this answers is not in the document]
joechenrh: Leaving this unchanged for now. The same UPDATE row can be materialized more than once in safe mode, but this PR path is limited to expression-unique-index tables and our target scenario does not use FK validation. I prefer not adding mutable per-RowChange cache state until we have a measured hotspot or a broader RowChange lifecycle cleanup.

dm/syncer/syncer.go
```diff
@@ -447,6 +447,12 @@ func (s *Syncer) Init(ctx context.Context) (err error) {
 	vars := map[string]string{
 		"time_zone": s.timezone.String(),
 	}
+	for k, v := range s.cfg.To.Session {
```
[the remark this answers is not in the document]
joechenrh: Partially handled in https://github.com/pingcap/tiflow/commit/5ad4143081ba495cf5999de6c48904a8c223a18d . I added the comment explaining why `sql_mode` is copied into the expression-eval session. I am not adding a broader sql_mode/collation-sensitive integration case in this PR: the current change only wires the configured downstream SQL mode into TiDB expression evaluation, while downstream charset/collation propagation is explicitly left behind the TODO.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,94 @@
...
+// keys. It also caches generated-column expressions needed to materialize
+// hidden columns backing unique expression indexes.
 type WhereHandle struct {
 	UniqueNotNullIdx *model.IndexInfo
 	// If the index and columns have no NOT NULL constraint, but all data is NOT
 	// NULL, we can still use it.
 	// every index that is UNIQUE should be added to UniqueIdxs, even for
-	// PK and NOT NULL.
+	// PK and NOT NULL. Indexes backed by a hidden column are excluded here.
 	UniqueIdxs []*model.IndexInfo
```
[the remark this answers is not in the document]
joechenrh: I am keeping the field name unchanged for API compatibility. `UniqueIdxs` is exported and already used outside `pkg/sqlmodel`; renaming it would widen this fix into an API migration. The updated field comment now states that it is the visible/addressable unique-index subset and that expression indexes live in `causalityIdxs`.

pkg/sqlmodel/causality.go
```diff
@@ -239,3 +249,79 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
 
 	return ret
 }
+
+// fillVirtualGeneratedValues returns a copy of values extended to the full
+// source column list, with hidden/virtual generated columns evaluated.
+func (r *RowChange) fillVirtualGeneratedValues(values []any) ([]any, bool) {
```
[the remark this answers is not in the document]
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/1707f4c0933817f8a9618440d3adc18e0d1c1a59 . Added a short sentence documenting that the bool means full values are available.

dm/tests/all_mode/data/db1.increment.sql
```diff
@@ -24,6 +24,17 @@ update t1 set name = 'gentestxxxxxx', dt = '2021-05-11 12:03:05', ts = '2021-05-
 -- delete with unique key
 delete from t1 where gen_id > 124;
 
+-- test unique functional index backed by a hidden generated column
+create table t_expr_unique (
+    id int primary key,
+    name varchar(64)
+);
+/*!80013 alter table t_expr_unique add unique key uk_lower_name ((lower(name))) */;
```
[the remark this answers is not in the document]
joechenrh: Leaving this unchanged for now. The functional-index DDL is intentionally guarded for MySQL 8.0.13+, and the actual hidden-column materialization path is covered by unit tests. I do not want to make the shared `all_mode` case fail on older upstream lanes just to assert optional coverage.

pkg/sqlmodel/where_handle.go
```diff
@@ -55,6 +127,14 @@ func GetWhereHandle(source, target *model.TableInfo) *WhereHandle {
 		if rewritten == nil {
 			continue
 		}
+
+		ret.causalityIdxs = append(ret.causalityIdxs, rewritten)
```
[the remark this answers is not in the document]
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/1707f4c0933817f8a9618440d3adc18e0d1c1a59 . Added the `causalityIdxs` superset invariant on the field comment.

pkg/sqlmodel/causality.go
```diff
@@ -239,3 +249,79 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
+	}
+
+	datums, err := utils.AdjustBinaryProtocolForDatum(r.tiSessionCtx, values, cols[:len(values)])
+	if err != nil {
+		log.L().Debug("cannot adjust row for generated column evaluation",
+			zap.String("table", r.sourceTable.String()), zap.Error(err))
+		return values, false
+	}
+
+	full := make([]any, len(cols))
```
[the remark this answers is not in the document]
joechenrh: Keeping this as-is in this PR. Expression unique indexes are a narrow path, and adding per-RowChange scratch/cache state would make the row object mutable for a small optimization. I would prefer doing that only with a measured hotspot.

pkg/sqlmodel/reduce_test.go
```diff
@@ -71,6 +71,121 @@ func TestIdentityUpdatedWithUniqueKeys(t *testing.T) {
...
+	}
+}
+
+func TestPrimaryOrUniqueKeyUpdatedExpressionIndexMaterializeFailure(t *testing.T) {
+	t.Parallel()
+
+	source := &cdcmodel.TableName{Schema: "db", Table: "tb1"}
+	sourceTI := mockTableInfo(t, "CREATE TABLE tb1 (id INT PRIMARY KEY, email VARCHAR(255) UNIQUE, name VARCHAR(255), "+
+		"UNIQUE KEY lower_name ((lower(name))))")
+	corruptHiddenGeneratedExpr(t, sourceTI)
```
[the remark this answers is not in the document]
joechenrh: Leaving this unchanged. Same-package test helpers are commonly shared across `_test.go` files, and this helper is small enough that moving it to a separate test helper file would mostly add churn.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,93 @@
...
+		c.exprs = exprs
+		c.ok = true
+	})
+	return c.exprs, c.exprCtx, c.ok
+}
+
+func generatedColumnExprContext(tiSessionCtx sessionctx.Context) *exprstatic.ExprContext {
+	vars := tiSessionCtx.GetSessionVars()
+	charset, collation := vars.GetCharsetInfo()
+	// TODO(joechenrh): Carry downstream charset/collation for collation-sensitive expression indexes.
```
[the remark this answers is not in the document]
joechenrh: Leaving this unchanged for now. The TODO already names the owner and the removal condition: carry downstream charset/collation for collation-sensitive expression indexes. I would rather link a real follow-up issue once we decide to support that path, instead of creating a placeholder reference in this PR.

dm/tests/all_mode/data/db1.increment.sql
```diff
@@ -24,6 +24,29 @@ update t1 set name = 'gentestxxxxxx', dt = '2021-05-11 12:03:05', ts = '2021-05-
...
+replace into t_stored_gen_unique (id, name) values (20, 'Carol');
+replace into t_stored_gen_unique (id, name) values (30, 'Dave');
+replace into t_stored_gen_unique (id, name) values (40, 'DAVE');
+
+-- test unique functional index backed by a hidden generated column
+create table t_expr_unique (
+    id int primary key,
+    name varchar(64)
+);
+/*!80013 alter table t_expr_unique add unique key uk_lower_name ((lower(name))) */;
```
D3Hunter, line 44: maybe add another test where the UK is inside the table definition

dm/tests/all_mode/data/db1.increment.sql
```diff
@@ -24,6 +24,29 @@ update t1 set name = 'gentestxxxxxx', dt = '2021-05-11 12:03:05', ts = '2021-05-
 -- delete with unique key
 delete from t1 where gen_id > 124;
 
+-- test unique index on a stored generated column
+create table t_stored_gen_unique (
+    id int primary key,
+    name varchar(64),
+    lower_name varchar(64) generated always as (lower(name)) stored,
```
D3Hunter, line 31: inside binlog, is the column value already converted to lower?

pkg/sqlmodel/reduce.go
```diff
@@ -130,11 +130,26 @@ func (r *RowChange) IsPrimaryOrUniqueKeyUpdated() bool {
...
+		postValues    = r.postValues
+		checkIndexes  = r.whereHandle.UniqueIdxs
+		preOK, postOK bool
+	)
+
+	if r.whereHandle.hiddenGeneratedColumnExprCache != nil {
+		// Use causality indexes only when both row images are fully materialized.
+		preValues, preOK = r.fillVirtualGeneratedValues(r.preValues)
+		postValues, postOK = r.fillVirtualGeneratedValues(r.postValues)
+		if preOK && postOK {
```
D3Hunter, line 141: will one of the `ok`s be false happen? in that case some indices inside causalityIdxs is not considered

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,94 @@
...
+func (c *generatedColumnExprCache) getOrBuildExprs(
+	tiSessionCtx sessionctx.Context,
+) (map[int]expression.Expression, *exprstatic.ExprContext, bool) {
+	c.once.Do(func() {
+		c.exprCtx = generatedColumnExprContext(tiSessionCtx)
+		exprs := make(map[int]expression.Expression)
+		for _, col := range c.columns {
+			e, err := expression.ParseSimpleExprWithTableInfo(c.exprCtx, col.GeneratedExprString, c.sourceTableInfo)
+			if err != nil {
+				log.Warn("cannot build generated column expression, its index will be skipped for causality",
```
D3Hunter: we cannot safely ignore this error for non-trivial expression based UK, maybe expose it

dm/tests/all_mode/data/db1.increment.sql
joechenrh: Yes, for stored column, the value stored are the converted one. (It can already be handled before this PR, I just add tests to cover this case.)

pkg/sqlmodel/where_handle.go
joechenrh: I agree that ignoring the error might not be safe. But exposing this requires big changes with related APIs.

Besides, since the table schema has already been tracked by DM, the chance of failing to build the generated expression should be low. If expression building or evaluation does fail, it likely indicates a schema/session mismatch, and the corresponding downstream DML may also fail. So I prefer to keep the current behavior in this PR.

pkg/sqlmodel/reduce.go
joechenrh: Yes, acutually it's the same problem as the below comment. I think that if such a situation occurs, DML will also likely fail to execute (like some incompatibility between upstream and downstream).

pkg/sqlmodel/causality.go
```diff
@@ -232,10 +242,86 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
+	fullDatums := make([]types.Datum, len(cols))
+	copy(full, values)
+	copy(fullDatums, datums)
```
kennytm: My 5.5 isn't quite happy with this section, saying that the hidden columns do not necessarily always appear at the end of the columns list:

```sql
use db;
create table tb1 (
  id int primary key, 
  a varchar(32), 
  unique key uk_a ((lower(a)))
);
alter table tb1 add column b varchar(32);
alter table tb1 add unique key uk_b ((lower(b)));
...
insert into tb1 (id, a, b) values (1, 'Alice', 'Bob');
```

The downstream table info via http://127.0.0.1:10080/schema/db/tb1 should show 5 columns in this order (the `.Offset` field is the same):

0. `id`
1. `a`
2. `_V$_uk_a_0`
3. `b`
4. `_v$_uk_b_0`

There is a (slightly wrong) generated unit test, that after manually swapping the orders of `b` and `_V$_uk_a_0`, the Causality Key only has 2 entries (`["alice._v$_uk_a_0.db.tb1", "1.id.db.tb1"]`), missing the `uk_b` index.

<details>

```go
package sqlmodel

import (
	"testing"

	cdcmodel "github.com/pingcap/tiflow/cdc/model"
	"github.com/stretchr/testify/require"
)

func TestReviewHiddenExpressionBeforeVisibleColumn(t *testing.T) {
	ti := mockTableInfo(t, "CREATE TABLE tb1 ("+
		"id INT PRIMARY KEY, a VARCHAR(32), b VARCHAR(32), "+
		"UNIQUE KEY uk_a ((lower(a))), UNIQUE KEY uk_b ((lower(b))))")

	require.Len(t, ti.Columns, 5)
	hiddenA, visibleB, hiddenB := ti.Columns[3], ti.Columns[2], ti.Columns[4]
	ti.Columns[2], ti.Columns[3], ti.Columns[4] = hiddenA, visibleB, hiddenB
	for i, col := range ti.Columns {
		col.Offset = i
	}
	for _, idx := range ti.Indices {
		for _, idxCol := range idx.Columns {
			for _, col := range ti.Columns {
				if col.Name.L == idxCol.Name.L {
					idxCol.Offset = col.Offset
				}
			}
		}
	}

	change := NewRowChange(
		&cdcmodel.TableName{Schema: "db", Table: "tb1"},
		nil,
		nil,
		[]any{1, "Alice", "Bob"},
		ti,
		nil,
		nil,
	)
	keys := change.CausalityKeys()
	require.Contains(t, keys, "alice.%!hidden!%_v$_expr_index_0.db.tb1") // Human note: the key name being checked is incorrect. but that's not the main point.
	require.Contains(t, keys, "bob.%!hidden!%_v$_expr_index_1.db.tb1")
}
```

</details>

I haven't verified on a real E2E test. Please verify the behavior after this CREATE TABLE → ADD COLUMN → CREATE INDEX sequence.

joechenrh: Ohh, I mixed it up. The invariant is: **public** columns need to be continuous. It's not related to **hidden** column. 😿

Upated in https://github.com/pingcap/tiflow/pull/12697/commits/1d3e9621c3f0ba0c923c571d8c6b12cfc3dd8785

dm/tests/all_mode/data/db1.increment.sql
joechenrh: Added in https://github.com/pingcap/tiflow/pull/12697/changes/4ce9dcf2a7c3039990b82c6426034c2f8d5b4157 (work for MySQL >= 8.0.13, TiDB)

pkg/sqlmodel/where_handle.go
D3Hunter: maybe add a comment to explain this

joechenrh: Done in https://github.com/pingcap/tiflow/commit/b282753d32be1527c376811b84eea66d9a758e87 . Added a comment on the expression build failure path to document the current degraded behavior: the error cannot be surfaced through the existing causality scheduling API, so we skip the hidden-column index even though it can reduce causality. The comment also records why this is expected to be rare after DM has tracked the DDL.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,99 @@
...
+			cols = append(cols, col)
+		}
+	}
+	return &generatedColumnExprCache{
+		sourceTableInfo: source,
+		columns:         cols,
+	}
+}
+
+// getOrBuildExprs uses tiSessionCtx only on the first cache build.
```
D3Hunter, line 142: #### ⚠️ **[Major]** Schema-level WhereHandle now owns session-scoped expression state

**Why**
WhereHandle is cached as part of DownstreamTableInfo, but the new generatedColumnExprCache builds and stores ExprContext/expression values from the first tiSessionCtx that reaches getOrBuildExprs. That makes causality and safe-mode identity checks depend on hidden first-call state instead of an explicit row/session-owned materializer.

**Scope**
pkg/sqlmodel/where_handle.go:68

**Risk if unchanged**
A cached handle can be reused across row changes and DML worker goroutines with evaluator state tied to an earlier session context, so future session-sensitive expression indexes or caller paths can silently use stale SQL mode/charset/eval state or share mutable evaluator state concurrently.

**Evidence**
GetWhereHandle is documented as cacheable and stores WhereHandle in DownstreamTableInfo (dm/pkg/schema/tracker.go:85, dm/pkg/schema/tracker.go:502). The new cache keeps exprCtx/exprs behind sync.Once using the first tiSessionCtx (pkg/sqlmodel/where_handle.go:46, pkg/sqlmodel/where_handle.go:68). RowChanges are assigned that cached handle (dm/syncer/dml.go:169), and DML workers run multiple goroutines that can call IsPrimaryOrUniqueKeyUpdated through the same handle (dm/syncer/dml_worker.go:146, dm/syncer/dml_worker.go:403).

**Change request**
Maybe move this session-dependent materializer out of the schema-level WhereHandle, or keep only session-independent parsed metadata there and build/key the EvalContext by explicit session settings at the RowChange call site. If the shared cache is intentional, please add a short why-comment and a focused test/guard for first-caller session context and concurrent worker use.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,99 @@
...
+func (c *generatedColumnExprCache) getOrBuildExprs(
+	tiSessionCtx sessionctx.Context,
+) (map[int]expression.Expression, *exprstatic.ExprContext, bool) {
+	c.once.Do(func() {
+		c.exprCtx = generatedColumnExprContext(tiSessionCtx)
+		exprs := make(map[int]expression.Expression)
+		for _, col := range c.columns {
+			e, err := expression.ParseSimpleExprWithTableInfo(c.exprCtx, col.GeneratedExprString, c.sourceTableInfo)
```
D3Hunter, line 150: #### ⚠️ **[Major]** Expression-index causality fails open when materialization fails

**Why**
When generated-column expression parsing or evaluation fails, the new path logs and skips the hidden-column index instead of returning an actionable error or falling back to a conservative conflict key.

**Scope**
pkg/sqlmodel/where_handle.go:76, pkg/sqlmodel/causality.go:265

**Risk if unchanged**
Rows with different primary keys but the same unique expression-index value can be scheduled concurrently after a schema/session mismatch or unsupported expression, which can still produce transient downstream unique-key conflicts during upgrade or compatibility-sensitive replication.

**Evidence**
`getOrBuildExprs` returns `ok=false` after a parse error and comments that the index will be skipped for causality; `fillVirtualGeneratedValues` propagates `fullValuesOK=false`, and `getCausalityString` then skips only hidden-column indexes while retaining other keys such as the PK. The fallback test asserts only the PK key remains for an expression-index table, so this is an intentional silent degradation rather than fail-closed behavior.

**Change request**
Can we fail closed for this path, for example by using a conservative whole-row/table-level causality key for tables with unmaterializable hidden expression indexes, or by surfacing an actionable error that tells operators the expression index cannot be used safely?

dm/tests/all_mode/data/db1.increment.sql
```diff
@@ -24,6 +24,40 @@ update t1 set name = 'gentestxxxxxx', dt = '2021-05-11 12:03:05', ts = '2021-05-
...
+replace into t_stored_gen_unique (id, name) values (20, 'Carol');
+replace into t_stored_gen_unique (id, name) values (30, 'Dave');
+replace into t_stored_gen_unique (id, name) values (40, 'DAVE');
+
+-- test unique functional index backed by a hidden generated column
+create table t_expr_unique (
+    id int primary key,
+    name varchar(64)
+);
+/*!80013 alter table t_expr_unique add unique key uk_lower_name ((lower(name))) */;
```
D3Hunter, line 44: #### 🟡 **[Minor]** Expression-index integration coverage depends on MySQL version comments

**Why**
The added all_mode fixture wraps the functional-index DDL and one full table scenario in `/*!80013 ... */`, so older upstream versions can ignore the exact statements intended to exercise hidden generated columns.

**Scope**
dm/tests/all_mode/data/db1.increment.sql:44

**Risk if unchanged**
The integration case may pass without validating the new expression-index causality behavior in environments where the version comment is ignored, reducing deterministic evidence for the compatibility-sensitive path.

**Evidence**
`t_expr_unique` is created unconditionally, but the unique functional index is guarded by `/*!80013 ... */`; `t_expr_unique_inline` and all its rows are also guarded. If the source does not execute 8.0.13 version comments, the test either exercises ordinary primary-key replaces only or skips the inline table entirely.

**Change request**
Please add a deterministic guard or assertion for this path, such as checking the source supports functional indexes before running the case, splitting it into a MySQL-8.0.13+ case, or adding a validation that the functional index actually exists before relying on the integration result.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,99 @@
...
+		c.exprs = exprs
+		c.ok = true
+	})
+	return c.exprs, c.exprCtx, c.ok
+}
+
+func generatedColumnExprContext(tiSessionCtx sessionctx.Context) *exprstatic.ExprContext {
+	vars := tiSessionCtx.GetSessionVars()
+	charset, collation := vars.GetCharsetInfo()
+	// TODO(joechenrh): Carry downstream charset/collation for collation-sensitive expression indexes.
```
D3Hunter: #### 🧹 **[Nit]** TODO omits tracking and removal condition

**Scope**
pkg/sqlmodel/where_handle.go:98

**Change request**
Please add an issue/reference and the condition for removing this TODO, or convert it into a short limitation comment if there is no planned follow-up.

D3Hunter (commented):
### Summary

- Total findings: 5
- Inline comments: 4
- Summary-only findings (no inline anchor): 1

<details>

#### ⚠️ **[Major]** (3)

1. Expression-only unique tables can still build invalid WHERE predicates (`Fallback WHERE construction in `whereColumnsAndValues` in `pkg/sqlmodel/row_change.go``)
2. Schema-level WhereHandle now owns session-scoped expression state (`pkg/sqlmodel/where_handle.go:68`)
3. Expression-index causality fails open when materialization fails (`pkg/sqlmodel/where_handle.go:76, pkg/sqlmodel/causality.go:265`)

#### 🟡 **[Minor]** (1)

1. Expression-index integration coverage depends on MySQL version comments (`dm/tests/all_mode/data/db1.increment.sql:44`)

#### 🧹 **[Nit]** (1)

1. TODO omits tracking and removal condition (`pkg/sqlmodel/where_handle.go:98`)

</details>

### Unanchored findings

#### ⚠️ **[Major]** (1)

1. Expression-only unique tables can still build invalid WHERE predicates
   - Scope: `Fallback WHERE construction in `whereColumnsAndValues` in `pkg/sqlmodel/row_change.go``
   - Request: Can we filter hidden generated columns out of the fallback WHERE column list, or otherwise make the fallback use the same visible-column projection as `ColumnCount`, and add a case for UPDATE/DELETE on a table whose only unique key is a functional index and has no PK or ordinary unique key?

dm/tests/all_mode/data/db1.increment.sql
joechenrh: Done in https://github.com/pingcap/tiflow/commit/225b331c7b3b1216618f14b8358d211fa42f3c56 . I added a guard after loading the incremental SQL.

If the upstream does not execute the `/*!80013 */` functional-index statements, the test prints an explicit skip message. If the case is enabled, it asserts that both expression-index fixtures have `uk_lower_name` in `information_schema.statistics`, so the integration result no longer silently relies on a skipped version-comment path.

pkg/sqlmodel/where_handle.go
joechenrh: Done in https://github.com/pingcap/tiflow/commit/225b331c7b3b1216618f14b8358d211fa42f3c56 . I kept the cache on the per-table `WhereHandle` intentionally.

The main reason is to avoid rebuilding/parsing generated-column expressions on every row, which would put expression-index tables on a hot causality path.

For the session-state concern: the handle is rebuilt when DDL invalidates the downstream schema cache, and within one Syncer lifetime DM builds one `s.sessCtx` from the downstream apply SQL mode/timezone and reuses it for all RowChanges. Runtime changes to global SQL mode/timezone are not reflected in existing DM sessions anyway. I added a short comment documenting that lifecycle assumption.

pkg/sqlmodel/where_handle.go
joechenrh: I kept this as a TODO intentionally. The missing piece is carrying downstream charset/collation into this expression context; once DM passes those session settings here, this TODO can be removed.

For this PR, SQL mode/timezone are already handled by the fixed downstream apply session context.

pkg/sqlmodel/where_handle.go
joechenrh: I agree that ignoring this error is not ideal, but I prefer to keep the existing degraded behavior in this PR.

Exposing it as an error would require changing the current causality APIs and the scheduler call path. Also, by this point DM has already tracked the DDL and built the hidden generated-column metadata from the downstream schema. So a generated-expression build failure should be rare, and if it does happen it likely indicates a schema/session mismatch that may also make the corresponding downstream DML fail.

I added a comment in the build-failure path earlier to document this existing behavior and the correctness tradeoff: https://github.com/pingcap/tiflow/commit/b282753d32be1527c376811b84eea66d9a758e87 .

pkg/sqlmodel/reduce.go
```diff
@@ -120,7 +120,7 @@ func (r *RowChange) IsPrimaryOrUniqueKeyUpdated() bool {
 	}
 
 	if idx := r.whereHandle.UniqueNotNullIdx; idx != nil {
-		pre, post := r.identityValuesByIndex(idx)
+		pre, post := r.identityValuesByIndex(idx, r.preValues, r.postValues)
```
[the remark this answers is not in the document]
joechenrh: Done in https://github.com/pingcap/tiflow/commit/d3bb004cf7ee43ec5a1596b0075c8da356e97f14 and simplified further in https://github.com/pingcap/tiflow/commit/0863b462810ea3cb229c767ae22626b9255b3ca7 .

The fix is not to build a full row for all callers, but to centralize the row-image mapping: when the row image is visible-only, full column offsets are mapped to visible offsets before extracting index values. This now covers `UniqueNotNullIdx`, `RowIdentity`, `whereColumnsAndValues`, and causality index extraction. I also added regression coverage for a NOT NULL unique visible column ordered after a hidden generated column.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,102 @@
...
+	// backed by hidden generated columns are tracked in causalityIdxs instead.
 	UniqueIdxs []*model.IndexInfo
+
+	// causalityIdxs is a superset of UniqueIdxs and also includes expression indexes.
+	causalityIdxs                  []*model.IndexInfo
+	hiddenGeneratedColumnExprCache *generatedColumnExprCache
+}
+
+type generatedColumnExprCache struct {
+	// ExprContext is cached with the per-table WhereHandle. The handle is rebuilt
```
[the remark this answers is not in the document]
joechenrh: Done in https://github.com/pingcap/tiflow/commit/d3bb004cf7ee43ec5a1596b0075c8da356e97f14 .

Moved the ExprContext lifecycle comment so it documents `exprCtx` directly.

pkg/sqlmodel/reduce.go
```diff
@@ -130,11 +130,26 @@ func (r *RowChange) IsPrimaryOrUniqueKeyUpdated() bool {
...
-	for _, idx := range r.whereHandle.UniqueIdxs {
+	var (
+		preValues     = r.preValues
+		postValues    = r.postValues
+		checkIndexes  = r.whereHandle.UniqueIdxs
+		preOK, postOK bool
+	)
+
+	if r.whereHandle.hiddenGeneratedColumnExprCache != nil {
+		// Use causality indexes only when both row images are fully materialized.
```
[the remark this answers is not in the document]
joechenrh: Done in https://github.com/pingcap/tiflow/commit/fd071e6e8249dbdabee971b9676f7e8ff0b8687f .

The comment now says that expression indexes are used only when both row images are fully materialized, and otherwise the check falls back to visible-column unique indexes.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,102 @@
...
+		exprs := make(map[int]expression.Expression)
+		for _, col := range c.columns {
+			e, err := expression.ParseSimpleExprWithTableInfo(c.exprCtx, col.GeneratedExprString, c.sourceTableInfo)
+			if err != nil {
+				// Current causality callers cannot surface this error to the
+				// scheduler, so keep the existing degraded behavior: skip the
+				// hidden-column index. This can reduce causality, but after DM
+				// has tracked the DDL, a build failure usually means a
+				// schema/session mismatch that may also fail the downstream DML.
+				log.Warn("cannot build generated column expression, its index will be skipped for causality",
```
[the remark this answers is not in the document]
joechenrh: Done in https://github.com/pingcap/tiflow/commit/d3bb004cf7ee43ec5a1596b0075c8da356e97f14 .

The log and inline comment now describe the table-level behavior: if building any hidden generated-column expression fails, hidden-column indexes for this table are skipped for causality. I kept this all-or-nothing behavior explicit instead of making it look like a per-index skip.

dm/tests/all_mode/run.sh
```diff
@@ -8,6 +8,17 @@ WORK_DIR=$TEST_DIR/$TEST_NAME
 API_VERSION="v1alpha1"
 ILLEGAL_CHAR_NAME='t-Ë!s`t'
 
+function check_expression_index_case_enabled() {
+	run_sql_source1 "select count(*) as expression_index_table_count from information_schema.tables where table_schema = 'all_mode' and table_name = 't_expr_unique_inline';"
+	if grep -Fq "expression_index_table_count: 0" "$TEST_DIR/sql_res.$TEST_NAME.txt"; then
+		echo "unique functional index case is skipped because upstream does not support /*!80013 */ syntax"
```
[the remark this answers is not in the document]
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/225b331c7b3b1216618f14b8358d211fa42f3c56 .

The all_mode case now checks whether the functional-index table exists after loading the incremental SQL. If upstream does not execute the MySQL 8.0.13 version-comment path, it prints an explicit skip message. If the path is enabled, it asserts both expression-index fixtures have `uk_lower_name` in `information_schema.statistics`.

pkg/sqlmodel/where_handle.go
```diff
@@ -14,21 +14,102 @@
...
+// getOrBuildExprs uses tiSessionCtx only on the first cache build.
+func (c *generatedColumnExprCache) getOrBuildExprs(
+	tiSessionCtx sessionctx.Context,
+) (map[int]expression.Expression, *exprstatic.ExprContext, bool) {
+	c.once.Do(func() {
+		c.exprCtx = generatedColumnExprContext(tiSessionCtx)
+		exprs := make(map[int]expression.Expression)
+		for _, col := range c.columns {
+			e, err := expression.ParseSimpleExprWithTableInfo(c.exprCtx, col.GeneratedExprString, c.sourceTableInfo)
+			if err != nil {
```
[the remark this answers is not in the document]
joechenrh: Leaving this all-or-nothing for this PR.

Partial per-column success sounds attractive, but it would need more bookkeeping than the current materializer has: we would need to track which hidden generated columns were buildable/evaluable, map indexes to the exact hidden columns they depend on, and preserve generated-column dependency order. Since the current causality API cannot surface errors anyway, I prefer the simpler table-level degraded behavior and making that behavior explicit in the comment/log.

pkg/sqlmodel/where_handle.go
joechenrh: I agree fail-closed would be stronger, but I am keeping the current degraded behavior in this PR.

Surfacing this as an error would require changing the causality API and the scheduler/DML caller path. At this point DM has already tracked the table schema and built hidden generated-column metadata from it, so a build/eval failure should be rare and likely indicates a schema/session mismatch that may also make the downstream DML fail.

The behavior is now documented and covered by fallback tests: the expression index is skipped, ordinary visible keys still participate, and whole-row fallback keeps visible columns aligned.

pkg/sqlmodel/causality.go
```diff
@@ -232,10 +242,93 @@ func (r *RowChange) getCausalityString(values []interface{}) []string {
...
+	if len(values) >= len(cols) {
+		return values, true
+	}
+
+	exprs, exprCtx, ok := r.whereHandle.hiddenGeneratedColumnExprCache.getOrBuildExprs(r.tiSessionCtx)
+	if !ok {
+		return values, false
+	}
+
+	visibleCols := visibleColumns(cols)
```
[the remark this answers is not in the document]
joechenrh: Partially improved in https://github.com/pingcap/tiflow/commit/d3bb004cf7ee43ec5a1596b0075c8da356e97f14 .

The visible-column slice is now cached in `rowValueMapper`, so `fillVirtualGeneratedValues` no longer rebuilds it per call. I am not adding pre/post materialized-row caches on `RowChange` in this PR: that would add mutable per-row cache state for an expression-index-only optimization. I would rather do that if we see a measured hotspot or clean up RowChange lifecycle more broadly.

pkg/sqlmodel/where_handle.go
joechenrh: I kept the expression cache on the per-table `WhereHandle` intentionally.

The main reason is to avoid rebuilding/parsing generated-column expressions on the row hot path. In DM, RowChanges in this path use the syncer apply session context built from downstream SQL mode/timezone. Runtime global session changes are not reflected in the existing syncer session anyway, and the handle is rebuilt when DDL invalidates the schema cache.

I added the lifecycle comment in https://github.com/pingcap/tiflow/commit/225b331c7b3b1216618f14b8358d211fa42f3c56 . The cache is also protected by `sync.Once`; after build, callers only read the parsed expressions and ExprContext.

dm/tests/all_mode/data/db1.increment.sql
joechenrh: Added in https://github.com/pingcap/tiflow/commit/4ce9dcf2ad8852705331253f5af3a0af095f85dc .

The all_mode fixture now covers both `ALTER TABLE ... ADD UNIQUE KEY ((expr))` and a functional unique index declared inside the table definition.

pkg/sqlmodel/causality.go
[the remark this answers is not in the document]
joechenrh: Follow-up in https://github.com/pingcap/tiflow/commit/720cca91e26a55928bfdf9877c0808e766d5b8f1 .

I simplified `datumValue` further: every non-NULL materialized datum now first uses `Datum.ToString()`, with `GetValue()` only as fallback. The generated value stored in the `full` slice is only used for causality key generation and update-key equality checks; later generated-column evaluation still uses the original `Datum` in `fullDatums`, so this keeps comparison stable without changing expression evaluation.

pkg/sqlmodel/causality.go
joechenrh: Verified and fixed in https://github.com/pingcap/tiflow/commit/d3bb004cf7ee43ec5a1596b0075c8da356e97f14 .

The wrong assumption was that hidden columns always stay at the end. The safer invariant is that public columns form the row image, while hidden generated columns can be interleaved in `TableInfo.Columns`. The fix centralizes full-offset to visible-offset mapping through `rowValueMapper`, and the tests now explicitly reorder columns into `id, a, hidden_a, b, hidden_b` / `a, hidden_a, b, c` shapes.

pkg/sqlmodel/where_handle.go
joechenrh: Done in https://github.com/pingcap/tiflow/commit/fd071e6e8249dbdabee971b9676f7e8ff0b8687f .

I kept the TODO but added the removal condition: once DM initializes downstream charset/collation session settings for expression-index evaluation, this TODO can be removed.

pkg/sqlmodel/where_handle.go
joechenrh: Added in https://github.com/pingcap/tiflow/commit/b282753d32be1527c376811b84eea66d9a758e87 and refined later in https://github.com/pingcap/tiflow/commit/d3bb004cf7ee43ec5a1596b0075c8da356e97f14 .

The comment now documents the existing degraded behavior and the table-level scope when generated expression build fails.

pkg/sqlmodel/reduce.go
joechenrh: Follow-up in https://github.com/pingcap/tiflow/commit/0863b462810ea3cb229c767ae22626b9255b3ca7 .

I removed the extra `ok`-style defensive fallback from the row mapper path. The mapper now only handles the necessary visible-row/full-row offset conversion; it no longer silently treats invalid offsets or invalid row images as a degraded path.

dm/tests/all_mode/data/db1.increment.sql
joechenrh: Handled in https://github.com/pingcap/tiflow/commit/225b331c7b3b1216618f14b8358d211fa42f3c56 .

After loading the incremental SQL, the test checks whether the inline functional-index table exists. If the upstream does not execute MySQL 8.0.13 version comments, it prints an explicit skip message. If the case is enabled, it asserts both expression-index fixtures have `uk_lower_name` in `information_schema.statistics`.

joechenrh:
For the summary-only finding about expression-only unique tables building invalid fallback WHERE predicates: handled in https://github.com/pingcap/tiflow/commit/fd071e6e8249dbdabee971b9676f7e8ff0b8687f and https://github.com/pingcap/tiflow/commit/d3bb004cf7ee43ec5a1596b0075c8da356e97f14 .

`whereColumnsAndValues` now uses the cached visible-column projection for fallback WHERE, and index value extraction goes through the row mapper when a WHERE-eligible unique index is selected. I also added DELETE regression coverage for a table whose only unique key is a functional index and has no PK or ordinary unique key.

D3Hunter (approved):

dm/syncer/syncer.go
```diff
@@ -447,6 +447,13 @@ func (s *Syncer) Init(ctx context.Context) (err error) {
 	vars := map[string]string{
 		"time_zone": s.timezone.String(),
 	}
+	// Expression-index causality uses the downstream apply SQL mode.
+	for k, v := range s.cfg.To.Session {
+		if strings.EqualFold(k, "sql_mode") {
+			vars["sql_mode"] = v
```
OliverS929: Could we avoid applying downstream sql_mode to the shared s.sessCtx and keep it isolated to expression-index causality, since it may also affect expression filters and schema rendering?

pkg/sqlmodel/where_handle.go
```diff
@@ -14,27 +14,182 @@
...
+		c.exprs = exprs
+		c.ok = true
+	})
+	return c.exprs, c.exprCtx, c.ok
+}
+
+func generatedColumnExprContext(tiSessionCtx sessionctx.Context) *exprstatic.ExprContext {
+	vars := tiSessionCtx.GetSessionVars()
+	charset, collation := vars.GetCharsetInfo()
+	// TODO(joechenrh): Carry downstream charset/collation for collation-sensitive expression indexes.
```
OliverS929: Could we expand this TODO slightly to mention the current behavior? Until downstream charset/collation are carried here, collation-sensitive functional indexes may generate causality keys that do not exactly match the database's unique-index comparison semantics. A short note would make this limitation easier to reason about later.

OliverS929 (approved):
rest LGTM.

joechenrh:
/hold
There are still some comments.

pkg/sqlmodel/where_handle.go
joechenrh: Done in https://github.com/pingcap/tiflow/commit/ea057a8a66a3fea91f0d186a32eb45e1aa6a1297 .

I expanded the TODO to describe the current behavior: expression-index causality currently uses the session ctx charset/collation, so collation-sensitive functional indexes may not exactly match downstream unique-index comparison semantics until DM carries downstream charset/collation into this context.

joechenrh:
/unhold
I've pasted manual test result in the PR description.

PR #12698, opened 2026-06-10
loader(dm): strip external ID from import-into S3 URI

### LINKED ISSUE #12699: DM import-into should not pass explicit S3 external ID to TiDB
## Bug Report

When a DM task uses S3 storage with `role-arn` and an explicit `external-id`, Dumpling may need the external ID to assume the AWS role. DM currently reuses the same S3 URI as the Lightning/IMPORT INTO source directory.

For `import-mode: import-into`, older TiDB versions reject an explicit `external-id` in the S3 URI before TiDB is upgraded with pingcap/tidb#69045. This means DM cannot use the same URI for both Dumpling and IMPORT INTO in this compatibility scenario.

## Expected behavior

DM should keep the original S3 URI for Dumpling and normal Lightning, but avoid passing explicit `external-id` / `external_id` to TiDB IMPORT INTO so TiDB can handle its own keyspace external ID logic.

## Related PR

- https://github.com/pingcap/tiflow/pull/12698

GMHDBJD:

Issue Number: close #12699

DM reuses the same S3 data-dir for Dumpling and Lightning/IMPORT INTO. When the URI contains an explicit `external-id`, Dumpling may need it for AWS AssumeRole, but older TiDB IMPORT INTO rejects explicit external ID before pingcap/tidb#69045.

### What is changed and how does it work?

Keep the original DM data-dir unchanged for Dumpling and normal Lightning. Only when DM configures Lightning with `import-mode: import-into`, strip `external-id` and `external_id` from the S3 URI passed as `Mydumper.SourceDir` so TiDB can inject its own keyspace external ID.

The stripping logic is implemented as a reusable DM storage helper backed by TiDB objstore URL parsing, while preserving other S3 query parameters such as `role-arn`, `region`, `endpoint`, and `provider`.

### Check List

Tests

- Unit test

### Release note

```release-note
Fix DM import-into compatibility with S3 role ARN URIs that include explicit external ID.
```

### Test

- `go test ./dm/pkg/storage -run TestStripS3ExternalID -count=1`
- `go test ./dm/loader -run TestMakeGlobalConfigStripsS3ExternalIDOnlyForImportInto -count=1`
- `go test ./dm/pkg/storage -count=1`
- `go test ./dm/loader -count=1`

### DIFF
--- a/dm/pkg/storage/utils.go
+++ b/dm/pkg/storage/utils.go
@@ -31,6 +31,7 @@ import (
 	backuppb "github.com/pingcap/kvproto/pkg/brpb"
 	"github.com/pingcap/tidb/pkg/objstore"
 	"github.com/pingcap/tidb/pkg/objstore/objectio"
+	"github.com/pingcap/tidb/pkg/objstore/s3like"
 	"github.com/pingcap/tidb/pkg/objstore/storeapi"
 )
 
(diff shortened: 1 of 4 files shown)


dm/pkg/storage/utils.go
```diff
@@ -96,6 +97,34 @@ func IsS3Path(rawURL string) bool {
...
+	}
+	u, err := objstore.ParseRawURL(rawURL)
+	if err != nil || u.Scheme != "s3" {
+		return rawURL
+	}
+
+	query := u.Query()
+	changed := false
+	for key := range query {
+		if strings.EqualFold(strings.ReplaceAll(key, "_", "-"), s3like.S3ExternalID) {
```
D3Hunter, line 116: if tidb upgraded, we can use `objstore.NormalizeQueryParameterKey`

D3Hunter (approved):
rest lgtm

dm/pkg/storage/utils.go
joechenrh: Upgraded in https://github.com/pingcap/tiflow/pull/12674, but it's waiting CI 😂

joechenrh (approved):

OliverS929 (approved):

PR #12729, opened 2026-06-18
checker(dm): improve same table precheck error

### LINKED ISSUE #12737: DM precheck reports confusing error for case-insensitive route conflicts
## Bug Report

When DM precheck validates case-insensitive route rules whose source table patterns only differ by case, the checker can return the table-router duplicate match error before the existing same-target-table validation gets a chance to report the clearer diagnostic.

Example user-visible error:

```text
RawCause: table test.T matches more than one rule
```

This is confusing compared with the clearer same-target-table error that points out the conflicting target table names.

### Expected behavior

DM precheck should report an actionable same-target-table error, for example:

```text
same target table `test`.`T` vs `test`.`t`
```

### Actual behavior

For some case-insensitive route-rule collisions, precheck reports the lower-level table-router error:

```text
RawCause: table test.T matches more than one rule
```

### Related PR

https://github.com/pingcap/tiflow/pull/12729

GMHDBJD:

Issue Number: close #12737

DM precheck can report a confusing raw table-router error when multiple case-insensitive table route rules match the same selected source table, for example:

```text
RawCause: table test.T matches more than one rule
```

This hides the existing clearer same-target-table diagnostic.

After validating the regular-expression router, DM now discovers source tables and applies the block/allow list before checking table-route collisions. For each selected source table, the checker uses the same filter semantics as the router to find matching table-level route rules.

When multiple matching table-level rules map to target tables whose fully qualified names differ only in letter case, the checker returns the existing `ErrTaskCheckSameTableName` error with actionable details such as:

```text
same target table `test`.`T` vs `test`.`t`
```

The check runs immediately before routing, so a real target-name case collision receives the clearer diagnostic while rules for nonexistent or block/allow-list-excluded tables keep their previous behavior. Exact-identical targets, different targets, and schema-level route conflicts continue to use the router's existing duplicate-rule error so the workaround correctly points users to the route rules.

#### Tests 

 - Unit test

Commands run:

```text
gofmt -w dm/checker/checker.go dm/checker/check_test.go dm/pkg/conn/db.go dm/pkg/conn/db_test.go
go test ./dm/checker ./dm/pkg/conn -count=1
git diff --check
```

Coverage includes actual selected-table collisions, nonexistent rules, block/allow-list-excluded tables, wildcard/exact pattern overlap, implicit target-table fallback, exact-identical targets retaining the source-route error, different targets, schema-level routes retaining the router error, invalid target schema, and the extracted routing helper.

No compatibility break is expected. The new diagnostic is limited to selected source tables that would already fail with the router's multiple-rule error; unused and filtered-out route rules do not introduce new failures. The checker performs one additional route-rule matching pass during initialization.

No.

### Release note 

```release-note
Fix a confusing DM precheck error message when case-insensitive table route rules for a selected source table map to target table names that differ only in letter case.
```

### DIFF
--- a/dm/checker/checker.go
+++ b/dm/checker/checker.go
@@ -567,12 +567,12 @@ func (c *Checker) fetchSourceTargetDB(
 		return nil, nil, terror.ErrTaskCheckGenBAList.Delegate(err)
 	}
 	instance.baList = bAList
-	if err := sameTargetTableNameDetectionForRouteRules(instance.cfg.CaseSensitive, instance.cfg.RouteRules); err != nil {
-		return nil, nil, err
-	}
 	r, err := regexprrouter.NewRegExprRouter(instance.cfg.CaseSensitive, instance.cfg.RouteRules)
 	if err != nil {
 		return nil, nil, terror.ErrTaskCheckGenTableRouter.Delegate(err)
+	}
+	if err := sameTargetTableNameDetectionForRouteRules(instance.cfg.CaseSensitive, instance.cfg.RouteRules); err != nil {
+		return nil, nil, err
 	}
 
 	if err != nil {
@@ -892,7 +892,7 @@ func sameTargetTableNameDetectionForRouteRules(caseSensitive bool, rules []*rout
 	sourceTableToTargetNameSets := make(map[string]map[string]string)
 	var messages []string
 	for _, rule := range rules {
-		if rule == nil || rule.TablePattern == "" || rule.TargetSchema == "" {
+		if rule == nil || rule.TablePattern == "" {
 			continue
 		}
 
@@ -900,10 +900,6 @@ func sameTargetTableNameDetectionForRouteRules(caseSensitive bool, rules []*rout
 		if targetTable == "" {
 			targetTable = rule.TablePattern
 		}
-		if targetTable == "" {
-			continue
-		}
-
 		source := filter.Table{
 			Schema: rule.SchemaPattern,
 			Name:   rule.TablePattern,
--- a/dm/checker/checker.go
+++ b/dm/checker/checker.go
@@ -898,7 +898,9 @@ func checkTableRouteCaseCollisions(
 	// reports "matches more than one rule" before the checker can build target
 	// table mappings. Match table-level route rules with the same filter
 	// semantics, but only against source tables that exist and remain after
-	// block/allow filtering.
+	// block/allow filtering. Schema-level route conflicts intentionally keep the
+	// router's generic duplicate-rule error because they are outside this target-
+	// table diagnostic.
 	type matchableRule struct {
 		rule    *router.TableRule
 		matcher *filter.Filter
@@ -950,7 +952,7 @@ func checkTableRouteCaseCollisions(
 			nameL := strings.ToLower(name)
 			if nameO, ok := targetNameSets[nameL]; !ok {
 				targetNameSets[nameL] = name
-			} else {
+			} else if nameO != name {
 				message := fmt.Sprintf("same target table %v vs %s", nameO, name)
 				if _, ok := messageSet[message]; !ok {
 					messageSet[message] = struct{}{}
--- a/dm/checker/checker.go
+++ b/dm/checker/checker.go
@@ -37,6 +37,7 @@ import (
 	"github.com/pingcap/tidb/pkg/util/dbutil"
 	"github.com/pingcap/tidb/pkg/util/filter"
 	regexprrouter "github.com/pingcap/tidb/pkg/util/regexpr-router"
+	router "github.com/pingcap/tidb/pkg/util/table-router"
 	"github.com/pingcap/tiflow/dm/config"
 	"github.com/pingcap/tiflow/dm/config/dbconfig"
 	"github.com/pingcap/tiflow/dm/loader"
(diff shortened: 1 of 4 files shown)


dm/checker/checker.go
```diff
@@ -876,6 +880,61 @@ func sameTableNameDetection(tables map[filter.Table][]filter.Table) error {
...
+			Schema: rule.TargetSchema,
+			Name:   targetTable,
+		}
+		name := target.String()
+		nameL := strings.ToLower(name)
+		if nameO, ok := tableNameSets[nameL]; !ok {
+			tableNameSets[nameL] = name
+		} else if nameO != name {
+			messages = append(messages, fmt.Sprintf("same target table %v vs %s", nameO, name))
+		}
```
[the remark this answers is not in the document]
GMHDBJD: Thanks, this is reasonable. Updated the preflight to catch any case-insensitive source-rule collision, including collisions that map to the exact same target table name, and added regression coverage for that case.

dm/checker/checker.go
```diff
@@ -876,6 +880,61 @@ func sameTableNameDetection(tables map[filter.Table][]filter.Table) error {
...
+			continue
+		}
+
+		targetTable := rule.TargetTable
+		if targetTable == "" {
+			targetTable = rule.TablePattern
+		}
+		if targetTable == "" {
+			continue
+		}
```
joechenrh:
```suggestion
```

dm/checker/checker.go
```diff
@@ -566,6 +567,9 @@ func (c *Checker) fetchSourceTargetDB(
 		return nil, nil, terror.ErrTaskCheckGenBAList.Delegate(err)
 	}
 	instance.baList = bAList
+	if err := sameTargetTableNameDetectionForRouteRules(instance.cfg.CaseSensitive, instance.cfg.RouteRules); err != nil {
+		return nil, nil, err
+	}
```
joechenrh: How about moving this detection after `NewRegExprRouter`, so we don't need to check `TargetSchema`.

joechenrh (commented):
Rest LGTM

dm/checker/checker.go
```diff
@@ -876,6 +880,57 @@ func sameTableNameDetection(tables map[filter.Table][]filter.Table) error {
 	return nil
 }
 
+func sameTargetTableNameDetectionForRouteRules(caseSensitive bool, rules []*router.TableRule) error {
```
D3Hunter: The current name is long and noun-heavy. How about `checkTableRouteCaseCollisions`? It is shorter and more closely describes the actual behavior: checking table-level route rules for collisions under case-insensitive matching.

dm/checker/checker.go
```diff
@@ -570,6 +571,9 @@ func (c *Checker) fetchSourceTargetDB(
 	if err != nil {
 		return nil, nil, terror.ErrTaskCheckGenTableRouter.Delegate(err)
 	}
+	if err := sameTargetTableNameDetectionForRouteRules(instance.cfg.CaseSensitive, instance.cfg.RouteRules); err != nil {
```
D3Hunter: This preflight examines every configured route rule before source tables are discovered and filtered. As a result, case-colliding rules for a nonexistent table or a table excluded by the block/allow list now fail precheck, whereas previously they never reached `router.Route`. It changes behavior for unused and allow-list-excluded rules. Is this stricter validation intentional? If so, please document it and add coverage; otherwise, the detection should be tied to the actual selected source tables.

dm/checker/checker.go
```diff
@@ -876,6 +880,57 @@ func sameTableNameDetection(tables map[filter.Table][]filter.Table) error {
...
+
+		targetTable := rule.TargetTable
+		if targetTable == "" {
+			targetTable = rule.TablePattern
+		}
+		source := filter.Table{
+			Schema: rule.SchemaPattern,
+			Name:   rule.TablePattern,
+		}
+		sourceNameL := strings.ToLower(source.String())
```
D3Hunter: This groups source patterns only when their case-folded strings are exactly equal; it does not detect textually different but overlapping patterns. For example, `foo* -> dst.T` and `foo1 -> dst.t` produce different map keys, but a concrete source table `test.foo1` matches both rules, so the router still returns `matches more than one rule` and this clearer diagnostic is not used. Is that narrow scope intentional? If so, please document it and add a negative test; otherwise, overlap detection needs to use the actual selected source tables rather than comparing pattern text.

dm/checker/checker.go
```diff
@@ -876,6 +880,57 @@ func sameTableNameDetection(tables map[filter.Table][]filter.Table) error {
...
+	}
+
+	// regexprrouter matches source patterns case-insensitively in this mode and
+	// reports "matches more than one rule" before the checker can build target
+	// table mappings. Detect the same-target-table conflict from the route
+	// rules first so the precheck reports the actionable table names.
+	sourceTableToTargetNameSets := make(map[string]map[string]string)
+	var messages []string
+	for _, rule := range rules {
+		if rule == nil || rule.TablePattern == "" {
```
D3Hunter, line 910: An empty `TablePattern` is a valid schema-level route, so this skips all schema-route collisions. For example, case-insensitive rules `Sales -> archive` and `sales -> Archive` both match a concrete table such as `Sales.orders`; the helper ignores both rules and the router still returns the raw `matches more than one rule` error. If this preflight is intentionally table-route-only, please make that scope explicit in the name/comment and add coverage. Otherwise, schema-level routes need corresponding handling.

dm/checker/checker.go
```diff
@@ -876,6 +880,57 @@ func sameTableNameDetection(tables map[filter.Table][]filter.Table) error {
...
+
+		target := filter.Table{
+			Schema: rule.TargetSchema,
+			Name:   targetTable,
+		}
+		name := target.String()
+		nameL := strings.ToLower(name)
+		if nameO, ok := tableNameSets[nameL]; !ok {
+			tableNameSets[nameL] = name
+		} else {
```
D3Hunter: The unconditional duplicate branch also classifies an exact-same-target case such as `T -> dst.T` and `t -> dst.T` as `ErrTaskCheckSameTableName`, producing `same target table dst.T vs dst.T`. That is primarily a duplicate source-route match, not a target-name collision, and the error workaround to check `target-table` is not actionable: changing the second target to `dst.X` only allows this helper to pass, after which the router still reports `matches more than one rule`. Consider retaining `else if nameO != name` for the target-name diagnostic and reporting exact duplicates as a source-route conflict, with an appropriate workaround.

joechenrh (approved):

D3Hunter (approved):

PR #12731, opened 2026-06-18
syncer, sqlmodel(dm): fix generated index schema tracking

joechenrh:

Issue Number: ref #12696

DM's schema tracker and row-value mapping could disagree for expression-index hidden generated columns.

One bad tracked `TableInfo` shape seen during runtime DDL was:

```text
id                  offset=0 state=public
name                offset=1 state=public
hidden lower(name)  offset=2 state=none
payload             offset=3 state=public
```

The hidden generated column backing the expression index should be public after the DDL finishes. Also, MySQL row events still carry only visible column values. For the table above, a row image is `[id, name, payload]`, so using `ColumnInfo.Offset` directly can read the wrong value or run past the row-value slice.

This PR bumps the TiDB dependency to include pingcap/tidb#69317, so DM's schema tracker gets valid `TableInfo` for expression-index hidden generated columns.

It also centralizes source row-value mapping for DML generation. DML code translates table column offsets to row-image value offsets, and routed source tables get separate cached `WhereHandle`s when they share one downstream table.

#### Tests 

 - Unit test
   - `go test ./pkg/sqlmodel -run 'TestGenDMLWithHiddenColumnBeforeVisibleColumn|TestGenUpdateMultiRowsWithHiddenColumnBeforeVisibleColumn|TestGenInsertMultiRowsWithHiddenColumnBeforeVisibleColumn|TestCausalityKeysInterleavedHiddenColumn|TestPrimaryOrUniqueKeyUpdatedUniqueAfterHiddenColumn|TestGetWhereIdxByDataUniqueAfterHiddenColumn' -count=1`
   - `go test ./dm/pkg/schema -count=1`
   - `go test ./dm/syncer -run 'Test.*ForeignKey|Test.*DownStream|Test.*DML|Test.*Causality' -count=1`
   - `go test ./pkg/sqlmodel -count=1`
 - Manual test
   - Built DM binaries from this PR branch.
   - Ran incremental MySQL 8.0.46 -> TiDB DM tasks with both `multiple-rows: false` and `multiple-rows: true`.
   - Covered normal tables, visible generated columns, hidden generated columns backing expression indexes, mixed DDL/DML, table rename, `MODIFY COLUMN`, normal index changes, and multi-row DML.
   - Reran the real-environment interleaved workload with sleep prevention enabled and a 3600s runtime, so every scheduled DDL block was naturally triggered without manual compensation.
   - The long run continuously wrote DML and naturally covered: add columns plus expression indexes, add a visible generated column, drop an expression index and column, rename the expression-index table, create/alter/add expression index/rename a late table, add/drop a normal index, rename a plain column, and binlog rotation.
   - The long run finished successfully after 1409 iterations. DM ended in `Running / synced=true` with no blocking DDL.
   - Verified sync-diff passed for `real_generated_ddl.long2_expr_r`, `real_generated_ddl.long2_plain`, and `real_generated_ddl.long2_late_r`.
   - Added a short targeted stress test for the sharp DDL-immediate-DML window on a separate expression-index table: `ADD COLUMN` followed immediately by DML, `ADD INDEX` on expression followed immediately by DML, visible generated column DDL followed immediately by DML, `DROP INDEX` followed immediately by DML, and table rename followed immediately by DML.
   - Verified the targeted stress table matched between upstream and downstream by row count, id sum, and checksum: `COUNT(*) = 2005`, `SUM(id) = 2016015`, checksum `4296700167039`.
   - Worker logs did not show the hidden-column layout issue, placeholder/argument mismatch, incorrect post-DDL DML column mapping, `panic`, `index out of range`, `unknown column`, or `execute statement failed`.

No expected performance regression. DML generation reuses cached row-value mapping, and the insert/replace path avoids building a full `WhereHandle` per row.

No.

### Release note 

```release-note
Fix DM replication failures after expression-index hidden generated columns are followed by runtime DDL that interleaves hidden and visible columns.
```


dm/pkg/schema/tracker.go
```diff
@@ -209,15 +209,35 @@ func (tr *Tracker) Exec(ctx context.Context, db string, stmt ast.StmtNode) (errR
 	case *ast.DropDatabaseStmt:
 		return tr.upstreamTracker.DropSchema(tr.se, v)
 	case *ast.CreateTableStmt:
-		return tr.upstreamTracker.CreateTable(tr.se, v)
+		if err := tr.upstreamTracker.CreateTable(tr.se, v); err != nil {
```
joechenrh: You can skip reviewing this part (and corresponding test), since it's better to be fixed in upstream: https://github.com/pingcap/tidb/pull/69317.

pkg/sqlmodel/multirow_test.go
```diff
@@ -233,3 +233,75 @@ func TestGenInsertMultiRows(t *testing.T) {
 	require.Equal(t, "INSERT INTO `db`.`tb` (`c`,`c2`) VALUES (?,?),(?,?) ON DUPLICATE KEY UPDATE `c`=VALUES(`c`),`c2`=VALUES(`c2`)", sql)
 	require.Equal(t, []interface{}{1, 2, 3, 4}, args)
 }
+
+func TestGenInsertMultiRowsWithHiddenColumnBeforeVisibleColumn(t *testing.T) {
```
D3Hunter: maybe unify with below `TestGenUpdateMultiRowsWithHiddenColumnBeforeVisibleColumn`, and have inner subtests

pkg/sqlmodel/multirow.go
```diff
@@ -135,18 +135,16 @@ func GenUpdateSQL(changes ...*RowChange) (string, []any) {
 		whenCaseStmts[i] = whereBuf.String()
 	}
 
-	// Build gegerated columns lower name set to accelerate the following check
-	targetGeneratedColSet := generatedColumnsNameSet(first.targetTableInfo.Columns)
+	first.lazyInitWhereHandle()
+	writableColumns := writableSourceColumns(
```
D3Hunter: `writableColumns` and the corresponding source row-image offsets should be fixed for a given source/target `TableInfo` pair, can we compute/cache them once on the per-source `WhereHandle` (or equivalent source/target cache) instead of rebuilding `writableSourceColumns` for each generated SQL and translating each column through `valueByOffset` in the hot path?

D3Hunter:
rest lgtm

dm/syncer/data_validator.go
```diff
@@ -833,7 +833,13 @@ func (v *DataValidator) genValidateTableInfo(sourceTable *filter.Table, columnCo
 		// todo: might be connection error, then return error, or downstream table not exists, then set state to stopped.
 		return res, err
 	}
-	pk := downstreamTableInfo.WhereHandle.UniqueNotNullIdx
+	// if current TI has more columns, clone and strip columns
+	if len(tableInfo.Columns) > columnCount {
+		tableInfo = tableInfo.Clone()
+		tableInfo.Columns = tableInfo.Columns[:columnCount]
```
OliverS929: Do we need to handle hidden columns here as well?

dm/pkg/schema/tracker.go
```diff
@@ -82,15 +82,52 @@ type downstreamTracker struct {
...
+		return nil
+	}
+	return &DownstreamTableInfo{
+		TableInfo:        dti.TableInfo,
+		whereHandleCache: dti.whereHandleCache,
+	}
+}
+
+// WhereHandle gets or builds the handle for the given source table.
+func (dti *DownstreamTableInfo) WhereHandle(sourceTable *filter.Table, sourceTI *model.TableInfo) *sqlmodel.WhereHandle {
```
OliverS929, line 119: I see per-source WhereHandle caching was added. Shall we add test coverage for shard merge cases where multiple source tables route to the same downstream table but have different hidden-column layouts?

dm/syncer/dml.go
```diff
@@ -142,6 +142,7 @@ func (s *Syncer) genAndFilterInsertDMLs(tctx *tcontext.Context, param *genDMLPar
 	if err != nil {
 		return nil, err
 	}
+	whereHandle := downstreamTableInfo.WhereHandle(param.sourceTable, param.sourceTableInfo)
```
OliverS929, line 145: Since row-value mapping now depends on the source-specific WhereHandle, should the expression-filter path below use the same visible-column mapping?

pkg/sqlmodel/multirow_test.go
joechenrh: Done in https://github.com/joechenrh/tiflow/commit/f9986038ea93cce2414f10a985584140e80a0a53 . Merged the insert/update hidden-column multi-row coverage into one test with inner subtests.

pkg/sqlmodel/multirow.go
joechenrh: Done in https://github.com/joechenrh/tiflow/commit/f9986038ea93cce2414f10a985584140e80a0a53 . `WhereHandle` now caches `writableColumns` together with the per-source row mapper, and DML generation reuses that mapping without building a full where handle on the CDC insert path.

pkg/sqlmodel/row_change.go
```diff
@@ -239,6 +239,18 @@ func (r *RowChange) lazyInitWhereHandle() {
 	r.whereHandle = GetWhereHandle(r.sourceTableInfo, r.targetTableInfo)
 }
 
+// dmlRowMapping returns DML column/value mapping without initializing the full
+// where handle. CDC insert/update paths can use this without paying the WHERE
+// handle build cost, while DM paths reuse the injected cached handle.
+func (r *RowChange) dmlRowMapping() (rowValueMapper, []*timodel.ColumnInfo) {
```
D3Hunter, line 245: Could we consider introducing an explicit internal row-image representation normalized to the source table column offsets here? The binlog row image only carries source visible columns, but after decoding we currently have multiple places translating between visible-row indexes and `TableInfo.Columns` offsets (`rowValueMapper`, expression-filter row expansion, DML arg generation). If the syncer/sqlmodel path converted visible binlog rows into a source-column-addressable row once, DML generation and `SkipDMLByExpression` could share one contract instead of each caller deciding whether the row is visible-width or full-width. The data validator can keep using visible-column rows and avoid this transformation unless needed.

Pros:
- Centralizes the visible-row -> source-column-offset mapping instead of repeating it in expression filtering, DML argument construction, causality, and reduce paths.
- Makes the internal contract clearer: SQL/model logic can address values by source `ColumnInfo.Offset`, while external/raw binlog data remains visible-column based.
- Reduces the chance of future hidden-column offset bugs when a hidden generated column is interleaved before visible columns.

Cons:
- This is a broader refactor than the immediate bug fix, so it may be better as a follow-up if we want to keep this PR small.
- We need to preserve existing visible-column contracts for validator/checkpoint data and `RowValues()` callers, otherwise full-width rows with hidden slots could break comparison/persistence logic.
- Hidden/generated slots must be represented explicitly as absent/materialized, not ordinary SQL NULL values, or downstream logic could accidentally treat a missing binlog value as real data.

D3Hunter: this pr can keep it as is, we can consider this as a later enhancement refactor

joechenrh: I agree that a centralized representation would be better, but this requires a larger refactor. The rationale for the current PR is to avoid introducing additional overhead in the DM/CDC paths while keeping the fix scoped (adds a lightweight mapping for correctness).

D3Hunter (approved):
LGTM

OliverS929 (approved):

joechenrh:
/hold
Let me run another round of real test locally

joechenrh:
I reran the complex real-environment DDL test with the current PR binaries and strengthened the coverage compared with the previous run.

What I ran:
- Started a DM task against MySQL 8.0 -> TiDB using the current PR build of `dm-master`, `dm-worker`, and `dmctl`.
- Reran the interleaved workload with sleep prevention enabled and extended the runtime to 3600s, so every scheduled DDL block could be naturally triggered by the script.
- The workload continuously wrote DML while DDLs were spread across the run.

DDL coverage from the long run:
- `ADD COLUMN extra` + `ADD UNIQUE KEY uk_expr_extra ((LOWER(extra)))`
- `ADD COLUMN mid_col` + `ADD KEY idx_expr_mid ((mid_col + id))`
- `ADD COLUMN tail` + visible generated column `tail_lower`
- `DROP INDEX uk_expr_extra` + `DROP COLUMN extra`
- `RENAME TABLE long2_expr TO long2_expr_r`
- create/alter/add expression index/rename `long2_late` -> `long2_late_r`
- `long2_plain` add `note`, add/drop `idx_payload`, and rename `payload` -> `payload2`
- binlog rotation during the post-DDL DML phase

Validation from the long run:
- The workload finished successfully after 1409 iterations.
- DM ended in `Running` state with `synced: true` and no blocking DDL.
- `sync_diff_inspector` passed for all 3 target tables:
  - `real_generated_ddl.long2_expr_r`
  - `real_generated_ddl.long2_plain`
  - `real_generated_ddl.long2_late_r`

I also added a short targeted stress test for the sharp DDL-immediate-DML window on a separate expression-index table:
- `ADD COLUMN` followed immediately by DML
- `ADD INDEX` on an expression followed immediately by DML
- visible generated column DDL followed immediately by DML
- `DROP INDEX` followed immediately by DML
- table rename followed immediately by DML

The targeted table matched between upstream and downstream by row count, id sum, and checksum: `COUNT(*) = 2005`, `SUM(id) = 2016015`, checksum `4296700167039`.

I did not see the hidden-column layout issue, placeholder/argument mismatch, incorrect post-DDL DML column mapping, `panic`, `index out of range`, `unknown column`, or `execute statement failed` with this PR.

PR #12738, opened 2026-06-23
metrics(ticdc): filter inactive tikv cdc metrics

### LINKED ISSUE #12734: tikv_cdc_min_resolved_ts is not reset when CDC captured region count becomes zero
## What did you do?

We observed TiKV CDC metrics after a region leader transfer.

A region was transferred away from one TiKV store. The old leader TiKV reported `not_leader` to the CDC downstream successfully, and the region was later initialized/listened on the new leader normally.

Example logs from the old leader:

```json
{
  "message": "cdc send region error success",
  "origin_error": "not_leader { region_id: 206739 leader { id: 206954 store_id: 3 } }",
  "region_id": 206739
}
The region was then initialized quickly:
{
  "message": "cdc async incremental scan finished",
  "takes": "1ms",
  "region_id": 206739
}
However, the CDC min resolved-ts lag panel continued to increase linearly on the old TiKV instance.
The dashboard expression is similar to:
scalar(max(pd_cluster_tso{k8s_cluster="$k8s_cluster", tidb_cluster="$tidb_cluster"})) / 1000
-
avg(tikv_cdc_min_resolved_ts{k8s_cluster="$k8s_cluster", tidb_cluster="$tidb_cluster", instance=~"$tikv_instance"} / 1000) by (instance)
> 0
At the same time, the old TiKV instance showed:
tikv_cdc_captured_region_total = 0
tikv_cdc_region_resolve_status{type="resolved"} = 2
tikv_cdc_region_resolve_status{type="unresolved"} = 0



## What did you expect to see?

When a TiKV instance no longer captures any CDC region, stale CDC resolved-ts metrics from this instance should not continue to contribute to CDC min resolved-ts lag.
Either:
tikv_cdc_min_resolved_ts / tikv_cdc_min_resolved_ts_region / tikv_cdc_min_resolved_ts_lag should be reset when there are no captured regions, or
the Grafana/PromQL query should filter out inactive TiKV instances with tikv_cdc_captured_region_total == 0.

## What did you see instead?

What did happen?
The old TiKV instance had tikv_cdc_captured_region_total = 0, which means it was no longer listening to CDC regions.
But tikv_cdc_min_resolved_ts appeared to keep its last reported value. Since PD TSO kept increasing, the dashboard expression kept showing an increasing lag:
pd_cluster_tso - stale tikv_cdc_min_resolved_ts
This made the dashboard look like CDC resolved-ts was stuck, even though logs indicated the region had already moved away and the old TiKV was no longer capturing it.
Investigation
From the code, CDC metrics are updated in Endpoint::on_timeout:
if self.min_resolved_ts != TimeStamp::max() {
    CDC_MIN_RESOLVED_TS_REGION.set(self.min_ts_region_id as i64);
    CDC_MIN_RESOLVED_TS.set(self.min_resolved_ts.physical() as i64);
    CDC_MIN_RESOLVED_TS_LAG.set(...);
}
self.min_resolved_ts = TimeStamp::max();
self.current_ts = TimeStamp::max();
self.min_ts_region_id = 0;
When there is no new min resolved-ts result, the in-memory fields are reset, but the Prometheus gauges are not reset. Therefore, the previously exported gauge value may remain visible and continue to be used by Grafana queries.
This is especially confusing after leader transfer or region scheduling, where an old leader TiKV may no longer capture any region but still exposes the last tikv_cdc_min_resolved_ts.
Related issue
This may be related to, but not exactly the same as:
https://github.com/tikv/tikv/issues/18821
That issue describes resolved-ts stuck after region leader transfer. In our case, the current evidence points more toward stale CDC metrics / dashboard query behavior because tikv_cdc_captured_region_total on the old instance is already 0.
Possible fix
One possible fix is to reset CDC min resolved-ts gauges when no valid min resolved-ts is produced or when capture_regions.len() == 0.
Another possible fix is to update the Grafana query to filter out inactive TiKV instances, for example:
(
  scalar(max(pd_cluster_tso{k8s_cluster="$k8s_cluster", tidb_cluster="$tidb_cluster"})) / 1000
-
  tikv_cdc_min_resolved_ts{k8s_cluster="$k8s_cluster", tidb_cluster="$tidb_cluster", instance=~"$tikv_instance"} / 1000
)
and on(instance)
(
  tikv_cdc_captured_region_total{k8s_cluster="$k8s_cluster", tidb_cluster="$tidb_cluster", instance=~"$tikv_instance"} > 0
)
Could you help confirm whether tikv_cdc_min_resolved_ts is expected to remain exported after the instance no longer captures any CDC region?

### Versions of the cluster

```console
```

```console
(paste TiKV version here)
```

```console
v8.5.5
```

Meowooh:

Issue Number: close #12734

This PR updates the TiCDC Grafana dashboard queries for TiKV CDC min resolved-ts metrics to filter out inactive TiKV instances whose `tikv_cdc_captured_region_total` is `0`.

After CDC region leadership moves away from a TiKV instance, TiKV can still expose stale `tikv_cdc_min_resolved_ts`, `tikv_cdc_min_resolved_ts_region`, and `tikv_cdc_min_resolved_ts_lag` values. Filtering by `tikv_cdc_captured_region_total > 0` prevents these stale values from being shown as an increasing resolved-ts lag in the dashboard.

#### Tests 

 - Manual test
   - `jq empty metrics/grafana/ticdc.json`
   - `bash scripts/check-ticdc-dashboard.sh`
   -  Verified Grafana dashboard after the fix
   before fix
   修复前
   now 
修复后

 - No code

No. This only changes Grafana PromQL expressions and does not affect TiCDC runtime behavior.

No. This PR updates the TiCDC Grafana dashboard directly.

### Release note 

```release-note
Fix the TiCDC Grafana dashboard to ignore inactive TiKV CDC instances when showing TiKV CDC min resolved-ts metrics.
```

### DIFF
--- a/metrics/grafana/ticdc.json
+++ b/metrics/grafana/ticdc.json
@@ -732,7 +732,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "scalar(max(pd_cluster_tso{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}))/1000 - avg(tikv_cdc_min_resolved_ts{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}/1000) by (instance) > 0",
+              "expr": "scalar(max(pd_cluster_tso{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}))/1000 - avg((tikv_cdc_min_resolved_ts{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}/1000) and (tikv_cdc_captured_region_total{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"} > 0)) by (instance) > 0",
               "format": "time_series",
               "hide": false,
               "instant": false,
@@ -11439,7 +11439,7 @@
             },
             {
               "exemplar": true,
-              "expr": "max(tikv_cdc_min_resolved_ts_lag{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}/1000)",
+              "expr": "max((tikv_cdc_min_resolved_ts_lag{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}/1000) and (tikv_cdc_captured_region_total{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"} > 0))",
               "hide": false,
               "interval": "",
               "legendFormat": "tikv",
@@ -16549,7 +16549,7 @@
           "targets": [
             {
               "exemplar": true,
-              "expr": "scalar(max(pd_cluster_tso{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}))/1000 - avg(tikv_cdc_min_resolved_ts{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}/1000) by (instance) > 0",
+              "expr": "scalar(max(pd_cluster_tso{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}))/1000 - avg((tikv_cdc_min_resolved_ts{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}/1000) and (tikv_cdc_captured_region_total{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"} > 0)) by (instance) > 0",
               "format": "time_series",
               "hide": false,
               "instant": false,
@@ -16570,7 +16570,7 @@
             },
             {
               "exemplar": true,
-              "expr": "sum(tikv_cdc_min_resolved_ts{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}) by (instance)",
+              "expr": "sum(tikv_cdc_min_resolved_ts{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"} and (tikv_cdc_captured_region_total{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"} > 0)) by (instance)",
               "format": "time_series",
               "hide": false,
               "interval": "",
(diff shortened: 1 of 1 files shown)


wlwilliamx (approved):

Meowooh:
CI note for `pull_cdc_integration_kafka_test` #489:

This failure looks unrelated to this PR. The job failed in `G00` / `debezium` while running:

`tests/integration_tests/debezium/sql/debezium/mysql-dbz-193.ddl`

The failing DDL contains a `FULLTEXT KEY`, and the current TiDB master artifact used by CI rejects it with:

`Error 8200 (HY000): FULLTEXT index is only supported in starter deployment mode`

The stack trace points to `tests/integration_tests/debezium/src/db_helper.go:59` and `tests/integration_tests/debezium/src/test_cases.go:298`.

This PR only changes `metrics/grafana/ticdc.json`, so it should not affect the Debezium Kafka integration test. I think this is caused by a TiDB artifact / existing test fixture incompatibility rather than this PR.

metrics/grafana/ticdc.json
```diff
@@ -732,7 +732,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "scalar(max(pd_cluster_tso{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}))/1000 - avg(tikv_cdc_min_resolved_ts{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}/1000) by (instance) > 0",
+              "expr": "(scalar(max(pd_cluster_tso{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\"}))/1000 - avg(tikv_cdc_min_resolved_ts{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}/1000) by (instance) > 0) and on(instance) (avg(tikv_cdc_captured_region_total{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$tikv_instance\"}) by (instance) > 0)",
```
lance6716: I see the new expression uses `avg`. I guess it's because the output is a vector with 1 element, and you want to use `avg` to convert the vector to scalar?

If the vector has multiple values, for example, [0.0, 3.0], `avg` will not work. Can you explain in PR description what's the shape and labels of tikv_cdc_captured_region_total?

lance6716 (commented):
please post the screenshots after the fix in the test part of PR description

metrics/grafana/ticdc.json
Meowooh: Thanks for pointing this out. The `avg` here is not intended to convert the vector to a scalar. With `by (instance)`, it still returns one series per TiKV instance, so it can be used with `and on(instance)` to filter the left-hand side resolved-ts series.

`tikv_cdc_captured_region_total` is emitted by TiKV as an `IntGauge` without custom metric labels, and it is set from `capture_regions.len()`. After Prometheus scraping/relabeling, it may have target labels such as `instance`, `job`, `k8s_cluster`, and `tidb_cluster`. The query aggregates by `instance` to collapse possible duplicate scrape labels while preserving one RHS series per TiKV instance.

The value is only used as a boolean filter (`> 0`). For duplicate samples like`[0, 3]`, `avg(...) > 0` still keeps the instance, which matches the intended condition.

lance6716 (commented):
please upload the screenshot to verify the fix. Rest lgtm

lance6716 (approved):

PR #12741, opened 2026-06-24
puller(ticdc): cap resolve lock target ts by PD tso

### LINKED ISSUE #12740: [Bug] TiCDC ScanLock can advance MaxTS and break async commit lock resolution
## Description

This is the TiFlow counterpart of pingcap/ticdc#5418.

In an on-call case, TiCDC stale-lock resolving scanned a region and later failed to resolve residual async-commit locks. The suspected root cause is that TiCDC `ScanLock` advanced the local TiKV store `MaxTS`.

The same stale-lock target timestamp calculation exists in TiFlow's TiCDC puller code:

- `cdc/puller/multiplexing_puller.go`
- `pkg/txnutil/lock_resolver.go`

## Context

Original TiCDC issue: https://github.com/pingcap/ticdc/issues/5418
Original TiCDC fix: https://github.com/pingcap/ticdc/pull/5419
Related TiDB context: https://github.com/pingcap/tidb/issues/65757#issuecomment-4570926210

Observed sequence from the original issue:

1. TiCDC `ScanLock` used `maxVersion=466162474942464000`.
2. The scanned async-commit lock had `min_commit_ts=466162474942464001`.
3. Later lock resolving used the lower primary commit timestamp `466162474850733727`.
4. TiKV rejected it because the commit timestamp was smaller than `min_commit_ts`.
5. Later resolve attempts continued to fail with `commit_ts_expired`.

## Expected Behavior

TiCDC stale-lock scanning and resolving should not advance TiKV local `MaxTS` in a way that makes an async-commit lock's `min_commit_ts` exceed the primary commit timestamp.

The resolve-lock checker should cap the `ScanLock` target timestamp by the latest PD TSO, instead of deriving it only from local clock based resolved-time calculations.

## Actual Behavior

Residual async-commit locks may remain and fail to resolve.

Key evidence from the original issue:

```text
ScanLock maxVersion = 466162474942464000
Lock min_commit_ts = 466162474942464001
Primary commit_ts = 466162474850733727
```

TiKV rejected the lower commit timestamp:

```text
trying to commit with smaller commit_ts than min_commit_ts
```

A later resolve-lock attempt failed with:

```text
commit_ts_expired
```

## Proposed Solution

Port the fix from pingcap/ticdc#5419 to TiFlow:

- Fetch the current TSO from PD before resolving stale locks.
- Cap the resolve-lock target timestamp by that PD current TSO.
- Keep the freshness check for `resolvedTsUpdated` based on the local clock, because it is written with `time.Now()`.
- Add unit tests covering the cap, fence conditions, and CDC/PD clock skew cases.

tenfyzhong:

Issue Number: close #12740

TiCDC stale-lock resolving can derive the `ScanLock` target timestamp from local clock based resolved-time calculations. If the local CDC clock is ahead of PD time, the target timestamp can exceed the latest PD TSO, which may advance TiKV local `MaxTS` and make async-commit residual locks fail to resolve with `commit_ts_expired`.

This ports the same fix as pingcap/ticdc#5419 to TiFlow.

This PR changes the TiCDC multiplexing puller resolve-lock checker to fetch the current TSO from PD before scanning stale locks and cap the stale-lock target timestamp by that PD `currentTs`.

The freshness fence for `resolvedTsUpdated` still uses the local clock because `resolvedTsUpdated` is written with `time.Now()`. The PD-derived timestamp is used only as the upper bound for the `ScanLock` target timestamp.

It also adds unit coverage for clock skew, target timestamp capping, and the existing fence conditions.

#### Tests 

 - Unit test

```console
go test ./cdc/puller -run TestGetResolveLockTargetTs -count=1
go test -tags=intest ./cdc/puller -count=1
go test ./cdc/kv -run TestResolveLock -count=1
```

No compatibility impact is expected. The resolve-lock checker now performs one PD `GetTS` call per resolve-lock tick before scheduling stale-lock scans.

No.

### Release note 

```release-note
Fix a TiCDC stale-lock resolving issue where the ScanLock target timestamp could advance TiKV local MaxTS beyond the latest PD TSO and make async-commit residual locks fail to resolve.
```

### DIFF
--- a/cdc/kv/shared_client.go
+++ b/cdc/kv/shared_client.go
@@ -276,6 +276,14 @@ func (s *SharedClient) AllocSubscriptionID() SubscriptionID {
 	return SubscriptionID(subscriptionIDGen.Add(1))
 }
 
+// GetTS returns the current PD TSO.
+func (s *SharedClient) GetTS(ctx context.Context) (int64, int64, error) {
+	if s.pd == nil {
+		return 0, 0, errors.New("pd client is nil")
+	}
+	return s.pd.GetTS(ctx)
+}
+
 // Subscribe the given table span.
 // NOTE: `span.TableID` must be set correctly.
 // It new a subscribedTable and store it in `s.totalSpans`,
--- a/cdc/puller/multiplexing_puller.go
+++ b/cdc/puller/multiplexing_puller.go
@@ -111,18 +111,26 @@ func (p *tableProgress) handleResolvedSpans(ctx context.Context, e *model.Resolv
 	return
 }
 
-func (p *tableProgress) resolveLock(currentTime time.Time) {
+func (p *tableProgress) getResolveLockTargetTs(currentTime time.Time, currentTs uint64) uint64 {
 	resolvedTsUpdated := time.Unix(p.resolvedTsUpdated.Load(), 0)
 	if !p.initialized.Load() || time.Since(resolvedTsUpdated) < resolveLockFence {
-		return
+		return 0
 	}
 	resolvedTs := p.resolvedTs.Load()
 	resolvedTime := oracle.GetTimeFromTS(resolvedTs)
 	if currentTime.Sub(resolvedTime) < resolveLockFence {
+		return 0
+	}
+
+	return min(currentTs, oracle.GoTimeToTS(resolvedTime.Add(resolveLockFence)))
+}
+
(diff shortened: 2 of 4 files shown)


tenfyzhong:
close https://github.com/tikv/tikv/issues/19755

asddongmen (approved):

hongyunyan (approved):

Issue #12747, opened 2026-06-25
DM: online-DDL cut-over splits a same-name index redefinition as ADD-before-DROP → [ddl:1061] index already exist (ErrCode 44006)


choudharypankaj:
### What did you do?

Replicating Aurora MySQL → TiDB with DM in `task-mode: incremental` and `online-ddl: true`, run a **gh-ost** migration that converts a secondary index from non-unique `KEY` to `UNIQUE KEY` **reusing the same index name**, with the `ADD` listed before the `DROP` in the alter.

1. Start a DM incremental task with `online-ddl: true` for a table that has a non-unique secondary index:
   ```sql
   CREATE TABLE onboarding_table_statuses (
     id BIGINT NOT NULL AUTO_INCREMENT,
     member_employment_id BIGINT NOT NULL,
     status VARCHAR(64) DEFAULT 'pending',
     updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
     PRIMARY KEY (id),
     KEY index_onboarding_table_statuses_on_member_employment_id (member_employment_id)  -- NON-UNIQUE
   ) ENGINE=InnoDB;   -- seeded with rows whose member_employment_id values are unique
   ```
2. Run gh-ost against the upstream, converting the index to UNIQUE **reusing the same name**, with `ADD` before `DROP`:
   ```bash
   gh-ost --host=<aurora> --database=naid --table=onboarding_table_statuses \
     --alter="ADD UNIQUE index_onboarding_table_statuses_on_member_employment_id (member_employment_id), \
              DROP INDEX index_onboarding_table_statuses_on_member_employment_id" \
     --allow-on-master --assume-rbr --execute
   ```
   MySQL/Aurora accepts this combined `ADD,DROP` of a same-named index as one atomic ALTER, so
   gh-ost completes upstream and the cut-over `RENAME` reaches the binlog.

### What did you expect to see?

DM reconstructs the index redefinition at the gh-ost cut-over and applies it cleanly to its internal schema tracker, the same way it does when the alter is written `DROP` before `ADD`, and replication continues. The downstream table ends with the UNIQUE index.

### What did you see instead?

At the cut-over RENAME, DM's online-DDL splitter (`ddl.go:309 "ready to split ddl"` → `ddl.go:329 "resolve sql"`) emits the split statements **preserving the upstream spec order —`ADD` before `DROP`**:

```
[ddl.go:329] ["resolve sql"] appliedDDLs=[
[...]

PR #12750, opened 2026-07-01
syncer(dm): centralize row image layout mapping

joechenrh:

This is a follow-up draft for centralizing row image layout handling after the hidden/generated column fixes. The current code has similar source-column/visible-row/writable-column mapping logic in `WhereHandle`, `RowChange`, expression filters, and validator code, which makes hidden-column layout handling easy to drift.

Issue Number: ref #12747

This PR adds `sqlmodel.RowImageLayout` as the stage-1 shared representation for row image layout metadata:

- source table columns
- visible columns present in binlog row images
- writable downstream columns
- source column offset to row value offset mapping
- expansion from visible-only row image to full source-column layout

`WhereHandle` and `RowChange.dmlRowMapping()` now build their DML/WHERE value mapping from this layout. DM expression filter and validator paths also reuse the same layout helpers instead of maintaining local hidden-column filtering/mapping logic.

#### Tests

 - Unit test

Verified locally:

```bash
go test ./pkg/sqlmodel
go test ./dm/syncer -run 'TestTableInfoForVisibleColumnCount|TestValidatorWorkerUsesVisibleColumns|TestSkipDMLByExpressionWithHiddenColumnBeforeVisibleColumn'
git diff --check
```

No expected compatibility impact. This is intended as an internal refactor. Performance impact should be neutral: the new layout replaces existing mapper/helper construction and removes duplicated mapping logic.

No. This is an internal code structure change.

### Release note

```release-note
None
```

### DIFF
--- a/dm/syncer/validate_worker.go
+++ b/dm/syncer/validate_worker.go
@@ -320,7 +320,7 @@ func (vw *validateWorker) batchValidateRowChanges(rows []*rowValidationJob, dele
 	firstRow := rows[0].row
 	cond := &Cond{
 		TargetTbl: firstRow.TargetTableID(),
-		Columns:   visibleColumns(firstRow.SourceTableInfo().Columns),
+		Columns:   sqlmodel.VisibleColumns(firstRow.SourceTableInfo().Columns),
 		PK:        firstRow.UniqueNotNullIdx(),
 		PkValues:  pkValues,
 	}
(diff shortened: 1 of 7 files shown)


dm/syncer/validate_worker.go
```diff
@@ -453,18 +453,6 @@ func (vw *validateWorker) resetErrorRows() {
...
 }
 
-// TODO(joechenrh): Share visible-row/source-column offset mapping with sqlmodel
-// instead of maintaining validator-specific hidden column helpers.
-func visibleColumns(columns []*model.ColumnInfo) []*model.ColumnInfo {
-	ret := make([]*model.ColumnInfo, 0, len(columns))
-	for _, col := range columns {
-		if !col.Hidden {
-			ret = append(ret, col)
-		}
```
GMHDBJD: test

Issue #12751, opened 2026-07-02
DM precheck fails on HeatWave SHOW GRANTS and FLUSH_TABLES privilege


GMHDBJD:
## Bug Report

DM precheck fails against Oracle HeatWave on AWS even when the source account can run the required dump operation.

### What happened

When running DM `check-task` or `start-task` with precheck enabled against HeatWave MySQL 9.7, the precheck can fail in two ways:

1. `SHOW GRANTS` parsing fails on HeatWave/MySQL 8 role grants such as:

```sql
GRANT `administrator`@`%` TO `user`@`%` WITH ADMIN OPTION
```

2. The dump privilege checker reports missing global `RELOAD`, although HeatWave grants the `FLUSH_TABLES` dynamic privilege and `FLUSH TABLES WITH READ LOCK` succeeds.

### Expected behavior

DM precheck should support HeatWave/MySQL 8 `SHOW GRANTS` output and should treat global `FLUSH_TABLES` as satisfying the FTWRL-style dump privilege requirement.

### Impact

This blocks the standard DM migration workflow for HeatWave sources when precheck is enabled. Users must skip precheck even though the actual full dump and incremental sync can work.

### Environment

- Upstream: Oracle HeatWave on AWS MySQL 9.7
- DM: upstream master local build
- Scenario: HeatWave source to TiDB downstream, `task-mode: all`, precheck enabled

GMHDBJD:
Fix PR: https://github.com/pingcap/tiflow/pull/12752

PR #12752, opened 2026-07-02
checker: support HeatWave grants in precheck

GMHDBJD:

Issue Number: ref #12751

This PR fixes DM precheck compatibility with Oracle HeatWave/MySQL 8 style `SHOW GRANTS` output.

Problems addressed:

- HeatWave may return role grants with `WITH ADMIN OPTION`, which TiDB parser cannot parse as `GrantRoleStmt` yet.
- HeatWave may return partial revoke statements in `SHOW GRANTS`; overlapping revokes must not be reported as sufficient privileges.
- HeatWave grants `FLUSH_TABLES` as a dynamic privilege. It should satisfy the FTWRL dump privilege requirement currently modeled as `RELOAD` in DM precheck.

- Add a DM checker-local `showGrants` helper to tolerate role grants with `WITH ADMIN OPTION` while still discovering active roles via `SHOW GRANTS ... USING ...`.
- Make role discovery best-effort: unsupported non-role grant syntax, such as MariaDB-specific grant rows, no longer fails before privilege verification can run with the correct parser mode.
- Skip role grants with `WITH ADMIN OPTION` during direct privilege verification, because role privileges are collected through `SHOW GRANTS ... USING ...`.
- Model overlapping `REVOKE` statements against the required privilege set, so a global grant plus a partial revoke on a checked schema/table fails conservatively.
- Treat global extended privilege `FLUSH_TABLES` as satisfying the dump privilege checker’s FTWRL requirement, without treating sibling `FLUSH_*` dynamic privileges as equivalent.
- Add unit coverage for HeatWave-style grants, role grants with `WITH ADMIN OPTION`, partial revokes, MariaDB-specific grant rows in role discovery, non-ASCII role grant suffix trimming, and `FLUSH_TABLES` dynamic privilege handling.

### Check List

Tests:

- [x] Unit test
- [x] Manual test with real HeatWave source

Manual validation:

- `go test ./dm/pkg/checker ./dm/checker`
- Built `dm-master`, `dm-worker`, and `dmctl` from this branch.
- Ran DM against Oracle HeatWave on AWS MySQL 9.7 with precheck enabled and without `ignore-checking-items`.
- `check-task` passed with only the MySQL version warning.
- `start-task` succeeded.
- Full load plus incremental insert/update/delete synced correctly to TiDB.

Related issue: https://github.com/pingcap/tiflow/issues/12751

### Release note

```release-note
Fix DM precheck compatibility with Oracle HeatWave `SHOW GRANTS` output and dynamic `FLUSH_TABLES` privileges.
```

### DIFF
--- a/dm/pkg/checker/privilege.go
+++ b/dm/pkg/checker/privilege.go
@@ -86,7 +86,7 @@ func (pc *SourceDumpPrivilegeChecker) Check(ctx context.Context) *Result {
 		Extra: fmt.Sprintf("address of db instance - %s:%d", pc.dbinfo.Host, pc.dbinfo.Port),
 	}
 
-	grants, err := dbutil.ShowGrants(ctx, pc.db, "", "")
+	grants, err := showGrants(ctx, pc.db, "", "")
 	if err != nil {
 		markCheckError(result, err)
 		return result
@@ -149,7 +149,7 @@ func (pc *SourceReplicatePrivilegeChecker) Check(ctx context.Context) *Result {
 		Extra: fmt.Sprintf("address of db instance - %s:%d", pc.dbinfo.Host, pc.dbinfo.Port),
 	}
 
-	grants, err := dbutil.ShowGrants(ctx, pc.db, "", "")
+	grants, err := showGrants(ctx, pc.db, "", "")
 	if err != nil {
 		markCheckError(result, err)
 		return result
@@ -193,7 +193,7 @@ func (t *TargetPrivilegeChecker) Check(ctx context.Context) *Result {
 		State: StateSuccess,
 		Extra: fmt.Sprintf("address of db instance - %s:%d", t.dbinfo.Host, t.dbinfo.Port),
 	}
-	grants, err := dbutil.ShowGrants(ctx, t.db, "", "")
+	grants, err := showGrants(ctx, t.db, "", "")
 	if err != nil {
 		markCheckError(result, err)
 		return result
@@ -303,9 +303,13 @@ func VerifyPrivileges(
 		p.SetMariaDB(true)
 	}
 
+	requiredPrivs := cloneRequiredPrivs(lackPrivs)
 	for _, grant := range grants {
-		if len(lackPrivs) == 0 {
-			break
+		if len(lackPrivs) == 0 && !isRevokeGrant(grant) {
+			continue
+		}
+		if shouldSkipGrantForPrivilegeCheck(grant) {
+			continue
 		}
 		node, err := p.ParseOneStmt(grant, "", "")
 		if err != nil {
@@ -313,8 +317,11 @@ func VerifyPrivileges(
 		}
 		grantStmt, ok := node.(*ast.GrantStmt)
 		if !ok {
-			switch node.(type) {
+			switch stmt := node.(type) {
 			case *ast.GrantProxyStmt, *ast.GrantRoleStmt:
+				continue
+			case *ast.RevokeStmt:
+				restoreRevokedPrivs(lackPrivs, requiredPrivs, stmt)
 				continue
 			default:
 				return nil, errors.Errorf("%s is not grant statement", grant)
@@ -330,6 +337,16 @@ func VerifyPrivileges(
 		switch grantStmt.Level.Level {
 		case ast.GrantLevelGlobal:
 			for _, privElem := range grantStmt.Privs {
+				if privElem.Priv == mysql.ExtendedPriv {
+					// MySQL 8.0/HeatWave expose FLUSH_TABLES as a dynamic privilege.
+					// It is sufficient for FLUSH TABLES WITH READ LOCK, which is the
+					// actual operation behind DM's historical RELOAD privilege check.
+					// Do not treat sibling FLUSH_* dynamic privileges as equivalent.
+					if strings.EqualFold(privElem.Name, "FLUSH_TABLES") {
+						delete(lackPrivs, mysql.ReloadPriv)
+					}
+					continue
+				}
 				// all privileges available at a given privilege level (except GRANT OPTION)
 				// from https://dev.mysql.com/doc/refman/5.7/en/privileges-provided.html#priv_all
 				if privElem.Priv == mysql.AllPriv {
@@ -339,7 +356,8 @@ func VerifyPrivileges(
 						}
 						continue
 					}
-					return nil, nil
+					lackPrivs = make(map[mysql.PrivilegeType]priv)
+					continue
 				}
 				// mysql> show master status;
 				// ERROR 1227 (42000): Access denied; you need (at least one of) the SUPER, REPLICATION CLIENT privilege(s) for this operation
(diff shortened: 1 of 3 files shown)


dm/pkg/checker/privilege.go
```diff
@@ -436,6 +445,114 @@ func VerifyPrivileges(
...
+	for i, role := range roles {
+		if i > 0 {
+			builder.WriteString(", ")
+		}
+		builder.WriteString(role)
+	}
+	return readGrants(builder.String())
+}
+
+func shouldIgnoreGrant(grant string) bool {
```
D3Hunter: maybe called containsWithAdminOption

dm/pkg/checker/privilege.go
```diff
@@ -307,14 +307,17 @@ func VerifyPrivileges(
 		if len(lackPrivs) == 0 {
 			break
 		}
+		if shouldIgnoreGrant(grant) {
+			continue
+		}
```
D3Hunter: as we have consider the WITH ADMIN OPTION inside `showGrants`, we can remove this check?

dm/pkg/checker/privilege.go
```diff
@@ -330,6 +333,12 @@ func VerifyPrivileges(
 		switch grantStmt.Level.Level {
 		case ast.GrantLevelGlobal:
 			for _, privElem := range grantStmt.Privs {
+				if privElem.Priv == mysql.ExtendedPriv {
+					if strings.EqualFold(privElem.Name, "FLUSH_TABLES") {
```
D3Hunter, line 342: is this HW specific? maybe add a comment

D3Hunter (commented):
### Summary

- Total findings: 4
- Inline comments: 4

<details>

#### ⚠️ **[Major]** (2)

1. Partial revokes can be reported as sufficient privileges (`dm/pkg/checker/privilege.go:320`)
2. Version-specific MariaDB grants can fail before MariaDB parsing is enabled (`dm/pkg/checker/privilege.go:501`)

#### 🟡 **[Minor]** (1)

1. Grant predicate name hides the role-admin case (`dm/pkg/checker/privilege.go:505 and dm/pkg/checker/privilege.go:534`)

#### ℹ️ **[Info]** (1)

1. FLUSH_TABLES-to-RELOAD mapping needs rationale (`dm/pkg/checker/privilege.go:336`)

</details>

dm/pkg/checker/privilege.go
```diff
@@ -307,14 +307,17 @@ func VerifyPrivileges(
...
+		}
 		node, err := p.ParseOneStmt(grant, "", "")
 		if err != nil {
 			return nil, errors.New(err.Error())
 		}
 		grantStmt, ok := node.(*ast.GrantStmt)
 		if !ok {
 			switch node.(type) {
-			case *ast.GrantProxyStmt, *ast.GrantRoleStmt:
+			case *ast.GrantProxyStmt, *ast.GrantRoleStmt, *ast.RevokeStmt:
```
D3Hunter: #### ⚠️ **[Major]** Partial revokes can be reported as sufficient privileges

**Why**
The new `*ast.RevokeStmt` branch treats parsed partial-revoke rows from `SHOW GRANTS` as non-operative metadata, but MySQL partial revokes subtract privileges from earlier global grants for a specific schema or table. The verifier only removes privileges from `lackPrivs`; it never applies a revoke back to the required table or schema set.

**Scope**
dm/pkg/checker/privilege.go:320

**Risk if unchanged**
A source account with a global required privilege plus `REVOKE ... ON checked_db.*` can pass DM's dump privilege precheck even though the account is denied on the checked schema or table. DM may then start after a checker success and fail later during dump/load/replication access, and an unsupported partial-revoke case becomes a silent compatibility pass.

**Evidence**
`VerifyPrivileges` now continues on `*ast.RevokeStmt`, while the global grant branch deletes required privileges from `lackPrivs`. The concrete state transition is: pending `SelectPriv` for `app.t` -> `GRANT SELECT ON *.*` clears it -> `REVOKE SELECT ON app.*` is skipped -> final result incorrectly has no missing `SELECT`. The parser AST exposes `RevokeStmt.Privs` and `RevokeStmt.Level`, and the added test at `dm/pkg/checker/privilege_test.go:219` uses `REVOKE ... ON `sys`.*` without a checked `sys` table, so it does not cover an overlapping revoke.

**Change request**
Please model `RevokeStmt` for required privileges, or fail the precheck conservatively when a partial revoke overlaps the checked schema/table; add a unit test for global `SELECT` plus `REVOKE SELECT ON checked_db.*` where the checker must fail.

dm/pkg/checker/privilege.go
```diff
@@ -436,6 +445,114 @@ func VerifyPrivileges(
...
+
+	grants, err := readGrants(query)
+	if err != nil {
+		return nil, err
+	}
+
+	// For MySQL 8.0, collect granted roles and read grants using those roles.
+	// HeatWave SHOW GRANTS may append `WITH ADMIN OPTION` to role grants, which
+	// TiDB parser cannot parse yet. Strip the suffix for role discovery only.
+	var roles []string
```
D3Hunter, line 688: #### ⚠️ **[Major]** Version-specific MariaDB grants can fail before MariaDB parsing is enabled

**Why**
The new `showGrants` helper pre-parses every returned grant only to discover roles, but it does so with a plain TiDB parser before the existing MariaDB parser mode in `VerifyPrivileges` can run.

**Scope**
dm/pkg/checker/privilege.go:501

**Risk if unchanged**
MariaDB sources can get a checker error for grant syntax that the existing verifier is explicitly prepared to parse, changing the precheck result from a privilege decision into an early parse failure.

**Evidence**
`VerifyPrivileges` enables `p.SetMariaDB(true)` when the stored version contains `MariaDB`, with a comment calling out `BINLOG MONITOR and other MariaDB things`. The new `showGrants` has no version parameter and returns immediately on any parse error before `verifyPrivilegesWithResult` receives the grant list and version.

**Change request**
Please either pass the DB version into `showGrants` and use the same parser mode, or avoid failing role discovery on non-role grant parse errors; add a case for a MariaDB-specific grant row to keep this path covered.

dm/pkg/checker/privilege.go
```diff
@@ -436,6 +445,114 @@ func VerifyPrivileges(
...
+	p := parser.New()
+	for _, grant := range grants {
+		grantForParse := grant
+		if shouldIgnoreGrant(grant) {
```
D3Hunter: #### 🟡 **[Minor]** Grant predicate name hides the role-admin case

**Why**
`shouldIgnoreGrant` reads like a general privilege-check decision, but the implementation only detects role grants with `WITH ADMIN OPTION`; in `showGrants` the same predicate is used to trim and parse the role instead of ignoring it.

**Scope**
dm/pkg/checker/privilege.go:505 and dm/pkg/checker/privilege.go:534

**Risk if unchanged**
Future changes can reuse the predicate as a broad ignore gate and accidentally skip grant forms that still need parsing, or miss that role discovery depends on this exact HeatWave role-grant shape.

**Evidence**
The predicate returns true only for `GRANT ... TO ... WITH ADMIN OPTION` statements without ` ON `. `VerifyPrivileges` continues on it, while `showGrants` treats the same true value as `trimAdminOption(grant)` before parsing roles.

**Change request**
Prefer a more specific name, for example `isRoleGrantWithAdminOption`, and use a separate `shouldSkipGrantForPrivilegeCheck` wrapper if the verification path needs the ignore wording.

dm/pkg/checker/privilege.go
```diff
@@ -330,6 +333,12 @@ func VerifyPrivileges(
 		switch grantStmt.Level.Level {
 		case ast.GrantLevelGlobal:
 			for _, privElem := range grantStmt.Privs {
+				if privElem.Priv == mysql.ExtendedPriv {
```
D3Hunter, line 337: #### ℹ️ **[Info]** FLUSH_TABLES-to-RELOAD mapping needs rationale

**Why**
The new dynamic-privilege branch maps `FLUSH_TABLES` to `mysql.ReloadPriv`, which is a non-obvious cross-vendor privilege equivalence.

**Scope**
dm/pkg/checker/privilege.go:336

**Risk if unchanged**
Maintainers may treat other `FLUSH_*` dynamic privileges the same way, or remove the special case, because the intent is not captured near the enforcement point.

**Evidence**
The code deletes `mysql.ReloadPriv` only for `mysql.ExtendedPriv` named `FLUSH_TABLES`, while the adjacent comments still describe only `AllPriv`; the added tests separately assert `FLUSH_STATUS` does not satisfy `RELOAD`.

**Change request**
Please add a short why-comment for this branch explaining why `FLUSH_TABLES` is accepted as satisfying `RELOAD` and why sibling `FLUSH_*` extended privileges are not.

dm/pkg/checker/privilege.go
```diff
@@ -436,6 +454,319 @@ func VerifyPrivileges(
...
+			builder.WriteString(", ")
+		}
+		builder.WriteString(role)
+	}
+	return readGrants(builder.String())
+}
+
+func shouldSkipGrantForPrivilegeCheck(grant string) bool {
+	return isRoleGrantWithAdminOption(grant)
+}
```
joechenrh: Seem redundant

dm/pkg/checker/privilege.go
```diff
@@ -436,6 +454,319 @@ func VerifyPrivileges(
...
+	// HeatWave SHOW GRANTS may append `WITH ADMIN OPTION` to role grants, which
+	// TiDB parser cannot parse yet. Strip the suffix for role discovery only.
+	var roles []string
+	p := parser.New()
+	for _, grant := range grants {
+		grantForParse := grant
+		if isRoleGrantWithAdminOption(grant) {
+			grantForParse = trimAdminOption(grant)
+		}
+		node, err := p.ParseOneStmt(grantForParse, "", "")
```
joechenrh: Not sure if we can trim directly
```go
		node, err := p.ParseOneStmt(trimAdminOption(grant), "", "")
```

dm/pkg/checker/privilege.go
```diff
@@ -436,6 +454,319 @@ func VerifyPrivileges(
...
+	if requiredPriv.needGlobal {
+		lackPrivs[privName] = priv{needGlobal: true}
+		return
+	}
+
+	switch revokeLevel.Level {
+	case ast.GrantLevelGlobal:
+		mergePriv(lackPrivs, privName, requiredPriv)
+	case ast.GrantLevelDB:
+		dbPatChar, dbPatType := stringutil.CompilePattern(revokeLevel.DBName, '\\')
```
D3Hunter: [P2] This partial-revoke path treats schema names as patterns. `CompilePattern` makes `_` and `%` wildcards, but MySQL partial revokes are schema-level restrictions and, when `partial_revokes` is enabled, unescaped `_` and `%` in schema names are interpreted literally.

Readable example:

~~~sql
GRANT SELECT ON *.* TO u1;
REVOKE SELECT ON `db_%`.* FROM u1;
~~~

The `REVOKE` row restricts only the literal schema named `db_%`; it should not remove `SELECT` for tables in schemas such as:

~~~text
db_01.t1
db_foo.t1
~~~

With the current matcher, `db_%` matches those schema names, so a precheck for `db_01.t1` can report a missing `SELECT` privilege even though that partial revoke does not apply. Consider comparing partial-revoke schema names literally, or escaping `_` and `%` before calling `CompilePattern`, and add a regression test for literal `_`/`%` schema names.

D3Hunter (commented):
Suggestion: consider modeling `SHOW GRANTS` as a privilege snapshot before comparing it with `requiredPrivs`.

The current implementation starts from `lackPrivs`, subtracts requirements when it sees a `GRANT`, then adds requirements back when it sees a partial `REVOKE`. That is workable for this patch, but the mental model is a little indirect because `SHOW GRANTS` is not really an ordered mutation log. It is a description of effective grants plus restrictions. A model like “parse grants/revokes into a checker-local privilege abstraction, then ask whether each required privilege is satisfied” would be more direct and easier to review.

One possible shape:

```go
type privilegeSnapshot struct {
    grants       []privilegeFact
    restrictions []privilegeFact // partial REVOKE rows from SHOW GRANTS
}

type privilegeFact struct {
    privs []*ast.PrivElem
    level *ast.GrantLevel
}

type requiredScope struct {
    needGlobal bool
    db         string
    table      string
}
```

Then `VerifyPrivileges` could become:

```go
func VerifyPrivileges(
    grants []string,
    requiredPrivs map[mysql.PrivilegeType]priv,
    version string,
) (map[mysql.PrivilegeType]priv, error) {
    if len(grants) == 0 {
        return nil, errors.New("there is no such grant defined for current user on host '%%'")
    }

    snapshot, err := buildPrivilegeSnapshot(grants, version)
    if err != nil {
        return nil, err
    }

    return snapshot.Missing(requiredPrivs), nil
}
```

And the core evaluation could be requirement-driven:

```go
func (s *privilegeSnapshot) Missing(required map[mysql.PrivilegeType]priv) map[mysql.PrivilegeType]priv {
    missing := cloneRequiredPrivs(required)

    for privName, requirement := range required {
        if requirement.needGlobal {
            if s.Has(privName, requiredScope{needGlobal: true}) {
                delete(missing, privName)
            }
            continue
        }

        for dbName, dbReq := range requirement.dbs {
            if dbReq.wholeDB {
                if s.Has(privName, requiredScope{db: dbName}) {
                    delete(missing[privName].dbs, dbName)
                }
                continue
            }

            for tableName := range dbReq.tables {
                if s.Has(privName, requiredScope{db: dbName, table: tableName}) {
                    delete(missing[privName].dbs[dbName].tables, tableName)
                }
            }
        }
    }

    pruneEmptyPrivs(missing)
    return missing
}
```

I would make `Has` predicate-based rather than materializing all effective privileges. For example, `GRANT SELECT ON *.*` covers every schema/table except the schemas restricted by partial revoke rows, so the checker should keep facts and evaluate coverage lazily:

```go
func (s *privilegeSnapshot) Has(priv mysql.PrivilegeType, target requiredScope) bool {
    for _, grant := range s.grants {
        if !grant.covers(priv, target) {
            continue
        }
        if s.isRestricted(grant, priv, target) {
            continue
        }
        return true
    }
    return false
}
```

Suggested helper split:

```go
func buildPrivilegeSnapshot(grants []string, version string) (*privilegeSnapshot, error)
func (s *privilegeSnapshot) Missing(required map[mysql.PrivilegeType]priv) map[mysql.PrivilegeType]priv
func (s *privilegeSnapshot) Has(priv mysql.PrivilegeType, target requiredScope) bool
func (f privilegeFact) covers(priv mysql.PrivilegeType, target requiredScope) bool
func (s *privilegeSnapshot) isRestricted(grant privilegeFact, priv mysql.PrivilegeType, target requiredScope) bool
func pruneEmptyPrivs(missing map[mysql.PrivilegeType]priv)
```

Pros:

- The model becomes order-independent. `SHOW GRANTS` rows are treated as facts, not as a sequence of mutations to `lackPrivs`.
- The code reads closer to the problem statement: parse current grants, then check required privileges.
- Names like `restoreRevokedPrivs`, `restoreRequiredPrivAtLevel`, and `toRestore` become unnecessary; the concepts become `grants`, `partialRevokes`, `covers`, `isRestricted`, and `missing`.
- It is easier to test parsing separately from privilege coverage, including `ALL`, `SUPER` satisfying `REPLICATION CLIENT`, and `FLUSH_TABLES` satisfying the current `RELOAD`/FTWRL requirement.
- It avoids relying on `SHOW GRANTS` output order. The current subtract/add-back approach only works naturally if broad grants are processed before partial revokes.

Cons / cautions:

- This can grow into a full MySQL ACL engine if we are not careful. I would keep it checker-local and only model the privilege shapes DM currently checks.
- Partial revoke semantics need to be precise. A partial `REVOKE SELECT ON db1.*` should invalidate a global `SELECT ON *.*` grant for `db1`, but we should be careful about how it interacts with explicit narrower grants such as `GRANT SELECT ON db1.t1` if those can appear together. That case is worth either verifying against MySQL/HeatWave or documenting as out of scope.
- The refactor is larger than the current patch. If the goal is the smallest compatibility fix, the current approach is acceptable; if we want long-term readability and fewer corner-case surprises, the snapshot model is a better direction.

In short, I think this PR is solving the right compatibility issue, but the privilege evaluation would be easier to maintain if `SHOW GRANTS` were first normalized into a small `privilegeSnapshot` abstraction and `lackPrivs` were computed from that snapshot, instead of being mutated directly while parsing each grant row.

dm/pkg/checker/privilege.go
```diff
@@ -436,6 +454,319 @@ func VerifyPrivileges(
...
+				}
+				clonedDBPrivs.tables[tableName] = clonedTablePrivs
+			}
+		}
+		cloned.dbs[dbName] = clonedDBPrivs
+	}
+	return cloned
+}
+
+func restoreRevokedPrivs(
```
D3Hunter, line 474: Could you add table-driven tests for the revoke-restoration path? The current coverage mostly proves one `REVOKE SELECT ON db1.*` case, but this function restores privileges across global/db/table levels and then merges them back into `lackPrivs`.

Suggested cases:
```go
// FLUSH_TABLES satisfied RELOAD, then a global revoke should make it missing again.
grants: []string{
    "GRANT SELECT ON *.* TO `dmtest`@`%`",
    "GRANT FLUSH_TABLES ON *.* TO `dmtest`@`%`",
    "REVOKE FLUSH_TABLES ON *.* FROM `dmtest`@`%`",
}
// expect: lack of RELOAD global (*.*) privilege

// Table-level revoke should only restore the revoked table.
grants: []string{
    "GRANT RELOAD, SELECT ON *.* TO `dmtest`@`%`",
    "REVOKE SELECT ON `db1`.`tb1` FROM `dmtest`@`%`",
}
checkTables: []filter.Table{{Schema: "db1", Name: "tb1"}, {Schema: "db1", Name: "tb2"}}
// expect: only `db1`.`tb1` is reported missing
```

That would make the new clone/restore/merge behavior observable instead of relying only on the broad HeatWave happy path.

dm/pkg/checker/privilege.go
```diff
@@ -436,6 +454,319 @@ func VerifyPrivileges(
...
+	}
+}
+
+func revokePrivCoversRequiredPriv(
+	revokedPriv *ast.PrivElem,
+	requiredPriv mysql.PrivilegeType,
+	level ast.GrantLevelType,
+	requiredGlobal bool,
+) bool {
+	switch revokedPriv.Priv {
```
D3Hunter, line 500: Please cover the special mapping branches here. The focused package coverage shows this helper is only partially exercised, and `isGlobalOnlyPriv` is currently not reached by tests.

Examples worth adding:
```go
// DB-level ALL revoke should not restore global-only requirements.
grants: []string{
    "GRANT ALL PRIVILEGES ON *.* TO `dmtest`@`%`",
    "REVOKE ALL PRIVILEGES ON `db1`.* FROM `dmtest`@`%`",
}
// For replication checks, expect no missing REPLICATION CLIENT / REPLICATION SLAVE from this DB-level revoke.

// SUPER is accepted as satisfying REPLICATION CLIENT, so revoking SUPER globally should restore that requirement.
grants: []string{
    "GRANT REPLICATION SLAVE, SUPER ON *.* TO `dmtest`@`%`",
    "REVOKE SUPER ON *.* FROM `dmtest`@`%`",
}
// expect: lack of REPLICATION CLIENT global (*.*) privilege

// FLUSH_TABLES is special, but sibling dynamic privileges must not be equivalent.
// REVOKE FLUSH_STATUS ON *.* should not restore RELOAD; REVOKE FLUSH_TABLES ON *.* should.
```

Direction: prefer adding these as end-to-end cases in `TestVerifyDumpPrivileges` and `TestVerifyReplicationPrivileges`, so the user-facing error strings are verified rather than only private helpers.

D3Hunter (commented):
I checked the focused checker tests. They pass, but the new tests still leave important revoke/role-handling branches unexercised. Suggested targeted coverage below.

dm/pkg/checker/privilege.go
```diff
@@ -436,6 +454,319 @@ func VerifyPrivileges(
...
+		}
+		for tableName, tablePrivs := range dbPrivs.tables {
+			existingDBPrivs.tables[tableName] = cloneTablePriv(tablePrivs)
+		}
+		existingPriv.dbs[dbName] = existingDBPrivs
+	}
+	lackPrivs[privName] = existingPriv
+}
+
+func showGrants(ctx context.Context, db dbutil.QueryExecutor, user, host string) ([]string, error) {
```
D3Hunter, line 635: Since this local helper replaces `dbutil.ShowGrants` for dump, replication, target, and connection-number checkers, could you add sqlmock coverage for the branches that are still untested?

Suggested cases:
```go
// Named user should keep the original helper behavior.
grants, err := showGrants(ctx, db, "dmuser", "")
// expect query: SHOW GRANTS FOR 'dmuser'@'%'

// Multiple roles should be included in the USING query.
// first rows include: GRANT `r1`@`%`,`r2`@`%` TO `dmtest`@`%` WITH ADMIN OPTION
// expect second query: SHOW GRANTS FOR CURRENT_USER USING `r1`@`%`, `r2`@`%`

// If the second SHOW GRANTS ... USING query fails, showGrants should return that error.
```

These would protect compatibility with the original helper while covering the HeatWave-specific role-discovery changes.

D3Hunter (approved):

dm/pkg/checker/privilege.go
```diff
@@ -436,6 +454,319 @@ func VerifyPrivileges(
...
+}
+
+func clonePriv(privs priv) priv {
+	cloned := priv{needGlobal: privs.needGlobal}
+	if privs.dbs == nil {
+		return cloned
+	}
+	cloned.dbs = make(map[string]dbPriv, len(privs.dbs))
+	for dbName, dbPrivs := range privs.dbs {
+		clonedDBPrivs := dbPriv{wholeDB: dbPrivs.wholeDB}
```
joechenrh: Seem duplicated with `cloneDBPriv`

joechenrh (approved):

Issue #12765, opened 2026-07-24
DM: stopped task triggers an inflated binlog gap alert after a worker-instance switch


expxiaoli:
## What did you do?

1. Start a DM task and let it enter incremental replication.
2. Stop the task through the DM OpenAPI.
3. After the source is assigned to another DM worker instance, inspect the task status and the following alert:

`DM_binlog_file_gap_between_master_syncer`

In the observed case:

- The stop request completed successfully.
- `dm_worker_task_state` changed from `2` (Running) to `3` (Paused).
- Replication throughput became zero.
- No subsequent start or resume request was found.
- The same task and source were later observed on another worker instance.

The available evidence confirms a worker-instance switch, but does not distinguish whether it was caused by a restart, rescheduling, or failover.

## What did you expect to see?

1. A task that has been intentionally stopped or paused should not trigger the master-to-syncer binlog gap alert.
2. When a stopped task is loaded on another worker, its syncer binlog-file and position gauges should be initialized from the persisted checkpoint instead of starting at zero.

## What did you see instead?

On the new worker instance, the metrics were:

```text
dm_syncer_binlog_file{node="master"} = 347650
dm_syncer_binlog_file{node="syncer"} = 0
```

The alert therefore reported a gap of `347650`.

However, the persisted checkpoint was:

```text
binary-log.346652:560567
```

The real file-number gap was approximately:

```text
347650 - 346652 = 998
```

The task remained stopped, with:

```text
rps:        0
total_rows: 0
```

The actual gap continued growing because the upstream database was still generating binlogs, but the reported alert value was significantly inflated.

## Root cause analysis

This symptom is caused by two independent DM observability defects.

### 1. The alert is not task-state-aware

The current rule evaluates every exported task metric without excluding paused or stopped tasks:

```promql
dm_syncer_binlog_file{node="master"}
- ON(instance, task, job)
dm_syncer_binlog_file{node="syncer"} > 1
```

Source:

https://github.com/pingcap/tiflow/blob/e8736df4fe74199cc8b8089530af55a49fdfecf7/dm/metrics/alertmanager/dm_worker.rules.yml#L127-L137

### 2. The syncer gauge is not restored from the checkpoint

When a stopped task is loaded on a worker, DM initializes its units but does not enter the replication loop:

https://github.com/pingcap/tiflow/blob/e8736df4fe74199cc8b8089530af55a49fdfecf7/dm/worker/subtask.go#L201-L223

[...]

PR #12766, opened 2026-07-27
metrics(dm): ignore binlog gap for non-running tasks

expxiaoli:

Issue Number: ref #12765

`DM_binlog_file_gap_between_master_syncer` evaluates paused and stopped tasks. These tasks can continue exposing master and syncer binlog metrics and trigger misleading critical alerts even though replication is intentionally not running.

- Join the binlog gap expression with `dm_worker_task_state == 2`, so the alert only evaluates Running tasks.
- Keep the expression stored in the alert labels consistent with the rule expression.

#### Tests

- Manual test

```console
git diff --check
```

No automated rule test is added. The repository does not have an existing promtool test setup, and a YAML metadata assertion would not validate the PromQL behavior.

No performance regression or compatibility break is expected. This change only affects alert evaluation and does not change DM runtime behavior. Paused and stopped tasks will no longer trigger this alert.

No. The alert rule is updated directly without changing metric names, labels, or user-facing configuration.

### Release note

```release-note
Fix the DM master-to-syncer binlog gap alert to ignore tasks that are not running.
```


dm/metrics/alertmanager/dm_worker_rules_test.go
```diff
@@ -0,0 +1,72 @@
...
+	} `yaml:"groups"`
+}
+
+func TestBinlogFileGapAlertRuleRequiresRunningTask(t *testing.T) {
+	rule := loadAlertRule(t, "dm_worker.rules.yml", binlogFileGapAlertName)
+
+	require.Equal(t, binlogFileGapAlertExpr, rule.Expr)
+	require.Equal(t, binlogFileGapAlertExpr, rule.Labels["expr"])
+	require.Equal(t, "10m", rule.For)
+}
```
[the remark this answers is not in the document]
expxiaoli: Fix it now.

D3Hunter (commented):
### Summary

- Total findings: 2
- Inline comments: 2

<details>

#### ⚠️ **[Major]** (1)

1. Test-only PromQL coverage changes repository-wide production dependencies (`dm/metrics/alertmanager/dm_worker_rules_test.go:21; go.mod:11; go.mod:56; go.mod:82; pkg/util/memory.go:34; cdc/sink/dmlsink/mq/dmlproducer/pulsar_dml_producer.go:99; cdc/sink/ddlsink/mq/ddlproducer/pulsar_ddl_producer.go:121; cdc/sink/dmlsink/txn/mysql/mysql.go:158`)

#### 🟡 **[Minor]** (1)

1. One-second wall-clock query deadline makes the rule test load-sensitive (`dm/metrics/alertmanager/dm_worker_rules_test.go:55`)

</details>

dm/metrics/alertmanager/dm_worker_rules_test.go
```diff
@@ -0,0 +1,99 @@
...
+// limitations under the License.
+
+package alertmanager
+
+import (
+	"os"
+	"testing"
+	"time"
+
+	"github.com/prometheus/prometheus/promql"
```
D3Hunter: #### ⚠️ **[Major]** Test-only PromQL coverage changes repository-wide production dependencies

**Why**
The new alert-rule test imports the full Prometheus PromQL engine from the root module, which also upgrades dependencies used by production TiFlow code outside DM. Go's module selection is repository-wide rather than test-only, so production binaries will compile against `automemlimit` v0.5.0 instead of v0.2.4 and `golang-lru` v0.6.0 instead of v0.5.1. Those releases change cgroup v1/hybrid/no-limit detection and move LRU eviction callbacks outside the cache lock, making a monitoring-only fix alter unrelated runtime behavior.

**Scope**
dm/metrics/alertmanager/dm_worker_rules_test.go:21; go.mod:11; go.mod:56; go.mod:82; pkg/util/memory.go:34; cdc/sink/dmlsink/mq/dmlproducer/pulsar_dml_producer.go:99; cdc/sink/ddlsink/mq/ddlproducer/pulsar_ddl_producer.go:121; cdc/sink/dmlsink/txn/mysql/mysql.go:158

**Risk if unchanged**
DM and TiCDC binaries can regress in memory-limit selection on cgroup v1 or hybrid hosts and in eviction or close concurrency for Pulsar producers, prepared statements, and other sink caches. These behavior changes would ship as part of a two-line alert fix even though the new validation exercises only PromQL, expanding the compatibility surface and making failures difficult to attribute.

**Evidence**
The only new Prometheus import is the test import on this line, while the root `go.mod` changes 27 requirement lines, adds roughly 20 other module entries, and `go.sum` changes 384 lines. Production code calls `memlimit.FromCgroup` from `pkg/util/memory.go`, whose result controls the redo applier memory cap, and several `cdc/sink` packages use the upgraded LRU library. No cgroup-mode, redo memory-limit, Pulsar eviction, prepared-statement eviction, or concurrency test is changed in this PR.

**Change request**
Please keep the production dependency graph unchanged by running the rule test through isolated Prometheus tooling, such as the existing promtool rule-test format or a dedicated test module. If the runtime dependency upgrades are intentional, split them into a dedicated change and add focused coverage for cgroup v1, v2, hybrid, and no-limit detection as well as the affected LRU eviction and sink-cache concurrency paths.

dm/metrics/alertmanager/dm_worker_rules_test.go
```diff
@@ -0,0 +1,99 @@
...
+func TestBinlogFileGapAlertRuleRequiresRunningTask(t *testing.T) {
+	rule := loadAlertRule(t, "dm_worker.rules.yml", binlogFileGapAlertName)
+
+	require.Equal(t, binlogFileGapAlertExpr, rule.Expr)
+	require.Equal(t, binlogFileGapAlertExpr, rule.Labels["expr"])
+	require.Equal(t, "10m", rule.For)
+
+	engine := promql.NewEngine(promql.EngineOpts{
+		MaxSamples:           10000,
+		Timeout:              time.Second,
```
D3Hunter: #### 🟡 **[Minor]** One-second wall-clock query deadline makes the rule test load-sensitive

**Why**
The tested alert expression has no timing semantics, but the engine is configured with a one-second real-time deadline. PromQL applies this setting through `context.WithTimeout` during query execution, so a constrained or race-enabled CI worker can fail the test because of host speed rather than an incorrect rule.

**Scope**
dm/metrics/alertmanager/dm_worker_rules_test.go:55

**Risk if unchanged**
The new regression test can fail intermittently under CPU contention, reducing confidence in the alert-rule signal and creating non-reproducible CI failures.

**Evidence**
`EngineOpts.Timeout` is set to `time.Second`; the upstream PromQL test setup in the selected v0.50.1 dependency uses 10-100 second safety timeouts for comparable engine-backed tests. The fixture itself is deterministic and does not require a one-second behavioral boundary.

**Change request**
Please use a generous safety timeout consistent with the upstream PromQL tests so this case fails on expression behavior, not machine scheduling speed.

dm/metrics/alertmanager/dm_worker_rules_test.go
D3Hunter, line 1: this is not the right place to write tests, and the test seems not that necessary

D3Hunter (commented):
rest lgtm

dm/metrics/alertmanager/dm_worker_rules_test.go
expxiaoli: I’ll remove PromQL evaluation and keep lightweight YAML assertions. I won’t add promtool tests because downloading promtool adds over a minute to test execution.

dm/metrics/alertmanager/dm_worker_rules_test.go
expxiaoli: The PromQL evaluation and one-second timeout will be removed, so this issue no longer applies.

dm/metrics/alertmanager/dm_worker_rules_test.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+type alertRulesFile struct {
+	Groups []struct {
+		Name  string      `yaml:"name"`
+		Rules []alertRule `yaml:"rules"`
+	} `yaml:"groups"`
+}
+
+func TestBinlogFileGapAlertRuleRequiresRunningTask(t *testing.T) {
```
D3Hunter: I don't think this test test anything except that the yaml contains xxx

expxiaoli: You're right. It only checks YAML fields, not real behavior. The new PR uses runtime tests instead.

D3Hunter: the test is still there

expxiaoli: You're right. I removed this test because it only checks YAML fields.

D3Hunter (approved):

GMHDBJD (approved):

PR #12767, opened 2026-07-27
syncer(dm): initialize binlog metrics from checkpoint

expxiaoli:

Issue Number: ref #12765

When a DM task is loaded on another worker but remains stopped or paused, the syncer is initialized and its metrics are registered, but the replication loop does not run. As a result, the syncer binlog file and position gauges keep their default values even when a valid persisted checkpoint exists.

The master binlog gauge continues to be updated by status polling, so the default syncer file number `0` can produce a significantly inflated master-to-syncer binlog gap.

This PR addresses the checkpoint metric initialization part of the issue.

- After loading the persisted checkpoint during syncer initialization, initialize the syncer binlog position gauge from the global checkpoint.
- Parse the checkpoint binlog filename and initialize the syncer binlog file gauge with its file index.
- Set the file gauge to `NaN` when the checkpoint has no valid binlog filename, such as a fresh or GTID-only checkpoint, instead of exposing the default value `0` as a valid file number.
- Add unit tests for file-position, missing-position, and GTID-only checkpoints.

#### Tests

- [x] Unit test

```console
GOPRIVATE=none GONOPROXY=none GONOSUMDB=none GOSUMDB=sum.golang.org GOPROXY=https://proxy.golang.org go test ./dm/syncer -run '^TestInitSyncerBinlogMetrics$' -count=1
git diff --check
```

No performance regression or compatibility break is expected. The gauges are initialized once during syncer initialization, and no replication or checkpoint behavior is changed. For checkpoints without a binlog filename, the file gauge now reports `NaN` until a valid file position is available instead of reporting a misleading zero.

No. Metric names and labels are unchanged; this PR only corrects their initialization values.

### Release note

```release-note
Fix DM syncer binlog metrics to initialize from the persisted checkpoint after a worker restart or rescheduling.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure binlog synchronization gauges are set as soon as synchronization tasks begin.
  * Keep binlog position metrics updated after checkpoint adjustments during synchronization.
  * When binlog filename parsing isn’t possible, the binlog file gauge now reports an unavailable value safely while preserving the binlog position.
* **Tests**
  * Expanded coverage for valid, empty, malformed, and GTID-only checkpoint formats.
  * Added lifecycle coverage for initializing and refreshing binlog metrics.

### DIFF
--- a/dm/syncer/syncer.go
+++ b/dm/syncer/syncer.go
@@ -545,9 +545,26 @@ func (s *Syncer) Init(ctx context.Context) (err error) {
 		metricProxies.Init(s.cfg.MetricsFactory)
 	}
 	s.metricsProxies = metricProxies.CacheForOneTask(s.cfg.Name, s.cfg.WorkerName, s.cfg.SourceID)
+	s.initSyncerBinlogMetrics(s.checkpoint.GlobalPoint())
 
 	s.ddlWorker = NewDDLWorker(&s.tctx.Logger, s)
 	return nil
+}
+
+func (s *Syncer) initSyncerBinlogMetrics(checkpoint binlog.Location) {
+	s.metricsProxies.Metrics.BinlogSyncerPosGauge.Set(float64(checkpoint.Position.Pos))
+
+	index, err := utils.GetFilenameIndex(checkpoint.Position.Name)
+	if err != nil {
+		// An empty binlog filename is expected for a fresh or GTID-only checkpoint.
+		// Use NaN so the unknown file number is not exposed as a valid zero value.
+		s.metricsProxies.Metrics.BinlogSyncerFileGauge.Set(math.NaN())
+		if checkpoint.Position.Name != "" {
+			s.tctx.L().Warn("fail to get index number of checkpoint binlog file", log.ShortError(err))
+		}
+		return
+	}
+	s.metricsProxies.Metrics.BinlogSyncerFileGauge.Set(float64(index))
 }
 
 // buildLowerCaseTableNamesMap build a lower case schema map and lower case table map for all tables
--- a/dm/syncer/syncer.go
+++ b/dm/syncer/syncer.go
@@ -1854,6 +1854,11 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 	if err != nil {
 		return err
 	}
+	// Init initializes metrics for tasks that do not enter Run. Refresh them
+	// here because Run may select a different checkpoint from start time or
+	// metadata, and GTID adjustment may further update the global checkpoint.
+	s.initSyncerBinlogMetrics(s.checkpoint.GlobalPoint())
+
 	if fresh && config.HasLoad(s.cfg.Mode) {
 		delLoadTask = true
 		flushCheckpoint = true
--- a/dm/syncer/syncer.go
+++ b/dm/syncer/syncer.go
@@ -545,13 +545,13 @@ func (s *Syncer) Init(ctx context.Context) (err error) {
 		metricProxies.Init(s.cfg.MetricsFactory)
 	}
 	s.metricsProxies = metricProxies.CacheForOneTask(s.cfg.Name, s.cfg.WorkerName, s.cfg.SourceID)
-	s.initSyncerBinlogMetrics(s.checkpoint.GlobalPoint())
+	s.updateSyncerBinlogMetrics(s.checkpoint.GlobalPoint())
 
 	s.ddlWorker = NewDDLWorker(&s.tctx.Logger, s)
 	return nil
 }
 
-func (s *Syncer) initSyncerBinlogMetrics(checkpoint binlog.Location) {
+func (s *Syncer) updateSyncerBinlogMetrics(checkpoint binlog.Location) {
 	s.metricsProxies.Metrics.BinlogSyncerPosGauge.Set(float64(checkpoint.Position.Pos))
 
 	index, err := utils.GetFilenameIndex(checkpoint.Position.Name)
@@ -1857,7 +1857,7 @@ func (s *Syncer) Run(ctx context.Context) (err error) {
 	// Init initializes metrics for tasks that do not enter Run. Refresh them
 	// here because Run may select a different checkpoint from start time or
 	// metadata, and GTID adjustment may further update the global checkpoint.
-	s.initSyncerBinlogMetrics(s.checkpoint.GlobalPoint())
+	s.updateSyncerBinlogMetrics(s.checkpoint.GlobalPoint())
 
 	if fresh && config.HasLoad(s.cfg.Mode) {
 		delLoadTask = true
--- a/dm/syncer/syncer_metrics_test.go
+++ b/dm/syncer/syncer_metrics_test.go
@@ -14,20 +14,27 @@
 package syncer
 
 import (
+	"context"
+	"errors"
 	"math"
 	"testing"
 
+	sqlmock "github.com/DATA-DOG/go-sqlmock"
 	"github.com/go-mysql-org/go-mysql/mysql"
+	"github.com/pingcap/tiflow/dm/config"
 	"github.com/pingcap/tiflow/dm/pkg/binlog"
+	"github.com/pingcap/tiflow/dm/pkg/conn"
 	tcontext "github.com/pingcap/tiflow/dm/pkg/context"
 	"github.com/pingcap/tiflow/dm/pkg/gtid"
+	"github.com/pingcap/tiflow/dm/pkg/schema"
 	"github.com/pingcap/tiflow/dm/syncer/metrics"
+	"github.com/pingcap/tiflow/engine/pkg/promutil"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/require"
 )
 
-func TestInitSyncerBinlogMetrics(t *testing.T) {
+func TestUpdateSyncerBinlogMetrics(t *testing.T) {
 	gtidSet, err := gtid.ParserGTID(
 		mysql.MySQLFlavor,
 		"3ccc475b-2343-11e7-be21-6c0b84d59f30:1-3",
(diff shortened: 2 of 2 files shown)


expxiaoli (commented):
​

expxiaoli:
> 🧹 Nitpick comments (1)
> > dm/syncer/syncer_metrics_test.go (1)> `42-65`: _🎯 Functional Correctness_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Add coverage for malformed non-empty binlog filenames.**
> > > The cases cover empty and GTID-only checkpoints, but not a checkpoint with a non-empty, unparsable filename. Add one to verify the file gauge becomes `NaN` while the position gauge preserves the checkpoint position.
> > > 
> > > Proposed test case
> > > ```diff
> > >  		{
> > >  			name:         "fresh checkpoint with empty binlog filename",
> > >  			checkpoint:   binlog.MustZeroLocation(mysql.MySQLFlavor),
> > >  			expectedFile: math.NaN(),
> > >  			expectedPos:  float64(binlog.MinPosition.Pos),
> > >  		},
> > > +		{
> > > +			name: "checkpoint with malformed binlog filename",
> > > +			checkpoint: binlog.NewLocation(mysql.Position{
> > > +				Name: "not-a-binlog-name",
> > > +				Pos:  123,
> > > +			}, nil),
> > > +			expectedFile: math.NaN(),
> > > +			expectedPos:  123,
> > > +		},
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > As per coding guidelines, DM fixes and features should include unit tests.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@dm/syncer/syncer_metrics_test.go` around lines 42 - 65, Add a test case
> > > alongside the existing checkpoint cases in the syncer metrics test table using a
> > > non-empty, malformed binlog filename and a known position. Assert that the file
> > > gauge result is math.NaN() while the position gauge retains the checkpoint’s
> > > position, reusing the existing checkpoint construction and expected-value
> > > conventions.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > _Source: Coding guidelines_
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Nitpick comments:
> In `@dm/syncer/syncer_metrics_test.go`:
> - Around line 42-65: Add a test case alongside the existing checkpoint cases in
> the syncer metrics test table using a non-empty, malformed binlog filename and a
> known position. Assert that the file gauge result is math.NaN() while the
> position gauge retains the checkpoint’s position, reusing the existing
> checkpoint construction and expected-value conventions.
> ```
> 
> ℹ️ Review info



> 🧹 Nitpick comments (1)
> > dm/syncer/syncer_metrics_test.go (1)> `42-65`: _🎯 Functional Correctness_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Add coverage for malformed non-empty binlog filenames.**
> > > The cases cover empty and GTID-only checkpoints, but not a checkpoint with a non-empty, unparsable filename. Add one to verify the file gauge becomes `NaN` while the position gauge preserves the checkpoint position.
> > > 
> > > Proposed test case
> > > ```diff
> > >  		{
> > >  			name:         "fresh checkpoint with empty binlog filename",
> > >  			checkpoint:   binlog.MustZeroLocation(mysql.MySQLFlavor),
> > >  			expectedFile: math.NaN(),
> > >  			expectedPos:  float64(binlog.MinPosition.Pos),
> > >  		},
> > > +		{
> > > +			name: "checkpoint with malformed binlog filename",
> > > +			checkpoint: binlog.NewLocation(mysql.Position{
> > > +				Name: "not-a-binlog-name",
> > > +				Pos:  123,
> > > +			}, nil),
> > > +			expectedFile: math.NaN(),
> > > +			expectedPos:  123,
> > > +		},
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > As per coding guidelines, DM fixes and features should include unit tests.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@dm/syncer/syncer_metrics_test.go` around lines 42 - 65, Add a test case
> > > alongside the existing checkpoint cases in the syncer metrics test table using a
> > > non-empty, malformed binlog filename and a known position. Assert that the file
> > > gauge result is math.NaN() while the position gauge retains the checkpoint’s
> > > position, reusing the existing checkpoint construction and expected-value
> > > conventions.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > _Source: Coding guidelines_
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Nitpick comments:
> In `@dm/syncer/syncer_metrics_test.go`:
> - Around line 42-65: Add a test case alongside the existing checkpoint cases in
> the syncer metrics test table using a non-empty, malformed binlog filename and a
> known position. Assert that the file gauge result is math.NaN() while the
> position gauge retains the checkpoint’s position, reusing the existing
> checkpoint construction and expected-value conventions.
> ```
> 
> ℹ️ Review info

Fix it

dm/syncer/syncer.go
```diff
@@ -545,11 +545,28 @@ func (s *Syncer) Init(ctx context.Context) (err error) {
 		metricProxies.Init(s.cfg.MetricsFactory)
 	}
 	s.metricsProxies = metricProxies.CacheForOneTask(s.cfg.Name, s.cfg.WorkerName, s.cfg.SourceID)
+	s.initSyncerBinlogMetrics(s.checkpoint.GlobalPoint())
 
 	s.ddlWorker = NewDDLWorker(&s.tctx.Logger, s)
 	return nil
 }
 
+func (s *Syncer) initSyncerBinlogMetrics(checkpoint binlog.Location) {
```
D3Hunter: #### 🟡 **[Minor]** Initializer name contradicts the helper's refresh behavior

**Why**
The helper is not limited to initialization: `Run` deliberately calls it again after the effective global checkpoint may change. Naming this repeated gauge update `initSyncerBinlogMetrics` obscures that it is safe and intended to overwrite existing metric values.

**Scope**
dm/syncer/syncer.go:554

**Risk if unchanged**
Future callers may treat the helper as a one-time setup operation, miss required refreshes after checkpoint changes, or add initialization-only work that is unsafe on the second call.

**Evidence**
The new comment at `dm/syncer/syncer.go:1857` explicitly says to "Refresh" the metrics, but line 1860 invokes `initSyncerBinlogMetrics`; the same method is first called from `Init` at line 548, and the test name `TestInitSyncerBinlogMetrics` reinforces the one-time interpretation.

**Change request**
Prefer `setSyncerBinlogMetrics` or `updateSyncerBinlogMetrics`; the current name is confusing. Rename the helper and its test so both initialization and later refresh call sites describe the same repeated update semantics.

dm/syncer/syncer_metrics_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/pingcap/tiflow/dm/pkg/binlog"
+	tcontext "github.com/pingcap/tiflow/dm/pkg/context"
+	"github.com/pingcap/tiflow/dm/pkg/gtid"
+	"github.com/pingcap/tiflow/dm/syncer/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInitSyncerBinlogMetrics(t *testing.T) {
```
D3Hunter: #### 🟡 **[Minor]** Checkpoint-to-metric lifecycle wiring is not covered

**Why**
The changed behavior depends on publishing the loaded checkpoint in `Init` and then republishing the final start checkpoint after metadata, start-time, and GTID adjustments in `Run`, but the new test validates only the conversion helper in isolation.

**Scope**
dm/syncer/syncer_metrics_test.go:30; dm/syncer/syncer.go:548; dm/syncer/syncer.go:1860

**Risk if unchanged**
A missing or misplaced lifecycle call can leave an idle or newly started task reporting zero, NaN, or a stale persisted checkpoint until its first binlog event, while this regression test still passes and the intended observability fix is lost.

**Evidence**
`TestInitSyncerBinlogMetrics` constructs standalone gauges and invokes `s.initSyncerBinlogMetrics(tc.checkpoint)` directly; it never exercises `Syncer.Init`, `LoadMeta`, `setGlobalPointByTime`, `adjustGlobalPointGTID`, or either production call site.

**Change request**
Please add an integration test for this path that asserts the exposed gauges after `Init` loads a persisted checkpoint and after `Run` selects a different metadata, start-time, or GTID-adjusted checkpoint before any binlog event is consumed.

D3Hunter (commented):
### Summary

- Total findings: 2
- Inline comments: 2

<details>

#### 🟡 **[Minor]** (2)

1. Initializer name contradicts the helper's refresh behavior (`dm/syncer/syncer.go:554`)
2. Checkpoint-to-metric lifecycle wiring is not covered (`dm/syncer/syncer_metrics_test.go:30; dm/syncer/syncer.go:548; dm/syncer/syncer.go:1860`)

</details>

D3Hunter (commented):
### Additional review finding

#### 🟠 **[Major] The original alerting issue is only partially fixed**

The production change correctly restores the syncer file and position gauges from the persisted checkpoint. It changes the reported example from `347650 - 0 = 347650` to the real file gap, `347650 - 346652 = 998`.

However, `dm/metrics/alertmanager/dm_worker.rules.yml:128` still evaluates every exported syncer metric without checking task state. Because `998 > 1`, an intentionally paused task can still trigger `DM_binlog_file_gap_between_master_syncer` after this PR. This fixes the misleading zero-based inflation, but it does not satisfy issue #12765's first expected behavior: paused or stopped tasks should not trigger the gap alert.

Please treat this as a partial fix and keep #12765 open until the alert is restricted to running tasks, for example by filtering or joining with `dm_worker_task_state == 2`. The join labels should preserve task and source identity so multi-source tasks do not produce ambiguous matches.

The checkpoint-to-metric lifecycle test gap is already covered by the existing inline comment at `dm/syncer/syncer_metrics_test.go:30`, so I did not duplicate it.

D3Hunter (commented):
rest lgtm

dm/syncer/syncer.go
expxiaoli: Make sense, I will fix it.

dm/syncer/syncer_metrics_test.go
expxiaoli: Make sense, I will fix it.

expxiaoli:
> ### Additional review finding
> #### 🟠 **[Major] The original alerting issue is only partially fixed**
> The production change correctly restores the syncer file and position gauges from the persisted checkpoint. It changes the reported example from `347650 - 0 = 347650` to the real file gap, `347650 - 346652 = 998`.
> 
> However, `dm/metrics/alertmanager/dm_worker.rules.yml:128` still evaluates every exported syncer metric without checking task state. Because `998 > 1`, an intentionally paused task can still trigger `DM_binlog_file_gap_between_master_syncer` after this PR. This fixes the misleading zero-based inflation, but it does not satisfy issue #12765's first expected behavior: paused or stopped tasks should not trigger the gap alert.
> 
> Please treat this as a partial fix and keep #12765 open until the alert is restricted to running tasks, for example by filtering or joining with `dm_worker_task_state == 2`. The join labels should preserve task and source identity so multi-source tasks do not produce ambiguous matches.
> 
> The checkpoint-to-metric lifecycle test gap is already covered by the existing inline comment at `dm/syncer/syncer_metrics_test.go:30`, so I did not duplicate it.



> ### Additional review finding
> #### 🟠 **[Major] The original alerting issue is only partially fixed**
> The production change correctly restores the syncer file and position gauges from the persisted checkpoint. It changes the reported example from `347650 - 0 = 347650` to the real file gap, `347650 - 346652 = 998`.
> 
> However, `dm/metrics/alertmanager/dm_worker.rules.yml:128` still evaluates every exported syncer metric without checking task state. Because `998 > 1`, an intentionally paused task can still trigger `DM_binlog_file_gap_between_master_syncer` after this PR. This fixes the misleading zero-based inflation, but it does not satisfy issue #12765's first expected behavior: paused or stopped tasks should not trigger the gap alert.
> 
> Please treat this as a partial fix and keep #12765 open until the alert is restricted to running tasks, for example by filtering or joining with `dm_worker_task_state == 2`. The join labels should preserve task and source identity so multi-source tasks do not produce ambiguous matches.
> 
> The checkpoint-to-metric lifecycle test gap is already covered by the existing inline comment at `dm/syncer/syncer_metrics_test.go:30`, so I did not duplicate it.

filtering `dm_worker_task_state == 2` fix is covered by https://github.com/pingcap/tiflow/pull/12766

D3Hunter (approved):

GMHDBJD (approved):

PR #12782, opened 2026-07-31
openapi(dm): support task timezone

OliverS929:

Issue Number: ref #12756

DM task configuration supports `timezone`, but this option is not available
through OpenAPI.

- Add an optional `timezone` field to the OpenAPI task schema.
- Validate it with DM's existing time-zone parser.
- Preserve it when converting between OpenAPI tasks and subtask configurations.
- Keep the existing downstream database time zone behavior when the field is
  omitted or empty.

The DST fall-back repeated-hour behavior observed during testing is an existing
DM runtime limitation, not introduced by OpenAPI. This PR only aligns the
OpenAPI capability, and changing the runtime behavior is out of scope.

#### Tests 

 - Unit test

Manual testing with DM, MySQL 8.4, and TiDB covered valid, invalid, and omitted
time zones, create and readback, update rejection, downstream-default behavior,
and full and incremental replication.

No. The field is optional, and omitted or empty values preserve the existing
behavior.

No additional documentation update is needed.

### Release note 

```release-note
Support configuring the DM task time zone through OpenAPI.
```

## Summary by CodeRabbit

* **New Features**
  * Tasks now support an optional timezone field that is validated upon submission. Timezone is immutable after task creation and defaults to database configuration when omitted.

* **Documentation**
  * Updated API specification to document the timezone property and its fallback behavior.


OliverS929:
@joechenrh PTAL.

dm/config/task_converters_test.go
```diff
@@ -464,11 +471,34 @@ func TestConvertBetweenOpenAPITaskAndTaskConfig(t *testing.T) {
...
+	for _, tc := range []struct {
+		name     string
+		timezone *string
+	}{
+		{name: "omitted", timezone: nil},
+		{name: "empty", timezone: &emptyTimezone},
+	} {
+		t.Run(tc.name+" timezone uses downstream default", func(t *testing.T) {
+			task.Timezone = tc.timezone
+			taskCfgWithoutTimezone, convertErr := OpenAPITaskToTaskConfig(&task, sourceCfgMap)
```
lance6716, line 491: why the timezone is only propagated for subtask in `OpenAPITaskToSubTaskConfig`, not task, as the test shows?

lance6716 (commented):
rest lgtm

dm/config/task_converters_test.go
OliverS929: It is eventually propagated to `TaskConfig`. `OpenAPITaskToTaskConfig` uses `SubTaskConfig` as an intermediate representation, and the parameter values are then propagated into `TaskConfig` by `SubTaskConfigsToTaskConfig`. This is the [existing conversion flow](https://github.com/pingcap/tiflow/pull/12782/changes#diff-2faa7688fd40433fd1f30b6fd008f08dd81f274826a3d71c6556f27a6b3188e3L751-L763) and also applies to other task-level fields such as `name` and `task_mode`. This PR adds timezone support without changing that flow.

joechenrh (approved):

PR #12790, opened 2026-08-04
conn(dm): cap pooled DB connection lifetime and idle time

### LINKED ISSUE #12789: dm: pooled DB connections live forever, causing spurious `invalid connection` unit exits
## Problem

`dm/pkg/conn/basedb.go#Apply` sets only `db.SetMaxIdleConns`. Neither
`db.SetConnMaxLifetime` nor `db.SetConnMaxIdleTime` are ever called,
so a pooled `*sql.DB` connection to the upstream MySQL or downstream
TiDB/MySQL lives for as long as the DM worker process does.

When the peer server silently drops such a connection -- rolling
restart, `wait_timeout` expiry, or a stateful middlebox timing out the
flow -- the next statement re-using it surfaces as `mysql.ErrInvalidConn`.

`dm/pkg/retry/errors.go#IsUnretryableConnectionError` classifies
`ErrInvalidConn` as unretryable (because it can't tell whether a
preceding write reached the downstream), so the syncer / loader /
dumper unit exits with an error and the corresponding
`dm_{syncer,loader,mydumper}_exit_with_error_count{resumable_err=\"true\"}`
increments. The task-checker auto-resumes seconds later, which makes
each individual failure look spurious, but with many tasks the aggregate
turns into a steady stream of DM "resumable error" alerts even when
nothing is wrong.

## Reproduction

1. Start a DM task against a MySQL / TiDB downstream that has
   `wait_timeout` set (default 28800s / 8h is enough).
2. Pause the workload for longer than `wait_timeout`, or restart the
   downstream, so that at least one pooled connection is dropped
   server-side without the client noticing.
3. Resume the workload. On the first statement using the stale
   connection the syncer logs
   `[error=... RawCause: invalid connection]` and the unit exits;
   `dm_syncer_exit_with_error_count{resumable_err="true"}` increments.

## Suggested fix

Set conservative defaults for `SetConnMaxLifetime` and
`SetConnMaxIdleTime` in `Apply`, well below the MySQL/TiDB default
`wait_timeout` (28800s) and any typical middlebox timeout, so the pool
never hands out a long-dead connection. Also expose the two values on
`RawDBConfig` so callers that build a `DBConfig` programmatically can
override them.

I'll open a PR shortly.

andreifedorov-bolt:

Issue Number: close #12789

`dm/pkg/conn/basedb.go#Apply` only calls `db.SetMaxIdleConns`; neither
`db.SetConnMaxLifetime` nor `db.SetConnMaxIdleTime` is ever set, so a
pooled `*sql.DB` connection to the upstream MySQL or the downstream
TiDB / MySQL lives for as long as the DM worker process.

When the peer server silently drops such a connection (rolling restart,
`wait_timeout` expiry, or a stateful middlebox timing out the flow), the
next statement re-using it surfaces as `mysql.ErrInvalidConn`, which
`dm/pkg/retry/errors.go#IsUnretryableConnectionError` classifies as
unretryable. The syncer / loader / dumper unit therefore exits with an
error and `dm_{syncer,loader,mydumper}_exit_with_error_count{resumable_err="true"}`
increments. Task-checker auto-resumes seconds later, so each individual
failure looks spurious, but with many concurrent tasks the aggregate
becomes a steady stream of DM "resumable error" alerts.

Apply two conservative defaults in `Apply` so the pool proactively
closes and re-opens a connection before its peer can drop it:

* `SetConnMaxLifetime(30m)` — well below the MySQL/TiDB default
  `wait_timeout` of 28800 s (8 h) and any typical middlebox idle timeout.
  At the default `MaxIdleConns` this is roughly two reconnects per minute
  per pooled slot — negligible next to steady-state syncer traffic.
* `SetConnMaxIdleTime(5m)` — retires connections that are only used
  intermittently before the wall clock does.

Both defaults are also exposed as `ConnMaxLifetime` / `ConnMaxIdleTime`
on `RawDBConfig` (with matching `Set…` methods for symmetry), so callers
that build a `DBConfig` programmatically can raise or lower them without
touching package-level defaults. A zero (or negative) `RawDBCfg` value
keeps the package default, so this is fully backwards-compatible with
every existing caller.

#### Tests

 - No code — behaviour of `database/sql` connection lifetime is
   documented and its handlers are simple setters; there is no additional
   observable behaviour beyond what `database/sql` already tests. The
   existing `dm/pkg/conn` test suite (including `TestGetBaseConn`,
   `TestFailDBPing`, `TestGetBaseConnWontBlock`) passes unchanged.

No. Steady-state churn is on the order of `MaxIdleConns / 30 min` — a
handful of extra `sql.Open`s per minute per worker per unit — which is
negligible against the normal statement rate. Every caller keeps its
current behaviour unless it explicitly overrides `RawDBConfig`.

No.

### Release note

```release-note
dm: cap pooled DB connection lifetime and idle time so a peer server or middlebox silently dropping an idle TCP connection no longer surfaces as a non-retryable "invalid connection" unit exit.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable database connection lifetime and idle-time limits.
  * Introduced default limits to proactively refresh database connections.

* **Bug Fixes**
  * Reduced stale connection errors and improved database connection reliability during retries.

### DIFF
--- a/dm/config/dbconfig/config.go
+++ b/dm/config/dbconfig/config.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"strings"
+	"time"
 
 	"github.com/BurntSushi/toml"
 	"github.com/pingcap/tiflow/dm/config/security"
@@ -49,6 +50,14 @@ type RawDBConfig struct {
 	MaxIdleConns int
 	ReadTimeout  string
 	WriteTimeout string
+	// ConnMaxLifetime caps the total lifetime of a pooled connection.
+	// A zero (or negative) value means the process-wide default is used
+	// (see basedb.Apply). Set explicitly to override.
+	ConnMaxLifetime time.Duration
+	// ConnMaxIdleTime caps how long a connection may sit idle in the pool
+	// before being closed. A zero (or negative) value means the process-wide
+	// default is used (see basedb.Apply). Set explicitly to override.
+	ConnMaxIdleTime time.Duration
 }
 
 // SetReadTimeout set readTimeout for raw database config.
@@ -68,6 +77,18 @@ func (c *RawDBConfig) SetWriteTimeout(writeTimeout string) *RawDBConfig {
 // set value > 0 then `value` idle connections are retained.
 func (c *RawDBConfig) SetMaxIdleConns(value int) *RawDBConfig {
 	c.MaxIdleConns = value
+	return c
+}
+
+// SetConnMaxLifetime sets the maximum lifetime of a pooled connection.
+func (c *RawDBConfig) SetConnMaxLifetime(d time.Duration) *RawDBConfig {
+	c.ConnMaxLifetime = d
+	return c
+}
+
+// SetConnMaxIdleTime sets the maximum idle time of a pooled connection.
+func (c *RawDBConfig) SetConnMaxIdleTime(d time.Duration) *RawDBConfig {
+	c.ConnMaxIdleTime = d
 	return c
 }
 
--- a/dm/pkg/conn/basedb.go
+++ b/dm/pkg/conn/basedb.go
@@ -44,6 +44,20 @@ var customID int64
 
 var netTimeout = DefaultDBTimeout
 
+const (
+	// defaultConnMaxLifetime is the default maximum lifetime of a pooled
+	// database/sql connection. It is set well below any reasonable server-side
+	// `wait_timeout` (MySQL/TiDB default is 8h) and any typical stateful
+	// middlebox idle timeout, so the pool never hands out a connection whose
+	// peer has silently gone away (rolling restart, `wait_timeout` expiry,
+	// keepalive miss). See https://pkg.go.dev/database/sql#DB.SetConnMaxLifetime.
+	defaultConnMaxLifetime = 30 * time.Minute
+	// defaultConnMaxIdleTime bounds how long a connection may sit idle in the
+	// pool. It complements defaultConnMaxLifetime for connections that are
+	// used rarely. See https://pkg.go.dev/database/sql#DB.SetConnMaxIdleTime.
+	defaultConnMaxIdleTime = 5 * time.Minute
+)
+
 // DBProvider providers BaseDB instance.
 type DBProvider interface {
 	Apply(config ScopedDBConfig) (*BaseDB, error)
@@ -148,6 +162,8 @@ func (d *DefaultDBProviderImpl) ApplyWithPingTimeout(config ScopedDBConfig, ping
 	}
 
 	var maxIdleConns int
+	connMaxLifetime := defaultConnMaxLifetime
+	connMaxIdleTime := defaultConnMaxIdleTime
 	rawCfg := config.RawDBCfg
 	if rawCfg != nil {
 		if rawCfg.ReadTimeout != "" {
@@ -157,6 +173,12 @@ func (d *DefaultDBProviderImpl) ApplyWithPingTimeout(config ScopedDBConfig, ping
 			dsn += fmt.Sprintf("&writeTimeout=%s", rawCfg.WriteTimeout)
 		}
 		maxIdleConns = rawCfg.MaxIdleConns
+		if rawCfg.ConnMaxLifetime > 0 {
+			connMaxLifetime = rawCfg.ConnMaxLifetime
+		}
+		if rawCfg.ConnMaxIdleTime > 0 {
+			connMaxIdleTime = rawCfg.ConnMaxIdleTime
+		}
 	}
 
 	var setFK bool
@@ -200,6 +222,14 @@ func (d *DefaultDBProviderImpl) ApplyWithPingTimeout(config ScopedDBConfig, ping
 	}
 
 	db.SetMaxIdleConns(maxIdleConns)
+	// Cap connection lifetime so the pool proactively closes and re-opens
+	// TCP connections before the server or any middlebox drops them.
+	// Without this, a pooled connection re-used after the peer has silently
+	// gone away surfaces as `mysql.ErrInvalidConn` on the next statement,
+	// which the DM retry classifier treats as unretryable and causes the
+	// syncer/loader/dumper unit to exit with an error.
+	db.SetConnMaxLifetime(connMaxLifetime)
+	db.SetConnMaxIdleTime(connMaxIdleTime)
 
 	return NewBaseDB(db, config.Scope, version, doFuncInClose), nil
 }


andreifedorov-bolt:
/label needs-cherry-pick-release-8.5

Requesting a backport to `release-8.5` (aimed at v8.5.6+) since the failure mode this fixes exists in that line as well. Happy to open the manual cherry-pick PR if the bot is not enabled for external contributors.

dm/config/dbconfig/config.go
```diff
@@ -71,6 +80,18 @@ func (c *RawDBConfig) SetMaxIdleConns(value int) *RawDBConfig {
 	return c
 }
 
+// SetConnMaxLifetime sets the maximum lifetime of a pooled connection.
+func (c *RawDBConfig) SetConnMaxLifetime(d time.Duration) *RawDBConfig {
```
dveeden, line 84: Do we really need `SetConnMaxLifetime()` if `ConnMaxLifetime` isn't private?

andreifedorov-bolt: Kept for symmetry with the existing `SetReadTimeout` / `SetWriteTimeout` / `SetMaxIdleConns` setters on the same struct — DM's `RawDBConfig` is used almost exclusively via the chained fluent-builder pattern, e.g.

```go
dbCfg.RawDBCfg = dbconfig.DefaultRawDBConfig().
    SetReadTimeout(maxDMLConnectionTimeout).
    SetMaxIdleConns(s.cfg.WorkerCount)
```

14 non-test call sites use it that way (e.g. `dm/syncer/syncer.go:3173,3216,3226`, `dm/syncer/data_validator.go:287,295,1326,1355,1395`, `dm/syncer/checkpoint.go:484`, `dm/syncer/sharding_group.go:457`, `dm/syncer/online-ddl-tools/online_ddl.go:136`, `dm/checker/checker.go:587,599`, `dm/config/source_config.go:280`). Only one non-test caller assigns the field directly (`dm/relay/relay.go:1099`) and it does so on an already-built config, not in a builder chain.

Without `SetConnMaxLifetime` / `SetConnMaxIdleTime` a caller wanting to override the new fields in a chain would have to break out to a temporary and mutate the public field, which reads worse and diverges from the convention. Happy to drop them if you'd prefer to move the codebase away from the fluent pattern going forward — the fields are public either way, so it's purely stylistic.

dveeden (approved):

D3Hunter (commented):
### Summary

- Total findings: 5
- Inline comments: 5

<details>

#### ⚠️ **[Major]** (4)

1. Connection-lifetime comments promise a guarantee the pool cannot provide (`dm/pkg/conn/basedb.go:48 and dm/pkg/conn/basedb.go:225`)
2. Pool limits do not retire DM's long-lived worker connections (`dm/pkg/conn/basedb.go:231`)
3. New pool policy lacks testable deterministic coverage (`dm/pkg/conn/basedb.go:165 and dm/pkg/conn/basedb.go:231; dm/config/dbconfig/config.go:53`)
4. Connection-lifetime overrides are not available through DM configuration (`dm/config/dbconfig/config.go:53 and dm/config/dbconfig/config.go:108; applied in dm/pkg/conn/basedb.go:165`)

#### 🟡 **[Minor]** (1)

1. Exported setters hide their non-positive-value semantics (`dm/config/dbconfig/config.go:53 and dm/config/dbconfig/config.go:83`)

</details>

dm/pkg/conn/basedb.go
```diff
@@ -44,6 +44,20 @@ var customID int64
 
 var netTimeout = DefaultDBTimeout
 
+const (
+	// defaultConnMaxLifetime is the default maximum lifetime of a pooled
+	// database/sql connection. It is set well below any reasonable server-side
+	// `wait_timeout` (MySQL/TiDB default is 8h) and any typical stateful
+	// middlebox idle timeout, so the pool never hands out a connection whose
```
D3Hunter, line 51: #### ⚠️ **[Major]** Connection-lifetime comments promise a guarantee the pool cannot provide

**Why**
The new rationale says the pool closes and reopens connections before any server or middlebox drops them and therefore never returns a connection whose peer has disappeared. The configured limits only retire connections after a fixed age or idle interval; they cannot predict a rolling restart, keepalive failure, or a server or middlebox timeout shorter than those limits, and database/sql opens replacements on demand rather than as part of these setters.

**Scope**
dm/pkg/conn/basedb.go:48 and dm/pkg/conn/basedb.go:225

**Risk if unchanged**
Maintainers can treat the invalid-connection failure mode as eliminated and rely on these defaults in deployments where peers disappear sooner, obscuring the remaining operational risk and making later tuning or incident diagnosis misleading.

**Evidence**
The implementation always uses the hard-coded 30-minute lifetime and 5-minute idle defaults unless a positive override is supplied; it does not read the server wait_timeout, discover a middlebox timeout, or detect restarts. The standard-library contract for both setters states that expired connections may be closed lazily before reuse, which is weaker than the comments' before and never claims.

**Change request**
Please rewrite both comments to say that the limits reduce the chance of reusing stale connections by retiring sufficiently old or idle entries, remove the universal before/never guarantee and the claim that the setters reopen TCP connections, and state that shorter external timeouts or abrupt peer loss can still produce ErrInvalidConn.

dm/pkg/conn/basedb.go
```diff
@@ -200,6 +222,14 @@ func (d *DefaultDBProviderImpl) ApplyWithPingTimeout(config ScopedDBConfig, ping
 	}
 
 	db.SetMaxIdleConns(maxIdleConns)
+	// Cap connection lifetime so the pool proactively closes and re-opens
+	// TCP connections before the server or any middlebox drops them.
+	// Without this, a pooled connection re-used after the peer has silently
+	// gone away surfaces as `mysql.ErrInvalidConn` on the next statement,
+	// which the DM retry classifier treats as unretryable and causes the
+	// syncer/loader/dumper unit to exit with an error.
+	db.SetConnMaxLifetime(connMaxLifetime)
```
D3Hunter, line 231: #### ⚠️ **[Major]** Pool limits do not retire DM's long-lived worker connections

**Why**
SetConnMaxLifetime and SetConnMaxIdleTime are enforced by database/sql when pooled connections are idle or checked out again, but DM reserves a *sql.Conn in each BaseConn and keeps it for the lifetime of the syncer or loader worker. Those connections never become idle in the pool and are not checked out again, so neither new limit rotates the connections that the change is intended to protect.

**Scope**
dm/pkg/conn/basedb.go:231

**Risk if unchanged**
After a task has run longer than the configured lifetime, a server restart, wait_timeout, or middlebox drop can still make the next DML on a reserved worker connection return mysql.ErrInvalidConn; DM will continue to classify the ambiguous write as unretryable and exit the unit, leaving the reported production failure mode in place.

**Evidence**
BaseDB.GetBaseConn obtains d.DB.Conn and stores the resulting fixed connection in BaseConn (dm/pkg/conn/basedb.go:304-319); CreateConns creates these once and retains them in each DBConn (dm/syncer/dbconn/db.go:333-346). The standard pool cleaner only scans idle freeConn entries, while a reserved sql.Conn owns its driver connection until Conn.Close, so the settings added at lines 231-232 cannot expire these worker connections.

**Change request**
Can we rotate long-lived BaseConn instances at a safe transaction boundary (or stop pinning them and acquire through sql.DB) and add a case that keeps a worker connection past the limit and verifies it is replaced before the next statement?

dm/pkg/conn/basedb.go
```diff
@@ -148,6 +162,8 @@ func (d *DefaultDBProviderImpl) ApplyWithPingTimeout(config ScopedDBConfig, ping
 	}
 
 	var maxIdleConns int
+	connMaxLifetime := defaultConnMaxLifetime
```
D3Hunter, line 165: #### ⚠️ **[Major]** New pool policy lacks testable deterministic coverage

**Why**
The change adds default, non-positive fallback, and positive-override policy directly inside ApplyWithPingTimeout, a method that must open and ping a real MySQL connection, and also changes the lifecycle of every database pool. It adds neither a focused test seam nor unit or integration coverage proving that the selected limits are applied and that an expired connection is retired before reuse.

**Scope**
dm/pkg/conn/basedb.go:165 and dm/pkg/conn/basedb.go:231; dm/config/dbconfig/config.go:53

**Risk if unchanged**
A later refactor can silently alter nil-config handling, invert zero or negative fallback semantics, ignore positive overrides, or stop applying the limits across syncer, loader, dumper, checker, and relay. The stale-connection production failure could then return without deterministic regression coverage catching it.

**Evidence**
Lines 165-180 select two defaults and conditionally replace them for positive RawDBConfig durations, while lines 231-232 apply the selected values. The diff changes only config.go and basedb.go and adds no *_test.go, integration, upgrade, or failure-path coverage for default selection, override selection, or stale-connection replacement.

**Change request**
Can we extract effective connection-pool option selection into a private pure helper and add table-driven unit cases for nil RawDBCfg, zero or negative values, and positive overrides? Please also add a deterministic lifecycle test using a controllable driver or server timeout and condition-based synchronization rather than an arbitrary time.Sleep, so it verifies that an expired pooled connection is replaced before the next statement.

dm/config/dbconfig/config.go
```diff
@@ -49,6 +50,14 @@ type RawDBConfig struct {
 	MaxIdleConns int
 	ReadTimeout  string
 	WriteTimeout string
+	// ConnMaxLifetime caps the total lifetime of a pooled connection.
```
D3Hunter, line 53: #### ⚠️ **[Major]** Connection-lifetime overrides are not available through DM configuration

**Why**
The new defaults are intended to stay below the database or middlebox timeout, but the only overrides live in RawDBConfig, while DBConfig.RawDBCfg is excluded from TOML, JSON, and YAML and SourceConfig.GenerateDBConfig replaces it with an internally constructed value. A DM operator therefore cannot adapt the 30-minute lifetime or 5-minute idle limit to an environment whose wait_timeout or network idle timeout is shorter.

**Scope**
dm/config/dbconfig/config.go:53 and dm/config/dbconfig/config.go:108; applied in dm/pkg/conn/basedb.go:165

**Risk if unchanged**
After upgrade, deployments with a timeout below these fixed defaults can continue receiving the same stale pooled connection and unretryable mysql.ErrInvalidConn that this change is meant to prevent, with no supported configuration-based mitigation.

**Evidence**
ApplyWithPingTimeout only accepts positive overrides from config.RawDBCfg, but the enclosing RawDBCfg field has toml:"-" json:"-" yaml:"-" tags. The normal source path at dm/config/source_config.go:280 always creates DefaultRawDBConfig() and does not carry user-supplied lifetime values.

**Change request**
Can we expose backward-compatible optional lifetime and idle-time settings in the supported source and downstream configuration path, carry them into RawDBConfig, validate their units and bounds, and document how operators should choose values relative to server and middlebox timeouts?

dm/config/dbconfig/config.go
```diff
@@ -71,6 +80,18 @@ func (c *RawDBConfig) SetMaxIdleConns(value int) *RawDBConfig {
 	return c
 }
 
+// SetConnMaxLifetime sets the maximum lifetime of a pooled connection.
```
D3Hunter, line 83: #### 🟡 **[Minor]** Exported setters hide their non-positive-value semantics

**Why**
SetConnMaxLifetime and SetConnMaxIdleTime use names and one-line docs that closely mirror database/sql, where a non-positive duration disables the corresponding limit, but ApplyWithPingTimeout interprets the same values as a request for DM's positive default. The field comments mention this behavior, but callers using the exported fluent setters should not have to inspect a separate field declaration to discover that the familiar sentinel means something different here.

**Scope**
dm/config/dbconfig/config.go:53 and dm/config/dbconfig/config.go:83

**Risk if unchanged**
A programmatic caller can pass zero intending to disable connection retirement and silently receive the 30-minute or 5-minute default instead; the phrase process-wide default also overstates the scope because the fallback is local to the default DB provider path.

**Evidence**
Both setters assign d without validation or explanation, while DefaultDBProviderImpl.ApplyWithPingTimeout accepts an override only when the stored value is greater than zero. This differs from the standard-library methods whose names these APIs reuse.

**Change request**
Please document on both exported setters that only positive durations override the DM provider defaults and that zero or negative durations do not disable the limits; also replace process-wide default/see basedb.Apply with the precise provider behavior, or expose an explicit way to request the disabled state if that is intended to be supported.

dm/pkg/conn/basedb.go
D3Hunter, line 231: These pool settings do not rotate DM’s long-lived worker connections, so the primary syncer failure remains.

`SetConnMaxLifetime` and `SetConnMaxIdleTime` are enforced when a `database/sql` connection is idle in the pool or is checked out/returned. However, `BaseDB.GetBaseConn` calls `DB.Conn`, and `dm/syncer/dbconn.CreateConns` retains the resulting `*sql.Conn` for the unit lifetime. Operations on an already checked-out `*sql.Conn` reuse its owned driver connection without evaluating the pool lifetime or idle-time limits.

Concrete failure: if a running syncer receives no events for longer than the downstream `wait_timeout`, its pinned DML/DDL/checkpoint connection remains checked out and is not retired by either new setting. The next `Begin`, `Exec`, or `Commit` can still encounter the server-closed socket and return `mysql.ErrInvalidConn`.

Please rotate or validate the pinned `BaseConn` handles at a safe boundary, or avoid task-lifetime `*sql.Conn` ownership. A regression test should cover the long-lived `BaseConn` path; testing ordinary `*sql.DB` pool reuse would not exercise the affected syncer behavior.

andreifedorov-bolt:
Thanks @D3Hunter for the thorough review — F2 is the load-bearing one and it's correct. I confirmed by re-reading BaseDB.GetBaseConn (d.DB.Conn(ctx) reserves a *sql.Conn) and dbconn.CreateConns (retains reserved conns for the unit's lifetime, only rotated by ResetBaseConnFn on error paths). database/sql's SetConnMaxLifetime / SetConnMaxIdleTime don't reach reserved connections, so this PR does not fix the syncer-exit failure mode I motivated it with.

A proper fix needs lifetime-based rotation of the pinned BaseConn at safe transaction boundaries (or moving away from pinning *sql.Conn in BaseConn), plus a driver-controllable test. That's a larger change and I'm not going to attempt it here.

On the operator side, the failure mode DM-9739 saw in production is now mitigated at the transport layer by shortening the host TCP keepalive on DM hosts (net.ipv4.tcp_keepalive_time=300 / _intvl=30 / _probes=4 → ~7 min detection ceiling, via a systemd oneshot in our ansible. Kernel probes are unaffected by whether the socket is pinned in *sql.Conn or in the pool, so this reaches the syncer's reserved connections regardless.

Closing this PR. Thanks again for the detailed review and for pointing out the Reserved *sql.Conn semantics — very useful reading.

dveeden:
> On the operator side, the failure mode DM-9739 saw in production is now mitigated at the transport layer by shortening the host TCP keepalive on DM hosts (net.ipv4.tcp_keepalive_time=300 / _intvl=30 / _probes=4 → ~7 min detection ceiling, via a systemd oneshot in our ansible. Kernel probes are unaffected by whether the socket is pinned in *sql.Conn or in the pool, so this reaches the syncer's reserved connections regardless.

Should this be added to the documentation of DM?

PR #12792, opened 2026-08-06
openapi(dm): support target session parameters

OliverS929:

Issue Number: ref #12756

DM task configuration supports `foreign_key_checks` for target database sessions, but OpenAPI does not expose this option.

- Add an optional `session` field to the OpenAPI target database configuration.
- Initially support only `foreign_key_checks`, with case-insensitive keys and an exact string value of `"0"` or `"1"`.
- Preserve the setting through task configuration conversions and apply it to incremental replication.
- Reject changing `foreign_key_checks` through PUT because the syncer session must be reinitialized.

Full import continues to use `foreign_key_checks=0`. This existing behavior is not changed by this PR.

#### Tests 

 - Unit test

Manual testing with DM, MySQL 8.4, and TiDB nightly covered validation and readback, both `foreign_key_checks` values during incremental replication, PUT rejection, and `all` and `full` replication.

No. The field is optional, and omitting it preserves the existing `foreign_key_checks=0` behavior.

No additional documentation update is needed. The field, supported parameter, and accepted values are documented in the OpenAPI schema.

### Release note 

```release-note
Support configuring `foreign_key_checks` for DM task target database sessions through OpenAPI.
```

## Summary by CodeRabbit

- **New Features**
  - Added optional target database session configuration for `foreign_key_checks`.
  - Supports values `0` and `1`, with consistent normalization across task creation, retrieval, and subtasks.
  - OpenAPI documentation now describes the new session setting, defaults, and supported values.

- **Bug Fixes**
  - Prevented unsupported session properties from being accepted or exported.
  - Configuration updates that change `foreign_key_checks` are now rejected.
  - Ensured source settings generate the correct Lightning configuration behavior.


OliverS929:
@joechenrh PTAL.

dm/syncer/syncer.go
```diff
@@ -3451,6 +3454,28 @@ func (s *Syncer) CheckCanUpdateCfg(newCfg *config.SubTaskConfig) error {
...
+			if !strings.EqualFold(key, "foreign_key_checks") {
+				continue
+			}
+			if found {
+				return false
+			}
+			found = true
+		}
+		return true
+	}
```
joechenrh: Could we centralize the normalization of target session settings? Then we can simplify this function as well as `projectTaskTargetSession`.

dm/openapi/spec/dm.yaml
```diff
@@ -1662,6 +1662,15 @@ components:
         security:
           $ref: "#/components/schemas/Security"
           description: "downstram database ssl config"
+        session:
```
D3Hunter, line 1665: #### ⚠️ **[Major]** Session schema advertises keys and values that the API rejects

**Why**
The machine-readable OpenAPI contract models `target_config.session` as an unrestricted `map[string]string`, while the implementation accepts only the case-insensitive `foreign_key_checks` key and only the exact values `"0"` or `"1"`. The prose narrows the contract, but generated clients and schema validators still expose every other key and string as valid.

**Scope**
dm/openapi/spec/dm.yaml:1665

**Risk if unchanged**
API consumers can construct requests that pass generated-client or OpenAPI validation and then fail only at runtime with HTTP 400, and the generated Go surface remains the generic `TaskTargetDataBase_Session.AdditionalProperties` map instead of communicating the supported domain setting.

**Evidence**
The schema uses `additionalProperties: {type: string}` with no named property or value constraint, but `NormalizeTaskTargetSession` in `dm/openapi/task.go:56` rejects every normalized key except `foreign_key_checks` and every value except `"0"`/`"1"`.

**Change request**
Prefer an explicit canonical `foreign_key_checks` property with an enum for `"0"` and `"1"` (and reject other properties in the schema); document case-insensitive input compatibility separately if it must remain supported. The current generic map contract is confusing and does not match runtime validation.

D3Hunter, line 1665: #### ⚠️ **[Major]** The public contract does not say that the new session setting is immutable through PUT

**Why**
The same `Task` schema, including the new `session` field, is used by both create and update requests, so consumers have no contract-level signal that changing `foreign_key_checks` on an existing task is rejected.

**Scope**
dm/openapi/spec/dm.yaml:1665 and dm/syncer/syncer.go:3420

**Risk if unchanged**
Clients may build an update workflow around a field that looks writable and discover only in production that PUT returns HTTP 400; the current error text also refers vaguely to session `fields` rather than naming the immutable setting or the supported way to change it.

**Evidence**
`UpdateTaskRequest.task` references the shared `Task` schema at `dm/openapi/spec/dm.yaml:2331`, while the added guard rejects an effective `foreign_key_checks` change and the added API test at `dm/master/openapi_view_test.go:1128` establishes that PUT must fail. Neither the field description nor the update operation documents this lifecycle rule.

**Change request**
Please document when this rejection happens: mark `target_config.session.foreign_key_checks` as immutable after task creation in the field/update contract, state the supported procedure for changing it, and make the update error name `foreign_key_checks` explicitly.

D3Hunter, line 1665: #### ⚠️ **[Major]** N-1 masters silently discard the new target session field

**Why**
The unversioned `session` request field is accepted only by the new generated task type, while every API request is forwarded to whichever DM master is leader. An N-1 leader has neither this field nor a strict unknown-field rejection, so a create or update can return success after decoding the request without `foreign_key_checks`.

**Scope**
dm/openapi/spec/dm.yaml:1665; rolling-upgrade and rollback paths for POST/PUT /api/v1/tasks

**Risk if unchanged**
During a rolling upgrade, a task submitted with `foreign_key_checks=1` can actually be created with the default disabled behavior. After rollback, an old leader can also omit the setting from GET responses and, for a single-thread syncer, accept a round-trip PUT that changes the effective setting. This silently changes downstream constraint enforcement and can invalidate the operator's data-integrity assumptions.

**Evidence**
The new field is added without an API version or capability gate at `dm/openapi/spec/dm.yaml:1665`. `dm/master/openapi_view.go:46-84` forwards all requests to the elected leader, but the N-1 `TaskTargetDataBase` type at parent commit 496b57f3b has only host/password/port/security/user, and ordinary JSON binding ignores the unknown `session` property. The new tests at `dm/master/openapi_view_test.go:1022-1052` and `dm/syncer/syncer_test.go:2119-2177` exercise only new-master/new-worker behavior; no N-1 leader, rollback, or mixed-version fixture proves that the setting cannot be lost. The same OpenAPI spec already documents an all-masters-upgraded precondition for another unknown field at `dm/openapi/spec/dm.yaml:552-560`, but this field has no equivalent safeguard.

**Change request**
Is this backward compatible with N-1 masters? Add an explicit version/capability guard that rejects use until every master can preserve the field, or document and enforce an all-masters-upgraded precondition; also add mixed-version create, GET/PUT round-trip, leader-transfer, and rollback coverage so incompatibility fails with an actionable error rather than silently falling back to `0`.

dm/config/task_converters.go
```diff
@@ -719,11 +726,53 @@ func SubTaskConfigsToOpenAPITask(subTaskConfigList []*SubTaskConfig) *openapi.Ta
...
+		value = "1"
+	}
+	return &openapi.TaskTargetDataBase_Session{
+		AdditionalProperties: map[string]string{"foreign_key_checks": value},
+	}
+}
+
 // TaskConfigToOpenAPITask converts TaskConfig to an openapi task.
 func TaskConfigToOpenAPITask(c *TaskConfig, sourceCfgMap map[string]*SourceConfig) (*openapi.Task, error) {
+	if c.TargetDB != nil {
```
D3Hunter: #### ⚠️ **[Major]** The task converter now rejects previously valid session settings

**Why**
`TaskConfigToOpenAPITask` now sends every target session entry except `time_zone` through a whitelist that accepts only `foreign_key_checks`. This changes the existing `/api/v1/tasks/converters` contract from accepting valid DM task files to returning an error for established downstream session parameters.

**Scope**
dm/config/task_converters.go:761; POST /api/v1/tasks/converters when `task_config_file` contains non-OpenAPI target session variables

**Risk if unchanged**
Existing automation and upgrade workflows that convert otherwise valid task files can start failing after the binary upgrade. Common configurations using `sql_mode`, `tidb_skip_utf8_check`, `tidb_disable_txn_auto_retry`, or `tidb_retry_limit` become unconvertible even though DM still supports those settings in task YAML.

**Evidence**
Before this diff, `TaskConfigToOpenAPITask` proceeded directly to conversion; the new block at `dm/config/task_converters.go:761-773` returns on any session key other than `foreign_key_checks` or `time_zone`. Repository fixtures demonstrate supported legacy input, including `sql_mode` at `dm/tests/handle_error/conf/double-source-no-sharding.yaml:13` and several TiDB session variables at `dm/tests/all_mode/conf/dm-task.yaml:16`. The added unit test at `dm/config/task_converters_test.go:536-538` explicitly expects the new rejection, but there is no compatibility test or versioned converter contract covering existing task files.

**Change request**
Can we keep the converter backward compatible by separating conversion-time projection from create/update validation, so valid legacy session entries remain accepted and preserved in the returned task config while only the exposed `foreign_key_checks` field is projected into OpenAPI? Add an HTTP regression case with representative legacy session keys, or version the behavior and provide an explicit migration path if rejection is intentional.

D3Hunter (commented):
### Summary

- Total findings: 4
- Inline comments: 4

<details>

#### ⚠️ **[Major]** (4)

1. Session schema advertises keys and values that the API rejects (`dm/openapi/spec/dm.yaml:1665`)
2. The public contract does not say that the new session setting is immutable through PUT (`dm/openapi/spec/dm.yaml:1665 and dm/syncer/syncer.go:3420`)
3. N-1 masters silently discard the new target session field (`dm/openapi/spec/dm.yaml:1665; rolling-upgrade and rollback paths for POST/PUT /api/v1/tasks`)
4. The task converter now rejects previously valid session settings (`dm/config/task_converters.go:761; POST /api/v1/tasks/converters when `task_config_file` contains non-OpenAPI target session variables`)

</details>

dm/syncer/syncer.go
OliverS929: Good point. I initially kept them separate because projection needs to keep legacy tasks readable, while update comparison needs to reject duplicate keys. I’ve now centralized the shared target session normalization in config and reused it for both OpenAPI projection and update comparison in 4870e1ec2, which simplifies the code a bit.

joechenrh: > I initially kept them separate because projection needs to keep legacy tasks readable

Got it, I once thought if we can canonicalize the session when `TaskConfig` / `SubTaskConfig` is adjusted

dm/openapi/spec/dm.yaml
OliverS929: Good point. The original design represented `session` as a generic map to keep target session variables extensible as a group. Since `foreign_key_checks` is currently the only concrete supported case, the implementation separately restricted and validated the map to that key.

Given the current controlled support model, expressing the restriction directly in the OpenAPI schema is simpler. It makes the code responsibilities clearer and gives schema-driven clients the exact request contract, while future supported variables can still be added explicitly. In e7b795654, the schema now allows only `foreign_key_checks` with values `"0"` or `"1"`, and the redundant implementation-side validation has been removed.

OliverS929: `target_config.session` was not part of the existing OpenAPI hot-update surface. That boundary is enforced in `Syncer.CheckCanUpdateCfg`. `Syncer.Update` does not apply a new `To.Session` to the running syncer: it neither updates `s.cfg.To.Session` nor reinitializes the downstream connections. The explicit rejection added by this PR only prevents such an update from being accepted while silently left unapplied. The hot-update limitation itself comes from the existing DM update path and is not introduced by this PR.

The dmctl `update-task` path is already disabled as unsupported, and OpenAPI parameter changes will likewise be applied through task restart. Therefore, expanding PUT hot-update support for this field here would not add practical value. In e7b795654, I documented this limitation in the schema.

joechenrh (approved):

dm/openapi/spec/dm.yaml
OliverS929: OpenAPI currently has no mechanism to gate a field on all masters being upgraded. This PR keeps the existing behavior instead of introducing an out-of-scope capability mechanism, and adds a concise upgrade note to the schema in e7b795654.

dm/config/task_converters.go
OliverS929: Addressed by the change above. The restriction is now handled by the OpenAPI schema, so this concern no longer applies.

dm/openapi/gen.types.go
```diff
@@ -91,6 +91,13 @@ const (
 	TaskStageStopped TaskStage = "Stopped"
 )
 
+// Defines values for TaskTargetSessionForeignKeyChecks.
+const (
+	TaskTargetSessionForeignKeyChecksN0 TaskTargetSessionForeignKeyChecks = "0"
+
+	TaskTargetSessionForeignKeyChecksN1 TaskTargetSessionForeignKeyChecks = "1"
+)
+
```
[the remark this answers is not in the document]
OliverS929: This file was already regenerated from the updated OpenAPI schema. I reran the type generation and the resulting gen.types.go is identical, so there is no generated-code change to make here.

dm/syncer/syncer.go
```diff
@@ -3417,6 +3417,12 @@ func (s *Syncer) CheckCanUpdateCfg(newCfg *config.SubTaskConfig) error {
 	if err := s.checkForeignKeyCausalityConfigUpdate(newCfg); err != nil {
 		return err
 	}
+	if config.IsForeignKeyChecksEnabled(s.cfg.To.Session) != config.IsForeignKeyChecksEnabled(newCfg.To.Session) {
+		return terror.ErrWorkerUpdateSubTaskConfig.Generatef(
+			"can't update foreign_key_checks for syncer because it requires reinitialization, task: %s",
+			s.cfg.Name,
+		)
```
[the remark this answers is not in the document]
OliverS929: The map behavior is real, but duplicate keys are not reachable through the supported OpenAPI path. Previous OpenAPI tasks could not specify target session parameters, while the new schema accepts only the exact foreign_key_checks property and converts it to one canonical entry. The normal dmctl path also does not expose task updates. Reaching this case would require an already ambiguous dmctl configuration to be updated through OpenAPI, which is not a supported or known mixed-interface workflow, so I would prefer not to add PR-specific validation for it.

GMHDBJD (approved):

PR #12808, opened 2026-08-11
openapi(dm): support table schema from source and target

OliverS929:

Issue Number: ref #12756

DM OpenAPI requires callers to provide a `CREATE TABLE` statement when replacing a table schema, while dmctl can read the schema from the upstream source or routed downstream target. The handler also dereferences the optional `flush` and `sync` fields before applying their defaults.

- Add an optional `schema_source` field to load the table schema from the upstream source or routed downstream target.
- Keep caller-provided SQL as the default for compatibility with existing requests.
- Reuse the existing worker paths for reading schemas from the source and target.
- Safely default `flush` and `sync` to `true` while preserving explicit `false` values.
- Validate the relationship between `schema_source` and `sql_content`.

As before, `flush=false` does not make the update non-persistent, and `sync` is ignored outside optimistic sharding mode. These behaviors are not introduced by this PR, and changing them is out of scope.

#### Tests 

 - Unit test
 - Integration test

Manual testing with DM, MySQL 8.4, and TiDB nightly covered legacy SQL requests, omitted defaults and explicit false values, upstream and routed downstream schema sources, invalid requests, checkpoint persistence across restart, and continued incremental replication.

No. Existing SQL requests remain valid, and omitting `schema_source` preserves the existing SQL behavior. The new modes reuse the existing worker schema-operation paths.

No additional documentation update is needed. The request modes and existing `flush` and `sync` behavior are documented in the OpenAPI schema.

### Release note 

```release-note
Support loading DM table schemas from upstream sources and routed downstream targets through OpenAPI.
```

## Summary by CodeRabbit

- **New Features**
  - Table-structure operations now support SQL, upstream, and downstream schema sources.
  - Added clearer defaults for synchronization and flushing, with validation tailored to the selected source.
  - Updated API documentation and schema definitions to reflect the new options.

- **Bug Fixes**
  - Improved handling of optional operation settings and invalid request combinations.
  - Added configurable metadata removal when starting tasks.

- **Tests**
  - Expanded coverage for schema-source behavior, defaults, validation, task status transitions, and synchronization results.


OliverS929:
@joechenrh PTAL.

joechenrh (approved):

GMHDBJD (approved):

PR #12819, opened 2026-08-14
sqlmodel(dm): fix panic comparing binary key columns

### LINKED ISSUE #9834: DM worker panic when sync some tables
### What did you do?

```
panic: runtime error: comparing uncomparable type []uint8

goroutine 1734018212 [running]:
github.com/pingcap/tiflow/pkg/sqlmodel.(*RowChange).IsIdentityUpdated(0x2?)
        github.com/pingcap/tiflow/pkg/sqlmodel/reduce.go:106 +0x108
github.com/pingcap/tiflow/dm/syncer.genDMLsWithSameOp({0x4003862a80, 0x106, 0x267ad60?})
        github.com/pingcap/tiflow/dm/syncer/dml.go:466 +0x134
github.com/pingcap/tiflow/dm/syncer.(*DMLWorker).genSQLs(0x400ffb8150?, {0x4003862a80?, 0x3?, 0x3?})
        github.com/pingcap/tiflow/dm/syncer/dml_worker.go:285 +0x8c
github.com/pingcap/tiflow/dm/syncer.(*DMLWorker).executeBatchJobs(0x4002790780, 0x9, {0x4003862a80?, 0x106, 0x12c})
        github.com/pingcap/tiflow/dm/syncer/dml_worker.go:239 +0x110
github.com/pingcap/tiflow/dm/syncer.(*DMLWorker).executeJobs(0x4002790780, 0x9, 0x40620fd0e0)
        github.com/pingcap/tiflow/dm/syncer/dml_worker.go:184 +0x2c0
created by github.com/pingcap/tiflow/dm/syncer.(*DMLWorker).run
        github.com/pingcap/tiflow/dm/syncer/dml_worker.go:107 +0x5c
```

Here we may compare `[]byte`, should use bytes.Compare for this type.

https://github.com/pingcap/tiflow/blob/5d6aaa216ef8b8fc36eb2ec91680dc59b923d9c5/pkg/sqlmodel/reduce.go#L105

mengxian-li:

Issue Number: close #9834

`IsIdentityUpdated` and `IsPrimaryOrUniqueKeyUpdated` in `pkg/sqlmodel/reduce.go` compare column values with `!=` on an `interface{}`. Go panics when that interface holds a slice:

```
panic: runtime error: comparing uncomparable type []uint8

github.com/pingcap/tiflow/pkg/sqlmodel.(*RowChange).IsPrimaryOrUniqueKeyUpdated.func1(...)
	pkg/sqlmodel/reduce.go:111
github.com/pingcap/tiflow/dm/syncer.(*DMLWorker).genSQLs(...)
	dm/syncer/dml_worker.go:413
```

#9834 reported this for `IsIdentityUpdated` in 2023. Since then `IsPrimaryOrUniqueKeyUpdated`, added by #12351, repeated the same pattern, and that widened the blast radius considerably. The original report's stack goes through `genDMLsWithSameOp`, where the suggested workaround was `multiple-rows: false` — the default. The newer call site in `DMLWorker.genSQLs` has no equivalent gate:

```go
case sqlmodel.RowChangeUpdate:
    if j.safeMode {
        if j.dml.IsPrimaryOrUniqueKeyUpdated() {   // no gate
```

So on a build carrying #12351, every safe-mode UPDATE on an affected table panics deterministically. We hit this in production: DM workers entered a crash loop and could not recover, because safe mode was enabled by config for the task rather than as a transient window after start, so nothing self-healed. No data was lost — the workers die before they write — but replication stopped until the build was rolled back.

#### Which columns trigger it

It is narrower than "binary columns", which is likely why test coverage has missed it. Two kinds reach `sqlmodel` as `[]byte`:

| DDL | parser type | `isBinary` at `dm/syncer/dml.go` | value compared |
|---|---|---|---|
| `BINARY(n)`, binary `CHAR` | 254 `TypeString` | true → converted to `[]byte` | `[]byte` → panics |
| `BLOB`, `TEXT` | 252 | n/a — binlog decoder returns `[]byte` | `[]byte` → panics |
| `VARBINARY(n)` | 15 `TypeVarchar` | false | `string` → safe |
| `VARCHAR(n)` | 15 `TypeVarchar` | false | `string` → safe |

`adjustValues` converts a value to `[]byte` only when the column is `TypeString` **and** carries the binary flag, so `VARBINARY` parses to `TypeVarchar` and stays a `string`. GBK columns take the same conversion a few lines below, so they behave like the first row.

A single `BINARY(16)` primary key therefore faults on the very first comparison. That is the shape we saw.

Existing coverage misses it because `TestIdentityUpdatedWithUniqueKeys` uses `INT` keys only, and the `dm/tests/safe_mode_foreign_key` integration test uses `INT` primary keys with a `VARCHAR` unique key, which compares fine. No case puts a `TypeString`-with-binary-flag or `BLOB` column in a key.

- `pkg/sqlmodel/utils.go`: new `colValEqual`, next to `ColValAsStr`, which already type-switches on `[]byte`. It compares `[]byte` with `bytes.Equal` and falls back to `reflect.DeepEqual`, so no other uncomparable type can panic here either.
- `pkg/sqlmodel/reduce.go`: both comparison loops now route through one shared `colValsUpdated` helper. This also removes the duplicated closure inside `IsPrimaryOrUniqueKeyUpdated`, so the two functions cannot drift apart again.
- `pkg/sqlmodel/reduce_test.go`: `TestIdentityUpdatedWithBinaryKeys` covers a single `BINARY(16)` primary key, a multi-column primary key with trailing `BINARY` columns (so the fault lands on a later comparison), a secondary nullable `UNIQUE` index over a `BLOB` prefix that only `IsPrimaryOrUniqueKeyUpdated` reads, and a one-sided NULL. `VARBINARY` is covered with both `string` and `[]byte` values, so the comparison does not depend on the `adjustValues` type check staying as it is.

Values with equal contents but distinct backing arrays now compare equal, so a safe-mode UPDATE no longer reports a spurious key change and no longer writes a needless DELETE.

#### Tests

 - Unit test

The new test reproduces the panic against unmodified `master` (`comparing uncomparable type []uint8` at `reduce.go:91`) and passes with the fix. `pkg/sqlmodel` passes, `gofmt` and `go vet` are clean.

No compatibility change. The comparison runs per key column on safe-mode UPDATE. Non-`[]byte` values now go through `reflect.DeepEqual` rather than `!=`, which costs more per call. If that matters on this path, the fallback can be narrowed to a type switch over the concrete types DM decodes — happy to change it if a reviewer prefers that.

No.

This affects `release-8.5` as well, and `IsIdentityUpdated` has carried the bug since at least the 7.5 line, so cherry-picks are probably wanted.

### Release note

```release-note
Fix a DM worker panic ("comparing uncomparable type []uint8") when a table has BINARY, binary CHAR, BLOB, TEXT or GBK columns in its primary or unique key and safe mode is enabled.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved identity, primary-key, and unique-key change detection for binary values.
  * Values with identical byte content are now correctly recognized as unchanged, while differing content is detected as an update.
  * Improved handling of nullable binary values, composite keys, and supported string or byte representations.
  * Prevented comparison errors for previously unsupported value types.

### DIFF
--- a/pkg/sqlmodel/reduce.go
+++ b/pkg/sqlmodel/reduce.go
@@ -83,12 +83,17 @@ func (r *RowChange) IsIdentityUpdated() bool {
 
 	r.lazyInitWhereHandle()
 	pre, post := r.IdentityValues()
+	return colValsUpdated(pre, post)
+}
+
+// colValsUpdated returns true when any value of pre differs from post.
+func colValsUpdated(pre, post []any) bool {
 	if len(pre) != len(post) {
 		// should not happen
 		return true
 	}
 	for i := range pre {
-		if pre[i] != post[i] {
+		if !colValEqual(pre[i], post[i]) {
 			return true
 		}
 	}
@@ -102,26 +107,14 @@ func (r *RowChange) IsPrimaryOrUniqueKeyUpdated() bool {
 	}
 
 	r.lazyInitWhereHandle()
-	identityUpdated := func(pre, post []any) bool {
-		if len(pre) != len(post) {
-			// should not happen
-			return true
-		}
-		for i := range pre {
-			if pre[i] != post[i] {
-				return true
-			}
-		}
-		return false
-	}
 
 	if idx := r.whereHandle.UniqueNotNullIdx; idx != nil {
 		pre, post := r.identityValuesByIndex(idx, r.preValues, r.postValues)
-		if identityUpdated(pre, post) {
+		if colValsUpdated(pre, post) {
 			return true
 		}
 	} else {
-		if identityUpdated(r.preValues, r.postValues) {
+		if colValsUpdated(r.preValues, r.postValues) {
 			return true
 		}
 	}
@@ -147,7 +140,7 @@ func (r *RowChange) IsPrimaryOrUniqueKeyUpdated() bool {
 			continue
 		}
 		pre, post := r.identityValuesByIndex(idx, preValues, postValues)
-		if identityUpdated(pre, post) {
+		if colValsUpdated(pre, post) {
 			return true
 		}
 	}
--- a/pkg/sqlmodel/reduce_test.go
+++ b/pkg/sqlmodel/reduce_test.go
@@ -72,6 +72,125 @@ func TestIdentityUpdatedWithUniqueKeys(t *testing.T) {
 	require.True(t, change.IsPrimaryOrUniqueKeyUpdated())
 }
 
+// TestIdentityUpdatedWithBinaryKeys covers key columns that reach sqlmodel as
+// []byte. Applying != to an interface that holds a slice panics with
+// "comparing uncomparable type []uint8", so equality has to go through
+// bytes.Equal. Values with equal contents but distinct backing arrays must
+// compare equal, otherwise a safe-mode UPDATE reports a spurious key change and
+// writes a needless DELETE.
+//
+// Two column kinds reach here as []byte:
+//   - BINARY and binary CHAR, which adjustValues converts from string because
+//     the column is TypeString with the binary flag set;
+//   - BLOB and TEXT, which the binlog decoder already returns as []byte.
+//
+// A GBK column takes the same conversion, so it behaves like the first kind.
+func TestIdentityUpdatedWithBinaryKeys(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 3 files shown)


mengxian-li:
I've signed the CLA

PR #12848, opened 2026-09-04
[DNM] kafka: bump sarama dependency to v1.60.1 and do not backoff on the first time error handling

3AceShowHand:

Issue Number: close #xxx

#### Tests 

 - Unit test
 - Integration test
 - No code

### Release note 

```release-note

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Retry sequences now begin immediately, while subsequent retries use randomized backoff intervals.
  * Retry backoff resets correctly after an extended period without errors.
  * Capture ownership reliably tracks peer additions and removals, including without processor activity.
  * Waiting operations now promptly return cancellation and deadline errors.
  * Kafka sink recovery is verified after an injected failure is cleared.

* **Chores**
  * Updated Kafka client and supporting module dependencies.
  * Removed obsolete dependency references and adjusted Kafka retry configuration documentation.


pkg/retry/error_retry.go
```diff
@@ -97,6 +98,10 @@ func (r *ErrorRetry) GetRetryBackoff(err error) (time.Duration, error) {
 
 	r.lastInternalError = err
 	r.lastErrorRetryTime = time.Now()
+	if isFirstRetry {
+		// Retry immediately when a new series of errors starts.
+		return 0, nil
+	}
```
Tema, line 104: Likely new CI flake: tests/integration_tests/kafka_sink_error_resume/run.sh

This test injects a one-shot error (KafkaSinkAsyncSendError=1*return(true)) and then polls check_changefeed_status ... "warning" "last_warning" "kafka sink injected error". In sinkmanager/manager.go the warning is sent before util.Hang(backoff), so the window during which the changefeed stays in warning was exactly this backoff (>= 5s). With return 0, nil the factory is rebuilt immediately, the checkpoint advances within an owner tick, and feed_state_manager.checkAndChangeState flips the state back to normal (and the v2 API then omits last_warning too). ensure probes at t=0, 2, 6, 12s..., so a sub-second warning window can be missed entirely, and because the failpoint fires once the window never reopens: the test burns all retries and fails. The script's own comment already warns: "if owner retry changefeed frequently, it may break the test."

changefeed_error/run.sh is not affected (persistent InjectChangefeedDDLError=return(true), so the second error still gets 5-30s). Suggest running kafka_sink_error_resume on this branch before merge and, if it flakes, switching the failpoint to 2*return(true) or asserting on last_warning history instead of the transient state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

pkg/retry/error_retry.go
```diff
@@ -97,6 +98,10 @@ func (r *ErrorRetry) GetRetryBackoff(err error) (time.Duration, error) {
 
 	r.lastInternalError = err
 	r.lastErrorRetryTime = time.Now()
+	if isFirstRetry {
+		// Retry immediately when a new series of errors starts.
+		return 0, nil
```
Tema, line 103: util.Hang(ctx, 0) is no longer a reliable cancellation point

Both callers feed this straight into util.Hang(ctx, backoff) (ddl_sink.go:189, sinkmanager/manager.go:327). Hang (pkg/util/time.go:22) is timer := time.NewTimer(dur); select { case <-ctx.Done(): ...; case <-timer.C: ... }. With dur == 0 the timer channel is already populated when select polls readiness (Go 1.23+ runs a due timer in selectgo before building pollorder), so if ctx is already cancelled both cases are ready and the pick is ~50/50. Previously backoff >= 5s guaranteed ctx.Done() won.

Effect: ~half of shutdowns/pauses that race a sink error run one extra loop iteration on a dead ctx. In ddl_sink.go that re-runs makeSinkReady -> factory.New(cancelledCtx), and because makeSinkReady returns errors.New("ddlSink not ready") (drops the cause), errors.Cause(err) != context.Canceled is true and s.reportWarning(err) fires for an ordinary stop. In manager.go the loop re-enters initSinkFactory() before the managerCtx.Done() select, so a fresh factory is dialed during teardown. Bounded to one iteration (the second GetRetryBackoff in the series returns >= 5s), so severity is low, but cheap to fix in Hang:

func Hang(ctx context.Context, dur time.Duration) error {
	if err := ctx.Err(); err != nil {
		return err
	}
	...

🤖 Generated with Claude Code

